PR #36211, opened 2023-01-18
fix(dav): catch MaxInstancesExceededException for recurring events without end date

miaulalala:
* Resolves part of https://github.com/nextcloud/calendar/issues/4811

## Summary

The VEVENT iterator throws a `MaxInstancesExceededException` on recurring events that do not have an end date. This exception should be caught and logged as to not break the reminder service.

## Checklist

- [ ] ~~Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration]~~(https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included

### DIFF
--- a/apps/dav/lib/CalDAV/Reminder/ReminderService.php
+++ b/apps/dav/lib/CalDAV/Reminder/ReminderService.php
@@ -482,49 +482,54 @@ class ReminderService {
 			return;
 		}
 
-		while ($iterator->valid()) {
-			$event = $iterator->getEventObject();
+		try {
+			while ($iterator->valid()) {
+				$event = $iterator->getEventObject();
 
-			// Recurrence-exceptions are handled separately, so just ignore them here
-			if (\in_array($event, $recurrenceExceptions, true)) {
-				$iterator->next();
-				continue;
-			}
-
-			$recurrenceId = $this->getEffectiveRecurrenceIdOfVEvent($event);
-			if ($reminder['recurrence_id'] >= $recurrenceId) {
-				$iterator->next();
-				continue;
-			}
-
-			foreach ($event->VALARM as $valarm) {
-				/** @var VAlarm $valarm */
-				$alarmHash = $this->getAlarmHash($valarm);
-				if ($alarmHash !== $reminder['alarm_hash']) {
+				// Recurrence-exceptions are handled separately, so just ignore them here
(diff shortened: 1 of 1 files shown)


apps/dav/lib/CalDAV/Reminder/ReminderService.php
```diff
@@ -482,49 +482,59 @@ private function deleteOrProcessNext(array $reminder,
...
+				if ($reminder['recurrence_id'] >= $recurrenceId) {
+					$iterator->next();
+					continue;
+				}
 
-				// If effective trigger time is in the past
-				// just skip and generate for next event
-				$diff = $now->diff($triggerTime);
-				if ($diff->invert === 1) {
+				if(!isset($event->VALARM)) {
```
ChristophWurst, line 501: why is this check necessary?

miaulalala: https://github.com/nextcloud/server/security/code-scanning/18339

ChristophWurst: the code was there before and worked. I think Psalm is just unable to derive the type of the VALARM component

https://github.com/nextcloud/server/pull/36211/files?w=1 for a clean diff

tcitworld: If the event doesn't have a VALARM that means we have a logic issue somewhere else, as all events and occurrences here should have some. If it doesn't have one it's better to be informed of the issue so we can try to fix it instead of ignoring it, even though that makes psalm happy.

miaulalala: but psalm isn't wrong, the VALARM property could be NULL.

tcitworld: It could theoretically, for events without reminders, but here it *shouldn't*. As I said, if we end up here with an event with no reminder, something else is wrong.

tcitworld (approved):

ChristophWurst (approved):
:+1: https://github.com/nextcloud/server/pull/36211/files?w=1

PR #36214, opened 2023-01-18
Specify the parameters of FederatedFileSharing controllers

nickvergessen:
## Summary

In order to get closer to OpenAPI all controllers should be up to the state of art how to defining the parameters so they are documented.

## Checklist

### DIFF
--- a/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
+++ b/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
@@ -33,7 +33,6 @@ use OCA\FederatedFileSharing\Notifications;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\OCS\OCSBadRequestException;
 use OCP\AppFramework\OCS\OCSException;
-use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\AppFramework\OCSController;
 use OCP\App\IAppManager;
 use OCP\Constants;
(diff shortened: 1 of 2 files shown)


provokateurin:
Shouldn't the parameters also be documented in the comment with explanations?

nickvergessen:
> Shouldn't the parameters also be documented in the comment …

Did that, just repushed because I forgot it on the first method

> … with explanations?

Not sure I'm fully qualified to provide this.

provokateurin:
Explanations could be copied from https://github.com/cs3org/OCM-API/blob/develop/spec.yaml

provokateurin:
It also doesn't matter too much right now, I'd just copy them later

apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -438,18 +440,20 @@ protected function ncPermissions2ocmPermissions($ncPermissions) {
...
+	 * @param string|null $remote
+	 * @param string|null $remote_id
 	 * @return Http\DataResponse
-	 * @throws \InvalidArgumentException
+	 * @throws OCSBadRequestException
 	 * @throws OCSException
+	 * @throws \OCP\DB\Exception
 	 */
-	public function move($id) {
+	public function move(int $id, ?string $token = null, ?string $remote = null, ?string $remote_id = null) {
```
provokateurin, line 451: psalm complains remote can't be null so i guess it should not be nullable here?

provokateurin (changes requested):

apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
nickvergessen: Previous controller code allowed it, but yeah it would have broken on that call then.
So save to assume it never was called with that.
Actually I assume all of the parameters were mandatory. It was just not enforce and defaulted to null and I'm unsure if anything else would break anything.

provokateurin: if it would fail with null values anyway, why not move it to an earlier stage where it fails before the method is even called

nickvergessen: I guess we could copy the types from https://github.com/cs3org/OCM-API/blob/develop/spec.yaml ?
Because that is what others would send anyway?

provokateurin: No, this API isn't implementing OCM (see discussion in the chat group)

nickvergessen:
Closing for now as it's the wrong endpoint

provokateurin:
This is relevant now. I reopened because I would have just created the same PR again.

provokateurin (approved):

ChristophWurst (approved):

Issue #25283, opened 2021-01-22
Nextcloud displays wrong used storage


olizimmermann:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Steps to reproduce
1. Install Nextcloud
2. Create some users
3. Upload some data 
4. check at the users overview the used storage
5. compare it with the real used storage within the users folder via shell

### Expected behaviour
Tell us what should happen
Should be 1:1 like all my other users.
### Actual behaviour
Tell us what happens instead
Real used storage: 9.3 Gb
Nextcloud shows me: 18.2
### Server configuration

**Operating system:**
FreeBsd
**Web server:**
Nginx
**Database:**

**PHP version:**
7.04
**Nextcloud version:** (see Nextcloud admin page)
20.0.5
**Updated from an older Nextcloud/ownCloud or fresh install:**

**Where did you install Nextcloud from:**
TrueNas - Plugin - latest Update via Updater
**Signing status:**
<details>
<summary>Signing status</summary>

```
Login as admin user into your Nextcloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
```
No errors have been found.
</details>

**List of activated apps:**
<details>
<summary>App list</summary>
Notes, Decks, Tasks

```
If you have access to your command line run e.g.:
sudo -u www-data php occ app:list
from within your Nextcloud installation folder
```
---
</details>

**Nextcloud configuration:**
<details>
<summary>Config report</summary>

```
If you have access to your command line run e.g.:
sudo -u www-data php occ config:list system
from within your Nextcloud installation folder

or 

Insert your config.php content here. 
Make sure to remove all sensitive content such as passwords. (e.g. database password, passwordsalt, secret, smtp password, …)
```

CONFIG = array (
  'apps_paths' =>
  array (
    0 =>
    array (
      'path' => '/usr/local/www/nextcloud/apps',
      'url' => '/apps',
      'writable' => true,
    ),
    1 =>
    array (
      'path' => '/usr/local/www/nextcloud/apps-pkg',
      'url' => '/apps-pkg',
      'writable' => true,
    ),
  ),
  'logfile' => '/var/log/nextcloud/nextcloud.log',
[...]

PR #36220, opened 2023-01-18
Add maintenance:repair step for file size of unencrypted files

simonspa:
* Related: #25283

## Summary

This resets the `unencrypted_size` of files with `encryption = 0` to `0`. As described in the linked issue at length, there seem to be setups that either have or had encryption enabled in the past, where even for unencrypted files the `unencrypted_size` column shows non-zero values.

These seem to completely confuse the quota calculation and result in partially unusable systems because users might have a huge quota usage reported while the actual used size is just a few MB.

## TODO

- [ ] Add tests

## Checklist

### DIFF
--- a/lib/private/Repair/RepairUnencryptedSize.php
+++ b/lib/private/Repair/RepairUnencryptedSize.php
@@ -38,18 +38,17 @@ class RepairUnencryptedSize implements IRepairStep {
 		return 'Repair file size of unencrypted files';
 	}
 
-    /**
+	/**
 	 * Fix unencrypted file size
 	 */
 	public function run(IOutput $output) {
-
-        $update = $this->connection->getQueryBuilder();
+		$update = $this->connection->getQueryBuilder();
 		$update->update('filecache')
 			->set('unencrypted_size', "0")
-			->where($update->expr()->eq('encrypted',"0", IQueryBuilder::PARAM_INT));
+			->where($update->expr()->eq('encrypted', "0", IQueryBuilder::PARAM_INT));
 
-        if ($update->execute()) {
-            $output->info('Fixed file size of unencrypted files');
-        }
+		if ($update->execute()) {
+			$output->info('Fixed file size of unencrypted files');
+		}
 	}
 }
--- a/lib/private/Repair/RepairUnencryptedSize.php
+++ b/lib/private/Repair/RepairUnencryptedSize.php
@@ -45,8 +45,8 @@ class RepairUnencryptedSize implements IRepairStep {
 		$update = $this->connection->getQueryBuilder();
 		$update->update('filecache')
 			->set('unencrypted_size', "0")
-			->where($update->expr()->eq('encrypted', "0", IQueryBuilder::PARAM_INT))
-			->andWhere($update->expr()->neq('unencrypted_size', "0", IQueryBuilder::PARAM_INT));
+			->where($update->expr()->gt('unencrypted_size', "0", IQueryBuilder::PARAM_INT))
+			->andWhere($update->expr()->eq('encrypted', "0", IQueryBuilder::PARAM_INT));
 
 		if ($update->execute()) {
 			$output->info('Fixed file size of unencrypted files');
--- a/lib/private/Repair/RepairUnencryptedSize.php
+++ b/lib/private/Repair/RepairUnencryptedSize.php
@@ -1,0 +1,53 @@
+<?php
+/**
+ * @copyright Copyright (c) 2023, Nextcloud GmbH.
+ *
+ * @author Simon Spannagel <simonspa@kth.se>
+ *
+ * @license AGPL-3.0
+ *
+ * This code is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License, version 3,
+ * as published by the Free Software Foundation.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


lib/private/Repair/RepairUnencryptedSize.php
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	/**
+	 * Fix unencrypted file size
+	 */
+	public function run(IOutput $output) {
+		$update = $this->connection->getQueryBuilder();
+		$update->update('filecache')
+			->set('unencrypted_size', "0")
+			->where($update->expr()->eq('encrypted', "0", IQueryBuilder::PARAM_INT));
```
PVince81, line 48: I wonder if we should add another `where unencrypted_size > 0` because I don't know if the current query will cause perf issues on big setups and it would make it select less rows to update ?

@icewind1991 @juliushaertl do you know ?

PVince81 (approved):
:+1: from me, I hope the second reviewer knows more about DB perf

good luck appeasing the bots :smile: 
thanks a lot!

lib/private/Repair/RepairUnencryptedSize.php
miaulalala, line 48:
```suggestion
			->where($update->expr()->eq('encrypted', "0", IQueryBuilder::PARAM_INT)
			->andWhere($update->expr()->neq('unencrypted_size', "0", IQueryBuilder::PARAM_INT);
```

maybe? so you don't touch rows that are already correct.

I would also wrap the whole query in a batch.

simonspa: Implemented, doesn't hurt. :)

szaimen:
Drone failures look related

lib/private/Repair/RepairUnencryptedSize.php
```diff
@@ -0,0 +1,55 @@
...
+
+	/**
+	 * Fix unencrypted file size
+	 */
+	public function run(IOutput $output) {
+		$update = $this->connection->getQueryBuilder();
+		$update->update('filecache')
+			->set('unencrypted_size', "0")
+			->where($update->expr()->eq('encrypted', "0", IQueryBuilder::PARAM_INT))
+			->andWhere($update->expr()->neq('unencrypted_size', "0", IQueryBuilder::PARAM_INT));
```
juliushaertl, line 49: There is no index on neither of those columns, so we may want to add at least to the unencrypted size.

I think the most safe way to use such an index then would be:

```suggestion
			->andWhere($update->expr()->gt('unencrypted_size', "0", IQueryBuilder::PARAM_INT));
			->andWhere($update->expr()->eq('encrypted', "0", IQueryBuilder::PARAM_INT))
```

juliushaertl: For the reasoning:
- neq may not use the index
- the order of where statements sometimes also matters to decide if the index is applied

juliushaertl: EXPLAIN of the current query does a full table scan:

```

explain update oc_filecache set unencrypted_size = 0 where encrypted = 0 and unencrypted_size != 0;
+------+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
| id   | select_type | table        | type  | possible_keys | key     | key_len | ref  | rows | Extra       |
+------+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
|    1 | SIMPLE      | oc_filecache | index | NULL          | PRIMARY | 8       | NULL | 8479 | Using where |
+------+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
```

simonspa: I don't see how the index is applied after reversal of arguments and using `>`:

```
explain update oc_filecache set unencrypted_size = 0 where unencrypted_size > 0 and encrypted = 0;
+------+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
| id   | select_type | table        | type  | possible_keys | key     | key_len | ref  | rows | Extra       |
+------+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
|    1 | SIMPLE      | oc_filecache | index | NULL          | PRIMARY | 8       | NULL | 490  | Using where |
+------+-------------+--------------+-------+---------------+---------+---------+------+------+-------------+
```

simonspa: I implemented the suggested change - but adding an index to either of these columns would have to done separately, and probably run manually, right? How would that work together with an automatic repair step?

simonspa:
Performance tests is failing with

```
fatal: remote error: upload-pack: not our ref 03c3817ff132653c794fd04410977952f69fd614
```

- is this owing to the fact that this pull comes from my fork rather than a branch within this repo?

sorbaugh:
@simonspa this PR looks nice :) I'm reopening in the hopes that you're still interested in pushing this forward! Since you're a member of Nextcloud, could you recreate this PR in a branch directly in the nextcoud/server repo? Feel free and encouraged to check the Nextcloud Commit guidelines. Hope we can keep this cool improvement alive! 🚀

joshtrichards:
I wonder if instead of another repair step, we integrated into our existing scanner. The scanner-based approach gives us a path to cover all relevant scenarios consistently, including the stale `unencrypted_size` cases this PR is targeting, and it helps the system converge naturally on correct values instead of relying on a one-off DB cleanup. We can still keep that broader repair behavior behind a feature flag for now if we want to limit rollout risk and make it opt-in initially.

PR #36221, opened 2023-01-18
Use current git branch by default when setting up cypress container

artonge:
This allow to tests changes that contains migrations.

### DIFF
--- a/cypress/dockerNode.ts
+++ b/cypress/dockerNode.ts
@@ -26,6 +26,7 @@
 import Docker from 'dockerode'
 import waitOn from 'wait-on'
 import tar from 'tar'
+import { execSync } from 'child_process'
 
 export const docker = new Docker()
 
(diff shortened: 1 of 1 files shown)


szaimen (approved):
🐘

cypress.config.ts
```diff
@@ -76,7 +77,7 @@ export default defineConfig({
 
 			// Before the browser launches
 			// starting Nextcloud testing container
-			return startNextcloud(process.env.BRANCH)
+			return startNextcloud(process.env.BRANCH || execSync('git rev-parse --abbrev-ref HEAD').toString().trim())
```
skjnldsv, line 80: Ha, so, I wanted to see if we could do something like that, it's fancy, I like it.
Not fond of this big exec straight in the function call, can we make this a bit nicer? Maybe put this into the dockerNode script straight away?

skjnldsv (approved):
Awesome! Works!

Pytal (approved):

PR #35299, opened 2022-11-21
Prevents the selection of encrypted folders when clicking on the "Select all" checkbox

TSI-kavitasonawane:

PR #36223, opened 2023-01-18
Allow interaction with encrypted folders

artonge:
This change improve on #35299 by only blocking interactions with the checkbox and not the whole entry


szaimen (approved):
LGTM but didnt test

PVince81 (approved):
:+1: saw the fix working in a screen share

Pytal (approved):

Issue #36201, opened 2023-01-18
[BITV] User menu must be focus trapped


Pytal:
- [x] #36232
- [x] https://github.com/nextcloud/server/pull/36668

PR #36232, opened 2023-01-19
Port user menu to Vue

Pytal:
* For https://github.com/nextcloud/server/issues/36201

## Summary

Move PHP user menu markup to Vue wrapped by the focus trapped NcHeaderMenu component

Before | After
--- | ---
image | image

## Requires
- ~https://github.com/nextcloud/server/pull/36449~
- [x] https://github.com/nextcloud/server/pull/36508

## Related
- https://github.com/nextcloud/firstrunwizard/pull/815
- https://github.com/nextcloud/firstrunwizard/pull/816

## Checklist


core/src/views/UserMenu.vue
```diff
@@ -0,0 +1,174 @@
...
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+-->
+
+<template>
+	<NcHeaderMenu id="settings"
+		class="user-menu"
+		:aria-label="t('core', 'Open user menu')">
```
nickvergessen, line 26: This was:
```suggestion
		:aria-label="t('core', 'Open settings menu')">
```

not sure which is better.

nickvergessen (commented):
Not breaking the external sites app, if that is why my review was requested.
From the technical changes in this PR I'm the wrong person to judge

lib/private/TemplateLayout.php
```diff
@@ -131,6 +131,7 @@ public function __construct($renderAs, $appId = '') {
 			$this->assign('navigation', $navigation);
 			$settingsNavigation = $this->navigationManager->getAll('settings');
 			$this->assign('settingsnavigation', $settingsNavigation);
+			$this->initialState->provideInitialState('core', 'settingsNavEntries', $settingsNavigation);
```
juliushaertl, line 134: Do we still need the above `settingsnavigation` template variable?

core/templates/layout.user.php
```diff
@@ -70,45 +70,7 @@
...
-								<?php if ($entry["active"]): ?> class="active"<?php endif; ?>>
-								<img alt="" src="<?php print_unescaped($entry['icon'] . '?v=' . $_['versionHash']); ?>">
-								<?php p($entry['name']) ?>
-							</a>
-						</li>
-					<?php endforeach; ?>
-					</ul>
-					</nav>
-				</div>
+				<div id="settings"></div>
```
juliushaertl, line 73: Maybe we could give this a more descriptive id in the same run?

apps/user_status/src/menu.js
```diff
@@ -24,42 +24,31 @@
...
-	preloadedUserStatus: {
-		message: userStatusData.message,
-		icon: userStatusData.icon,
-		status: userStatusData.status,
-	},
-	user: avatarDiv.dataset.user,
-	displayName: avatarDiv.dataset.displayname,
-	disableMenu: true,
-	disableTooltip: true,
+const mountMenu = () => {
```
juliushaertl, line 39: Can we be fully sure that the UserMenu has already been rendered at this point? Otherwise this might need some way of handling to not end up with a menu without the profile link which might become hard to debug later on.

juliusknorr (commented):
Just some smaller comments, but looks good otherwise 👍

lib/private/TemplateLayout.php
nickvergessen: Wondered the same, but I guess we do for other templates.

juliushaertl: `layout.user.php` seems to be the only occurrence

core/src/views/UserMenu.vue
Pytal: Not sure either but thinking was that the menu contains more than just settings, @nimishavijay @jancborchardt what should we label this?

lib/private/TemplateLayout.php
Pytal: Removed!

core/templates/layout.user.php
Pytal: Now `id="user-menu"`

apps/user_status/src/menu.js
Pytal: Should be as this is only executed when UserMenu has been mounted and emits `core:user-menu:mounted`

This only works as long as `apps/user_status/src/menu.js` is loaded before `core/src/components/UserMenu.js`, but if `UserMenu.js` loads and emits `core:user-menu:mounted` before `menu.js` can subscribe to the event then it wouldn't

For this I adjusted the logic to handle both cases

apps/user_status/src/menu.js
```diff
@@ -24,42 +24,35 @@
...
-	displayName: avatarDiv.dataset.displayname,
-	disableMenu: true,
-	disableTooltip: true,
-}
+const mountPoint = document.querySelector('li[data-id="user_status-menu-entry"]')
 
-const NcAvatarInMenu = Vue.extend(NcAvatar)
-new NcAvatarInMenu({ propsData }).$mount('#avatardiv-menu')
+const mountMenuEntry = () => {
+	const mountPoint = document.querySelector('li[data-id="user_status-menu-entry"]')
```
skjnldsv, line 42: `data-id` ?

This doesn't seem like a valid selector to use for a mountpoint.
a proper `id=` should be used. Data selectors have a different purpose

skjnldsv, line 42: `data-id` ?

This doesn't seem like a valid selector to use for a mountpoint.
a proper `id=` should be used. Data selectors have a different purpose

core/src/views/UserMenu.vue
```diff
@@ -0,0 +1,174 @@
...
+	<NcHeaderMenu id="user-menu"
+		class="user-menu"
+		:aria-label="t('core', 'Open user menu')">
+		<template #trigger>
+			<NcAvatar class="user-menu__avatar"
+				:disable-menu="true"
+				:disable-tooltip="true"
+				:user="userId" />
+		</template>
+		<nav class="user-menu__nav">
```
skjnldsv, line 33: Missing `aria-label` https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role

apps/user_status/src/menu.js
skjnldsv: I just saw that was what was used in the first place... :see_no_evil: 
Fee free to adjust this

skjnldsv: While it's probably what everyone else is using, (like [firstrunwizard](https://github.com/nextcloud/firstrunwizard/blob/4ce69aef027411d52fba85eb63d46cde28cca5f1/js/about.js#L2)), they are also using various obsolete selectors like `#expanddiv`. So this PR is a breaking change anyway.
I would fall back and adjust to a simple `id`

skjnldsv (changes requested):

core/src/views/UserMenu.vue
skjnldsv: If no reply from the design team, the original strings should be used.
Keep in mind there is also admin entries in here. `user menu` might also be misleading

apps/user_status/src/menu.js
Pytal: Using `id` now with required https://github.com/nextcloud/firstrunwizard updates:
- https://github.com/nextcloud/firstrunwizard/pull/815
- https://github.com/nextcloud/firstrunwizard/pull/816

skjnldsv (dismissed):

szaimen:
I tested this together with https://github.com/nextcloud/firstrunwizard/pull/816 and clicking on `Personal settings` leads to internal server error. Also the firstrunwizard entry is in the list somehow not shown...

<details>
<summary>Click to show</summary>

image

</details>
<details>
<summary>Tested with </summary>

```
docker run -it \
--name nextcloud-easy-test \
-p 127.0.0.1:8443:443 \
-e SERVER_BRANCH=enh/a11y-user-menu \
-e FIRSTRUNWIZARD_BRANCH=enh/a11y-user-menu-id \
```

</details>

core/src/views/UserMenu.vue
nimishavijay: Agreed with @skjnldsv that it could be misleading, also we're trying not to use the wording "user" so "settings menu" is probably a better fit anyway

szaimen: For the admin, it is personal settings...

Pytal:
Rebased

szaimen (approved):
Tested together with https://github.com/nextcloud/firstrunwizard/pull/816 and works now but didnt review the code.

(Looks like the internal server issue was somehow caused by the viewer app not being installed).

artonge (approved):

PR #36257, opened 2023-01-19
non-static app_dir

### LINKED ISSUE #170: Recent upgrade broke Nextcloud
After the recent upgrade a few hours ago I no longer could access my Nextcloud instance (AIO on Ubuntu 20 LTS). The message "Update needed" greeted me on the start screen. I disabled Related resouces via occ command, and Nextcloud started working again

### LINKED ISSUE #18673: Reset app dir on upgrade with non-writable apps paths.
Snap community is holding back the upgrade to 16.0.6. They experience a second upgrade screen after the actual upgrade. 

Screenshot from 2020-01-05 15-07-25

### Steps to reproduce
1. git checkout v16.0.5 (it should also affect newer versions)
2. Add a second app path to config.php and make the default one `writable => false`
```
	'apps_paths' => [
		[
			'path' => 'nextcloud/apps',
			'url' => '/apps',
			'writable' => false,
		],
		[
			'path' => 'nextcloud/extra-apps',
			'url' => '/extra-apps',
			'writable' => true,
		],
	],
```
3. Put `files_rightclick` (0.13.0 from app store https://github.com/nextcloud/files_rightclick/releases/download/v0.13.0/files_rightclick.tar.gz) to `nextcloud/apps`.
4. Run setup
5. git checkout v16.0.7
6. Run upgrade

### Expected behaviour
No second upgrade screen

### Actual behaviour
1. Upgrade will check app store for updates.
2. There is an update for files_rightclick available.
3. `nextcloud/apps` is not writable. App will be installed to `nextcloud/extra-apps`.
4. [`OC_App::updateApp($appId);`](https://github.com/nextcloud/server/blob/f021db956c1e0c9da4a9829dce0d27d1ba45ecef/lib/private/Installer.php#L186) where $appId is `files_rightclick`.
5. [`OC_App::getAppPath($appId);`](https://github.com/nextcloud/server/blob/139e45f3602ea97af2f290a38ec40ec0addcfda0/lib/private/legacy/app.php#L897)
6. [`OC_App::findAppInDirectories($appId);`](https://github.com/nextcloud/server/blob/139e45f3602ea97af2f290a38ec40ec0addcfda0/lib/private/legacy/app.php#L499)

https://github.com/nextcloud/server/blob/139e45f3602ea97af2f290a38ec40ec0addcfda0/lib/private/legacy/app.php#L451-L455

7. Step 6 should actually check both app directories for apps and return the latest one. But above cache already has a entry for files_rightclick (the one from nextcloud/apps which is still 0.13.0). 

8. Run upgrade again. No cache for files_rightclick and the right one is picked from nextcloud/extra-apps).

Also the documentation for `getAppPath` and `getAppWebPath` seems to be wrong

> If the app is defined in multiple directories, the first one is taken. (false if not found)

Its actually the newest version. Or the documentation is right and the code is wrong :confused: 

**I'm a bit lost. How can we fix this in a backportable way?**  

A) Make $app_dir a static class member of OC_App and add a method to unset the cache for a appId

B) Introduce a $useCache value similar to `getAppVersion` and pass it around.

C) You tell me ;)

cc @nickvergessen @georgehrke @rullzer @juliushaertl thanks in advance :+1: 

cc @kyrofa because you initially reported this https://github.com/nextcloud/server/issues/17338#issuecomment-555779273.

ArtificialOwl:
Fix https://github.com/nextcloud/related_resources/issues/170
Close https://github.com/nextcloud/server/issues/18673

Should fix https://github.com/nextcloud/all-in-one/discussions/1787#discussioncomment-4710650

There is an edge use case where custom apps folder seems ignored during server upgrade.

When upgrading the server:

- version of apps included with server (bundle) are compared with appstore,
- in case of new versions, last available version is downloaded,
- if custom_apps is used, the downloaded app will be installed in the writable app folder (and not the default app folder),
- unfortunately, upgrade of apps installed in custom_apps are not triggered during this process,
- upgrade finalize and an admin will need to start a second upgrading process to trigger the upgrade of apps downloaded in custom_apps.

While it seems a minor issue because it only require to initiate a second upgrading process, it locks **AIO Nextcloud** because the upgrade cannot be initiated from the webclient (but only by running `./occ upgrade`)

This PR will trigger the upgrading process of the apps installed in custom apps during the upgrading process of the server.

### DIFF
--- a/lib/private/legacy/OC_App.php
+++ b/lib/private/legacy/OC_App.php
@@ -477,16 +477,17 @@ class OC_App {
 	 * search for an app in all app-directories
 	 *
 	 * @param string $appId
+	 * @param bool $ignoreCache ignore cache and rebuild it
 	 * @return false|string
 	 */
-	public static function findAppInDirectories(string $appId) {
+	public static function findAppInDirectories(string $appId, bool $ignoreCache = false) {
 		$sanitizedAppId = self::cleanAppId($appId);
 		if ($sanitizedAppId !== $appId) {
 			return false;
 		}
 		static $app_dir = [];
 
-		if (isset($app_dir[$appId])) {
+		if (isset($app_dir[$appId]) && !$ignoreCache) {
 			return $app_dir[$appId];
 		}
 
@@ -527,15 +528,16 @@ class OC_App {
 	 * @psalm-taint-specialize
 	 *
 	 * @param string $appId
+	 * @param bool $refreshAppPath should be set to true only during install/upgrade
 	 * @return string|false
 	 * @deprecated 11.0.0 use \OC::$server->getAppManager()->getAppPath()
 	 */
-	public static function getAppPath(string $appId) {
+	public static function getAppPath(string $appId, bool $refreshAppPath = false) {
 		if ($appId === null || trim($appId) === '') {
 			return false;
 		}
 
-		if (($dir = self::findAppInDirectories($appId)) != false) {
+		if (($dir = self::findAppInDirectories($appId, $refreshAppPath)) != false) {
 			return $dir['path'] . '/' . $appId;
 		}
 		return false;
@@ -973,7 +975,9 @@ class OC_App {
 	 * @return bool
 	 */
 	public static function updateApp(string $appId): bool {
-		$appPath = self::getAppPath($appId);
+		// for apps distributed with core, we refresh app path in case the downloaded version
+		// have been installed in custom apps and not in the default path
+		$appPath = self::getAppPath($appId, true);
 		if ($appPath === false) {
 			return false;
 		}


szaimen (approved):
This seems to fix the issue in my testing. Thanks a lot for fixing this @ArtificialOwl! :)

However I am not sure if this has some negative performance implications...

blizzz:
I would not say it is the final fix, for it may have performance impact (not 100% sure, but otherwise the static var would not make sense). Alternatively it could be an override parameter, for instance, that would avoid the runtime cache in update scenarios.

ArtificialOwl:
> I would not say it is the final fix, for it may have performance impact (not 100% sure, but otherwise the static var would not make sense). Alternatively it could be an override parameter, for instance, that would avoid the runtime cache in update scenarios.

please check last commit; it should ignores cached data only during install/upgrade

szaimen (approved):
Works as well in my testing (but cannot judge the code). Thanks Maxence! :)

szaimen (approved):

blizzz (approved):

szaimen:
CI failure unrelated

PR #36260, opened 2023-01-20
Fix WHERE condition when selecting user's availability for the status…

nickvergessen:
… automation

## Summary

Over some time now I noticed that my user status was not toggling correctly with my working schedule.
Since I had quite a "complex" time schedule I assumed the math/checking of IN or OUT is not working.
But after trying to extract my data from our instance to replay it locally, I found the root-cause and it's so much simpler.

Since this is not the first time we have a where-whereOr/whereAnd-where situation, I wonder if we should add some safety net by logging an error when `where` is being used on a query builder while the current where is not empty. But it reduces reusability of QueryBuilder objects (a pattern I dislike anyway 😅)

## Checklist


AndyScherzinger (approved):
LGTM

vitormattos (approved):

nickvergessen:
New in 25, so okay

PR #36287, opened 2023-01-22
Deprecate and cleanup `OC.L10N` in favor of `@nextcloud/l10n` 

susnux:
## Summary
This PR consists of three parts:
1. Cleanup long standing deprecation from `OC`: `OC.addTranslations` was deprecated with NC17 and is not used anymore by any app within the github nextcloud organization
2. Deprecate `OC.getLocale` and `OC.getLanguage` as they are provided by `@nextcloud/l10n` and drop duplicated code
3. Deprecate and replace the translation functions implementation with the `@nextcloud/l10n`.

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated
  - See PR: https://github.com/nextcloud/documentation/pull/9575 (upgrade guide)


ChristophWurst (approved):
:broom:

Pytal (approved):

core/src/OC/l10n.js
```diff
@@ -28,79 +27,56 @@
...
-		registerAppTranslations(appName, bundle, this._getPlural)
-	},
+	register,
 
 	/**
 	 * @private
+	 * @deprecated 26.0.0 use `_unregister` from https://www.npmjs.com/package/@nextcloud/l10n
 	 */
-	_unregister: unregisterAppTranslations,
+	_unregister,
```
skjnldsv, line 73: is it a private method? Then it should not be exposed or it should be made public?

skjnldsv (approved):

core/src/OC/l10n.js
susnux: This is currently only used in CI tests, but I do not see a reason why this function should not be on the public API.
Maybe we should change this on the `@nextcloud/l10n` package.

But for this PR I wanted to keep the changes small and do not introduce a breaking API change.

susnux:
Rebased onto current master

susnux:
Blocked by nextcloud/nextcloud-l10n#571 (and we need to wait for a new release)
There was an issue introduced in the standalone implementation which caused the function to synchronously load translations instead of asynchronously, this is the reason the test fail here.

susnux:
The drone CI failure on the mysql DB tests seem unrelated.

PR #36002, opened 2023-01-04
feat(files): Quota in navigation

skjnldsv:
Followup #35772 

| Unlimited | With quota |
|:---------:|:------:|
|||

| Warning >80% | Full |
|:---------:|:------:|
|||

## Feat
- [x] Click refreshes the quota
- [x] Auto-refresh in background
- [x] Auto-refresh from global files event (deletion, upload, update)
- [x] Testing

### Associated fixes
- https://github.com/nextcloud/nextcloud-vue/pull/3613
- https://github.com/nextcloud/nextcloud-vue/pull/3617

PR #36301, opened 2023-01-23
feat(files) Update quota when updating storage statistics

### LINKED ISSUE #36300: [Files] Atomatically refresh quota of current folder when using  folders with custom quota
### How to use GitHub

When we open a folder with a custom quota, the quota displayed is always the quota of the root folder of the user, which can be confusing

In apps/files when using folders with custom quota (with groupfolders or any other apps that allow it) it would be great to refresh the quota displayed at the bottom left of the page with the quota/size of the current opened folder. 

I don't see any

**Additional context**

Users have root folder with a quota and multiple groupfolders with others quotas. So quotas and folders size are always different and need to be refresh to be right

tpayen:
* Resolves: #36300 

## Summary

It's free to update the quota at the same time as the storage statistics. This is very useful for folders with different quotas (like groupfolders)

## Checklist

### DIFF
--- a/apps/files/js/files.js
+++ b/apps/files/js/files.js
@@ -31,6 +31,7 @@
 				state.dir = null;
 				state.call = null;
 				Files.updateMaxUploadFilesize(response);
+				Files.updateQuota(response);
 			});
 		},
 		// update quota


Pytal (approved):

artonge (approved):

artonge (changes requested):
Might actually be irelevant with https://github.com/nextcloud/server/pull/36002

@skjnldsv can you confirm?

tpayen:
> We should maybe discuss what it should look like when we're seeing a folder-quota. That could be confusing.

In fact I think it works just like any e-mail application (like Thunderbird or Roundcube for the ones I'm most familliar with) the quota shown is relative to the current mailbox/folder and that's pretty clear

skjnldsv:
@nimishavijay @jancborchardt, currently thi sis how it look:
image

When we enter a folder that have a different quota, could you tell us what it should look like? If we just adjust the display of the image above, I am afraid it would be confusing to the user to see their quota suddenly change. How can we make sure they understand the current folder they're in have a different quota than their global storage? :)

nimishavijay:
> I am afraid it would be confusing to the user to see their quota suddenly change.

Definitely agreed. A simple enough idea is: if the user is in a special folder, add the folder name to the end of the space info with a middle dot separator (looks great but will probably be ellipsised for most folder names) or on a new line (folder names more visible).

How it would look (switched the positions of the progress bar and text so that the progress bar remains closer to the quota info):
image

image

What do you think? I'm leaning towards the second one because it is more clear :)

skjnldsv:
Got it @nimishavijay 
Will do for 26 and above.

@tpayen considering this changed, we might drop that PR and only do this for the upcoming nc version

tpayen:
> @tpayen considering this changed, we might drop that PR and only do this for the upcoming nc version

Clearly, I close this one

PR #32631, opened 2022-05-27
Get access list share by email recipients

ASerbinski:
Close https://github.com/nextcloud/server/issues/32629

## Summary: 

Return the recipients for an email share. 

We already have `mail`, but it's unused / broken.

The backstory: 

In https://github.com/nextcloud/server/pull/2834 `mail => bool` was documented but not implemented. 

image

Commit to change mail to public: https://github.com/nextcloud/server/pull/2834/commits/cab41118f67e6a4743bac2cfcfa69a4bee28f2a3

PR #2834, opened 2016-12-22
Access list to share manager

rullzer:
Revived this PR to get rid of the old code

What we do now is offload the accesslist creation to the share providers. Which allows for way more efficient queries.

The manager function now can handle recursion and directly filter out users the currently have no access to the Node (delete group shares for example).

The helper class should be able to assist for notifications.

@nickvergessen @schiessle lets see if we can get this moving!

TODO: 
- [X] add  accesslist function to circles provider
  * Issue at https://github.com/nextcloud/circles/issues/46

PR #36305, opened 2023-01-23
Bugfix/32630/get access list use correct array key

### LINKED ISSUE #32630: [Bug]: getAccessList in ShareByMailProvider uses incorrect array key

### Bug description

Per documentation https://github.com/nextcloud/server/blob/master/lib/private/Share20/Manager.php#L1671 -- share by mail should use the array key "mail". Instead, it is using the array key "public", which collides with array key used in DefaultShareProvider for share by link; https://github.com/nextcloud/server/blob/master/lib/private/Share20/DefaultShareProvider.php#L1368

Change is required here;
https://github.com/nextcloud/server/blob/master/apps/sharebymail/lib/ShareByMailProvider.php#L1211

### Steps to reproduce

Not applicable

### Expected behavior

ShareByMailProvider should use "mail" as array key.

### Installation method

_No response_

### Operating system

_No response_

### PHP engine version

_No response_

### Web server

_No response_

### Database engine version

_No response_

_No response_

_No response_

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

```shell
N/A
```

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

ASerbinski:
* Resolves: #32630 

## Summary
Switch to correct array key in ShareByMailProvider/getAccessList -- should be 'mail' not 'public'.

### DIFF
--- a/apps/sharebymail/lib/ShareByMailProvider.php
+++ b/apps/sharebymail/lib/ShareByMailProvider.php
@@ -1227,7 +1227,7 @@ class ShareByMailProvider implements IShareProvider {
 		$mail = $cursor->fetch() !== false;
 		$cursor->closeCursor();
 
-		return ['public' => $mail];
+		return ['mail' => $mail];
 	}
 
 	public function getAllShares(): iterable {
--- a/apps/sharebymail/tests/ShareByMailProviderTest.php
+++ b/apps/sharebymail/tests/ShareByMailProviderTest.php
@@ -1086,11 +1086,11 @@ class ShareByMailProviderTest extends TestCase {
 		$folder = $rootFolder->getUserFolder($u1->getUID())->newFolder('foo');
 
 		$accessList = $provider->getAccessList([$folder], true);
-		$this->assertArrayHasKey('public', $accessList);
-		$this->assertFalse($accessList['public']);
+		$this->assertArrayHasKey('mail', $accessList);
+		$this->assertFalse($accessList['mail']);
 		$accessList = $provider->getAccessList([$folder], false);
-		$this->assertArrayHasKey('public', $accessList);
-		$this->assertFalse($accessList['public']);
+		$this->assertArrayHasKey('mail', $accessList);
+		$this->assertFalse($accessList['mail']);
 
 		$share1 = $this->shareManager->newShare();
 		$share1->setSharedWith('user@server.com')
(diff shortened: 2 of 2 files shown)


kesselb:
Thank you very much :+1: 

Can you have a look at the failing test:

```
1) OCA\ShareByMail\Tests\ShareByMailProviderTest::testGetAccessList
Failed asserting that an array has the key 'public'.

/home/runner/work/server/server/apps/sharebymail/tests/ShareByMailProviderTest.php:1089
```

kesselb:
Thank you :+1: 

Please revert the changes to composer.lock

ASerbinski:
Hi @kesselb ; that commit changing composer.lock seems to be somehow added by github, not by me. I'm not sure how to prevent it from being added.

kesselb:
Thank you :+1: 

I read https://github.com/nextcloud/server/pull/2834 to understand `public` and `mail`. 
If I understand it correct, public is fine and the documentation is wrong. 

https://github.com/nextcloud/server/blob/f5c361cf44739058b79f322576a1bad2d8c142d9/lib/private/Share20/Manager.php#L1761-L1763

The above code will "merge" the values together. 
If something is shared by MailShareProvider or LinkShareProvider public should be true. 

I believe that's good news for your other pull request.

ASerbinski:
Hi @kesselb ; I believe that the documentation is correct and that the PR #2834 you referenced is from a more primitive state before the 'mail' key was added.

Per @schiessle ; "... To make it more generic we could also add another key 'mail' to the array, which would probably be the better solution."

kesselb:
> I believe that the documentation is correct and that the PR https://github.com/nextcloud/server/pull/2834 you referenced is from a more primitive state before the 'mail' key was added.

The documentation for `mail` was added with the mentioned pr.  

> To make it more generic we could also add another key 'mail' to the array, which would probably be the better solution.

I guess that was an idea, but not done. 

The takeaway for your pull request to list the mail shares for a node is to keep the current behavior (add public = true when at least one mail share exist) and add the mail attribute to the result. 

I'm unsure if we need this pr. There is nothing broken to be fixed apart from the documentation. Sorry, for asking you to create a pull request for this change.

ASerbinski:
Hi @kesselb ; I added a commit to https://github.com/nextcloud/server/pull/32631 to comply with your directive to both preserve the prior behavior with respect to the 'public' array key, and to use the 'mail' key per proposed new usage.

PR #36316, opened 2023-01-23
Log to `error_log` in fatal case

R0Wi:
## Summary

Write a log message via [`error_log`](https://www.php.net/manual/en/function.error-log.php) in case we can't regularly write to our logfile. This improves troubleshooting.

## Use case

**Example:**
We forgot to install the `php-apcu` module but configured it in `config.php`. This could happen for example when upgrading the PHP version or when not properly reading the manual.

**Before:**
When trying to access NC, we're presented with a generic error message. 

```
Internal Server Error

The server encountered an internal error and was unable to complete your request.
Please contact the server administrator if this error reappears multiple times, please include the technical details below in your report.
More details can be found in the server log.
```

Neither `nextcloud.log` nor `/var/log/apache2/error.log` (or something similar) is written. As an admin we absolutely don't know what's going on (Btw: I only solved this by debugging via XDebug).

**After:**
There is a log entry in `/var/log/apache2/error.log` telling us:

```
[Mon Jan 23 17:57:01.559577 2023] [php:notice] [pid 189438] [client ::1:48076] Error when trying to log exception: Memcache \\OC\\Memcache\\APCu not available for local cache 
#0 /var/www/html/nextcloud/lib/private/Server.php(745): OC\\Memcache\\Factory->__construct()
#1 /var/www/html/nextcloud/lib/private/AppFramework/Utility/SimpleContainer.php(162): OC\\Server->OC\\{closure}()
#2 /var/www/html/nextcloud/3rdparty/pimple/pimple/src/Pimple/Container.php(122): OC\\AppFramework\\Utility\\SimpleContainer->OC\\AppFramework\\Utility\\{closure}()
#3 /var/www/html/nextcloud/lib/private/AppFramework/Utility/SimpleContainer.php(129): Pimple\\Container->offsetGet()
...
```

This is not formatted nicely but it tells the admin what's the problem and why the NC server instance is down.

## Checklist

- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included - not needed
- [x] Screenshots before/after for front-end changes - no changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required - no doc changes
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes) - non critical

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -359,6 +359,7 @@ class Log implements ILogger, IDataLogger {
 			$context['level'] = $level;
 		} catch (Throwable $e) {
 			// make sure we dont hard crash if logging fails
+			error_log('Error when trying to log exception: ' . $e->getMessage() . ' ' . $e->getTraceAsString());
 		}
 	}
 


psyray:
Thanks for this PR it helps me to identify the problem
It was APCu and memcached PHP extensions missing for me

I don't understand why there is a try..catch without log

lib/private/Log.php
```diff
@@ -359,6 +359,7 @@ public function logData(string $message, array $data, array $context = []): void
 			$context['level'] = $level;
 		} catch (Throwable $e) {
 			// make sure we dont hard crash if logging fails
+			error_log('Error when trying to log exception: ' . $e->getMessage() . ' ' . $e->getTraceAsString());
```
szaimen, line 362: Isn't that enough? Otherwise we might risk that secrets get logged?

```suggestion
			error_log('Error when trying to log exception: ' . $e->getMessage());
```

R0Wi: Well in general I would agree but only the exception message might not be enough to investigate what went wrong. And this handler should only be called as a last resort.

In my case the home directory of the user was missing so I'd only get something like "File not found". The stacktrace helped my finding the issue. But yeah if this is a security concern the message would be better than nothing :-)

nickvergessen: agreed, this is the error after an error, so fine to log with details.

R0Wi:
@nickvergessen I think the CI/CD errors are not related to these changes, right? So could we merge this?

szaimen (approved):
Fine by me then but didnt test

R0Wi:
The only test currently failing is `postgres10-php8.0 `: 

```
There was 1 failure:
--
115 |  
116 | 1) Test\User\DatabaseTest::testSearch
117 | Failed asserting that actual size 3 matches expected size 2.
118 |  
119 | /drone/src/tests/lib/User/Backend.php:114
120 | /drone/src/tests/lib/User/DatabaseTest.php:125
```

Would assume this is not related to the current change?

R0Wi:
Any chance to get this minor change merged, please?

nickvergessen (approved):

nickvergessen:
CI fail is known and unrelated.
So rebasing and setting to merge

szaimen:
CI failure unrelated

PR #36333, opened 2023-01-24
Fix default value for files_no_background_scan

max-nextcloud:
The value is read in https://github.com/nextcloud/server/blob/b888c6146327d842e21988b0d90d6ade4f3a9435/apps/files/lib/BackgroundJob/ScanFiles.php#L105 and the default value is false.

## Summary

The documented default value for `files_no_background_scan` was wrong.

## Checklist

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -2223,7 +2223,7 @@ $CONFIG = [
  * scan to sync filesystem and database. Only users with unscanned files
  * (size < 0 in filecache) are included. Maximum 500 users per job.
  *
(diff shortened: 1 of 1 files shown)


max-nextcloud:
I'm not sure how best to update the documentation based on this. I can change the value there directly or try and rebuild it from the `config.sample.php`.

szaimen (approved):
LGTM

kesselb (approved):
Thank you

kesselb:
The documentation should be updated automatically ;)

PR #36341, opened 2023-01-25
fix(app-framework): Specify return type of Middleware::beforeController

ChristophWurst:
Resolves: n/a

## Summary

The abstract class has no defined return type. Adding one makes it easier to write a correct subclass override for that method.

We can not add a PHP return type yet because that would potentially cause trouble with untyped subclasses: https://3v4l.org/dRU5Q

## Checklist


ChristophWurst:
Tests are all happy. Let's do this?

susnux (approved):
👍

miaulalala (approved):

PR #36348, opened 2023-01-25
Fix primary color usage in emails and federation

nickvergessen:
* Resolves: N/A

## Summary

* Federation box to be implemented on own website also used the background image color instead of the instance theme
* Emails used the color of the triggering user, so if you invite 5 users with different custom background image to an event, you will get 5 differently colored emails instead.
2 sample emails from the same day, same instance, no settings changed on my side:

From person 1 | From person 2
---|---
Bildschirmfoto vom 2023-01-24 17-08-23 | Bildschirmfoto vom 2023-01-24 17-08-29

* https://github.com/nextcloud/server/blob/304c1b9b61b20dfa3c1f07e8a914f0661d91704d/apps/encryption/templates/mail.php#L9 is also affected, but I will instead of patching that migrate it to the proper templating mechanism => https://github.com/nextcloud/server/pull/36351

## Screenshots

* Theming color: `#DF2254`
* User custom picked theming color: `#ffffff`

Before | After
---|---
Bildschirmfoto vom 2023-01-25 10-46-02 | Bildschirmfoto vom 2023-01-25 10-44-19
Bildschirmfoto vom 2023-01-25 10-52-59 | Bildschirmfoto vom 2023-01-25 10-39-40

## Checklist

- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are adjusted where applicable

### DIFF
--- a/lib/public/Defaults.php
+++ b/lib/public/Defaults.php
@@ -206,6 +206,18 @@ class Defaults {
 	}
 
 	/**
+	 * Return the default color primary
+	 * @return string
+	 * @since 25.0.4
+	 */
+	public function getDefaultColorPrimary(): string {
+		if (method_exists($this->defaults, 'getDefaultColorPrimary')) {
+			return $this->defaults->getDefaultColorPrimary();
+		}
+		return $this->defaults->getColorPrimary();
+	}
+
+	/**
 	 * @param string $key
 	 * @return string URL to doc with key
 	 * @since 12.0.0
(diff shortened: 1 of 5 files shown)


lib/public/Defaults.php
```diff
@@ -205,6 +205,18 @@ public function getColorPrimary(): string {
...
+	 * Return the default color primary
+	 * @return string
+	 * @since 25.0.4
+	 */
+	public function getDefaultColorPrimary(): string {
+		if (method_exists($this->defaults, 'getDefaultColorPrimary')) {
+			return $this->defaults->getDefaultColorPrimary();
+		}
+		return $this->defaults->getColorPrimary();
+	}
```
skjnldsv, line 218: `getDefaultColorPrimary` should always be available no?

nickvergessen: `OC_Defaults` only has `getColorPrimary` and I'm unwilling to debug which version of the defaults object is used where and why.
https://github.com/nextcloud/server/blob/3860dad9eaac937288d125c75e7f58cfbe44848a/lib/private/legacy/OC_Defaults.php#L301-L309

That's a cleanup task for another day.

skjnldsv: ok then

skjnldsv (approved):

szaimen (approved):
🐘

Pytal (approved):

nickvergessen:
Of course there are tests which checks which method is invoked, when getting color in the email... will have a look tomorrow.

PR #36355, opened 2023-01-25
Add generic templates for DataResponse and JSONResponse

provokateurin:
## Summary

Add templates to indicate generic type. Not required, but very useful for OpenAPI.

## Checklist

### DIFF
--- a/lib/public/AppFramework/Http/JSONResponse.php
+++ b/lib/public/AppFramework/Http/JSONResponse.php
@@ -32,18 +32,19 @@ use OCP\AppFramework\Http;
 /**
  * A renderer for JSON calls
  * @since 6.0.0
+ * @template T
  */
 class JSONResponse extends Response {
 	/**
 	 * response data
-	 * @var array|object
+	 * @var T
 	 */
 	protected $data;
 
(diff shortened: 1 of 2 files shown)


come-nc:
This adds a new constraints that one cannot use setData with another type than what was used in the constructor. No idea if this is a problem or not.

lib/public/AppFramework/Http/JSONResponse.php
```diff
@@ -32,18 +32,19 @@
 /**
  * A renderer for JSON calls
  * @since 6.0.0
+ * @template T
  */
 class JSONResponse extends Response {
 	/**
 	 * response data
-	 * @var array|object
+	 * @var T
```
ChristophWurst, line 40: this weakens the type. before it was an array or object. now it can be anything

provokateurin:
Waiting on a solution for https://github.com/nextcloud/server/issues/36513 before continuing since this change might not be needed after all.

provokateurin:
I just realized this isn't possible, since psalm will complain about missing template types in every place that isn't updated.

PR #36356, opened 2023-01-25
cloud_federation_api: Introduce OpenAPI spec

provokateurin:
## Summary

Adapts the code to make it friendly for https://github.com/nextcloud-gmbh/openapi-extractor and add the automatically generated spec.

## TODO
- [x] https://github.com/nextcloud/server/issues/36513

## Checklist

### DIFF
--- a/apps/cloud_federation_api/lib/Controller/RequestHandlerController.php
+++ b/apps/cloud_federation_api/lib/Controller/RequestHandlerController.php
@@ -5,6 +5,7 @@
  * @author Bjoern Schiessle <bjoern@schiessle.org>
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -22,9 +23,12 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
+
 namespace OCA\CloudFederationAPI\Controller;
 
+use Exception;
 use OCA\CloudFederationAPI\Config;
+use OCA\CloudFederationAPI\ResponseDefinitions;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\JSONResponse;
@@ -41,6 +45,7 @@ use OCP\IRequest;
 use OCP\IURLGenerator;
 use OCP\IUserManager;
 use OCP\Share\Exceptions\ShareNotFound;
+use OCP\Util;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -77,15 +82,15 @@ class RequestHandlerController extends Controller {
 	private $cloudIdManager;
 
 	public function __construct($appName,
-								IRequest $request,
-								LoggerInterface $logger,
-								IUserManager $userManager,
-								IGroupManager $groupManager,
-								IURLGenerator $urlGenerator,
-								ICloudFederationProviderManager $cloudFederationProviderManager,
-								Config $config,
-								ICloudFederationFactory $factory,
-								ICloudIdManager $cloudIdManager
+		IRequest $request,
+		LoggerInterface $logger,
+		IUserManager $userManager,
+		IGroupManager $groupManager,
+		IURLGenerator $urlGenerator,
+		ICloudFederationProviderManager $cloudFederationProviderManager,
+		Config $config,
+		ICloudFederationFactory $factory,
+		ICloudIdManager $cloudIdManager
 	) {
 		parent::__construct($appName, $request);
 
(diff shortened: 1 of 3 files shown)


apps/cloud_federation_api/lib/Controller/RequestHandlerController.php
```diff
@@ -22,10 +22,14 @@
...
  */
+
 namespace OCA\CloudFederationAPI\Controller;
 
+use CloudFederationAddShare;
+use CloudFederationError;
+use CloudFederationValidationError;
+use Exception;
 use OCA\CloudFederationAPI\Config;
-use OCP\AppFramework\Controller;
```
nickvergessen, line 28: ???

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.response.php
```diff
@@ -0,0 +1,13 @@
+<?php
```
nickvergessen, line 1: file name mismatch
`CloudFederationAddShare.php` is expected.
https://www.php-fig.org/psr/psr-4/

nickvergessen: Missing
```php
declare(strict_types=1);
```

nickvergessen: Missing copyright header

come-nc: Missing namespace

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.response.php
```diff
@@ -0,0 +1,13 @@
...
+
+class CloudFederationAddShare implements JsonSerializable {
+	public ?string $recipientDisplayName;
+
+	public function __construct(?string $recipientDisplayName) {
+		$this->recipientDisplayName = $recipientDisplayName;
+	}
```
come-nc, line 8:
```suggestion
	public function __construct(
		public ?string $recipientDisplayName
	) {
	}
```

provokateurin: Thanks for this hint, I made a big refactor and the extractor is a lot better now and this is now the only supported way to declare models.

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.php
```diff
@@ -0,0 +1,40 @@
...
+declare(strict_types=1);
+
+namespace OCA\CloudFederationAPI\Model;
+
+use JsonSerializable;
+
+class CloudFederationAddShare implements JsonSerializable {
+	/**
+	 * @param string|null $recipientDisplayName
+	 */
```
nickvergessen, line 33: Can be removed. The typehint is enough

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.php
```diff
@@ -0,0 +1,40 @@
...
+	public function __construct(public ?string $recipientDisplayName) {
+	}
+
+	public function jsonSerialize(): array {
+		return ["recipientDisplayName" => $this->recipientDisplayName];
```
nickvergessen, line 38:
```suggestion
		return ['recipientDisplayName' => $this->recipientDisplayName];
```

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.php
```diff
@@ -0,0 +1,40 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+declare(strict_types=1);
```
nickvergessen, line 24: Needs to be the first thing even before comments/copyright

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace OCA\CloudFederationAPI\Model;
```
nickvergessen, line 26: If we don't add Response to the class name, we should have it in the namespace.
Could image `ResponseModel` or `Model\Response`

nickvergessen, line 26: If we don't add Response to the class name, we should have it in the namespace.
Could image `ResponseModel` or `Model\Response`

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.php
```diff
@@ -0,0 +1,40 @@
...
+
+use JsonSerializable;
+
+class CloudFederationAddShare implements JsonSerializable {
+	/**
+	 * @param string|null $recipientDisplayName
+	 */
+	public function __construct(public ?string $recipientDisplayName) {
```
nickvergessen, line 34:
```suggestion
	public function __construct(protected ?string $recipientDisplayName) {
```

apps/cloud_federation_api/lib/Model/CloudFederationAddShare.php
provokateurin: The parsing currently only relies on the doc comments. I've done this because it's easier to only have a single way of parsing and the parameters for the controller methods all need to have a comment anyway. I'll add this though so it's cleaner in the models.

nickvergessen: We could document this on the JsonSerialize method. There is is more useful as it allows psalm to inherit the knowledge later on.
Sample https://github.com/nextcloud/notifications/pull/1429/files#diff-5f4fa2d752904832306c56f7832e2126a1e12db8dd1679abd71e526b77989821R215

provokateurin: It don't see the point in doing that. The jsonSerialize method will never be called by hand anyway.

nickvergessen: Only in the current cases

provokateurin: Then let's only add it when it's required?

apps/cloud_federation_api/lib/Model/Response/CloudFederationValidationError.php
```diff
@@ -0,0 +1,41 @@
...
+
+namespace OCA\CloudFederationAPI\Model\Response;
+
+use JsonSerializable;
+
+class CloudFederationValidationError extends CloudFederationError implements JsonSerializable {
+	/**
+	 * @param array{name: string|null, message: string|null}[]|null $validationErrors
+	 */
+	public function __construct(string $message, protected ?array $validationErrors = null) {
```
come-nc, line 34:
```suggestion
	public function __construct(
		string $message, 
		protected ?array $validationErrors = null,
	) {
```
Please break constructors like this so that properties are easier to spot.

PR #36360, opened 2023-01-25
also set the node when re-using existing share on share create

icewind1991:
This not only helps with performance by not having to get the node later, but also prevents errors if the origin share owner no longer has access to the node.


come-nc (approved):

marcelklehr (approved):
Customer confirms this fixes the error

PR #36371, opened 2023-01-26
Mark compiled CSS as binary

Pytal:
## Summary

Mark compiled CSS as binary in the `.gitattributes` file as viewing the textual diff of a single long line in e.g. https://github.com/nextcloud/server/blob/4caa90dfb643aea958b10d0741a8f21be5e18f39/core/css/server.css and https://github.com/nextcloud/server/blob/4caa90dfb643aea958b10d0741a8f21be5e18f39/core/css/server.css.map is not necessary when the source SCSS files e.g. https://github.com/nextcloud/server/blob/4caa90dfb643aea958b10d0741a8f21be5e18f39/core/css/styles.scss are much more readable

## Checklist

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 /dist/* binary
 /package-lock.json merge=binary
+/core/css/*.css* binary


szaimen (approved):
LGTM but didnt test

ChristophWurst (approved):
Might also make sense to move generated css files to /dist and leave core/css for source files only

skjnldsv:
> Might also make sense to move generated css files to /dist and leave core/css for source files only

Won't work with our `addScript` php handler

artonge (approved):

ChristophWurst:
> Won't work with our `addScript` php handler

yet ;)

Issue #35763, opened 2022-12-14
[BITV] User status dialog "Clear status after" menu entries need border for focus feedback


Pytal:
- [x] Bump https://github.com/nextcloud/server/pull/35717
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3656
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3663
- [x] https://github.com/nextcloud/server/pull/36321
  - [x] https://github.com/nextcloud/server/pull/36322
- [x] https://github.com/nextcloud-deps/vue-select/pull/14
- [x] https://github.com/nextcloud-deps/vue-select/pull/16
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3748
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3755
- [x] https://github.com/nextcloud/server/pull/36813
- ~https://github.com/nextcloud/server/pull/36372~
- [x] https://github.com/nextcloud/server/pull/36843
- [x] https://github.com/nextcloud/server/pull/36932

Pytal:
While porting the selector it seemed that the dropdown does not position as expected so we need to implement a configurable dropdown position in NcSelect

Pytal:
While porting the selector it seemed that the dropdown did not position as expected so we need to implement a configurable dropdown position in NcSelect

PR #36372, opened 2023-01-26
[WIP] Port clear status after select

Pytal:
- For https://github.com/nextcloud/server/issues/35763

## Summary

Port the clear status after select to NcSelect

- [x] FIXME Initial dropdown width is not equal to input width https://github.com/nextcloud-deps/vue-select/pull/14

## Checklist


Pytal:
The `width` property
https://github.com/sagalbot/vue-select/blob/857753655adbebae612518aab072ed7f4a8deaa0/src/directives/appendToBody.js#L4-L9
is ~10px less than it should be when opening the dropdown for the first time

Further investigations to be done

Pytal:
The initial `width` issue on first open can be reproduced on Chrome on Windows and Linux

| | Screenshot |
--- | ---
**Chrome (Windows)** | image
**Chrome (Linux)** | image

<details>
<summary>Expected width</summary>

The width is normal when opening subsequent times

| | Screenshot |
--- | ---
**Opening subsequent times** | image

</details>

And also reproduces on Firefox on Windows but not Linux

| | Screenshot |
--- | ---
**Firefox (Windows)** | image
**Firefox (Linux)** | image

Further investigations needed

Pytal:
Does the bug in https://github.com/nextcloud/server/pull/36372#issuecomment-1413037603 reproduce for you @artonge?

And based on the source of the `width` in https://github.com/sagalbot/vue-select/blob/v3.20.2/src/directives/appendToBody.js#L4-L9 and the assignment in https://github.com/nextcloud/nextcloud-vue/blob/v7.5.0/src/components/NcSelect/NcSelect.vue#L790-L791, any ideas how this could occur?

artonge:
> Does the bug in [#36372 (comment)](https://github.com/nextcloud/server/pull/36372#issuecomment-1413037603) reproduce for you @artonge?

I can reproduce in Chrome on Linux and not on Firefox.

Maybe it is because the computation happens when the scale is less than 1 ?
It looks like we are animating the modal:
https://github.com/nextcloud/nextcloud-vue/blob/c1a15b3f6c5cec758cd0461b154821ef07fd5013/src/components/NcModal/NcModal.vue#L980-L983

artonge:
It also happens if I open the modal, resize the window, which resize NcSelect. So probably nothing to do with the animation.

Pytal:
Fixed upstream in https://github.com/nextcloud-deps/vue-select/pull/14

Pytal:
Superseded by https://github.com/nextcloud/server/pull/36843

PR #36363, opened 2023-01-25
feat(app-framework): Add UseSession attribute to replace annotation

ChristophWurst:
* Resolves: the urge to do something fancy

## Summary

https://help.nextcloud.com/t/how-should-we-use-php8-attributes/104278

## TODO

- [x] Add new attribute
- [x] Adjust session middleware
- [x] Migrate core controllers

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
  - [x] https://github.com/nextcloud/documentation/pull/9589
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #36408, opened 2023-01-27
user_status: Add OpenAPI spec

provokateurin:
## Summary

Adapts the code to make it friendly for https://github.com/nextcloud-gmbh/openapi-extractor and add the automatically generated spec.

## TODO
- [x] https://github.com/nextcloud/server/issues/36513

## Checklist

### DIFF
--- a/apps/user_status/lib/Controller/HeartbeatController.php
+++ b/apps/user_status/lib/Controller/HeartbeatController.php
@@ -6,6 +6,7 @@ declare(strict_types=1);
  * @copyright Copyright (c) 2020, Georg Ehrke
  *
  * @author Georg Ehrke <oc.list@georgehrke.com>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -23,12 +24,12 @@ declare(strict_types=1);
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
+
 namespace OCA\UserStatus\Controller;
 
 use OCA\UserStatus\Db\UserStatus;
+use OCA\UserStatus\ResponseDefinitions;
 use OCA\UserStatus\Service\StatusService;
-use OCP\AppFramework\Controller;
-use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
@@ -53,12 +54,13 @@ class HeartbeatController extends OCSController {
 	/** @var StatusService */
 	private $service;
 
-	public function __construct(string $appName,
-								IRequest $request,
-								IEventDispatcher $eventDispatcher,
-								IUserSession $userSession,
-								ITimeFactory $timeFactory,
-								StatusService $service) {
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		IEventDispatcher $eventDispatcher,
+		IUserSession $userSession,
+		ITimeFactory $timeFactory,
+		StatusService $service) {
 		parent::__construct($appName, $request);
 		$this->eventDispatcher = $eventDispatcher;
 		$this->userSession = $userSession;
@@ -69,11 +71,14 @@ class HeartbeatController extends OCSController {
 	/**
 	 * @NoAdminRequired
 	 *
-	 * @param string $status
-	 * @return DataResponse
+	 * @param string $status online, away
+	 * @psalm-import-type PrivateUserStatus from ResponseDefinitions
+	 * @return DataResponse<PrivateUserStatus> 200 Status successfully updated
+	 * @return DataResponse 204 User has no status to update
+	 * @return DataResponse 400 Invalid status to update
 	 */
 	public function heartbeat(string $status): DataResponse {
-		if (!\in_array($status, [IUserStatus::ONLINE, IUserStatus::AWAY], true)) {
+		if (!in_array($status, [IUserStatus::ONLINE, IUserStatus::AWAY], true)) {
 			return new DataResponse([], Http::STATUS_BAD_REQUEST);
 		}
 
--- a/apps/user_status/lib/Controller/PredefinedStatusController.php
+++ b/apps/user_status/lib/Controller/PredefinedStatusController.php
@@ -6,6 +6,7 @@ declare(strict_types=1);
  * @copyright Copyright (c) 2020, Georg Ehrke
  *
  * @author Georg Ehrke <oc.list@georgehrke.com>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -23,8 +24,10 @@ declare(strict_types=1);
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
+
 namespace OCA\UserStatus\Controller;
 
+use OCA\UserStatus\ResponseDefinitions;
 use OCA\UserStatus\Service\PredefinedStatusService;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
@@ -47,9 +50,10 @@ class PredefinedStatusController extends OCSController {
 	 * @param IRequest $request
 	 * @param PredefinedStatusService $predefinedStatusService
 	 */
-	public function __construct(string $appName,
-								IRequest $request,
-								PredefinedStatusService $predefinedStatusService) {
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		PredefinedStatusService $predefinedStatusService) {
 		parent::__construct($appName, $request);
 		$this->predefinedStatusService = $predefinedStatusService;
 	}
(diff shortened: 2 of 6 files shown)


apps/user_status/lib/Controller/HeartbeatController.php
```diff
@@ -53,12 +53,12 @@ class HeartbeatController extends OCSController {
...
-								IRequest $request,
+	public function __construct(string           $appName,
+								IRequest         $request,
 								IEventDispatcher $eventDispatcher,
-								IUserSession $userSession,
-								ITimeFactory $timeFactory,
-								StatusService $service) {
+								IUserSession     $userSession,
+								ITimeFactory     $timeFactory,
+								StatusService    $service) {
```
come-nc, line 61: This spacing is not consistent with the rest of the code base.

apps/user_status/lib/Controller/HeartbeatController.php
```diff
@@ -69,11 +69,13 @@ public function __construct(string $appName,
 	/**
 	 * @NoAdminRequired
 	 *
-	 * @param string $status
-	 * @return DataResponse
+	 * @param string $status online, away
+	 * @return DataResponse<PrivateUserStatus> 200 Status successfully updated
+	 * @return DataResponse 204 User has no status to update
+	 * @return DataResponse 400 Invalid status to update
```
come-nc, line 75: Several `@return` is not allowed it seems, see psalm errors.

apps/user_status/lib/Controller/HeartbeatController.php
```diff
@@ -95,20 +97,6 @@ public function heartbeat(string $status): DataResponse {
 			return new DataResponse([], Http::STATUS_NO_CONTENT);
 		}
 
-		/** @psalm-suppress UndefinedInterfaceMethod */
```
come-nc, line 98: You need to put back this psalm-suppress comment, or to fix the psalm error in an clean way.

apps/user_status/lib/Controller/PredefinedStatusController.php
```diff
@@ -57,12 +60,12 @@ public function __construct(string $appName,
 	/**
 	 * @NoAdminRequired
 	 *
-	 * @return DataResponse
+	 * @return DataResponse<PredefinedStatus[]> 200
```
come-nc, line 63: DataResponse is not a template in the current code base. Remove the parameter or update the docblocks of DataRepsonse to make it a template.

apps/user_status/lib/Controller/PredefinedStatusController.php
```diff
@@ -57,12 +60,12 @@ public function __construct(string $appName,
...
 	 */
-	public function findAll():DataResponse {
+	public function findAll(): DataResponse {
 		// Filtering out the invisible one, that should only be set by API
-		return new DataResponse(array_filter($this->predefinedStatusService->getDefaultStatuses(), function (array $status) {
-			return !array_key_exists('visible', $status) || $status['visible'] === true;
+		return new DataResponse(array_filter($this->predefinedStatusService->getDefaultStatuses(), function (PredefinedStatus $status) {
+			return $status->isVisible();
 		}));
```
come-nc, line 69:
```suggestion
		return new DataResponse(array_filter($this->predefinedStatusService->getDefaultStatuses(), fn (PredefinedStatus $status) => $status->isVisible()));
```
In this instance using the `fn` keyword is improving readability I think.

apps/user_status/lib/Model/Response/ClearAt.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ */
+
+namespace OCA\UserStatus\Model\Response;
+
+use JsonSerializable;
+use OCA\UserStatus\Db\UserStatus;
+
+class ClearAt implements JsonSerializable {
+	public function __construct(protected string $type, protected string|int $time) {
```
come-nc, line 32:
```suggestion
	public function __construct(
		protected string $type, 
		protected string|int $time,
	) {
```

apps/user_status/lib/Model/Response/PredefinedStatus.php
```diff
@@ -0,0 +1,48 @@
...
+ *
+ */
+
+namespace OCA\UserStatus\Model\Response;
+
+use JsonSerializable;
+
+class PredefinedStatus implements JsonSerializable {
+
+	public function __construct(protected string $id, protected string $icon, protected string $message, protected ?ClearAt $clearAt, protected ?bool $visible) {
```
come-nc, line 32: Same thing, one property per line please.

apps/user_status/lib/Model/Response/PrivateUserStatus.php
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+namespace OCA\UserStatus\Model\Response;
+
+use JsonSerializable;
+use OCA\UserStatus\Db\UserStatus;
+
+class PrivateUserStatus implements JsonSerializable {
+
+	public function __construct(protected string $userId, protected ?string $message, protected ?string $messageId, protected bool $messageIsPredefined, protected ?string $icon, protected int|ClearAt|null $clearAt, protected string $status, protected bool $statusIsUserDefined) {
```
come-nc, line 33: Same

apps/user_status/lib/Model/Response/PredefinedStatus.php
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	public function isVisible(): bool {
+		return $this->visible != false;
```
come-nc, line 36:
```suggestion
		return $this->visible ?? false;
```
Would it be possible to instead type the property as bool and have a default value in constructor?

apps/user_status/lib/Model/Response/PredefinedStatus.php
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	public function jsonSerialize(): array {
+		return [
+			'id' => $this->id,
+			'icon' => $this->icon,
+			'message' => $this->message,
+			'clearAt' => $this->clearAt,
+			'visible' => $this->visible,
```
come-nc, line 45: It’s okay to have null here in the json?

apps/user_status/lib/Model/Response/PublicUserStatus.php
```diff
@@ -0,0 +1,61 @@
...
+
+namespace OCA\UserStatus\Model\Response;
+
+use JsonSerializable;
+use OCA\UserStatus\Db\UserStatus;
+use OCP\UserStatus\IUserStatus;
+
+class PublicUserStatus implements JsonSerializable {
+
+	public function __construct(protected string $userId, protected ?string $message, protected ?string $icon, protected int|ClearAt|null $clearAt, protected string $status) {
```
come-nc, line 34: One prop per line

come-nc (changes requested):

apps/user_status/lib/Controller/HeartbeatController.php
provokateurin: Yeah this is still a problem I have no solution for. I really need to have this working, but I don't know how. Only a single return annotation is kind of impossible or will get really ugly.

apps/user_status/lib/Controller/PredefinedStatusController.php
provokateurin: Didn't know that existed!

apps/user_status/lib/Model/Response/PredefinedStatus.php
provokateurin: Not sure, I've gone with the default value now so it will never be null.

apps/user_status/lib/Controller/HeartbeatController.php
provokateurin: Wdym? We probably found a solution for this problem, but I first need to test it.

nickvergessen: https://www.php.net/manual/de/language.attributes.syntax.php it something we could use in Nextcloud 26+ as it requires PHP8+

Sample in https://github.com/nextcloud/server/pull/36363

provokateurin: Hm I see, but it wouldn't help with static analysis, right? The idea seems pretty good though.

christianlupus: If you define am attribute and scan statically for that attribute, you can find these. You might be able to use reflections to let the PHP parser do the scanning even. 
As PHP 7 is EOL, it should be good to go in general.
You could add a parameter to the attribute that contains an array. The keys would be the status codes and the value either a string of even a structure to describe the status.

provokateurin: Yes, but this won't help psalm validate the code in the method, right?

christianlupus: Maybe i got the problem wrong. Are you concerned about static code analysis using psalm or about the extraction of the API specification from the code?
For psalm a single return should be sufficient, right?

provokateurin: Both. Please let us discuss this in https://github.com/nextcloud/server/issues/36513#issuecomment-1424276606

PR #36417, opened 2023-01-28
Make the container fully fulfill PSR container interface

susnux:
## Summary
To fulfill the PSR container interface it must throw `NotFoundExceptionInterface` if the requested name was not found within the container. 
Added a simple private wrapper over the `QueryException` implementing the `NotFoundExceptionInterface` which is only thrown if the query was not found, this way the public API should be unchanged and only the thrown error **additionally** implements `NotFoundExceptionInterface`.

## Checklist

### DIFF
--- a/lib/composer/composer/autoload_classmap.php
+++ b/lib/composer/composer/autoload_classmap.php
@@ -720,6 +720,7 @@ return array(
     'OC\\AppFramework\\Services\\AppConfig' => $baseDir . '/lib/private/AppFramework/Services/AppConfig.php',
     'OC\\AppFramework\\Services\\InitialState' => $baseDir . '/lib/private/AppFramework/Services/InitialState.php',
     'OC\\AppFramework\\Utility\\ControllerMethodReflector' => $baseDir . '/lib/private/AppFramework/Utility/ControllerMethodReflector.php',
+    'OC\\AppFramework\\Utility\\QueryNotFoundException' => $baseDir . '/lib/private/AppFramework/Utility/QueryNotFoundException.php',
     'OC\\AppFramework\\Utility\\SimpleContainer' => $baseDir . '/lib/private/AppFramework/Utility/SimpleContainer.php',
     'OC\\AppFramework\\Utility\\TimeFactory' => $baseDir . '/lib/private/AppFramework/Utility/TimeFactory.php',
     'OC\\AppScriptDependency' => $baseDir . '/lib/private/AppScriptDependency.php',
--- a/lib/public/AppFramework/QueryException.php
+++ b/lib/public/AppFramework/QueryException.php
@@ -32,7 +32,7 @@ use Psr\Container\ContainerExceptionInterface;
 /**
  * Class QueryException
  *
- * The class extends `NotFoundExceptionInterface` since 20.0.0
+ * The class extends `ContainerExceptionInterface` since 20.0.0
  *
  * @since 8.1.0
  * @deprecated 20.0.0 catch \Psr\Container\ContainerExceptionInterface
--- a/tests/lib/AppFramework/Utility/SimpleContainerTest.php
+++ b/tests/lib/AppFramework/Utility/SimpleContainerTest.php
@@ -26,6 +26,7 @@ declare(strict_types=1);
 namespace Test\AppFramework\Utility;
 
 use OC\AppFramework\Utility\SimpleContainer;
+use Psr\Container\NotFoundExceptionInterface;
 
 interface TestInterface {
 }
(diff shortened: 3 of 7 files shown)


nickvergessen (approved):
https://www.php-fig.org/psr/psr-11/#33-psrcontainernotfoundexceptioninterface

ChristophWurst (approved):
Thanks!

ChristophWurst:
https://github.com/nextcloud/3rdparty/blob/1d53ed4d3282427854fca7ee6ecbb945304272db/composer.json#L47 time to bump that to v2

susnux:
> time to bump that to v2

Would include extending `Throwable` to the exceptions and adding the boolean return type hint to the `has` method.
As far as I can see this is already done on our implementation (extending `Exception` which implements `Throwable`), so this should be no breaking change.
Should I include this within this PR?

---

ChristophWurst:
Let's only fix the PSR compliance here to keep this backportable and do the bump for master/26 only.

susnux:
I have no clue why drone is failing, but it seems unrelated

ChristophWurst:
Test\AppFramework\Utility\SimpleContainerTest::testNothingRegistered sounds related

tests/lib/AppFramework/Utility/SimpleContainerTest.php
```diff
@@ -78,9 +79,13 @@ public function testRegister() {
...
 
 	public function testNothingRegistered() {
-		$this->expectException(\OCP\AppFramework\QueryException::class);
-
-		$this->container->query('something really hard');
+		try {
+			$this->container->query('something really hard');
+			$this->fail('Expected `QueryException` exception was not thrown');
+		} catch (Exception $exception) {
+			$this->assertTrue($exception instanceof \OCP\AppFramework\QueryException);
```
ChristophWurst, line 86: nit: ``self::assertInstanceOf(\OCP\AppFramework\QueryException::class, $exception)``

lib/public/AppFramework/QueryException.php
```diff
@@ -32,7 +32,7 @@
 /**
  * Class QueryException
  *
- * The class extends `NotFoundExceptionInterface` since 20.0.0
+ * The class extends `ContainerExceptionInterface` since 20.0.0
```
susnux, line 35: Noticed this commend was wrong, as the class did not extend the not found interface but the container exception interface. Fixed.

ChristophWurst (approved):

susnux:
I fixed that issue, but that CI still fails with some issues on the encryption, which seem definitely unrelated, but I rebased onto current master to hopefully fix that drone nodb  check.

susnux:
Well the remaining drone issues seem to be unrelated, as the same issues are reported for every PR currently ...

nickvergessen:
> There was 1 failure:
> 1) Test\AppFramework\Utility\SimpleContainerTest::testNothingRegistered
> Failed asserting 
that OCP\AppFramework\QueryException Object (...) is an instance of 
interface "Psr\Container\NotFoundExceptionInterface".
> 
> /drone/src/tests/lib/AppFramework/Utility/SimpleContainerTest.php:87

nickvergessen:
For the encryption error just rebase, the above error is your code change and needs fixing/adjusting

susnux:
@nickvergessen thank you! I forgot to handle querying with autoloading enabled, this is fixed now and the test passes.

lib/composer/composer/autoload_classmap.php
```diff
@@ -720,6 +720,7 @@
     'OC\\AppFramework\\Services\\AppConfig' => $baseDir . '/lib/private/AppFramework/Services/AppConfig.php',
     'OC\\AppFramework\\Services\\InitialState' => $baseDir . '/lib/private/AppFramework/Services/InitialState.php',
     'OC\\AppFramework\\Utility\\ControllerMethodReflector' => $baseDir . '/lib/private/AppFramework/Utility/ControllerMethodReflector.php',
+    'OC\\AppFramework\\Utility\\QueryNotFoundException' => $baseDir . '/lib/private/AppFramework/Utility/QueryNotFoundException.php',
```
ChristophWurst, line 723: nit: this is the only change in the autoloader files you need to commit

susnux: There was a drone error:
```
The autoloaders are not up to date
Please run: bash build/autoloaderchecker.sh
```
Just committed all changed files, but I can undo this

nickvergessen:
Something to document in https://github.com/nextcloud/server/issues/34692 ?

PR #3966, opened 2017-03-21
Override config.php values through environment variables 

MorrisJobke:
* added functionality to override config.php values with 'OC_' prefixed environment variables
* use getenv to read environment variables since apache does not set $_ENV variables, fixed test
* fixes #3926 
* downstream of https://github.com/owncloud/core/pull/26570

@LukasReschke @nickvergessen @rullzer @karlitschek Should we rename the env variables to NC_ or should leave them?

PR #36428, opened 2023-01-29
add option to disable db user creation trough environment variable

icewind1991:
When providing root/superuser db credentials, instead of using them as-is Nextcloud uses them to create a new non-root/superuser user.

While this can be useful to protect admins against accidentally running with elevated db access, this also interferes with some setup mechanisms that want more control over what db user is used. Currently the only option for these systems is to patch out the behavior themselves (see for [NixOs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/servers/nextcloud/patches/v26/0001-Setup-remove-custom-dbuser-creation-behavior.patch) for example).

This adds a way to disable this behavior by setting the `setup_create_db_user` config variable to `false`, this can also be done trough an environment variable (`NC_setup_create_db_user=false`) for the setup command.

### DIFF
--- a/lib/private/Setup/AbstractDatabase.php
+++ b/lib/private/Setup/AbstractDatabase.php
@@ -57,6 +57,8 @@ abstract class AbstractDatabase {
 	protected $logger;
 	/** @var ISecureRandom */
 	protected $random;
+	/** @var bool */
+	protected $tryCreateDbUser;
 
 	public function __construct(IL10N $trans, SystemConfig $config, LoggerInterface $logger, ISecureRandom $random) {
 		$this->trans = $trans;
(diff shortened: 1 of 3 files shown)


come-nc (approved):

lib/private/Setup/AbstractDatabase.php
```diff
@@ -88,6 +90,8 @@ public function initialize($config) {
 		$dbPort = !empty($config['dbport']) ? $config['dbport'] : '';
 		$dbTablePrefix = isset($config['dbtableprefix']) ? $config['dbtableprefix'] : 'oc_';
 
+		$this->tryCreateDbUser = getenv("SETUP_CREATE_DB_USER") !== "false";
```
juliushaertl, line 93: Any reason for not using a config.php value for this? Could also be set through env variables then with https://github.com/nextcloud/server/pull/3966

juliusknorr (approved):
Code looks good, non blocking question inline

lib/private/Setup/AbstractDatabase.php
icewind1991: Good catch, I forgot about the config env variables

szaimen:
This was not documented, was it?

joshtrichards:
> This was not documented, was it?

Indeed. Created tracker in docs repo for the time being: nextcloud/documentation#11858

PR #36453, opened 2023-01-31
Fix MigrateBackgroundImages on oracle

juliusknorr:
## Summary

https://github.com/nextcloud/server/blob/09099cfaeceada44dd4454a400d4753e2cc2f0b0/apps/theming/lib/Jobs/MigrateBackgroundImages.php#L95 fails on oracle as configvalue is a longtext which is not allowed in where clauses without casting.

This is the same approach also used in https://github.com/nextcloud/server/blob/361e69f19fd3af632452dbe586065ab42103cf47/lib/private/AllConfig.php#L497-L499

## Checklist

### DIFF
--- a/apps/theming/lib/Jobs/MigrateBackgroundImages.php
+++ b/apps/theming/lib/Jobs/MigrateBackgroundImages.php
@@ -30,6 +30,7 @@ use OCA\Theming\AppInfo\Application;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\BackgroundJob\IJobList;
 use OCP\BackgroundJob\QueuedJob;
+use OCP\DB\QueryBuilder\IQueryBuilder;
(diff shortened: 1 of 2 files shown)


juliushaertl:
Pushed another commit to test if passing PARAM_STR to the eq function also works, as this should cast also looking at the implementation of the OCIExpressionBuilder

nickvergessen:
> Pushed another commit to test if passing PARAM_STR to the eq function also works, as this should cast also looking at the implementation of the OCIExpressionBuilder

Should adjust the runtime code now as well?

skjnldsv (approved):

nickvergessen (approved):

juliushaertl:
Failures unrelated.

juliushaertl:
/backport ecccff8bd232b579f08bc7fe387a5379cc6436cd to stable25

PR #36455, opened 2023-01-31
Reset theming in cypress test

artonge:
The test in `admin-settings` does not reset the theming. So when it is run before `user-background`, the `"Default cloud background is not rendered"` test fails.

This makes sure that the theming is reset before running `"User default background settings"`.

@skjnldsv please tell me if I missed something.

Here are two examples:

- :red_circle:  Failing: https://github.com/nextcloud/server/actions/runs/4044870936/jobs/6974051507 because both tests are run one after the other.
- :heavy_check_mark: Succeeding: https://github.com/nextcloud/server/actions/runs/4053245003/jobs/6973699491 because both tests are run separately.

### DIFF
--- a/cypress/e2e/theming/user-background.cy.ts
+++ b/cypress/e2e/theming/user-background.cy.ts
@@ -19,15 +19,18 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>.
  *
  */
-import type { User } from '@nextcloud/cypress'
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

skjnldsv:
Nice catch!

Pytal (approved):

PR #36462, opened 2023-01-31
fix(theming): Theme specific variables must be set on the root node

### LINKED ISSUE #3369: Black table on light theme
**Describe the bug**
If light theme is chosen, tables are shown in black.

**To Reproduce**
1. Choose light theme
2. Open a markdown (text) file
3. Create a table

**Expected behavior**
Table should be "normally" visible (black lines/letters on white ground).

**Screenshots**
Black table

**Client details:**
 - OS: Ubuntu 22.04
 - Browser: Firefox
 - Version: 106.0.2
 - Device: Desktop

<details>
<summary>Server details</summary>

**Text app version:** 3.6.0

**Operating system:** Ubuntu 20.04

**Web server:** apache2 (2.4.54)

**Database:** mariadb (Ver 15.1 Distrib 10.3.34-MariaDB)

**PHP version:** 7.4.32

**Nextcloud version:** 25.0.1

</details>

### LINKED ISSUE #36303: Theming: Scoping issue when overriding CSS variables for user-themes
To prevent confusion, here a note on the nomenclature I use:
_"user theme"_ → theme enabled by the individual user in their personal settings
_"nextcloud theme"_ → instance-wide design customization which is enabled by the administrator

**TL;DR** I propose changing the element the user theme data attributes are set to the root `html` element to solve scoping issues.

I'm developing a nextcloud theme. It should work with the inbuilt user themes (light, dark, light-highcontrast, dark-highcontrast).

In order to do this I am using a SASS build-process to reduce the amount of duplicate code and still catch all the possible edge-cases. (personal user account settings vs. inbuilt browser settings/media queries).

Example SASS mixin for theming:

```scss
@mixin dark-mode {
  @media (prefers-color-scheme: dark) {
    body:not([data-theme-light], [data-theme-light-hightcontrast]) {
      // browser says this is dark mode, nextcloud user settings do not disagree
      @content;
    }
  }
  body[data-theme-dark],
  body[data-theme-dark-highcontrast] {
    // nextcloud user settings say this should be dark mode
    @content;
  }
}
```

As you can see I'm using the `body` attribute selector to catch the user theme-preference (e.g. `body[data-theme-dark]`). This is sub optimal since this limits the customization to elements within the `body` scope. (For example the `html` tag has a background color, this is visible when the browser window is resized).

I propose the theme attributes are set on the `html` to fix this problem.

The code which would need to be adjusted lies in https://github.com/nextcloud/server/blob/master/apps/theming/src/UserThemes.vue#L240

As I have no deeper knowledge of the nextcloud code base I cannot foresee which problems/conflicts might arise from such a change.

susnux:
* Resolves: https://github.com/nextcloud/server/issues/36303
* <del>Resolves: https://github.com/nextcloud/text/issues/3369</del>

## Summary
Currently if you use a css variable which links to an other css variable, like `--color-main-background` and a theme is enabled, the theme is not used for the variable. See linked issue (same e.g. on forms, see https://github.com/nextcloud/forms/pull/1471#pullrequestreview-1277505842).

The problem is the css variable evaluation order, e.g. consider the browser requests dark color theme but the user selected the bright nextcloud theme, then:
* HTML tag
  * defines `--color-main-background` for dark color theme (as requested by the browser)
  * defines `--table-color-background` to be the same as `--color-main-background`
* Body tag
  * redefines `--color-main-background` to match the theme (because `data-theme-light` is set on the body tag)
* Table tag
  * Uses `--table-color-background` which evaluates to `HTML-Tag`::`--color-main-background` and in the context it is set to dark instead of bright.

So the solution is to set `data-theme-light`  on the HTML tag rather than the body tag.

## Screenshots
System uses dark color theme, but nextcloud is configured to use the bright theme.
You can see the `NcSelect` has the wrong background color:
| before (currently) | after (with this PR) |
|---|---|
| dark background on bright color theme, hard to read | correct bright background on bright color theme, good to read |

## Checklist

### DIFF
--- a/apps/theming/src/UserThemes.vue
+++ b/apps/theming/src/UserThemes.vue
@@ -242,13 +242,13 @@ export default {
 			const enabledFontsIDs = this.fonts.filter(font => font.enabled === true).map(font => font.id)
 
 			this.themes.forEach(theme => {
-				document.body.toggleAttribute(`data-theme-${theme.id}`, theme.enabled)
+				document.documentElement.toggleAttribute(`data-theme-${theme.id}`, theme.enabled)
 			})
 			this.fonts.forEach(font => {
-				document.body.toggleAttribute(`data-theme-${font.id}`, font.enabled)
+				document.documentElement.toggleAttribute(`data-theme-${font.id}`, font.enabled)
 			})
 
-			document.body.setAttribute('data-themes', [...enabledThemesIDs, ...enabledFontsIDs].join(','))
+			document.documentElement.setAttribute('data-themes', [...enabledThemesIDs, ...enabledFontsIDs].join(','))
 		},
 
 		/**
--- a/apps/theming/lib/Controller/ThemingController.php
+++ b/apps/theming/lib/Controller/ThemingController.php
@@ -370,7 +370,7 @@ class ThemingController extends Controller {
 		} else {
 			// If not set, we'll rely on the body class
 			$compiler = new Compiler();
-			$compiledCss = $compiler->compileString("[data-theme-$themeId] { $variables $customCss }");
+			$compiledCss = $compiler->compileString(":root[data-theme-$themeId], [data-theme-$themeId] { $variables $customCss }");
 			$css = $compiledCss->getCss();;
 		}
 
--- a/core/templates/layout.user.php
+++ b/core/templates/layout.user.php
@@ -13,7 +13,10 @@ $getUserAvatar = static function (int $size) use ($_): string {
 }
 
 ?><!DOCTYPE html>
-<html class="ng-csp" data-placeholder-focus="false" lang="<?php p($_['language']); ?>" data-locale="<?php p($_['locale']); ?>" translate="no" >
+<html class="ng-csp" data-placeholder-focus="false" lang="<?php p($_['language']); ?>" data-locale="<?php p($_['locale']); ?>" translate="no"
+<?php foreach ($_['enabledThemes'] as $themeId) {
+	p("data-theme-$themeId ");
+}?> data-themes=<?php p(join(',', $_['enabledThemes'])) ?>>
 	<head data-user="<?php p($_['user_uid']); ?>" data-user-displayname="<?php p($_['user_displayname']); ?>" data-requesttoken="<?php p($_['requesttoken']); ?>">
 		<meta charset="utf-8">
 		<title>
@@ -42,9 +45,7 @@ p($theme->getTitle());
 		<?php emit_script_loading_tags($_); ?>
 		<?php print_unescaped($_['headers']); ?>
 	</head>
-	<body id="<?php p($_['bodyid']);?>" <?php foreach ($_['enabledThemes'] as $themeId) {
-		p("data-theme-$themeId ");
-	}?> data-themes=<?php p(join(',', $_['enabledThemes'])) ?>>
+	<body id="<?php p($_['bodyid']);?>">
 	<?php include 'layout.noscript.warning.php'; ?>
 
 		<?php foreach ($_['initialStates'] as $app => $initialState) { ?>


apps/theming/src/UserThemes.vue
```diff
@@ -242,13 +242,13 @@ export default {
...
+				document.documentElement.toggleAttribute(`data-theme-${theme.id}`, theme.enabled)
 			})
 			this.fonts.forEach(font => {
-				document.body.toggleAttribute(`data-theme-${font.id}`, font.enabled)
+				document.documentElement.toggleAttribute(`data-theme-${font.id}`, font.enabled)
 			})
 
-			document.body.setAttribute('data-themes', [...enabledThemesIDs, ...enabledFontsIDs].join(','))
+			document.documentElement.setAttribute('data-themes', [...enabledThemesIDs, ...enabledFontsIDs].join(','))
 		},
```
skjnldsv, line 252: This is only for live preview of the theme. Are we sure this will take precedence over the default `:root` variables from the css themes ?

skjnldsv:
@susnux can you explain what the NcSelect is doing differently that makes our variables behave differently in body than html? :thinking:

susnux:
> @susnux can you explain what the NcSelect is doing differently that makes our variables behave differently in body than html? 🤔

As explained above it uses variables defined on the root, that evaluate to another variable, [see e.g. here](https://github.com/nextcloud/nextcloud-vue/blob/093ce19071b66b076b975255810f6f1642eaf855/src/components/NcSelect/NcSelect.vue#L907).

`--vs-search-input-bg: var(--color-main-background);` is defined on the root node, so it will not pickup the overridden variable but the default one.

skjnldsv:
Ok, let me explain my concerns (onot founded/verified yet)

When loading the page, we load (in order)
1. the default `light` theme on `:root`
2. the `light`, `dark`,  `highcontrast-light` and `highcontrast-dark` with appropriate media queries
3. the user defined theme with `[data-theme-xxx]` attribute on body
    image


The goal was to ensure we have priorities and fallback. Since body css vars would always be superior to `:root` this ensure that:
1. if no user theme is defined (`body[data-theme-xxx]`), it fallback to any matching media query (`prefer-color-scheme` or `prefers-contrast`)
2. if no media query match, we fallback to the default `:root` light theme

I'm trying to ensure this doesn't create regressions and trouble the css priority.
I don't have much time to fully review this in depth and testing all uses cases :)

susnux:
The variables from the theme should have precedence over the default ones, because they are one level more specific (because of the data tag).

apps/theming/src/UserThemes.vue
susnux: This PR basically moves the data tags to the root node, so this is just the same for the user preview. So this should work as expected.

skjnldsv:
> To prevent regression from loading order the theming css should be `:root[data-theme-...]`

No, because that would prevent us from applying themes to other areas of the page (e.g., the Viewer is always in dark mode, whatever theme is currently active)

susnux:
> No, because that would prevent us from applying themes to other areas of the page (e.g., the Viewer is always in dark mode, whatever theme is currently active)

More like `:root[data-theme], [data-theme] {...}`, this allows to set the theme for different areas while still override the root variables if necessary.

susnux:
It can be fixed on the `NcSelect`, especially to fix apps for older nextcloud versions.
But having the theming not overriding the root variables will create a pitfall for app developers as it is quite common to use `:root { variables}` when defining variables, e.g. I never saw `body { variables}` or similar.
Currently I only know of the `NcSelect` and the text app issue, but  I guess more issues like this will appear over time.

susnux:
> https://github.com/nextcloud/nextcloud-vue/blob/093ce19071b66b076b975255810f6f1642eaf855/src/components/NcSelect/NcSelect.vue#L907 Use the root selector (`.select` it seems) of the component instead of `:root` ?

Some of the variables are for styling the options list, which is not a child of the `select` but a modal, so styling must be applied to at least `body`.

susnux:
I think we can close this

PR #28438, opened 2021-08-15
Don't die with LockedException when removing/restoring multiple files from trash

csware:
fixes issue #16491

TODO: Maybe in medium term the whole restore/delete from trash activity should be guarded by a lock.

PR #36465, opened 2023-02-01
Avoid locking on restore/delete and limit concurrent requests

juliusknorr:
## Summary

This is a frontend safeguard to avoid running into backend file locks when deleting or restoring multiple files. Even after https://github.com/nextcloud/server/pull/28438/files#diff-5450a6734b54a8b5e2c2f7521098ba6d05993e8bbebffeee140e93ec349e4127R978 this could still happen if the retry timeout was exceeding 15 seconds.

## Checklist


skjnldsv (approved):

szaimen (approved):
🐘

icewind1991:
I think it's fine to change the scanning logic to only scan `'files_trashbin/versions/'. $filename` instead of `'files_trashbin/versions/'`, which should help prevent conflicts

icewind1991:
Note that I'm not sure if the scan logic is still needed, I know that in the past we didn't actively maintain the cache outside of `/$user/files` but I believe that is no longer the case

juliushaertl:
> Note that I'm not sure if the scan logic is still needed, I know that in the past we didn't actively maintain the cache outside of `/$user/files` but I believe that is no longer the case

I had the same assumption in https://github.com/nextcloud/server/pull/28438/files#r691270467 so should we just drop it?

juliushaertl:
Deleting and restoring versions seems to still work fine with that code part removed.

icewind1991:
Do we want to increase the concurrency a bit if we remove the scan?

juliushaertl:
Increased to 4 which seems a reasonable value for now.

icewind1991 (approved):

PR #32572, opened 2022-05-24
Fix a bunch of deprecation in the phpunit for core

CarlSchwan:

PR #36497, opened 2023-02-02
Fix autotest.sh execution

nickvergessen:
Some regressions from https://github.com/nextcloud/server/pull/32572


come-nc (approved):
:elephant:

blizzz (approved):

PR #36068, opened 2023-01-10
extend fix-key-location to handle cases from broken cross-storage moves

icewind1991:
Fixes files not being decrypted when moved cross-storage due to (https://github.com/nextcloud/server/pull/35894) (ab)using the fact that the encrypted flag wasn't cleared for them (https://github.com/nextcloud/server/pull/35961) to find these cases.
If a working key can be found for the files they are decrypted in place, leaving the encrypted file as a backup to reduce chances of errors leading to further data loss.

This currently only works if the file hasn't been renamed after being moved because the file name is being used to try and find possible encryption keys. A possible further step would be to add an option to check every available key, covering more cases at the cost of performance.

A special case was also added for when encryption is enabled for the storage after encrypted files have been moved. In those cases the found key is copied to the correct location.

The "fix-key-location" name of the command isn't fully applicable anymore so it might make sense to give it a new name.

---

To test:

1. setup an instance with encryption and a groupfolder
2. move an encrypted file from the home storage to the groupfolder
3. try to open the file
4. run `occ encryption:fix-key-location <user>`
5. try to open the file again

Alternatively add step 3.5 `occ config:app:set groupfolders enable_encryption --value true` to enable groupfolder encryption

---

Backport note: this currently reuses the normal encryption logic to determine which mountpoints are "system" mountpoints, previous nc versions didn't handle groupfolders for that properly. So some extra logic for that will probsbly need to be added for backports.

PR #36526, opened 2023-02-03
Add repair command to search for encryption keys

icewind1991:
For cases where for any reason the relation between encrypted file and encrypted key is lost, this allows trying all keys in an attempt to find one that works. This is slow but allows recovery in cases where no other option would work

- [ ] Based on https://github.com/nextcloud/server/pull/36068


come-nc:
I’d like a guided tour for this one it’s big and complicated

PR #36534, opened 2023-02-04
[F2V] migrate  files_trashbin to vue

skjnldsv:
image

- [x] Base API
- [x] Fix and update side libraries
- [x] Implement store
  - [x] Fix reactivity issue :(
- [x] Implement initial view
  - [x] Virtual scrolling
  - [x] Table mode
- [x] Selection handling
- [x] Very basic action (without API for now ?)
- [x] Accessibility
- [ ] Testing

### Needs:
- [x] https://github.com/nextcloud/nextcloud-files/pull/608
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3936
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3927

### Follow-up:
- [ ] Improve loading indicator for actions
- [ ] Deepen virtual scrolling performances
- [ ] Implement keyboard arrow/pageUp-Down navigation and focus management
- [ ] Hide some columns on small screens


apps/files/lib/Controller/ApiController.php
```diff
@@ -417,4 +428,22 @@ public function getNodeType($folderpath) {
 		$node = $this->userFolder->get($folderpath);
 		return $node->getType();
 	}
+	
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 */
+	public function serviceWorker(): StreamResponse {
+		$response = new StreamResponse(__DIR__ . '/../../../../dist/preview-service-worker.js');
```
come-nc, line 437: Looks ugly, I cannot find other occurences of this in our code, we have no better way of referencing the dist folder?

skjnldsv: Not that I'm aware off lol.
That's the issue with all those nested files :wink: 

We could maybe have something like getting the instance root and build from that, but tbh, I don't think it's that much worth it :grimacing:

apps/files_trashbin/src/css/trashbin.css
```diff
@@ -0,0 +1,3 @@
+.files-list__row-trashbin-deleted {
```
artonge, line 1: Unecessary?

apps/files_trashbin/src/actions/restoreAction.ts
```diff
@@ -0,0 +1,77 @@
...
+					destination,
+				},
+			})
+
+			// Let's pretend the file is deleted since
+			// we don't know the restored location
+			emit('files:file:deleted', node)
+			return true
+		} catch (error) {
+			console.error(error)
```
artonge, line 67: logger

apps/files_trashbin/src/actions/restoreAction.ts
```diff
@@ -0,0 +1,77 @@
...
+import { translate as t } from '@nextcloud/l10n'
+import axios from '@nextcloud/axios'
+import History from '@mdi/svg/svg/history.svg?raw'
+
+import { registerFileAction, FileAction } from '../../../files/src/services/FileAction.ts'
+
+registerFileAction(new FileAction({
+	id: 'restore',
+	displayName() {
+		return t('files_trashbin', 'Restore')
```
artonge, line 35: "Restore all" ?

apps/files/src/components/FileEntry.vue
```diff
@@ -0,0 +1,486 @@
...
+			default: false,
+		},
+		isSizeAvailable: {
+			type: Boolean,
+			default: false,
+		},
+		source: {
+			type: Object,
+			required: true,
+		},
```
artonge, line 156: What kind of object is source?

apps/files/src/components/FileEntry.vue
```diff
@@ -0,0 +1,486 @@
...
+				url.searchParams.set('y', '32')
+				// Handle cropping
+				url.searchParams.set('a', this.cropPreviews === true ? '1' : '0')
+				return url.href
+			} catch (e) {
+				return null
+			}
+		},
+
+		mimeUrl() {
```
artonge, line 272:
```suggestion
		mimeIconUrl() {
```

apps/files/src/components/FileEntry.vue
```diff
@@ -0,0 +1,486 @@
...
+		},
+	},
+
+	watch: {
+		active(active, before) {
+			if (active === false && before === true) {
+				this.resetState()
+
+				// When the row is not active anymore
+				// remove the tabindex from the row
```
artonge, line 309: This is about display, not tabindex, or am I missing something?

apps/files/src/components/FileEntry.vue
```diff
@@ -0,0 +1,486 @@
...
+				return
+			}
+
+			// If any image is being processed, reset it
+			if (this.previewPromise) {
+				this.clearImg()
+			}
+
+			// Ensure max 5 previews are being fetched at the same time
+			const controller = new AbortController()
```
artonge, line 377: When is controller used? It is aborted, but not passed to anything right?

apps/files/src/components/FileEntry.vue
artonge: This is about display, not tabindex

apps/files/src/components/FileEntry.vue
```diff
@@ -0,0 +1,486 @@
...
+		},
+
+		async onActionClick(action) {
+			const displayName = action.displayName([this.source], this.currentView)
+			try {
+				this.loading = action.id
+				await action.exec(this.source, this.currentView)
+			} catch (e) {
+				logger.error('Error while executing action', { action, e })
+				showError(this.t('files', 'Error while executing action "{displayName}"', { displayName }))
```
artonge, line 443: Will never be run as action code catch every error without throwing anything.

apps/files_trashbin/src/actions/restoreAction.ts
```diff
@@ -0,0 +1,77 @@
...
+		// Only available in the trashbin view
+		if (view.id !== 'trashbin') {
+			return false
+		}
+
+		// Only available if all nodes have read permission
+		return nodes.length > 0 && nodes
+			.map(node => node.permissions)
+			.every(permission => (permission & Permission.READ) !== 0)
+	},
```
artonge, line 49: Is it enabled on a per file basis or for all files?
If per file, then you can't restore all files if one file do not have read permission?

apps/files_trashbin/src/services/trashbin.ts
```diff
@@ -0,0 +1,98 @@
...
+		<oc:size />
+		<d:getcontentlength />
+	</d:prop>
+</d:propfind>`
+
+
+const resultToNode = function(node: FileStat): File | Folder {
+	const permissions = parseWebdavPermissions(node.props?.permissions)
+	const owner = getCurrentUser()?.uid as string
+	const previewUrl = generateUrl('/apps/files_trashbin/preview?fileId={fileid}&x=32&y=32', node.props)
```
artonge, line 56: As mentionned, add the etag?

apps/files_trashbin/src/services/trashbin.ts
```diff
@@ -0,0 +1,98 @@
...
+			previewUrl,
+		},
+	}
+
+	return node.type === 'file'
+		? new File(nodeData)
+		: new Folder(nodeData)
+}
+
+export default async (path: string = '/'): Promise<ContentsWithRoot> => {
```
artonge, line 81: Name the function for better debugging and readability

apps/files_trashbin/src/services/client.ts
```diff
@@ -0,0 +1,33 @@
...
+ */
+import { createClient } from 'webdav'
+import { generateRemoteUrl } from '@nextcloud/router'
+import { getCurrentUser, getRequestToken } from '@nextcloud/auth'
+
+export const rootPath = `/trashbin/${getCurrentUser()?.uid}/trash`
+export const rootUrl = generateRemoteUrl('dav' + rootPath)
+const client = createClient(rootUrl, {
+	headers: {
+		requesttoken: getRequestToken(),
```
artonge, line 30: No need for `'X-Requested-With': 'XMLHttpRequest'` here?
Not sure why it is needed in other places.

apps/files/src/views/Navigation.vue
```diff
@@ -151,7 +160,16 @@ export default {
 
 	watch: {
 		currentView(view, oldView) {
-			logger.debug('View changed', { id: view.id, view })
+			// If undefined, it means we're initializing the view
+			// This is handled by the legacy-view:initialized event
+			if (view?.id === oldView?.id) {
```
artonge, line 165: `// TODO: remove when legacy views are dropped` ?

apps/files/src/components/CustomElementRender.vue
```diff
@@ -0,0 +1,59 @@
...
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+  -
+  -->
+<template>
+	<span />
+</template>
+
+<script>
+export default {
+	name: 'CustomElementRender',
```
artonge, line 28: Needed to render inline action and headers ?
A comment would help :)

apps/files/src/components/FilesListHeader.vue
```diff
@@ -0,0 +1,222 @@
...
+			selectionStore,
+			sortingStore,
+		}
+	},
+
+	computed: {
+		...mapState(useSortingStore, ['filesSortingConfig']),
+
+		/** @return {Navigation} */
+		currentView() {
```
artonge, line 123:
```suggestion
		currentView(): Navigation {
```

apps/files/src/components/FilesListHeader.vue
```diff
@@ -0,0 +1,222 @@
...
+import NcCheckboxRadioSwitch from '@nextcloud/vue/dist/Components/NcCheckboxRadioSwitch.js'
+import Vue from 'vue'
+
+import { useFilesStore } from '../store/files.ts'
+import { useSelectionStore } from '../store/selection.ts'
+import { useSortingStore } from '../store/sorting.ts'
+import FilesListHeaderActions from './FilesListHeaderActions.vue'
+import FilesListHeaderButton from './FilesListHeaderButton.vue'
+import logger from '../logger.js'
+import Navigation from '../services/Navigation.ts'
```
artonge, line 80:
```suggestion
import type Navigation from '../services/Navigation.ts'
```

apps/files/src/views/FilesList.vue
```diff
@@ -0,0 +1,361 @@
...
+
+				// Update store
+				this.filesStore.updateNodes(contents)
+
+				// Define current directory children
+				folder.children = contents.map(node => node.attributes.fileid)
+
+				// If we're in the root dir, define the root
+				if (dir === '/') {
+					console.debug('files', 'Setting root', { service: currentView.id, folder })
```
artonge, line 279: logger

apps/files/src/views/FilesList.vue
```diff
@@ -0,0 +1,361 @@
...
+					this.filesStore.setRoot({ service: currentView.id, root: folder })
+				} else
```
artonge, line 281: No {} around else logic ?

apps/files/js/app.js
```diff
@@ -117,7 +116,9 @@
 						},
 					],
 					sorting: {
-						mode: $('#defaultFileSorting').val(),
+						mode: $('#defaultFileSorting').val() === 'basename'
+							? 'name'
+							: $('#defaultFileSorting').val(),
```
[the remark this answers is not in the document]
skjnldsv: it's only for legacy compatibility, as the other views will be added, the old name will fade off
The settings also changed on the backend, so the old name will be obsolete anyway

apps/files/js/filelist.js
```diff
@@ -2181,8 +2181,10 @@
 
 			if (persist && OC.getCurrentUser().uid) {
 				$.post(OC.generateUrl('/apps/files/api/v1/sorting'), {
-					mode: sort,
-					direction: direction
+					// Compatibility with new files-to-vue API
+					mode: sort === 'name' ? 'basename' : sort,
```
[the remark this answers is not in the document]
skjnldsv: `basename` is the right one, name is the old one.
But considering the new backend config, we should not get "name" from the backend config anyway, this is just a legacy failsafe

apps/files/lib/Controller/ApiController.php
```diff
@@ -279,20 +281,29 @@ public function getStorageStats($dir = '/'): JSONResponse {
 	 *
 	 * @param string $mode
 	 * @param string $direction
-	 * @return Response
+	 * @return JSONResponse
 	 * @throws \OCP\PreConditionNotMetException
 	 */
-	public function updateFileSorting($mode, $direction) {
-		$allowedMode = ['name', 'size', 'mtime'];
```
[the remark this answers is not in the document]
skjnldsv: No because the new API allow apps to register their custom columns and provide a sort function.
Like the trashbin deleted column. And it means you can decide to sort by this custom column ID which the backend is not aware off.
I think it's alright, worth case the data is invalid and the front will fallback to the view default sorting key

apps/files/src/actions/deleteAction.ts
```diff
@@ -0,0 +1,64 @@
...
+		try {
+			await axios.delete(node.source)
+
+			// Let's delete even if it's moved to the trashbin
+			// since it has been removed from the current view
+			//  and changing the view will trigger a reload anyway.
+			emit('files:file:deleted', node)
+			return true
+		} catch (error) {
+			console.error(error)
```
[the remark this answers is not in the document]
skjnldsv: Show Error is handled by the files list. If at least 1 item is false, it means something failed and the font will:
- Tell the user
- Adjust the current selection to remove the successful elements and keep only the one that failed

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -0,0 +1,89 @@
...
+		},
+	},
+
+	computed: {
+		dirs() {
+			const cumulativePath = (acc) => (value) => (acc += `${value}/`)
+			const paths = this.path.split('/').filter(Boolean).map(cumulativePath('/'))
+			// Strip away trailing slash
+			return ['/', ...paths.map(path => path.replace(/^(.+)\/$/, '$1'))]
+		},
```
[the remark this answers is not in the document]
skjnldsv: Absolutely! It generate a cumulative path from a path:
```js
const path = '/this/is/a/path'
const dirs =  ['/', '/this', '/this/is', '/this/is/a', '/this/is/a/path']
```

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -0,0 +1,89 @@
...
+				}
+			})
+		},
+	},
+
+	methods: {
+		onClick(to) {
+			if (to?.query?.dir === this.$route.query.dir) {
+				this.$emit('reload')
+			}
```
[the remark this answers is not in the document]
skjnldsv: no, it's already handled by the `router-link` element :)

apps/files/src/components/FileEntry.vue
```diff
@@ -0,0 +1,486 @@
...
+
+		previewUrl() {
+			try {
+				const url = new URL(window.location.origin + this.source.attributes.previewUrl)
+				// Request tiny previews
+				url.searchParams.set('x', '32')
+				url.searchParams.set('y', '32')
+				// Handle cropping
+				url.searchParams.set('a', this.cropPreviews === true ? '1' : '0')
+				return url.href
```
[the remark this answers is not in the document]
skjnldsv: Will do as followup! :+1:

apps/files/src/services/PreviewService.ts
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+// The preview service worker cache name (see webpack config)
+const SWCacheName = 'previews'
+
+/**
+ * Check if the preview is already cached by the service worker
+ */
+export const isCachedPreview = function(previewUrl: string) {
+	return caches.open(SWCacheName)
```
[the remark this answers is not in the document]
skjnldsv: I learned that recently too :see_no_evil:

apps/files/src/store/paths.ts
```diff
@@ -0,0 +1,86 @@
...
+				if (!this[payload.service]) {
+					Vue.set(this, payload.service, {})
+				}
+
+				// Now we can set the provided path
+				Vue.set(this[payload.service], payload.path, payload.fileid)
+			},
+
+			onCreatedNode() {
+				console.debug(arguments)
```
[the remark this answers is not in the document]
skjnldsv: I should totally remove or comment those, it's a mistake :wink:

apps/files/src/store/files.ts
```diff
@@ -0,0 +1,115 @@
...
+
+			onMovedNode() {
+				console.debug(arguments)
+			},
+		}
+	})
+
+	const fileStore = store()
+	// Make sure we only register the listeners once
+	if (!fileStore.initialized) {
```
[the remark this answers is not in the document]
skjnldsv: I was afraid of making it an official prop in that case, but I can rename and declare it as `_initialized` if necessary :thinking:

apps/files/src/store/sorting.ts
```diff
@@ -0,0 +1,80 @@
...
+const filesSortingConfig = loadState('files', 'filesSortingConfig', {}) as SortingStore
+
+export const useSortingStore = defineStore('sorting', {
+	state: () => ({
+		filesSortingConfig,
+	}),
+
+	getters: {
+		isAscSorting: (state) => (view: string = 'files') => state.filesSortingConfig[view]?.direction !== 'desc',
+		getSortingMode: (state) => (view: string = 'files') => state.filesSortingConfig[view]?.mode,
```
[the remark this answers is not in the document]
skjnldsv: Asc is the direction, the default mode depends on the view (trashbin is the `deleted` column)

apps/files/src/mixins/fileslist-row.scss
```diff
@@ -0,0 +1,170 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+/**
+ * ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠
+ * This file is for every column styling that must be
+ * shared between BOTH the files list AND the list header.
+ * ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠ ⚠
+ */
```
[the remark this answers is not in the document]
skjnldsv: Because that would duplicate the code a lot or I'd have to use v-deep everywhere.
I figured it was cleaner to have one scss imported in multiple files? No? :see_no_evil:

apps/files_trashbin/src/actions/restoreAction.ts
skjnldsv: Even if there is a selection, you cannot know if all the files are selected.
So I'd leave Restore for now as this is the current behaviour on old Files

apps/files/src/components/FileEntry.vue
skjnldsv: Old comment! Nice catch :)

apps/files/src/components/FileEntry.vue
skjnldsv: Leftover from another ongoing work (different PR)

apps/files_trashbin/src/actions/restoreAction.ts
skjnldsv: You cannot ensure all devs will comply and do it properly :wink: 
This is a safety measure

apps/files/src/components/FileEntry.vue
skjnldsv: But we do need to do the same as the batch action and warn the user, this is an issue, thanks for catching it!

apps/files_trashbin/src/actions/restoreAction.ts
skjnldsv: Yep, currently the select works like that.
We can improve later on by warning the user or something I don't know.
But if we allow an `exec` on a node that doesn't have the requirement, it will obviously fail. Not sure we should let the user do that. Either you can do it for the whole selection, or not at all. Doing it partially might lead to unexpected behaviour and issues for the user :)

apps/files_trashbin/src/services/trashbin.ts
skjnldsv: Will do as followup

apps/files_trashbin/src/services/client.ts
skjnldsv: Will investigate afterwards

apps/files/src/components/FilesListHeader.vue
skjnldsv: I wish! The ts webpack fails if I use ts typing in the vue script.
I want to investigate afterwards why exactly. This is babel complaining :(

apps/files/src/views/FilesList.vue
skjnldsv: There is a bit bellow. There is a comment in between.
Otherwise the comment location is weird (that's why I hate the else/if indent syntax :see_no_evil:

apps/files/src/components/FileEntry.vue
skjnldsv: ... as a followup lol

babel.config.js
```diff
@@ -10,6 +10,7 @@ module.exports = {
 			'@babel/preset-env',
 			{
 				useBuiltIns: false,
+				modules: 'auto',
```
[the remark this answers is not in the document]
skjnldsv: I don't recall, I was getting errors with TS iirc :thinking:

apps/comments/src/views/Comments.vue
```diff
@@ -206,25 +206,25 @@ export default {
...
+				const { request, abort } = cancelableRequest(getComments)
+				this.cancelRequest = abort
 
 				// Fetch comments
-				const comments = await request({
+				const { data } = await request({
 					commentsType: this.commentsType,
 					ressourceId: this.ressourceId,
-				}, { offset: this.offset })
+				}, { offset: this.offset }) || []
```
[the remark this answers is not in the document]
skjnldsv: Because it's coming from the webdav lib. I would have to reassign to a 

```js
// Fetch comments
const { data } = await request({
	commentsType: this.commentsType,
	ressourceId: this.ressourceId,
}, { offset: this.offset }) || []
```

I guess I could have done `const { data: comments } = await request()` :wink:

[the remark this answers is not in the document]
artonge: Oh yes, offcourse.

What about `|| { data: []}` instead of `|| []` ?

apps/files/src/views/FilesList.vue
artonge: So it is an intentional `else if` ? Feels really less readable that if it was on the same line. I would place the comment inside the logic.

apps/files_trashbin/src/actions/restoreAction.ts
artonge: Could we filter out the node for which the action is not enabled?

artonge: Having a per file `enabled` method, and filtering out nodes for which the action is not enabled in the exec, would be nice indeed.

apps/comments/src/views/Comments.vue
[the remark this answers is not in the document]
skjnldsv: Aouch, nice catch :+1:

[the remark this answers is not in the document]
skjnldsv: Will push fix

apps/files/src/mixins/fileslist-row.scss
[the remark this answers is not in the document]
artonge: In the current state, the code will be, bear with me, tri-plicated?
I think single :deep should be enough.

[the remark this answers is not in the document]
skjnldsv: So put it in [`FilesListVirtual`](https://github.com/nextcloud/server/blob/495c65060476a5ea40b68bc99f8039067a8bad19/apps/files/src/components/FilesListVirtual.vue#L148-L180) ?

apps/files/src/views/FilesList.vue
skjnldsv: Ahaha, personal preferences I guess :wink:

apps/files/src/components/FilesListHeader.vue
skjnldsv: Not a type :)

apps/files/src/store/files.ts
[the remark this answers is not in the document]
skjnldsv: More like this would suddenly be available in the store state
It's not really a store variable, but a private variable not really related to the components usage of the store

apps/files/src/mixins/fileslist-row.scss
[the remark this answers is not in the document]
artonge: I think so, yes

apps/files/src/store/files.ts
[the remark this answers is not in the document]
artonge: Fair enough, I do not have any strong opinion on it.

apps/files_trashbin/src/actions/restoreAction.ts
skjnldsv: > Having a per file enabled method

Performance could heavily suffer from this. Not sure how other files explorer works :thinking: 
I'll raise the topic after this is merged

juliushaertl:
Looks very nice already :) Just some findings, could also be follow ups
- [ ] Switching back to other file list does not work once entered the trash bin once
- [ ] Previews show a black box
- [ ] Restore/delete permanently a file does not remove it from the list

skjnldsv:
> * Previews show a black box

You mean when it's loading? Because I do see previews :)
image

> * Restore/delete permanently a file does not remove it from the list

It used to... :facepalm: 
I'll investigate in followup

> *  Breadcrumbs show `Foldername.dTIMESTAMP` when browsing a deleted folder, in the old implementation we just showed `Foldername

Yeah, I need to see how I can mitigate this without performance drawback :)

> *   Sorting by size/name doesn't work yet

Works here! :thinking: 
https://user-images.githubusercontent.com/14975046/230409035-3f40beac-0d82-490a-a9d0-c95055de68bc.mp4

szaimen:
Additionally to Julius' findings, I also found something: 
- right-click does not work anymore
- the mobile view is currently completely broken (but already in the top post as I can see)
- clicking on a file in the trashbin downloads the file but with incorrect mimetype ending
- there is at least one warning in the console log when I open the trashbin:
image


Apart from that looks good to me!

skjnldsv:
@szaimen self signed certificate?
Service workers cannot be loaded on self signed, this is a browser issue.

> clicking on a file in the trashbin downloads the file but with incorrect mimetype ending

Can you clarify what do oyou mean with the ending? The extention?
If so this is an issue with the dav backend :wink:  (while we could mitigate with the front)
Will open a ticket afterwards :+1:

szaimen:
> @szaimen self signed certificate?
> Service workers cannot be loaded on self signed, this is a browser issue.

Ah I see. Then this is probably a no-issue.

szaimen:
BTW is  there already some chunking implemented so that only a few files are loaded and then only when scrolling further down? (I have not tested this)...

juliushaertl:
> > * Previews show a black box
> 
> You mean when it's loading? Because I do see previews :) image

Still doesn't after retesting. I have the following console trace, that looks related:

```
PreviewService.ts:30 Uncaught (in promise) ReferenceError: caches is not defined
    at isCachedPreview (PreviewService.ts:30:1)
    at _callee$ (FileEntry.vue?vue&type=script&lang=ts&:269:1)
    at tryCatch (core-common.js:223538:40)
    at Generator.invoke (core-common.js:223773:22)
    at Generator.next (core-common.js:223598:21)
    at asyncGeneratorStep (FileEntry.vue?vue&type=script&lang=ts&:2:1)
    at _next (FileEntry.vue?vue&type=script&lang=ts&:3:1)
    at FileEntry.vue?vue&type=script&lang=ts&:3:1
    at new Promise (<anonymous>)
    at FileEntry.vue?vue&type=script&lang=ts&:3:1
```

 
> > * Sorting by size/name doesn't work yet
> 
> Works here! 🤔

Interesting, retested and works now.

apps/files/src/services/PreviewService.ts
juliushaertl, line 30: > Secure context: This feature is available only in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS), in some or all [supporting browsers](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage?retiredLocale=de#browser_compatibility).

This would mean that preview don't work with http

skjnldsv: No, it means you will not have enhanced caching with service worker on http/self-signed.
It will fallback to default caching.
But I will fix the cache test as a followup! :+1:

juliushaertl: Confirmed, works fine with https

PR #36562, opened 2023-02-06
fix(authtoken): Store only one hash for authtokens with the current password per user

juliusknorr:
Checking auth token hashes takes quite some time. We can avoid checking each of them if we make sure that the same password also uses the same hash so we can quickly confirm that they do not need updates by just verifying one hash.

Waiting for CI at https://drone.nextcloud.com/nextcloud/spreed/11676/1/4

### DIFF
--- a/lib/private/Authentication/Token/PublicKeyTokenMapper.php
+++ b/lib/private/Authentication/Token/PublicKeyTokenMapper.php
@@ -229,4 +229,31 @@ class PublicKeyTokenMapper extends QBMapper {
 			);
 		$update->executeStatement();
 	}
+
+	public function updateHashesForUser(string $userId, string $passwordHash): void {
+		$qb = $this->db->getQueryBuilder();
+		$update = $qb->update($this->getTableName())
+			->set('password_hash', $qb->createNamedParameter($passwordHash))
+			->where(
+				$qb->expr()->eq('uid', $qb->createNamedParameter($userId))
+			);
+		$update->executeStatement();
+	}
+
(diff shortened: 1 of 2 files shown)


marcelklehr (approved):
Nice!

lib/private/Authentication/Token/PublicKeyTokenMapper.php
```diff
@@ -229,4 +229,31 @@ public function updateActivity(IToken $token, int $now): void {
...
+	}
+
+	public function getFirstTokenForUser(string $userId): ?PublicKeyToken {
+		$qb = $this->db->getQueryBuilder();
+		$result = $qb->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('uid', $qb->createNamedParameter($userId)))
+			->setMaxResults(1)
+			->orderBy('id')
+			->executeQuery();
```
nickvergessen, line 250:
```suggestion
		$qb->select('*')
			->from($this->getTableName())
			->where($qb->expr()->eq('uid', $qb->createNamedParameter($userId)))
			->setMaxResults(1)
			->orderBy('id');
		$result = $qb->executeQuery();
```
For clearer type handling and easier debugging :)

nickvergessen:
Talk testing times look good again.

But CI on server is unhappy (unit tests)

nickvergessen:
Also started notifications tests at https://github.com/nextcloud/notifications/pull/1445 to see if they are "green" again.

ChristophWurst (approved):

PR #36565, opened 2023-02-06
[user_ldap] Update profile from LDAP fields

### LINKED ISSUE #32085: Import personal data from LDAP
### How to use GitHub

We use the `LDAP/AD integration` app as a login backend and t is frustrating that our users have to fill in personal informations in their profile when they are already in the attributes of our LDAP.

A simple "Attributes mapping" menu in the `LDAP/AD integration` app where the administrator could link Nextcloud attributes and LDAP fields. For exemple, in our case, we'd like to link the phone number on Nextcloud and the LDAP attribute `mobile`. Likewise for their address, their nickname, their date of birth...

We could create a cron job to synchronize our LDAP attributes with the ones displayed on Nextcloud.
It sound doable by looking at the [occ documentation for users](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/occ_command.html#user-commands) (specifically under the `Set a setting` section).

**Additional context**
We use openLDAP, and Nextcloud 23.0.4.
This feature was already reported in 2017, but was closed in 2019 since the issuer deleted his account. (#4365)

march42:
* Resolves: #32085

## Summary
Implement feature, to update user profile data from LDAP attributes.
Add configuration items for LDAP attribute mapping (Administration - LDAP/AD integration - Advanced - User Profile Attributes)
Add methods for user profile properties to lib/private/User/User.php
Add connections for LDAP attributes to user profile properties

## TODO
- [ ] Testing with various setting and scenario

## Checklist

Screenshot (Nextcloud 2)
Screenshot (1)

EDIT: [documentation PR 9614](https://github.com/nextcloud/documentation/pull/9614)


apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
 		}
 		unset($attr);
 
+		//User profile visibility
+		$profileScope = $this->connection->ldapProfileScope;
+		if (is_null($profileScope) || '' === $profileScope || 'unset' === $profileScope) {
```
come-nc, line 250:
```suggestion
		if ($profileScope === '' || $profileScope === 'unset') {
```
But when is it 'unset' ?

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
...
+			$profileScope = null;
+		}
+		//User Profile Field - Phone number
+		$attr = strtolower($this->connection->ldapAttributePhone);
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_PHONE, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
```
come-nc, line 258:
```suggestion
```
What are these unset for?

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +580,69 @@ private function verifyQuotaValue(string $quotaValue) {
 		return $quotaValue === 'none' || $quotaValue === 'default' || \OC_Helper::computerFileSize($quotaValue) !== false;
 	}
 
+	/**
+	 * fetches values from LDAP and stores it as Nextcloud user value
+	 * @param string $valueFromLDAP if known, to save an LDAP read request
+	 * @return null
+	 */
+	private function updateProfile(string $property, $valueFromLDAP, $scope=null) {
```
come-nc, line 588:
```suggestion
	 */
	private function updateProfile(string $property, $valueFromLDAP, ?string $scope=null): void {
```

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +580,69 @@ private function verifyQuotaValue(string $quotaValue) {
...
+					// setScope(IAccountManager::SCOPE_FEDERATED);
+					// setVerified(IAccountManager::VERIFIED);
+				}
+			} catch (PropertyDoesNotExistException $e) {
+				$this->logger->error('property does not exist: '.$profileProperty.' for user '.$this->getUsername().'', ['app' => 'user_ldap']);
+				return;
+			}
+			$this->logger->debug('property updated: '.$profileProperty.'='.$value.' for user '.$this->getUsername().'', ['app' => 'user_ldap']);
+			$this->config->setUserValue($this->getUsername(), 'user_ldap', $property, $value);
+			return $value;
```
come-nc, line 638:
```suggestion
			return;
```

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +580,69 @@ private function verifyQuotaValue(string $quotaValue) {
...
+		} else {
+			// FIXME: I decided, to leave profile untouched, if attribute gets removed from LDAP
+			$this->config->deleteUserValue($this->getUsername(), 'user_ldap', $property);
+			return '';
```
come-nc, line 642:
```suggestion
			return;
```

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +580,69 @@ private function verifyQuotaValue(string $quotaValue) {
...
+		}
+		if ($valueFromLDAP !== null) {
+			//$propertyValue = (string)$valueFromLDAP;
+			$propertyValue = [$valueFromLDAP];
+		}
+		$this->logger->debug('user profile data ('.$profileProperty.') from LDAP '.$this->dn.' ='.((string)$valueFromLDAP), ['app' => 'user_ldap']);
+		if ($propertyValue && isset($propertyValue[0])) {
+			$value = $propertyValue[0];
+			try {
+				$user = $this->userManager->get($this->uid);
```
come-nc, line 623: For each configured profile property you will fetch the user again. You should prepare an array of profile values to apply and give that to your method, which could then fetch the user only once.

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +580,69 @@ private function verifyQuotaValue(string $quotaValue) {
...
+			} catch (PropertyDoesNotExistException $e) {
+				$this->logger->error('property does not exist: '.$profileProperty.' for user '.$this->getUsername().'', ['app' => 'user_ldap']);
+				return;
+			}
+			$this->logger->debug('property updated: '.$profileProperty.'='.$value.' for user '.$this->getUsername().'', ['app' => 'user_ldap']);
+			$this->config->setUserValue($this->getUsername(), 'user_ldap', $property, $value);
+			return $value;
+		} else {
+			// FIXME: I decided, to leave profile untouched, if attribute gets removed from LDAP
+			$this->config->deleteUserValue($this->getUsername(), 'user_ldap', $property);
```
come-nc, line 641: profile properties need to be set both on user and in config?

lib/private/User/User.php
```diff
@@ -587,4 +588,104 @@ public function triggerChange($feature, $value = null, $oldValue = null) {
 			$this->emitter->emit('\OC\User', 'changeUser', [$this, $feature, $value, $oldValue]);
 		}
 	}
+
+	/**
+	 * @param string $property name of the AccountProperty
+	 * @return string|null AccountProperty value
+	 * @throws InvalidArgumentException when the property name is invalid or null
+	 */
+	public function getProfilePropertyValue($property): ?string {
```
come-nc, line 597: Are you sure all those proxy methods are needed? What prevents user_ldap from getting the account itself?

come-nc (commented):
The feature will be popular I think but the code needs to be refined.
Also we must be very careful about performance in user_ldap, this should have no performance impact when not used.

@Pytal Can you review this? It’s a lot related to the profile system.

apps/user_ldap/lib/User/User.php
march42: habit: cleaning up the memory used, to ease garbage collection
I would remove all but the last, to unclutter it

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
...
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_PHONE, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
+		//User Profile Field - website
+		$attr = strtolower($this->connection->ldapAttributeWebsite);
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_WEBSITE, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
```
march42, line 264:
```suggestion
```

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
...
+		//User Profile Field - Address
+		$attr = strtolower($this->connection->ldapAttributeAddress);
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_ADDRESS, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
```
march42, line 270:
```suggestion
```

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
...
+		//User Profile Field - Twitter
+		$attr = strtolower($this->connection->ldapAttributeTwitter);
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_TWITTER, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
```
march42, line 276:
```suggestion
```

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
...
+		//User Profile Field - organisation
+		$attr = strtolower($this->connection->ldapAttributeOrganisation);
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_ORGANISATION, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
```
march42, line 282:
```suggestion
```

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
...
+		//User Profile Field - role
+		$attr = strtolower($this->connection->ldapAttributeRole);
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_ROLE, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
```
march42, line 288:
```suggestion
```

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +245,60 @@ public function processAttributes($ldapEntry) {
...
+		//User Profile Field - headline
+		$attr = strtolower($this->connection->ldapAttributeHeadline);
+		if (isset($ldapEntry[$attr])) {
+			$this->updateProfile(self::USER_PREFKEY_HEADLINE, $ldapEntry[$attr][0], $profileScope);
+		}
+		unset($attr);
```
march42, line 294:
```suggestion
```

march42 (commented):
cleanup: remove unneeded unset

apps/user_ldap/lib/User/User.php
march42: When choosen in the advanced configuration tab. The administrator chooses profileScope from unset, v2-private, v2-local, v2-federated, v2-published. If using 'unset' it means the user can set these self in the profile.

apps/user_ldap/lib/User/User.php
march42: Sounds good, I will look into this.

lib/public/IUser.php
```diff
@@ -270,4 +271,47 @@ public function getQuota();
...
 	 */
 	public function setQuota($quota);
+
+	/**
+	 * get users' profile property value.
+	 *
+	 * @param string $property name see IAccountManager::PROPERTY_*
+	 * @return string AccountProperty value
+	 * @throws InvalidArgumentException when the property name is invalid or null
+	 * @since 25.0.0
```
Pytal, line 281:
```suggestion
	 * @since 27.0.0
```

lib/public/IUser.php
```diff
@@ -270,4 +271,47 @@ public function getQuota();
...
+	 */
+	public function getProfilePropertyValue(string $property): ?string;
+
+	/**
+	 * get users' profile property scope.
+	 *
+	 * @param string $property name see IAccountManager::PROPERTY_*
+	 * @return string AccountProperty scope IAccountManager::SCOPE_*
+	 * @throws InvalidArgumentException when the property name is invalid or null
+	 * @since 25.0.0
```
Pytal, line 291:
```suggestion
	 * @since 27.0.0
```

lib/public/IUser.php
```diff
@@ -270,4 +271,47 @@ public function getQuota();
...
+	 */
+	public function getProfilePropertyScope(string $property): ?string;
+
+	/**
+	 * get users' profile property verified.
+	 *
+	 * @param string $property name see IAccountManager::PROPERTY_*
+	 * @return string AccountProperty verification status IAccountManager::NOT_VERIFIED/VERIFICATION_IN_PROGRESS/VERIFIED
+	 * @throws InvalidArgumentException when the property name is invalid or null
+	 * @since 25.0.0
```
Pytal, line 301:
```suggestion
	 * @since 27.0.0
```

lib/public/IUser.php
```diff
@@ -270,4 +271,47 @@ public function getQuota();
...
+	/**
+	 * set users' profile property value,scope,verified.
+	 *
+	 * @param string $property name from IAccountManager::PROPERTY_*
+	 * @param string $value AccountProperty value
+	 * @param string $scope AccountProperty scope
+	 * @param string $verified AccountProperty verified
+	 * @return void
+	 * @throws InvalidArgumentException when the property name is invalid or null
+	 * @since 25.0.0
```
Pytal, line 314:
```suggestion
	 * @since 27.0.0
```

lib/private/User/User.php
```diff
@@ -587,4 +588,104 @@ public function triggerChange($feature, $value = null, $oldValue = null) {
 			$this->emitter->emit('\OC\User', 'changeUser', [$this, $feature, $value, $oldValue]);
 		}
 	}
+
+	/**
+	 * @param string $property name of the AccountProperty
+	 * @return string|null AccountProperty value
```
Pytal, line 594:
```suggestion
	 * @return ?string AccountProperty value
```

lib/private/User/User.php
```diff
@@ -587,4 +588,104 @@ public function triggerChange($feature, $value = null, $oldValue = null) {
...
+		//return $this->config->getUserValue($this->uid, 'user_ldap', $property, null);
+		$this->ensureAccountManager();
+		$account = $this->accountManager->getAccount($this);
+		$property = $account->getProperty($property);
+		return $property->getValue();
+	}
+
+	/**
+	 * @param string $property name of the AccountProperty
+	 * @return string|null AccountProperty scope
```
Pytal, line 613:
```suggestion
	 * @return ?string AccountProperty scope
```

lib/private/User/User.php
```diff
@@ -587,4 +588,104 @@ public function triggerChange($feature, $value = null, $oldValue = null) {
...
+		$this->ensureAccountManager();
+		$account = $this->accountManager->getAccount($this);
+		// TODO: this should be stored locally, to reduce database overhead
+		$property = $account->getProperty($property);
+		return $property->getScope();
+	}
+
+	/**
+	 * @param string $property name of the AccountProperty
+	 * @return string|null AccountProperty verified
```
Pytal, line 629:
```suggestion
	 * @return ?string AccountProperty verified
```

apps/user_ldap/lib/User/User.php
march42: No, I remove this. I used it to easily check the data with `occ user:setting` command.

march42:
I reworked the code:

- prepare and propagate an array with the profile data from LDAP and fetching user just once. Like @come-nc suggested.
- added Fediverse profile property
- did some cleanup and simplification
- fixed messages from psalm

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +233,70 @@ public function processAttributes($ldapEntry) {
...
+		if (isset($ldapEntry[$attr])) {
+			$profileValues[\OCP\Accounts\IAccountManager::PROPERTY_BIOGRAPHY] = $ldapEntry[$attr][0];
+		}
+		//User Profile Field - fediverse
+		$attr = strtolower($this->connection->ldapAttributeFediverse);
+		if (isset($ldapEntry[$attr])) {
+			$profileValues[\OCP\Accounts\IAccountManager::PROPERTY_FEDIVERSE] = $ldapEntry[$attr][0];
+		}
+		// Update user profile
+		if(0 < count($profileValues)) {
```
come-nc, line 293:
```suggestion
		if (count($profileValues) > 0) {
```

apps/user_ldap/lib/User/User.php
```diff
@@ -231,6 +233,70 @@ public function processAttributes($ldapEntry) {
...
+			$this->updateProfile($profileValues, $profileScope);
+			unset($profileValues);
+		}
+		unset($profileScope);
+		unset($attr);
```
come-nc, line 298: I’m still not comfortable with those unset, we are not micromanaging memory anywhere else in our code like this.

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +578,52 @@ private function verifyQuotaValue(string $quotaValue) {
...
+	 * @param array $profileValues associaive array of property keys and values from LDAP
+	 * @param string|null $profileScope the scope of visibility to set
+	 */
+	private function updateProfile(array $profileValues, ?string $profileScope=null): void {
+		// check if user profile was refreshed before
+		if ($this->wasRefreshed('profile')) {
+			return;
+		}
+		// check if parameter array is empty
+		if(0 == count($profileValues)) {
```
come-nc, line 593:
```suggestion
		if (count($profileValues) == 0) {
```

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +578,52 @@ private function verifyQuotaValue(string $quotaValue) {
...
+		}
+		// loop through the properties and handle them
+		/** @var string $property the array key (property name from AccountManager class) */
+		/** @var string $valueFromLDAP the value as read from LDAP */
+		foreach($profileValues as $property => $valueFromLDAP) {
+			$this->logger->debug('user profile data ('.$property.') from LDAP '.$this->dn, ['app' => 'user_ldap']);
+			// check and update profile properties
+			/** @var string $propertyValue */
+			$propertyValue = [$valueFromLDAP];
+			if (isset($propertyValue[0])) {
```
come-nc, line 609: This is always true, no?

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +578,52 @@ private function verifyQuotaValue(string $quotaValue) {
...
+		if (is_null($user)) {
+			return;
+		}
+		// loop through the properties and handle them
+		/** @var string $property the array key (property name from AccountManager class) */
+		/** @var string $valueFromLDAP the value as read from LDAP */
+		foreach($profileValues as $property => $valueFromLDAP) {
+			$this->logger->debug('user profile data ('.$property.') from LDAP '.$this->dn, ['app' => 'user_ldap']);
+			// check and update profile properties
+			/** @var string $propertyValue */
```
come-nc, line 607:
```suggestion
			/** @var string[] $propertyValue */
```

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +578,52 @@ private function verifyQuotaValue(string $quotaValue) {
...
+			$propertyValue = [$valueFromLDAP];
+			if (isset($propertyValue[0])) {
+				/** @var string $value */
```
come-nc, line 610:
```suggestion
```

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +578,52 @@ private function verifyQuotaValue(string $quotaValue) {
...
+				$value = $propertyValue[0];
+				try {
+					/** @var string $currentValue */
+						$currentValue = (string)$user->getProfilePropertyValue($property);
+						if ($currentValue !== $value) {
+							$user->setProfileProperty($property,$value,$profileScope,null);
+							$this->logger->debug('property updated: '.$property.'='.$value.' for user '.$this->getUsername().'', ['app' => 'user_ldap']);
+						}
+				} catch (PropertyDoesNotExistException $e) {
+					$this->logger->error('property does not exist: '.$property.' for user '.$this->getUsername().'', ['app' => 'user_ldap']);
```
come-nc, line 620:
```suggestion
					$this->logger->error('property does not exist: '.$property.' for user '.$this->getUsername().'', ['app' => 'user_ldap', 'exception' => $e]);
```

lib/private/User/User.php
come-nc: You should use `IAccountManager::getAccount($user)` to get the account,
then `IAccount::setProperty` for each property, and then `IAccountManager::updateAccount($account)`.

There is no need to add methods to OCP\IUser for this.

come-nc (changes requested):
See my last comment, please go through the accountmanager instead of adding methods to `OCP\User`.

Also, could you please detail how this will behave for multivalue fields from LDAP?

Regarding scope, if I understand correctly the admin can force a given scope for all profile data synced from LDAP?
What happens if the user then changes the scope?

Is it possible for the admin to leave the scope option empty to let users decide the scope of each information?

come-nc:
> I have found, that some commits are not correctly signed-off. Seems to be like all commits through GitHub Desktop. So I would probably best rebase the branch before you finally merge it. Any suggestions?

You can use `rebase` with `--signoff` option.
For instance `git rebase HEAD~25 --signoff` will add a signoff to the last 25 commits.

come-nc:
> Now, you got me a little pissed, because you were unspecific and I think I'm not supposed to read your mind. I'll be specific, so we could find an agreement on how to handle. First of all. I feel unappreciated, even unseen. I created my branch in May 2022, after issue 32085 was opened. see [#32085 (comment)](https://github.com/nextcloud/server/issues/32085#issuecomment-1119313753) I invested time, to package my quick'n'dirty hack and pretty it up, before committing to GitHub. I updated the documentation, run numerous test and checked different scenarios. Nobody seemed to care, until February 2023, when I was asked to make a pull request. 

I’m sorry you feel that way, that was not my intention of course.
I do not follow feature requests in the issues closely as we already have a lot on our plate. So I only became aware of this when the PR was open and I got pinged for a review.
I’m used to reviewing code from collegues and being picky about it, to try to merge the best code as possible. Sorry if that felt like too much complains, your work is appreciated of course. As I previously stated I think this feature will be welcomed by users.

> I don't need this to be integrated. I don't have any intentions on wasting my time or your time. I have two working solutions and patching after update is done in minutes. If this is unwanted or unappreciated, I'm happy with closing and deletion. This is your call to make. After your suggestions two weeks ago I took time, to rework the code and test on current version 25 and version 26 beta 3 and 4. […] I took time, to rewrite the code and even more time to test it and cleanup the coding. Now I have to go back and do it again, because your suggestions were too cautious phrased. I'm not very happy with this and I will have to make time for this next weekend. 

I am happy with the refactoring you did regarding fetching the user and I do not think it overlaps that much with going through the account intead of `OCP\IUser`.
Only the calls to `$user->setProfileProperty` will need to change on user_ldap side.

> 2. What do you want me to do with those unset. I don't want you to be uncomfortable. I did, what I thought was intended. Looking at the original code of processAttributes, for example version 25 starting line 162, you can see a number of unset($attr) in lines 172, 192, 201, 232. I used that present code as guideline for my coding. I am pretty unaffected by wether or not those are present. I’m still not comfortable with those unset, we are not micromanaging memory anywhere else in our code like this. Like I said, You are doing it in your code exactly in the questioned processAttributes function. What should I do to make you comfortable? 3. I will take time, to ckeck and test this specifically and add more clearification to the documentation.

Thank you for the pointers on processAttribute, that’s code from before I joined Nextcloud, I did not see much unset in the rest of the code.
I was just afraid by merging those unset of setting a precedent and having unsets added everywhere in the code. Because I do not understand why these local vars should be unset more than any other.
@blizzz Do you remember what is the rationale behind the `unset` calls in `processAttribute`?

Thank you for your details on scope handling, that sounds quite reasonable.

If this is eating too much of your time I can take it from here if your prefer.

march42:
> I’m sorry you feel that way, that was not my intention of course.
> I’m used to reviewing code from collegues and being picky about it, to try to merge the best code as possible.
> Sorry if that felt like too much complains,

Thanks. I needed to address it anyways.
I'm here with you. I, myself too, love being picky about coding and following rules.
I felt like being pushed to decide things, not my decision to make. Espacially the `unset` commands. I would have just removed all of them, if you had just told me to do so. I was just not willing to make this decision on my own.

> I am happy with the refactoring you did regarding fetching the user and I do not think it overlaps that much with going through the account intead of `OCP\IUser`. Only the calls to `$user->setProfileProperty` will need to change on user_ldap side.

Thanks. Now I have done the rest, like you suggested and I removed the added code from IUser interface.

> If this is eating too much of your time I can take it from here if your prefer.

Thank you for the offer. I think I have it almost finished for now. I finished the code and finally did the rebase. For now, I'm testing and trying to find scenarios where it fails or crashes.

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +574,51 @@ private function verifyQuotaValue(string $quotaValue) {
...
+		if ($this->wasRefreshed('profile')) {
+			return;
+		}
+		// fetch/prepare user
+		$user = $this->userManager->get($this->uid);
+		if (is_null($user)) {
+			return;
+		}
+		// prepare AccountManager and Account
+		$accountManager = \OC::$server->get(IAccountManager::class);
```
march42, line 594: Is this a good way to get an accountManager instance, to work with?

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +574,51 @@ private function verifyQuotaValue(string $quotaValue) {
...
+			return;
+		}
+		// loop through the properties and handle them
+		foreach($profileValues as $property => $valueFromLDAP) {
+			// check and update profile properties
+			$value = (is_array($valueFromLDAP) ? $valueFromLDAP[0] : $valueFromLDAP); // take ONLY the first value, if multiple values specified
+			try {
+				$accountProperty = $account->getProperty($property);
+				$currentValue = $accountProperty->getValue();
+				$scope = ($profileScope ? $profileScope : ($accountProperty->getScope() ? $accountProperty->getScope() : AccountManager::DEFAULT_SCOPES[$property]));
```
march42, line 606: I got the idea, that `AccountManager::DEFAULT_SCOPES[$property]` doesn't get changed by setting `account_manager.default_property_scope` in `config.php` - do I need to check this config setting extra?

apps/user_ldap/lib/User/User.php
come-nc:
```suggestion
		$accountManager = Server::get(IAccountManager::class);
```
This with a `use OCP\Server` at the beginning of the file is the preferred way now.
The best is to use DI (Dependency Injection) in the constructor, but here for a user object it’s more complicated because it’s not a singleton. In this instance I think it’s fine to use a get here.

apps/user_ldap/lib/User/User.php
come-nc: DEFAULT_SCOPES is a const, it never changes.
You can see in AccountManager.php line 674 that it merges that with values from the config file to build the actual default scope.

Ideally you should not have to use AccountManager since it’s in `OC` and not `OCP`.
But it seems you found a missing feature from OCP here.
So either we’d need to add a getPropertyDefaultScope method to `OCP\IAccountManager`, or a magic value to give setProperty when wanting to use the default scope. (either null or a new IAccountManager::VISIBILITY_DEFAULT)
@Pytal What do you think, did I miss something?

apps/user_ldap/lib/User/User.php
```diff
@@ -512,6 +574,51 @@ private function verifyQuotaValue(string $quotaValue) {
...
+			}
+			catch (PropertyDoesNotExistException $e) { // thrown at getProperty
```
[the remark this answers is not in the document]
come-nc: It can be thrown by getProperty if the property do not exist. You should leave it there, if Account is ever modified to remove one of the property you use we will get a clean handling instead of a misterious crash. Better safe than  sorry

come-nc (approved):
Apart from the tiny default scope question, all good!

Just tested and confirm it works.

apps/user_ldap/lib/User/User.php
march42: Looking into the code again, I think about removing the scope handling from my code and only use `Account::getProfileScope`

- I added this, to make sure all profile data stays on my server setting `v2-local`
- using the `config.php` `account_manager.default_property_scope` is more flexible and the user stays in control
- I want to rely on the defaults - using config.php and existing mechanisms
- new user will have empty profile with default scope applied
- if user changes property scope it will be used (CAN be both, pro and con ... and IS users choice)

Pytal:
Are the FIXMEs still to be done in this PR @march42?

march42:
> Are the FIXMEs still to be done in this PR @march42?

@Pytal yes. I couldn't find a scenario, where those two checks against null in User.php (line 590 and 596) updateProfile will be true. I left them in there for "better safe than sorry" purpose.

Currently I'm doing some more testing and digging and found another question with profile handling. If I set phoneNumber to "0971..." in LDAP it gets formated to "+49971..." in user profile. And the next check in updateProfile (line 615) it get set again. I found this is not a big performance issue and think I need to add to documentation and clearify for the LDAP admin how the attributes should be prepared.
While checking with multi line attributes streetAddress and userBiography I noticed streetAddress has CRLF line ending and userBiography has a $ dollar sign instead of CR or LF. Here I need to double check again, where those difference comes from - OpenLDAP server or the configuration client. And I want to check Nextcloud, how multi line property should be formated here.
So, I was/am thinking about adding data formatting to processAttributes, at least for multi line values.

come-nc:
> Currently I'm doing some more testing and digging and found another question with profile handling. If I set phoneNumber to "0971..." in LDAP it gets formated to "+49971..." in user profile. And the next check in updateProfile (line 615) it get set again. I found this is not a big performance issue and think I need to add to documentation and clearify for the LDAP admin how the attributes should be prepared. While checking with multi line attributes streetAddress and userBiography I noticed streetAddress has CRLF line ending and userBiography has a $ dollar sign instead of CR or LF. Here I need to double check again, where those difference comes from - OpenLDAP server or the configuration client. And I want to check Nextcloud, how multi line property should be formated here. So, I was/am thinking about adding data formatting to processAttributes, at least for multi line values.

The $ thing is the postal address format in LDAP: https://www.rfc-editor.org/rfc/rfc4517#section-3.3.28
This is the complicated thing about allowing the admin to choose any attribute to map, there are many different syntaxes out there.

march42:
> The $ thing is the postal address format in LDAP: https://www.rfc-editor.org/rfc/rfc4517#section-3.3.28
Yes, thank you, you're absolutely right. I did not remember this.

> This is the complicated thing about allowing the admin to choose any attribute to map, there are many different syntaxes out there.
And in most cases I would just tell the admin which syntax I want/need.
Last year I thought about implementing a formating syntax instead of plain attribute names. So the admin could format like "street+<CR>+postalCode+<SPACE>+l" and I decided this would be too much and dropped it.
My second thought was using the attributes syntax OID and mapping based on that. And right now I think this would be my preffered way to go.

BTW: now I can explain the reformatting of my phone numbers - E.123 format in LDAP allows many different syntaxes, i used international with dashes for separation, while E.164 in Nextcloud wants international without delimiters.

BUT: every checking and formatting here can/will have a performance impact.

march42:
> Are the FIXMEs still to be done in this PR @march42?

@Pytal I finished and removed the FIXMEs.

Pytal (approved):

march42:
I changed my LDAP directory and have `nextcloud-user-profile.schema` a modified version of @blizzz `nextcloud.schema` (i think from the univention app) with attributes for all the fields added. I have no idea, where to put this or/and if I should ignore.

blizzz:
> I changed my LDAP directory and have `nextcloud-user-profile.schema` a modified version of @blizzz `nextcloud.schema` (i think from the univention app) with attributes for all the fields added. I have no idea, where to put this or/and if I should ignore.

That .schema file is an old format to define LDAP schemas, which may be imported to the LDAP server. This is nothing Nextcloud deals with.

szaimen:
Merge?

march42:
> Merge?

If you're asking me, then yes.

come-nc:
There are open questions in https://github.com/nextcloud/documentation/pull/9614

Issue #28923, opened 2021-09-21
Add event dispatcher to OC\DB\QueryBuilder::execute


summersab:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
For advanced app integration, it would be nice to have an event dispatcher run before and after `OC\DB\QueryBuilder::execute`. While this may seem a little odd, there are certainly situations where this functionality would be useful:
- Security apps could check SQL queries and parameters for dangerous patterns and prevent them from executing.
- Some apps don't fire events themselves, but they do access the DB. A `QueryBuilder` event listener would allow developers to execute code when a query matching certain conditions is executed.
- Deep integration with the platform and other apps would be possible.
- In my particular case, I'm wanting to encrypt/decrypt certain fields used by other apps. I have written an app to transparently encrypt DAV data, but this required some rather kludge code to avoid editing core libraries. An event listener would have been a simpler solution.

**Describe the solution you'd like**
It would be great to have an event dispatcher added to `OC\DB\QueryBuilder::execute` both before and after `$result = $this->queryBuilder->execute();`. The dispatcher should be able to check if propagation has been halted, and if so, it will return the result provided by the listener.

**Describe alternatives you've considered**
- I have tried to use the existing middleware and events to add certain functionalities, but these were neither efficient, reliable, nor ideal.
- In order to modify DB calls made by existing apps, I have "overloaded" their classes by unregistering their existing classmap, autoloading a modified classmap specifying my modified classes, and executed queries this way. While this works and doesn't modify core code . . . it is gross and not recommended.
- I have tried editing core libraries, but this is obviously not recommended.

joshtrichards:
Related PR is #36572

PR #36572, opened 2023-02-07
Add events to `QueryBuilder` class

### LINKED ISSUE #13103: Add support for database encyrption
The NextCloud database contains sensitive/personal data about users. Sensitive data could be leaked due a vulnerability in NextCloud. On most installations, the database is hosted on a third parity infrastructure, that could also leak data.

PostgreSQL, MariaDB and MySQL all support symmetric and asymmetric encryption of columns natively.
It would be great to have the ability to optionally encrypt sensitive/personal content of the database.
Similar to the file encryption module, this can be done with a master key and/or user specific keys.
This will make it much harder to steal useful data, because you'd also need all user keys to decrypt the content.

MariaDB does also support database encryption:
https://mariadb.com/resources/blog/table-and-tablespace-encryption-on-mariadb-10-1/
https://mariadb.com/kb/en/library/why-encrypt-mariadb-data/
But this would not protect users if data is leaked due a vulnerability in NextCloud (e.g. SQLI).

summersab:
* Would provide the functionality to resolve: #13103

## Summary
This provides events before and after the execution of `QueryBuilder` objects in order to modify `QueryBuilder` elements, the query results, or both. While it may seem like this is only useful for niche applications (or potentially dangerous), there are valid instances where intercepting DB queries and results are extremely useful (or even necessary).

Being able to intercept queries and/or results in order to restrict access, redact information, or even generate query-level logs would require the ability to access the `QueryBuilder` class before and/or after execution.

One application that I have implemented for my organization is adding column-level encryption on a per-user basis. Without being able to access the `QueryBuilder` class immediately before execution, this integration would be impossible. Once this PR is accepted and my integration is tested more thoroughly in production, I plan to make my code publicly available so administrators may encrypt information in the NC database such as contact and calendar data.

## Checklist

### DIFF
--- a/lib/private/DB/QueryBuilder/QueryBuilder.php
+++ b/lib/private/DB/QueryBuilder/QueryBuilder.php
@@ -48,11 +48,14 @@ use OC\DB\QueryBuilder\FunctionBuilder\SqliteFunctionBuilder;
 use OC\DB\ResultAdapter;
 use OC\SystemConfig;
 use OCP\DB\IResult;
+use OCP\DB\QueryBuilder\Events\BeforeQueryExecutedEvent;
+use OCP\DB\QueryBuilder\Events\QueryExecutedEvent;
 use OCP\DB\QueryBuilder\ICompositeExpression;
 use OCP\DB\QueryBuilder\ILiteral;
 use OCP\DB\QueryBuilder\IParameter;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\DB\QueryBuilder\IQueryFunction;
+use OCP\EventDispatcher\IEventDispatcher;
 use Psr\Log\LoggerInterface;
 
 class QueryBuilder implements IQueryBuilder {
@@ -76,6 +79,9 @@ class QueryBuilder implements IQueryBuilder {
 	/** @var string */
 	protected $lastInsertedTable;
 
+	/** @var IEventDispatcher */
+	private $dispatcher;
+
 	/**
 	 * Initializes a new QueryBuilder.
 	 *
@@ -88,6 +94,7 @@ class QueryBuilder implements IQueryBuilder {
 		$this->logger = $logger;
 		$this->queryBuilder = new \Doctrine\DBAL\Query\QueryBuilder($this->connection->getInner());
 		$this->helper = new QuoteHelper();
+		$this->dispatcher = \OC::$server->get(IEventDispatcher::class);
 	}
 
 	/**
@@ -277,7 +284,18 @@ class QueryBuilder implements IQueryBuilder {
 			]);
 		}
 
-		$result = $this->queryBuilder->execute();
+		$event = new BeforeQueryExecutedEvent($this);
+		$this->dispatcher->dispatchTyped($event);
+		$result = $event->getResult();
+
+		if ($result === null) {
+			$result = $this->queryBuilder->execute();
+		}
+
+		$event = new QueryExecutedEvent($this, $result);
+		$this->dispatcher->dispatchTyped($event);
+		$result = $event->getResult();
+		
 		if (is_int($result)) {
 			return $result;
 		}
--- a/lib/composer/composer/autoload_classmap.php
+++ b/lib/composer/composer/autoload_classmap.php
@@ -219,6 +219,8 @@ return array(
     'OCP\\DB\\IPreparedStatement' => $baseDir . '/lib/public/DB/IPreparedStatement.php',
     'OCP\\DB\\IResult' => $baseDir . '/lib/public/DB/IResult.php',
     'OCP\\DB\\ISchemaWrapper' => $baseDir . '/lib/public/DB/ISchemaWrapper.php',
+    'OCP\\DB\\QueryBuilder\\Events\\BeforeQueryExecutedEvent' => $baseDir . '/lib/public/DB/QueryBuilder/Events/BeforeQueryExecutedEvent.php',
+    'OCP\\DB\\QueryBuilder\\Events\\QueryExecutedEvent' => $baseDir . '/lib/public/DB/QueryBuilder/Events/QueryExecutedEvent.php',
     'OCP\\DB\\QueryBuilder\\ICompositeExpression' => $baseDir . '/lib/public/DB/QueryBuilder/ICompositeExpression.php',
     'OCP\\DB\\QueryBuilder\\IExpressionBuilder' => $baseDir . '/lib/public/DB/QueryBuilder/IExpressionBuilder.php',
     'OCP\\DB\\QueryBuilder\\IFunctionBuilder' => $baseDir . '/lib/public/DB/QueryBuilder/IFunctionBuilder.php',
--- a/lib/composer/composer/autoload_static.php
+++ b/lib/composer/composer/autoload_static.php
@@ -252,6 +252,8 @@ class ComposerStaticInit749170dad3f5e7f9ca158f5a9f04f6a2
         'OCP\\DB\\IPreparedStatement' => __DIR__ . '/../../..' . '/lib/public/DB/IPreparedStatement.php',
         'OCP\\DB\\IResult' => __DIR__ . '/../../..' . '/lib/public/DB/IResult.php',
         'OCP\\DB\\ISchemaWrapper' => __DIR__ . '/../../..' . '/lib/public/DB/ISchemaWrapper.php',
+        'OCP\\DB\\QueryBuilder\\Events\\BeforeQueryExecutedEvent' => __DIR__ . '/../../..' . '/lib/public/DB/QueryBuilder/Events/BeforeQueryExecutedEvent.php',
+        'OCP\\DB\\QueryBuilder\\Events\\QueryExecutedEvent' => __DIR__ . '/../../..' . '/lib/public/DB/QueryBuilder/Events/QueryExecutedEvent.php',
         'OCP\\DB\\QueryBuilder\\ICompositeExpression' => __DIR__ . '/../../..' . '/lib/public/DB/QueryBuilder/ICompositeExpression.php',
         'OCP\\DB\\QueryBuilder\\IExpressionBuilder' => __DIR__ . '/../../..' . '/lib/public/DB/QueryBuilder/IExpressionBuilder.php',
         'OCP\\DB\\QueryBuilder\\IFunctionBuilder' => __DIR__ . '/../../..' . '/lib/public/DB/QueryBuilder/IFunctionBuilder.php',
--- a/lib/public/DB/QueryBuilder/Events/BeforeQueryExecutedEvent.php
+++ b/lib/public/DB/QueryBuilder/Events/BeforeQueryExecutedEvent.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Andrew Summers
+ *
+ * @author Andrew Summers
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\DB\QueryBuilder\Events;
+
+use OCP\EventDispatcher\Event;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+use Doctrine\DBAL\Result;
+
+/**
+ * This event is used by apps to intercept, inspect, and potentially modify
+ * database queries prior to execution. This can be used for deep integrations,
+ * restricting user access to certain data, redacting information, etc.
+ *
+ * The result field can optionally be set by the event listener by executing the
+ * query in the event handler logic. This allows apps to inspect the results,
+ * use try/catch blocks around the query execution, and/or modify/re-execute the
+ * query if necessary. If the result field is not set, the QueryBuilder class
+ * will execute the query as expected by default.
+ *
+ * @see https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/projects.html
+ * @since 28.0.0
+ */
+class BeforeQueryExecutedEvent extends Event {
+	private IQueryBuilder $queryBuilder;
+	private $result = null;
+
+	/**
+	 * @param IQueryBuilder $queryBuilder
+	 * @since 28.0.0
+	 */
+	public function __construct(IQueryBuilder $queryBuilder) {
+		$this->queryBuilder = $queryBuilder;
+	}
+
+	/**
+	 * @return IQueryBuilder
+	 * @since 28.0.0
+	 */
+	public function getQueryBuilder(): IQueryBuilder {
+		return $this->queryBuilder;
+	}
+
+	/**
+	 * @param IQueryBuilder $queryBuilder
+	 * @since 28.0.0
+	 */
+	public function setQueryBuilder(IQueryBuilder $queryBuilder) {
+		$this->queryBuilder = $queryBuilder;
+	}
+
+	/**
+	 * @return Result|int|string
+	 * @since 28.0.0
+	 */
+	public function getResult() {
+		return $this->result;
+	}
+
+	/**
+	 * @param Result|int|string $result
+	 * @since 28.0.0
+	 */
+	public function setResult($result) {
+		$this->result = $result;
+	}
+}
--- a/lib/public/DB/QueryBuilder/Events/QueryExecutedEvent.php
+++ b/lib/public/DB/QueryBuilder/Events/QueryExecutedEvent.php
@@ -1,0 +1,87 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Andrew Summers
+ *
+ * @author Andrew Summers
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\DB\QueryBuilder\Events;
+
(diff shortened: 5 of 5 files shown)


summersab:
I didn't see how to implement tests for `Event` or `EventDispatcher` classes. There are tests for the deprecated dispatcher. There is also the ability to create a mock `IEventDispatcher`, but my class doesn't create a new `IEventDispatcher` - rather, it gets the class from `OC::$server`.

That said, if there are applicable tests which I need to implement, I would be happy to add them.

nickvergessen:
> or even generate query-level logs

This is already available by setting the following config:
https://github.com/nextcloud/server/blob/dde5c46a3eb7c6dcee47795590619ccd393577d2/config/config.sample.php#L2225-L2232

Regarding the events:From my POV this looks like an interesting idea, but it has the option to fail very hard performance-wise.

Regarding column encryption: Also any user that has access to the database mostlikely also has access to the source code, so the encryption is not helping a lot. Which use/attack case are you trying to prevent?

summersab:
> Regarding the events:From my POV this looks like an interesting idea, but it has the option to fail very hard performance-wise.

Yes, this is true - if used incorrectly, these events can be a huge performance failure. However, based on my early usage of this code, there really isn't a noticible performance hit when done well. (Also, let's be honest. Plenty of integrations have the _potential_ for bad performance unless you build them right. Some would be even worse than this). 

> Regarding column encryption: Also any user that has access to the database mostlikely also has access to the source code, so the encryption is not helping a lot. Which use/attack case are you trying to prevent?

The integration I wrote uses `AES_ENCRYPT`/`AES_DECRYPT` (MariaDB) or `pgcrypto` (PostgreSQL) in combination with the `IProvideUserSecretBackend` class to encrypt user data in the DB using a user-specific secret. Based on the tables and columns present in a given query, my code adds the required functions to the `QueryBuilder` object in order to encrypt and decrypt user data. It may seem questionable, but it works  quite well, and I have not noticed a performance hit.

That said, yes - this is a very unique implementation with very specific use-cases, and it could lead to negative effects. However, those use-cases DO exist. Also, there are other NC APIs that can impact performance even worse when used incorrectly.

Thank you for your feedback. Let me know if you have further questions, comments, or suggestions - happy to discuss!

nickvergessen:
> in combination with the IProvideUserSecretBackend class to encrypt user data in the DB using a user-specific secret

I get that, but the server might need to write data without the user password being available (e.g. in a background job, or when others trigger an update of "your" data. So the server is able to encrypt/decrypt and therefor also the sysadmin and any one that hacks themselves access to the server.
That's the reason why I wonder which of the attack vectors you are trying to prevent.

summersab:
Well, I decided to use `IProvideUserSecretBackend` in my implementation (which comes with a number of limitations like you mentioned like background jobs - I have taken this into account with my setup). It is also possible to use something like the NC server secret for encrypting database columns, and all features work as expected.

Database encryption is my particular use-case for this PR, but I am sure there are other reasons why these events might be usefulm

summersab:
@nickvergessen, I took some metrics to address the subject of performance. This is not the most scientific calculation, but it provides a rough estimate. I created a contact in NC and edited it multiple times. Then, I enabled my `QueryBuilder` events and encryption logic, created a new contact, and edited it the same way. This was done on a pretty small VM with 1vCPU and 8GB RAM. Here are the results:
```
# of Queries  Avg (Default)    Avg (Encrypted)  Difference         Query
6             0.0009931723277  0.00150179863    0.0005086263021    DELETE FROM `*PREFIX*cards_properties` WHERE (`cardid` = :dcValue1) AND (`addressbookid` = :dcValue2)
36            0.0005187855826  0.00106041299    0.0005416274071    INSERT INTO `*PREFIX*cards_properties` (`addressbookid`, `cardid`, `name`, `value`, `preferred`) VALUES(:dcValue1, :dcValue2, :name, :value, :preferred)
6             0.0003132025401  0.000682870547   0.0003696680069    SELECT `id` FROM `*PREFIX*cards` WHERE (`uri` = :dcValue1) AND (`addressbookid` = :dcValue2)
18            0.0004307428996  0.0009056540097  0.0004749111101    SELECT `id`, `uri`, `lastmodified`, `etag`, `calendarid`, `size`, `calendardata`, `componenttype`, `classification`, `deleted_at` FROM `*PREFIX*calendarobjects` WHERE (`calendarid` = :dcValue1) AND (`uri` = :dcValue2) AND (`calendartype` = :dcValue3)
30            0.0004283030828  0.002046084404   0.001617781321     SELECT `id`, `uri`, `lastmodified`, `etag`, `size`, `carddata`, `uid` FROM `*PREFIX*cards` WHERE (`addressbookid` = :dcValue1) AND (`uri` = :dcValue2) LIMIT 1
5             0.001100826263   0.002524375916   0.001423549652     UPDATE `*PREFIX*cards` SET `carddata` = :dcValue1, `lastmodified` = :dcValue2, `size` = :dcValue3, `etag` = :dcValue4, `uid` = :dcValue5 WHERE (`uri` = :dcValue6) AND (`addressbookid` = :dcValue7)
```

Again, this is just _my_ use-case for adding `Event` dispatchers to the `QueryBuilder` class. There are probably other uses. With that said, it is pretty clear that the performance hit is negligible.

summersab:
What else is required for this PR? I'm not sure how to make CI happy...

nickvergessen:
> The autoloaders are not up to date
Please run: bash build/autoloaderchecker.sh
And commit the result

But the biggest problem is technical. This is a very fragile thing to be added. It comes with great risk (performance and listing to the event while triggering a SQL query will cause an infinite loop) and therefore we are currently unsure if this is the way to go, especially since the target usecase is soo limited. It does not even prevent admins from doing "bad" things, as they have access to the keys as well. So the only thing it would help against is when your database storage is stolen but not the nextcloud one?

summersab:
I completely understand the concerns regarding the events being very fragile and risky. However, I would like to respectfully push back a little.

Many existing events in NC are fragile, risky, can lead to infinite loops, and allow admins to do bad things. There are events that allow developers to interact with the authentication process, user/group creation, and even password updates. Are there really that many use cases for `OCP\Security\Events\GenerateSecurePasswordEvent`? As far as I can see, it is only used by the `password_policy` app. However, it IS available for all developers to use if they wish. That would allow them to do some pretty bad things.

My proposed events really aren't much different. Yes, the use cases are limited, but so are many events. That doesn't mean that there aren't valid use cases.

Yes, directly accessing database queries is generally avoided because it is risky, but there are TONS of internal elements of the NC framework already exposed to developers. Many of them are just as risky.

Yes, this could lead to infinite loops if improperly used, but even the most benign APIs can lead to infinite loops if not used correctly. This has happened to every developer at some point.

I think this is the nature of open source and security. Systems should be as malleable as possible (even if it is risky) so developers don't have to "hack" system internals. In turn, developers should make their code available for public review and scrutiny. With that in mind, which option is more responsible?
1. I could fork NC, edit these classes, and implement my code. This would allow me to move forward, but I would have to edit core files, and it breaks the code signing in NC.
2. I could continue to politely petition for this PR to officially provide access to the `QueryBuilder` class even though there are risks and very few use cases (ahem - `GenerateSecurePasswordEvent`). This would make NC more malleable so I can write a proper integration for NC without hacking into core classes.

Either method will accomplish the same goal, but I think the second option is much more responsible. That is why I am pursuing it.

summersab:
Any thoughts on how to better structure this and still provide the same sort of functionality, @nickvergessen? It's a touchy integration, but I'd like to stop overriding classes in my NC instances. Plus, I could provide my app that enables per-user column encryption, and I think a lot of people would benefit from that.

lib/private/DB/QueryBuilder/Events/AfterQueryExecuted.php
```diff
@@ -0,0 +1,65 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\DB\QueryBuilder\Events;
```
nickvergessen, line 27: Should be moved to the public namespace `OCP` and located in `lib/public/...`

lib/private/DB/QueryBuilder/Events/BeforeQueryExecuted.php
```diff
@@ -0,0 +1,69 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\DB\QueryBuilder\Events;
```
nickvergessen, line 27: Should be moved to the public namespace `OCP` and located in `lib/public/...`

summersab: Done - thank you!

nickvergessen:
You modified 3rdparty? That should not be needed.

On that note, what is needed is `bash build/autoload....`

summersab:
Crap. I don't know how I managed to modify 3rdparty. That certainly wasn't intentional. I'm trying to figure out how to revert all of this - sorry for being such a noob.

summersab:
@nickvergessen, the postgres11-php8.0 CI test is failing because Postgres fails to respond. I don't think this is an issue with my code - the test just failed to properly run, I believe.

nickvergessen:
Yeah, seems unrelated

summersab:
Looks like I've finally gotten this past CI - hurrah!

Sorry for being such a noob on this PR. It's my first PR to a project this complex, so I had a lot to learn.

summersab:
@nickvergessen - my docblocks had issues, but I sorted them. Everything looks good except the `performance-8.0` check. It's failing on cloning `3rdparty` even though I don't touch those libs in my PR. The same thing happened in another PR I have open, and a maintainer said it didn't appear to be caused by me.

Let me know it there's anything else I need to do for this PR. Thank you for all of your help and patience!

nickvergessen:
Yeah should be unrelated here too. Mostlikely because it's a fork and not pushed to our org

summersab:
Cool. Anything else I need to do for review and approval except patiently wait on the devs (who are swamped with finalizing v26)?

nickvergessen:
> Anything else I need to do for review and approval

I guess it's fine

lib/public/DB/QueryBuilder/Events/AfterQueryExecuted.php
```diff
@@ -0,0 +1,87 @@
...
+	 */
+	public function getQueryBuilder(): IQueryBuilder {
+		return $this->queryBuilder;
+	}
+
+	/**
+	 * @param IQueryBuilder $queryBuilder
+	 * @since 26.0.0
+	 */
+	public function setQueryBuilder(IQueryBuilder $queryBuilder) {
```
ChristophWurst, line 68: I'm not a fan of mutable events and therefore would prefer to not have any setters on this object

dispatching an event should allow other code to hook into another application event (something that happened), not alter it in my opinion

lib/public/DB/QueryBuilder/Events/BeforeQueryExecuted.php
```diff
@@ -0,0 +1,90 @@
...
+ * restricting user access to certain data, redacting information, etc.
+ *
+ * The result field can optionally be set by the event listener by executing the
+ * query in the event handler logic. This allows apps to inspect the results,
+ * use try/catch blocks around the query execution, and/or modify/re-execute the
+ * query if necessary. If the result field is not set, the QueryBuilder class
+ * will execute the query as expected by default.
+ *
+ * @see https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/projects.html
+ * @since 26.0.0
```
ChristophWurst, line 45: 27

ChristophWurst (commented):
My (opinionated) design review

https://event-driven.io/en/whats_the_difference_between_event_and_command/

ChristophWurst:
> [ChristophWurst](/ChristophWurst) added the [ pending documentation ](/nextcloud/server/labels/pending%20documentation)

New OCP api requires documentation. E.g. https://docs.nextcloud.com/server/18/developer_manual/app/events.html#available-events

joshtrichards:
Linked open Issue this PR is associated with seems to be #28923

lib/public/DB/QueryBuilder/Events/AfterQueryExecuted.php
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * This event is used by apps to intercept, inspect, and potentially modify
+ * the results of database queries after execution. This can be used for deep
+ * integrations, restricting user access to certain data, redacting information,
+ * etc.
+ *
+ * @see https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/projects.html
+ * @since 26.0.0
+ */
+class AfterQueryExecuted extends Event {
```
nickvergessen, line 42: According to https://docs.nextcloud.com/server/latest/developer_manual/basics/events.html#naming-scheme this should be named:
```suggestion
class QueryExecutedEvent extends Event {
```

lib/public/DB/QueryBuilder/Events/AfterQueryExecuted.php
nickvergessen: While I agree, I think the idea here was to exactly do that. The planned app rewrites the queries and decrypts the results where needed

lib/public/DB/QueryBuilder/Events/BeforeQueryExecuted.php
```diff
@@ -0,0 +1,90 @@
...
+ * The result field can optionally be set by the event listener by executing the
+ * query in the event handler logic. This allows apps to inspect the results,
+ * use try/catch blocks around the query execution, and/or modify/re-execute the
+ * query if necessary. If the result field is not set, the QueryBuilder class
+ * will execute the query as expected by default.
+ *
+ * @see https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/projects.html
+ * @since 27.0.0
+ */
+class BeforeQueryExecuted extends Event {
```
nickvergessen, line 47: According to https://docs.nextcloud.com/server/latest/developer_manual/basics/events.html#naming-scheme this should be named:
```suggestion
class BeforeQueryExecutedEvent extends Event {
```

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -277,7 +284,18 @@ public function execute() {
...
-		$result = $this->queryBuilder->execute();
+		$event = new BeforeQueryExecuted($this);
+		$this->dispatcher->dispatchTyped($event);
+		$result = $event->getResult();
+
+		if ($result === null) {
+			$result = $this->queryBuilder->execute();
+		}
+
+		$event = new AfterQueryExecuted($this, $result);
```
nickvergessen, line 295:
```suggestion
		$event = new QueryExecutedEvent($this, $result);
```

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -277,7 +284,18 @@ public function execute() {
 			]);
 		}
 
-		$result = $this->queryBuilder->execute();
+		$event = new BeforeQueryExecuted($this);
```
nickvergessen, line 287:
```suggestion
		$event = new BeforeQueryExecutedEvent($this);
```

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -48,11 +48,14 @@
 use OC\DB\ResultAdapter;
 use OC\SystemConfig;
 use OCP\DB\IResult;
+use OCP\DB\QueryBuilder\Events\BeforeQueryExecuted;
+use OCP\DB\QueryBuilder\Events\AfterQueryExecuted;
```
nickvergessen, line 52:
```suggestion
use OCP\DB\QueryBuilder\Events\BeforeQueryExecutedEvent;
use OCP\DB\QueryBuilder\Events\QueryExecutedEvent;
```

3rdparty
nickvergessen, line 1: I guess this is a bug from rebasing.

nickvergessen (changes requested):

3rdparty
summersab: Yeah, I'm not sure what happened, here. Anything I need to do to resolve?

lib/public/DB/QueryBuilder/Events/BeforeQueryExecuted.php
summersab: > While I agree, I think the idea here was to exactly do that. The planned app rewrites the queries and decrypts the results where needed

This is correct. However, I CAN get by with just executing the query in my app and setting the result - the QueryBuilder::execute method checks for the existence of a result and uses that if necessary. I added the setter just as a convenience. I believe I can take it out and still have the same functionality.

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -88,6 +94,7 @@ public function __construct(ConnectionAdapter $connection, SystemConfig $systemC
 		$this->logger = $logger;
 		$this->queryBuilder = new \Doctrine\DBAL\Query\QueryBuilder($this->connection->getInner());
 		$this->helper = new QuoteHelper();
+		$this->dispatcher = \OC::$server->get(IEventDispatcher::class);
```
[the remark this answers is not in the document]
summersab: I think I can make that happen. However, I'm going to push back a little. There are a number of places throughout Nextcloud where this exact method is used (many of which include you as an author, actually):
https://github.com/nextcloud/server/blob/8b3b5ff8d86613c2f9c85ff5c9c5441a451961b8/core/Application.php#L85

https://github.com/nextcloud/server/blob/8b3b5ff8d86613c2f9c85ff5c9c5441a451961b8/lib/private/Files/Cache/Cache.php#L127

https://github.com/nextcloud/server/blob/8b3b5ff8d86613c2f9c85ff5c9c5441a451961b8/apps/dav/lib/Server.php#L107

https://github.com/nextcloud/server/blob/8b3b5ff8d86613c2f9c85ff5c9c5441a451961b8/apps/dav/lib/RootCollection.php#L67

If I were to inject it as a constructor argument, I would need to edit anywhere the QueryBuilder is instantiated, correct (looks like `OC\DB\Connection` is the only place)?

lib/public/DB/QueryBuilder/Events/AfterQueryExecuted.php
summersab: > While I agree, I think the idea here was to exactly do that. The planned app rewrites the queries and decrypts the results where needed

As noted elsewhere in the discussion:

This is correct. However, I CAN get by with just executing the query in my app and setting the result - the QueryBuilder::execute method checks for the existence of a result and uses that if necessary. I added the setter just as a convenience, if I recall. I believe I can take it out and still have the same functionality.

summersab:
> > [ChristophWurst](/ChristophWurst) added the [ pending documentation ](/nextcloud/server/labels/pending%20documentation)
> 
> 
> New OCP api requires documentation. E.g. https://docs.nextcloud.com/server/18/developer_manual/app/events.html#available-events

I opened the following PR as a draft so it can be merged pending the approval of this PR:
https://github.com/nextcloud/documentation/pull/10791

summersab:
Okay, why does updating the autoloaders wind up modifying some two dozen files with my PR? That's not right, but if I don't, CI isn't happy. I'm not sure what's going on, here.

tcitworld:
Keep the changes in `lib/`, drop the ones in `apps/`.

summersab:
> Keep the changes in `lib/`, drop the ones in `apps/`.

I didn't make any changes to apps. I just ran the composer update script, and it decided to modify all of these. Why?

PR #36584, opened 2023-02-07
feat(translations): Add translation provider API

juliusknorr:
## TODO

- [x] Wrap registration into app bootstrap for lazyness
- [x] Clarify if defined API makes sense
- [x] Improve error handling

## Checklist


lib/public/Translation/ITranslationProvider.php
```diff
@@ -0,0 +1,52 @@
...
+
+	/**
+	 * @since 26.0.0
+	 */
+	public function getName(): string;
+
+	/**
+	 * @since 26.0.0
+	 */
+	public function getAvailableLanguages(): array;
```
marcelklehr, line 41: Do we assume that the provider can translate to and from all languages (ie. the cartesian product)?

lib/private/Translation/TranslationManager.php
```diff
@@ -0,0 +1,61 @@
...
+		$languages = [];
+		foreach ($this->providers as $provider) {
+			$languages = array_merge($languages, $provider->getAvailableLanguages());
+		}
+		return $languages;
+	}
+
+	public function translate(string $text, string $fromLanguage, string $toLanguage): string {
+		foreach ($this->providers as $provider) {
+			return $provider->translate($fromLanguage, $toLanguage, $text) ?? 'no translation';
```
marcelklehr, line 53: This should probably be ?:

lib/public/Translation/ITranslationManager.php
```diff
@@ -0,0 +1,48 @@
...
+
+	/**
+	 * @since 26.0.0
+	 */
+	public function hasProviders(): bool;
+
+	/**
+	 * @since 26.0.0
+	 */
+	public function getLanguages(): array;
```
marcelklehr, line 42: Can we document the format of this array in the docblock here?

juliushaertl: Added a LanguageTuple class to represent this

lib/private/Translation/TranslationManager.php
```diff
@@ -0,0 +1,119 @@
...
+		}
+		return $languages;
+	}
+
+	public function translate(string $text, ?string $fromLanguage, string $toLanguage): string {
+		if ($fromLanguage === null) {
+			foreach ($this->getProviders() as $provider) {
+				if ($provider instanceof IDetectLanguageProvider) {
+					$fromLanguage = $provider->detectLanguage($text);
+					break;
```
marcelklehr, line 65: should keep looping here if the result is null?

lib/public/Translation/ITranslationManager.php
marcelklehr: you didn't commit it, i believe

lib/private/Translation/TranslationManager.php
juliushaertl: Yep, simplified that to only loop once and keep detection and translation for a provider combined.

lib/private/Translation/TranslationManager.php
juliushaertl: I've moved to throwing exceptions if there are errors with the translation from the provider side.

juliushaertl:
> Most providers accept translating full documents, maybe add an extra interface for those?

Might be a nice idea for a follow up extension, for now I'd like to keep this PR as simple as possible.

lib/private/Translation/TranslationManager.php
```diff
@@ -0,0 +1,120 @@
...
+		private IServerContainer $serverContainer,
+		private Coordinator $coordinator,
+		private LoggerInterface $logger,
+	) {
+	}
+
+	public function getLanguages(): array {
+		$languages = [];
+		foreach ($this->getProviders() as $provider) {
+			$languages = array_merge($languages, $provider->getAvailableLanguages());
```
marcelklehr, line 56: I believe this doesn't deduplicate the LangaugeTuple objects correctly

juliushaertl: I'd say we should consider them unique for each provider. I was trying to map deepl language codes to Nextcloud ones and failed because they have a larger variety of languages and their flavours (formal/informal). Felt more straight forward to just keep them separate then and by that also allow to provide translations for languages that are not available through the Nextcloud IL10N mapping.

marcelklehr: How can we avoid something like the following in the UI later, though?

* American English
* English
* English (American)
* French
* Francais
* Deutsch (Du)
* Deutsch (Sie)
* Deutsch, formell
* Deutsch, informell
* German
* German (formal)
* German (informal)

juliushaertl: From a user perspective I'd say it also doesn't really make sense to have multiple providers. We can of course introduce fixed languages key, label pairs in the server, but that would limit to what we define there. However I don't really have a strong opinion on that.

juliushaertl: Maybe we indeed just come up with a list of common languages and just let apps map their own offerings to that.

marcelklehr: > From a user perspective I'd say it also doesn't really make sense to have multiple providers.

True, then lets ignore this for now.

lib/private/Translation/TranslationManager.php
```diff
@@ -0,0 +1,120 @@
...
+			}
+
+			if ($fromLanguage === null) {
+				throw new InvalidArgumentException('Could not detect language');
+			}
+
+			try {
+				return $provider->translate($fromLanguage, $toLanguage, $text);
+			} catch (RuntimeException $e) {
+				$this->logger->warning("Failed translate from {$fromLanguage} to {$toLanguage}", ['exception' => $e]);
```
julien-nc, line 78: Nitpicking
```suggestion
				$this->logger->warning("Failed to translate from {$fromLanguage} to {$toLanguage}", ['exception' => $e]);
```

lib/private/Translation/TranslationManager.php
```diff
@@ -0,0 +1,120 @@
...
+
+		if ($this->providers !== null) {
+			return $this->providers;
+		}
+
+		$this->providers = [];
+		foreach ($context->getTranslationProviders() as $provider) {
+			$class = $provider->getService();
+			try {
+				$this->providers[$class] = $this->serverContainer->get($class);
```
julien-nc, line 96: I find this a bit confusing. `$provider` is actually an instance of `OC\Translation\ServiceRegistration` and not of `OCP\Translation\ITranslationProvider`. Not a big deal but there might be a more explicit variable name to avoid the confusion.

julien-nc (approved):
Very nice!
Apart from my small comments: :+1:

marcelklehr (approved):

lib/private/Translation/TranslationManager.php
juliushaertl: Adapted to providerRegistration

lib/private/Translation/TranslationManager.php
juliushaertl: Fixed

nickvergessen:
Missing API docs are in https://github.com/nextcloud/documentation/pull/10047

PR #36591, opened 2023-02-07
Refactor app loading

come-nc:
## TODO

- [x] Move loadApp to AppManager
- [x] Move away from server::get as much as possible
- [x] Move needed methods to AppManager to avoid calling OC_App

## Checklist

### DIFF
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -70,46 +70,32 @@ class AppManager implements IAppManager {
 		'prevent_group_restriction',
 	];
 
-	/** @var IUserSession */
-	private $userSession;
-
-	/** @var IConfig */
-	private $config;
-
-	/** @var AppConfig */
-	private $appConfig;
-
-	/** @var IGroupManager */
-	private $groupManager;
-
-	/** @var ICacheFactory */
-	private $memCacheFactory;
-
-	/** @var EventDispatcherInterface */
-	private $legacyDispatcher;
-
+	private IUserSession $userSession;
+	private IConfig $config;
+	private AppConfig $appConfig;
+	private IGroupManager $groupManager;
+	private ICacheFactory $memCacheFactory;
+	private EventDispatcherInterface $legacyDispatcher;
 	private IEventDispatcher $dispatcher;
-
-	/** @var LoggerInterface */
-	private $logger;
+	private LoggerInterface $logger;
 
 	/** @var string[] $appId => $enabled */
-	private $installedAppsCache;
+	private array $installedAppsCache = [];
 
-	/** @var string[] */
-	private $shippedApps;
+	/** @var string[]|null */
+	private ?array $shippedApps = null;
 
 	private array $alwaysEnabled = [];
 	private array $defaultEnabled = [];
 
 	/** @var array */
-	private $appInfos = [];
+	private array $appInfos = [];
 
 	/** @var array */
-	private $appVersions = [];
+	private array $appVersions = [];
 
 	/** @var array */
-	private $autoDisabledApps = [];
+	private array $autoDisabledApps = [];
 
 	/** @var array<string, true> */
 	private array $loadedApps = [];
@@ -135,7 +121,7 @@ class AppManager implements IAppManager {
 	/**
 	 * @return string[] $appId => $enabled
 	 */
-	private function getInstalledAppsValues() {
+	private function getInstalledAppsValues(): array {
 		if (!$this->installedAppsCache) {
 			$values = $this->appConfig->getValues(false, 'enabled');
 
@@ -234,12 +220,7 @@ class AppManager implements IAppManager {
 		}
 	}
 
-	/**
-	 * @param string $enabled
-	 * @param IUser $user
-	 * @return bool
-	 */
-	private function checkAppForUser($enabled, $user) {
+	private function checkAppForUser(string $enabled, ?IUser $user): bool {
 		if ($enabled === 'yes') {
 			return true;
 		} elseif ($user === null) {
@@ -267,16 +248,9 @@ class AppManager implements IAppManager {
 		}
 	}
 
-	/**
-	 * @param string $enabled
-	 * @param IGroup $group
-	 * @return bool
-	 */
 	private function checkAppForGroups(string $enabled, IGroup $group): bool {
 		if ($enabled === 'yes') {
 			return true;
-		} elseif ($group === null) {
-			return false;
 		} else {
 			if (empty($enabled)) {
 				return false;
@@ -706,7 +680,7 @@ class AppManager implements IAppManager {
 		return in_array($appId, $this->shippedApps, true);
 	}
 
-	private function isAlwaysEnabled($appId) {
+	private function isAlwaysEnabled(string $appId): bool {
 		$alwaysEnabled = $this->getAlwaysEnabledApps();
 		return in_array($appId, $alwaysEnabled, true);
 	}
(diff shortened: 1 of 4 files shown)


come-nc:
This is not bringing the expected performance boost, but it does cleanup things and move away from legacy OC_App, so I think it’s still worth to be merged.

ChristophWurst:
> This is not bringing the expected performance boost, but it does cleanup things and move away from legacy OC_App, so I think it’s still worth to be merged.

Clean code > fast code :wink:

lib/private/App/AppManager.php
```diff
@@ -174,6 +172,92 @@ public function getEnabledAppsForGroup(IGroup $group): array {
...
+	 *
+	 * This function walks through the Nextcloud directory and loads all apps
+	 * it can find. A directory contains an app if the file /appinfo/info.xml
+	 * exists.
+	 *
+	 * if $types is set to non-empty array, only apps of those types will be loaded
+	 */
+	public function loadApps(array $types = []): bool {
+		if ($this->config->getSystemValueBool('maintenance', false)) {
+			// TODO check if this if makes sense
```
ChristophWurst, line 189: todo in this PR or later?

lib/private/App/AppManager.php
```diff
@@ -174,6 +172,92 @@ public function getEnabledAppsForGroup(IGroup $group): array {
...
+	 * check if an app is of a specific type
+	 *
+	 * @param string $app
+	 * @param array $types
+	 * @return bool
+	 */
+	public function isType(string $app, array $types): bool {
+		$appTypes = $this->getAppTypes($app);
+		foreach ($types as $type) {
+			if (in_array($type, $appTypes)) {
```
ChristophWurst, line 235: pass `true` for strict comparison

lib/private/App/AppManager.php
```diff
@@ -174,6 +172,92 @@ public function getEnabledAppsForGroup(IGroup $group): array {
...
+			}
+		}
+		return false;
+	}
+
+	/**
+	 * get the types of an app
+	 *
+	 * @param string $app
+	 * @return array
```
ChristophWurst, line 246:
```suggestion
	 * @return string[]
```

lib/private/App/AppManager.php
```diff
@@ -174,6 +172,92 @@ public function getEnabledAppsForGroup(IGroup $group): array {
...
+	 */
+	private function getAppTypes(string $app): array {
+		//load the cache
+		if (count($this->appTypes) == 0) {
```
ChristophWurst, line 250:
```suggestion
		if (count($this->appTypes) === 0) {
```

or

```suggestion
		if (empty($this->appTypes)) {
```

lib/private/legacy/OC_App.php
```diff
@@ -295,51 +167,16 @@ public static function registerAutoloading(string $app, string $path, bool $forc
...
-		require_once $app . '/appinfo/app.php';
-	}
-
 	/**
 	 * check if an app is of a specific type
 	 *
 	 * @param string $app
 	 * @param array $types
 	 * @return bool
+	 * @deprecated 26.0.0 call \OC::$server->get(IAppManager::class)->isType($app, $types)
```
ChristophWurst, line 176:
```suggestion
	 * @deprecated 26.0.0 use IAppManager::isType
```

to not encourage service locators

ChristophWurst (approved):
Looks good

ChristophWurst:
Nit: clean up the git history. Either in the three steps described in the PR or one atomic change. Future bisectors will thank you :)

lib/private/App/AppManager.php
come-nc: The sooner the better.
The problem here is that I had to keep checks in OC_App before calling the AppManager loadApps, because when not installed we cannot built the AppManager. I do not know if this is the same for maintenance flag or if we can check it here instead.

come-nc:
Squashed and rebased on master

juliusknorr (approved):

juliushaertl:
Conflicts ;)

come-nc:
> Conflicts ;)

:disappointed: These are not easy to fix, the loadApp method modifications from @icewind1991 about event logging need to be transposed in the appmanager instead

lib/public/App/IAppManager.php
```diff
@@ -182,6 +196,27 @@ public function clearAppsCache();
...
+	 * exists.
+	 *
+	 * if $types is set to non-empty array, only apps of those types will be loaded
+	 * @since 26.0.0
+	 */
+	public function loadApps(array $types = []): bool;
+
+	/**
+	 * Check if an app is of a specific type
+	 * @since 26.0.0
```
nickvergessen, line 216:
```suggestion
	 * @since 27.0.0
```

lib/public/App/IAppManager.php
```diff
@@ -182,6 +196,27 @@ public function clearAppsCache();
...
+	 *
+	 * @param string[] $types
+	 * @return bool
+	 *
+	 * This function walks through the Nextcloud directory and loads all apps
+	 * it can find. A directory contains an app if the file /appinfo/info.xml
+	 * exists.
+	 *
+	 * if $types is set to non-empty array, only apps of those types will be loaded
+	 * @since 26.0.0
```
nickvergessen, line 210:
```suggestion
	 * @since 27.0.0
```

lib/public/App/IAppManager.php
```diff
@@ -93,6 +93,20 @@ public function isInstalled($appId);
...
+	 * Load an app, if not already loaded
+	 * @param string $app app id
+	 * @since 26.0.0
+	 */
+	public function loadApp(string $app): void;
+
+	/**
+	 * Check if an app is loaded
+	 * @param string $app app id
+	 * @since 26.0.0
```
nickvergessen, line 106:
```suggestion
	 * @since 27.0.0
```

lib/public/App/IAppManager.php
```diff
@@ -93,6 +93,20 @@ public function isInstalled($appId);
 	 */
 	public function isDefaultEnabled(string $appId):bool;
 
+	/**
+	 * Load an app, if not already loaded
+	 * @param string $app app id
+	 * @since 26.0.0
```
nickvergessen, line 99:
```suggestion
	 * @since 27.0.0
```

lib/private/legacy/OC_App.php
```diff
@@ -306,51 +165,16 @@ public static function registerAutoloading(string $app, string $path, bool $forc
...
-		require_once $app . '/appinfo/app.php';
-	}
-
 	/**
 	 * check if an app is of a specific type
 	 *
 	 * @param string $app
 	 * @param array $types
 	 * @return bool
+	 * @deprecated 26.0.0 use IAppManager::isType
```
nickvergessen, line 174:
```suggestion
	 * @deprecated 27.0.0 use IAppManager::isType
```

lib/private/legacy/OC_App.php
```diff
@@ -119,161 +117,22 @@ public static function isAppLoaded(string $app): bool {
...
-		return true;
+		return \OC::$server->get(IAppManager::class)->loadApps($types);
 	}
 
 	/**
 	 * load a single app
 	 *
 	 * @param string $app
 	 * @throws Exception
+	 * @deprecated 26.0.0 use IAppManager::loadApp
```
nickvergessen, line 132: 27 (and for the others too)

nickvergessen (changes requested):
Please document in https://github.com/nextcloud/server/issues/37039

PR #36592, opened 2023-02-07
fix default values and type hints for GroupManager::search

icewind1991:
ints really are ints

Signed-off-by: Robin Appelman <robin@icewind.nl>

### DIFF
--- a/lib/private/Group/Database.php
+++ b/lib/private/Group/Database.php
@@ -266,7 +266,7 @@ class Database extends ABackend implements
 	 *
 	 * Returns a list with all groups
 	 */
-	public function getGroups($search = '', $limit = null, $offset = null) {
+	public function getGroups(string $search = '', int $limit = -1, int $offset = 0) {
 		$this->fixDI();
 
 		$query = $this->dbConn->getQueryBuilder();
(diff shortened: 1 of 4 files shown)


lib/private/Group/Manager.php
```diff
@@ -252,7 +252,7 @@ public function search($search, $limit = null, $offset = null) {
 					$this->logger->debug('Group "' . $groupId . '" was returned by search but not found through direct access', ['app' => 'core']);
 				}
 			}
-			if (!is_null($limit) and $limit <= 0) {
+			if ($limit === 0) {
```
come-nc, line 255: Hum, the behavior when passing -1 explicitely changes then, no?

icewind1991: tbh, I don't know if the previous behavior with `-1` was intentional, and I can't find any code calling it with `-1` with a quick search

icewind1991:
changed the `IGroupManager` to accept `null`s to prevent messing with code that relied on passing `null` as a default value, the `null`s are changed to the proper `-`/`0` before passing them to the backends so the backends still get the stronger typing

lib/private/Group/Database.php
```diff
@@ -279,8 +279,12 @@ public function getGroups($search = '', $limit = null, $offset = null) {
 			)));
 		}
 
-		$query->setMaxResults($limit)
-			->setFirstResult($offset);
+		if (!$limit > 0) {
```
come-nc, line 282:
```suggestion
		if ($limit <= 0) {
```

come-nc (approved):
See nitpick, fine otherwise.

skjnldsv (approved):

PR #36609, opened 2023-02-08
optimize LazyUserFolder::getMountPoint

icewind1991:
no need to do a full setup when we can just get the mountpoint we want.


come-nc (changes requested):
Psalm is angry

icewind1991:
made psalm happy

come-nc (approved):

juliusknorr (approved):

blizzz:
@icewind1991 should this be backported to 26?

PR #36610, opened 2023-02-08
only fetch the data for mounts inside a folder when needed

icewind1991:
for most operations we don't actually care about any mounts inside a folder, only for metadata that needs to propagate across storage boundaries (size, etag, mtime) do we need all the submount info.

By only loading this data when needed we can save a bunch of storage setup in a number of cases


come-nc (approved):

juliusknorr (approved):
Nice one :)

PR #36612, opened 2023-02-08
fix(collaboration-resources): Ensure nullable columns are nullable

ChristophWurst:
* Resolves: # 

## Summary

``\OC\Collaboration\Resources\Manager::cacheAccessForCollection`` inserts null values

## Checklist

### DIFF
--- a/core/Migrations/Version23000Date20230208144411.php
+++ b/core/Migrations/Version23000Date20230208144411.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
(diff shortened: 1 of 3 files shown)


ChristophWurst:
Does not work because mysql converts the columns to notnull when they are used in a primary key

ref https://stackoverflow.com/questions/11001333/null-value-in-multi-column-primary-key

`php occ db:add-missing-primary-keys` implicitly converts the columns if they are nullable after the initial migration

PR #36620, opened 2023-02-08
Better Calendar Widget Data

miaulalala:
Signed-off-by: Anna Larch <anna@nextcloud.com>

* Resolves part of https://github.com/nextcloud/calendar/issues/4813 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/CalDavBackend.php
+++ b/apps/dav/lib/CalDAV/CalDavBackend.php
@@ -1891,13 +1891,20 @@ class CalDavBackend extends AbstractBackend implements SyncSupport, Subscription
 
 		$outerQuery->andWhere($outerQuery->expr()->in('c.id', $outerQuery->createFunction($innerQuery->getSQL())));
 
+		$allowedSorts = ['id','calendarid', 'lastmodified', 'firstoccurence', 'lastoccurence'];
 		if(!empty($options['sort_asc'])) {
 			foreach($options['sort_asc'] as $sort) {
+				if(!in_array($sort, $allowedSorts)) {
+					continue;
+				}
 				$outerQuery->orderBy($sort, 'ASC');
 			}
 		}
 		if(!empty($options['sort_desc'])) {
 			foreach($options['sort_desc'] as $sort) {
+				if(!in_array($sort, $allowedSorts)) {
+					continue;
+				}
 				$outerQuery->orderBy($sort, 'DESC');
 			}
 		}
--- a/apps/dav/lib/CalDAV/CalDavBackend.php
+++ b/apps/dav/lib/CalDAV/CalDavBackend.php
@@ -356,6 +356,7 @@ class CalDavBackend extends AbstractBackend implements SyncSupport, Subscription
 				'{' . Plugin::NS_CALDAV . '}supported-calendar-component-set' => new SupportedCalendarComponentSet($components),
 				'{' . Plugin::NS_CALDAV . '}schedule-calendar-transp' => new ScheduleCalendarTransp($row['transparent']?'transparent':'opaque'),
 				'{' . \OCA\DAV\DAV\Sharing\Plugin::NS_OWNCLOUD . '}owner-principal' => $this->convertPrincipal($principalUri, !$this->legacyEndpoint),
+
 			];
 
 			$calendar = $this->rowToCalendar($row, $calendar);
@@ -1873,6 +1874,16 @@ class CalDavBackend extends AbstractBackend implements SyncSupport, Subscription
 
 		$outerQuery->andWhere($outerQuery->expr()->in('c.id', $outerQuery->createFunction($innerQuery->getSQL())));
 
+		if(!empty($options['sort_asc'])) {
+			foreach($options['sort_asc'] as $sort) {
+				$outerQuery->orderBy($sort, 'ASC');
+			}
+		}
+		if(!empty($options['sort_desc'])) {
+			foreach($options['sort_desc'] as $sort) {
+				$outerQuery->orderBy($sort, 'DESC');
+			}
+		}
 		if ($offset) {
 			$outerQuery->setFirstResult($offset);
 		}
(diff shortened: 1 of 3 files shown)


lib/public/Calendar/IGetTimezone.php
```diff
@@ -0,0 +1,41 @@
...
+
+interface IGetTimezone extends ICalendar {
+	/**
+	 * Get the calendar timezone as a string
+	 * i.e. Europe/Vienna
+	 * as set in the VTIMEZONE->TZID for a calendar
+	 *
+	 * @since 26.0.0
+	 */
+	public function getCalendarTimezoneString(): ?string;
```
tcitworld, line 39: **suggestion:** I would name it `getCalendarTimezoneID`

apps/dav/lib/CalDAV/CalendarImpl.php
```diff
@@ -235,4 +237,14 @@ public function handleIMipMessage(string $name, string $calendarData): void {
 	public function getInvitationResponseServer(): InvitationResponseServer {
 		return new InvitationResponseServer(false);
 	}
+
+	public function getCalendarTimezoneString(): ?string {
+		$vTimezoneString = $this->calendarInfo['{urn:ietf:params:xml:ns:caldav}calendar-timezone'];
+		if(empty($vTimezoneString)) {
+			return null;
+		}
+		$cal = Reader::read($vTimezoneString);
```
tcitworld, line 246: **nitpick:** I would catch exceptions here and return `null` properly. I'm not sure server validates this property, so a client could put garbage inside it.

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -1873,6 +1874,16 @@ public function search(array $calendarInfo, $pattern, array $searchProperties,
 
 		$outerQuery->andWhere($outerQuery->expr()->in('c.id', $outerQuery->createFunction($innerQuery->getSQL())));
 
+		if(!empty($options['sort_asc'])) {
+			foreach($options['sort_asc'] as $sort) {
+				$outerQuery->orderBy($sort, 'ASC');
```
tcitworld, line 1879: **thought (non-blocking):** Should we validate or restrict the value of `$sort` before using it (against valid column names)? In order not to explode at runtime.

(same applies just below)

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -356,6 +356,7 @@ public function getCalendarsForUser($principalUri) {
 				'{' . Plugin::NS_CALDAV . '}supported-calendar-component-set' => new SupportedCalendarComponentSet($components),
 				'{' . Plugin::NS_CALDAV . '}schedule-calendar-transp' => new ScheduleCalendarTransp($row['transparent']?'transparent':'opaque'),
 				'{' . \OCA\DAV\DAV\Sharing\Plugin::NS_OWNCLOUD . '}owner-principal' => $this->convertPrincipal($principalUri, !$this->legacyEndpoint),
+
```
tcitworld, line 359: **nitpick (non-blocking):** Extra line that came out of nowhere

lib/public/Calendar/IGetTimezone.php
miaulalala: good point

apps/dav/lib/CalDAV/CalendarImpl.php
```diff
@@ -235,4 +237,14 @@ public function handleIMipMessage(string $name, string $calendarData): void {
 	public function getInvitationResponseServer(): InvitationResponseServer {
 		return new InvitationResponseServer(false);
 	}
+
+	public function getCalendarTimezoneString(): ?string {
```
tcitworld, line 241: **suggestion:** There is a very similar block of code in https://github.com/nextcloud/server/pull/36192, it would be nice to reuse it somehow. :)

lib/public/Calendar/IGetTimezone.php
```diff
@@ -0,0 +1,41 @@
...
+ *  *
+ *
+ */
+namespace OCP\Calendar;
+
+use OCP\Calendar\Exceptions\CalendarException;
+use OCP\Calendar\ICalendar;
+use Sabre\VObject\Component\VTimeZone;
+
+interface IGetTimezone extends ICalendar {
```
ChristophWurst, line 31: what is the purpose of this new interface and public API? it's not programmed against anywhere, is it?

miaulalala: The CalendarImpl implements it. But Thomas suggested using the new Timezone stuff in https://github.com/nextcloud/server/pull/36192 so I will look into that one

ChristophWurst: It implements it but there is no code that uses it. Hence there is no need for a public API here.

miaulalala: https://github.com/nextcloud/calendar/pull/4969/files

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -1847,7 +1848,24 @@ public function search(array $calendarInfo, $pattern, array $searchProperties,
...
+				$outerQuery->andWhere($or1);
+			}
+			if (isset($options['timerange']['end']) && $options['timerange']['end'] instanceof DateTimeInterface) {
+				$or2 = $outerQuery->expr()->orX();
+				$or2->add($outerQuery->expr()->gt('firstoccurence',
+					$outerQuery->createNamedParameter($options['timerange']['start']->getTimeStamp())));
+				$or2->add($outerQuery->expr()->isNull('firstoccurence'));
+				$outerQuery->andWhere($or2);
+			}
+		} else {
```
miaulalala, line 1868: The following problem presents itself when adding VTODO support for compined searches where `[options => [ types => [VEVENT, VTODO]], [timerange => [start => 123456]]]`

A VTODO will never have a `firstoccurence`/`lastoccurence` set (db entry `null`), but a VEVENT timerange filter will always look for it, thus ignoring `null` values.

My changes would allow the query to do an `AND WHERE(firtsoccurence = 12346 OR firstoccurence IS NULL)` when the `BTODO` type is passed.

(The filters need adjusting too further below)

miaulalala:
Superceded by https://github.com/nextcloud/server/pull/39937

PR #36621, opened 2023-02-08
fix(performance): Only search for auth tokens when the provided login…

nickvergessen:
… is long enough

## Summary

All our generated auth tokens have a a fixed length (apart from session id based one which has a minimum). So it's save to assume that shorter strings are not actually valid apptokens we need to look up but the password when basic auth is used.
Additionally a min length check was added to help developers on generating new tokens in the future.

## Query count of notifications app integration tests

* Ref https://github.com/nextcloud/notifications/pull/1452

Before | After
---|---
https://github.com/nextcloud/notifications/actions/runs/4108063021/jobs/7088301789 | https://github.com/nextcloud/notifications/actions/runs/4128695573/jobs/7133455445
10354 | 9150

## Checklist

### DIFF
--- a/lib/private/Authentication/Token/PublicKeyTokenProvider.php
+++ b/lib/private/Authentication/Token/PublicKeyTokenProvider.php
@@ -112,6 +112,27 @@ class PublicKeyTokenProvider implements IProvider {
 	}
 
 	public function getToken(string $tokenId): IToken {
+		/**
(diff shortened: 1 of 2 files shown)


miaulalala (approved):
:elephant:

juliusknorr (approved):

come-nc (approved):

PR #36637, opened 2023-02-09
Safer URL match in FileReferenceProvider

julien-nc:
This was matching URLs like `https://instance.yours/apps/files_sharing/accept/ocinternal:123456`.

Also, it was crashing when the URL does not have any query params.


nickvergessen (approved):
🙈

julien-nc:
This can be safely backported to stable25 IMO.

juliusknorr (approved):

PR #36639, opened 2023-02-09
also cache backend for user in memory instead of always going to redis

icewind1991:
remote caches aren't *that* cheap.

There are probably some other cases that do this manually that can be replaced with `WithLocalCache`, but that is left as an exercise for a follow up PR.

### DIFF
--- a/lib/private/Memcache/WithLocalCache.php
+++ b/lib/private/Memcache/WithLocalCache.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace OC\Memcache;
+
+use OCP\Cache\CappedMemoryCache;
+use OCP\ICache;
+
+/**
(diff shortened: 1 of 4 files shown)


come-nc (approved):
CappedMemoryCache in user_ldap could benefit from using this instead I think

lib/private/Memcache/WithLocalCache.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace OC\Memcache;
+
+use OCP\Cache\CappedMemoryCache;
+use OCP\ICache;
+
+/**
+ * Wrap a cache instance with an extra later of local, in-memory caching
```
ChristophWurst, line 9: yo dawg

ChristophWurst (approved):

PR #36646, opened 2023-02-09
perf: No need to setup the full filesystem for getting storage info

juliusknorr:
## Summary

getStorageInfo also works without a full filesystem setup, so we can save a decent amount of time on the initial page load.

full setup might still happen from other places but this can avoid it depending on the setup and if the user has favourites or not.

## Checklist


icewind1991:
See also https://github.com/nextcloud/server/pull/36608

juliushaertl:
Even better.

juliushaertl:
Ah this one makes still sense.

icewind1991 (approved):

come-nc (approved):

PR #36656, opened 2023-02-10
more routing performance instrumentation

icewind1991:
Signed-off-by: Robin Appelman <robin@icewind.nl>

### DIFF
--- a/lib/private/Route/Router.php
+++ b/lib/private/Route/Router.php
@@ -35,8 +35,11 @@ namespace OC\Route;
 use OC\AppFramework\Routing\RouteParser;
 use OCP\AppFramework\App;
 use OCP\Diagnostics\IEventLogger;
+use OCP\IConfig;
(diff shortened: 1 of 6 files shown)


juliusknorr (approved):

lib/private/Route/Router.php
```diff
@@ -82,6 +84,7 @@ public function __construct(LoggerInterface $logger) {
 		$this->context = new RequestContext($baseUrl, $method, $host, $schema);
 		// TODO cache
 		$this->root = $this->getCollection('root');
+		$this->eventLogger = \OC::$server->get(IEventLogger::class);
```
come-nc, line 87: Why no DI?

come-nc (approved):

lib/private/Route/Router.php
icewind1991: lazyness :see_no_evil:

PR #36660, opened 2023-02-10
enh: Allow prefetch-src self in CSP

juliusknorr:
## Summary

Make sure that we can prefetch resources from our own domain e.g. with [webpack module prefetching](https://webpack.js.org/guides/code-splitting/#prefetchingpreloading-modules)

## Checklist

### DIFF
--- a/lib/public/AppFramework/Http/EmptyContentSecurityPolicy.php
+++ b/lib/public/AppFramework/Http/EmptyContentSecurityPolicy.php
@@ -446,6 +446,7 @@ class EmptyContentSecurityPolicy {
 		$policy = "default-src 'none';";
 		$policy .= "base-uri 'none';";
 		$policy .= "manifest-src 'self';";
+		$policy .= "prefetch-src 'self';";
 
 		if (!empty($this->allowedScriptDomains) || $this->inlineScriptAllowed || $this->evalScriptAllowed) {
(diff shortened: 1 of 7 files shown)


tcitworld:
Not exposed to public API?

marcelklehr (approved):

juliushaertl:
> Not exposed to change via public API?

Mainly did this for apps to prefetch their javascript, so I'd keep it rather simple for now. We can extend this if we have use cases for adding external sources.

marcelklehr:
[No browser seems to support this](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/prefetch-src#browser_compatibility), yet.

tcitworld:
Latest Safari [has support for it](https://webkit.org/blog/13691/webkit-features-in-safari-16-3/), but Chromium [wants to remove it](https://bugs.chromium.org/p/chromium/issues/detail?id=1406444). It was also removed from the spec a few weeks ago: https://github.com/w3c/webappsec-csp/pull/582

juliushaertl:
Good catch, so I actually had a CSP error when testing this the first time, but I cannot reproduce that anymore, so maybe I was just lacking he nonce to be set with web pack there. Will close this.

PR #36667, opened 2023-02-10
Fix OCM validation errors

provokateurin:
## Summary

According to the [spec](https://github.com/cs3org/OCM-API/blob/c590a3d6c2c0388bb63b76c7c4c30cc9aed5f75d/spec.yaml#L190) the `validationErrors` is an array of objects.

## Checklist


ChristophWurst (approved):

nickvergessen (approved):

provokateurin:
Technically this should be backported to every version, since it's a non-compliant implementation of the spec, but so far nobody complained, so it's probably not a big deal anyway, right?

PR #33608, opened 2022-08-18
Remove the loop of calendars when only one is needed

miaulalala:
Fixes #33599

PR #36681, opened 2023-02-13
feat(dav): Directly access shared calendars by URI

tcitworld:
Each time we access a shared calendar details (with getChild), this loads every user's calendars and iterates until the correct one is found. Now we lookup from the shared calendars which ones has the correct URI, just iterating on the possible shares (for instance a calendar can be shared directly through user-to-user share or through user-to-group share).

Didn't mesure properly the performance benefits, but should be really nice with many shared calendars.

Follow-up to https://github.com/nextcloud/server/pull/33608

## TODO

- [ ] Add some tests for this new method

## Checklist


susnux:
Hello :wave: 
currently here have been no activity in the last years.
If this is still actual feel free to reopen the pull request and resolve the conflicts so we can incorporate it into Nextcloud.
Thank you very much for your contribution!

PR #36683, opened 2023-02-13
[stable25] Fix missing files:navigation:changed when clicking on a dir in the favorites view

julien-nc:
While being in the favorites view, if one clicks a directory, the `files:navigation:changed` event is not emitted so Text can't update the rich workspace visibility.

This fixes https://github.com/nextcloud/text/issues/3697 but has the side effect to emit the event twice when clicking on the "Favorites" nav item while being in the "All files" view.
As `files:navigation:changed` is only listened by Text, it does not seems too bad.

An alternative fix would be to remove the silent option from https://github.com/nextcloud/server/blob/6bb0985e59d7ba3c4cf9928d8bb766ac9975cd12/apps/files/js/favoritesplugin.js#L98
but we don't know exactly what side effect it could have.

### DIFF
--- a/apps/files/js/app.js
+++ b/apps/files/js/app.js
@@ -200,6 +200,7 @@
 				const isGridView = this.$showGridView.is(':checked');
 				this.currentFileList.setGridView(isGridView);
 			}
+			window._nc_event_bus.emit('files:navigation:changed')
 		},
 
 		/**


juliushaertl:
Doesn't the urlChanged jQuery event trigger and catch this in text?

https://github.com/nextcloud/server/blob/fix/stable25/files.navigation.changed-when-entering-favorite-dir/apps/files/js/app.js#L320

https://github.com/nextcloud/text/blob/stable25/src/helpers/files.js#L180-L182

julien-nc:
@juliushaertl Problem is, in https://github.com/nextcloud/server/blob/fix/stable25/files.navigation.changed-when-entering-favorite-dir/apps/files/js/app.js#L311
`_onNavigationChanged` is not called when clicking on a directory in the Favorites view. IMO it should be called because we "leave" the favorites view to enter a directory so we're back in the "All files" view. This is what's missing for Text to recompute the rich workspace visibility.

I checked and the `urlChanged` event is emitted when clicking on the "All files" entry in the navigation menu but **not** when entering a directory in the favorites view.

julien-nc:
Rebased on stable 25
Updated the PR description.
Ready for review.

juliusknorr (approved):

skjnldsv (approved):

PR #36701, opened 2023-02-14
Add parent index on filecache

artonge:
To speed up:
https://github.com/nextcloud/server/blob/fd2afe604ef1d8ef6d5ae9d497b0f01ae159b2f6/lib/private/Files/Cache/Cache.php#L878-L880

The current query uses the `parent_name_hash` index, which is not optimal with long names.

|               | id  | select_type | table        | type | possible_keys                           | key                 | key_len | ref   | rows |
| ------------- | --- | ----------- | ------------ | ---- | --------------------------------------- | ------------------- | ------- | ----- | ---- |
| **Without index** | 1   | SIMPLE      | oc_filecache | ref  | fs_parent_name_hash                     | fs_parent_name_hash | 8       | const | 5    |
| **With index**    | 1   | SIMPLE      | oc_filecache | ref  | fs_parent_name_hash,oc_filecache_parent | oc_filecache_parent | 8       | const | 5    |

### DIFF
--- a/core/Application.php
+++ b/core/Application.php
@@ -132,6 +132,10 @@ class Application extends App {
 					if (!$table->hasIndex('fs_storage_path_prefix') && !$schema->getDatabasePlatform() instanceof PostgreSQL94Platform) {
 						$subject->addHintForMissingSubject($table->getName(), 'fs_storage_path_prefix');
 					}
+
+					if (!$table->hasIndex('fs_parent')) {
+						$subject->addHintForMissingSubject($table->getName(), 'fs_parent');
+					}
 				}
 
 				if ($schema->hasTable('twofactor_providers')) {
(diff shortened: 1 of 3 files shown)


icewind1991 (approved):

icewind1991 (dismissed):
Also needs to be added to the migration that creates the filecache table

artonge:
Waiting for more practical feedbacks before moving on.

icewind1991 (approved):

juliushaertl:
Can you outline or do you have a reference why the single index would be faster? At least from the explain there seems to be no difference.

juliusknorr (approved):

juliushaertl:
Failure unrelated

PR #36727, opened 2023-02-15
feat(directediting): Allow opening by file id

juliusknorr:
## Summary

Make sure that it is easy for the notes iOS/Android apps to open files by just knowing the file id. For cases where multiple files with the id are found we prefer the one with update permissions over others.

## Checklist

### DIFF
--- a/lib/private/DirectEditing/Manager.php
+++ b/lib/private/DirectEditing/Manager.php
@@ -26,10 +26,11 @@
 namespace OC\DirectEditing;
 
 use Doctrine\DBAL\FetchMode;
-use OC\Files\Node\Folder;
+use \OCP\Files\Folder;
 use OCP\AppFramework\Http\NotFoundResponse;
 use OCP\AppFramework\Http\Response;
 use OCP\AppFramework\Http\TemplateResponse;
+use OCP\Constants;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\DirectEditing\ACreateFromTemplate;
 use OCP\DirectEditing\IEditor;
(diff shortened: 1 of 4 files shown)


lib/private/DirectEditing/Manager.php
```diff
@@ -149,9 +150,25 @@ public function create(string $path, string $editorId, string $creatorId, $templ
 		throw new \RuntimeException('No creator found');
 	}
 
-	public function open(string $filePath, string $editorId = null): string {
-		/** @var File $file */
+	public function open(string $filePath, string $editorId = null, ?int $fileId = null): string {
+		$userFolder = $this->rootFolder->getUserFolder($this->userId);
 		$file = $this->rootFolder->getUserFolder($this->userId)->get($filePath);
```
mejo-, line 155:
```suggestion
		$file = $userFolder->get($filePath);
```

juliushaertl: Thanks, good catch. Fixed.

AlvaroBrey:
@juliushaertl any possibility to add a capability indicating whether this change is available? Otherwise I'll just check if server version >= 26

juliushaertl:
Capability added:

Screenshot 2023-02-16 at 15 02 33

max-nextcloud (approved):
Code looks good.

max-nextcloud:
Maybe add a test?
I did not find any tests for this controller - but I may have missed them. However there's https://github.com/nextcloud/server/blob/master/tests/lib/DirectEditing/ManagerTest.php which might be fairly easy to extend?

juliushaertl:
Pushed two further cases to cover the open method with and without a file id.

mejo- (approved):
Code looks good to me.

marcelklehr (approved):

juliushaertl:
Failure unrelated.

marinofaggiana:
@juliushaertl  thanks

marinofaggiana:
@juliushaertl I have tested with:

https://cloud.nextcloud.com/ocs/v2.php/apps/files/api/v1/directEditing/open?fileId=7155303&editorId=text
Response Error 500

AlvaroBrey:
> @juliushaertl I have tested it, example:
> 
> https://cloud.nextcloud.com/ocs/v2.php/apps/files/api/v1/directEditing/open?fileId=7155303&editorId=text Response Error 500

You need to also provide `path`, with the value being the notes storage path (from https://github.com/nextcloud/notes/blob/master/docs/api/v1.md#settings).

PR #32635, opened 2022-05-27
Fix User profile picture when performing the search

AndyXheli:
Signed-off-by: Andy Xheli <axheli@axtsolutions.com>

Before 
image

After
2022-05-27 14_47_23-Dashboard - AXTSolutions Cloud

Fix for #31065

PR #36735, opened 2023-02-15
fix the contacts-menu vcf-contact avatars

szaimen:
Regression of https://github.com/nextcloud/server/pull/32635

Fixes avatars in the contacts-menu of vcf-contacts

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/lib/private/Contacts/ContactsMenu/ContactsStore.php
+++ b/lib/private/Contacts/ContactsMenu/ContactsStore.php
@@ -287,7 +287,13 @@ class ContactsStore implements IContactsStore {
 		if (isset($contact['UID'])) {
 			$uid = $contact['UID'];
 			$entry->setId($uid);
-			$avatar = $this->urlGenerator->linkToRouteAbsolute('core.avatar.getAvatar', ['userId' => $uid, 'size' => 64]);
+			if (isset($contact['isLocalSystemBook'])) {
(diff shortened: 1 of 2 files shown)


lib/private/Contacts/ContactsMenu/ContactsStore.php
```diff
@@ -287,8 +287,15 @@ private function contactArrayToEntry(array $contact): Entry {
 		if (isset($contact['UID'])) {
 			$uid = $contact['UID'];
 			$entry->setId($uid);
-			$avatar = $this->urlGenerator->linkToRouteAbsolute('core.avatar.getAvatar', ['userId' => $uid, 'size' => 64]);
-			$entry->setAvatar($avatar);
+			$avatar = '';
+			if ($this->userManager->userExists($uid)) {
```
szaimen, line 291: @nextcloud/server-backend do you know of a better way to distinguish users from vcf-contacts? Both seem to hav a uid and fn and other stuff ...

nickvergessen: See isLocalSystemBook above

szaimen: Thanks!

blizzz (approved):

artonge (approved):

PR #36736, opened 2023-02-15
Use proc_open to avoid spawning a shell

Glandos:
## Summary
The use of `exec` will spawn a shell, using `/bin/sh` on POSIX platforms. But in restricted environment, such as AppArmor, this means giving execution to `/bin/sh`, which renders the execution restriction quite useless.
Using an array with `proc_open` reduces this, and paved the way for file streaming instead of temporary file.

Signed-off-by: Glandos <bugs-github@antipoul.fr>

## Checklist


come-nc (approved):

skjnldsv (approved):
Code make sense

Issue #36713, opened 2023-02-15
[BITV] 9.1.1.1a/2.2 - The avatar buttons, which have been implemented as links, in the "Search contacts" dialog do not have alternative text. (9)


Pytal:
- [x] https://github.com/nextcloud/server/pull/36740
- [x] https://github.com/nextcloud/server/pull/37047

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v2-n2

![image](https://user-images.githubusercontent.com/24800714/218899672-d2700f98-643e-495c-a0aa-8d02b6ea1f1f.png)

</details>

- [x] 9.1.1.1a/3.6 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v3-n6
- [x] 9.1.1.1a/4.2 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v4-n2
- [x] 9.1.1.1a/5.4 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v5-n4
- [x] 9.1.1.1a/6.3 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v6-n3
- [x] 9.1.1.1a/7.3 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v7-n3
- [x] 9.1.1.1a/8.2 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v8-n2
- [x] 9.1.1.1a/9.3 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v9-n3
- [x] 9.1.1.1a/10.6 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v10-n6

PR #36740, opened 2023-02-16
Add avatar contact img description

Pytal:
- For https://github.com/nextcloud/server/issues/36713

## Summary

Add descriptive text to the contacts menu avatar `img` elements

## Checklist


jotoeri (commented):
Hmm, isn't that worse, as the screenreader reads the name on the avatar and then once again on the actual full-name text?

Pytal:
> Hmm, isn't that worse, as the screenreader reads the name on the avatar and then once again on the actual full-name text?

Do you think we should add a separate label for the avatar @michaelnissenbaum?

michaelnissenbaum:
@JuliaKirschenheuter From my perspective, Avatar should not be implemented as a link because we already have a 'View Profile' link that provides the same functionality. Avatar should simply be implemented as an img element with an "alt" attribute set to an empty string, so that screen readers will ignore this element. However, if you really want the Avatar element to remain interactive and be displayed, then you can use the aria-label attribute with an "a"-element or the "alt" attribute of the image.

JuliaKirschenheuter (commented):
would be better to set `Avatar of __name__`? Or should we through away a link from avatar https://github.com/nextcloud/server/pull/36740#issuecomment-1439878082?

Pytal:
> would be better to set `Avatar of __name__`? Or should we through away a link from avatar [#36740 (comment)](https://github.com/nextcloud/server/pull/36740#issuecomment-1439878082)?

Makes sense UX-wise to have it clickable so updated the description!

kesselb (approved):

JuliaKirschenheuter (approved):

Issue #26668, opened 2021-04-20
Send invitations for shared calendars


kesselb:
Ref: https://github.com/nextcloud/server/pull/15676

It's not possible (anymore) to send invitations for an event for shared calendars. This behavior is confusing for end users.

> 
> Allowing sharees to send invitation in shared calendars was a mistake back then, there are many unresolved topics, especially if the sharee is also an attendee of the event:
> 
> - when editing, does the sharee edit for themselves or on behalf of the calendar owner?
> - when deleting, is the sharee declining the event for themselves or cancelling it for everyone?
> - ...
>
> The CalDAV standard does not have answers to these problems yet. Until then, we should not allow sharees to send invitations in shared calendars.

https://sabre.io/dav/caldav-sharing/ mentions 
- https://tools.ietf.org/html/draft-pot-webdav-resource-sharing-04 
- https://tools.ietf.org/html/draft-pot-caldav-sharing-00

Unfortunately there is no final RFC how shared calendars should work. Yet the drafts could be a starting point. draft-pot-caldav-sharing-00 describes those open issues:

   1.  DAV:owner requirement for scheduling.  I think this is
       problematic...

   2.  I don't think we should allow sharees that have access to an
       invite for which they are the attendee for, via the organizers
       shared calendar, to allow them to make attendee-related changes.
       The entire collection should operate as if the operation is on
       behalf of the organizer.

We don't implement https://github.com/sabre-io/dav/blob/master/lib/CalDAV/Backend/SharingSupport.php interface yet. That might be something to start with. There is already some logic connected to the above interface in sabre that might be useful.

The example implementation https://github.com/sabre-io/dav/blob/c1afdc77a95efea6ee40c03c45f57c3c0c80ec22/lib/CalDAV/Backend/PDO.php#L1308-L1465 uses a database table [calendarinstances](https://github.com/sabre-io/dav/blob/c1afdc77a95efea6ee40c03c45f57c3c0c80ec22/examples/sql/mysql.calendars.sql#L23-L41) that looks similar to `oc_calendars` and `oc_dav_shares`. For example `calendarcolor` is something we store with the calendar. But for a shared calendar the color is something a user may change. So this needs to go to `oc_dav_shares` (calendarinstance in the sabre example).

ChristophWurst:
Some related sharing limitations are discussed at https://github.com/nextcloud/server/issues/5050.

favincen:
[...]

PR #36766, opened 2023-02-17
CalDavBackend, implement SharingSupport interface.

rotdrop:
* Related: #26668

## Summary

Implement the `\Sabre\CalDAV\Backend\SharingSupport` in `\OCA\DAV\CalDAV\CalDavBackend`. The starting comment of the linked issue #26668 suggests that this might be the first step in supporting invitations for events in shared writeable calendars. However, after coding #36756 I have now the impression that the CalDAV sharing interface is orthogonal to sending invitations for events in shared calendars. It is nice to have and may help the calendar frontend (see nextcloud/calendar#4983). However, implementing the "SharingSupport" interface in the CalDAV backend does not bring us closer to sending invitations.

## TODO

- [ ] testing

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/CalDavBackend.php
+++ b/apps/dav/lib/CalDAV/CalDavBackend.php
@@ -80,6 +80,7 @@ use Sabre\CalDAV\Backend\AbstractBackend;
 use Sabre\CalDAV\Backend\SchedulingSupport;
 use Sabre\CalDAV\Backend\SubscriptionSupport;
 use Sabre\CalDAV\Backend\SyncSupport;
+use Sabre\CalDAV\Backend\SharingSupport;
 use Sabre\CalDAV\Xml\Property\ScheduleCalendarTransp;
 use Sabre\CalDAV\Xml\Property\SupportedCalendarComponentSet;
 use Sabre\DAV;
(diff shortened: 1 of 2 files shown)


tcitworld (commented):
First of all, thanks for the effort. As #26668 hints, this might be valuable in the long term.

In the short term, though, it shouldn't bring much: the exposed API is very similar to what we currently have with the `oc-resource-sharing` custom plugin, which was based on Sabre's experiments at the time, themselves based on [CalendarServer custom's spec](https://github.com/apple/ccs-calendarserver/blob/master/doc/Extensions/caldav-sharing.txt).

Since all of @evert's drafts have long expired and Apple has abandoned CalendarServer, it's not sure if it's worth it to pursue in this direction until we have hints of the ecosystem going into the same direction (it's been 7 years since the drafts publication, and 4 years since the last mention of this in the CalConnect/IETF discussions that I can find).

Another thing is that SharingSupport is very linked to `NotificationSupport` (as seen in [the Sharing plugin calling `sharingReply`](https://github.com/nextcloud/3rdparty/blob/0ec73636ee36558960a1df60828b645ef3d4a53c/sabre/dav/lib/CalDAV/SharingPlugin.php#L244-L250) (as derived from the CS origin). 
The notification plugin [even says CalDAV-sharing requires it](https://github.com/nextcloud/3rdparty/blob/c86fa3ed21786c646ef6c5e066a3a1b6f7505a71/sabre/dav/lib/CalDAV/Notifications/Plugin.php#L157). So that's a whole other thing, even if it would bring nice to have features, such as macOS/iOS calendar clients eventual compatibility.

sorbaugh:
@rotdrop following @tcitworld, it looks like this PR won't be pursued any longer? Please let me know if you'd like us to reopen this!

Issue #36720, opened 2023-02-15
[BITV] "Search" dialog - The items in the search result list have poor visibility when focused with the keyboard.


Pytal:
- [ ] Fix
- [ ] Backport

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-91f2646afa-v2-n1

![image](https://user-images.githubusercontent.com/24800714/218904002-575f51e9-30d1-4277-9199-1b0e6ce3d06b.png)

</details>

PR #36771, opened 2023-02-18
Add focus border for search results

Pytal:
- For https://github.com/nextcloud/server/issues/36720

## Summary

Before | After
--- | ---
image | image

## Checklist


szaimen (approved):
LGTM but didnt test

core/src/components/UnifiedSearch/SearchResult.vue
```diff
@@ -180,11 +180,15 @@ $margin: 10px;
...
-	&--focused,
+	&--focused {
+		background-color: var(--color-background-hover);
+	}
+
 	&:active,
 	&:hover,
 	&:focus {
 		background-color: var(--color-background-hover);
+		box-shadow: inset 0 0 0 2px var(--color-border-maxcontrast);
```
artonge, line 191: Isn't the border enough?

Pytal: As in use the `border` CSS property?

If so, `box-shadow` was used here instead because it requires less code changes due to `border` taking up space

artonge: If it is really too many changes, then ok, but if possible, I think it would be better to use the semantically correct attribute here.

Would `box-sizing; border-box` help ?

Pytal: Since we depend on padding to fill out the desired highlight area it seems using a transparent border is the more intutive styling option vs. `box-sizing: border-box` + changing the height and width and increasing the margins of child elements to achieve the same design

artonge (approved):

Issue #34932, opened 2022-11-02
[Bug]: `strpos()` and/or `pathinfo()`: Passing null to parameter #1 ($haystack) of type string is deprecated


killerb77:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

I am using the official docker image nextcloud 25.  I am getting all kinds of error PHP messages regarding of type string is deprecated at.  It's happening on file uploads, when nextcloud sends email notifications as well.  I think it's related to the new version 25 using PHP 8.1.  It started happening for version 25.  Didn't happen on version 24 which uses PHP 8.0.

![198488279-52dbfcb6-429e-45a0-b0b0-d4c941842b21](https://user-images.githubusercontent.com/56096557/199617763-20043c14-60af-4136-982b-1e20b046fa52.png)

### Steps to reproduce

1.Send a test email notification.
2.Or upload a file.
3.Then check the logs to see the error messages.

### Expected behavior

Should not be getting these error messages.

### Installation method

Community Docker image

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "htaccess.RewriteBase": "\/",
        "memcache.local": "\\OC\\Memcache\\APCu",
        "apps_paths": [
            {
                "path": "\/var\/www\/html\/apps",
                "url": "\/apps",
                "writable": false
            },
            {
                "path": "\/var\/www\/html\/custom_apps",
                "url": "\/custom_apps",
                "writable": true
            }
        ],
[...]

PR #36589, opened 2023-02-07
fix(performance): Do not set up filesystem on every call

miaulalala:
Removed old OO_FileChunking logic that produced GC- collectable chunks

Added BG Job that will run a garbage collection for every user

Signed-off-by: Anna Larch <anna@nextcloud.com>

* Resolves: https://github.com/nextcloud/server/issues/31357
## Summary

Comparison for OCS requests:
https://blackfire.io/profiles/compare/2c3c3699-137e-49a3-a363-e0950bb278d4...41a33cf0-71ee-42aa-9791-9ef74f99e6c3/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main()&constraintDoc=
Screenshot 2023-02-16 at 16 52 52

## TODO

- [ ] Testing

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [ ] ~~Screenshots before/after for front-end changes~~ 
- [ ] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] ~~[Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)~~

PR #36774, opened 2023-02-18
fix: always use proper path on node api when calling the view

juliusknorr:
Make sure that the nodes API always uses the full path and root view.

This is only noticeable on places where the dav app already uses the new nodes api but could cause trouble in the future.

Seen this being an issue when the dav app passes a relative path and fake root view for the user but the full file system setup was dropped by #36589 when:

- Incoming share was moved to a subfolder A/B/share
- Perform the following query: `curl -X PROPFIND http://admin:admin@nextcloud.dev.local/remote.php/webdav/A -H 'Depth: 2'`

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Node.php
+++ b/apps/dav/lib/Connector/Sabre/Node.php
@@ -91,11 +91,13 @@ abstract class Node implements \Sabre\DAV\INode {
 		if ($info instanceof Folder || $info instanceof File) {
 			$this->node = $info;
 		} else {
+			// The Node API assumes that the view passed doesn't have a fake root
+			$rootView = \OC::$server->get(View::class);
 			$root = \OC::$server->get(IRootFolder::class);
 			if ($info->getType() === FileInfo::TYPE_FOLDER) {
-				$this->node = new Folder($root, $view, $this->path, $info);
+				$this->node = new Folder($root, $rootView, $this->fileView->getAbsolutePath($this->path), $info);
 			} else {
-				$this->node = new File($root, $view, $this->path, $info);
+				$this->node = new File($root, $rootView, $this->fileView->getAbsolutePath($this->path), $info);
 			}
 		}
 	}
@@ -107,10 +109,11 @@ abstract class Node implements \Sabre\DAV\INode {
 		}
 		$this->info = $info;
 		$root = \OC::$server->get(IRootFolder::class);
+		$rootView = \OC::$server->get(View::class);
 		if ($this->info->getType() === FileInfo::TYPE_FOLDER) {
-			$this->node = new Folder($root, $this->fileView, $this->path, $this->info);
+			$this->node = new Folder($root, $rootView, $this->path, $this->info);
 		} else {
-			$this->node = new File($root, $this->fileView, $this->path, $this->info);
+			$this->node = new File($root, $rootView, $this->path, $this->info);
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


miaulalala (approved):
Makes sense to me but I'm no expert.

come-nc (approved):
Does that fix https://github.com/nextcloud/server/issues/34932#issuecomment-1369040458 ?

juliushaertl:
> Does that fix [#34932 (comment)](https://github.com/nextcloud/server/issues/34932#issuecomment-1369040458) ?

Not directly I think, but we could probably refactor to just avoid using the View in https://github.com/nextcloud/server/blob/5d4efb4d5fd8e4389856df5d94c3b92c7019e603/apps/dav/lib/Connector/Sabre/Directory.php#L115-L151 and go directly through the Nodes API instead. This would get rid of the separate handling of hooks in this case in the dav app.

artonge (approved):

lib/private/Files/Node/Folder.php
```diff
@@ -98,7 +98,8 @@ public function isSubNode($node) {
 	 * @throws \OCP\Files\NotFoundException
 	 */
 	public function getDirectoryListing() {
-		$folderContent = $this->view->getDirectoryContent($this->path, '', $this->getFileInfo(false));
+		$relativePath = $this->view->getRelativePath($this->path);
+		$folderContent = $this->view->getDirectoryContent($relativePath, '', $this->getFileInfo(false));
```
icewind1991, line 102: This would break things if any code path ends up passing a relative `path` and a non-root `View` where it would currently work.

If we're going to rely on always having a root `View`, we should check for this in the `Node` constructor (and adjust the path/view and log when a non-root `View` is encountered) and these changes can then be removed.

icewind1991 (dismissed):
`Sabre\Node::refreshInfo` will also need updating.

Overall the change is the right direction imo, and ensuring some consistency here should also make an eventual move to using the `Node` api directly a bit easier

lib/private/Files/Node/Folder.php
juliushaertl: Sounds good, checking makes sense as overall the node code seems to rely on the absolute path

juliushaertl: Added a constructor check, let's see if this causes any breakage on the tests

nickvergessen:
I think we have to revert https://github.com/nextcloud/server/pull/36589 because it causes a major regression with Talk/ attachment handling, see https://github.com/nextcloud/server/issues/36787

I also tested with this PR here and Talk/ is still not working. Given that we are behind feature freeze I vote to delay this here, revert the other PR on master until stable26 is branched off and try again for 27.l

nickvergessen (dismissed):
blocking as per above

apps/dav/lib/Connector/Sabre/Node.php
```diff
@@ -91,11 +91,13 @@ public function __construct(View $view, FileInfo $info, IManager $shareManager =
 		if ($info instanceof Folder || $info instanceof File) {
 			$this->node = $info;
 		} else {
+			// The Node API assumes that the view passed doesn't have a fake root
```
nickvergessen, line 94: Can we add an error log in case this happens, so devs become more aware?

juliushaertl: It actually throws an exception at https://github.com/nextcloud/server/pull/36774/files#diff-edc136e5646ea4171607db53b38833e8f6caaacc64874a0a850aa4444f0a56edR77

PR #36776, opened 2023-02-18
Fix wrong class name in CSS for UserList

sreith1:
Class name is .subAdminsGroups not .subadmins


skjnldsv:
As there is no feedback since a while I will close this ticket. If this is still happening please make sure to upgrade to the latest version. After that, feel free to reopen.

PR #35223, opened 2022-11-17
Smaller performance optimizations

juliusknorr:
This is a collection of smaller performance improvements in different areas that I've found when looking at some profiles in regards to class loading, dependency injection and server initialization. The commit messages contain more details about the individual changes.

- Avoid container dance for appName
- Skip querying the app container for server namespace
- Take some shortcut when setting up the application class
- Use single resource locator instance
- Set apcu prefix for composer
- Cache available languages locally

https://blackfire.io/profiles/compare/b5552647-5f7f-4848-93c7-862084400a89/graph for comparison

While the changes itself are rather small optimizations in total they contribute quite a bit to the overall CPU time, especially since some of the calls depend on the number of enabled apps.

Screenshot 2022-11-17 at 12 07 42

PR #36780, opened 2023-02-20
fix: Avoid throwing errors on public auth page

juliusknorr:
Avoid throwing errors on the public share page like:

> Cannot read properties of null (reading 'addEventListener')

https://juliushaertl.sentry.io/share/issue/2d96706d8b094bd7b5b780761c11969c/

## Checklist

### DIFF
--- a/lib/private/Template/CSSResourceLocator.php
+++ b/lib/private/Template/CSSResourceLocator.php
@@ -43,8 +43,7 @@ class CSSResourceLocator extends ResourceLocator {
 	 */
 	public function doFind($style) {
 		$app = substr($style, 0, strpos($style, '/'));
-		if (strpos($style, '3rdparty') === 0
-			&& $this->appendIfExist($this->serverroot, $style.'.css')
+		if ($this->appendIfExist($this->serverroot, $style.'.css')
 			|| $this->appendIfExist($this->serverroot, 'core/'.$style.'.css')
 		) {
 			return;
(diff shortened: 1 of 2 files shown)


szaimen (approved):
LGTM but didnt test

juliushaertl:
Pushed another related fix as styles from core/css were not loaded if called with `\OC_Util:.addStyle('core', 'publicshareauth`)`.

lib/private/Template/CSSResourceLocator.php
```diff
@@ -43,8 +43,7 @@ public function __construct(LoggerInterface $logger) {
 	 */
 	public function doFind($style) {
 		$app = substr($style, 0, strpos($style, '/'));
-		if (strpos($style, '3rdparty') === 0
```
artonge, line 46: Seems wrong. Is this removing the check that the style name should not contain `'3rdparty'` ?
Not sure about why this helps.

juliushaertl: This is actually a regression that I produced when dropping the 3rdparty handling in the ResourceLocator in https://github.com/nextcloud/server/pull/35223/commits/4bcaeb6c2cd0e9c5bb87995aac8df849bedd0b70 but didn't remove this part. 

Movable 3rdparty support was dropped ages ago anyways https://github.com/nextcloud/server/commit/dc061bae42293df936c138f0db32fb6d308042f6

artonge: Thanks for the explanation, wasn't even aware of that feature :)

juliushaertl: Side-by-side diff from before 4bcaeb6 makes it clearer on why this worked before:

`strpos($style, '3rdparty')` is always false here so the `&&` is skipped early but the next check for `||` will evaluate. 

Screenshot 2023-02-20 at 12 33 04

artonge (approved):

PR #36792, opened 2023-02-21
New BeforeGroupChangedEvent before setDisplayName on groups

julien-nc:
This could be useful to let apps prevent renaming of some groups.
This was mentioned by the developers of the [OpenProject integration](https://github.com/nextcloud/integration_openproject/#openproject-integration).

@miaulalala Why isn't the old value passed to the `GroupChangedEvent`?
https://github.com/nextcloud/server/blob/master/lib/private/Group/Group.php#L116


miaulalala (approved):
To answer your question: I'm not sure, must have been an oversight.

PR looks good!

icewind1991 (approved):

julien-nc:
Thanks for the reviews!

Psalm failure is unrelated.

juliushaertl:
Drone failure was related. Pushed updated autoloaders in https://github.com/nextcloud/server/pull/36799

PR #36795, opened 2023-02-21
Avoid extra parenthesis around SQL in IN expressions

### LINKED ISSUE #34836: Subquery inside IN must not use additional () on the createFunction() call or it breaks in SQLite
And seems there are at least 2 others in server (middle one is not an IN, so it might still work?):
image

So we should try to find a programmatic way to prevent this.
Can you raise a tech debt ticket in the server about it?

_Originally posted by @nickvergessen in https://github.com/nextcloud/spreed/issues/8220#issuecomment-1288651806_

come-nc:
## Summary

The expression builder already suround the SQL with parenthesis when
 using in(), so we must not add another pair, this confuses at least
 sqlite.

## Checklist

### DIFF
--- a/apps/files_sharing/lib/External/Manager.php
+++ b/apps/files_sharing/lib/External/Manager.php
@@ -660,7 +660,7 @@ class Manager {
 
 
 		$query->delete('federated_reshares')
-			->where($query->expr()->in('share_id', $query->createFunction('(' . $select . ')')));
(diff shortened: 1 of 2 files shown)


come-nc:
Follow up of https://github.com/nextcloud/spreed/pull/8220#issuecomment-1288651806

nickvergessen (approved):

icewind1991 (approved):

PR #36803, opened 2023-02-22
[db]: Remove not supported column comments for SQLite

susnux:
* Resolves https://github.com/nextcloud/forms/pull/1479#issuecomment-1439234202

## Summary
Some times column comments are used, e.g. to make clear an integer is used as a timestamp.
For SQLite column comments are not supported and migration that use column comments will not work (see linked comment above for an example).
Somehow it works when you have a clean install, then all migrations pass, but when executing single migrations they will fail.

This solves the problem by simply removing column comments when using SQLite.

## Checklist

### DIFF
--- a/lib/private/DB/SQLiteMigrator.php
+++ b/lib/private/DB/SQLiteMigrator.php
@@ -39,9 +39,13 @@ class SQLiteMigrator extends Migrator {
 		$platform->registerDoctrineTypeMapping('smallint unsigned', 'integer');
 		$platform->registerDoctrineTypeMapping('varchar ', 'string');
 
-		// with sqlite autoincrement columns is of type integer
 		foreach ($targetSchema->getTables() as $table) {
 			foreach ($table->getColumns() as $column) {
+				// column comments are not supported on SQLite
+				if ($column->getComment() !== null) {
+					$column->setComment(null);
+				}
+				// with sqlite autoincrement columns is of type integer
 				if ($column->getType() instanceof BigIntType && $column->getAutoincrement()) {
 					$column->setType(Type::getType('integer'));
 				}
--- a/tests/lib/DB/MigratorTest.php
+++ b/tests/lib/DB/MigratorTest.php
@@ -237,6 +237,30 @@ class MigratorTest extends \Test\TestCase {
 		$this->addToAssertionCount(1);
 	}
 
+	/**
+	 * Test for nextcloud/server#36803
+	 */
+	public function testColumnCommentsInUpdate() {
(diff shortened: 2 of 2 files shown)


kesselb:
Sounds weird. 

Comments are supported by doctrine and sqlite: https://github.com/doctrine/dbal/pull/819

susnux:
> Sounds weird.
> 
> Comments are supported by doctrine and sqlite: [doctrine/dbal#819](https://github.com/doctrine/dbal/pull/819)

The documentation does not list SQLite as supported:
https://www.doctrine-project.org/projects/doctrine-dbal/en/3.2/reference/schema-representation.html#common-options

Also SQL comments are not supported by SQLite (only inline comments within the schema).

How you can reproduce the error:
1. Setup a nextcloud server using SQLite (e.g. docker image)
2. Install and enable the forms app (version 3.1.0) 
3. Update forms to https://github.com/nextcloud/forms/commit/3181423ee0ed9cd378f630737db90dbb87463cdc
4. Try to run migrations and see error.

Or of cause any other app you like and try to run a migration containing
```php
$table->addColumn('foo', Types::INTEGER, [
	'notnull' => false,
	'default' => 0,
	'comment' => 'unix-timestamp',
]);
```

kesselb:
Hi,

I created a bug report for doctrine/dbal: https://github.com/doctrine/dbal/issues/5934.

> Also SQL comments are not supported by SQLite (only inline comments within the schema).

`'comment' => 'unix-timestamp',` that's an inline comment within the schema ;)

> Some times column comments are used, e.g. to make clear an integer is used as a timestamp.

I don't like the idea of having code in Nextcloud to remove comments on SQLite. 
My recommendation is to remove the comments from the migrations. 
A comment in the code is anyway a better place to document that a column should store a certain type of data.

kesselb (approved):
Seems valid, should work, still I don't like it ;)

jotoeri (approved):
Looks good to me. 👍 

I'd rather have the comment at least when using mysql,... (and e.g. phpmyadmin), than not having it at all. To have code-comments is independently of course a good thing.
- One could perhaps here add a comment to link to the dbal-issue. So once resolved one could re-check, if the comment-removal is still necessary.

nickvergessen:
Can we add a test to https://github.com/nextcloud/server/blob/master/tests/lib/DB/MigratorTest.php
So that it is ensured that SQLite keeps working?

susnux:
> Can we add a test so that it is ensured that SQLite keeps working?

@nickvergessen done :)

PR #36807, opened 2023-02-22
Avoid getting null mimetype when getting reference preview

julien-nc:
Fix getting null mimetype if the cached reference lacks an image content type.


nickvergessen (approved):

juliusknorr (approved):

julien-nc:
Drone failure is unrelated (postres service does not start).

PR #36810, opened 2023-02-22
do onetime user setup before getting any mount from providers

icewind1991:
this fixes a race condition where wrappers weren't properly applied if when a mount provider creates the storage instance one mountpoint creation instead of lazily

To test:

- Setup an instance with encryption enabled, groupfolders and text
- Enable groupfolder encryption: `occ config:app:set groupfolders enable_encryption --value true`
- Create a groupfolder and navigate into it
- Click the "Add notes, lists, or links..." in the webui to create a `Readme.md` in the folder
- Check if the text editor gets filled with an encrypted blob or an empty textfile

### DIFF
--- a/lib/private/Files/SetupManager.php
+++ b/lib/private/Files/SetupManager.php
@@ -235,10 +235,12 @@ class SetupManager {
 	 * part of the user setup that is run only once per user
 	 */
 	private function oneTimeUserSetup(IUser $user) {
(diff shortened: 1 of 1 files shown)


come-nc (approved):

skjnldsv (approved):

Pytal:
Backportable to `stable24` @icewind1991?

PR #36812, opened 2023-02-23
Set notification email automatically after deleting additional email

Pytal:
## Summary

When an additional email is set as the notification email and is deleted, the deleted email will remain as the notification email

This is unexpected behaviour so we fix this by automatically setting the system email as the notification email on deletion

### Additional email as notification email
image

### After deletion
| | Screenshot |
--- | ---
**Before** | image
**After** | image

## Checklist


artonge (approved):
Would it make sense to do the change on the backend side instead of here?

skjnldsv:
> Would it make sense to do the change on the backend side instead of here?

I agree with Louis @Pytal :)

blizzz (approved):

blizzz:
also would prefer it on the backend side

skjnldsv:
Closing this pull request due to lack of recent activity and updates; also seeing the two comments about backend :)

PR #36813, opened 2023-02-23
Bump @nextcloud/vue from 7.5.0 to 7.7.0 and add node-polyfill-webpack-plugin

Pytal:
## Summary

Bump `@nextcloud/vue` from https://github.com/nextcloud/nextcloud-vue/releases/tag/v7.5.0 to https://github.com/nextcloud/nextcloud-vue/releases/tag/v7.7.0

Add https://github.com/Richienb/node-polyfill-webpack-plugin to fix the `process is not defined` and resulting `OC is not defined` errors

## Checklist


szaimen (approved):
🐘

artonge:
Shouldn't the process thing be solved on the `@nextcloud/files` side ?

webpack.common.js
```diff
@@ -161,6 +164,9 @@ module.exports = {
 
 			// https://github.com/webpack/changelog-v5/issues/10
 			Buffer: ['buffer', 'Buffer'],
+
+			// Provide process module shim for dependencies which import it
+			process: 'process/browser',
```
juliushaertl, line 169: We could think about using https://www.npmjs.com/package/node-polyfill-webpack-plugin

juliusknorr (approved):

nickvergessen:
> FAIL apps/dav/src/views/CalDavSettings.spec.js

Pytal:
> Shouldn't the process thing be solved on the `@nextcloud/files` side ? I never really played with modules, so I have no idea.

`@nextcloud/files` (https://github.com/nextcloud/nextcloud-files/blob/94bfe0e1a093dcb8ed6b0710e7bdd8214ba16427/lib/files/node.ts#L22) seems to have an implicit dependency on running within a https://github.com/nextcloud/server runtime environment as it neither has a `path` dependency in [package.json](https://github.com/nextcloud/nextcloud-files/blob/94bfe0e1a093dcb8ed6b0710e7bdd8214ba16427/package.json) nor a polyfill for it in https://github.com/nextcloud/nextcloud-files/blob/94bfe0e1a093dcb8ed6b0710e7bdd8214ba16427/rollup.config.js while the server [package.json](https://github.com/nextcloud/server/blob/00a7478d8df2007697d5caf2e0cb1c298afbb3fd/package.json#L89) does

From the `process is not defined` error stack trace we see that the `path` import is resolved to https://github.com/jinder/path which is also used in https://github.com/nextcloud/server https://github.com/nextcloud/server/blob/00a7478d8df2007697d5caf2e0cb1c298afbb3fd/apps/files/src/views/TemplatePicker.vue#L67 https://github.com/nextcloud/server/blob/00a7478d8df2007697d5caf2e0cb1c298afbb3fd/apps/files_sharing/src/components/SelectShareFolderDialogue.vue#L43

For these reasons let's handle the dependencies and polyfills here for now, while explicit dependencies and polyfills may be added to https://github.com/nextcloud/nextcloud-files separately if needed

PR #25768, opened 2021-02-23
Use mimetype from cache for workflow checks

icewind1991:
- Use the mimetype out of the cache instead of re-calculating it
- Only detect mimetype from content for local storages
- Remove redundant `file_exists` check
- Add some tests

Motivation is that when using object stores, the file is added to the cache (and thus triggering workflow hooks) before it's written to the storage.

PR #24459, opened 2020-11-30
Only execute plain mimetype check for directories and do the fallback…

MorrisJobke:
… only for non-directories

Introduced with #23096

Fixes https://github.com/nextcloud/files_accesscontrol/issues/182

@juliushaertl I'm open for other approaches, but this seems to be useful here.

PR #23096, opened 2020-09-29
Also check the path based mimetype for flow rule checks

juliusknorr:
Before this PR it is not possible to match workflow rules against application/javascript or application/x-ms-dos-executable since they are not part of the mimetypemappings and the detectContents always uses the secure fallback, so checking for those file types for example always fails:

| file | detectContent | detectPath | 
|---|---|---|
| .exe | application/octet-stream | application/x-ms-dos-executable |
| .js | text/plain | application/javascript |

This PR makes sure that when checking for a workflow rule both the detected type on the content and path are checked.

PR #36830, opened 2023-02-23
fix: Avoid failing with duplicate checks

juliusknorr:
## Summary

When receiving a share the workflow mimetype check the patch was always checked but was an empty string (considered as application/octet-stream then).

There is no need for performing this check anymore here since it is covered within getActualValue these days.

Steps to reproduce:

- Block access with files_accesscontrol with the followin rules
	- If not mimetype httpd/directory
	- If not mimetype application/pdf
- Uplaod a pdf file and share it to user B

Before this the file was not visible to the recipient

Note this seems to occur only after https://github.com/nextcloud/server/pull/25768

Covered with integration tests by https://github.com/nextcloud/files_accesscontrol/pull/321

## Checklist

### DIFF
--- a/apps/workflowengine/lib/Check/FileMimeType.php
+++ b/apps/workflowengine/lib/Check/FileMimeType.php
@@ -107,13 +107,7 @@ class FileMimeType extends AbstractStringCheck implements IFileCheck {
 	 * @return bool
 	 */
 	public function executeCheck($operator, $value) {
-		$actualValue = $this->getActualValue();
-		$plainMimetypeResult = $this->executeStringCheck($operator, $value, $actualValue);
-		if ($actualValue === 'httpd/unix-directory') {
-			return $plainMimetypeResult;
-		}
-		$detectMimetypeBasedOnFilenameResult = $this->executeStringCheck($operator, $value, $this->mimeTypeDetector->detectPath($this->path));
-		return $plainMimetypeResult || $detectMimetypeBasedOnFilenameResult;
+		return $this->executeStringCheck($operator, $value, $this->getActualValue());
 	}
 
 	/**
--- a/apps/workflowengine/tests/Check/FileMimeTypeTest.php
+++ b/apps/workflowengine/tests/Check/FileMimeTypeTest.php
@@ -127,7 +127,7 @@ class FileMimeTypeTest extends TestCase {
 		$check = new FileMimeType($this->l10n, $this->request, $this->mimeDetector);
 		$check->setFileInfo($storage, 'foo/bar.txt');
 
-		$this->assertTrue($check->executeCheck('is', 'text/plain-path-detected'));
+		$this->assertTrue($check->executeCheck('is', 'text/plain-content-detected'));
 	}
 
 	public function testFallback() {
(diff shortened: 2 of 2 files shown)


juliushaertl:
Needs another careful look after discussion with @nickvergessen 

Different approach could be this patch:

```patch
diff --git a/apps/workflowengine/lib/Check/FileMimeType.php b/apps/workflowengine/lib/Check/FileMimeType.php
index 42b93bd9513..3269638e12d 100644
--- a/apps/workflowengine/lib/Check/FileMimeType.php
+++ b/apps/workflowengine/lib/Check/FileMimeType.php
@@ -106,12 +106,14 @@ class FileMimeType extends AbstractStringCheck implements IFileCheck {
         * @return bool
         */
        public function executeCheck($operator, $value) {
+               $path = $this->storage->getCache('')->get($this->path)->getName();
                $actualValue = $this->getActualValue();
                $plainMimetypeResult = $this->executeStringCheck($operator, $value, $actualValue);
                if ($actualValue === 'httpd/unix-directory') {
                        return $plainMimetypeResult;
                }
-               $detectMimetypeBasedOnFilenameResult = $this->executeStringCheck($operator, $value, $this->mimeTypeDetector->detectPath($this->path));
+               $detectPath = $this->mimeTypeDetector->detectPath($this->storage->getCache('')->get('')->getName());
+               $detectMimetypeBasedOnFilenameResult = $this->executeStringCheck($operator, $value, $detectPath);
                return $plainMimetypeResult || $detectMimetypeBasedOnFilenameResult;
```

juliushaertl:
Will check the failing test which happens because we run a test for a failure upload case with the new upload that should only fail on s3 on local primary storage. Will look into

apps/workflowengine/lib/Check/FileMimeType.php
```diff
@@ -107,13 +107,7 @@ protected function cacheAndReturnMimeType(string $storageId, ?string $path, stri
...
 	 */
 	public function executeCheck($operator, $value) {
-		$actualValue = $this->getActualValue();
-		$plainMimetypeResult = $this->executeStringCheck($operator, $value, $actualValue);
-		if ($actualValue === 'httpd/unix-directory') {
-			return $plainMimetypeResult;
-		}
-		$detectMimetypeBasedOnFilenameResult = $this->executeStringCheck($operator, $value, $this->mimeTypeDetector->detectPath($this->path));
-		return $plainMimetypeResult || $detectMimetypeBasedOnFilenameResult;
+		return $this->executeStringCheck($operator, $value, $this->getActualValue());
```
nickvergessen, line 110: This basically matches deleting the method, which should basically bring back the issue you yourself fixed in:
* https://github.com/nextcloud/server/pull/23096

The final result of the current code was done in a follow up:
* https://github.com/nextcloud/server/pull/24459

juliushaertl: Proven to work by https://github.com/nextcloud/files_accesscontrol/pull/321

nickvergessen (approved):

juliushaertl:
Another ping for review here

artonge (approved):

PR #36836, opened 2023-02-23
Tidy up typing in OC\Files\View

come-nc:
## Summary

Add return types and fix psalm complaints in View and related classes.

## Checklist

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -235,20 +235,6 @@ class View {
 	}
 
 	/**
-	 * @param string $path
-	 */
-	public function getLocalFolder($path): string|bool {
-		$parent = substr($path, 0, strrpos($path, '/') ?: 0);
-		$path = $this->getAbsolutePath($path);
-		[$storage, $internalPath] = Filesystem::resolvePath($path);
-		if (Filesystem::isValidPath($parent) and $storage) {
-			return $storage->getLocalFolder($internalPath);
-		} else {
-			return false;
-		}
-	}
-
-	/**
 	 * the following functions operate with arguments and return values identical
 	 * to those of their PHP built-in equivalents. Mostly they are merely wrappers
 	 * for \OC\Files\Storage\Storage via basicOperation().
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -1336,7 +1336,7 @@ class View {
 	 * @param string $path
 	 * @param bool|string $includeMountPoints true to add mountpoint sizes,
 	 * 'ext' to add only ext storage mount point sizes. Defaults to true.
-	 * @return \OCP\Files\FileInfo|false False if file does not exist
+	 * @return \OC\Files\FileInfo|false False if file does not exist
 	 */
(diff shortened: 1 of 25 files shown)


come-nc:
Needs https://github.com/icewind1991/Streams/pull/17

come-nc:
@icewind1991 https://github.com/nextcloud/server/blob/master/apps/dav/lib/Connector/Sabre/Directory.php#L139-L141 this is blowing up because it puts null in non-nullable parameters.
Any idea how to refactor this properly?

come-nc:
> @icewind1991 https://github.com/nextcloud/server/blob/master/apps/dav/lib/Connector/Sabre/Directory.php#L139-L141 this is blowing up because it puts null in non-nullable parameters. Any idea how to refactor this properly?

For now I went back to weak typing for internalPath in FileInfo.

ChristophWurst (commented):
:+1: looks ok but I'm not familiar enough with this area of the code to given an approval

lib/private/Files/View.php
```diff
@@ -192,74 +181,56 @@ public function getRelativePath($path) {
...
 		} else {
-			return null;
-		}
-	}
-
-	/**
-	 * @param string $path
-	 * @return string
-	 */
-	public function getLocalFolder($path) {
```
icewind1991, line 255: is deleting this intentional?

icewind1991 (approved):
Good work otherwise :+1:

icewind1991:
> @icewind1991 https://github.com/nextcloud/server/blob/master/apps/dav/lib/Connector/Sabre/Directory.php#L139-L141 this is blowing up because it puts null in non-nullable parameters. Any idea how to refactor this properly?

probably makes most sense to change those `null`s to empty strings. But it probably makes more sense to do a that in a separate PR

lib/private/Files/View.php
come-nc: Yes, the only place it was used was a test.

lib/private/Files/View.php
```diff
@@ -1163,9 +1124,9 @@ private function basicOperation($operation, $path, $hooks = [], $extraParam = nu
 			}
 
 			$run = $this->runHooks($hooks, $path);
-			/** @var \OC\Files\Storage\Storage $storage */
 			[$storage, $internalPath] = Filesystem::resolvePath($absolutePath . $postFix);
 			if ($run and $storage) {
+				/** @var \OC\Files\Storage\Storage $storage */
```
blizzz, line 1129: Namespace not necessary, use statement already present

lib/public/Files/FileInfo.php
```diff
@@ -250,7 +250,7 @@ public function getMountPoint();
 	/**
 	 * Get the owner of the file
 	 *
-	 * @return \OCP\IUser
+	 * @return ?\OCP\IUser
```
blizzz, line 253: should be added to critical changes (even if not a de-facto change to raise awareness)

blizzz (approved):

PR #36845, opened 2023-02-24
Fix max instances exception in get vevent call

akhil1508:
- [This commit](https://github.com/nextcloud/server/commit/a9264b007ffa4bf4ae04069bb14f56b6b3f591ea) handles the `MaxInstancesExceededException` in most of the cases but misses one
- Fixing that in this one case as exception is causing reminders to not be fired at all in one of our servers :(

### DIFF
--- a/apps/dav/lib/CalDAV/Reminder/ReminderService.php
+++ b/apps/dav/lib/CalDAV/Reminder/ReminderService.php
@@ -149,8 +149,7 @@ class ReminderService {
 
 			try {
 				$vevent = $this->getVEventByRecurrenceId($vcalendar, $reminder['recurrence_id'], $reminder['is_recurrence_exception']);
-			}
-			catch (MaxInstancesExceededException $e) {
+			} catch (MaxInstancesExceededException $e) {
 				$this->logger->debug('Recurrence with too many instances detected, skipping VEVENT', ['exception' => $e]);
 				continue;
 			}
--- a/apps/dav/lib/CalDAV/Reminder/ReminderService.php
+++ b/apps/dav/lib/CalDAV/Reminder/ReminderService.php
@@ -151,6 +151,7 @@ class ReminderService {
 				$vevent = $this->getVEventByRecurrenceId($vcalendar, $reminder['recurrence_id'], $reminder['is_recurrence_exception']);
 			} catch (MaxInstancesExceededException $e) {
 				$this->logger->debug('Recurrence with too many instances detected, skipping VEVENT', ['exception' => $e]);
+				$this->backend->removeReminder($reminder['id']);
 				continue;
 			}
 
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

miaulalala (approved):

tcitworld (approved):
Thanks!

(btw, your commit is marked as unverified on Github, you need to upload your key)

akhil1508:
@tcitworld Fixed the sign-off. Thanks

apps/dav/lib/CalDAV/Reminder/ReminderService.php
```diff
@@ -147,7 +147,14 @@ public function processReminders() :void {
 				continue;
 			}
 
-			$vevent = $this->getVEventByRecurrenceId($vcalendar, $reminder['recurrence_id'], $reminder['is_recurrence_exception']);
+			try {
+				$vevent = $this->getVEventByRecurrenceId($vcalendar, $reminder['recurrence_id'], $reminder['is_recurrence_exception']);
+			}
+			catch (MaxInstancesExceededException $e) {
+				$this->logger->debug('Recurrence with too many instances detected, skipping VEVENT', ['exception' => $e]);
```
akhil1508, line 154: @ChristophWurst @miaulalala @tcitworld 
Does it make sense to add a `$this->backend->removeReminder($reminder['id']);` over here in the catch before the  `continue`?

tcitworld: I'd say the current changes are preferable. We should avoid inserting them in the first place otherwise.

akhil1508: Got it. But also we do remove the reminder in all the other conditions(like if it does not belong to a calendar or to an event e.g. [here](https://github.com/nextcloud/server/blob/9f9da964f678aed3276fb8f71c553253423a819c/apps/dav/lib/CalDAV/Reminder/ReminderService.php#L146)).

apps/dav/lib/CalDAV/Reminder/ReminderService.php
```diff
@@ -147,7 +147,14 @@ public function processReminders() :void {
 				continue;
 			}
 
-			$vevent = $this->getVEventByRecurrenceId($vcalendar, $reminder['recurrence_id'], $reminder['is_recurrence_exception']);
+			try {
+				$vevent = $this->getVEventByRecurrenceId($vcalendar, $reminder['recurrence_id'], $reminder['is_recurrence_exception']);
+			}
+			catch (MaxInstancesExceededException $e) {
```
ChristophWurst, line 153:
```suggestion
			} catch (MaxInstancesExceededException $e) {
```

formatting

apps/dav/lib/CalDAV/Reminder/ReminderService.php
tcitworld: That's a good point. It makes sense.

akhil1508: Thanks. Added it @tcitworld .

PR #36852, opened 2023-02-24
fix(user_status): Fix the user status automation on the day availablity rules are adjusted

nickvergessen:
- [x] Needs to be rebased after #37167

## Summary

I was trying to work towards #29339 but the availability I configured did not kick in and set myself to DND

So after lengthy debugging it turns out we need to do some shenanigans to fix the automation on the day the rules were adjusted
Otherwise the $originalStart would match rules for Thursdays on a Friday, etc. So we simply wind back a week and then fastForward to the next occurrence since today's midnight, which then also accounts for the week days.

## Checklist


miaulalala (approved):
LGTM

tcitworld (approved):

PR #36854, opened 2023-02-24
Allow to revert automated status

### LINKED ISSUE #29339: Missing action to revert the status back after an auto change in calendar/talk
> - In the Status modal, between the "What’s your status" input and the 5 suggestions, we put a block like this (with the second line being a text link / button:
> > Automatically set during calls and events.
> > Reset to **📅 Vacation next week**

Still missing, but not related to the Talk change

_Originally posted by @nickvergessen in https://github.com/nextcloud/spreed/issues/6110#issuecomment-946424372_

nickvergessen:
* Resolves: #29339

## Summary

DND because of Availability | Away because of "In call"
---|---
Peek 2023-02-27 21-50 | Peek 2023-02-27 21-52

## TODO

- [x] Allow to manually revert the status when it was set by "Availability" or "Joining a call"

## Follow up
- [ ] Currently requires a page reload before, as the "messageId" is only populated via initial state and not transmitted by any other API calls and therefore not updated "live" (e.g. when joining a call the reset button is not visible in the same window) => https://github.com/nextcloud/server/issues/37733

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required => https://github.com/nextcloud/documentation/pull/10042


nimishavijay:
Good idea 👍🏽 For the UI we could do it like this:
- Show the current status and emoji in status input field (because that is where people would look first for the current status)
- Show "Your status was set automatically" as a subline below the input field (info when people are confused that their status has changed)
- Show previous status as the first entry in the list of predefined statuses (visually follows the pattern of the list of statuses while still being easy to find because it's the first entry)
- Show a reset button on the right of the previous status (doesn't take up too much vertical space) 
- When you click on "reset" button it will be reflected in the contents of the status input field and the emoji

Rough idea of the layout:


cc @jancborchardt

nickvergessen:
> When you click on "reset" button it will be reflected in the contents of the status input field and the emoji

Is it okay to directly "save" this as well? That would be easier as we can restore the old online-status, emoji, message and clear-at with a single request instead of having to do multiple requests later on.

Current state sample gifs

DND because of Availability | Away because of "In call"
---|---
Peek 2023-02-27 21-50 | Peek 2023-02-27 21-52

jancborchardt (changes requested):
Nice! To make it a bit less jumpy as @nimishavijay noted, it would be good to add a little animation here. Best of the row you click "Reset" on moving up and fading away, sort of so it fades into the input field. :)

nickvergessen:
> Best of the row you click "Reset" on moving up and fading away, sort of so it fades into the input field. :)

I'm not a frontender so can't help with that. The other option I would see is closing the modal which also happens when "Set status message" is used.

jancborchardt (approved):
Nice as is now, the animation could be something for the future. Nice one @nickvergessen! :)

Antreesy (approved):

PR #36857, opened 2023-02-24
fix unencrypted_size for folders when scanning the filesystem with encryption enabled

icewind1991:
To test:

- setup an instance with encryption
- create folder `test` and write some files into it
- "accidentally" mess up the `unencrypted_size` for the folder with
  
    `update oc_filecache set unencrypted_size = 100 where name = 'test';`
  
- run `occ files:scan --all` in an attempt to repair it
- check if the folder size is reported correctly

### DIFF
--- a/lib/private/Files/Cache/Scanner.php
+++ b/lib/private/Files/Cache/Scanner.php
@@ -208,9 +208,17 @@ class Scanner extends BasicEmitter implements IScanner {
 								$data['etag'] = $etag;
 							}
 						}
+
+						// we only updated unencrypted_size if it's already set
+						if ($cacheData['unencrypted_size'] === 0) {
+							unset($data['unencrypted_size']);
+						}
+
 						// Only update metadata that has changed
 						$newData = array_diff_assoc($data, $cacheData->getData());
 					} else {
+						// we only updated unencrypted_size if it's already set
+						unset($data['unencrypted_size']);
 						$newData = $data;
 						$fileId = -1;
 					}
--- a/lib/private/Files/Cache/Scanner.php
+++ b/lib/private/Files/Cache/Scanner.php
@@ -36,6 +36,7 @@
 namespace OC\Files\Cache;
 
 use Doctrine\DBAL\Exception;
+use OC\Files\Storage\Wrapper\Encryption;
 use OCP\Files\Cache\IScanner;
 use OCP\Files\ForbiddenException;
 use OCP\Files\Storage\IReliableEtagStorage;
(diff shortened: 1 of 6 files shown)


lib/private/Files/Cache/Scanner.php
```diff
@@ -390,8 +391,15 @@ protected function scanChildren($path, $recursive = self::SCAN_RECURSIVE, $reuse
 			}
 		}
 		$oldSize = $data['size'] ?? null;
-		if ($this->cacheActive && $oldSize !== $size) {
-			$this->cache->update($folderId, ['size' => $size]);
+
+		// for encrypted storages, we trigger a regular folder size calculation instead of using the calculated size
+		// to make sure we also updated the unencrypted-size where applicable
```
come-nc, line 396: What would be the downside of doing so for all storages?
Also, since `$this->cache->calculateFolderSize($path);` does not use $size is the above code needed?

icewind1991: The old method is more performant since it doesn't require querying all the child sizes

come-nc (approved):

PVince81:
is this related / should this be linked with https://github.com/nextcloud/server/issues/25283 which has the whole discussion about unencrypted_size ? also there might be people there who could test this PR ?

lib/private/Files/Cache/Scanner.php
```diff
@@ -207,9 +208,17 @@ public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData =
 								$data['etag'] = $etag;
 							}
 						}
+
+						// we only updated unencrypted_size if it's already set
+						if ($cacheData['unencrypted_size'] === 0) {
```
artonge, line 213: Shouldn't we use the `encrypted` column for those kinds of conditions? What if `unencrypted_size` is set to 0?

icewind1991: `unencrypted_size` isn't set in all cases for encrypted files.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -494,7 +496,8 @@ protected function verifyUnencryptedSize(string $path, int $unencryptedSize): in
 		$result = $unencryptedSize;
 
 		if ($unencryptedSize < 0 ||
-			($size > 0 && $unencryptedSize === $size)
+			($size > 0 && $unencryptedSize === $size) ||
+			$unencryptedSize > $size
```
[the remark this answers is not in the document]
artonge: Yes, sorry, I mixed both :+1:

lib/private/Files/Cache/Scanner.php
artonge: Then wouldn't `$cacheData['encrypted'] && !isset($cacheData['unencrypted_size'])` be better?
I am just wondering what would append if unencrypted_size is set to 0 on purpose?

icewind1991: my current approach is to be conservative for when to set the `unencrypted_size` since it isn't needed in most cases. (the main use case is object store which set's it itself)

Maybe at a later date we can decide to actively populate the field but for now this PR focuses on repairing invalid values.

artonge (approved):

PR #36883, opened 2023-02-27
dispatch BeforeUserLoggedInEvent

ArtificialOwl:
legacy code ...

This fix an issue with gss+saml (regreation post migration to `IEventDispatcher` as login event is not detected anymore). 
The idea is to copy the normal process available in https://github.com/nextcloud/server/blob/master/lib/private/Server.php#L615-L621 and implement it on login process on SSO

### DIFF
--- a/lib/private/legacy/OC_User.php
+++ b/lib/private/legacy/OC_User.php
@@ -40,6 +40,7 @@ use OC\User\LoginException;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\ILogger;
 use OCP\IUserManager;
+use OCP\User\Events\BeforeUserLoggedInEvent;
 use OCP\User\Events\UserLoggedInEvent;
 
 /**
@@ -172,6 +173,10 @@ class OC_User {
 			if (self::getUser() !== $uid) {
 				self::setUserId($uid);
 				$userSession = \OC::$server->getUserSession();
+
+				/** @var IEventDispatcher $dispatcher */
+				$dispatcher = \OC::$server->get(IEventDispatcher::class);
+
 				if ($userSession->getUser() && !$userSession->getUser()->isEnabled()) {
 					$message = \OC::$server->getL10N('lib')->t('User disabled');
 					throw new LoginException($message);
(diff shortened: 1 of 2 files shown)


Pytal (approved):

come-nc:
Why is the IApacheBackend needed?
Is it really specific to Apache web server?
Can you link to the code using that?

blizzz:
It used to be that legacy hooks were listened to in the server, and modern style ones fired in those cases. Here it would be a listener against  `OC_Hook::emit("OC_User", "pre_login", ["run" => &$run, "uid" => $uid, 'backend' => $backend]);'. Is it not the case (anymore)?

ArtificialOwl:
example of `IApacheBackend`: https://github.com/nextcloud/user_saml/blob/master/lib/UserBackend.php
No idea if there is still in use elsewhere.

Currently, the modern style hook were not triggered during the sso auth because it was missing.
The `globalsiteselector` from master needs to interrupt the process of sso and generate session on the `slave`, not on the `master`. Because of the missing event, the session was created on master

come-nc:
> It used to be that legacy hooks were listened to in the server, and modern style ones fired in those cases. Here it would be a listener against `OC_Hook::emit("OC_User", "pre_login", ["run" => &$run, "uid" => $uid, 'backend' => $backend]);'. Is it not the case (anymore)?

It’s weird actually there is https://github.com/nextcloud/server/blob/master/lib/private/Server.php#L615-L621

But it’s the wrong way around, it listens to `\OC\User::preLogin` and emits both `OC_User::pre_login` and `BeforeUserLoggedInEvent`.
And it does not have the backend parameter.

Not sure what the clean fix is here, events are a mess.

ArtificialOwl:
> It’s weird actually there is https://github.com/nextcloud/server/blob/master/lib/private/Server.php#L615-L621

Yes, my first thought was to replace the current `OC_Hook::emit("OC_User", "pre_login", [...` to use the code your linked, but I think it was safer to add the call to `IEventDispatcher` directly in the code as it was already available later in the function.

This event will be catched by https://github.com/nextcloud/globalsiteselector/pull/89

come-nc (approved):

PR #36928, opened 2023-02-28
feat(middleware): Migrate BruteForceProtection annotation to PHP Attribute and allow multiple

nickvergessen:
### Previously (and still supported)
```php
	/**
	 * @BruteForceProtection(action=login)
	 */
	public function testMethodWithAnnotation() {
```

### New syntax
```php
	#[BruteForceProtection(action: 'login')]
	public function testMethodWithAttribute() {
```

### Multiple attributes now supported
By specifying an `'action'` in the `throttle($metadata)` only the dedicated attribute will trigger it's BFP, if none is specified all will throttle.
This is useful e.g. in controllers that need to protect different data, e.g. Talk has the tokens and the signaling secret in the same method and needs to do a lot of manual work instead of just calling `throttle` on the response:
https://github.com/nextcloud/spreed/blob/29a442cf86d49b426ecd47776c86ab6f6de5da58/lib/Controller/SignalingController.php#L154-L179

```php
	#[BruteForceProtection(action: 'token')]
	#[BruteForceProtection(action: 'password')]
	public function testMethodWithMultipleAttributes() {
```

## 🚧 ToDo

- [x] Allow as attribute
- [x] Add multi support
- [x] Log debug when the request was throttled, but there was no Annotation nor matching Attribute
- [ ] Add a header to the response when the request is throttled(?)

## Checklist

- [ ] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required https://github.com/nextcloud/documentation/pull/9742

### DIFF
--- a/lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 /**
+ * @copyright Copyright (c) 2023 Joas Schilling <coding@schilljs.com>
  * @copyright Copyright (c) 2017 Lukas Reschke <lukas@statuscode.ch>
  *
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
@@ -31,6 +32,7 @@ use OC\AppFramework\Utility\ControllerMethodReflector;
 use OC\Security\Bruteforce\Throttler;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\BruteForceProtection;
 use OCP\AppFramework\Http\Response;
 use OCP\AppFramework\Http\TooManyRequestsResponse;
 use OCP\AppFramework\Middleware;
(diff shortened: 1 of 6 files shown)


lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
```diff
@@ -68,18 +68,55 @@ public function beforeController($controller, $methodName) {
 		if ($this->reflector->hasAnnotation('BruteForceProtection')) {
 			$action = $this->reflector->getAnnotationParameter('BruteForceProtection', 'action');
 			$this->throttler->sleepDelayOrThrowOnMax($this->request->getRemoteAddress(), $action);
+		} else {
```
come-nc, line 71: Would it make sense to add `hasAttribute` / `getAttributes` to `ControllerMethodReflector` ?

come-nc (approved):

lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
nickvergessen: THought about that too. I'm still up for it as a follow up, to combine the logic our for UseSession and this one.
Just thought it's easier to first get this in and then do more rework.

come-nc: I’m always afraid the follow-up never comes in these cases :-)

We should try to make the new way as easy to use as the previous one if we want the code to get migrated at some point.

nickvergessen: > I’m always afraid the follow-up never comes in these cases

I guess you have a point, at the same time there is the RateLimit annotation waiting for me as a task this and next month, so will come.

vitormattos (approved):

PR #36931, opened 2023-03-01
fix(theming): select "no-background" option and fix layout order

### LINKED ISSUE #35479: Theming refactoring follow-ups
- should the `Change color` option really be part of the background grid? In my opinion it should be rather outisde as it is not a background option...
-  `Remove background` should be shown as selected if it is selected imho and it should have the background color that is currently active, imho.
        image

skjnldsv:
Fix https://github.com/nextcloud/server/issues/35479

|No background|Custom background|
|:--:|:--:|
|||
|Shipped background|Default background|
|||

- Renamed "Remove background" to "No background" to better reflect that this is an option and not just an action
- Make the "Remove background" reflect the user choice
- Moved the "Remove background" option in the first row to better separate the shipped backgrounds
- Added an icon on the "Custom background" option


apps/theming/src/components/BackgroundSettings.vue
```diff
@@ -60,6 +61,17 @@
...
 
+		<!-- Remove background -->
+		<button class="background background__delete"
+			:class="{ 'background--active': isBackgroundDisabled }"
+			data-user-theming-background-clear
+			tabindex="0"
+			@click="removeBackground">
+			{{ t('theming', 'No background') }}
+			<Close v-if="!isBackgroundDisabled" :size="32" />
+			<Check :size="44" />
```
artonge, line 72: Any reason you used css instead `v-else` ?

skjnldsv: I don't recall :shrug: 
I think it was like that before, and I left it when I switched from `icon-confirm` to the Check vue component
This is not related tot his PR anyway :)

artonge (approved):

szaimen (approved):
Works in my testing 🚀

szaimen:
But conflicts and jsunit fails...

Pytal (approved):

skjnldsv:
> and jsunit fails

I don't think it's related

PR #36942, opened 2023-03-01
unmark deleted users if their last login date is more recent than deleted date

icewind1991:
If the user didn't get unmarked on login for some reason we can at least unmark them when listing the deleted users.

### DIFF
--- a/apps/user_ldap/lib/User/DeletedUsersIndex.php
+++ b/apps/user_ldap/lib/User/DeletedUsersIndex.php
@@ -56,7 +56,12 @@ class DeletedUsersIndex {
 
 		$userObjects = [];
 		foreach ($deletedUsers as $user) {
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

apps/user_ldap/lib/User/DeletedUsersIndex.php
```diff
@@ -56,7 +56,12 @@ private function fetchDeletedUsers(): array {
 
 		$userObjects = [];
 		foreach ($deletedUsers as $user) {
-			$userObjects[] = new OfflineUser($user, $this->config, $this->mapping, $this->shareManager);
+			$userObject = new OfflineUser($user, $this->config, $this->mapping, $this->shareManager);;
```
come-nc, line 59:
```suggestion
			$userObject = new OfflineUser($user, $this->config, $this->mapping, $this->shareManager);
```

come-nc (approved):

PR #37006, opened 2023-03-02
also unmark deleted ldap user when checking the ldap entry

icewind1991:


blizzz:
Reasoning: the early return prevents the unmark logic in line 348-350.

blizzz (approved):

come-nc (approved):

PR #37070, opened 2023-03-07
add a check for disk_free_space in Config.php

szaimen:
Follow-up to https://github.com/nextcloud/server/pull/36691

This is best reviewed like this: https://github.com/nextcloud/server/pull/37070/files?diff=unified&w=1

### DIFF
--- a/lib/private/Config.php
+++ b/lib/private/Config.php
@@ -286,10 +286,12 @@ class Config {
 		}
 
 		// Never write file back if disk space should be too low
-		$df = disk_free_space($this->configDir);
-		$size = strlen($content) + 10240;
-		if ($df !== false && $df < (float)$size) {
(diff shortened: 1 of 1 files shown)


artonge (approved):
When does the `disk_free_space` function not exist ?

szaimen:
> When does the `disk_free_space` function not exist ?

If it is whyever disabled by purpose by the admin...

come-nc (approved):

come-nc:
> > When does the `disk_free_space` function not exist ?
> 
> If it is whyever disabled by the admin...

You had an actual usecase or is that extra caution?
Because any function can be disabled in php.ini and we cannot test all functions we use.

szaimen:
> You had an actual usecase or is that extra caution?
> Because any function can be disabled in php.ini and we cannot test all functions we use.

https://github.com/nextcloud/server/pull/37062#discussion_r1127475570

PR #37146, opened 2023-03-09
Migrate metadata JSON column to new value TEXT column 

artonge:
`oc_files_metadata.metadata` is created as `Types::JSON`. Beside not bringing any advantages, this creates the limitation of not being able to use the `DISTINCT` operator on it when using PostgreSQL.

The `DISTINCT` operator is for example used by photos: https://github.com/nextcloud/photos/blob/65d222db10ab5f005125d56576b8bd223b4dd054/lib/DB/Place/PlaceMapper.php#L56

This PR migrates the `metadata` column to a new `value` column as `Types:STRING`
It also renames and adapt the code accordingly.

Photos PR to adjust: https://github.com/nextcloud/photos/pull/1692

Tested both on fresh install and after a migration.

### DIFF
--- a/core/Migrations/Version24000Date20220404230027.php
+++ b/core/Migrations/Version24000Date20220404230027.php
@@ -52,11 +52,15 @@ class Version24000Date20220404230027 extends SimpleMigrationStep {
 				'notnull' => true,
 				'length' => 50,
 			]);
-			$table->addColumn('metadata', Types::JSON, [
-				'notnull' => true,
+			$table->addColumn('value', Types::TEXT, [
+				'notnull' => false,
+				'default' => '',
 			]);
 			$table->setPrimaryKey(['id', 'group_name'], 'file_metadata_idx');
+
+			return $schema;
 		}
-		return $schema;
+
+		return null;
 	}
 }
--- a/core/Migrations/Version27000Date20230309104325.php
+++ b/core/Migrations/Version27000Date20230309104325.php
@@ -1,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Louis Chmn <louis@chmn.me>
+ *
+ * @author Louis Chmn <louis@chmn.me>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\Core\Migrations;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\IDBConnection;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Migrate oc_file_metadata.metadata as JSON type to oc_file_metadata.value a STRING type
+ * @see \OC\Metadata\FileMetadata
+ */
+class Version27000Date20230309104325 extends SimpleMigrationStep {
+	public function __construct(
+		private IDBConnection $connection
+	) {
+	}
+
+	/**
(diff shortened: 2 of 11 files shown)


core/Migrations/Version24000Date20220404230027.php
```diff
@@ -52,11 +52,14 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 				'notnull' => true,
 				'length' => 50,
 			]);
-			$table->addColumn('metadata', Types::JSON, [
+			$table->addColumn('value', Types::STRING, [
```
artonge, line 55: Ok to simply change type here, instead of commenting the migration and creating new one ?

ArtificialOwl:
why Types::STRING and not Types::TEXT ?

blizzz:
@nickvergessen i recall, `DISTINCT` should be avoided for performance reasons?

artonge:
> why `Types::STRING` and not `Types::TEXT` ?

No reason, why use one over the other ?

nickvergessen:
String is limited to 4000 chars IIRC

blizzz:
`String` translates to varchar, while `TEXT` translates up to `Longtext` (Mysql), respectively `TEXT` or `CLOB` on other DBs. `Json` translates to `Json` native types, apart of Sqlite and Oracle where it `CLOB`.  So, going to `TEXT` would be the sorta 1:1 option, otherwise it may fail loosing data (MySQL?) or erroring out on migration.

Still, the question on distinct?

core/Migrations/Version27000Date20230309104325.php
```diff
@@ -0,0 +1,125 @@
...
+		$results = $selectQuery
+			->setFirstResult($offset)
+			->executeQuery();
+
+		while ($row = $results->fetch()) {
+			$updateQuery
+				->setParameter('id', (int)$row['id'])
+				->setParameter('group_name', $row['group_name'])
+				->setParameter('value', $row['metadata'])
+				->executeStatement();
```
nickvergessen, line 117: A single direct update with `set('value', 'metadata')` does not work?

artonge: You mean not using parameters or actually setting 'value' by using the value of the 'metadata' column ? Not sure that the later is possible.

nickvergessen: Yeah, without a parameter just `UPDATE table SET newcolumn = oldcolumn;` basically.
That would allow you to get rid of the loop over all the entries in the database.

artonge: Seems like the only way to do it is

```php
		$this->connection->prepare('UPDATE *PREFIX*file_metadata SET value = metadata;')->execute();
```

Is that ok?

nickvergessen: The query builder can do it the same way, no?

artonge: Didn't manage to have it do the same

nickvergessen:
```php

		$updateQuery = $this->connection->getQueryBuilder();
		$updateQuery->update('file_metadata')
			->set('value', 'metadata');
```

Should work just fine?
https://github.com/nextcloud/spreed/blob/c07251f0c67d3c02debe1f6dbe0e12ee3fe9f99a/lib/Migration/Version2001Date20180103144447.php#L130-L137

artonge: Ok, now I feel stupid because I was convinced that it would result in a column full of the string 'metadata'.
Thanks for the handholding…

artonge:
Squashed, any other comments ?

nickvergessen (approved):
LGTM

szaimen:
Autoloaders seem to not be up-to-date?

artonge:
> Autoloaders seem to not be up-to-date?

Indeed, fixed :)

lib/private/Metadata/FileMetadata.php
```diff
@@ -28,16 +28,24 @@
...
 		$this->addType('groupName', 'string');
-		$this->addType('metadata', Types::JSON);
+		$this->addType('value', Types::STRING);
+	}
+
+	public function getDecodedValue(): array {
+		return json_decode($this->getValue(), true) ?? [];
+	}
+
+	public function setObjectAsValue(array $value): void {
```
datenangebot, line 48: For me the function name sounds a bit misleading, expecting a object not an array. But not that important.

datenangebot (approved):
Looks good, did not test it.

lib/private/Metadata/FileMetadata.php
artonge: True, fixed :)

Andreaux:
> Gnaa. This breaks recognize: [nextcloud/recognize#806](https://github.com/nextcloud/recognize/issues/806)

Hahh. I'm looking into fixing recognize (which fails  and I landed here. Has anyone got a suggestion?

PR #32798, opened 2022-06-09
[S3] Add option to specify an SSE-C customer provided key

juliusknorr:
- [x] Document sse_c_key to provide the 256-bit encryption key for Amazon S3 to use to encrypt or decrypt your data.
  - Requires a base64 encoded value e.g. generated with `openssl rand 32 | base64`
  - `'sse_c_key' => 'o9d3Q9tHcPMv6TIpH53MSXaUmY91YheZRwuIhwCFRSs=',`
  - https://github.com/nextcloud/documentation/pull/9414
- [x] Needs more testing, especially large file upload, copy

https://docs.aws.amazon.com/AmazonS3/latest/userguide/ServerSideEncryptionCustomerKeys.html#specifying-s3-c-encryption

PR #27034, opened 2021-05-19
Use MultipartUpload for uploading chunks to s3

juliusknorr:
Implements https://github.com/nextcloud/server/issues/19414

This allows chunked upload streaming the chunks directly to S3 so that the final MOVE request no longer needs to assemble the final file on the Nextcloud server.

Ref: https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html

## Current API changes to trigger the new upload mechanism (required changes on the client side)

The DAV capabilities expose a feature flag `s3-multipart` if available.

Additional requirements on top of the current chunking API in order to trigger it:

- New additional header is required on all chunked upload requests (MKCOL/PUT/MOVE)
  - `Destination: http://nextcloud.dev.local/remote.php/dav/files/admin/myfiletarget`
- Limitations
  - Chunk filename must be numberic between 1 and 10000
  - Chunks must be at least 5MB of size (except for the last one)
  - Chunks cannot be downloaded from the upload directory

## Summary of how S3 handles MultipartUpload

- CreateMultipartUpload is called with the destination bucket, object key and will return an uploadId
- UploadPart gets called for each part with the bucket, objectKey, uploadId, contentLength and the content, it returns a set of an ETag and the UploadId which need to be stored for the CompleteMultipartUpload 
- CompleteMultipartUpload Receives the list of parts and merges them directly on S3 to the final object

## How it currently works

- MKCOL Create the UploadFolder, initiate the MultipartUpload through the IChunkedFileWrite storage and persist the returned chunkToken (uploadId) in the dav properties of the folder
  - If the destination file does not exist we create a .target file in the upload folder to have a filecache entry
  - The destination is required upfront to allow writing directly to the target file
- PUT Writes the stream data directly to the S3 object through the IChunkedFileWrite storage
  - Validates the DAV chunk filename name to be numeric between 1 and 10000
- MOVE the MultipartUpload will be completed 
  - if the file exists we are done 
  - if the file didn't exist, call rename to move the .target file from the upload folder to the destination
- If a storage doesn't exist the new API calls will just fallback to the previous chunking mechanism

## Test script

<details>
 <summary>Bash script to test the upload</summary>

```
#!/bin/bash
# Usage: ./upload.sh targetname1
# Uploads a file consisting of 3 chunks and downloads it a again to verify the upload

if [ -z "$1" ]
  then
    echo "Usage: ./s3-chunk filepath"
	exit 1
fi

CHUNKSIZE=10
CHUNKS=2
[...]

PR #37150, opened 2023-03-09
chore: Address review comments

juliusknorr:
Addressing review follow ups from https://github.com/nextcloud/server/pull/27034

- [x] review from blizzz
- [x] review from come-nc
- [x] add SSE-C headers here to support #32798 with multipart upload


blizzz:

PR #37152, opened 2023-03-09
Use native tooltip for version's date

artonge:
And properly declare the method as computed property
And use a proper format


JuliaKirschenheuter (approved):

Pytal (approved):

artonge:
CI failure unrelated

PR #37170, opened 2023-03-10
fix(profile): Set profile config cache directly after writing it to DB

tcitworld:
In `ProfileManager::getProfileParams` we make a lot of indirect calls to `ProfileManager::getProfileConfig`. The first time we get the call we get into the catch and insert data, and the next time we miss the cache and do the select. There's a possibility of read-after-write issue here (the database r/o-mirror doesn't have the inserted config yet), which leads to conflicts when it tries to reinsert the config. To avoid that, let's put the config value in the configcache directly when it's written.

Fixes:
```
OC\DB\Exceptions\DbalException

An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERREUR:  la valeur d'une clé dupliquée rompt la contrainte unique « profile_config_user_id_idx »
DETAIL:  La clé « (user_id)=(Directrice) » existe déjà.
```

## Checklist

### DIFF
--- a/lib/private/Profile/ProfileManager.php
+++ b/lib/private/Profile/ProfileManager.php
@@ -347,6 +347,7 @@ class ProfileManager implements IProfileManager {
 				$this->filterNotStoredProfileConfig($config->getConfigArray()),
 			));
 			$this->configMapper->update($config);
+			$this->configCache[$targetUser->getUID()] = $config;
 			$configArray = $config->getConfigArray();
 		} catch (DoesNotExistException $e) {
 			// Create a new default config if it does not exist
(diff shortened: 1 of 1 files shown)


tcitworld:
This makes `AccountMigrator` tests unhappy for some reason. If you know why @Pytal

lib/private/Profile/ProfileManager.php
```diff
@@ -385,13 +385,15 @@ public function getProfileConfig(IUser $targetUser, ?IUser $visitingUser): array
 				$this->filterNotStoredProfileConfig($config->getConfigArray()),
 			));
 			$this->configMapper->update($config);
```
tcitworld, line 387: **suggestion:** BTW, do we really need to update the config at each call? Can't we just check if the merged config has been updated and only update it in this case?

tcitworld:
I think this is fixed already

PR #37178, opened 2023-03-10
feat: only load available commands in maintenance mode

kesselb:
* Resolves: # 

## Summary

```
Nextcloud is in maintenance mode, hence the database isn't accessible.
Cannot perform any command except 'maintenance:mode --off'
```

In maintenance mode, the above message is printed for every command. 
A customer brought to our attention that the message is misleading.

"isn't accessible" is a simplified statement. 
Technically, the database is accessible, but it's probably unsafe to run commands that modify the system. 

https://github.com/nextcloud/server/blob/bb4b34ff69350bb65672146406cb6c57c65326de/lib/base.php#L1032-L1053

https://github.com/nextcloud/server/blob/426c0341ffff262f58d1b7f031de4f0c53c8bec5/lib/private/Console/Application.php#L116-L121

In maintenance mode, only a minimum set of apps is loaded.  

The commands available in maintenance mode are mainly related to system operations. 
It's fine and for some even recommended to run those commands in maintenance mode.

However, some commands are difficult:

- `user:delete` will emit `BeforeUserDeletedEvent` and `UserRemovedEvent`. We have a couple of apps listening for those events: https://github.com/search?q=org%3Anextcloud+%28UserRemovedEvent+OR+BeforeUserDeleted%29&type=code

- `user:add` 

- `user:resetpassword`

- `group:add`

- `group:delete`

- `group:adduser`

- `group:removeuser`

Just to name some examples. 

My thought is, if we already know that a command is potentially unsafe, we should prevent the execution. 

## TODO

- [x] Make interface public
- [x] Rename IAvailableInMaintenanceMode to IUnavailableInMaintenanceMode (and flip the logic)

## Checklist


lib/private/Console/IAvailableInMaintenanceMode.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+
+namespace OC\Console;
+
+/**
+ * Implement if a core command is available in maintenance mode.
+ *
+ * A core command is a command registered in core/register_command.php.
+ * Don't implement the interface if the command is registered in IBootstrap or info.xml
```
nickvergessen, line 29: Why not? Talk has commands that could be run while maintenance is on?
Similar for files and others, maybe an admin restored a folder from a backup and wants to run file scan before sync clients interfere?
support:system-report, transfer-ownership, user-migration and others are also good examples for commands that could (maybe even should) be run in maintenance-mode.
As mentioned in #37365 also all the db:* commands are recommended to run in maintenance mode.

lib/private/Console/IAvailableInMaintenanceMode.php
```diff
@@ -0,0 +1,34 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\Console;
```
nickvergessen, line 23: Should definitely be in OCP for better usage.

lib/private/Console/IAvailableInMaintenanceMode.php
nickvergessen: I would even argue to default to "is available in maintenance mode" and have commands opt out of it.
But at the same time I would not know which command I would opt-out of it?

nickvergessen (changes requested):

violethaze74 (approved):
reviewed

lib/private/Console/Application.php
```diff
@@ -71,6 +76,22 @@ public function __construct(IConfig $config,
...
+	/**
+	 * Adds a command object.
+	 *
+	 * If a command with the same name already exists, it will be overridden.
+	 * If the command is not enabled it will not be added.
+	 *
+	 * @return Command|null The registered command if enabled or null
+	 */
+	public function add(Command $command) {
+		if ($this->maintenanceMode && in_array(IUnavailableInMaintenanceMode::class, class_implements($command, false))) {
```
nickvergessen, line 91:
```suggestion
		if ($this->maintenanceMode && $command instanceof IUnavailableInMaintenanceMode) {
```

???

core/register_command.php
nickvergessen, line 1: Can't find any changes apart from `s/$application/$this/` correct?
The new interface is not being added anywhere (totally okay, just asking)?

lib/private/Console/Application.php
kesselb: I understand your puzzlement ;)

class_implements is a leftover from an experiment. My idea was to change `Command` to `Command|string` and use the dependency injection to build the class when a string is given to clean up register_command.php a bit. 

Will use instanceof for now and submit another pr sometime for the dependency injection feature.

core/register_command.php
kesselb: Yep.

nickvergessen:
> In maintenance mode, only a minimum set of apps is loaded.

So if that is the reason, I would change the other half and also make sure to load all apps in CLI?
I would find it quite disturbing to not be able to prepopulate and manage an instance during maintenance mode.

Since we are approaching the next deadline in light speed maybe we can merge the interface and Application change in a single PR and discuss the implementing this quite restrictive behaviour in another one.

kesselb:
> So if that is the reason, I would change the other half and also make sure to load all apps in CLI?

As reference:

https://github.com/owncloud/core/issues/20939
https://github.com/owncloud/core/pull/21076

> I would find it quite disturbing to not be able to prepopulate and manage an instance during maintenance mode.

You put your instance in maintenance mode to delete a user :thinking: 

> Since we are approaching the next deadline in light speed maybe we can merge the interface and Application change in a single PR and discuss the implementing this quite restrictive behaviour in another one.

I see your concern, we don't need to rush. Let's continue the work on this feature when 27 is done.

My next task is to update the documentation: https://github.com/nextcloud/documentation/issues/10063

PR #37201, opened 2023-03-13
Fix: Escape group names for LDAP

AaronDewes:
## Summary

Nextcloud's LDAP feature allows to automatically generate queries so only certain groups are available in Nextcloud.

However, I noticed that groups may contain special characters (Like "(" or ")") that should be escaped to ensure generated queries are correct. Currently, the generated queries will be invalid and lead to 0 groups found if any  groups contain such characters.

I originally encountered this issue when trying to get Nextcloud running with a [MNS+](https://mnsplus.bildung-rp.de) system, a very popular school managment system in Rhineland-Palatinate.

## TODO

- [x] Are tests needed for this? There are no tests for this function at all, so I'm not sure if/how to write new ones.

## Checklist


come-nc (commented):
It would make sense to escape all the other values which are concatenated into ldap filters in the function.

AaronDewes:
> It would make sense to escape all the other values which are concatenated into ldap filters in the function.

Yes, I fixed this now. I think these are all values that need escaping, but please let me know if I missed any.

AaronDewes:
I did not intend to remove the other two requests, this seems to be a bug in GitHub (I just clicked "Re-request review from @come-nc").

come-nc (approved):

come-nc:
> Also, if this gets merged, can this be backported to Nextcloud 26(.0.2)?

I’m afraid it will be for 26.0.1, we are too close to 26 release to backport in time most likely.

blizzz:
we also have `Access::escapeFilterPart()` (not sure why i used this and not the escape method back then, maybe i simply was not aware of it, or we supported PHP < 5.6).

come-nc:
> we also have `Access::escapeFilterPart()` (not sure why i introduced this in 2014 and not the escape method back then, maybe i simply was not aware of it, or we supported PHP < 5.6).

escapeFilterPart also has a special handling of the asterisk as first character, if allowed.
But I do think it would work as well to do that:
```diff
diff --git a/apps/user_ldap/lib/Access.php b/apps/user_ldap/lib/Access.php
index 1cc0c62ff1d..658de8c0b83 100644
--- a/apps/user_ldap/lib/Access.php
+++ b/apps/user_ldap/lib/Access.php
@@ -1411,9 +1411,7 @@ class Access extends LDAPUtility {
                        $asterisk = '*';
                        $input = mb_substr($input, 1, null, 'UTF-8');
                }
-               $search = ['*', '\\', '(', ')'];
-               $replace = ['\\*', '\\\\', '\\(', '\\)'];
-               return $asterisk . str_replace($search, $replace, $input);
+               return $asterisk . ldap_escape($input, '', LDAP_ESCAPE_FILTER);
        }
 
        /**

```

AaronDewes:
>  `escapeFilterPart` also has a special handling of the asterisk as first character, if allowed.

For this purpose, `ldap_escape` should be the correct method though, because the user selects these groups from a list and does not manually enter them (This menu could be improved imho, but for now, I think this is the best fix).

AaronDewes:
@blizzz @ArtificialOwl is there anything I can do to Help this PR get merged? I'd really like to have this in the next release to save school IT admins in Rhineland-Palatinate some work at the beginning of the next school year.

blizzz:
@AaronDewes from my PoV also fix the other occurrences as described in https://github.com/nextcloud/server/pull/37201#issuecomment-1469835365 f

AaronDewes:
I did that in https://github.com/nextcloud/server/pull/37201/commits/f1b6ddcc310989c7d6db93f1e53464c0d21742dc, did I miss any?

blizzz:
In the Access class as pointed out.

come-nc (approved):

AaronDewes:
ldap_escape seems to convert a `*` to `\2a`. This is completely valid, but I'm not sure if it is desired?

come-nc:
> ldap_escape seems to convert a `*` to `\2a`. This is completely valid, but I'm not sure if it is desired?

The point of relying on the official function is to let it worry about these kind of things. It’s suppose to properly escape to a valid LDAP filter, if it does not we should file a bug to PHP.

AaronDewes:
It escapes properly, but it does it differently than Nextcloud previously did, so there's no bug.

AaronDewes:
@blizzz Could you please give this another review?

AaronDewes:
@ArtificialOwl @blizz Any news?

AaronDewes:
@joshtrichards @ArtificialOwl @blizz Can someone have a look again? This bug is really annoying to me, and it would be great to have it fixed soon.

artonge (approved):

Altahrim (approved):
Nit: maybe add a test with parenthesis as it was the original bug

AaronDewes:
> Nit: maybe add a test with parenthesis as it was the original bug

Parenthesis are correctly escaped by `escapeFilterPart`. Originally, this PR was just to make some generated LDAP filters use `escapeFilterPart`, all of which do not have tests yet at all AFAIK and I don't know how to implement myself.

PR #37206, opened 2023-03-14
fix: wrong redirect when skipping recommended apps

kesselb:
* Resolves: # 

## Summary

To access the page after setup: `index.php/core/apps/recommended`

**Before**

[Screencast from 2023-03-13 20-49-14.webm](https://user-images.githubusercontent.com/3902676/224959336-ddbc3490-50d1-450f-bc8d-21684c8ef337.webm)

**After**

[Screencast from 2023-03-13 20-48-10.webm](https://user-images.githubusercontent.com/3902676/224959365-9b75b02c-a145-4d81-9b72-9bd3708246f0.webm)

## TODO

- [x] CI

## Checklist


core/src/components/setup/RecommendedApps.vue
```diff
@@ -59,7 +59,6 @@
 			<NcButton v-if="showInstallButton"
 				type="tertiary"
 				role="link"
-				href="defaultPageUrl"
```
tcitworld, line 62: **suggestion:** We can keep it a link, just set
```vue
:href="defaultPageUrl"
```
and remove the `@click` event (and the useless method `goTo`).

kesselb: Sounds good. `defaultPageUrl` is defined outside the vue component hence we need an computed property to access the variable from the template.

core/src/components/setup/RecommendedApps.vue
```diff
@@ -129,10 +128,12 @@ export default {
 			loadingApps: true,
 			loadingAppsError: false,
 			apps: [],
-			defaultPageUrl,
 		}
 	},
 	computed: {
+		defaultPageUrl() {
+			return defaultPageUrl
+		},
```
tcitworld, line 136: **suggestion (non-blocking):** Data would be fine as well, since the value will not change.

kesselb: Thanks! 

image

This version, with defaultPageUrl as a dedicated variable, does not seem to work. 

image

It seems you need a computed property to access global variables or use loadState in data directly.

Pytal (approved):

ChristophWurst (approved):

PR #37209, opened 2023-03-14
Make it possible to strong type Entity properties

come-nc:
PropertyNotSetInConstructor still have to be suppressed because properties are
 actually set in fromRow/fromParams methods, and the init path is too
 convoluted anyway.

* Resolves: # 

## Summary

I tried several solutions to improve typing for QBMapper/Entity, I could not find a clean way to type properties without suppressing PropertyNotSetInConstructor manually and adding the @method annotations by hand as well.
But at least we can now type those properties and have psalm-compliant entities.
Also id property is actually nullable and null by default for autoincrement, made that clear in the code.

## Checklist

### DIFF
--- a/lib/private/Authentication/Token/IToken.php
+++ b/lib/private/Authentication/Token/IToken.php
@@ -38,7 +38,7 @@ interface IToken extends JsonSerializable {
 	/**
 	 * Get the token ID
 	 */
-	public function getId(): ?int;
+	public function getId(): int;
 
 	/**
 	 * Get the user UID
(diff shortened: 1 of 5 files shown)


lib/private/Authentication/Token/IToken.php
```diff
@@ -74,6 +72,7 @@ public function getLastCheck(): int;
 	 * Set the timestamp of the last password check
 	 *
 	 * @param int $time
+	 * @return void
```
Pytal, line 75: Inline typing doesn't work?

come-nc: Inline typing forces implementations to be updated with the same typing. Maybe I could here as it’s not OCP but I wanted to avoid too big changes.

Pytal (approved):

artonge (approved):

come-nc:
Giving up on this one :cry:

PR #37214, opened 2023-03-14
fix EOL workflow for pending major releases

szaimen:


.github/workflows/block-merge-eol.yml
```diff
@@ -27,4 +27,11 @@ jobs:
  
       - name: Checking if ${{ env.server_major }} is EOL
         run: |
-          php -r 'echo json_encode(require_once "config.php");' | jq --arg version "${{ env.server_major }}" '.stable[$version]["100"].eol' | grep --silent -i 'false'
+          if php -r 'echo json_encode(require_once "config.php");' | jq --arg version "${{ env.server_major }}" '.stable[$version]["100"].eol' | grep --silent -i 'false'; then
+            exit 0
+          fi
+
+          PREVIOUS_MAJOR="$((env.server_major-1))"
```
skjnldsv, line 34: This does not seem compliant to the way github workflow variables works? 
`env.server_major` would not be transformed here without `${{ env.server_major }}`, no?

szaimen: right, I've solved it by making a copy into a usual variable.

szaimen:
> Instead we could check if beta has it covered?

done

max-nextcloud:
> > Instead we could check if beta has it covered?
> 
> done

I think if you check beta you don't need to look at the previous release, right? For example stable26 -> 26 is in beta right now. We don't have a stable27 branch yet.

szaimen:
> I think if you check beta you don't need to look at the previous release, right? For example stable26 -> 26 is in beta right now. We don't have a stable27 branch yet.

right, my mistake

szaimen:
> > I think if you check beta you don't need to look at the previous release, right? For example stable26 -> 26 is in beta right now. We don't have a stable27 branch yet.
> 
> right, my mistake

fixed

max-nextcloud (approved):

blizzz (changes requested):
please open and complete against https://github.com/nextcloud/.github first

szaimen:
> please open and complete against https://github.com/nextcloud/.github first

done in https://github.com/nextcloud/.github/pull/130

skjnldsv (approved):

.github/workflows/block-merge-eol.yml
```diff
@@ -27,4 +27,10 @@ jobs:
...
       - name: Checking if ${{ env.server_major }} is EOL
         run: |
-          php -r 'echo json_encode(require_once "config.php");' | jq --arg version "${{ env.server_major }}" '.stable[$version]["100"].eol' | grep --silent -i 'false'
+          if php -r 'echo json_encode(require_once "config.php");' | jq --arg version "${{ env.server_major }}" '.stable[$version]["100"].eol' | grep --silent -i 'false'; then
+            exit 0
+          fi
+
+          if ! php -r 'echo json_encode(require_once "config.php");' | jq --arg version "${{ env.server_major }}" '.beta[$version]["100"].eol' | grep --silent -i 'false'; then
+            exit 1
+          fi
```
skjnldsv, line 36: You can use fallback in jq: https://stedolan.github.io/jq/manual/#Alternativeoperator://
```suggestion
          if php -r 'echo json_encode(require_once "config.php");' | jq --arg version "${{ env.server_major }}" '.stable[$version]["100"].eol // .beta[$version]["100"].eol'
' | grep --silent -i 'false'; then
            exit 0
          fi
```

skjnldsv (changes requested):

PR #37228, opened 2023-03-15
[stable24] non-blocking copy encryption wrapper

ArtificialOwl:
backport of #31379


blizzz:
@ArtificialOwl what's the state about this PR?

PR #37243, opened 2023-03-15
Getting rid of openssl_seal and rc4 in server side encryption

come-nc:
* Resolves: # 

## Summary

Get rid of the intermediate fileKey and the obsolete RC4 encryption used on it.
Also used the opportunity to change the padding used to OAEP.
Files encrypted with previous version can still be read but upon write they will get migrated to the new encryption.

## Checklist

### DIFF
--- a/apps/encryption/lib/Crypto/Crypt.php
+++ b/apps/encryption/lib/Crypto/Crypt.php
@@ -185,14 +185,9 @@ class Crypt {
 	}
 
 	/**
-	 * @param string $plainContent
-	 * @param string $passPhrase
-	 * @param int $version
-	 * @param int $position
-	 * @return false|string
 	 * @throws EncryptionFailedException
 	 */
-	public function symmetricEncryptFileContent($plainContent, $passPhrase, $version, $position) {
+	public function symmetricEncryptFileContent(string $plainContent, string $passPhrase, int $version, string $position): string|false {
 		if (!$plainContent) {
 			$this->logger->error('Encryption Library, symmetrical encryption failed no content given',
 				['app' => 'encryption']);
@@ -409,7 +404,7 @@ class Crypt {
 			$privateKey,
 			$hash,
 			0,
-			0
+			'0'
 		);
 
 		return $encryptedKey;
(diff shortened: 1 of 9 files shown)


yahesh:
> So the version in the header would only be written by the normal process, and would be read only for recovery scenarios?

Yes, exactly. I remember that someone wrote a fix-file-version-in-database command which tests through a bunch of version numbers to find the correct one. I wrote one such script myself in the past as well. Having the version number within the file header would obsolete that brute-force approach. And as the file header is rewritten with any re-encryption anyway...

> But in any case I think this would need to go in a follow-up PR as it is not tied to this one.

Fine for me.

come-nc:
Follow-up task: write an occ command to migrate all files to the new key format, when master key is used. (it only needs to open/write/close all encrypted files.)

artonge (approved):

skjnldsv (approved):

come-nc:
> @come-nc Hi and thanks for this work, could the follow-up task be tracked by an issue or draft PR ?

https://github.com/nextcloud/server/issues/37918

biboc:
Hi,

Does it break recover-tool? 
https://github.com/nextcloud/encryption-recovery-tools

New files added with Nextcloud 27 have no fileKey and the script does not work anymore

biboc:
Also, the documentation needs to be updated: https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/encryption_configuration.html

yahesh:
> Does it break recover-tool? https://github.com/nextcloud/encryption-recovery-tools
> 
> New files added with Nextcloud 27 have no fileKey and the script does not work anymore

This should be fixed.

come-nc:
> Also, the documentation needs to be updated: https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/encryption_configuration.html

I read this page and I am not sure which part you are referring to which should be updated?
What I’ve read is still true I think.

biboc:
I don't remember, someone may have removed the part I mentioned. I should have link to a specific Nextcloud version

come-nc:
@joshtrichards Can you explain why you added the pending documentation label back?

joshtrichards:
Oops missed this first time around. Not sure now. I seem to remember fixing the encryption tools link, but that was a different issue. Fine by me.

PR #36579, opened 2023-02-07
[stable25] fix(user-management): Fix deleting user in hidden group(s)

ChristophWurst:
Backport of #36509

PR #37267, opened 2023-03-16
[stable25] fix(user-management): Fix deleting user in hidden group(s)

szaimen:
Supersedes https://github.com/nextcloud/server/pull/36579


apps/settings/src/store/users.js
```diff
@@ -184,6 +184,10 @@ const mutations = {
 				state.userCount-- // decrement Active Users count
 				user.groups.forEach(userGroup => {
 					const group = state.groups.find(groupSearch => groupSearch.id === userGroup)
+					if (!group) {
+						console.warn('User group ' + userGroup + ' does not exist during user removal')
+						return
+					}
```
artonge, line 190: I would map and filter instead of this

szaimen: @ChristophWurst could you have a look at the comment? Not sure if I can follow Louis' suggestion...

ChristophWurst: It gets the same result of `undefined`. This is a backport. Let's leave it like the original PR and refactoring is desired one can clean up the code on `master`

artonge (approved):

blizzz (approved):

PR #37279, opened 2023-03-16
Add a getAttributes to IControllerMethodReflector

nickvergessen:
As discussed in https://github.com/nextcloud/server/pull/36928#discussion_r1131224444

Looking at the diff, I'm not sure it's worth it.

## Checklist

### DIFF
--- a/lib/private/AppFramework/Utility/ControllerMethodReflector.php
+++ b/lib/private/AppFramework/Utility/ControllerMethodReflector.php
@@ -33,12 +33,15 @@ declare(strict_types=1);
  */
 namespace OC\AppFramework\Utility;
 
+use Attribute;
 use OCP\AppFramework\Utility\IControllerMethodReflector;
 
 /**
  * Reads and parses annotations from doc comments
  */
 class ControllerMethodReflector implements IControllerMethodReflector {
+
+	protected ?\ReflectionMethod $reflection;
 	public $annotations = [];
 	private $types = [];
 	private $parameters = [];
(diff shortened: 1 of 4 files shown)


ChristophWurst:
I'm on the fence here. It's a small enhancement in all fairness.

lib/private/AppFramework/Utility/ControllerMethodReflector.php
```diff
@@ -33,12 +33,15 @@
...
 
+use Attribute;
 use OCP\AppFramework\Utility\IControllerMethodReflector;
 
 /**
  * Reads and parses annotations from doc comments
  */
 class ControllerMethodReflector implements IControllerMethodReflector {
+
+	protected ?\ReflectionMethod $reflection;
```
vitormattos, line 44: To be more specific because exist other types of reflections:

```suggestion
	protected ?\ReflectionMethod $reflectionMethod;
```

vitormattos (changes requested):

come-nc (approved):
:+1: I like it

lib/private/AppFramework/Utility/ControllerMethodReflector.php
```diff
@@ -138,4 +141,18 @@ public function getAnnotationParameter(string $name, string $key): string {
 
 		return '';
 	}
+
+	/**
+	 * @template T of Attribute
```
come-nc, line 146: Attribute classes do not extend Attribute class, which is final. So you cannot do that.

nickvergessen:
Exactly my thoughts after seeing the resulting PR.
So also tend to close it now and continue as before

PR #37284, opened 2023-03-17
fix: add logging for failed fopen operations that don't throw exceptions

miaulalala:
Add logging that shows some more file details when the fopen operation returns false.

### DIFF
--- a/lib/private/Streamer.php
+++ b/lib/private/Streamer.php
@@ -34,6 +34,7 @@ use OCP\Files\Folder;
 use OCP\Files\InvalidPathException;
 use OCP\Files\NotFoundException;
 use OCP\Files\NotPermittedException;
+use OCP\ILogger;
 use OCP\IRequest;
 use ownCloud\TarStreamer\TarStreamer;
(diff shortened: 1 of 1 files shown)


lib/private/Streamer.php
```diff
@@ -126,6 +127,9 @@ public function addDirRecursive(string $dir, string $internalDir = ''): void {
 			if ($file instanceof File) {
 				try {
 					$fh = $file->fopen('r');
+					if($fh === false) {
+						\OCP\Util::writeLog('core', 'unable to open file for stream: ' . print_r($file, true) , ILogger::ERROR);
```
icewind1991, line 131: please use `LoggerInterface` instead of the deprecated `writeLog`

icewind1991 (changes requested):

miaulalala:
Tests are passing on master

blizzz:
What do we do with this PR now?

miaulalala:
> What do we do with this PR now?

Nothing

PR #37295, opened 2023-03-19
fix right-click menu for open locally

szaimen:
<details>

```
docker run -it \
--name nextcloud-easy-test \
-p 8443:443 \
-e TRUSTED_DOMAIN=192.168.24.128 \
-e SERVER_BRANCH=enh/noid/fix-rightlick-open-locally \
-e RIGHTCLICK_BRANCH=master \
-e COMPILE_SERVER=1 \
```

</details>


apps/files/js/fileactions.js
```diff
@@ -716,15 +716,15 @@
...
 						}
 					},
 					mime: 'all',
 					order: -23,
-					icon: function(filename, context) {
+					iconClass: function(filename, context) {
 						var locked = context.$file.data('locked');
 						if (!locked) {
-							return OC.imagePath('files', 'computer.svg')
+							return 'icon-computer'
```
szaimen, line 727: doesnt seem to work yet...

apps/files/js/fileactions.js
```diff
@@ -716,15 +716,15 @@
...
 						var locked = context.$file.data('locked');
 						if (!locked) {
-							return t('files', 'Edit locally');
+							return t('files', 'Open locally');
 						}
 					},
 					mime: 'all',
 					order: -23,
-					icon: function(filename, context) {
+					iconClass: function(filename, context) {
```
szaimen, line 724: @Pytal changing to iconClass seems to make it visible in the right-click menu. However I still need to find out how to add the icon-computer class then.

Pytal:
What issue does this fix? The "Edit locally" action shows up correctly on NC26

szaimen:
> What issue does this fix? The "Edit locally" action shows up correctly on NC26

Not in the right-clik menu afaics...

apps/files/js/fileactions.js
Pytal: Instead of adding `icon-computer` which is the deprecated way, would probably be better to make it work with the `icon` property

apps/files/js/fileactions.js
```diff
@@ -716,15 +716,15 @@
 					displayName: function(context) {
 						var locked = context.$file.data('locked');
 						if (!locked) {
-							return t('files', 'Edit locally');
+							return t('files', 'Open locally');
```
Pytal, line 719: Should we change the wording from "Edit" to "Open" @jancborchardt?

apps/files/js/fileactions.js
szaimen: I fear that this would need to be fixed in the horrible rightclick app and unfortunately I dont see an easy way to achieve this in there :/

@skjlndsv would it be fine for now to create a new icon-computer class until it gets rewritten with vue?

Pytal: @skjnldsv 👆

skjnldsv:
Not relevant anymore

PR #28422, opened 2021-08-13
Allow users to choose a different email for notifications

blizzz:
- solves #27465
- part of  #26866

* [x] Implement local email verification
* [x] job to clean up expired verification tokens
* [x] Implement methods to for setting and getting notification emails, as laid out in #27465
* [ ] GUI controls (might be split off)

Screenshot_20210828_000844

Screenshot_20210820_210826

### Getting and setting primary mail via provisioning API

`curl -u $userid -X PUT -d 'key=notify_email' -d 'value=myname%40mydomain.com' -H 'OCS-APIRequest: true' https://my.nxt.cld/ocs/v2.php/cloud/users/$userid`

(mind the address must be added as additional email address and also be confirmed)

`curl -u $userid -X GET -H 'OCS-APIRequest: true' https://nc.zara/master/ocs/v2.php/cloud/users/$userid`

PR #37303, opened 2023-03-20
fix(provisionning_api): Remove parameters that are not set into template

tcitworld:
## Summary

Fixes
* `Notice: Undefined index: parameters`
* `Warning: Invalid argument supplied for foreach()`
* `Undefined index: targetUrl at /var/www/nc/nextcloud/core/templates/confirmation.php#9`

The template is only called by the Provisioning API in `apps/provisioning_api/lib/Controller/VerificationController.php`, brought in https://github.com/nextcloud/server/pull/28422.

## Checklist

### DIFF
--- a/core/templates/confirmation.php
+++ b/core/templates/confirmation.php
@@ -8,15 +8,12 @@
 /** @var \OCP\Defaults $theme */
 ?>
 <div class="update">
-	<form method="POST" action="<?php print_unescaped($_['targetUrl']);?>">
+	<form method="POST">
 		<h2><?php p($_['title']) ?></h2>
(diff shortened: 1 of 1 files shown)


nickvergessen:
Since it's pretty clear this thing is not used from any other point, let's move it also into the app, to avoid it being in core/ and therefore seen as "public API"

provokateurin (approved):

kesselb (approved):

skjnldsv (approved):

provokateurin:
Rebased to fix CI :+1:

PR #37328, opened 2023-03-21
Save user's first login time and output it in `occ user:info`

### LINKED ISSUE #22640: User created date
### How to use GitHub

Hi! It will be nice to see in `occ user:info` the date and time, when user was created. For now there is only last_seen value. But sometimes there is need to know also the creation date. 

Add to `occ user:info` new key with creation date and time

To have this information in interface in /settings/users

akhil1508:
* Resolves: https://github.com/nextcloud/server/issues/22640 & https://github.com/bpcurse/nextcloud-userexport/issues/52

### DIFF
--- a/core/Command/User/Info.php
+++ b/core/Command/User/Info.php
@@ -79,7 +79,8 @@ class Info extends Base {
 			'storage' => $this->getStorageInfo($user),
 			'last_seen' => date(\DateTimeInterface::ATOM, $user->getLastLogin()), // ISO-8601
 			'user_directory' => $user->getHome(),
-			'backend' => $user->getBackendClassName()
+			'backend' => $user->getBackendClassName(),
+			'created' => date(\DateTimeInterface::Atom, $user->getFirstLogin())
 		];
 		$this->writeArrayInOutputFormat($input, $output, $data);
 		return 0;
--- a/core/Command/User/Info.php
+++ b/core/Command/User/Info.php
@@ -80,7 +80,7 @@ class Info extends Base {
 			'last_seen' => date(\DateTimeInterface::ATOM, $user->getLastLogin()), // ISO-8601
 			'user_directory' => $user->getHome(),
 			'backend' => $user->getBackendClassName(),
-			'created' => date(\DateTimeInterface::Atom, $user->getFirstLogin())
+			'created' => date(\DateTimeInterface::ATOM, $user->getFirstLogin())
 		];
 		$this->writeArrayInOutputFormat($input, $output, $data);
 		return 0;
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -560,7 +560,8 @@ class Session implements IUserSession, Emitter {
 				// read only uses
 			}
 
-			$user->updateFirstLoginTimestamp($user->getLastLogin());
+			$userObj = $this->getUser();
+			$userObj->updateFirstLoginTimestamp($userObj->getLastLogin());
 			// trigger any other initialization
 			\OC::$server->getEventDispatcher()->dispatch(IUser::class . '::firstLogin', new GenericEvent($this->getUser()));
 		}
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -560,6 +560,7 @@ class Session implements IUserSession, Emitter {
 				// read only uses
 			}
 
+			$user->updateFirstLoginTimestamp($user->getLastLogin());
 			// trigger any other initialization
 			\OC::$server->getEventDispatcher()->dispatch(IUser::class . '::firstLogin', new GenericEvent($this->getUser()));
 		}
--- a/lib/private/User/User.php
+++ b/lib/private/User/User.php
@@ -254,7 +254,7 @@ class User implements IUser {
 	 */
 
 	public function getFirstLogin() {
-		if($this->firstLogin === null) {
+		if ($this->firstLogin === null) {
 			$this->firstLogin = (int) $this->config->getUserValue($this->uid, 'login', 'firstLogin', 0);
 		}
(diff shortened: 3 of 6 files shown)


akhil1508:
It would be very simple to just use `setUserValue` maybe even in the "firstLogin" hook but seems like the correct way is to add to the `IUser` interface and proceed from there. If a maintainer could comment on whether this is the right approach or whether it is better to just add to the hook directly, it would be great.

lib/private/User/User.php
```diff
@@ -89,6 +89,9 @@ class User implements IUser {
 	/** @var int|null */
 	private $lastLogin;
 
+	/** @var int|null */
+	private $firstLogin;
```
come-nc, line 93:
```suggestion
	private ?int $firstLogin;
```
Please strong type new code

lib/public/IUser.php
```diff
@@ -270,4 +270,19 @@ public function getQuota();
...
 	public function setQuota($quota);
+
+	/**
+	 * returns the timestamp of the user's first login or 0 if the user did never
+	 * login
+	 *
+	 * @return int
+	 * @since 27.0.0
+	 */
+	public function getFirstLogin();
```
come-nc, line 281:
```suggestion
	public function getFirstLogin(): int;
```

lib/public/IUser.php
```diff
@@ -270,4 +270,19 @@ public function getQuota();
...
+
+	/**
+	 * updates the timestamp of the first login of this user
+	 * @since 27.0.0
+	 */
+	public function updateFirstLoginTimestamp(int $timestamp);
```
come-nc, line 287:
```suggestion
	public function updateFirstLoginTimestamp(int $timestamp): void;
```

core/Command/User/Info.php
```diff
@@ -79,7 +79,8 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			'storage' => $this->getStorageInfo($user),
 			'last_seen' => date(\DateTimeInterface::ATOM, $user->getLastLogin()), // ISO-8601
 			'user_directory' => $user->getHome(),
-			'backend' => $user->getBackendClassName()
+			'backend' => $user->getBackendClassName(),
+			'created' => date(\DateTimeInterface::ATOM, $user->getFirstLogin())
```
come-nc, line 83: Creation and first login are not the same thing.
Also, this should handle properly the case where firstlogin is 0.

come-nc (commented):
I am not against the idea but a choice needs to be made on whether we want creation date or first login date (or both?), this is not the same thing.

I have no strong opinion about whether to advertize it in IUser, it feels pretty natural how it is done in the PR as it mimics lastLogin API.

lib/private/User/User.php
```diff
@@ -260,6 +277,14 @@ public function updateLastLoginTimestamp() {
 		return $firstTimeLogin;
 	}
 
+	/**
+	 * updates the timestamp of the first login of this user
+	 */
+	public function updateFirstLoginTimestamp(int $timestamp) {
+		$this->firstLogin = $timestamp;
+		$this->config->setUserValue($this->uid, 'login', 'firstLogin', (string)$this->firstLogin);
+	}
```
come-nc, line 286: Should there be some kind of check about whether a first login timestamp was already set for this user?
But not sure what could be done if it’s a case, maybe at least log an error.

lib/private/User/Session.php
```diff
@@ -560,6 +560,8 @@ protected function prepareUserLogin($firstTimeLogin, $refreshCsrfToken = true) {
 				// read only uses
 			}
 
+			$userObj = $this->getUser();
+			$userObj->updateFirstLoginTimestamp($userObj->getLastLogin());
 			// trigger any other initialization
 			\OC::$server->getEventDispatcher()->dispatch(IUser::class . '::firstLogin', new GenericEvent($this->getUser()));
```
come-nc, line 566: It would feel prettier if the call to updateFirstLoginTimestamp is instead in a Listener for the event firstLogin.
Sadly it seems there is no modern event for this yet, only the deprecated GenericEvent.
There is a `UserCreatedEvent` on the other hand.

lib/private/User/User.php
```diff
@@ -243,6 +246,20 @@ public function getLastLogin() {
...
 	}
 
+	/**
+	 * returns the timestamp of the user's first login or 0 if the user did never
+	 * login
+	 *
+	 * @return int
+	 */
+
+	public function getFirstLogin() {
```
icewind1991, line 256: might be better to return `false` for cases when the user first logged in, but I have no strong feelings either way.

lib/public/IUser.php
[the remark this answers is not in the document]
come-nc: Well even in core most of the time the objects will be typed as IUser and not User, most likely this is the case here as well? Especially if it gets moved to a listener, most likely the event will give an IUser instance.

skjnldsv:
Hey @akhil1508 !
Thanks for the pull request. As we would like to push https://github.com/nextcloud/server/issues/22640 forward, and this PR had some unresolved comments, we'll close and move this to a different ticket.

I suggest you keep an eye on https://github.com/nextcloud/server/issues/22640 to be notified of its resolution :rocket: 
Thank you for your help and opening this ticket!
Have a nice day and weekend!
Cheers,
John

akhil1508:
> Thanks for the pull request. As we would like to push https://github.com/nextcloud/server/issues/22640 forward, and this PR had some unresolved comments, we'll close and move this to a different ticket.

@skjnldsv  I've unfortunately been unable to continue my work on this PR in a timely manner. No worries at all, hopefully the other PR gets merged, I would like first and foremost for the feature to make it to a release :)

PR #29423, opened 2021-10-24
SetupController#run(): prevent admin auto-login after autoconfig-only setup

Al2Klimov:

PR #37330, opened 2023-03-21
Omit "Accessing site insecurely via HTTP." warning if HTTP is secure enough

Al2Klimov:
## Summary

This is the case for localhost and Tor hidden services.

## Checklist

### DIFF
--- a/core/js/setupchecks.js
+++ b/core/js/setupchecks.js
@@ -717,7 +717,7 @@
 							type: OC.SetupChecks.MESSAGE_TYPE_WARNING
 						});
 					}
-				} else {
+				} else if (!/(?:^(?:localhost|127\.0\.0\.1|::1)|\.onion)$/.exec(window.location.hostname)) {
 					messages.push({
 						msg: t('core', 'Accessing site insecurely via HTTP. You are strongly advised to set up your server to require HTTPS instead, as described in the {linkstart}security tips ↗{linkend}.')
 							.replace('{linkstart}', '<a target="_blank" rel="noreferrer noopener" class="external" href="' + tipsUrl + '">')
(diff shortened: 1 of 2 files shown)


Al2Klimov:
Cool that you consider this for NC 27. https://github.com/nextcloud/server/pull/37330#event-8824318862

But IMAO there’s some a little more important of mine, a literal security hole IMAO:

* #29423

artonge (approved):
Makes sense

core/js/setupchecks.js
```diff
@@ -717,7 +717,7 @@
 							type: OC.SetupChecks.MESSAGE_TYPE_WARNING
 						});
 					}
-				} else {
+				} else if (!/(?:^(?:localhost|127\.0\.0\.1|::1)|\.onion)$/.exec(window.location.hostname)) {
```
Pytal, line 720: Could we add tests for the cases where the "Accessing site insecurely via HTTP" message is returned?

Pytal: Makes sense, then please adapt the removed test for when non-localhost IPs are accessed insecurely :)

Al2Klimov: Are they accessed at all? If they were, I wouldn’t have to remove the test I guess.

Al2Klimov: If you **really want** that test, I _could_ drop the localhost handling and just keep the Tor hidden services. Almost no one actually uses NC in prod via localhost.

Pytal: Only nice-to-have I guess so current state is fine with me @Al2Klimov :)

Pytal (approved):

PR #37345, opened 2023-03-22
Fix cypress tests for files_versions 

artonge:
~~In files_versions cypress tests, we wait for 1000ms between each versions uploads to make sure that the server properly create a version. Under 1s the serve does not create a version.~~

~~This commit increases the wait time as it might help to ensure that the time span is enough between two uploads.~~

The automatic versions expiration process was kicking in during the tests, making some versions unavailable. I changed the tests to isolate expiration process testing from other tests.

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -57,7 +57,7 @@ jobs:
       fail-fast: false
       matrix:
         # run multiple copies of the current job in parallel
-        containers: ['component', 1, 2]
+        containers: ["component", 1, 2]
 
     name: runner ${{ matrix.containers }}
 
@@ -90,6 +90,17 @@ jobs:
           TESTING: true
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+
+      - name: Extract NC logs
+        if: always()
+        run: docker-compose --project-directory cypress logs > nextcloud.log
+
+      - name: Upload NC logs
+        uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: nc_logs_${{ matrix.containers }}
+          path: nextcloud.log
 
   summary:
     runs-on: ubuntu-latest
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -92,7 +92,7 @@ jobs:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
 
       - name: Extract NC logs
-        if: ${{ matrix.containers != 'component' }}
+        if: ( success() || failure() ) && matrix.containers != 'component'
         run: docker logs nextcloud-cypress-tests-server > nextcloud.log
 
       - name: Upload NC logs
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -92,12 +92,12 @@ jobs:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
 
       - name: Extract NC logs
-        if: always()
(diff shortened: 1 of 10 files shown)


cypress/e2e/files_versions/filesVersionsUtils.ts
```diff
@@ -24,9 +24,9 @@ import path from "path"
 
 export function uploadThreeVersions(user) {
 	cy.uploadContent(user, new Blob(['v1'], { type: 'text/plain' }), 'text/plain', '/test.txt')
-	cy.wait(1000)
+	cy.wait(1500)
```
skjnldsv, line 27:
```suggestion
	// A new version will not be created if the changes occur
	// within less than one second of each other.
	cy.wait(1500)
```

skjnldsv (approved):

.github/workflows/cypress.yml
```diff
@@ -91,6 +91,17 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
 
+      - name: Extract NC logs
+        if: always()
+        run: docker-compose --project-directory cypress logs > nextcloud.log
```
skjnldsv, line 96: We're not using docker compose?

skjnldsv (changes requested):
We're not using docker compose?

.github/workflows/cypress.yml
skjnldsv: Also, it should not be run on the component testing runner :)

artonge: Indeed, good catch, copy pasting does not always work !

artonge: Not sure how to exclude component testing. Try `${{ matrix.containers !== 'component' }}`

skjnldsv: No need for the `${{}}` https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

artonge:
Strongly suspecting auto expiration from deleting files, let's see

cypress.config.ts
```diff
@@ -71,7 +71,9 @@ export default defineConfig({
 
 			// Remove container after run
 			on('after:run', () => {
-				stopNextcloud()
+				if (!process.env.CI) {
+					stopNextcloud()
+				}
```
artonge, line 76: Needed to extract logs and data directory

artonge (commented):
@skjnldsv I managed to get the tests working again. I was tricked by the auto expiration process...

The culprit is that I move the `occ config:system:set versions_retention_obligation` call from `dockerNode.ts` to `version_expiration.cy.ts`.

Note that: 

- I extended the cypress workflow to extract logs and data directory. Can be remove but seems useful. Ok to keep?
- The change also contains logic to use a random file name instead of 'test.txt'. Not needed, but it's here and it works.

.github/workflows/cypress.yml
```diff
@@ -91,6 +91,28 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
 
+      - name: Extract NC logs
+        if: ( success() || failure() ) && matrix.containers != 'component'
```
come-nc, line 95: `( success() || failure() )` is not always true? :thinking:

come-nc (commented):
It should only upload log and data on failure, no?

.github/workflows/cypress.yml
artonge: Yes, it is uneeded. Let's reduce our data consumption :)

skjnldsv (approved):
Looks good! :+1:

come-nc (approved):

PR #37365, opened 2023-03-23
Block occ commands when in maintenance mode

miaulalala:
## Summary
Currently, occ allows execution of commands even while in maintenance mode. The CLI does warn the admin that commands are limited but the execution still works.

This PR adds an exception to block any command that isn't `maintenance:mode --off` or `status` as executing other commands is unsafe.

(Either this, or remove the misleading sentence as commands can be executed and the database is indeed available - occ user:add for example works).

## Checklist

### DIFF
--- a/lib/private/Console/Application.php
+++ b/lib/private/Console/Application.php
@@ -32,6 +32,7 @@ namespace OC\Console;
 
 use OC\MemoryInfo;
 use OC\NeedsUpdateException;
+use OC\ServiceUnavailableException;
 use OC_App;
 use OCP\AppFramework\QueryException;
 use OCP\App\IAppManager;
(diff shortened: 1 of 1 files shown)


miaulalala:
Alternative PR by Daniel here: https://github.com/nextcloud/server/pull/37178

nickvergessen:
We have quite some commands in the docs where we recommend to put the instance into maintenance before (and even turn off the webserver), so I think it's intended that it still works :/

miaulalala:
> We have quite some commands in the docs where we recommend to put the instance into maintenance before (and even turn off the webserver), so I think it's intended that it still works :/
> 
> Long running and expensive things like:
> 
> * Add missing index
> * Add primary key
> * Convert to utf8mb4
> 
> And so on come to mind.

Then Daniel's version is the superior approach. Closing this then.

violethaze74 (commented):
reviewed

PR #37370, opened 2023-03-23
handle trashed groupfolder files as non-accessible

icewind1991:
Not really happy about adding a special case to the sharing logic but I can't think of a good generic approach atm.


come-nc (approved):

icewind1991:
#37448

PR #37390, opened 2023-03-24
Update to psalm 5

provokateurin:
## Checklist

### DIFF
--- a/build/psalm-baseline-ocp.xml
+++ b/build/psalm-baseline-ocp.xml
@@ -1,130 +1,152 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<files psalm-version="4.30.0@d0bc6e25d89f649e4f36a534f330f8bb4643dd69">
+<files psalm-version="5.9.0@8b9ad1eb9e8b7d3101f949291da2b9f7767cd163">
   <file src="lib/private/legacy/OC_Template.php">
-    <UndefinedClass occurrences="1">
+    <UndefinedClass>
       <code>OC</code>
     </UndefinedClass>
   </file>
   <file src="lib/public/AppFramework/ApiController.php">
-    <NoInterfaceProperties occurrences="1">
-      <code>$this-&gt;request-&gt;server</code>
+    <NoInterfaceProperties>
+      <code><![CDATA[$this->request->server]]></code>
     </NoInterfaceProperties>
   </file>
   <file src="lib/public/AppFramework/App.php">
-    <InternalMethod occurrences="1">
-      <code>new RouteConfig($this-&gt;container, $router, $routes)</code>
+    <InternalMethod>
+      <code><![CDATA[new RouteConfig($this->container, $router, $routes)]]></code>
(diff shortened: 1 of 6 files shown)


apps/dav/lib/Connector/Sabre/Principal.php
```diff
@@ -434,6 +434,7 @@ protected function searchUserPrincipals(array $searchProperties, $test = 'allof'
 
 			case 'allof':
 			default:
+				/** @var  array<array-key, mixed> $results */
```
ChristophWurst, line 437: does this express anything?

build/stubs/psr_container.php
```diff
@@ -0,0 +1,52 @@
...
+ * Base interface representing a generic exception in a container.
+ */
+interface ContainerExceptionInterface extends Throwable
+{
+}
+
+/**
+ * Describes the interface of a container that exposes methods to read its entries.
+ */
+interface ContainerInterface
```
ChristophWurst, line 19: why do we need a stub now?

composer.json
```diff
@@ -47,7 +47,7 @@
 		"cs:fix": "php-cs-fixer fix",
 		"cs:check": "php-cs-fixer fix --dry-run --diff",
 		"lint": "find . -name \\*.php -not -path './lib/composer/*' -not -path './build/stubs/*' -print0 | xargs -0 -n1 php -l",
-		"psalm": "psalm --threads=1",
```
ChristophWurst, line 50: unrelated to the update to v5, isnt it?

ChristophWurst (commented):
I would prefer to split this into more than one PR. It is not trivial to comprehend, review and approve this PR. I don't feel like waving this through.

provokateurin:
@ChristophWurst I can definitely split it into multiple PRs. I would try to land the individual fixes first and then upgrade psalm afterwards, right? Then I can separate those fixes by type (there are a few fixes that just repeat over and over throughout the whole codebase).

ChristophWurst:
Yes, submit the fixes separately. Psalm 4 will like them too and they also seem legit from looking at the diff.

build/stubs/psr_container.php
provokateurin: lmao when running with a single thread the errors don't show up, but using more than one thread breaks it... no idea what is going on there. i'd like to be able to run psalm with more than one thread though, since it speeds up things a lot (in CI too).

provokateurin:
I already tried fixing a lot of the new found issues, but it's getting out of hand. What I did now is just ignore all the new errors so that we can finally have psalm 5.

provokateurin:
No idea what the failing security analysis means, the output by psalm seems to be incorrect.

kesselb (approved):
LGTM :+1:

provokateurin:
Yeah I filed an issue at https://github.com/vimeo/psalm/issues/9712, I hope it gets fixed soon.

come-nc (approved):

PR #37397, opened 2023-03-25
check-setup: recommend exif extension

pulsejet:
JPEG previews are not correctly oriented when exif is not enabled along with other issues. There's no easy way to discover this either.


szaimen (approved):
LGTM

skjnldsv (approved):

PR #37403, opened 2023-03-25
Optimize to return quicker

solracsf:
## Summary

Rewrite code to return quicker

## Checklist

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -183,11 +183,7 @@ class View {
 			return null;
 		} else {
 			$path = substr($path, strlen($this->fakeRoot));
-			if (strlen($path) === 0) {
-				return '/';
-			} else {
-				return $path;
-			}
+			return strlen($path) === 0 ? '/' : $path;
 		}
 	}
 
@@ -241,11 +237,10 @@ class View {
 		$parent = substr($path, 0, strrpos($path, '/'));
 		$path = $this->getAbsolutePath($path);
 		[$storage, $internalPath] = Filesystem::resolvePath($path);
-		if (Filesystem::isValidPath($parent) and $storage) {
+		if (Filesystem::isValidPath($parent) && $storage) {
 			return $storage->getLocalFile($internalPath);
-		} else {
-			return null;
 		}
+		return null;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


lib/private/Files/View.php
```diff
@@ -347,11 +341,8 @@ public function rmdir($path) {
 		if ($mount->getInternalPath($absolutePath) === '') {
 			return $this->removeMount($mount, $absolutePath);
 		}
-		if ($this->is_dir($path)) {
-			$result = $this->basicOperation('rmdir', $path, ['delete']);
-		} else {
-			$result = false;
-		}
+
+		$result = $this->is_dir($path) ? $this->basicOperation('rmdir', $path, ['delete']) : false;
```
ChristophWurst, line 345: this feels dangerous. does php only evaluate the ternary branches if they match? even if so, I don't like it here. let's use the ternary operator only for pure expressions

lib/private/Files/View.php
```diff
@@ -932,11 +919,7 @@ public function copy($source, $target, $preserveMtime = false) {
 					$lockTypePath2 = ILockingProvider::LOCK_EXCLUSIVE;
 
 					if ($mount1->getMountPoint() == $mount2->getMountPoint()) {
-						if ($storage1) {
-							$result = $storage1->copy($internalPath1, $internalPath2);
-						} else {
-							$result = false;
-						}
+						$result = $storage1 ? $storage1->copy($internalPath1, $internalPath2) : false;
```
ChristophWurst, line 922: same here

ChristophWurst (commented):
Not  a fan of using the ternary operator for statements

lib/private/Files/View.php
ChristophWurst: https://3v4l.org/Y6SaA the ternary operation is lazy. But still not a fan of this.

solracsf: No problem. Symfony makes heavy use of it without no problems, but ok, reverted them :)

lib/private/Files/View.php
```diff
@@ -1727,11 +1706,7 @@ public function getETag($path) {
 		 * @var string $internalPath
 		 */
 		[$storage, $internalPath] = $this->resolvePath($path);
-		if ($storage) {
-			return $storage->getETag($internalPath);
-		} else {
-			return null;
-		}
+		return $storage->getETag($internalPath) ?? null;
```
ChristophWurst, line 1709: if storage is null you'll get an error here

solracsf: Doesn't the null coalescing operator `??` achieve the same result?
This operator returns the value of its left operand if it is not null, and otherwise returns the value of its right operand. Am I missing something here?

solracsf: oh sorry, you're right, my bad 😭

come-nc (commented):
Not a big fan of the changes apart from the switch to `&&` and `||`.
But unless I missed it nothing here is quicker, the behavior should be exactly the same?

PR #17784, opened 2019-11-02
Disable Clear-Site-Data for Chrom* (and Opera, Brave, etc)

kesselb:
Close #9179 (again)

I think Clear-Site-Data is a good feature but it will slow down the logout for chromium based browsers. For some reason clearing the site data (see https://github.com/nextcloud/server/issues/9179#issuecomment-502928769 and https://github.com/nextcloud/server/issues/9179#issuecomment-509629565) takes a while. There is nothing we can do.

People will always complain to their administrators about the slow logout. We should give them a way to opt out. I've been thinking about sending the header only for firefox but this makes things even more complicated. 

Since this patch has no real impact I would like to ask for a backport to 17.

PR #37405, opened 2023-03-25
Send Clear-Site-Data header and let browsers ignore it if unsupported

solracsf:
## Summary

According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data#browser_compatibility only Safari don't support it.

## Checklist

### DIFF
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -34,7 +34,6 @@ declare(strict_types=1);
  */
 namespace OC\Core\Controller;
 
-use OC\AppFramework\Http\Request;
 use OC\Authentication\Login\Chain;
 use OC\Authentication\Login\LoginData;
 use OC\Authentication\WebAuthn\Manager as WebAuthnManager;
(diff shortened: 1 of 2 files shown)


skjnldsv:
Do we gain anything to just not send it if safari?
Wouldn't safari just ignore the header?

If so, we should not even bother and send it for all browsers, no?

solracsf:
Sure but as i didn't knew the reason why some were balcklisted i assumed it was a compatiblity issue; but yes, this is just a header, browsers will just ignore it if unsupported.

solracsf:
@blizzz I've changed the code accordingly.

artonge (approved):

skjnldsv (approved):

skjnldsv (changes requested):
`tests/Core/Controller/LoginControllerTest.php:160` needs a fix :)

skjnldsv (approved):

joshtrichards:
The blacklist was due to #17784 which is Chrome/Chromium specific and apparently still a problem in the field (e.g. see #41196).

HLFH:
I have added this PR that fixes the issue: https://github.com/nextcloud/server/pull/41204.

PR #37413, opened 2023-03-26
Fix 2FA icon

solracsf:
## Summary
| before | after |
|---|---|
| image | image |

## Checklist


artonge (approved):

jancborchardt (changes requested):
This seems like an issue with the icon and should be fixed at the source, no?
- With light background, we need a dark icon
- With dark background, we need a light icon

We generally don’t put containers around our emptycontent icons like these. :)

solracsf:
@jancborchardt makes sense.

PR #37428, opened 2023-03-27
Ignore and log non integer versions

come-nc:
This avoids being stuck when there is one buggy file version with a
 non-numeric version. It allows the expiration process to not crash and
 continue with the other ones.

## Checklist

### DIFF
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -716,13 +716,13 @@ class Storage {
 		}
 
 		foreach ($versions as $key => $version) {
-			if (!is_int($version['version'])) {
+			if (!is_numeric($version['version'])) {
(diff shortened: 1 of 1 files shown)


blizzz:
> This avoids being stuck when there is one buggy file version with a
non-numeric version. 

What's the root cause?

icewind1991 (approved):

come-nc:
> > This avoids being stuck when there is one buggy file version with a
> > non-numeric version.
> 
> What's the root cause?

No idea, might be moving file to/from groupfolders, or a buggy application.
This is to help debugging this case and not have the whole version expiry system crash down because of one badly named file.

artonge (approved):

come-nc:
> /backport to stable27

It’s already in 27

PR #37429, opened 2023-03-27
improve objectstore rmdir handling

icewind1991:
Remove folder entries as they are cleared instead of in one go afterwards

Otherwise they stick around if some of the child entries can't be removed

Also includes some optimizations by querying the cache less.

### DIFF
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -27,6 +27,7 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>
  *
  */
+
 namespace OC\Files\ObjectStore;
 
 use Aws\S3\Exception\S3Exception;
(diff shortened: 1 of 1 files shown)


lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -177,63 +178,65 @@ public function getId() {
...
-			} catch (\Exception $ex) {
-				if ($ex->getCode() !== 404) {
-					$this->logger->logException($ex, [
-						'app' => 'objectstore',
-						'message' => 'Could not delete object ' . $this->getURN($stat['fileid']) . ' for ' . $path,
-					]);
-					return false;
-				}
-				//removing from cache is ok as it does not exist in the objectstore anyway
+		if ($entry && $entry->getId() > 0) {
```
come-nc, line 213: When is id negative or 0?

icewind1991: was adapted from the old check, presumable for "partial" cache entries, but on further consideration I don't think it's needed

come-nc (approved):

juliusknorr (approved):

PR #37432, opened 2023-03-27
Fix: Make OC\Console\Application::$application public

### LINKED ISSUE #5: NC26 - Exception: Cannot access private property OC\Console\Application::$application
Hello all,

I have the following error in my logging after updating to version 26.0.0.11:

Exception: Cannot access private property OC\Console\Application::$application in file '/nextcloud/apps/occweb/lib/Controller/OccController.php' line 81

and the app does not work anymore. 

Lines 80-87 are as follows:
  public function list() {
    $defs = $this->application->application->all();
    $cmds = array();
    foreach ($defs as $d) {
      array_push($cmds, $d->getName());
    }
    return new DataResponse($cmds);
  }

Line 81:
$defs = $this->application->application->all();

I know OCC Web is no longer supported but up to version 24 the app still worked to execute occ commands.
**My question now, how would line 81 have to be changed to make occ web work again, at least rudimentarily?**

### Steps to reproduce

1. update to version 26
2. open occ web
3. Black surface
4. look into the logging

### Expected behavior

the app opens normally again without a black screen

AaronDewes:
* Resolves: https://github.com/Adphi/occweb/issues/5

## Summary

The app [OCC-Web](https://github.com/adphi/occweb) relied on this being public.

OCC-Web is technically deprecated, and this API was probably never meant to be public, but I still think it'd be useful to make this public now, considering it has been before; and OCC-Web is really convenient.

## Checklist

### DIFF
--- a/lib/private/Console/Application.php
+++ b/lib/private/Console/Application.php
@@ -49,7 +49,7 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 class Application {
 	/** @var IConfig */
 	private $config;
-	private SymfonyApplication $application;
+	public SymfonyApplication $application;
 	/** @var EventDispatcherInterface */
 	private $dispatcher;
 	/** @var IRequest */


ChristophWurst:
> this API was probably never meant to be public, but I still think it'd be useful to make this public now, considering it has been before; and OCC-Web is really convenient.

Correct. Everthing in `OC\` is internal. If an app requires access to the occ commands we should create a real public API instead.

I vote against this change.

come-nc (changes requested):
Sorry but I feel the same, we cannot dirty patch this for a deprecated application.
Looking at occweb it only use this to access `all()` method, so it could be added in a new proxy method and in a `OCP\Console\IApplication` interface, but this only makes sense if there is development on the application side to actually use the new interface.

ChristophWurst:
Closing because the approach is not technically sound.

If we expose commands to apps it has to be through a public API.

Issue #36753, opened 2023-02-16
[BITV] 9.1.1.1a/2.1 - The Nextcloud logo (link) does not have an alternative text. A suitable alternative text could be "Nextcloud - Go to dashboard."


AndyScherzinger:
- [ ] Fix
- [ ] Backport

![2b66d051c0e7a904424fb940d9032598](https://user-images.githubusercontent.com/1315170/219386835-211b0c58-f7ba-425b-be2b-8314c97de6f4.png)

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-cd239865a5-v2-n1

</details>

JuliaKirschenheuter:
@michaelnissenbaum It is actually not easy to set a detailed name for the logo-link because a link for the logo is configurable in: https://github.com/nextcloud/server/blob/dde5c46a3eb7c6dcee47795590619ccd393577d2/config/config.sample.php#L224-L233
This means it depends on the config and on user access/rights which link is available for a current user. 

For now the implementation looks like this which always says "(name of the nextcloud instance) homepage", whatever the current homepage is: 

![Screenshot from 2023-02-28 14-35-23](https://user-images.githubusercontent.com/6078378/221869657-51690d1f-92db-4d39-ba6c-c80dc798d4b2.png)

Could we move on with this solution? 
Thank you!

skjnldsv:
The logo is irrelevant here. It should be treated as a decorative image.
What is important is the aria description of the link. You can edit this and have access to the `defaultapp` config with php.
Then use it to inject the proper aria attributes in the user template
https://github.com/nextcloud/server/blob/0d67fc23f4077e7b06a01bc519957f3f13d95f10/core/templates/layout.user.php#L61

skjnldsv:
Extract the logic from https://github.com/nextcloud/server/blob/71f698649f578db19a22457cb9d420fb62c10382/lib/private/URLGenerator.php#L302-L338

And create a method to get the default AppId somewhere it make sense (Maybe [User.php](https://github.com/nextcloud/server/blob/ecad09b38a152957c771139562e61f7ecb1a6879/lib/private/User/User.php) ? :shrug: )

PR #37435, opened 2023-03-28
Add label for logo link

Pytal:
- For https://github.com/nextcloud/server/issues/36753

## Summary

Add a descriptive label for the logo link
image

To be backported to `stable25` for accessibility

## TODO

- [x] Remove extraneous `span` translated string
- [x] Fix label when going to `logoUrl`
- [x] Translate app name
- [x] Fix existing tests
- [x] Add tests

## Checklist

### DIFF
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -827,4 +827,28 @@ class AppManager implements IAppManager {
 
 		return $this->defaultEnabled;
 	}
+
+	public function getDefaultAppForUser(?IUser $user = null): string {
+		// Set fallback to always-enabled files app
+		$appId = 'files';
+		$defaultApps = explode(',', $this->config->getSystemValueString('defaultapp', 'dashboard,files'));
+
+		$user ??= $this->userSession->getUser();
+
+		if ($user !== null) {
+			$userDefaultApps = explode(',', $this->config->getUserValue($user->getUID(), 'core', 'defaultapp'));
+			$defaultApps = array_filter(array_merge($userDefaultApps, $defaultApps));
+		}
+
+		// Find the first app that is enabled for the current user
+		foreach ($defaultApps as $defaultApp) {
+			$defaultApp = \OC_App::cleanAppId(strip_tags($defaultApp));
+			if ($this->isEnabledForUser($defaultApp, $user)) {
+				$appId = $defaultApp;
+				break;
+			}
+		}
+
+		return $appId;
+	}
 }
--- a/lib/public/App/IAppManager.php
+++ b/lib/public/App/IAppManager.php
@@ -242,4 +242,13 @@ interface IAppManager {
 	 * @since 17.0.0
 	 */
 	public function getAppRestriction(string $appId): array;
+
+	/**
+	 * Returns the id of the user's default app
+	 *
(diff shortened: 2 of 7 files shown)


lib/private/App/AppManager.php
```diff
@@ -827,4 +827,27 @@ public function getDefaultEnabledApps():array {
 
 		return $this->defaultEnabled;
 	}
+
+	public function getDefaultAppForUser(?IUser $user = null): string {
+		$appId = 'files';
```
skjnldsv, line 832: I think dashboard is the default :)

lib/private/App/AppManager.php
```diff
@@ -827,4 +827,27 @@ public function getDefaultEnabledApps():array {
...
+		$defaultApps = explode(',', $this->config->getSystemValueString('defaultapp', 'dashboard,files'));
+
+		$user ??= $this->userSession->getUser();
+
+		if ($user !== null) {
+			$userDefaultApps = explode(',', $this->config->getUserValue($user->getUID(), 'core', 'defaultapp'));
```
skjnldsv, line 838: `defaultapp` ?

lib/public/App/IAppManager.php
```diff
@@ -242,4 +242,13 @@ public function getEnabledAppsForGroup(IGroup $group): array;
 	 * @since 17.0.0
 	 */
 	public function getAppRestriction(string $appId): array;
+
+	/**
+	 * Returns the id of the user's default app
+	 *
+	 * If `user` is not passed, the currently logged in user will be used
+	 *
+	 * @since 25.0.0
```
come-nc, line 251:
```suggestion
	 * @since 25.0.6
```
I suppose?

come-nc (approved):
All good apart from skjnldsv comments.

lib/private/App/AppManager.php
skjnldsv: Make sense! We can keep it as files then since you can't disable files :)

lib/private/App/AppManager.php
Pytal: Does not seem to be updatable from the web UI but added in https://github.com/nextcloud/server/commit/db86bea18ce41ad73e9c1a06f2b2d89d8d4f2ef8 probably for admins to set with
```bash
occ user:setting admin core defaultapp files
```

JuliaKirschenheuter (approved):
tested different scenarios and it works! Thank you

lib/private/App/AppManager.php
skjnldsv: Do we actually support users setting their own default app?
Also my main concern was more about the string itself. What does "defaultApp" refers too? There is no app named like that?

Pytal: > Do we actually support users setting their own default app?

Can only be set by an admin from what I can see

> Also my main concern was more about the string itself. What does "defaultApp" refers too? There is no app named like that?

Probably thought the string was the default fallback? `'defaultapp'` is actually passed as the `key` argument for https://github.com/nextcloud/server/blob/31f4f4117e37849dabbf6b3c83790af878c01119/lib/public/IConfig.php#L196

Issue #36437, opened 2023-01-30
Deprecation and removal of PHP8.0 support


ChristophWurst:
Continuation of https://github.com/nextcloud/server/issues/35025

PHP8.0 will run out of security support end of year.

Some usage stats can be found on https://ncpw.mdns.eu/.
Ref https://www.php.net/supported-versions.php

We should do this in two steps again
* [x] Deprecate the version but still support it + documentation (Nextcloud 27)
* [ ] Drop support with Nextcloud 28 if released before EOL
* [ ] Drop support with Nextcloud 29 if v28 still supported it

cc @AndyScherzinger @juliushaertl @nickvergessen

ChristophWurst:
That is also fine. We can still deprecate with 27 to give admins more time and only make the breaking change with 29.

AndyScherzinger:
So let's do it like that then: deprecate PHP8.0 with v27, v28 last version supporting it (if released before PHP8.0 EOL), drop with v29 (except when 28 already would drop it).

ChristophWurst:
PHP8.0 runs out of security support in five months. With our current release pace we'll probably finish 28 slightly earlier. Thus 8.0 support is rather not to drop in 28 but in 29.

isdnfan:
hi team another heads up - for some reason NC30 still supports PHP 8.0 which is EOL since 26.11.**2023**!! 
please keep in mind PHP 8.1 "*Active support*" ends in less than 2 months on on 1 Jan 2025

nickvergessen:
30 does not support 8.0: https://github.com/nextcloud/server/blob/stable30/lib/versioncheck.php#L9

Also note, that we will keep 8.1 (as it's still security supported) until a dependency bump of non replaceable dependencies like symfony, doctrine dbal or the likes forces us to drop it.

AndyScherzinger:
@isdnfan yes, we decided for v30 to have 8.0 support given that at the time of the release SLES did not have php8.0 support, hence we were forced to keep it, else instances running SLES would have been locked on an older Nextcloud version since the underlying stack could not be upgraded.

isdnfan:
thank you for clarification @nickvergessen and @AndyScherzinger . In fact I was misleaded by unfortunate linking issue from https://docs.nextcloud.com which links NC30 to a "stable* docs branch and NC31 to a "latest" while in fact NC30 is "latest"

[...]

PR #37451, opened 2023-03-28
Deprecate PHP8.0

nfebe:
PHP8.0 would stop receiving security support on 26 Nov 2023 hence this deprecation. Next, is the eventual drop of PHP8.0 before its EOL.

See: https://www.php.net/supported-versions.php

Resolves the first part of #36437

### DIFF
--- a/core/js/setupchecks.js
+++ b/core/js/setupchecks.js
@@ -300,6 +300,14 @@
 							msg: t('core', 'You are currently running PHP {version}. Upgrade your PHP version to take advantage of {linkstart}performance and security updates provided by the PHP Group ↗{linkend} as soon as your distribution supports it.', { version: data.phpSupported.version })
 								.replace('{linkstart}', '<a target="_blank" rel="noreferrer noopener" class="external" href="https://secure.php.net/supported-versions.php">')
 								.replace('{linkend}', '</a>'),
+							type: OC.SetupChecks.MESSAGE_TYPE_WARNING
+						})
+					}
+					if (data.phpSupported && data.phpSupported.version.substr(0, 3) === '8.0') {
+						messages.push({
+							msg: t('core', 'PHP 8.0 is now deprecated in Nextcloud 27. Nextcloud 28 may require at least PHP 8.1. Please upgrade to {linkstart}one of the officially supported PHP versions provided by the PHP Group ↗{linkend} as soon as possible.')
+							.replace('{linkstart}', '<a target="_blank" rel="noreferrer noopener" class="external" href="https://secure.php.net/supported-versions.php">')
+							.replace('{linkend}', '</a>'),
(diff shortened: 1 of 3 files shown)


Fenn-CS:
Hey @ChristophWurst tagging you here for a possible review. Thank you!

ChristophWurst (approved):

ChristophWurst:
Assigning *Pending documentation* because this change should be noted in the admin docs https://docs.nextcloud.com/server/latest/admin_manual/ (https://github.com/nextcloud/documentation/tree/master/admin_manual)

AndyScherzinger (approved):

core/js/setupchecks.js
```diff
@@ -303,6 +303,12 @@
 							type: OC.SetupChecks.MESSAGE_TYPE_INFO
 						})
 					}
+					if (data.phpSupported && data.phpSupported.version.substr(0, 3) === '8.0') {
+						messages.push({
+							msg: t('core', 'PHP 8.0 is now deprecated in Nextcloud 27. Nextcloud 28 may require at least PHP 8.1.'),
```
nickvergessen, line 308: Could also link here to https://secure.php.net/supported-versions.php as in the above check.

That being said, this here is fine as a `MESSAGE_TYPE_INFO` but the above one could be upgraded to a `MESSAGE_TYPE_WARNING` ?

susnux (approved):

core/js/setupchecks.js
Fenn-CS: @nickvergessen Should I add that to this pr? 

Also should I modify this commit or add it as a separate commit?

ChristophWurst: The change belongs to the other commit so add it here and amend the last commit to form an *atomic* change.

Fenn-CS: Now resolved, thanks!

ChristophWurst (approved):

nickvergessen (approved):

come-nc (approved):

Fenn-CS:
Update

This change is now documented in the admin doc via : https://github.com/nextcloud/documentation/pull/9966

cc : @ChristophWurst

AndyScherzinger:
@Fenn-CS see https://github.com/nextcloud/server/actions/runs/4563355952/jobs/8053019695?pr=37451 the test is now failing and needs to be adapted since this PR changes the version check and needs to add this.

Fenn-CS:
Thanks for pointing that out @AndyScherzinger, that is now addressed :)

ChristophWurst (approved):

AndyScherzinger:
Nice, thanks a lot @Fenn-CS 🎉 CHange and doc change merged 🚀

PR #37452, opened 2023-03-28
Add missing interface methods

provokateurin:
## Summary

Psalm 5 shows a bunch of errors because of missing interface methods. Those methods are not on the interfaces because they should not be used by apps, but are used internally. Typing the variables as the implementation makes it work.

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/CachedSubscription.php
+++ b/apps/dav/lib/CalDAV/CachedSubscription.php
@@ -28,7 +28,6 @@ declare(strict_types=1);
 namespace OCA\DAV\CalDAV;
 
 use OCA\DAV\Exception\UnsupportedLimitOnInitialSyncException;
-use Sabre\CalDAV\Backend\BackendInterface;
 use Sabre\DAV\Exception\MethodNotAllowed;
 use Sabre\DAV\Exception\NotFound;
 use Sabre\DAV\INode;
@@ -38,7 +37,7 @@ use Sabre\DAV\PropPatch;
  * Class CachedSubscription
  *
  * @package OCA\DAV\CalDAV
- * @property BackendInterface|CalDavBackend $caldavBackend
+ * @property CalDavBackend $caldavBackend
  */
 class CachedSubscription extends \Sabre\CalDAV\Calendar {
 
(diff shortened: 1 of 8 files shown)


lib/public/Collaboration/Resources/IManager.php
```diff
@@ -79,6 +79,16 @@ public function invalidateAccessCacheForUser(?IUser $user): void;
 	 */
 	public function invalidateAccessCacheForResource(IResource $resource): void;
 
+	/**
+	 * @since 27.0.0
```
nickvergessen, line 83: Deprecated feature *cough*
Also this is intentionally not exposed as apps are not supposed to call this 🤷‍♂️

lib/private/Group/Manager.php
```diff
@@ -418,7 +418,7 @@ public function displayNamesInGroup($gid, $search = '', $limit = -1, $offset = 0
 	/**
 	 * @return \OC\SubAdmin
 	 */
-	public function getSubAdmin() {
+	public function getSubAdmin(): \OC\SubAdmin {
```
nickvergessen, line 421: Should be the interface?!

nickvergessen (dismissed):

lib/public/Collaboration/Resources/IManager.php
provokateurin: You mean I should just type the object as the implementation instead of the interface?

nickvergessen: You can type both `AClass|AnInterface` if necessary

provokateurin: That is already done in some of the places, but it complains that the method is only available on the implementation and not on the interface.

provokateurin: I'd say internally we switch to relying on the implementation but only expose it as the interface. That would als solve some problems with the CalDavBackend and Sabre since our implementation has some optional properties that are not present in the interface.

provokateurin: I did this now

ChristophWurst (approved):
I like it

julien-nc (approved):

provokateurin:
The test failure isn't related, right?

provokateurin:
@nickvergessen Can you remove your requested changes? I addressed them

julien-nc:
If Php comments could break Cypress tests...I would be devastated.
I've seen the drone failure in other PRs, not related IMO as well.

PR #37466, opened 2023-03-29
Refactor OCP\Share\IShare ID to int

provokateurin:
## Summary

There is a lot of confusion in the code about what the type of IShare ID is. In the database it's an integer, so I switched it to int everywhere where it is possible. Sadly this doesn't apply to all places, because sometimes you get the id of a remote share and those have to be handled as strings for compatibility reasons.

I also thought that we could make everything a string and only convert it to an integer for the database queries to remove all the different types and have it only as string. IMO this is the easier solution, but also makes the code more fragile since future devs might not be aware that the ids have to be valid integers although they are typed as string everywhere.

## Checklist

### DIFF
--- a/lib/public/Share/IShare.php
+++ b/lib/public/Share/IShare.php
@@ -142,21 +142,20 @@ interface IShare {
 	 * It is only allowed to set the internal id of a share once.
 	 * Attempts to override the internal id will result in an IllegalIDChangeException
 	 *
-	 * @param string $id
 	 * @return \OCP\Share\IShare
 	 * @throws IllegalIDChangeException
 	 * @throws \InvalidArgumentException
 	 * @since 9.1.0
 	 */
-	public function setId($id);
+	public function setId(int $id);
 
 	/**
 	 * Get the internal id of the share.
 	 *
-	 * @return string
+	 * @return int
 	 * @since 9.0.0
 	 */
-	public function getId();
+	public function getId(): int;
 
 	/**
 	 * Get the full share id. This is the <providerid>:<internalid>.
@@ -166,7 +165,7 @@ interface IShare {
 	 * @since 9.0.0
 	 * @throws \UnexpectedValueException If the fullId could not be constructed
 	 */
-	public function getFullId();
+	public function getFullId(): string;
 
 	/**
 	 * Set the provider id of the share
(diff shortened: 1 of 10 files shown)


lib/public/Share/IShare.php
```diff
@@ -153,7 +153,7 @@ public function setId($id);
 	/**
 	 * Get the internal id of the share.
 	 *
-	 * @return string
+	 * @return int
```
ChristophWurst, line 156: this is a breaking change if the phpdoc is/was correct

provokateurin: I know, how do I document it except for adding it to the 27 changes issue?
I think technically it's not even a breaking change since the casting from/to int/string should always work.

provokateurin: Would you still consider it a breaking change, since it definitely changes the outline of the interface, but it's only a bugfix and has been broken anyway?

ChristophWurst: in the light of your findings it is probably not a breaking change

ChristophWurst (commented):
Good clean-up. Let's enforce the types while we're here.

provokateurin:
Yes, I forgot that. Better do it now

nickvergessen (changes requested):
While I get the point, this again is going to break lot of things down the road.

lib/public/Share/IShare.php
provokateurin: Ah shit I forgot that it parses the int and converts it to a string. So while it was set to int everywhere, it's actually always a string... dead end again lol

nickvergessen: And similar code is also found in apps. So since it's public API it has to stay or be announced 3 years in advance *sigh*

* Talk https://github.com/nextcloud/spreed/blob/5f096c611cf5fa879941e9fd85a9d8ac869d6845/lib/Files/Util.php#L108
* Text https://github.com/nextcloud/text/blob/f8b122502316a02581185add11bc8b9941097ba2/lib/Controller/WorkspaceController.php#L157

provokateurin: I wanna cry :sob:

PR #37467, opened 2023-03-29
Refactor OCP\IConfig and OCP\IAppConfig to add stronger typing

provokateurin:
## Summary

IConfig and IAppConfig use very weak and often wrong and inconsistent typing. This change is completely backwards compatible and only makes the analysis better and the interface definitions clearer.

## Checklist

### DIFF
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -41,6 +41,8 @@ use OCP\IConfig;
 /**
  * This class provides an easy way for apps to store config values in the
  * database.
+ *
+ * @psalm-import-type Value from IConfig
  */
 class AppConfig implements IAppConfig {
 	/** @var array[] */
@@ -160,9 +162,9 @@ class AppConfig implements IAppConfig {
 
 	/**
 	 * @param string $app
-	 * @return array
+	 * @return array<string, Value>
 	 */
-	private function getAppValues($app) {
+	private function getAppValues(string $app): array {
 		$this->loadConfigValues();
 
 		if (isset($this->cache[$app])) {
@@ -175,12 +177,12 @@ class AppConfig implements IAppConfig {
 	/**
 	 * Get all apps using the config
 	 *
-	 * @return array an array of app ids
+	 * @return string[] an array of app ids
 	 *
 	 * This function returns a list of all apps that have at least one
 	 * entry in the appconfig table.
 	 */
-	public function getApps() {
+	public function getApps(): array {
 		$this->loadConfigValues();
 
 		return $this->getSortedKeys($this->cache);
@@ -190,12 +192,12 @@ class AppConfig implements IAppConfig {
 	 * Get the available keys for an app
 	 *
 	 * @param string $app the app we are looking for
-	 * @return array an array of key names
+	 * @return string[] an array of key names
 	 *
 	 * This function gets all keys of an app. Please note that the values are
 	 * not returned.
 	 */
-	public function getKeys($app) {
+	public function getKeys(string $app): array {
 		$this->loadConfigValues();
 
 		if (isset($this->cache[$app])) {
@@ -205,7 +207,7 @@ class AppConfig implements IAppConfig {
 		return [];
 	}
 
-	public function getSortedKeys($data) {
+	public function getSortedKeys($data): array {
 		$keys = array_keys($data);
 		sort($keys);
 		return $keys;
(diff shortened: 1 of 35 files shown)


lib/public/IAppConfig.php
```diff
@@ -49,24 +51,24 @@ public function hasKey($app, $key);
...
 	 */
-	public function getValues($app, $key);
+	public function getValues(mixed $app, mixed $key): mixed;
 
 	/**
 	 * get all values of the app or and filters out sensitive data
 	 *
 	 * @param string $app
-	 * @return array
+	 * @return array<string, Value>
```
nickvergessen, line 60: This is wrong, the appconfig can only store and return strings.

lib/private/AppConfig.php
```diff
@@ -214,15 +216,16 @@ public function getSortedKeys($data) {
...
-	 * @param string $default = null, default value if the key does not exist
-	 * @return string the value or $default
+	 * @param T $default = null, default value if the key does not exist
+	 * @return T the value or $default
 	 *
 	 * This function gets a value from the appconfig table. If the key does
 	 * not exist the default value will be returned
 	 */
-	public function getValue($app, $key, $default = null) {
+	public function getValue(string $app, string $key, mixed $default = null): mixed {
```
nickvergessen, line 228: This will break the guests app:
https://github.com/nextcloud/guests/blob/master/lib/AppConfigOverwrite.php

lib/public/IAppConfig.php
provokateurin: Ok I didn't know that. Then it's definitely not correctly done in some places...

lib/public/IConfig.php
```diff
@@ -127,29 +131,30 @@ public function deleteSystemValue($key);
...
 
 	/**
 	 * Looks up an app wide defined value
 	 *
+	 * @template T of Value
 	 * @param string $appName the appName that we stored the value under
 	 * @param string $key the key of the value, under which it was saved
-	 * @param string $default the default value to be returned if the value isn't set
-	 * @return string the saved value
+	 * @param T $default the default value to be returned if the value isn't set
```
nickvergessen, line 153: From my POV we should only allow string and null. The DB can't store anything else, so the return needs casting anyway (that's also why string is documented and not `mixed`)

lib/public/IConfig.php
```diff
@@ -175,36 +180,37 @@ public function deleteAppValues($appName);
...
-	 * @param string $value the value that you want to store
-	 * @param string $preCondition only update if the config value was previously the value passed as $preCondition
+	 * @param Value $value the value that you want to store
+	 * @param ?string $preCondition only update if the config value was previously the value passed as $preCondition
 	 * @throws \OCP\PreConditionNotMetException if a precondition is specified and is not met
 	 * @throws \UnexpectedValueException when trying to store an unexpected value
 	 * @since 6.0.0 - parameter $precondition was added in 8.0.0
 	 */
-	public function setUserValue($userId, $appName, $key, $value, $preCondition = null);
+	public function setUserValue(string $userId, string $appName, string $key, mixed $value, string $preCondition = null);
```
nickvergessen, line 189:
```suggestion
	public function setUserValue(string $userId, string $appName, string $key, mixed $value, ?string $preCondition = null);
```

nickvergessen: Same problem here, mixed is not needed for value, the DB can only store strings.

nickvergessen (changes requested):
Lot of alarm bells going off on this one.
There is a reason no one tried to ever touch it.

We should not allow certain things to be passed in when that makes a type conversation before writing to the DB.
But enforcing it with a typehint is a breaking change which we need to announce 3 years in advance.
So keeping the stricter type in the `@param` doc sounds like the best solution (and that's what we were doing already).

provokateurin:
OK, then let's only allow strings? Or will that also break apps?

nickvergessen:
> Or will that also break apps?

It will break all apps that do it wrongly but are not strict in type handling, so can't really do that either.

provokateurin:
So this is just a dead end if I understand correctly? Sucks but I guess I'll have to live with that...

nickvergessen:
> So this is just a dead end if I understand correctly?

yes :(

> Sucks but I guess I'll have to live with that...

s/I'll/we'll

Everyone is suffering from it. We could add additionally a typed implementation and deprecate the old one, but it's also a dangerous path if the 2 APIs are not using the same code-path the behaviour oculd diverge at some point unintentionally.

PR #37469, opened 2023-03-29
restore shared lock ttl to previous value when releasing

icewind1991:
With shared locks, each time the lock is acquired, the ttl for the path is reset.

Due to this "ttl extension" when a shared lock isn't freed correctly for any reason
the lock won't expire until no shared locks are required for the path for 1h.
This can lead to a client repeatedly trying to upload a file, and failing forever
because the lock never gets the opportunity to expire.

To help the lock expire in this case, we lower the TTL back to what it was before we
took the shared lock *only* if nobody else got a shared lock after we did.

This doesn't handle all cases where multiple requests are acquiring shared locks
but it should handle some of the more common ones and not hurt things further.

Ideally each copy of a shared lock will have it's own independent TTL so locks can expire while others are still being help, but we don't have that luxury.

To test:

- setup an instance with redis file locking without applying this patch
- set a breakpoint at [acquireLock](https://github.com/nextcloud/server/blob/fcae6a68c347e2913cc29f45648be37789f09c29/lib/private/Lock/MemcacheLockingProvider.php#L60)
- upload a file called `test.txt`
- note the `$path` and `$this->memcache->prefix` used for the lock for `test.txt` (check `$readablePath` to ensure you're look at the right lock)
- remove the breakpoint
- delete `test.txt`
- open a redis-cli and run
  - `set "<$this->memcache->prefix><$path>" 1` (key should be something like `bc2af67f80d0445df3c328c2a0f08faf/lockfiles/0ca47d1e59d0cbf07f715297a70f8417`
  - `expire "<$this->memcache->prefix><$path>" 3600`
- you have now created a "leaked" lock with a TTL of 1 hour
- wait a few seconds and run `expire "<$this->memcache->prefix><$path>"` in redis-cli to check that the TTL is counting down
- try to upload `test.txt` again, this should fail due to the leaked lock
- check the TTL of the leaked lock which should have been reset to ~1h
- apply the patch
- try uploading `test.txt again
- check that the TTL of the leaked lock didn't get reset to 1h

### DIFF
--- a/lib/private/Lock/MemcacheLockingProvider.php
+++ b/lib/private/Lock/MemcacheLockingProvider.php
@@ -27,21 +27,42 @@ declare(strict_types=1);
  */
 namespace OC\Lock;
 
+use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\IMemcache;
 use OCP\IMemcacheTTL;
 use OCP\Lock\LockedException;
 
 class MemcacheLockingProvider extends AbstractLockingProvider {
+	/** @var array<string, array{time: int, ttl: int}> */
+	private array $oldTTLs = [];
+
 	public function __construct(
 		private IMemcache $memcache,
+		private ITimeFactory $timeFactory,
 		int $ttl = 3600,
 	) {
 		parent::__construct($ttl);
 	}
 
-	private function setTTL(string $path): void {
+	private function setTTL(string $path, int $ttl = null, ?int $compare = null): void {
+		if (is_null($ttl)) {
+			$ttl = $this->ttl;
+		}
 		if ($this->memcache instanceof IMemcacheTTL) {
-			$this->memcache->setTTL($path, $this->ttl);
+			if ($compare !== null) {
+				$this->memcache->compareSetTTL($path, $compare, $ttl);
+			} else {
+				$this->memcache->setTTL($path, $ttl);
+			}
+		}
+	}
+
+	private function getTTL(string $path): int {
+		if ($this->memcache instanceof IMemcacheTTL) {
+			$ttl = $this->memcache->getTTL($path);
+			return $ttl === false ? -1 : $ttl;
+		} else {
+			return -1;
 		}
 	}
 
@@ -58,14 +79,22 @@ class MemcacheLockingProvider extends AbstractLockingProvider {
 
 	public function acquireLock(string $path, int $type, ?string $readablePath = null): void {
 		if ($type === self::LOCK_SHARED) {
+			// save the old TTL to for `restoreTTL`
+			$this->oldTTLs[$path] = [
+				"ttl" => $this->getTTL($path),
+				"time" => $this->timeFactory->getTime()
+			];
 			if (!$this->memcache->inc($path)) {
 				throw new LockedException($path, null, $this->getExistingLockForException($path), $readablePath);
 			}
 		} else {
+			// when getting exclusive locks, we know there are no old TTLs to restore
 			$this->memcache->add($path, 0);
+			// ttl is updated automatically when the `set` succeeds
 			if (!$this->memcache->cas($path, 0, 'exclusive')) {
 				throw new LockedException($path, null, $this->getExistingLockForException($path), $readablePath);
 			}
+			unset($this->oldTTLs[$path]);
 		}
 		$this->setTTL($path);
 		$this->markAcquire($path, $type);
@@ -88,6 +117,12 @@ class MemcacheLockingProvider extends AbstractLockingProvider {
 				$newValue = $this->memcache->dec($path);
 			}
 
+			if ($newValue > 0) {
+				$this->restoreTTL($path);
+			} else {
+				unset($this->oldTTLs[$path]);
+			}
+
 			// if we somehow release more locks then exists, reset the lock
 			if ($newValue < 0) {
 				$this->memcache->cad($path, $newValue);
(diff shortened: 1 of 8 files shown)


lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -58,13 +80,20 @@ public function isLocked(string $path, int $type): bool {
...
 	public function acquireLock(string $path, int $type, ?string $readablePath = null): void {
 		if ($type === self::LOCK_SHARED) {
+			// save the old TTL to for `restoreTTL`
+			$this->oldTTLs[$path] = [
+				"ttl" => $this->getTTL($path),
+				"time" => $this->timeFactory->getTime()
+			];
 			if (!$this->memcache->inc($path)) {
-				throw new LockedException($path, null, $this->getExistingLockForException($path), $readablePath);
+				throw new LockedException($path, null, $this->getExistingLockForException($path), $readablePath, $this->getTTL($path));
```
come-nc, line 89: Forgot to commit `LockedException` maybe?

lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -58,13 +80,20 @@ public function isLocked(string $path, int $type): bool {
...
 			}
 		} else {
+			// when getting exclusive locks, we know there are no old TTLs to restore
```
come-nc, line 92: No need to clear `$this->oldTTLs[$path]` ?

lib/private/Memcache/LoggerWrapperCache.php
```diff
@@ -167,10 +167,18 @@ public function cad($key, $old) {
 	}
 
 	/** @inheritDoc */
-	public function setTTL($key, $ttl) {
+	public function setTTL(string $key, int $ttl) {
 		$this->wrappedCache->setTTL($key, $ttl);
 	}
 
+	public function getTTL(string $key) {
+		$this->wrappedCache->getTTL($key);
```
come-nc, line 175:
```suggestion
		return $this->wrappedCache->getTTL($key);
```

lib/private/Memcache/LoggerWrapperCache.php
```diff
@@ -167,10 +167,18 @@ public function cad($key, $old) {
...
+	}
+
+	public function compareSetTTL(string $key, mixed $value, int $ttl): bool {
+		$this->wrappedCache->compareSetTTL($key, $value, $ttl);
```
come-nc, line 179:
```suggestion
		return $this->wrappedCache->compareSetTTL($key, $value, $ttl);
```

come-nc:
I am not sure I understand the problem description.
When uploading test.txt fails, it is because of the existing lock? If so, it means it fails to acquire the lock, no?
If it succeed to acquire the lock and extends it, what fails?

lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -106,13 +139,50 @@ public function changeLock(string $path, int $targetType): void {
...
+			$saved = $this->oldTTLs[$path];
+			$elapsed = $this->timeFactory->getTime() - $saved['time'];
+
+			// old value to compare to when setting ttl in case someone else changes the lock in the middle of this function
+			$value = $this->memcache->get($path);
+
+			$currentTtl = $this->getTTL($path);
+
+			// what the old ttl would be given the time elapsed since we acquired the lock
+			$remainingOldTtl = $saved['ttl'] - $elapsed;
```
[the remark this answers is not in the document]
icewind1991: Yes, but that would just lead to the key being removed when setting the ttl as desired.

Added a comment about it

icewind1991:
> When uploading test.txt fails, it is because of the existing lock?

Yes

> If so, it means it fails to acquire the lock, no?

on upload it first 1) acquires a shared lock, 2) changes it to an exclusive one, 3) changes it back to a shared one
2) fails here, but 1) still extended the TTL

So if a client repeatedly attempts to upload a file that has a leaked shared lock, it will never succeed to change it's shared lock into an exclusive one, but the TTL will never expire.

lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -27,21 +27,43 @@
...
 use OCP\IMemcache;
 use OCP\IMemcacheTTL;
 use OCP\Lock\LockedException;
 
 class MemcacheLockingProvider extends AbstractLockingProvider {
 	private IMemcache $memcache;
+	private ITimeFactory $timeFactory;
+
+	/** @var array<string, array{time: int, ttl: int}> */
+	private $oldTTLs = [];
```
come-nc, line 40:
```suggestion
	private array $oldTTLs = [];
```

lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -88,6 +118,12 @@ public function releaseLock(string $path, int $type): void {
 				$newValue = $this->memcache->dec($path);
 			}
 
+			if ($newValue > 0) {
+				$this->restoreTTL($path);
```
come-nc, line 122: Why is the restoreTTL from changeLock not enough, given the scenario you described?
What is the rationale behind this one in releaseLock? For other kind of exceptions/failures?

lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -27,21 +27,43 @@
...
+	/** @var array<string, array{time: int, ttl: int}> */
+	private $oldTTLs = [];
 
 	public function __construct(IMemcache $memcache, int $ttl = 3600) {
 		$this->memcache = $memcache;
 		$this->ttl = $ttl;
+		$this->timeFactory = \OC::$server->get(ITimeFactory::class);
+	}
+
+	private function setTTL(string $path, int $ttl = null, $compare = null): void {
```
blizzz, line 48:
```suggestion
	private function setTTL(string $path, int $ttl = null, ?int $compare = null): void {
```

blizzz (commented):
Logically it looks okay and I can follow your outlines, but I really cannot pretend to be very knowledgeable here. But contributed some bikeshedding ;)

lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -27,21 +27,43 @@
...
 	private IMemcache $memcache;
+	private ITimeFactory $timeFactory;
+
+	/** @var array<string, array{time: int, ttl: int}> */
+	private $oldTTLs = [];
 
 	public function __construct(IMemcache $memcache, int $ttl = 3600) {
 		$this->memcache = $memcache;
 		$this->ttl = $ttl;
+		$this->timeFactory = \OC::$server->get(ITimeFactory::class);
```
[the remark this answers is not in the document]
icewind1991: lazyness :see_no_evil:

lib/private/Lock/MemcacheLockingProvider.php
icewind1991: We still want to restore the ttl when a request just acquires + releases.

come-nc (approved):

skjnldsv (approved):

PR #37470, opened 2023-03-29
implemented Sorting of  shares by title and then creation date

SamantaTarun:
* Resolves: # 
implemented Sorting of  shares by title and then creation date
#36933 
## TODO

- [ ] ...

## Checklist

- [X] [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits

### DIFF
--- a/apps/files_sharing/src/views/SharingTab.vue
+++ b/apps/files_sharing/src/views/SharingTab.vue
@@ -246,7 +246,7 @@ export default {
 				this.$set(this.sharedWithMe, 'subtitle', t('files_sharing', 'this share just expired.'))
 			}
 		},
-
+		// .sort((a, b) => b.createdTime - a.createdTime)
 		/**
 		 * Process the current shares data
 		 * and init shares[]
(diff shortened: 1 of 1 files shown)


artonge (commented):
Not sure if we want that as the initial sort logic seem on purpose. @nextcloud/designers any thought on it ?
@tarunsamanta2k20 Also, the screenshot was not uploaded correctly ;)

tarunsamanta2k20:
@artonge i will upload the screen shot again in some time. I have tested it it's working fine.

tarunsamanta2k20:
@skjnldsv

> Not sure if we want that as the initial sort logic seem on purpose. @nextcloud/designers any thought on it ? @tarunsamanta2k20 Also, the screenshot was not uploaded correctly ;)

@artonge let me upload the screenshot.

jancborchardt:
Sorry, what's the short summary of the change and the use-case?

The sorting should be the same across the different views, just like in a desktop file manager. So if "All" are sorted by name, "Shared" should be too. Same for "Modified" etc.

PR #37476, opened 2023-03-30
ci(checkers): Fix autoloader (re)generation for extraneous apps

ChristophWurst:
* Resolves: # 

## Summary

Running ``./build/autoloaderchecker.sh`` should only touch apps that are part of this repo.

## TODO

- [x] Fix the script

## Checklist

### DIFF
--- a/build/autoloaderchecker.sh
+++ b/build/autoloaderchecker.sh
@@ -22,6 +22,11 @@ echo "Regenerating main autoloader"
 $COMPOSER_COMMAND dump-autoload -d $REPODIR
 
 for app in ${REPODIR}/apps/*; do
+	if git check-ignore ${app} -q ; then
+		echo
+		echo "${app} is not shipped. Ignoring autoloader regeneration"
+		continue
+	fi
     if [[ -d $app ]]; then
         if [[ -e ${app}/composer/composer.json ]]; then
             echo


build/autoloaderchecker.sh
```diff
@@ -22,6 +22,12 @@ echo "Regenerating main autoloader"
 $COMPOSER_COMMAND dump-autoload -d $REPODIR
 
 for app in ${REPODIR}/apps/*; do
+	git check-ignore ${app} -q
+	if git check-ignore ${app} -q ; then
```
come-nc, line 26: Why call the command twice?
```suggestion
	if git check-ignore ${app} -q ; then
```

come-nc (commented):
Any chance to make it stop creating a huge diff when there are no change in files?

ChristophWurst:
> Any chance to make it stop creating a huge diff when there are no change in files?

Composer files are not designed to be commited to git. I can't think of a way to disable any diffs for the generated autoloader code.

If we remove the autoloaders from git and generate them at dev and release time we circumvent the problem.

come-nc (approved):

andrey18106 (approved):

nickvergessen (approved):

go2sh (approved):

build/autoloaderchecker.sh
```diff
@@ -22,6 +22,11 @@ echo "Regenerating main autoloader"
 $COMPOSER_COMMAND dump-autoload -d $REPODIR
 
 for app in ${REPODIR}/apps/*; do
+	if git check-ignore ${app} -q ; then
```
go2sh, line 25: Should it be the other way arround? git check-ignore returns 0 if the file is ignored. This will only non ignored files will be skipped like this.

nickvergessen:
Btw this could also replace my local patch for `build/image-optimization.sh` right:
```diff
diff --git a/build/image-optimization.sh b/build/image-optimization.sh
index e559d8552f6..99ac9048e1a 100755
--- a/build/image-optimization.sh
+++ b/build/image-optimization.sh
@@ -33,6 +33,22 @@ function recursive_optimize_images() {
                return
        elif [[ "$DIR_NAME" == "tests" ]]; then
                return
+       elif [[ "$DIR_NAME" == "appsmine" ]]; then
+               return
+       elif [[ "$DIR_NAME" == "appscommunity" ]]; then
+               return
+       elif [[ "$DIR_NAME" == "appsmarketing" ]]; then
+               return
+       elif [[ "$DIR_NAME" == "appspackaged" ]]; then
+               return
+       elif [[ "$DIR_NAME" == "appssubscriptions" ]]; then
+               return
+       elif [[ "$DIR_NAME" == "appssupported" ]]; then
+               return
+       elif [[ "$DIR_NAME" == "appszips" ]]; then
+               return
+       elif [[ "$DIR_NAME" == "apps-action-templates" ]]; then
+               return
        fi
 
        # Optimize all PNGs
```

But I guess you don't want to add that here, do you?

Issue #32117, opened 2022-04-25
Changes for developers & admins for Nextcloud 25


nickvergessen:
Continuation of #29914

# Front-end

* Theming redone: https://github.com/nextcloud/server/issues/32060
* `body.theme--dark` classes removed. See above
* SCSS support removed! :warning: https://github.com/nextcloud/server/issues/32060
  https://help.nextcloud.com/t/scss-support-dropped-with-nextcloud-25/
* Files app ID changes in the DOM, to remove duplicate IDs for accessibility:
    * https://github.com/nextcloud/server/pull/33320#issuecomment-1195311502
    * https://github.com/nextcloud/server/pull/33373
    * https://github.com/nextcloud/server/pull/33379
* Non-Vue dialogs will stop working if triggered on top of Vue dialogs because of focus trap, example here: https://github.com/nextcloud/server/issues/33675.
* We changed the container layout to adapt to the design changes with Nextcloud 25
    * The body is no longer the main scroll area but instead the app content
    * https://github.com/nextcloud/server/pull/33568
    * Scroll listeners need to be adapted to watch for the scroll on the NcAppContent container instead of the main document
    * If you are using sticky headers in your app you might need to adjust the css top property as the app content container is now the scrollable element. For any contained sticky that had a header height top spacing before you should be able to set that to 0
    * If `@nextcloud/vue` components are used you'll need to update to the major release 7 of the library
        * Migration help for that can be found in https://github.com/nextcloud/nextcloud-vue/blob/master/CHANGELOG.md#v700-beta0-2022-09-01
        * v6.0.0 included a rename of the components to be prefixed with `Nc` which is something that would also be required if not already done
* A global option to disable keyboard shortcuts was added to the accessibility settings. Since it heavily depends on the screenreader and tools that you use if Ctrl and/or Alt or other things are okay to use or not and maintaining a more detailed list is too much effort, we went for a global on/off switch. Apps can use this public javascript API call to determine whether the user used the opt-out: `OCP.Accessibility.disableKeyboardShortcuts()`. If that is the case, no additional shortcuts shall be registered by any app. Only `space` to toggle checkboxes and `enter` to submit the currently active buttons/links are okay: https://github.com/nextcloud/server/pull/34081

# Backend

[...]

PR #37479, opened 2023-03-30
[stable25] dispatch BeforeUserLoggedInEvent

ArtificialOwl:
backport from #36883 (failed at #37071)


skjnldsv (approved):

artonge (approved):

lib/public/User/Events/BeforeUserLoggedInEvent.php
```diff
@@ -59,8 +66,19 @@ public function getUsername(): string {
 
 	/**
 	 * @since 18.0.0
+	 * @since 26.0.0 value can be null
```
nickvergessen, line 69: 25.0.6

Also that can break things. Should at least bei documented in https://github.com/nextcloud/server/issues/32117

PR #37495, opened 2023-03-30
Trim the user/email provided for password resets

### LINKED ISSUE #37408: UX: Forgot password form should trim user input
### How to use GitHub

Frequently users complain that they never received a password reset email. They are certain that the email is not in their spam folder and that the email address or user name they entered was correct.

The logs show that in such cases there is usually additional whitespace in the input, for example " alice" or "bob@example.com ".

Leading and trailing whitespace from the input should be trimmed and ignored like so:

- " alice" -> "alice"
- "bob@example.com " -> "bob@example.com"

There could be a note in the copy that it's the user's responsibility to check for whitespace... But that would probably be even more effort than just trimming in the first place.

**Additional comments**
The copy after submitting the form is misleading (actually it's a lie):

> A password reset message has been sent to the email address of this account.

I understand that for security reasons, it should not be revealed whether this accounts exists or not, but the copy should reflect that limitation. Something along the lines of "If this account exists, an email will be sent [...]".

joshtrichards:
* Resolves: #37408 

## Summary

Trims the username/email address provided by the user when requesting a lost password reset. This reduces support requests from users that complain about never receiving password reset emails (because we silently - from the user perspective - dropped the request as being a bogus username/email address).

Since this is a workflow where few indicators are provided to the user (to avoid security information leakage), it seems worthwhile to handle this for the user to catch this common and easy/no-risk scenario. We know with certainty that whitespaces are never allowed at the start/end of usernames so this is safe.

## TODO

(nothing)

## Checklist

### DIFF
--- a/core/Controller/LostController.php
+++ b/core/Controller/LostController.php
@@ -200,6 +200,8 @@ class LostController extends Controller {
 			return new JSONResponse($this->error($this->l10n->t('Password reset is disabled')));
 		}
 
+		$user = trim($user);
+		
 		\OCP\Util::emitHook(
 			'\OCA\Files_Sharing\API\Server2Server',
 			'preLoginNameUsedAsUserName',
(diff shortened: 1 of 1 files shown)


come-nc (commented):
I think it should be trimed before being sent to the hook then, no?

jtrees:
I am unfamiliar with the Nextcloud code so I cannot review this but I want to thank you for this PR.

joshtrichards:
> I think it should be trimed before being sent to the hook then, no?

Indeed @come-nc. Fixed

come-nc (approved):

blizzz (approved):

PR #37505, opened 2023-03-31
Set recommended psalm options

provokateurin:
## Summary

Psalm 5 tells you that you should set those options because psalm 6 will enforce them.

## Checklist

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -6,6 +6,8 @@
 	xmlns="https://getpsalm.org/schema/config"
 	xsi:schemaLocation="https://getpsalm.org/schema/config"
     errorBaseline="build/psalm-baseline.xml"
+	findUnusedBaselineEntry="true"
+	findUnusedCode="false"
 >
 	<plugins>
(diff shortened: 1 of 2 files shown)


psalm.xml
```diff
@@ -6,6 +6,8 @@
 	xmlns="https://getpsalm.org/schema/config"
 	xsi:schemaLocation="https://getpsalm.org/schema/config"
     errorBaseline="build/psalm-baseline.xml"
+	findUnusedBaselineEntry="false"
```
nickvergessen, line 9: We could set this to true?

provokateurin: Yeah, I didn't because it introduced a bunch of changes in my original PR. Will do it here though

provokateurin: Ah it has basically no effect for us, since we always update the baseline. I'll still enable it, since it can find some problems if there are any.

nickvergessen (approved):

ChristophWurst (approved):

provokateurin:
@nickvergessen @ChristophWurst Can I get a force-merge, please?

PR #37506, opened 2023-03-31
Drop IEventListener template parameter

provokateurin:
## Summary

## Checklist

### DIFF
--- a/lib/public/EventDispatcher/IEventListener.php
+++ b/lib/public/EventDispatcher/IEventListener.php
@@ -27,13 +27,10 @@ namespace OCP\EventDispatcher;
 
 /**
  * @since 17.0.0
- *
- * @template T of Event
  */
 interface IEventListener {
 	/**
 	 * @param Event $event
-	 * @psalm-param T $event
 	 *
 	 * @since 17.0.0
 	 */
--- a/apps/dav/lib/Listener/CalendarDeletionDefaultUpdaterListener.php
+++ b/apps/dav/lib/Listener/CalendarDeletionDefaultUpdaterListener.php
@@ -33,9 +33,6 @@ use Psr\Log\LoggerInterface;
 use Throwable;
 use function strpos;
 
-/**
- * @template-implements IEventListener<\OCA\DAV\Events\CalendarDeletedEvent>
- */
 class CalendarDeletionDefaultUpdaterListener implements IEventListener {
 
 	/** @var IConfig */
--- a/apps/settings/lib/Listener/AppPasswordCreatedActivityListener.php
+++ b/apps/settings/lib/Listener/AppPasswordCreatedActivityListener.php
@@ -34,9 +34,6 @@ use OCP\EventDispatcher\IEventListener;
 use OCP\IUserSession;
 use Psr\Log\LoggerInterface;
 
-/**
- * @template-implements IEventListener<\OC\Authentication\Events\AppPasswordCreatedEvent>
- */
 class AppPasswordCreatedActivityListener implements IEventListener {
 	/** @var IActivityManager */
 	private $activityManager;
--- a/lib/private/Accounts/Hooks.php
+++ b/lib/private/Accounts/Hooks.php
@@ -32,9 +32,6 @@ use OCP\IUser;
 use OCP\User\Events\UserChangedEvent;
 use Psr\Log\LoggerInterface;
 
-/**
- * @template-implements IEventListener<UserChangedEvent>
- */
 class Hooks implements IEventListener {
 	/** @var IAccountManager */
 	private $accountManager;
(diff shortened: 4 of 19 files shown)


lib/public/EventDispatcher/IEventListener.php
```diff
@@ -29,11 +29,11 @@
  * @since 17.0.0
  *
  * @template T of Event
+ * @template-covariant T
```
ChristophWurst, line 32: if you title your PR *Add type hints for all event listeners* please document these changes. otherwise it feels like this is sneaked in.

we had a bunch of issues with the typing of the event listeners so this has to be adjusted carefully to not break static analysis for all apps.

ChristophWurst: https://psalm.dev/docs/annotating_code/templated_annotations/#template-covariance for details

provokateurin: Ah right, I forgot that this then makes it mandatory to add the template parameter... Should I document it in the changes for 27 issue?

ChristophWurst: And I think making the template covariant but dropping it form the method argument erases the templating of the class

before: https://psalm.dev/r/feb99f7e5e
after: https://psalm.dev/r/dae1a0e06f

The purpose of the template is to tell static analysis what type of event is passed to the listener.

https://psalm.dev/r/b991b00aea becomes impossible.

provokateurin: Ah yes, so this isn't even a change from this patch, but when upgrading to psalm 5.

provokateurin: I removed the template from the method, because i thought than the check using `instanceof` would break because it would look like it's not possible to put any other event in the method anyway.

provokateurin: This is how it works in a real event listener: https://psalm.dev/r/c4754d62ca

ChristophWurst: Yeah, we added the `!($event instanceof LoginEvent)` often because PHPStorm didn't/doesn't understand the class templates.

provokateurin: And if you keep the template parameter on the method you get an error: https://psalm.dev/r/9c451952fa

provokateurin: So I think this change is just fine. The covariant is required so that also subclasses of `Event` can be passed as the template parameter.

provokateurin: Sorry, I thought it was resolved.

I think the template adds very little benefit, since the code still has to check for the right event anyway. I'd be up for dropping if that is fine.

come-nc:
See https://github.com/vimeo/psalm/issues/7549

provokateurin:
I see, so the type hint will actually be useful?

come-nc:
> I see, so the type hint will actually be useful?

Hopefully some day. I never found a way to make it work properly.

provokateurin:
Hm so we either keep it and the changes in this PR to make it work in the future, or we completely drop it now since it doesn't work. I'd go with the second option, what do you think @come-nc @ChristophWurst ?

come-nc:
> Hm so we either keep it and the changes in this PR to make it work in the future, or we completely drop it now since it doesn't work. I'd go with the second option, what do you think @come-nc @ChristophWurst ?

What happens if you keep your changes except for the ones in lib/public/EventDispatcher/IEventListener.php ?

provokateurin:
I think that'll work and have no effect as expected.

lib/public/EventDispatcher/IEventListener.php
provokateurin: Ok, so tried a few things around:
- We need the covariant, otherwise the template parameters don't work since they extend the base Event class
- We need to drop the `T` type from the method, because psalm complains: `InvalidTemplateParam: Template param T of OCP\EventDispatcher\IEventListener is marked covariant and cannot be used here`

Alternatively to the second requirement we could also choose to ignore the error, because it fixes all the analysis everywhere else, but I don't know if that is a good thing to do.

provokateurin: @ChristophWurst what do you think?

We could also drop the template parameter altogether, since it doesn't do anything right now due to the bug mentioned by @come-nc.

ChristophWurst: Then let's drop it and carefully document the change. Apps extending the template class right now will run into static analysis errors.

provokateurin:
@come-nc @ChristophWurst please review again.

PR #37507, opened 2023-03-31
Fix types for reading and writing config values

provokateurin:
## Summary

## Checklist


apps/user_ldap/lib/User/User.php
```diff
@@ -303,7 +303,7 @@ public function getHomePath($valueFromLDAP = null) {
 		}
 
 		if (!is_null($attr)
-			&& $this->config->getAppValue('user_ldap', 'enforce_home_folder_naming_rule', true)
+			&& (bool)$this->config->getAppValue('user_ldap', 'enforce_home_folder_naming_rule', 'true')
```
nickvergessen, line 306: && automatically casts to bool.
On that note, what is the config string value when this is "off"? Since string `'false'` is boolean true

provokateurin: I thought it might not cast to bool and take it as a string and would only be false if the string was empty or null.

nickvergessen (approved):

ChristophWurst (approved):

come-nc (approved):

PR #37511, opened 2023-03-31
Add type hints for mappers

provokateurin:
## Summary

## Checklist


lib/private/Updater/ChangesMapper.php
```diff
@@ -31,6 +31,9 @@
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
 
+/**
+ * @template-extends QBMapper<ChangesResult>
```
nickvergessen, line 35: Since the class name is missing from the constructor `ChangesResult` is not used but accoring to the parent constructor
https://github.com/nextcloud/server/blob/f5c361cf44739058b79f322576a1bad2d8c142d9/lib/public/AppFramework/Db/QBMapper.php#L70
`QBMapper<Changes>` would be correct but does not exist. I guess this mapper is not even working?!

lib/public/AppFramework/Db/QBMapper.php
```diff
@@ -55,8 +55,7 @@ abstract class QBMapper {
 	/**
 	 * @param IDBConnection $db Instance of the Db abstraction layer
 	 * @param string $tableName the name of the table. set this to allow entity
-	 * @param string|null $entityClass the name of the entity that the sql should be
-	 * @psalm-param class-string<T>|null $entityClass the name of the entity that the sql should be
+	 * @param class-string<T>|null $entityClass the name of the entity that the sql should be
```
ChristophWurst, line 58: sneaky

ChristophWurst: do we have to use `@param` only? I don't mind having both if it allows older/legacy tools to process `@param string` as long as Psalm only follows to the `@psalm-param`.

Does Psalm 5 merge the two annotations into `string|class-string<T>|null`?

lib/private/Updater/ChangesMapper.php
ChristophWurst: It works because `\OC\Updater\ChangesMapper::getChanges` is the only used method and it calls `\OCP\AppFramework\Db\Entity::fromRow`.

ChristophWurst: In other workds: the mapper doesn't rely on any of the parent logic. It would break as soon as `\OCP\AppFramework\Db\QBMapper::findEntities` is used.

lib/public/AppFramework/Db/QBMapper.php
provokateurin: Yeah, so I had to remove the string type. I don't know which older/legacy tools you mean? I think as long as psalm understands it it is fine since that's the tool we use to validate our code, right?

ChristophWurst: https://nextcloud-server.netlify.app/classes/ocp-appframework-db-qbmapper for example but it seems ok with the enhanced type.

ChristophWurst (approved):

lib/private/Updater/ChangesMapper.php
come-nc: It’s not in OCP, might as well rename it?

come-nc (approved):

provokateurin:
@come-nc @ChristophWurst Can I get a force-merge, please?

lib/private/Updater/ChangesMapper.php
ChristophWurst: @provokateurin can we do that before someone runs into surprises? Before you changes the mapper was untyped, so all methods take/return `Entity`. Now it wrongly advertises that the methods work with `ChangesResult`.

miaulalala (approved):

lib/private/Updater/ChangesMapper.php
provokateurin: @come-nc Should I rename the mapper or the data class?
@ChristophWurst Do you suggest changing the templated type to `Entity` or changing the return type of `getChanges`?

come-nc: > @come-nc Should I rename the mapper or the data class? 

I don’t know, whichever is more logical, or just rename the one that causes the less troubles in the rest of the code.

ChristophWurst (approved):
ChangesMapper/Changes looks good now

PR #37517, opened 2023-03-31
Create semantic list for comments in comments tab

### LINKED ISSUE #36919: [BITV] 9.1.3.1b/3.3 - "Details" section / "Comments" tab - The list of comments represents a list, but it was not marked as such with a list item using HTML or ARIA. (2)
- [ ] Fix
- [ ] Backport

92135bee115f2a82a05a26cbe01ccb03

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-1d5eadb58d-v3-n3

</details>

- [ ] 9\.1.3.1b/10.3 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-1d5eadb58d-v10-n3

JuliaKirschenheuter:
* Resolves: https://github.com/nextcloud/server/issues/36919

## Summary

Create semantic list for comments in comments tab. No visual changes.

| Screenshot from 2023-03-31 11-35-40 | Screenshot from 2023-03-31 11-34-37

## Checklist


apps/comments/src/components/Comment.vue
```diff
@@ -162,6 +162,11 @@ export default {
 			type: Function,
 			required: true,
 		},
+
+		isList: {
+			type: Boolean,
+			default: false,
+		}
```
ShGKme, line 169: `isList` name seems to be a bit unclean. A "Comment" is never a list. With `isList == true` it renders as a list item (`li`), not a list.

If I'm not mistaken, this component is used only in one place - in the `views/Components.vue` and we could change the root element's tag from `div` to `li` without new props.

In the case of better reusability of the component, if we want to add a prop, we may add a `tag` prop (with `div` as the default value) and use it directly in the root `<component :is="tag"`. Then component's user may provide any semantic tag/component for root element, including `li`. Also in the component usage, a developer would implicitly see that it renders as a list item:

```html
<ul>
  <Comment v-for="comment in comments"
    tag="li"
```

ShGKme (commented):
Don't forget to fix ESLint errors

apps/comments/src/components/Comment.vue
JuliaKirschenheuter: Thank you!
Sure, i'm agree with your point about `tag`. 
I'm not sure it is a good idea to make a list at all because sometimes `Comment` is used as stadalone and not as a list, like https://github.com/nextcloud/server/blob/a1aa4491acac6217061a1cecc8cb9d807292376c/apps/comments/src/views/Comments.vue#L27

Could i leave this solution with `<component :is="tag"`?

ShGKme: > I'm not sure it is a good idea to make a list at all because sometimes Comment is used as stadalone and not as a list, like

Oh, sorry, I missed this usage.

> Could i leave this solution with <component :is="tag"?

Yes, but `tag` prop should have `div` as default value

JuliaKirschenheuter: Thank you!

apps/comments/src/views/Comments.vue
```diff
@@ -40,18 +40,19 @@
...
+					:key="comment.props.id"
+					v-bind="comment.props"
+					:auto-complete="autoComplete"
+					:message.sync="comment.props.message"
+					:ressource-id="ressourceId"
+					:user-data="genMentionsData(comment.props.mentions)"
+					class="comments__list"
+					tag="li"
+					@delete="onDelete" />
+			</ul>
```
ShGKme, line 55: I'd put the `tag` prop to the top of the attrs list, before the `key` attr. Then it would be more visually visible, that `Comment` is a `li`

ShGKme (commented):
Tested locally, no visual changes, fixes the issue. But one ESLint error left.

apps/comments/src/views/Comments.vue
JuliaKirschenheuter: ok, thank you! I've put `tag` prop right after `key` because otherwise it is a next eslint error.

artonge (approved):

ShGKme (approved):
🦖

Pytal (approved):

PR #37542, opened 2023-04-03
feat(security): Allow to opt-out of ratelimit protection, e.g. for te…

nickvergessen:
…sting on CI

* Resolves: # 

## Summary

On talk integration tests are red on CI since middle of last week (Merge of https://github.com/nextcloud/server/pull/37497 ). Turns out it's the rate limiting hitting in disallowing to set the user's display name. See https://github.com/nextcloud/spreed/pull/9233 for debugging history.

## Checklist

### DIFF
--- a/lib/private/Security/RateLimiting/Backend/DatabaseBackend.php
+++ b/lib/private/Security/RateLimiting/Backend/DatabaseBackend.php
@@ -3,8 +3,10 @@
 declare(strict_types=1);
 
 /**
+ * @copyright Copyright (c) 2023 Joas Schilling <coding@schilljs.com>
  * @copyright Copyright (c) 2021 Lukas Reschke <lukas@statuscode.ch>
  *
+ * @author Joas Schilling <coding@schilljs.com>
  * @author Lukas Reschke <lukas@statuscode.ch>
  *
  * @license GNU AGPL version 3 or any later version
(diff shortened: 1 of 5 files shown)


ChristophWurst (approved):

lib/private/Security/RateLimiting/Backend/DatabaseBackend.php
```diff
@@ -115,7 +118,12 @@ public function registerAttempt(string $methodIdentifier,
 			->values([
 				'hash' => $qb->createNamedParameter($identifier, IQueryBuilder::PARAM_STR),
 				'delete_after' => $qb->createNamedParameter($deleteAfter, IQueryBuilder::PARAM_DATE),
-			])
-			->executeStatement();
+			]);
+
+		if (!$this->config->getSystemValueBool('ratelimit.protection.enabled', true)) {
```
juliushaertl, line 123: Any reason to not return early at the beginning of the method? Would feel easier to read to me.

juliusknorr (approved):

lib/private/Security/RateLimiting/Backend/DatabaseBackend.php
nickvergessen: I'm drown apart between:
* keep code as simple as possible (return early)
* Keep code as close to production as possible as soon every dev will have this active locally

PR #37575, opened 2023-04-04
Log a warning when a user or group is mapped with a suffix

come-nc:
Most of the time, this is the result of a misconfiguration by the admin
 and is not what is expected.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Access.php
+++ b/apps/user_ldap/lib/Access.php
@@ -589,6 +589,15 @@ class Access extends LDAPUtility {
 		$altName = $this->createAltInternalOwnCloudName($intName, $isUser);
 		if (is_string($altName)) {
 			if ($this->mapAndAnnounceIfApplicable($mapper, $fdn, $altName, $uuid, $isUser)) {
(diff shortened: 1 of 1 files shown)


blizzz (approved):

apps/user_ldap/lib/Access.php
```diff
@@ -589,6 +589,15 @@ public function dn2ocname($fdn, $ldapName = null, $isUser = true, &$newlyMapped
 		$altName = $this->createAltInternalOwnCloudName($intName, $isUser);
 		if (is_string($altName)) {
 			if ($this->mapAndAnnounceIfApplicable($mapper, $fdn, $altName, $uuid, $isUser)) {
+				$this->logger->warning(
```
szaimen, line 592: Should that be a warning or would info be enough?

come-nc: I prefer a warning so that it is not missed. It is not an error per se, but it is unexpected in most cases so it makes sense to warn the admin.

szaimen (approved):
🐘

PR #37617, opened 2023-04-06
handle not being able to write file for notify self-test

icewind1991:
And better handle errors in `file_put_contents`


szaimen (approved):
LGTM but didnt test

blizzz (approved):

Issue #37627, opened 2023-04-06
follow-up to files_trashbin to vue migration


szaimen:
Follow-ups to https://github.com/nextcloud/server/pull/36534

- [ ] Improve loading indicator for actions
- [ ] Deepen virtual scrolling performances
- [ ] Implement keyboard arrow/pageUp-Down navigation and focus management
- [ ] Hide some columns on small screens

- [ ] Switching back to other file list does not work once entered the trash bin once
- [ ] Restore/delete permanently a file does not remove it from the list
- [ ] Grid view is gone
- [ ] Breadcrumbs show `Foldername.dTIMESTAMP` when browsing a deleted folder, in the old implementation we just showed `Foldername

- [ ] right-click does not work anymore
- [ ] the mobile view is currently completely broken (but already in the top post as I can see)
- [ ] clicking on a file in the trashbin downloads the file but with incorrect mimetype ending

szaimen:
@skjnldsv I am not sure if it is related to the trashbin migration but uploading any folder into the files app now results in conflicts and a `NOTFOUNDEXCEPTIONS` files getting created multiple times: 

<details>
<summary>Click to show</summary>

https://user-images.githubusercontent.com/42591237/231138721-08fc9f8d-258c-4c0a-884f-245260d68694.mp4

</details>

joshtrichards:
This is great. Glad to see the trash bin getting some love (hah).

Question: Is there a reason Download is the new default behavior for clicking the filename rather than opening the file?

Seems like Download would just be another option like Restore and Delete. Whereas clicking on filename itself would open the file. Would address #19353 as well as be more consistent with Files UI behavior / experience.

skjnldsv:
@joshtrichards the trashbin never had any default action. Clicking did nothing with 25 and earlier versions.

joshtrichards:
Definitely an improvement. I could have worded my query better. I wasn't suggesting it was a regression or change. I get clicking on filename in the trashbin was a no-op before (so download is an improvement).

But in Files, clicking a filename typically opens a file for viewing (well, or editing) rather than downloading. I was trying to understand if it was a deliberate choice to do Download as the default action in the trashbin rather than viewer/edit mode like in Files generally.

skjnldsv:
All good :)
https://github.com/nextcloud/server/issues/1071 would be the proper ticket to wait for then :+1:

skjnldsv:
https://github.com/nextcloud/server/issues/39914

PR #37633, opened 2023-04-07
Trashbin followup fixes

skjnldsv:
1. Fix action store update (deleting/restore updates the list automatically, it was missing an update of the `@nextcloud/files` library)
2. Fix trashbin download filename header
3. Fix breadcrumbs displayname
  image
5. Fix loading state when executing multiple actions:
   - When executing a single action on a row of a selected file, the batch actions should temporarily be disabled
   - When executing an action on a selection, the individual action be disabled

| 1 | 2 |
|:---------:|:------:|
|Peek 07-04-2023 09-29|Peek 07-04-2023 09-30|

Reference #37627 #36534

There is [some php to review too](https://github.com/nextcloud/server/pull/37633/commits/d6a5fbad5837f10721558a1e849e227dfae29020) :wink: @icewind1991


apps/files/src/store/files.ts
```diff
@@ -81,34 +81,26 @@ export const useFilesStore = () => {
...
 		subscribe('files:file:deleted', fileStore.onDeletedNode)
-		subscribe('files:file:moved', fileStore.onMovedNode)
+		// subscribe('files:file:moved', fileStore.onMovedNode)
 		// subscribe('files:file:updated', fileStore.onUpdatedNode)
 
-		subscribe('files:folder:created', fileStore.onCreatedNode)
+		// subscribe('files:folder:created', fileStore.onCreatedNode)
 		subscribe('files:folder:deleted', fileStore.onDeletedNode)
-		subscribe('files:folder:moved', fileStore.onMovedNode)
+		// subscribe('files:folder:moved', fileStore.onMovedNode)
```
Pytal, line 100: Remove or to be uncommented?

Pytal (approved):

apps/files/src/store/files.ts
skjnldsv: I'll leave them here for reference until they're all used :)

szaimen (approved):
Tested and works but didnt review the code

apps/files/src/components/FilesListHeaderActions.vue
```diff
@@ -122,9 +126,16 @@ export default Vue.extend({
 			const displayName = action.displayName(this.nodes, this.currentView)
 			const selectionIds = this.selectedNodes
 			try {
+				// Set loading markers
 				this.loading = action.id
+				this.nodes.forEach(node => {
+					Vue.set(node, '_loading', true)
+				})
+
```
artonge, line 134: Would it be better to save this info in the component's data instead of hijacking the nodes ?
It would also improve the perf of `areSomeNodesLoading` that would become: `this.loadingNodes.length > 0`

artonge (approved):

apps/files/src/components/FilesListHeaderActions.vue
artonge: Yes, I missed that :)
Having this info in a shared store would be better then. Else we will end up with unspecified properties

PR #37639, opened 2023-04-07
feat(files): implement shift-select

### LINKED ISSUE #3484: 11.0.1 shift click bulk select odd behaviour
When clicking on a file, holding shift and clicking further down, it always tries to select from the very top.
Similarly when clicking on a file low down, holding shift and clicking a file higher up, it selects from the top to the file rather than from the bottom.

Here are examples:
https://youtu.be/bw3aCo9QSbU

https://youtu.be/ngwYFH70YcA

### Expected behaviour
It should simply select from the first file/folder clicked, to the last.

### Server configuration
**Operating system**: Ubuntu 16.04.2

**Web server:** Apache

**Database:** Maria

**PHP version:** 7

**Nextcloud version:** 11.0.1

**Updated from an older Nextcloud/ownCloud or fresh install:** Update

skjnldsv:
- When selecting a file, we store its reference
- When using Shift+select, we use the previous reference to select all files in between
- If the clicked file is already selected, we update the selection from the saved reference

https://user-images.githubusercontent.com/14975046/230588746-45c07d46-bc84-4274-a0cc-ec74b36b20d5.mp4

____

Fix #3484 
cc @jancborchardt @nimishavijay for UX
PS: I'll wait before compiling assets to avoid rebasing


skjnldsv:
An additional question for the design team, do we still want the "range select" option ?
It was implemented to mitigate the fact that we didn't support the shift+select option, correct? Do we still need it?

apps/files/src/store/keyboard.ts
```diff
@@ -0,0 +1,62 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+/* eslint-disable */
```
Pytal, line 22:
```suggestion

```

apps/files/src/store/keyboard.ts
```diff
@@ -0,0 +1,62 @@
...
+		state: () => ({			
+			altKey: false,
+			ctrlKey: false,
+			metaKey: false,
+			shiftKey: false,
+		}),
+
+		actions: {
+			onEvent(event: MouseEvent | KeyboardEvent) {
+				if (!event) event = window.event as MouseEvent | KeyboardEvent
```
Pytal, line 42:
```suggestion
				if (!event) {
					event = window.event as MouseEvent | KeyboardEvent
				}
```

apps/files/src/store/paths.ts
```diff
@@ -25,14 +25,15 @@ import type { PathOptions, ServicesState } from '../types.ts'
...
+import { FileId } from '../types'
 
 export const usePathsStore = () => {
 	const store = defineStore('paths', {
 		state: (): ServicesState => ({}),
 
 		getters: {
 			getPath: (state) => {
-				return (service: string, path: string): number|undefined => {
+				return (service: string, path: string): FileId|undefined => {
```
Pytal, line 36:
```suggestion
				return (service: string, path: string): ?FileId => {
```

Pytal (approved):

szaimen:
I get an error during compiling while trying to test this:
image

apps/files/src/store/paths.ts
skjnldsv: `?FileId` !== `FileId|undefined` :)

szaimen (changes requested):
This doesnt work in my testing.

I made a screen recording (with audio). See:
<details>
<summary>Click here</summary>


https://user-images.githubusercontent.com/42591237/231123568-1de9230b-d85e-4f99-8d70-ea6c0c9dc212.mp4



</details>

skjnldsv:
> Good question. For people who do not use a keyboard, this could be a nice way to select multiple files. It is hidden away in the 3 dot menu so I think it is okay to keep :)

Will do in a dedicated PR :+1:

szaimen (approved):
Works now but didnt review the code

PR #37642, opened 2023-04-07
feat(files): responsive design for mobile and narrow screens 

skjnldsv:
First commit is moving css around
The real code to review is in the second https://github.com/nextcloud/server/pull/37642/commits/7f3abbe4287dc06627ac5d89f41c2addfaefea7a

https://user-images.githubusercontent.com/14975046/230601960-2f91f71b-46cc-46a9-a92d-9b05c43f2b3e.mp4

____
@jancborchardt @nimishavijay 

Reference https://github.com/nextcloud/server/issues/37627 https://github.com/nextcloud/server/pull/36534
PS: I'll wait before compiling assets to avoid rebasing


szaimen:
I guess the multiselect buttons should also collapse? Apart from that looks good to me! :)
See 
image

skjnldsv:
Good catch Simon!! :rocket:

nimishavijay:
Very cool! Looks great in the screenshot, and completely agreed with Simon! :)

skjnldsv:
Done

szaimen (approved):
Tested and works :)
(but didnt review the code)

apps/files/src/components/FilesListHeaderActions.vue
```diff
@@ -122,6 +128,16 @@ export default Vue.extend({
...
 			},
 		},
+
+		inlineActions() {
+			if (this.clientWidth < 480) {
+				return 1
+			}
+			if (this.clientWidth < 768) {
+				return 2
+			}
```
artonge, line 138: How is this behaving with the sidebar opened ?
Could we use the container width instead of the client's width?

artonge: Also, would it make sense to have the hiding logic inside nc/vue instead of here?

skjnldsv: I used the values files had before.
It's fair to raise the question a bit more widely I think yes :)

When you say sidebar, you mean navigation or sidebar? (because I've seen it being used for both)
- If Navigation, the sidebar is floating on mobile, so it doesn't affect the main view which takes 100%
- If you say Sidebar (right), the Sidebar takes the full width on mobile, but we should definitely change those values if the sidebar is opened (see the video below)  

https://user-images.githubusercontent.com/14975046/231511828-c2025e35-e5be-44da-8b7d-417cfb594f1b.mp4

skjnldsv: I would merge this and improve the sidebar after raising the topic in another PR? :)

skjnldsv: Sidebar is splitting on `1024/2`
https://github.com/nextcloud/nextcloud-vue/blob/50f344487828c33dc9b0f7c5012abb4fd045512e/src/components/NcAppSidebar/NcAppSidebar.vue#L1064

artonge: Yes, I meant the right Sidebar. I am asking for the number of displayed actions. The sidebar reduce the width of the container, but the logic is based on the width of the client window, so it might not work as expected.

My advice would be to use a resize observer on the list container instead of the resize event on the window.

artonge (approved):

apps/files/src/components/FilesListHeaderActions.vue
skjnldsv: Wow, I will prepare an issue with RFC and screenshots, you're 100% right Louis, we can do better :+1:

skjnldsv: https://github.com/nextcloud/server/pull/37692

PR #37588, opened 2023-04-05
fix: Adjust filepicker breadcrumb structure to recent a11y changes

juliusknorr:
Fixes a regression from #37453 where the file picker bread crumbs were not adapted to the new structure:

Before:
Screenshot 2023-04-05 at 11 43 09

After:
Screenshot 2023-04-05 at 11 42 29

PR #37647, opened 2023-04-07
fix: Fix filepicker breadcrumbs

juliusknorr:
Follow up to #37588 as I broke the navigation and order of breadcrumbs


skjnldsv (approved):

JuliaKirschenheuter (approved):
Thank you Julius!

PR #37669, opened 2023-04-11
feat(files): right click

skjnldsv:
Implement right click = actions menu open

Reference https://github.com/nextcloud/server/issues/37627 https://github.com/nextcloud/server/pull/36534


szaimen:
First of all thanks for implementing this! :)

I've tested this and a few remarks:
- it feels strange that the right-click menu gets opened not directly at the tooltip but at where the expand button is triggered
- the size and date columns dont have a pointer (the mouse doesnt transform into a pointer)
- when multi-select is active, shouldn't right clicking show the multi-select options in a instead of the single file actions?

skjnldsv:
> it feels strange that the right-click menu gets opened not directly at the tooltip but at where the expand button is triggered

On previous discussions, it was decided to keep it simple like this. Maybe we can hijack the tooltip placement as a followup ?

> the size and date columns dont have a pointer (the mouse doesnt transform into a pointer)

Why would it? :thinking: 

> when multi-select is active, shouldn't right clicking show the multi-select options in opup instead of the single file actions?

Good point! :+1:

szaimen:
> > it feels strange that the right-click menu gets opened not directly at the tooltip but at where the expand button is triggered
> 
> On previous discussions, it was decided to keep it simple like this. Maybe we can hijack the tooltip placement as a followup ?

Follow-up sounds good!

> > the size and date columns dont have a pointer (the mouse doesnt transform into a pointer)
> 
> Why would it? 🤔

Do you see the mouse not transforming into a pointer on the right columts? But you can still right-click there and open the right-click menu by doing so.

https://user-images.githubusercontent.com/42591237/231115125-471f32eb-0c99-4e82-aac6-84d554482a18.mp4

skjnldsv:
> Do you see the mouse not transforming into a pointer on the right columts? But you can still right-click there and open the right-click menu by doing so.

The mouse cursor has never been about the right click? Only about the left click, no? :thinking:

szaimen:
> The mouse cursor has never been about the right click? Only about the left click, no? 🤔

Honestly I don't know but that would make sense 😅

skjnldsv:
Rebased and ready to review

szaimen (approved):
Works and fine by me then :)
(but didnt review the code)

artonge (approved):

damnshaneisthatu:
Does this change affect the "Rename" functionality as well? The focus after clicking "Rename" is out of sorts.

Issue #37039, opened 2023-03-03
Changes for developers & admins for Nextcloud 27


nickvergessen:
Continuation of #34692

# Readme

📝 Please always leave a comment when editing, so subscribed developers and admins are notified.

# Front-end

* …

# Backend

* …
* 
## Removed from public namespace:

* …

## Removed from private namespace:

* …

## Deprecations

* …

### Deprecations of Events & Hooks

* …

## Behavorial changes

* …

## Changes for admins

* …

nickvergessen:
Added:

> ## Removed from public namespace:
>
> * Intermediate transition event classes `OCP\WorkflowEngine\IEntityCompat` and `OCP\WorkflowEngine\IOperationCompat` have been removed as advertised for 2023: https://github.com/nextcloud/server/pull/37040

nickvergessen:
Added:


> ## Backend
>
> * `\OCP\AppFramework\Utility\ITimeFactory` is now a `\PSR\Clock\ClockInterface` following the [PSR-20 standard](https://www.php-fig.org/psr/psr-20/#21-clockinterface). Therefore `ITimeFactory::getTime` and `ITimeFactory::getDateTime` were deprecated: https://github.com/nextcloud/server/pull/35872

artonge:
Added:

> ## Behavorial changes
> 
>     * `CacheEntryRemovedEvent` will now be dispatched for all files and folders inside the deleted node.

provokateurin:
Added:

> ## Backend
> * `OCP\Accounts\IAccountManager::PROPERTY_DISPLAYNAME_LEGACY` was added to be used for keeping the provisioning API compatible: https://github.com/nextcloud/server/pull/36665

nickvergessen:
Added:

> ## Backend
> * A new system config `ratelimit.protection.enabled` (boolean, default true) was added to allow developers to **disable** rate limiting. It is **not** recommended to disable this on a production/live system.

marcelklehr:
Added:

> # Backend
>  * A new OCP API under `OCP\SpeechToText` was introduced to allow automatic transcription of media: #37674

nickvergessen:
Added:

> ## Backend
> 
> * BruteForceProtection can now be enabled using a native PHP attribute #36928  https://github.com/nextcloud/documentation/pull/9742

nickvergessen:
Added:

> # Backend
> 
[...]

PR #37674, opened 2023-04-11
feat(SpeechToText): Add SpeechToText OCP provider API

marcelklehr:
## Summary
Implement an OCP API register and call SpeechToText providers

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required: https://github.com/nextcloud/documentation/pull/10085


lib/public/SpeechToText/ISpeechToTextProvider.php
```diff
@@ -0,0 +1,45 @@
...
+	/**
+	 * @since 27.0.0
+	 */
+	public function getName(): string;
+
+	/**
+	 * @since 27.0.0
+	 * @throws RuntimeException If the text could not be transcribed
+	 */
+	public function transcribeFile(string $path): string;
```
juliushaertl, line 44: Since we talked about the path being a temp location, we may not be able to do this because the cron job may not run on the same physical server as where the job gets scheduled. So maybe we need to require having the file in the root folder

juliushaertl: Another option might be to just pass the context and emit an event also before transcribing that could just provide the file then as a local copy

nickvergessen: > So maybe we need to require having the file in the root folder


That would not scale for transcripting talk recordings, or am I missing something?

marcelklehr: Why would it not scale? The recordings are stored in some user folder eventually, right?

nickvergessen: > The recordings are stored in some user folder eventually, right?

Ah okay, if that is meant with rootfolder yeah. I thought it meant root as in /temp, server root, etc.

juliushaertl: Yes, I was referring to the data folder as in either user files or appdata

lib/public/SpeechToText/ISpeechToTextProvider.php
```diff
@@ -0,0 +1,46 @@
...
+	/**
+	 * @since 27.0.0
+	 */
+	public function getName(): string;
+
+	/**
+	 * @since 27.0.0
+	 * @throws RuntimeException If the text could not be transcribed
+	 */
+	public function transcribeFile(File $file): string;
```
marcelklehr, line 45: Changed this now to take a File instead of a path

lib/public/SpeechToText/ISpeechToTextManager.php
```diff
@@ -0,0 +1,62 @@
...
+	 * with the \OCP\SpeechToText\Events\TranscriptionFinishedEvent
+	 * You should add context information to the context array to re-identify the transcription result as
+	 * belonging to your transcription request.
+	 *
+	 * @since 27.0.0
+	 * @throws PreConditionNotMetException If no provider was registered but this method was still called
+	 * @throws InvalidArgumentException If the file could not be found or is not of a supported type
+	 * @throws RuntimeException If the transcription failed for other reasons
+	 */
+	public function scheduleFileTranscription(string $path, array $context): void;
```
nickvergessen, line 53: Also use `OCP\Files\File` here?

The unclear defined `context` is a bit weird as well.
Maybe just enforce a pattern? Things that come to mind:

- app id
- user
- file id (also solved by working with the Node everywhere)
- time of request

In case you are storing it somewhere the data length might be limited (e.g. 4k chars for background jobs).

So maybe just use a `referenceId` which is `<appid>-sha1(*)` this way we have a fixed non-colliding layout and also a very defined type and max length?

lib/public/SpeechToText/ISpeechToTextManager.php
```diff
@@ -0,0 +1,62 @@
...
+
+	/**
+	 * @since 27.0.0
+	 * @throws PreConditionNotMetException If no provider was registered but this method was still called
+	 * @throws InvalidArgumentException If the file could not be found or is not of a supported type
+	 * @throws RuntimeException If the transcription failed for other reasons
+	 */
+	public function transcribeFile(string $path) : string;
```
nickvergessen, line 61:
```suggestion
	public function transcribeFile(string $path): string;
```

nickvergessen: What is the returned string? The transcribe? A path to a file with the transcribe?

Also this method should work on a node?

lib/public/SpeechToText/Events/TranscriptionFinishedEvent.php
```diff
@@ -0,0 +1,73 @@
...
+
+/**
+ * @since 27.0.0
+ */
+class TranscriptionFinishedEvent extends Event {
+	/**
+	 * @since 27.0.0
+	 */
+	public function __construct(
+		private bool $successful,
```
nickvergessen, line 38: Could also make this class abstract and have a dedicated Success and Failed event extending this one?

lib/private/SpeechToText/TranscriptionJob.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+namespace OC\SpeechToText;
```
nickvergessen, line 3: missing declare strict_types + author + copyright

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,124 @@
...
+	}
+
+	public function transcribeFile(string $path): string {
+		$provider = current($this->getProviders());
+		if (!$provider) {
+			throw new PreConditionNotMetException('No SpeechToText providers have been registered');
+		}
+
+		try {
+			$node = $this->rootFolder->get($path);
```
nickvergessen, line 115: I would prefer to not do another roundtrip. In most cases the app should have the node already and would just pass that one on

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,124 @@
...
+		}
+		return !empty($context->getTranslationProviders());
+	}
+
+	public function scheduleFileTranscription(string $path, array $context): void {
+		if (!$this->hasProviders()) {
+			throw new PreConditionNotMetException('No SpeechToText providers have been registered');
+		}
+		try {
+			$node = $this->rootFolder->get($path);
```
nickvergessen, line 98: Same here, just require the node to be injected

nickvergessen (changes requested):

lib/public/SpeechToText/ISpeechToTextManager.php
marcelklehr: I was trying to spare you the need to do book-keeping and potentially having to create another db table just for this identifier. I like simply using the file Id, would this be enough, do you think, to simply drop the context param altogether?

lib/public/SpeechToText/Events/TranscriptionFinishedEvent.php
marcelklehr: :+1:

lib/public/SpeechToText/ISpeechToTextManager.php
nickvergessen: Should be fine as long as not 2 apps/actions do the same file :P

marcelklehr: Should still be fine in that case, as it's the same content then, no?

lib/private/SpeechToText/TranscriptionJob.php
marcelklehr: Fixed

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,116 @@
...
+
+		return $this->providers;
+	}
+
+	public function hasProviders(): bool {
+		$context = $this->coordinator->getRegistrationContext();
+		if ($context === null) {
+			return false;
+		}
+		return !empty($context->getTranslationProviders());
```
nickvergessen, line 89: **issue (blocking):** 
```suggestion
		return !empty($context->getSpeechToTextProviders());
```

lib/public/SpeechToText/Events/TranscriptionFailedEvent.php
```diff
@@ -0,0 +1,48 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+namespace OCP\SpeechToText\Events;
+
+class TranscriptionFailedEvent extends AbstractTranscriptionEvent {
```
nickvergessen, line 29: **issue (blocking):** InvalidDocblock

> lib/public/SpeechToText/Events/TranscriptionFailedEvent.php:29:1: InvalidDocblock: PHPDoc is required for classes/interfaces in OCP. (see https://psalm.dev/008)

lib/public/SpeechToText/Events/AbstractTranscriptionEvent.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @since 27.0.0
+ */
+abstract class AbstractTranscriptionEvent extends Event {
+	/**
+	 * @since 27.0.0
+	 */
+	public function __construct(
+		private int $fileIdId
```
nickvergessen, line 38: **suggestion:** you have the node object in all cases already, just pass that one through the events, to avoid the listeners having to load the node themselves (we will do it otherwise for each listener individually).

nickvergessen (changes requested):

lib/public/SpeechToText/Events/AbstractTranscriptionEvent.php
marcelklehr: I initially wanted to use the Node object directly, but we actually do not have the node object in all cases, eg. in case the file was removed before the background job is run. For the successful case we definitely have the Node object, but then the API would be different between successful and failed :shrug: I don't mind either way, personally, I like it better the way it is now, but of course it's less performant.

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,116 @@
...
+		}
+		return !empty($context->getSpeechToTextProviders());
+	}
+
+	public function scheduleFileTranscription(File $file): void {
+		if (!$this->hasProviders()) {
+			throw new PreConditionNotMetException('No SpeechToText providers have been registered');
+		}
+		try {
+			$this->jobList->add(TranscriptionJob::class, ['fileId' => $file->getId()]);
```
marcelklehr, line 97: Note to self: Using the jobs table as a queue is a bad idea if the jobs take longer than 5mins (which they do)

nickvergessen: Should still be fine? Jobs are reserved with a timestamp lock. Only becomes problematic after 1h, then the log is ignored

marcelklehr: I was thinking because multiple jobs may run at the same time if we leave it like this. That may not be expected for admins. If they set up a worker or even multiple for these it's fine, but for a normal 1machine-server it would clog up the CPUs.

marcelklehr: This one is kinda hard to do, as querying jobs inside a job class will likely manipulate the current job object :/

lib/public/SpeechToText/Events/AbstractTranscriptionEvent.php
marcelklehr: I've added a nullable File argument to the Abstract Event class now

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	public function scheduleFileTranscription(File $file): void {
+		if (!$this->hasProviders()) {
+			throw new PreConditionNotMetException('No SpeechToText providers have been registered');
+		}
+		try {
+			if ($this->jobList->has(TranscriptionJob::class, ['fileId' => $file->getId()])) {
+				return;
+			}
```
nickvergessen, line 99:
```suggestion
```

Should not be needed. It's already dealt with inside:
https://github.com/nextcloud/server/blob/d74044f6341e93861d820f21b6d37a2f98e1ef60/lib/private/BackgroundJob/JobList.php#L70

nickvergessen (approved):

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,116 @@
...
+		}
+		try {
+			$this->jobList->add(TranscriptionJob::class, ['fileId' => $file->getId()]);
+		} catch (NotFoundException|InvalidPathException $e) {
+			throw new InvalidArgumentException('Invalid file provided for file transcription: ' . $e->getMessage());
+		}
+	}
+
+	public function transcribeFile(File $file): string {
+		$provider = current($this->getProviders());
```
nickvergessen, line 104: I wonder if we should instead loop over all providers (maybe they are language dependent) and return the result from the first one that was successful. Similar to the translation API:
https://github.com/nextcloud/server/blob/3e6329838158892e3b89c7e5116fa313c282a8a1/lib/private/Translation/TranslationManager.php#L66-L80

lib/public/SpeechToText/ISpeechToTextManager.php
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * Will schedule a transcription process in the background. The result will become available
+	 * with the \OCP\SpeechToText\Events\TranscriptionFinishedEvent
+	 * You should add context information to the context array to re-identify the transcription result as
+	 * belonging to your transcription request.
+	 *
+	 * @since 27.0.0
+	 * @throws PreConditionNotMetException If no provider was registered but this method was still called
+	 * @throws InvalidArgumentException If the file could not be found or is not of a supported type
+	 * @throws RuntimeException If the transcription failed for other reasons
```
nickvergessen, line 52: **Suggestion:** Not applicable on scheduled transcript

lib/public/SpeechToText/ISpeechToTextManager.php
```diff
@@ -0,0 +1,65 @@
...
+	 */
+	public function scheduleFileTranscription(File $file): void;
+
+	/**
+	 * @since 27.0.0
+	 * @param File $file The media file to transcribe
+	 * @returns string The transcription of the the passed media file
```
nickvergessen, line 59: **Suggestion:**
```suggestion
	 * @returns string The transcription of the passed media file
```

lib/public/SpeechToText/ISpeechToTextManager.php
```diff
@@ -0,0 +1,65 @@
...
+	 * @throws PreConditionNotMetException If no provider was registered but this method was still called
+	 * @throws InvalidArgumentException If the file could not be found or is not of a supported type
+	 * @throws RuntimeException If the transcription failed for other reasons
+	 */
+	public function transcribeFile(File $file) : string;
```
nickvergessen, line 64: Coding style:
```suggestion
	public function transcribeFile(File $file): string;
```

lib/private/SpeechToText/SpeechToTextManager.php
marcelklehr: Yeah, good point, will change this

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,118 @@
...
+
+		$this->providers = [];
+
+		foreach ($context->getSpeechToTextProviders() as $providerRegistration) {
+			$class = $providerRegistration->getService();
+			try {
+				$this->providers[$class] = $this->serverContainer->get($class);
+			} catch (NotFoundExceptionInterface|ContainerExceptionInterface|Throwable $e) {
+				$this->logger->error('Failed to load SpeechToText provider ' . $class, [
+					'exception' => $e
```
julien-nc, line 76:
```suggestion
					'exception' => $e,
```
Nitpicking.

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		if ($this->providers !== null) {
+			return $this->providers;
+		}
+
+		$this->providers = [];
+
+		foreach ($context->getSpeechToTextProviders() as $providerRegistration) {
+			$class = $providerRegistration->getService();
```
julien-nc, line 71:
```suggestion
		foreach ($context->getSpeechToTextProviders() as $providerServiceRegistration) {
			$class = $providerServiceRegistration->getService();
```
Just a suggestion.

julien-nc (commented):
Just a few small remarks.

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,118 @@
...
+
+		foreach ($this->getProviders() as $provider) {
+			try {
+				return $provider->transcribeFile($file);
+			} catch (\Throwable $e) {
+				$this->logger->info('SpeechToText transcription using provider ' . $provider->getName() . ' failed', ['exception' => $e]);
+			}
+		}
+
+		throw new RuntimeException('Could not transcribe file');
```
nickvergessen, line 116: Missing import?
Class, interface or enum named OC\SpeechToText\RuntimeException does not exist

lib/private/SpeechToText/TranscriptionJob.php
```diff
@@ -0,0 +1,90 @@
...
+	) {
+		parent::__construct($timeFactory);
+	}
+
+
+	/**
+	 * @inheritDoc
+	 */
+	protected function run($argument) {
+		$fileId = $argument['fileId'];
```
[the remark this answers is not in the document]
nickvergessen: We never do this in background jobs. Whatever is meant to be given is expected.

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,119 @@
...
+		}
+		return !empty($context->getSpeechToTextProviders());
+	}
+
+	public function scheduleFileTranscription(File $file): void {
+		if (!$this->hasProviders()) {
+			throw new PreConditionNotMetException('No SpeechToText providers have been registered');
+		}
+		try {
+			$this->jobList->add(TranscriptionJob::class, ['fileId' => $file->getId()]);
```
nickvergessen, line 98: Looking at talk, we might want to store the acting user as well.
Will try to analyse tomorrow if the transript could end up for a wrong user when the recording was already shared to a room. But it would also lighten the logic in the background job if we additionally add the user.

marcelklehr: Added the user now as a job argument

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -0,0 +1,119 @@
...
+
+	public function hasProviders(): bool {
+		$context = $this->coordinator->getRegistrationContext();
+		if ($context === null) {
+			return false;
+		}
+		return !empty($context->getSpeechToTextProviders());
+	}
+
+	public function scheduleFileTranscription(File $file): void {
```
nickvergessen, line 93: As discussed, to make app live easier, we should allow passing in `string $appId` and `?string $userId` here and emit return them on events

lib/public/SpeechToText/ISpeechToTextManager.php
```diff
@@ -0,0 +1,67 @@
...
+	public function hasProviders(): bool;
+
+	/**
+	 * Will schedule a transcription process in the background. The result will become available
+	 * with the \OCP\SpeechToText\Events\TranscriptionFinishedEvent
+	 * You should add context information to the context array to re-identify the transcription result as
+	 * belonging to your transcription request.
+	 *
+	 * @param File $file The media file to transcribe
+	 * @param string $userId The user that triggered this request (only for convenience, will be available on the TranscriptEvents)
```
nickvergessen, line 50:
```suggestion
	 * @param ?string $userId The user that triggered this request (only for convenience, will be available on the TranscriptEvents)
```

nickvergessen (approved):

julien-nc (approved):

marcelklehr:
Documentation pr: https://github.com/nextcloud/documentation/pull/10085

nickvergessen:
Don't forget to add an entry in https://github.com/nextcloud/server/issues/37039 referencing this PR and the documentation page

marcelklehr:
Isn't #37039 only for (breaking) changes? I didn't know we also add additions there, that's also why I didn't add the translations API there. Will add this, though.

nickvergessen:
> that's also why I didn't add the translations API there.

Yeah I complained about this already.

No, it used to be breaking changes only, but now it's basically the OCP changelog.

PR #37680, opened 2023-04-11
app type extended_authentication

ArtificialOwl:
`globalsiteselector` is a single app used in `globalscale` that redirect users and/or sessions from one instance (master) to another (slaves) while working with local account, ldap, sso and saml.

Some use-cases:

- when using SAML, then authentication of the user is done on the master, a session is opened on the assigned slave and user is redirected. In that case, the globalsiteselector will comes with its own UserBackend that needs to be loaded soon enough in the process.
- on some OCS API Request, the `globalsiteselector` app **must** be loaded before the `files` app
- when using LDAP, the authentication is done on the slave. Based on the userid, the master will create a token with uid/password and redirect the user to the assigned slave. The slave will perform the authentication and `user_ldap` needs to be loaded before `globalsiteselector`

In conclusion, `globalsiteselector` needs to be loaded once all apps `type=authentication` are loaded and before `type=filesystem`

This fix creates a new type of apps for `globalsiteselector` that is loaded right after `authentication`.

### DIFF
--- a/lib/base.php
+++ b/lib/base.php
@@ -1027,6 +1027,7 @@ class OC {
 
 		// Always load authentication apps
 		OC_App::loadApps(['authentication']);
+		OC_App::loadApps(['extended_authentication']);
 
 		// Load minimum set of apps
 		if (!\OCP\Util::needUpgrade()
--- a/lib/private/Updater.php
+++ b/lib/private/Updater.php
@@ -331,7 +331,7 @@ class Updater extends BasicEmitter {
 	 */
 	protected function doAppUpgrade(): void {
 		$apps = \OC_App::getEnabledApps();
-		$priorityTypes = ['authentication', 'filesystem', 'logging'];
+		$priorityTypes = ['authentication', 'extended_authentication', 'filesystem', 'logging'];
 		$pseudoOtherType = 'other';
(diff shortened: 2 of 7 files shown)


blizzz (changes requested):
Please add new type also to `resources/app-info.xsd` and `resources/app-info-shipped.xsd` (this will need a PR to the appstore as well)

blizzz:
Unsure about the name, it is not saying much, however to my mind nothing great is coming either. I could think of `umbrella_authentication` or `catchall_authentication`  or `supre_authentication` but it is not necessarily better, latter might be confusing even. So, wording not a blocker, but open for improvements.

ArtificialOwl:
> 

good luck with that :-D
`extended_authentication` already took me a while. But the more I see it, the more it makes sense for me (from the `globalsiteselector` point of view).

ArtificialOwl:
> Please add new type also to `resources/app-info.xsd` and `resources/app-info-shipped.xsd` (this will need a PR to the appstore as well)

done and squashed

blizzz (approved):

icewind1991 (approved):

ChristophWurst:
This change may affect app devs and should be documented

PR #37626, opened 2023-04-06
fix(dav): add string comparison for diff

miaulalala:
* Resolves: #37282

## Summary
The diffing didn't diff. Now it does.

| Before | After |
|--------|--------|
| image | image | 

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #37683, opened 2023-04-12
[stable26] doc(auth): Warn about disabled token passwords and LDAP logout

ChristophWurst:
Backport of https://github.com/nextcloud/server/pull/37626


ChristophWurst:
Dup of https://github.com/nextcloud/server/pull/37684

PR #37691, opened 2023-04-12
repair -1 folder sizes for object store background scan

icewind1991:
Changes the object store background scan from a noop to repairing -1 sizes properly.


lib/private/Files/ObjectStore/ObjectStoreScanner.php
```diff
@@ -76,6 +73,55 @@ protected function scanChildren($path, $recursive = self::SCAN_RECURSIVE, $reuse
...
 	public function backgroundScan() {
-		//noop
+		$lastPath = null;
+		// find any path marked as unscanned and run the scanner until no more paths are unscanned (or we get stuck)
+		// we sort by path DESC to ensure that contents of a folder are handled before the parent folder
+		while (($path = $this->getIncomplete()) !== false && $path !== $lastPath) {
+			$this->runBackgroundScanJob(function () use ($path) {
+				$item = $this->cache->get($path);
+				if ($item && $item->getMimeType() !== FileInfo::MIMETYPE_FOLDER) {
+					$fh = $this->storage->fopen($path, 'r');
```
come-nc, line 83: Why not `$this->storage->stat($path)` ?

lib/private/Files/ObjectStore/ObjectStoreScanner.php
```diff
@@ -76,6 +73,55 @@ protected function scanChildren($path, $recursive = self::SCAN_RECURSIVE, $reuse
...
+	private function getIncomplete() {
+		$query = $this->connection->getQueryBuilder();
+		$query->select('path')
+			->from('filecache')
+			->where($query->expr()->eq('storage', $query->createNamedParameter($this->cache->getNumericStorageId(), IQueryBuilder::PARAM_INT)))
+			->andWhere($query->expr()->lt('size', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT)))
+			->orderBy('path', 'DESC')
+			->setMaxResults(1);
+
+		$result = $query->execute();
```
come-nc, line 116:
```suggestion
		$result = $query->executeQuery();
```

apps/files_sharing/lib/Scanner.php
```diff
@@ -72,7 +72,7 @@ private function getSourceScanner() {
 
 	public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData = null, $lock = true, $data = null) {
 		$sourceScanner = $this->getSourceScanner();
-		if ($sourceScanner instanceof NoopScanner) {
+		if ($sourceScanner instanceof ObjectStoreScanner) {
 			return [];
 		} else {
```
come-nc, line 77: What is this special treatment for?

lib/private/Files/ObjectStore/ObjectStoreScanner.php
icewind1991: `stat` with object stores just load the data from the cache.

`fopen` has some logic added to forward the `Content-Size` header from the object store. So it's the easiest way to get the real object size atm.

apps/files_sharing/lib/Scanner.php
icewind1991: `ObjectStoreScanner` doesn't do any scanning, so it just shorts circuits here.

This used to be clearer with the old `NoopScanner` name so I added a comment to explain

come-nc (approved):

lib/private/Files/ObjectStore/ObjectStoreScanner.php
```diff
@@ -76,6 +73,55 @@ protected function scanChildren($path, $recursive = self::SCAN_RECURSIVE, $reuse
...
+					if ($fh) {
+						$stat = fstat($fh);
+						if ($stat['size']) {
+							$this->cache->update($item->getId(), ['size' => $stat['size']]);
+						}
+					}
+				}
+			}, $path);
+			// FIXME: this won't proceed with the next item, needs revamping of getIncomplete()
+			// to make this possible
```
blizzz, line 93: when? ;)

icewind1991: eventually™.

it would require some deeper rework of how unscanned files are detected which I wouldn't want to do here

skjnldsv (approved):

Issue #37695, opened 2023-04-12
[Bug]: Expected filesize of X bytes but read (from Nextcloud client) and wrote (to Nextcloud storage) 0 bytes


mgutt:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Some of my users randomly face the issue that uploads are interrupted. Finally I found out that the user's browser does not respect the Chunking size (10 MiB default) and some or all uploads (if multiple) hit Apache's `LimitRequestBody` limit, which seems to be set by default to 1024MB in the Nextcloud Docker Container.

I found some other users complaining about this issue and solution:
https://help.nextcloud.com/t/experiencing-unwanted-1gb-upload-hard-limit-unable-to-find-the-culprit/157243/10

### Steps to reproduce

Can't be reliable reproduced as it happens randomly.

### Expected behavior

Upload should work in any case.

### Installation method

Community Docker image

### Nextcloud Server version

25

### Operating system

Other

### PHP engine version

None

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### List of activated Apps

```shell
irrelevant
```

### Nextcloud Logs

```shell
[...]

PR #37696, opened 2023-04-12
Add LimitRequestBody 0 to .htaccess

mgutt:
"LimitRequestBody 0" allows uploading files bigger than 1024 MB, which seems to be the default apache webserver value. This is important as some browsers do not respect the chunking (for some of my Safari users this happens randomly) and the upload fails with the error "Expected filesize of 1207907122 bytes but read (from Nextcloud client) and wrote (to Nextcloud storage) 0 bytes.".

In addition it allows setting a chunking size bigger than 1024 MB.

* Resolves:
https://github.com/nextcloud/server/issues/37695

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -95,3 +95,4 @@
 
 AddDefaultCharset utf-8
 Options -Indexes
+LimitRequestBody 0


szaimen (changes requested):
Hi, see https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/big_file_upload_configuration.html

mgutt:
> Hi, see https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/big_file_upload_configuration.html

I know that article, but finally it does not contain any explanation regarding the `LimitRequestBody` size, why it could be relevant and how to change it. To be exact: It is not possible to change it as every container upgrade overwrites the .htaccess file (that's why I use a cronjob to check and add the setting if its missing).

In addition the article describes solutions which shouldn't be needed at all as Nextcloud's chunking bypasses all those limits. With the default settings (10 MB chunking, 512MB PHP_UPLOAD_LIMIT/PHP_MEMORY_LIMIT) it was no problem to upload huge files to my servers for 99% of my users. Only a tiny amount of them had the problem that their browser randomly fail to chunk the files and by that those tiny amount hits the PHP, Apache and Proxy Server limits. So we try to create a workaround for random situations, which I think is ok as finally the user experience is the most important thing.

I think we have to solutions to solve this:
A) Add `LimitRequestBody 0` as the new default
B) Use the container's `PHP_UPLOAD_LIMIT` value and set the `LimitRequestBody` accordingly (in bytes of course)

szaimen:
> It is not possible to change it as every container upgrade overwrites the .htaccess file

See https://github.com/nextcloud/docker/issues/1796

mgutt:
> See [nextcloud/docker#1796](https://github.com/nextcloud/docker/issues/1796)

This is not a solution, it is only a workaround because the user has to create the following dirs with the proper perms BEFORE installing the container (unRAID paths as an example):

/mnt/user/appdata/Nextcloud/etc
/mnt/user/appdata/Nextcloud/etc/apache2
/mnt/user/appdata/Nextcloud/etc/apache2/conf-enabled

And file with the proper content:
/etc/apache2/conf-enabled/nextcloud-apache.conf

Even advanced users would now struggle as they can't know the perms and the needed content of the file before installing the container. So they will install the container first.

Then they would create the file with the needed configs and add a new mounted path and restart the container.

By that it is NOT possible to prepare the docker command (compose, docker run
..., or unraid templates) for non-advanced users as docker creates this empty dir if the file does not already exist:
/etc/apache2/conf-enabled/nextcloud-apache.conf/

In addition this solution blocks all changes through future apache upgrades.

And it does not make sense at all. I mean why does the container offer to change the `PHP_UPLOAD_LIMIT` if it fails with a higher value than 1024? This value must change `LimitRequestBody`, too, or it should be unlimited by default. I don't see an other solution for non-advanced users.

szaimen:
What I was pointing out that this is a bug in the docker image and needs to be fixed there 😉

It works fine e.g. with https://github.com/nextcloud/all-in-one

kesselb:
> sane default

Do you have a value in mind? 

If we add LimitRequestBody to our .htaccess file, you cannot override it anymore. 
In my tests, the value in .htaccess wins over the web server configuration.

mgutt:
> Do you have a value in mind?
0

I don't see a reason to have a limit at all.

> If we add LimitRequestBody to our .htaccess file, you cannot override it anymore.
Could be added to httpd.conf as well. Why not.

kesselb:
> How are upload_max_filesize/post_max_size handled?

.user.ini + ini_set

> Can't we use the same mechanics for LimitRequestBody?

Not possible

mgutt:
> Not possible

Why shouldn't it be possible to set the value in httpd.conf by dockers entrypoint script?!

kesselb:
@mgutt reworked my previous comment.

PR #34302, opened 2022-09-28
Fix: Prevent deadlocks during mtime/size/etag propagation

Raudius:
When uploading many files in one go deadlocks can occur when propagating the changes. This PR performs two changes which should help prevent deadlocks:

 * Combines the 2 propagation queries into 1 which halves the amount of queries that can cause the deadlock to occur
 * Adds a retry loop in case the deadlock occurs

PR #37709, opened 2023-04-13
fix: Check for wrapped retriable exceptions to handle deadlocks

juliusknorr:
Follow-up fix for https://github.com/nextcloud/server/pull/34302 which was trying to catch an exception that was wrapped by our own exception class so the catch would never have catched.

Exact code part https://github.com/nextcloud/server/pull/34302/files#diff-9cdfa82a2275acf10161656ddc7b70be83cdbf6be8663ecfb8f963e4eb516d8eR139


szaimen:
@juliushaertl what is missing here?

juliushaertl:
Split out the new API change and this is now ready for review for just the bugfix

ChristophWurst (approved):

come-nc (approved):

juliushaertl:
Failure unrelated

PR #32016, opened 2022-04-19
Use stored user for PasswordUpdatedEvent

artonge:
When handling `PasswordUpdatedEvent` event, we are calling `getLoginName` which does not exist.

This PR adds a condition to use the previously stored user when handling `PasswordUpdatedEvent`.

PR #37717, opened 2023-04-13
Do not override stored credentials when login in with SAML

artonge:
When login in with SAML, the password from `$event->getPassword()` is `null`.

This PR makes sure that this `null` value won't be used to override the stored password even though it is different.
This PR also allow for the password and user to be updated even though they were not set before.

 Followup of: https://github.com/nextcloud/server/pull/32016


blizzz:
can we have a simple unit test?

blizzz (approved):

come-nc (approved):

PR #37722, opened 2023-04-14
Fix preview generator semaphore typing

provokateurin:
## Summary

## Checklist

### DIFF
--- a/lib/private/Preview/Generator.php
+++ b/lib/private/Preview/Generator.php
@@ -46,6 +46,7 @@ use OCP\Preview\IProviderV2;
 use OCP\Preview\IVersionedPreviewFile;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
+use SysvSemaphore;
 
 class Generator {
 	public const SEMAPHORE_ID_ALL = 0x0a11;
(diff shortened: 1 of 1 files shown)


lib/private/Preview/Generator.php
```diff
@@ -311,7 +312,7 @@ private function getSmallImagePreview(ISimpleFolder $previewFolder, File $file,
 	 *
 	 * @param int $semId
 	 * @param int $concurrency
-	 * @return false|resource the semaphore on success or false on failure
+	 * @return false|resource|SysvSemaphore the semaphore on success or false on failure
```
provokateurin, line 315: For some reasone `resource` still needs to be listed here because psalm gets confused otherwise

lib/private/Preview/Generator.php
```diff
@@ -330,7 +331,7 @@ public static function guardWithSemaphore(int $semId, int $concurrency) {
 	/**
 	 * Releases the semaphore acquired from {@see Generator::guardWithSemaphore()}.
 	 *
-	 * @param resource|bool $semId the semaphore identifier returned by guardWithSemaphore
+	 * @param SysvSemaphore|bool $semId the semaphore identifier returned by guardWithSemaphore
 	 * @return bool
 	 */
 	public static function unguardWithSemaphore($semId): bool {
```
come-nc, line 337: You need to fix the call to is_resource in here, it will always be false

PR #37726, opened 2023-04-14
fix(files): standardize `files:node:event` event syntax

skjnldsv:
After discussion:
image

Note: For the other apps that already started using the old `files:file:updated` syntax, we should migrate.
  As those apps are shipped and have stablexx branches, I think it's safe to change it for master in both server and those apps. Only Notes would be an issue:
- https://github.com/nextcloud/notes/blob/f635c5c94d30130a5eaea317a05c7f558b78aaa4/src/components/NoteRich.vue#L63 @juliushaertl 
- https://github.com/nextcloud/photos/blob/805087317386dcd80c051a71df0e1774e42848b3/src/components/FilesListViewer.vue#L230 @artonge 
- https://github.com/nextcloud/text/blob/01ccc7c17eb21d39fd8f4077538b97aa36dd4f0d/src/components/Editor.vue#L637 @juliushaertl 
- https://github.com/nextcloud/viewer/blob/0c9b93c497184d069690ff8026a8bc19566c3005/src/components/ImageEditor.vue#L178 @skjnldsv


nickvergessen (approved):

juliusknorr (approved):

juliushaertl:
For notes we can add a version check then

PR #37731, opened 2023-04-14
feat(files): add view config service to store user-config per view

skjnldsv:
Introduce a dedicated API to save user config per view

I feel this is much cleaner and more flexible than one API for sorting, one API for expanded state... etc
It also allows to get the list of ALL views configs, easily, [which the current expanded 'api' does not allow](https://github.com/nextcloud/server/blob/84292422098bbbbf7955f9450f4c73d553f4fae6/apps/files/lib/Controller/ApiController.php#L407-L420)

Example:
 - Expanded navigation state
 - Sorting mode (size, name...)
 - Sorting order (asc, desc)

Later: 
- Grid display state per view?


artonge (approved):

apps/files/lib/Service/ViewConfig.php
```diff
@@ -0,0 +1,190 @@
+<?php
+/**
+ * @copyright Copyright (c) 2022 John Molakvoæ <skjnldsv@protonmail.com>
```
Pytal, line 3:
```suggestion
 * @copyright Copyright (c) 2023 John Molakvoæ <skjnldsv@protonmail.com>
```
📅

apps/files/lib/Service/ViewConfig.php
```diff
@@ -0,0 +1,190 @@
...
+			throw new \InvalidArgumentException('Unknown config key');
+		}
+	
+		if (!in_array($value, $this->getAllowedConfigValues($key))
+			&& !empty($this->getAllowedConfigValues($key))) {
+			throw new \InvalidArgumentException('Invalid config value');
+		}
+
+		if (is_bool($value)) {
+			$value = $value ? '1' : '0';
```
Pytal, line 130: **question (non-blocking):** Instead of having to cast boolean to string would it be worth adding boolean parsing or is the legacy code too unwieldy?

Pytal (approved):

apps/files/lib/Service/ViewConfig.php
skjnldsv: Ouups :see_no_evil:

apps/files/lib/Service/ViewConfig.php
skjnldsv: Cleaner with one cast to boolean instead of handling int.
Thanks for the suggestion!! :rocket:

PR #37734, opened 2023-04-14
feat(dav): expose system address book

### LINKED ISSUE #37797: [Feature] CardDAV - expose system address book
Part of https://github.com/nextcloud/server/issues/19575

Currently, the system principal address book is not available from the AddressBookHome.

## Todo

- [x] Add the system address book while still respecting the share enumeration settings as defined in the Share API.
- [x] Exclude any users coming from https://github.com/nextcloud/guests (uri: Guests:email.vcf)
- [x] Exclude any disabled users (in my tests, they disappeared from the system address books).

## Follow-ups

* Feature request: expose a guest users addressbook (use case still to be found)

miaulalala:
* Fixes #37797

## Summary
Expose the system address book

## How to test

Enable share enumeration in the backend

image

to see all users.

In the contacts app, the users will have the address book "system":

image

Disable it to see the logged in user only:

| With enumeration | Without enumeration |
|--------|--------|
| image | image | 

You can also check the DAV url <your-url>/remote.php/dav/addressbooks/users/<your-username> to see the new system address book:

image

## To Do
- [x] check if renaming address book to something else but "system" for the new user principal collection is possible to avoid naming conflicts
- [x] Rename the principal collection Display name to "Accounts" for the sabre collection.

## Checklist

- [x] https://github.com/nextcloud/documentation/issues/10021
- [x] https://github.com/nextcloud/documentation/issues/10020

### DIFF
--- a/apps/dav/appinfo/v1/carddav.php
+++ b/apps/dav/appinfo/v1/carddav.php
@@ -10,6 +10,7 @@
  * @author Morris Jobke <hey@morrisjobke.de>
  * @author Thomas Citharel <nextcloud@tcit.fr>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
+ * @author Anna Larch <anna.larch@gmx.net>
  *
  * @license AGPL-3.0
  *
@@ -72,7 +73,7 @@ $principalCollection = new \Sabre\CalDAV\Principal\Collection($principalBackend)
 $principalCollection->disableListing = !$debugging; // Disable listing
 
 $pluginManager = new PluginManager(\OC::$server, \OC::$server->query(IAppManager::class));
-$addressBookRoot = new AddressBookRoot($principalBackend, $cardDavBackend, $pluginManager);
+$addressBookRoot = new AddressBookRoot($principalBackend, $cardDavBackend, $pluginManager, \OC::$server->getUserSession()->getUser(), \OC::$server->get(\OCP\IGroupManager::class));
 $addressBookRoot->disableListing = !$debugging; // Disable listing
 
 $nodes = [
--- a/apps/dav/lib/CardDAV/AddressBookRoot.php
+++ b/apps/dav/lib/CardDAV/AddressBookRoot.php
@@ -5,6 +5,7 @@
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Joas Schilling <coding@schilljs.com>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
+ * @author Anna Larch <anna.larch@gmx.net>
  *
  * @license AGPL-3.0
  *
@@ -24,11 +25,15 @@
 namespace OCA\DAV\CardDAV;
 
 use OCA\DAV\AppInfo\PluginManager;
+use OCP\IGroupManager;
+use OCP\IUser;
 
 class AddressBookRoot extends \Sabre\CardDAV\AddressBookRoot {
 
 	/** @var PluginManager */
 	private $pluginManager;
+	private ?IUser $user;
+	private ?IGroupManager $groupManager;
 
 	/**
 	 * @param \Sabre\DAVACL\PrincipalBackend\BackendInterface $principalBackend
@@ -38,9 +43,13 @@ class AddressBookRoot extends \Sabre\CardDAV\AddressBookRoot {
 	public function __construct(\Sabre\DAVACL\PrincipalBackend\BackendInterface $principalBackend,
 								\Sabre\CardDAV\Backend\BackendInterface $carddavBackend,
 								PluginManager $pluginManager,
-								$principalPrefix = 'principals') {
+								?IUser $user,
+								?IGroupManager $groupManager,
+								string $principalPrefix = 'principals') {
 		parent::__construct($principalBackend, $carddavBackend, $principalPrefix);
 		$this->pluginManager = $pluginManager;
+		$this->user = $user;
+		$this->groupManager = $groupManager;
 	}
 
 	/**
@@ -55,7 +64,7 @@ class AddressBookRoot extends \Sabre\CardDAV\AddressBookRoot {
 	 * @return \Sabre\DAV\INode
 	 */
 	public function getChildForPrincipal(array $principal) {
-		return new UserAddressBooks($this->carddavBackend, $principal['uri'], $this->pluginManager);
+		return new UserAddressBooks($this->carddavBackend, $principal['uri'], $this->pluginManager, $this->user, $this->groupManager);
 	}
 
 	public function getName() {
--- a/apps/dav/lib/CardDAV/CardDavBackend.php
+++ b/apps/dav/lib/CardDAV/CardDavBackend.php
@@ -311,6 +311,11 @@ class CardDavBackend implements BackendInterface, SyncSupport {
 			'{http://sabredav.org/ns}sync-token' => $row['synctoken'] ?: '0',
 		];
 
+		// system address books are always read only
+		if ($principal === 'principals/system/system') {
+			$addressBook['{' . \OCA\DAV\DAV\Sharing\Plugin::NS_OWNCLOUD . '}read-only'] = true;
+		}
+
 		$this->addOwnerPrincipal($addressBook);
 
 		return $addressBook;
--- a/apps/dav/lib/CardDAV/SystemAddressbook.php
+++ b/apps/dav/lib/CardDAV/SystemAddressbook.php
@@ -8,6 +8,7 @@ declare(strict_types=1);
  * @author Joas Schilling <coding@schilljs.com>
  * @author Julius Härtl <jus@bitgrid.net>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Anna Larch <anna.larch@gmx.net>
  *
  * @license GNU AGPL version 3 or any later version
  *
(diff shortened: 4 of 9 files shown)


apps/dav/lib/CardDAV/AddressBookRoot.php
```diff
@@ -38,9 +40,10 @@ class AddressBookRoot extends \Sabre\CardDAV\AddressBookRoot {
 	public function __construct(\Sabre\DAVACL\PrincipalBackend\BackendInterface $principalBackend,
 								\Sabre\CardDAV\Backend\BackendInterface $carddavBackend,
 								PluginManager $pluginManager,
-								$principalPrefix = 'principals') {
+								$principalPrefix = 'principals', ?IUser $user = null) {
```
miaulalala, line 43: As not to destroy the order of parameters, the `IUser $user` needs to be initialised with null. Just in case other apps instantiate this class.

apps/dav/lib/CardDAV/UserAddressBooks.php
```diff
@@ -48,12 +48,15 @@
...
+	private IUser $user;
 
 	public function __construct(Backend\BackendInterface $carddavBackend,
 								string $principalUri,
-								PluginManager $pluginManager) {
+								PluginManager $pluginManager,
+								?IUser $user) {
 		parent::__construct($carddavBackend, $principalUri);
 		$this->pluginManager = $pluginManager;
+		$this->user = $user;
```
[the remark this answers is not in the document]
miaulalala: Should be fixed when rebased.

apps/dav/lib/CardDAV/SystemAddressbook.php
```diff
@@ -29,25 +29,40 @@
...
 		$shareEnumeration = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'yes') === 'yes';
 		$shareEnumerationGroup = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_group', 'no') === 'yes';
 		$shareEnumerationPhone = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_phone', 'no') === 'yes';
-		if (!$shareEnumeration || $shareEnumerationGroup || $shareEnumerationPhone) {
-			return [];
+		if (!$shareEnumeration) {
+			if ($this->user === null) {
+				return [];
+			}
+			$name = $this->user->getBackendClassName() . ':' . $this->user->getUID() . '.vcf';
```
ChristophWurst, line 55: add a public method to `\OCA\DAV\CardDAV\SyncService` and use it here so the naming creation happens at one place only

apps/settings/templates/settings/admin/sharing.php
```diff
@@ -209,7 +209,7 @@
 				<?php if ($_['allowShareDialogUserEnumeration'] === 'yes') {
 	print_unescaped('checked="checked"');
 } ?> />
-			<label for="shareapi_allow_share_dialog_user_enumeration"><?php p($l->t('Allow username autocompletion in share dialog'));?></label><br />
+			<label for="shareapi_allow_share_dialog_user_enumeration"><?php p($l->t('Allow username autocompletion in share dialog and expose system address book endpoint'));?></label><br />
```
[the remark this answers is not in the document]
miaulalala: Makes it sound as if username autocompletion is enabled for the system address book too. But it's an additional feature. Maybe not expose but `Allow username autocompletion in share dialog and allow access to the system address book` or similar?

apps/dav/lib/CardDAV/CardDavBackend.php
```diff
@@ -495,27 +493,69 @@ public function deleteAddressBook($addressBookId) {
...
+		if(empty($cardIds)) {
+			return $cards;
+		}
+
+		// Chunk the query so the IN doesn't fail
+		$query->where($query->expr()->in('id', $query->createParameter('ids'), IQueryBuilder::PARAM_STR_ARRAY));
+
+		foreach (array_chunk($cardIds, 1000) as $chunk) {
+			$query = $query->setParameter('ids', $chunk);
+			$result = $query->execute();
```
miaulalala, line 545: Probably just blind but the query never returns any results even though the query looks ok and runs locally. Taking a look tomorrow.

st3iny (commented):
I tested this locally (without federation). I'm impressed! It works quite good.

However, the contacts in the system address book are not indicated to be read-only. The fields can't actually be edited though because saving fails. Could this be a problem with the ACLS of the system address book?

miaulalala:
From @ChristophWurst for how we handle the share enumeration:

```
0) No checkbox checked -> Show only the same user
1) Only "Allow username autocompletion in share dialog" -> show everyone
2) "Allow username autocompletion in share dialog" + "Allow username autocompletion to users within the same groups" -> show only users in intersecting groups
3) "Allow username autocompletion in share dialog" + "Allow username autocompletion to users based on phone number integration" -> show only the same user
4) Allow username autocompletion in share dialog" + "Allow username autocompletion to users within the same groups" + "Allow username autocompletion to users based on phone number integration" -> show only users in intersecting groups
```

apps/dav/lib/CardDAV/CardDavBackend.php
```diff
@@ -145,87 +143,89 @@ public function getAddressBooksForUserCount($principalUri) {
 	 * @return array
 	 */
 	public function getAddressBooksForUser($principalUri) {
-		$principalUriOriginal = $principalUri;
-		$principalUri = $this->convertPrincipal($principalUri, true);
-		$query = $this->db->getQueryBuilder();
-		$query->select(['id', 'uri', 'displayname', 'principaluri', 'description', 'synctoken'])
-			->from('addressbooks')
-			->where($query->expr()->eq('principaluri', $query->createNamedParameter($principalUri)));
+		return $this->atomic(function () use ($principalUri) {
```
miaulalala, line 146: Will be gone after rebase

apps/settings/templates/settings/admin/sharing.php
```diff
@@ -40,88 +41,90 @@
 		<p id="enable">
 			<input type="checkbox" name="shareapi_enabled" id="shareAPIEnabled" class="checkbox"
 				   value="1" <?php if ($_['shareAPIEnabled'] === 'yes') {
-	print_unescaped('checked="checked"');
-} ?> />
+				   	print_unescaped('checked="checked"');
```
ChristophWurst, line 44: changes in this file can be reverted I guess

ChristophWurst (commented):
Looks good so far

ChristophWurst:
This PR is currently blocked from merge

- [ ] Requires https://github.com/nextcloud/server/pull/38048
- [ ] Requires https://github.com/nextcloud/server/pull/38013

apps/settings/templates/settings/admin/sharing.php
miaulalala: including text changes? or just whitespace changes?

ChristophWurst: Didn't see the text changes, sorry. They should stay of course.

miaulalala: Have commited only text changes

miaulalala:
Yesss, drone is passing

apps/dav/lib/CardDAV/UserAddressBooks.php
```diff
@@ -69,11 +77,24 @@ public function getChildren() {
...
 		$addressBooks = $this->carddavBackend->getAddressBooksForUser($this->principalUri);
+		// add the system address book
+		$systemAddressBook = null;
+		if (is_string($principal) && $principal !== 'principals/system/system' && $this->carddavBackend instanceof CardDavBackend) {
+			$systemAddressBook = $this->carddavBackend->getAddressBooksByUri('principals/system/system', 'system');
+		}
+		if (!is_null($systemAddressBook)) {
+			$addressBooks[] = $systemAddressBook;
+		}
+		if (empty($addressBooks)) {
```
ChristophWurst, line 91: premature optimization. there might be ABs proved by plugins at L103. they should still be returned even if the user has no owned ABs

apps/dav/lib/CardDAV/SystemAddressbook.php
```diff
@@ -45,6 +45,7 @@
 use Sabre\VObject\Reader;
 
 class SystemAddressbook extends AddressBook {
+	public const URI_SHARED = 'z-server-generated--system';
```
tcitworld, line 48: **question:** Why is the constant named like this?

ChristophWurst: to avoid a naming conflict. if a user has an addressbook `system` I don't want the URIs to overlap. and the naming is inspired by the app provided ABs/calendars, e.g. `z-app-generated--contactsinteraction--recent`

tcitworld: I didn't meant the constant's value, but the constant's name itself. What's shared here?

ChristophWurst: Right. The system/system principal's addressbook is "shared" with all users. Does that make sense?

The URI remains `system` for the system/system principal. It should only change when the system address book is added into users' address book home.

tcitworld: I feel "shared" mixes things with regular sharing. `URI_GLOBAL` would do it, but I'm fine either way.

ChristophWurst:
Checked current behavior with @jancborchardt 

* System contacts should show in the contacts menu – they do :heavy_check_mark: 
* System contacts should not show in the sharing dialogue – they do not :heavy_check_mark:

apps/dav/lib/CardDAV/SystemAddressbook.php
```diff
@@ -31,38 +32,113 @@
...
+	 * 'Allow username autocompletion in share dialog' + 'Allow username autocompletion to users within the same groups' + 'Allow username autocompletion to users based on phone number integration' -> show only users in intersecting groups
+	 */
+	public function getChildren() {
 		$shareEnumeration = $this->config->getAppValue('core', 'shareapi_allow_share_dialog_user_enumeration', 'yes') === 'yes';
 		$shareEnumerationGroup = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_group', 'no') === 'yes';
 		$shareEnumerationPhone = $this->config->getAppValue('core', 'shareapi_restrict_user_enumeration_to_phone', 'no') === 'yes';
-		if (!$shareEnumeration || $shareEnumerationGroup || $shareEnumerationPhone) {
-			return [];
+		if (!$shareEnumeration || !$shareEnumerationGroup && $shareEnumerationPhone) {
+			$name = SyncService::getCardUri($this->user);
```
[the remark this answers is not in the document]
miaulalala: ah yes, i knew there was something i forgot

apps/dav/lib/CardDAV/SystemAddressbook.php
```diff
@@ -72,6 +73,29 @@ public function __construct(BackendInterface $carddavBackend,
...
+	 * The properties list is a list of propertynames the client requested,
+	 * encoded in clark-notation {xmlnamespace}tagname
+	 *
+	 * If the array is empty, it means 'all properties' were requested.
+	 *
+	 * @param array $properties
+	 *
+	 * @return array
+	 */
+	public function getProperties($properties) {
```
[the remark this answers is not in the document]
miaulalala: yes

miaulalala:
Group enumeration breaks at the moment because the User session at the time of passing the object to the AddressBookRoot is still null.

Authentication happpens when adding the Auth plugin:

https://github.com/nextcloud/server/blob/9e9133a86368cac55879e469069e10c687264dc7/apps/dav/appinfo/v1/carddav.php#L91

Which calls `check()` on the authBackend. Can't move the calls around as the addressbookroot is needed to be added as a node to the server.

ChristophWurst:
> Group enumeration breaks at the moment because the User session at the time of passing the object to the AddressBookRoot is still null.

Is it just the order of execution? Inject the IUserSession and fetch the IUser object as late as possible. Would that help?

apps/dav/lib/CardDAV/SystemAddressbook.php
```diff
@@ -31,38 +32,113 @@
...
+			if ($this->groupManager === null) {
+				// Group manager is not available, so we can't determine which data is safe
+				return [];
+			}
+			$groups = $this->groupManager->getUserGroups($this->user);
+			$names = [];
+			foreach ($groups as $group) {
+				$users = $group->getUsers();
+				foreach ($users as $groupUser) {
+					if ($groupUser->getBackendClassName() === 'Guests') {
```
ChristophWurst, line 128: - [ ] Test this. I'm convinced this will be a class string of https://github.com/nextcloud/guests/blob/master/lib/UserBackend.php

ChristophWurst: image

ChristophWurst:
Rebasing

ChristophWurst (approved):
Works

apps/dav/lib/CardDAV/UserAddressBooks.php
```diff
@@ -33,18 +34,23 @@
...
+use OCP\IGroupManager;
 use OCP\IL10N;
+<<<<<<< HEAD
 use OCP\IRequest;
+use OCP\IUserSession;
 use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\NotFoundExceptionInterface;
+=======
+use OCP\IUser;
+>>>>>>> 25a5d9661f9 (feat(dav): expose system address book)
```
ChristophWurst, line 46: not again

ChristophWurst (approved):
Just tested with davx5 and it works (again)

apps/dav/appinfo/v1/carddav.php
```diff
@@ -72,7 +73,7 @@
 $principalCollection->disableListing = !$debugging; // Disable listing
 
 $pluginManager = new PluginManager(\OC::$server, \OC::$server->query(IAppManager::class));
-$addressBookRoot = new AddressBookRoot($principalBackend, $cardDavBackend, $pluginManager);
+$addressBookRoot = new AddressBookRoot($principalBackend, $cardDavBackend, $pluginManager, \OC::$server->getUserSession()->getUser(), \OC::$server->get(\OCP\IGroupManager::class));
```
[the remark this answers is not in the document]
blizzz: `\OCP\Server::get(…)` … but can be added in a follow up

JohannesGGE:
Testet with contacts app:
- I can delete the system adress book in the settings :upside_down_face:
- (Currently no option to disable the adressbook (same as recently contacted). Could be done in another issue but important for large sysAddBooks I think)

JohannesGGE (approved):

JohannesGGE (approved):

ChristophWurst:
Integration test failure is the ususal contact menu
Psalm fails randomly, possibly fixed by https://github.com/nextcloud/server/pull/38195

dac2020:
Is this a bug?

A new install. I created an admin who is in the admin group and two users in their own group. When the new users go to their Contacts page, they see the name of the admin in the Manager section.

It shows something like admin-z-server-generated-system

I’ve checked everything I can in the documentation but cannot find how I can hide the ‘Manager’ from the users Contacts view.

dac2020:
Cannot get help in forums, cannot get any answer here. Not sure it's worth spending any more time with this.

PR #37758, opened 2023-04-16
redis: use atomic operations everywhere

pulsejet:
This removes a lot of acrobatics in the code and does each operation
atomically using a lua script. This also reduces several round trips
to the server, and the scripts are compiled and cached server-side.

Notably, since all operations work only on a single key (except clear,
which is broken anyway and shouldn't be used), they will continue to
function and be atomic for Redis cluster.

EDIT: I did some simple benchmark with a single client process. Looks like locking is approximately 2x faster (probably even better with multiple clients). This is with Redis on the same host, so gains will be better when latency is higher between the servers.
```php
<?php
        $lock = \OC::$server->get(\OCP\Lock\ILockingProvider::class);

        $t1 = microtime(true);
        for($i = 0; $i < 20000; $i++) {
            $lock->acquireLock('benchmark', \OCP\Lock\ILockingProvider::LOCK_EXCLUSIVE);
            $lock->releaseLock('benchmark', \OCP\Lock\ILockingProvider::LOCK_EXCLUSIVE);
        }
        $t2 = microtime(true);
        $output->writeln("Total: " . ($t2 - $t1) . "s");
```

```
Before: 5.590616941452s
After: 2.6423881053925s
```

On a side note, using DB for locking is a joke:
```
MySQL: 232.49977707863s
```

### DIFF
--- a/lib/private/Memcache/Redis.php
+++ b/lib/private/Memcache/Redis.php
@@ -31,23 +31,23 @@ namespace OC\Memcache;
 
 use OCP\IMemcacheTTL;
 
-/** name => [script, sha1] */
-const LUA_SCRIPTS = [
-	'dec' => [
-		'if redis.call("exists", KEYS[1]) == 1 then return redis.call("decrby", KEYS[1], ARGV[1]) else return "NEX" end',
-		'720b40cb66cef1579f2ef16ec69b3da8c85510e9',
-	],
-	'cas' => [
-		'if redis.call("get", KEYS[1]) == ARGV[1] then redis.call("set", KEYS[1], ARGV[2]) return 1 else return 0 end',
-		'94eac401502554c02b811e3199baddde62d976d4',
-	],
-	'cad' => [
-		'if redis.call("get", KEYS[1]) == ARGV[1] then return redis.call("del", KEYS[1]) else return 0 end',
-		'cf0e94b2e9ffc7e04395cf88f7583fc309985910',
-	],
-];
-
 class Redis extends Cache implements IMemcacheTTL {
+	/** name => [script, sha1] */
+	public const LUA_SCRIPTS = [
+		'dec' => [
+			'if redis.call("exists", KEYS[1]) == 1 then return redis.call("decrby", KEYS[1], ARGV[1]) else return "NEX" end',
+			'720b40cb66cef1579f2ef16ec69b3da8c85510e9',
+		],
+		'cas' => [
+			'if redis.call("get", KEYS[1]) == ARGV[1] then redis.call("set", KEYS[1], ARGV[2]) return 1 else return 0 end',
+			'94eac401502554c02b811e3199baddde62d976d4',
+		],
+		'cad' => [
+			'if redis.call("get", KEYS[1]) == ARGV[1] then return redis.call("del", KEYS[1]) else return 0 end',
+			'cf0e94b2e9ffc7e04395cf88f7583fc309985910',
+		],
+	];
+
 	/**
 	 * @var \Redis|\RedisCluster $cache
 	 */
@@ -185,10 +185,10 @@ class Redis extends Cache implements IMemcacheTTL {
 		return \OC::$server->getGetRedisFactory()->isAvailable();
 	}
 
-	protected function evalLua($scriptName, $keys, $args) {
+	protected function evalLua(string $scriptName, array $keys, array $args) {
 		$keys = array_map(fn ($key) => $this->getPrefix() . $key, $keys);
 		$args = array_merge($keys, $args);
-		$script = LUA_SCRIPTS[$scriptName];
+		$script = self::LUA_SCRIPTS[$scriptName];
 
 		$result = $this->getCache()->evalSha($script[1], $args, count($keys));
 		if (false === $result) {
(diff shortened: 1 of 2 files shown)


lib/private/Memcache/Redis.php
```diff
@@ -54,18 +54,19 @@ public function getCache() {
 
 	public function get($key) {
 		$result = $this->getCache()->get($this->getPrefix() . $key);
-		if ($result === false && !$this->getCache()->exists($this->getPrefix() . $key)) {
```
pulsejet, line 57: This check is useless. Redis does not support boolean types, so getting a `false` means it did not exist.

lib/private/Memcache/Redis.php
```diff
@@ -82,6 +83,7 @@ public function remove($key) {
 	}
 
 	public function clear($prefix = '') {
+		// TODO: this is slow and would fail with Redis cluster
 		$prefix = $this->getPrefix() . $prefix . '*';
 		$keys = $this->getCache()->keys($prefix);
 		$deleted = $this->getCache()->del($keys);
```
pulsejet, line 89: This will fail for redis cluster and a large number of keys. The solution is to replace this with a lua script and run it on all masters.

lib/private/Memcache/Redis.php
```diff
@@ -145,18 +151,19 @@ public function dec($key, $step = 1) {
...
-			$new = json_encode($new);
-		}
-		$this->getCache()->watch($this->getPrefix() . $key);
-		if ($this->get($key) === $old) {
-			$result = $this->getCache()->multi()
-				->set($this->getPrefix() . $key, $new)
-				->exec();
-			return $result !== false;
-		}
-		$this->getCache()->unwatch();
```
pulsejet, line 158: 5 statements here vs 1 now. Ditto for `cad`.

lib/private/Memcache/Redis.php
```diff
@@ -167,15 +174,17 @@ public function cas($key, $old, $new) {
...
-		return false;
+		$old = self::encodeValue($old);
+
+		return $this->getCache()->eval(
+			'if redis.call("get", KEYS[1]) == ARGV[1] then
+				return redis.call("del", KEYS[1])
```
pulsejet, line 181: Using `unlink` here is a premature optimization that's probably harmful. In fact, unless the values are *huge*, `unlink` internally calls `del` after checking the size, so we just did those checks for no reason...

icewind1991 (approved):

icewind1991:
Added a unit test to validate the script hashes to make sure we don't forget to update them if we ever change the scripts

pulsejet:
> Added a unit test to validate the script hashes to make sure we don't forget to update them if we ever change the scripts

Good point, thanks!

pulsejet:
Bump?

CarlSchwan (approved):
This is pretty neat. I didn't knew that redis allowed to run lua script

tests/lib/Memcache/RedisTest.php
```diff
@@ -9,6 +9,8 @@
 
 namespace Test\Memcache;
 
+use const OC\Memcache\LUA_SCRIPTS;
```
CarlSchwan, line 12: php cs fixer doesn't like it :(

pulsejet: It was the incorrect namespace anyway. Moved it to inside the class instead.

lib/private/Memcache/Redis.php
```diff
@@ -185,4 +184,25 @@ public function setTTL($key, $ttl) {
 	public static function isAvailable(): bool {
 		return \OC::$server->getGetRedisFactory()->isAvailable();
 	}
+
+	protected function evalLua($scriptName, $keys, $args) {
```
juliushaertl, line 188: Can we type hint those arguments directly?

lib/private/Memcache/Redis.php
```diff
@@ -185,4 +184,25 @@ public function setTTL($key, $ttl) {
...
+
+		$result = $this->getCache()->evalSha($script[1], $args, count($keys));
+		if (false === $result) {
+			$result = $this->getCache()->eval($script[0], $args, count($keys));
+		}
+
+		return $result;
+	}
+
+	protected static function encodeValue($value) {
```
juliushaertl, line 201:
```suggestion
	protected static function encodeValue(mixed $value): string {
```

juliusknorr (commented):
Minor comments, otherwise this seems like a nice improvement. 

Is there any minimum version requirement for this on the redis side?

pulsejet:
> Minor comments, otherwise this seems like a nice improvement.

Thanks, will fix these tomorrow.

> Is there any minimum version requirement for this on the redis side?

[Redis 2.6](https://raw.githubusercontent.com/redis/redis/2.6/00-RELEASENOTES), released ~10 years ago.

pulsejet:
CI failure unrelated

juliusknorr (approved):
Let's get this in for 28 👍

pulsejet:
Great! A possible reason I see is that `Redis::dec` wasn't atomic earlier, maybe that led to a race. But transaction support with cluster is iffy to begin with anyway.

AndyScherzinger:
@ChristophWurst  yes, backporting would be fine from my pov 👍

solracsf:
@ChristophWurst any specific reason why not backporting to 25 also?

ChristophWurst:
Wrongly got the impression 25 is EOL but it's only 24

ChristophWurst:
I think I'll backport the 26 backport because there are conflicts to resolve

PR #37773, opened 2023-04-17
Scale ocdialog based on screen size to contain more information on big screens

SuperSandro2000:
## Summary

Before:

image

After:

image

## TODO

- [ ] test with a deployment, I only changed the values in the css because I didn't have the time to set up a nextcloud development environment 

## Checklist


core/src/OC/dialogs.js
```diff
@@ -534,9 +534,8 @@ const Dialogs = {
 
 			self.$filePicker.ocdialog({
 				closeOnEscape: true,
-				// max-width of 600
-				width: 600,
-				height: 500,
+				width: '70vw',
+				height: '60vh',
```
Pytal, line 538: Should have a max-width and max-height as well but not sure what values would be best, @jancborchardt @skjnldsv?

skjnldsv: and min too :+1:

SuperSandro2000: I am not super good in css, should I set the following:

```suggestion
				mix-width: '70vw',
				max-width: '70vw',
				min-height: '60vh',
				max-height: '60vh',
```
or

```suggestion
				mix-width: '600',
				width: '70vw',
				min-height: '500',
				height: '60vh',
```
?

PR #37787, opened 2023-04-18
fix: catch ManuallyLockedException and use app context

### LINKED ISSUE #123: Old versions can not be restored from the sidebar because file is locked
How to reproduce:
* open a file (.md or .odt)
* add some content so multiple versions exist
* open the versions tab in the sidebar
* try to restore previous version
* see `Could not restore version`.

The server responds with a 500 with this content
```xml
  <s:exception>OCP\Lock\ManuallyLockedException</s:exception>
  <s:message>"files/AA.mine/test text/restore file version with lock.md" is locked, existing lock on file: files_lock/bd8abcbd-a01a-40c5-8332-b9c2e5c42cd2</s:message>
</d:error>
```
<details>
<summary>
Trace from the server:
</summary>

```yaml
{
app: webdav,
method: MOVE,
url: /remote.php/dav/versions/<user-id>/16523/1838744392,
message: "files/filename.docx" is locked, existing lock on file: files_lock/15551a66-c88c-49ea-8c3c-e85585df40a3,
exception: {
      Exception: OCP\Lock\ManuallyLockedException
      Message: "files/filename.docx" is locked, existing lock on file: files_lock/15551a66-c88c-49ea-8c3c-e85585df40a3
      Code: 0
      Trace: [ 
           {
               file: /var/www/html/nextcloud/apps/files_lock/lib/Storage/LockWrapper.php
               line: 214
               function: checkPermission
               class: OCA\FilesLock\Storage\LockWrapper"
               type: ->
           },
           {
               file: /var/www/html/nextcloud/lib/private/Files/Storage/Wrapper/Wrapper.php
               line: 301
               function: fopen
               class: OCA\FilesLock\Storage\LockWrapper"
               type: ->
           },
           {
               file: /var/www/html/nextcloud/apps/files_versions/lib/Storage.php
               line: 442
               function: fopen
               class: OCA\FilesLock\Wrapper\Wrapper"
               type: ->
           },
           {
               file: /var/www/html/nextcloud/apps/files_versions/lib/Storage.php
               line: 401
               function: copyFileContents
               class: OCA\Files_Versions\Storage
               type: ::
           },
    ]
}
```
</details>

max-nextcloud:
* Resolves: https://github.com/nextcloud/files_lock/issues/123

## Summary

The files_lock app may throw ManuallyLockedExceptions when attempting to revert a file that is currently opened. This would prevent the user from rolling back a opened file.

Text and Richdocuments handle changes of the file while editing. Allow reverting files even when they are locked by these apps and let the apps handle the conflict.

## Alternatives considered

I tried to tackle this in lower levels first:
* Inside the `Storage\LockWrapper` in `files_lock` - but it turns out we use different functions (`rename` and `fopen`) to rollback to an old version depending on the storage type. The `fopen` happens separatedly for the source and the target and therefore it's impossible to detect that it was triggered by a version rollback.
* Inside the `Storage` of `files_version` - it's possible to get a hold of the actual node there. But it would only fix the issue in the normal version storage. Separate fixes would still be required for `collectives` and `group_folders` and any other versioning implementation.

So i think this is the easiest way to go.

## TODO

- [x] checked if this could be moved into the `files_lock` app - looks like it cannot due to load order issues. Once the files_lock app is loaded other apps might already have dependency injected the VersionsManager. So we cannot change it from within the `files_lock` app.

## Checklist

### DIFF
--- a/apps/files_versions/lib/Versions/VersionManager.php
+++ b/apps/files_versions/lib/Versions/VersionManager.php
@@ -27,8 +27,13 @@ namespace OCA\Files_Versions\Versions;
 
 use OCP\Files\File;
 use OCP\Files\FileInfo;
+use OCP\Files\IRootFolder;
+use OCP\Files\Lock\ILock;
+use OCP\Files\Lock\ILockManager;
+use OCP\Files\Lock\LockContext;
 use OCP\Files\Storage\IStorage;
 use OCP\IUser;
+use OCP\Lock\ManuallyLockedException;
 
 class VersionManager implements IVersionManager, INameableVersionBackend, IDeletableVersionBackend {
 	/** @var (IVersionBackend[])[] */
@@ -94,7 +99,7 @@ class VersionManager implements IVersionManager, INameableVersionBackend, IDelet
 
 	public function rollback(IVersion $version) {
 		$backend = $version->getBackend();
-		$result = $backend->rollback($version);
+		$result = self::handleAppLocks(fn(): ?bool => $backend->rollback($version));
 		// rollback doesn't have a return type yet and some implementations don't return anything
 		if ($result === null || $result === true) {
 			\OC_Hook::emit('\OCP\Versions', 'rollback', [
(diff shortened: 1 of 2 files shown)


apps/files_versions/lib/Versions/VersionManager.php
```diff
@@ -133,4 +138,45 @@ public function deleteVersion(IVersion $version): void {
...
+	/**
+	 * Catch ManuallyLockedException and retry in app context if possible.
+	 *
+	 * The files_lock app may throw ManuallyLockedExceptions
+	 * when attempting to revert a file that is currently opened.
+	 * This would prevent the user from rolling back a opened file.
+	 *
+	 * Text and Richdocuments handle changes of the file while editing.
+	 * Allow reverting files even when they are locked by these apps
+	 * and let the apps handle the conflict.
```
artonge, line 151: A bit nitpicky, but I find it cleared. I might be saying wrong things though.

```suggestion
	 * Context: The files_lock app will throw ManuallyLockedExceptions
	 * when writing to a file that is currently locked.
	 * This is to prevents a user from editing an already opened file.
	 *
	 * Exception: Text and Richdocuments can handle changes of the file while editing.
	 * So we execute the edition under their lock context
	 * to let them handle the conflict.
```

apps/files_versions/lib/Versions/VersionManager.php
```diff
@@ -133,4 +138,45 @@ public function deleteVersion(IVersion $version): void {
...
+			$appsThatHandleUpdates = array("text", "richdocuments");
+			if (!in_array($owner, $appsThatHandleUpdates)) {
+				throw $e;
+			}
+			// The LockWrapper in the files_lock app only compares the lock type and owner
+			// when checking the lock against the current scope.
+			// So we do not need to get the actual node here
+			// and use the root node instead.
+			$root = \OC::$server->get(IRootFolder::class);
+			$lockContext = new LockContext($root, ILock::TYPE_APP, $owner());
```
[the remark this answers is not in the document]
max-nextcloud: Good catch! Used to be `$e->getOwner()` but I now convert that to a `string` above to make sure it's not null. Funny that this is not caught by psalm.

apps/files_versions/lib/Versions/VersionManager.php
max-nextcloud: Thanks for the proposal. It's really only about rolling back to old versions - not about editing. I will take your changes as a starting point and rephrase from there. :)

max-nextcloud: I did rephrase it. Curious to hear what you think @artonge

apps/files_versions/lib/Versions/VersionManager.php
```diff
@@ -133,4 +138,48 @@ public function deleteVersion(IVersion $version): void {
...
+	 *
+	 * Allow users to go back to old versions via the versions tab in the sidebar
+	 * even when the file is opened in the viewer next to it.
+	 *
+	 * Context: If a file is currently opened for editing
+	 * the files_lock app will throw ManuallyLockedExceptions.
+	 * This prevented the user from rolling an opened file back to a previous version.
+	 *
+	 * Text and Richdocuments can handle changes of open files.
+	 * So we execute the rollback under their lock contex
```
artonge, line 153:
```suggestion
* So we execute the rollback under their lock context
```

artonge (approved):
Typo, else nice :+1:

juliusknorr (approved):

juliushaertl:
Failure unrelated

PR #37790, opened 2023-04-18
fix(files): fix sorting mixin usage

skjnldsv:
You cannot use a setup in a mixin :facepalm: 
https://github.com/vuejs/core/issues/1866

Regression from #37731


apps/files/src/mixins/filesSorting.ts
```diff
@@ -22,13 +22,13 @@
...
 export default Vue.extend({
-	setup() {
-		const viewConfigStore = useViewConfigStore()
+	data() {
 		return {
-			viewConfigStore,
+			// composition api setup cannot be used in mixins
+			viewConfigStore: useViewConfigStore(),
 		}
 	},
```
ShGKme, line 33: In general case, composables should not be used in `data` or anywhere outside `setup`. As they are literally "composable" of the Vue component in meaning of "a part of setup".

While creating a new instance of a component with `setup` the instance being setup is the **current instance**. This **current instance** is used by all composable functions (such as `onMounted` or `provide`) as well as reactivity effect scope (so effects stop when the component is unmounted).

In this usage `useViewConfigStore` has no connection to the instance with this mixin:
- `activePinia` is not injected with `inject` and default global is used
- Store instance is not in a scope of the component
- Anything added latter to the composable could be broken is it require a composable to be a composable (such as `onMounted`, `ref` and etc)

Currently it has no problem now, because this composable actually does nothing composable-related, except injecting a pinia store, what is allowed by pinia with globally initialized pinia. (Though, I don't know how it will work with many app with many global pinias). 

Proposals:

1. Use recommended way to use Pinia in Options API (also works in mixin): https://pinia.vuejs.org/core-concepts/state.html#usage-with-the-options-api
  - mixin will work fine, injected pinia will be used
  - useViewConfigStore in this case should be a pinia store, not a composable with `defineStore` inside. In this case the store initialization could be a part of a `defineStore` or a part of ESM.

2. Make a whole mixin - a composable

artonge (approved):

ShGKme (commented):
> You cannot use a setup in a mixin 🤦

This is intentional. As `setup` is component instance "creating" function in Composition while `mixin` is merging "configuration" for Options. `Setups` are not mergeable.

apps/files/src/mixins/filesSorting.ts
skjnldsv: > 1. Use recommended way to use Pinia in Options API (also works in mixin): [pinia.vuejs.org/core-concepts/state.html#usage-with-the-options-api](https://pinia.vuejs.org/core-concepts/state.html#usage-with-the-options-api)

Good from me, thanks for the feedback :+1:

apps/files/src/mixins/filesSorting.ts
```diff
@@ -21,18 +21,14 @@
...
-	setup() {
-		const viewConfigStore = useViewConfigStore()
-		return {
-			viewConfigStore,
-		}
-	},
-
 	computed: {
+		...mapState(useViewConfigStore, ['getConfig', 'setSortingBy', 'toggleSortingDirection']),
+
```
ShGKme, line 31: Injected Pinia is lost here, as `useViewConfigStore` is not Pinia's `useStore`. Was it intended?

ShGKme:
> Yep, see the thread I linked in the first post! :)

Sorry, missed the link in the text 😅

apps/files/src/mixins/filesSorting.ts
skjnldsv: `useViewConfigStore` is a wrapper around Pinia's `useStore`. Pinia injection seems to work fine :thinking:

skjnldsv: Capture d’écran_2023-04-19_08-41-13

skjnldsv: If that was not working, and if my memory is correct, it would have been throwing `[🍍]: getActivePinia was called with no active Pinia. Did you forget to install pinia?`

ShGKme: > useViewConfigStore is a wrapper around Pinia's useStore. Pinia injection seems to work fine 🤔
> EDIT: here is some explanations behind the fancy wrapper trick I'm doing https://github.com/vuejs/pinia/discussions/1176#discussioncomment-2551258

In this comment they are using Composition API. In Composition API it works because it uses `inject` during `setup`, when there is the `currentInstance`.

In Options API it is injected by `this.$pinia` here: https://github.com/vuejs/pinia/blob/1650c6efebc75fdc47b2ec082ba10c197b894aef/packages/pinia/src/mapHelpers.ts#L272

And used here: https://github.com/vuejs/pinia/blob/1650c6efebc75fdc47b2ec082ba10c197b894aef/packages/pinia/src/store.ts#L884-L902

In current usage `this.$pinia` is lost.

---

> If that was not working, and if my memory is correct, it would have been throwing
> [🍍]: getActivePinia was called with no active Pinia. Did you forget to install pinia?

This is a bit different things. This error appears when there is no active Pinia at all. For example, if you are using `useStore` before init pinia with `createPinia`.

It works, because after `createPinia` there is already globally set `activePinia`, and because we don't have several different `pinia` in several apps, (and don't have SSR).

But it doesn't follow `mapState` interface. Also, it may works now, but it depends on how pinia uses `activePinia` internally.

Here you provide `pinia` and this provide is ignored in this usage: https://github.com/nextcloud/server/blob/9db33055b208518601ff7b389188c032f8792e39/apps/files/src/main.js#L52-L56

Proposals:
1. Make `useViewConfigStore` more a wrapper - pass `...args` to `useStore(...args)`
2. Use general `useStore` without `useViewConfigStore` and move init into the store definition   
3. Use general `useStore` without `useViewConfigStore` and init the store just in ES module.

The first one seems to be easy and doesn't break anything. Then `useViewConfigStore` will be truly a wrapper.

skjnldsv: https://github.com/nextcloud/server/pull/37790/commits/98ee8e15e6c83063582f21b7e1074ac3d4f367e6

ShGKme (approved):
Good to me 🦖
Thank you!

PR #37808, opened 2023-04-18
Add security information (IP and browser) to the password reset email

### LINKED ISSUE #4345: Improve lost password email
Followup to #4308

Mockup by @Espina2 

new password

Comments by @eppfel:

> Yeah, @Espina2 copy would be much appreciated. 
> 
> Too minor edits:
> 
> This ~~reset~~ password _reset_ is only available... 
> 
> Or:
> 
> This password reset link is only.... 
> 
> And 
> 
> ~~For security,~~ this request...

susnux:
* Resolves: #4345

## Summary
Add some metadata like the IP address and the browser type to the password reset mail for fraud detection.

## Screenshots

before | after
---|---
image | image

## Checklist

### DIFF
--- a/core/Controller/LostController.php
+++ b/core/Controller/LostController.php
@@ -308,12 +308,37 @@ class LostController extends Controller {
 			htmlspecialchars($this->l10n->t('Click the following button to reset your password. If you have not requested the password reset, then ignore this email.')),
 			$this->l10n->t('Click the following link to reset your password. If you have not requested the password reset, then ignore this email.')
 		);
-
 		$emailTemplate->addBodyButton(
 			htmlspecialchars($this->l10n->t('Reset your password')),
 			$link,
 			false
 		);
+
+		if ($this->request->isUserAgent([IRequest::USER_AGENT_CLIENT_ANDROID])) {
+			$platform = $this->l10n->t('Android client');
+		} elseif ($this->request->isUserAgent([IRequest::USER_AGENT_CLIENT_IOS])) {
+			$platform = $this->l10n->t('iOS client');
+		} elseif ($this->request->isUserAgent([IRequest::USER_AGENT_CLIENT_DESKTOP])) {
+			$platform = $this->l10n->t('desktop client');
+		} elseif ($this->request->isUserAgent(['/Firefox\//'])) {
+			$platform = 'Firefox';
+		} elseif ($this->request->isUserAgent(['/Chromium\//'])) {
+			$platform = 'Chromium';
+		} elseif ($this->request->isUserAgent(['/Safari\//'])) {
+			$platform = 'Safari';
+		} elseif ($this->request->isUserAgent(['/(OPR|Opera)\//'])) {
+			$platform = 'Opera';
+		} elseif ($this->request->isUserAgent(['/Edg.*\//'])) {
+			$platform = 'Edge';
+		} elseif ($this->request->isUserAgent(['/Chrome\//'])) {
+			$platform = 'Chrome';
+		} else {
+			$platform = $this->l10n->t('unknown');
+		}
+		$emailTemplate->addBodyText(
+			htmlspecialchars($this->l10n->t('Security notice: This password reset was requested from %1$s (IP address) using %2$s.', [ $this->request->getRemoteAddress(), $platform ]))
+		);
+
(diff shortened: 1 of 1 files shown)


core/Controller/LostController.php
```diff
@@ -308,12 +308,31 @@ protected function sendEmail(string $input): void {
...
+			$platform = 'Opera';
+		} elseif ($this->request->isUserAgent(['/Edg.*\//'])) {
+			$platform = 'Edge';
+		} elseif ($this->request->isUserAgent(['/Chrome\//'])) {
+			$platform = 'Chrome';
+		} else {
+			$platform = $this->l10n->t('unknown');
+		}
+		$emailTemplate->addBodyText(
+			htmlspecialchars($this->l10n->t('Security notice: This password reset was requested from following IP address and browser: %1$s (%2$s).', [ $this->request->getRemoteAddress(), $platform ]))
```
nickvergessen, line 333: I would rephrase and reorder this information

> ## Password reset
>
> A password resset was requested from %1$s (IP address) and using %2$s.
> If you have not requested the password reset, then ignore this email.
> Otherwise, click the following link to reset your password.

core/Controller/LostController.php
```diff
@@ -308,12 +308,31 @@ protected function sendEmail(string $input): void {
...
+		} elseif ($this->request->isUserAgent(['/Safari\//'])) {
+			$platform = 'Safari';
+		} elseif ($this->request->isUserAgent(['/(OPR|Opera)\//'])) {
+			$platform = 'Opera';
+		} elseif ($this->request->isUserAgent(['/Edg.*\//'])) {
+			$platform = 'Edge';
+		} elseif ($this->request->isUserAgent(['/Chrome\//'])) {
+			$platform = 'Chrome';
+		} else {
+			$platform = $this->l10n->t('unknown');
```
nickvergessen, line 330: we should make sure that it does not show "unknown" when reset is done via a mobile app or the desktop client.

nickvergessen (changes requested):

core/Controller/LostController.php
susnux: Good point, fixed.

core/Controller/LostController.php
susnux: I have added that additional paragraph because of this comment: https://github.com/nextcloud/server/issues/4345#issuecomment-296418122

core/Controller/LostController.php
nickvergessen: Can you add the Talk clients too? 😅

core/Controller/LostController.php
nickvergessen: So I checked all reset password emails in my inbox and most of them did not have information like this.
But if they did, it was the first information given.

cc @jancborchard for your 6y old comment

nickvergessen: So I checked all reset password emails in my inbox and most of them did not have information like this.
But if they did, it was the first information given.

cc @jancborchardt for your 6y old comment

come-nc:
Stupid question maybe, but is there no data privacy implications here?
This is sending the IP and browser of a web visitor to another person by email. I understand in this case it’s the web visitor’s fault for entering the email address of someone else but laws may not be that precise?

tobiasKaminsky:
It is. Server is processing ip address and thus user must be informed prior to it.
(other services, e.g. Google go even further and try to locate ip address, to give user a better info where the login attempt was done, e.g "user tried to log in near Frankfurt")

As far as I know we do not ship by default such an information, so I tend to not include this at all.

core/Controller/LostController.php
skjnldsv: also cc @nimishavijay :)

nickvergessen: As per GDPR officer we can not do this:
https://github.com/nextcloud/server/pull/37808#issuecomment-1515931164

skjnldsv: Then closing :)

skjnldsv:
As per GDPR officer we can not do this:
https://github.com/nextcloud/server/pull/37808#issuecomment-1515931164

susnux:
Would it make sense to only include the browser?

PR #37818, opened 2023-04-19
debug log requests made by dav external storage/shares

icewind1991:
And add an event to the profiler while we're at it.

### DIFF
--- a/lib/private/Files/Storage/DAV.php
+++ b/lib/private/Files/Storage/DAV.php
@@ -44,6 +44,7 @@ use OC\Files\Filesystem;
 use OC\MemCache\ArrayCache;
 use OCP\AppFramework\Http;
 use OCP\Constants;
+use OCP\Diagnostics\IEventLogger;
 use OCP\Files\FileInfo;
 use OCP\Files\ForbiddenException;
 use OCP\Files\StorageInvalidException;
@@ -56,6 +57,7 @@ use Sabre\DAV\Xml\Property\ResourceType;
 use Sabre\HTTP\ClientException;
 use Sabre\HTTP\ClientHttpException;
 use Psr\Log\LoggerInterface;
+use Sabre\HTTP\RequestInterface;
 
 /**
  * Class DAV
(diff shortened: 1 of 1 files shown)


lib/private/Files/Storage/DAV.php
```diff
@@ -162,6 +168,18 @@ protected function init() {
...
+
+		$lastRequestStart = 0;
+		$this->client->on('beforeRequest', function(RequestInterface $request) use (&$lastRequestStart) {
+			$this->logger->debug("sending dav request to external storage: " . $request->getAbsoluteUrl());
+			$lastRequestStart = microtime(true);
+			$this->eventLogger->start('fs:storage:dav:request', "Sending dav request to external storage");
+		});
+		$this->client->on('afterRequest', function(RequestInterface $request) use (&$lastRequestStart) {
+			$elapsed = microtime(true) - $lastRequestStart;
+			$this->logger->debug("dav request to external storage: " . $request->getAbsoluteUrl() . " took " . round($elapsed * 1000, 1) . "ms");
```
blizzz, line 180: 'app' context would be useful to be able to set a log condition on it.

lib/private/Files/Storage/DAV.php
```diff
@@ -162,6 +168,18 @@ protected function init() {
...
 		}
+
+		$lastRequestStart = 0;
+		$this->client->on('beforeRequest', function(RequestInterface $request) use (&$lastRequestStart) {
+			$this->logger->debug("sending dav request to external storage: " . $request->getAbsoluteUrl());
+			$lastRequestStart = microtime(true);
+			$this->eventLogger->start('fs:storage:dav:request', "Sending dav request to external storage");
+		});
+		$this->client->on('afterRequest', function(RequestInterface $request) use (&$lastRequestStart) {
+			$elapsed = microtime(true) - $lastRequestStart;
```
come-nc, line 179: That works?
What is the scope of the `$lastRequestStart` var, PHP detects that both functions are using it and keeps it alive until both functions are deleted?

icewind1991: Yes, capturing it with the closure keeps the var alive. The trick only works because it's captured by reference though (else the closures get their unique copy).

come-nc:
@icewind1991 You pushed a change to 3rdparty

blizzz (changes requested):
accidental bump to 3rdparty

icewind1991:
fixed

blizzz (approved):

come-nc (approved):

PR #37820, opened 2023-04-19
ignore errors while trying to update parent storage_mtime

icewind1991:
in the worst case this should only cause an extra rescan later.

This should hopefully reduce some cases of deadlock exceptions

### DIFF
--- a/lib/private/Files/Cache/Updater.php
+++ b/lib/private/Files/Cache/Updater.php
@@ -27,10 +27,12 @@
  */
 namespace OC\Files\Cache;
 
+use OC\DB\Exceptions\DbalException;
 use OC\Files\FileInfo;
 use OCP\Files\Cache\ICacheEntry;
 use OCP\Files\Cache\IUpdater;
 use OCP\Files\Storage\IStorage;
+use Psr\Log\LoggerInterface;
 
 /**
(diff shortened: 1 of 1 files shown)


lib/private/Files/Cache/Updater.php
```diff
@@ -253,7 +254,13 @@ private function correctParentStorageMtime($internalPath) {
...
 			$mtime = $this->storage->filemtime($parent);
 			if ($mtime !== false) {
-				$this->cache->update($parentId, ['storage_mtime' => $mtime]);
+				try {
+					$this->cache->update($parentId, ['storage_mtime' => $mtime]);
+				} catch (DbalException $e) {
+					// ignore the failure.
+					// with failures concurrent updates, someone else would have already done it.
+					// in the worst case the `storage_mtime` isn't updated, which should at most only trigger an extra rescan
+				}
```
come-nc, line 263: Log it, at least in debug?
Silent ignore is always scary :ghost:

icewind1991: done

come-nc (approved):

blizzz (approved):

juliushaertl:
Thanks for checking that Robin :)

PR #37824, opened 2023-04-19
feat(files): add default action support and expose router

skjnldsv:
### Adds default action support
This means any default action not being bound to the file name click will be shown in the actions menu (so far there is none). But that means we could have `Open in Viewer` as default and `Open in Collabora`... etc in the menu :rocket: 
Reference: https://github.com/nextcloud/viewer/issues/2393

### Expose vue router to allow external apps to change the route
Example:
```js
OCP.Files.Router.goTo('/trashbin?dir=/Unsplash.d1680193199')
OCP.Files.Router.goToRoute('fileslist', { view: 'files' }, { dir: '/Folders/Group folder' })
```

____
Required for https://github.com/nextcloud/groupfolders/pull/2326


apps/files/src/components/FilesListHeaderActions.vue
```diff
@@ -167,11 +167,18 @@ export default Vue.extend({
 				// Dispatch action execution
 				const results = await action.execBatch(this.nodes, this.currentView)
 
+				// Check if all actions returned null
+				if (results.filter(result => result !== null).length === 0) {
```
artonge, line 171:
```suggestion
				if (!results.some(result => result !== null)) {
```

apps/files/src/main.js
```diff
@@ -22,6 +22,9 @@ import router from './router/router.js'
 window.OCA.Files = window.OCA.Files ?? {}
 window.OCP.Files = window.OCP.Files ?? {}
 
+// Expose router
+Object.assign(window.OCP.Files, { Router: router })
```
artonge, line 26: What would be the use case to access the router? To react in an action created by another app?
If so, I would include it in the API instead of exposing it publicly.

skjnldsv: The example on the first post is from https://github.com/nextcloud/groupfolders/pull/2326
Where we redirect to the files list from an action

But I'm guessing there would be other similar things. Like wanting to just redirect the user to a specific view/dir/fileid without refreshing the page. For that the vue router is the only one that can do it

ChristophWurst:
I'd highly prefer a simple wrapper around the Vue router. It potentially makes upgrades and the migration away from the Vue router easier. Todays fancy tech will be legacy one day.

skjnldsv:
Done, the router is implemented, see first comment

artonge (approved):

ChristophWurst (approved):
Looks good

Docs needed for the new OCP.Files.Router API

PR #37835, opened 2023-04-20
feat(BackgroundJobs): Allow preventing parallel runs for a job class

marcelklehr:
* Related: #37674 
 
## Summary
This PR adds a setAllowParallelRuns() method to the BackgroundJob\Job abstract class in order to allow jobs to disallow parallel runs.

## Checklist

### DIFF
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -396,7 +396,8 @@ class JobList implements IJobList {
 		}
 
 		try {
-			return $query->executeQuery()->rowCount() > 0;
+			$result = $query->executeQuery();
+			return count($result->fetchAll()) > 0;
 		} catch (Exception $e) {
 			return false;
 		}
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -39,16 +39,19 @@ use OCP\DB\Exception;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IConfig;
 use OCP\IDBConnection;
+use Psr\Log\LoggerInterface;
 
 class JobList implements IJobList {
 	protected IDBConnection $connection;
 	protected IConfig $config;
 	protected ITimeFactory $timeFactory;
+	protected LoggerInterface $logger;
 
-	public function __construct(IDBConnection $connection, IConfig $config, ITimeFactory $timeFactory) {
+	public function __construct(IDBConnection $connection, IConfig $config, ITimeFactory $timeFactory, LoggerInterface $logger) {
 		$this->connection = $connection;
 		$this->config = $config;
 		$this->timeFactory = $timeFactory;
+		$this->logger = $logger;
 	}
 
 	/**
@@ -401,6 +404,7 @@ class JobList implements IJobList {
 			$result->closeCursor();
 			return $hasReservedJobs;
 		} catch (Exception $e) {
+			$this->logger->debug('Querying reserved jobs failed', ['exception' => $e]);
 			return false;
 		}
 	}
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -388,7 +388,8 @@ class JobList implements IJobList {
 		$query = $this->connection->getQueryBuilder();
 		$query->select('*')
 			->from('jobs')
-			->where($query->expr()->neq('reserved_at', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT)));
+			->where($query->expr()->neq('reserved_at', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT)))
+			->setMaxResults(1);
 
 		if ($className !== null) {
 			$query->andWhere($query->expr()->eq('class', $query->createNamedParameter($className)));
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -384,7 +384,7 @@ class JobList implements IJobList {
 		$query->executeStatement();
 	}
 
-	public function hasReservedJob(?string $className): bool {
+	public function hasReservedJob(?string $className = null): bool {
 		$query = $this->connection->getQueryBuilder();
 		$query->select('*')
 			->from('jobs')
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -397,7 +397,9 @@ class JobList implements IJobList {
 
 		try {
 			$result = $query->executeQuery();
-			return count($result->fetchAll()) > 0;
+			$hasReservedJobs = $result->fetch() !== false;
+			$result->closeCursor();
+			return $hasReservedJobs;
 		} catch (Exception $e) {
 			return false;
 		}
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -35,6 +35,7 @@ use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\AutoloadNotAllowedException;
 use OCP\BackgroundJob\IJob;
 use OCP\BackgroundJob\IJobList;
+use OCP\DB\Exception;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IConfig;
 use OCP\IDBConnection;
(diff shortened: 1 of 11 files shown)


tests/lib/BackgroundJob/TestJob.php
```diff
@@ -8,7 +8,9 @@
 
 namespace Test\BackgroundJob;
 
-class TestJob extends \OC\BackgroundJob\Job {
+use OCP\AppFramework\Utility\ITimeFactory;
+
+class TestJob extends \OCP\BackgroundJob\Job {
```
marcelklehr, line 13: This class still used to extend the deprecated OC class. Since I didn't add the new methods to that class, I had to change this to extend the OCP class instead. I hope this is fine.

lib/private/BackgroundJob/JobList.php
```diff
@@ -382,4 +383,21 @@ public function resetBackgroundJob(IJob $job): void {
 			->where($query->expr()->eq('id', $query->createNamedParameter($job->getId()), IQueryBuilder::PARAM_INT));
 		$query->executeStatement();
 	}
+
+	public function hasReservedJob(?string $className = null): bool {
+		$query = $this->connection->getQueryBuilder();
+		$query->select('*')
+			->from('jobs')
+			->where($query->expr()->neq('reserved_at', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT)));
```
nickvergessen, line 391:
```suggestion
			->where($query->expr()->neq('reserved_at', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT)))
			->setMaxResults(1);
```

lib/public/BackgroundJob/Job.php
```diff
@@ -132,6 +138,25 @@ public function getArgument() {
...
 	}
 
+	/**
+	 * Set this to false to prevent two Jobs from this class from running in parallel
+	 *
+	 * @param bool $allow
+	 * @return void
+	 * @since 27.0.0
+	 */
+	public function setAllowParallelRuns(bool $allow) {
```
nickvergessen, line 148: Add to `OCP\BackgroundJob\IJob` ?
Same for the get?

nickvergessen: Or to be saver add them to a new interface that is optional and then checked by the joblist before trying to use the methods

nickvergessen: ISingletonJob? 😝 
I'm really bad with naming

marcelklehr: mh, not sure this is necessary, because the check is in OCP\BackgroundJob\Job itself

marcelklehr: But we can still make a separate interface and let Job implement that in addition to IJob

lib/private/BackgroundJob/JobList.php
```diff
@@ -382,4 +383,22 @@ public function resetBackgroundJob(IJob $job): void {
...
+			->from('jobs')
+			->where($query->expr()->neq('reserved_at', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT)))
+			->setMaxResults(1);
+
+		if ($className !== null) {
+			$query->andWhere($query->expr()->eq('class', $query->createNamedParameter($className)));
+		}
+
+		try {
+			return $query->executeQuery()->rowCount() > 0;
```
marcelklehr, line 399: I wasn't sure whether this is good, or if I should use a COUNT(*) query directly

lib/public/BackgroundJob/Job.php
marcelklehr: Maybe IParallelJob? Since it provides an option to control parallelity?

marcelklehr: Or IJObWithParallelityOptions

lib/private/BackgroundJob/JobList.php
nickvergessen: well with setMaxResults only 1 row is get.
That's also good, as after 1 row in a lets say 7m rows oc_jobs table we can abort the query as there is a result. We don't need to load all.

lib/public/BackgroundJob/Job.php
nickvergessen: > But we can still make a separate interface and let Job implement that in addition to IJob

Yeah that's what I meant

tests/lib/BackgroundJob/TestJob.php
marcelklehr: This results in two new test failures. I think the culprit is that OC\BackgroundJob\Job catches all \Throwable errors while OCP\BackgroundJob\Job only catches \Exception errors. Should I change OCP\BackgroundJob\Job ?

marcelklehr: Another issue is that the OCP version doesn't allow injecting a logger :(

lib/private/BackgroundJob/JobList.php
```diff
@@ -382,4 +383,23 @@ public function resetBackgroundJob(IJob $job): void {
...
+			->where($query->expr()->neq('reserved_at', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT)))
+			->setMaxResults(1);
+
+		if ($className !== null) {
+			$query->andWhere($query->expr()->eq('class', $query->createNamedParameter($className)));
+		}
+
+		try {
+			$result = $query->executeQuery();
+			return count($result->fetchAll()) > 0;
```
nickvergessen, line 400: we only load 1 entry, so you could also do:
```suggestion
			return $result->fetch() !== false;
```

Also I kind of die when `$result->closeCursor()` is not called.

tests/lib/BackgroundJob/TestJob.php
```diff
@@ -20,7 +22,8 @@ class TestJob extends \OC\BackgroundJob\Job {
 	 * @param JobTest $testCase
 	 * @param callable $callback
 	 */
-	public function __construct($testCase = null, $callback = null) {
+	public function __construct(ITimeFactory $time = null, $testCase = null, $callback = null) {
```
nickvergessen, line 25: If you add it as first argument all previous occurrences still need touching?

nickvergessen (approved):

nickvergessen:
you need to rebase on master to fix the red ci

marcelklehr:
Didn't help.

skjnldsv:
> Didn't help.

https://github.com/nextcloud/server/pull/37858

nickvergessen:
So time to rebase again

lib/private/BackgroundJob/JobList.php
```diff
@@ -382,4 +383,25 @@ public function resetBackgroundJob(IJob $job): void {
...
+		}
+
+		try {
+			$result = $query->executeQuery();
+			$hasReservedJobs = $result->fetch() !== false;
+			$result->closeCursor();
+			return $hasReservedJobs;
+		} catch (Exception $e) {
+			return false;
+		}
```
come-nc, line 405: Log it, at least in debug?

tests/lib/BackgroundJob/DummyJobList.php
```diff
@@ -135,6 +140,14 @@ public function setLastRun(IJob $job): void {
 		$job->setLastRun(time());
 	}
 
+	public function hasReservedJob(?string $className = null): bool {
+		return $this->reserved[$className];
+	}
+
+	public function setHasReservedJob(?string $className, bool $hasReserved): void {
+		$this->reserved[$className] = $hasReserved;
+	}
```
come-nc, line 149:
```suggestion
	public function hasReservedJob(?string $className = null): bool {
		return $this->reserved[$className ?? ''];
	}

	public function setHasReservedJob(?string $className, bool $hasReserved): void {
		$this->reserved[$className ?? ''] = $hasReserved;
	}
```

I did not know null was allowed as array key, it appears to be silently casted to string and so it works, but I would feel better if the cast is explicit.

tests/lib/BackgroundJob/JobListTest.php
```diff
@@ -244,4 +245,24 @@ public function testSetLastRun() {
...
+	public function testHasReservedJobs() {
+		$this->clearJobsList();
+		$job = new TestJob($this->timeFactory, $this, function () {
+			$this->assertTrue($this->instance->hasReservedJob());
+			$this->assertTrue($this->instance->hasReservedJob(TestJob::class));
+		});
+		$this->instance->add($job);
+
+		$this->assertFalse($this->instance->hasReservedJob());
+		$this->assertFalse($this->instance->hasReservedJob(TestJob::class));
```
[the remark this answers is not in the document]
marcelklehr: Yep, adding a job does not reserve it. Only, when cron.php starts the job is it set as reserved.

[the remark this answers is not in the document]
marcelklehr: Reserved at X basically means "expected to be running since / starting from X"

come-nc (approved):
We have both `TestJob.php` and `JobTest.php` files in the same folder xD

marcelklehr:
Can we merge this? CI failure seems unlreated.

PR #37864, opened 2023-04-21
feat(ratelimit): Add Attributes support to rate limit middleware

nickvergessen:
## Summary

* Sibling to #36928

### Previously (and still supported)
```php
	/**
	 * @UserRateThrottle(limit=20, period=200)
	 * @AnonRateThrottle(limit=10, period=100)
	 */
	public function testMethodWithAnnotation() {
```

### New syntax
```php
	#[UserRateLimit(limit: 20, period: 200)]
	#[AnonRateLimit(limit: 10, period: 100)]
	public function testMethodWithAttribute() {
```

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required https://github.com/nextcloud/documentation/pull/10100


kesselb (commented):
Nice :+1:

lib/public/AppFramework/Http/Attribute/ARateLimit.php
```diff
@@ -0,0 +1,59 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCP\AppFramework\Http\Attribute;
+
+use Attribute;
```
come-nc, line 28:
```suggestion
```

come-nc (approved):

ChristophWurst:
> Previously (and still supported)

I would prefer to deprecate the old system right away so we can move everyone to attribute in the next few years and drop the legacy system as soon as possible.

nickvergessen:
> I would prefer to deprecate the old system right away so we can move everyone to attribute in the next few years and drop the legacy system as soon as possible.

Yeah that was my plan. The problem is I didn't find a way (without totally spaming out of control on debug) to notify devs about the deprecation. So yeah a section in the "Changes for 27" will be added for BruteForceProtection and RateLimit later on

lib/public/AppFramework/Http/Attribute/ARateLimit.php
nickvergessen: Fixed

vitormattos (approved):

lib/public/AppFramework/Http/Attribute/ARateLimit.php
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Attribute for controller methods that want to limit the times a logged-in
+ * user can call the endpoint in a given time period.
+ *
+ * @since 27.0.0
+ */
+abstract class ARateLimit {
+	/**
+	 * @since 27.0.0
+	 */
```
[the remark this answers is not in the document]
ChristophWurst: `@since` is required for all functions, classes and methods in the `OCP\` namespace

PR #37870, opened 2023-04-21
Fix table view

### LINKED ISSUE #36921: [BITV] 9.1.3.1e/8.1 - The user table represents a grid element and should be implemented as such. (1)
### ⚠️⚠️⚠️ Regression after table implementation: https://github.com/nextcloud/server/pull/38424

At the moment, the data is implemented in an unstructured manner using "div" elements. More information on grid implementation can be found at the following URLs: https://sarahmhigley.com/writing/grids-part1/, https://sarahmhigley.com/writing/grids-part2/ or https://www.w3.org/WAI/ARIA/apg/patterns/grid/. 

- [x] https://github.com/nextcloud/server/pull/39050
- ~https://github.com/Akryum/vue-virtual-scroller/pull/825~

### If upstream additions can't be made
- [x] https://github.com/nextcloud/server/pull/40719

4b98244143224a248f230b7ff75c24c8

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-3f0ff0bc0e-v8-n1

</details>

JuliaKirschenheuter:
* Resolves: https://github.com/nextcloud/server/issues/36921

## Summary

Create a grid for users view.

One visual change: made input fields visible in editing mode.

| Peek 2023-04-26 10-34 | Peek 2023-04-26 10-32

**AFTER STRUCTURE CHANGES (TABLE) STYLES RESPONSIBLE FOR ACTIONS WHILE EDITING A ROW WERE BROKEN**
Related (same issue here): https://github.com/nextcloud/server/pull/37860

I've adapted styles and now action menu is not broken anymore. But i had to remove `position: sticky` for it. 

| Peek 2023-05-10 14-54 | Peek 2023-05-10 14-56

**It is absolutely valid to have a vertical scrolling inside of a table. `position:sticky` in this case causes too much problems and because of it i would like to leave implementation like this.**

## Checklist


artonge (commented):
Could the input fields be a little longer to fill up empty spaces?

artonge (approved):
Looks good anyway :+1:

apps/settings/src/components/UserList.vue
```diff
@@ -642,4 +652,8 @@ export default {
 		 * prevent it). */
 		width: 0;
 	}
+
+	#app-content tbody tr:hover, #app-content tbody tr:focus, #app-content tbody tr:active {
+		background-color: var(--color-main-background);
+	}
```
Pytal, line 658:
```suggestion
	#app-content tbody tr {
	  &:hover,
	  &:focus,
	  &:active {
	    background-color: var(--color-main-background);
	  }
	}
```

JuliaKirschenheuter:
> Is the `developing` label still applicable @JuliaKirschenheuter?

Thanks for remark

artonge (approved):

Pytal (approved):

ChristophWurst:
COnflicts. CI stopped to free resources for mergeable PRs

juliushaertl:
I think removing the stickyness of the right action buttons is quite bad for a general user experience. Can you elaborate on the reasons why sticky is not working?

Also cc @jancborchardt for that

JuliaKirschenheuter:
> I think removing the stickyness of the right action buttons is quite bad for a general user experience. Can you elaborate on the reasons why sticky is not working?

Sticky on master was already broken. Because z-indexes were wrong implemented, please look in: 
https://github.com/nextcloud/server/pull/37860#issuecomment-1517710248

> Current style was a bit incorrect:
.userActions had z-index: 10
.popovermenu had z-index: 110
.userActions is a parent of .popovermenu
A child cannot have z-index higher than parent
For position: relative we could just remove z-index: 10. But with position: sticky it works different... position: sticky always creates a new [stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context). It is possible neither to remove z-index from sticky parent, nor to make child with higher z-index...
If I'm not wrong, the only way to keep sticky here is to use another container for NcActions here. For example, #app-content. But I am really not sure, it positions element correct with sticky...

Current master:

Peek 2023-05-10 15-14

ShGKme:
Proposals regarding sticky + z-index issue:

**Render `NcActions` outside the table row**

- Props:
  - I'd expect it to be easy to implement
- Cons:
  - Requires rewriting of some integration tests because row buttons will be outside of the row

**Generate progressive z-index - set each next row less `z-index` value**

- Props:
  - Keep the current structure
- Cons:
  - More complex implementation
  - Not sure about performance and large tables

AndyScherzinger:
@juliushaertl any suggestion how to improve this? Also @jancborchardt 
The table is needed for a11y reasons, yet for the moment stickiness got removed but was already broken before. So I would then rather like to get the table issue fixed via this PR and have a follow-up fixing the stickiness which would not necessarily have to be done by the people working on a11y improvements.

JuliaKirschenheuter:
Ok, sticky issue has been already existed. I would suggest to move forwards without `sticky` for now and i will open an issue to create action buttons with right `sticky` implementation.

jancborchardt (approved):
Looks good design-wise! :)

ShGKme (approved):
Loos good by code, tested on light and dark theme, different sizes, desktop and mobile.

Found a problem with the table header sticking, but it also appears on the master.

AndyScherzinger (approved):

ChristophWurst:
two regressions

1) table doesn't align, td columns float to the right
2) actions background element doesn't match row background

Bildschirmfoto vom 2023-05-12 12-02-14

ChristophWurst:
^ @JuliaKirschenheuter

Issue #36943, opened 2023-03-01
[BITV] 9.1.3.2/3.4 - Modal Dialog "Choose target folder" - In the list view, the contents of the "Modified" column are no longer available to both visual and non-visual users because "display:none" was used. (2)


AndyScherzinger:
- [x] https://github.com/nextcloud/server/pull/37879

<img width="450" alt="2023-03-01 18_34_37-Product Management - Files - Nextcloud – Mozilla Firefox" src="https://user-images.githubusercontent.com/1315170/222218042-c31afb78-3ad6-4a90-a3f8-1aebf127c72b.png">

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-1782286125-v3-n4

</details>

- [x] 9\.1.3.2/10.4 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-1782286125-v10-n4

PR #37879, opened 2023-04-22
Show modified in filepicker grid

Pytal:
- For https://github.com/nextcloud/server/issues/36943

## Summary

Before | After
--- | ---
image | image

## Checklist


Jerome-Herbinet (commented):
Technically, for me, it's OK. I tested on a Nextcloud 26 instance which has the same problem as Nextcloud 27 for which this Pull Request is initially intended. Before I give my approval, I wonder why this is done here, but not also in the grid view of all Nextcloud Files (which in this case will require a different code change presumably). In my opinion, there should be consistency between this copy/move popup and Nextcloud Files at large (unless this is already planned in the Nextcloud 27 developments, and I am not aware of it). What do you think about this ?

Pytal:
This scoped change is only for the filepicker and it is true that this change does not also show it in the files grid view but that is because the code is, as you guessed, separate

I agree that there should be consistency but as the files and filepicker grid views already diverge, i.e. files does not show the size while filepicker already does, I'd like to not scope creep this beyond what is described by the PR title @Jerome-Herbinet :)

Jerome-Herbinet (approved):

szaimen (approved):
Tested and works

susnux (approved):
Works and resolved the specific issue :)

nimishavijay (approved):
Agreed with @Jerome-Herbinet about the consistency, so any adjustments in files can be a separate PR :) Look great, nice work!

JuliaKirschenheuter (approved):

Issue #37716, opened 2023-04-13
Cors origin allowed list check


aleixq:
I think is useful to be able to operate with data from Nextcloud from other websites cms such as moodle, drupal, wp's while using different domains. The app webappassword lets use the DAV resources, it uses the sabre pluggable system to modify the response checking against a list of allowed domains. But I cannot find a way to use share api from different domain as files_sharing app is not providing neither the necessary preflight OPTIONS route, nor the `@CORS` annotation in the relevant controller methods. Also using it will allow every domain to gain access, so I think it may open security breach in sites.

## workaround for parts of 
It's discussed also in https://github.com/digital-blueprint/webapppassword/issues/1 , but I cannot find a way to listen for kernel-response events like https://symfony.com/doc/current/reference/events.html#kernel-response (as using symfony events is discouraged). So the workaround I end up using to be able to make requests from other domain is extending the shareapi controller to modify the response( https://gitlab.com/communia/files_sharing_webapppassword  information about it is in app issue tracker:  ). Of course, feedback about the ugliness of this workaround is welcome.

## proposal
Maybe nextcloud could provide it natively, modifying the existing behavior in https://github.com/nextcloud/server/blob/master/lib/private/AppFramework/Middleware/Security/CORSMiddleware.php#L116 to use a similar logic as I am using in the checkOrigin trait (https://gitlab.com/communia/files_sharing_webapppassword/-/blob/main/lib/Controller/AccessControl.php#L25 ) . 

One thing that will remain is how to define somewhere the automatic adding of preflighted OPTIONS route if cors is used...

szaimen:
Hi, did you know about this app?
https://apps.nextcloud.com/apps/webapppassword

aleixq:
Hi, yes, but it has some limitations, as it only covers webdav and with a commit I have recently contributed it also adds share api support. 
But as I said it has some limitations, it needs to reimplement every controller methods of the desired endpoint, so we(me and main developer) are agree in https://github.com/digital-blueprint/webapppassword/pull/61#discussion_r1169930354  that a nextcloud native  way of offfering this will be better. 
what do you think?

aleixq:
[...]

PR #37896, opened 2023-04-23
Draft proposal of allowed origins cors list

aleixq:
* Covers: #37716

## Summary

This is a WIP draft PR, that partially implements allowed origins list, to not open everywhere the nextcloud ocs and external api, and allowing to open some parts as the interaction will be limited. This is planned in a way that the desired endpoints will use the apicontroller instead of the base controller, so they can use the inherited preflighted_cors method.

When using so and adding the annotation `@cors` in desired  method, filter will be done, The definition of the allowed origins will be done in system and in the (to be defined) app in a general and in a app manner, so for example in case of files_sharing will check in config.php for list of allowed_origins, then in (tbd) app settings,  then will check also in config.php  the value of files_sharing.allowed_origins and in (tbd) app settings the value of files_sharing.allowed_origins.

If http_origin header matches any entry in these settings will allow the use of api from this origin url.

As example I am allowing the preview endpoint only but it culd be opened to other endpoints easily, just grab the code and define  

```
  'allowed_origins' => array(
    'https://example.com',
  ),
```
in config.php

then check if you can get a preview when calling from the allowed origin url (a simple js fetch can be done).

The frontend to control allowed_origins is yet to be done depending on how accepted is this draft or if you think it's a bad and ugly way of achieving the filtering.

## todo in this pr


aleixq:
Adding an application to be used as frontend to set the allowed origins, it's in separated repo at: https://github.com/Communia/cors_origin_filter_settings  . I am marking this PR as it needs review right now although it needs to add and modify tests, let me know if it's better to set it back to DRAFT.

vitormattos:
Hi @aleixq is this PR WIP, as mentioned at body of descriptio?

If is WIP, I think that would be good convert to draft and change the label to `2. developing`
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request

aleixq:
Thank you for caring, vitor.  Yes, the code could be reviewed and its main idea evaluated: 

- Adding a way to control the origins which can interoperate with nextcloud endpoints and how it's achieved: applying whitelists to all `@cors` annotated controller methods.

But there are parts not done yet:

- Tests.
- Documentation.
- Also, as there will be more control over the origins, opening more endpoints could be proposed (like search endpoints, the activity api or everything that could be used by a 3rd party web)  as it's already done in this pr with preview endpoint just to show how it could be done.

Just looking for feedback before continue the work on this. As I don't want to waste everyone's time with a lot of code if it's futile.  Said this, if it's better to set back to draft until everything is done, no problem.

aleixq:
@skjnldsv I suppose that is closed in favor of https://github.com/nextcloud/server/pull/40537 , these are good news! btw a simple comment to note this would help understand the close.

skjnldsv:
@aleixq thanks for the ping!
Yes and no! We did a cleanup pass on old PRs that looked stale. I wasn't able to comment on all of the, but nothing being definitive on Github, I counted on devs pinging me or the nc team if they needed extra help! So thank you for this! :hugs: 

https://github.com/nextcloud/server/pull/40537 looks more active indeed, if you think this is also a good approach to the initial PR, then great news, we'll try to move the other one forward.

If you think this one still apply, maybe you're willing to take over and push it forward?
It's not always easy to get a PR to completion amongst the wide variety of other tickets :)
Hope you can understand
Cheers, John

aleixq:
Thank's for your comment. You are true about the activity, and tbh it's more broadly implemented, so better let's close this and wait and test how this will land in nextcloud and then suggest new features if it's still needed (in fact I am apporting my impressions there right now).

skjnldsv:
Thank you for taking the time to reply! :)
We'll try our best to make things happen and move forward!!

PR #37904, opened 2023-04-24
fix: Use proper link for navigating to files on click handler

juliusknorr:
Fixes the file link preview where the click for files might have triggered a navigation to an invalid URL.

We now take the same location as the href attribute and navigate to it if viewer is not supported for the mimetype.


marcelklehr (approved):
:elephant:

nickvergessen (approved):

juliushaertl:
> Needs a recompile?

Yep, I tend to wait until reviewed to avoid rebase dances :)

PR #37923, opened 2023-04-25
Rely on getconf() instead of /proc/cpuinfo

solracsf:
## Summary

`getconf()` is more compatible and less `open_basedir()` error prone.

## Checklist

### DIFF
--- a/lib/private/Preview/Generator.php
+++ b/lib/private/Preview/Generator.php
@@ -297,7 +297,7 @@ class Generator {
 		static $width;
 		if (!isset($width)) {
 			if (function_exists('getconf')) {
-				$width = is_int(getconf('_NPROCESSORS_ONLN')) ?? is_int(getconf('NPROCESSORS_ONLN')) ?? 0;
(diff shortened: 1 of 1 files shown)


lib/private/Preview/Generator.php
```diff
@@ -296,8 +296,8 @@ public static function unguardWithSemaphore($semId): bool {
 	public static function getHardwareConcurrency(): int {
 		static $width;
 		if (!isset($width)) {
-			if (is_file("/proc/cpuinfo")) {
-				$width = substr_count(file_get_contents("/proc/cpuinfo"), "processor");
+			if (function_exists('getconf')) {
+				$width = getconf('_NPROCESSORS_ONLN');
```
kesselb, line 300: getconf is not available on my local php. Does it belong to an extension?

solracsf: Oh, you're right! I'm not using a standard PHP and, indeed, this module is not a common one.

Issue #20515, opened 2020-04-16
File Scan not possible if nextcloud is in maintenance mode


soccermax215:
### Steps to reproduce
1. Put nextcloud in maintenance mode
2. try to start files scan `sudo -u www-data php occ files:scan`

### Expected behaviour
- files scan should be possible if nextcloud is in maintenance mode
- as it's currently not possible to do so, an appropriate error message should be displayed
- there should be a piece of information about this behavior in the documentation of occ

### Actual behaviour
- error message if you start a files scan in maintenance mode: ` There are no commands defined in the "files" namespace.`

### Server configuration

**Operating system:** Ubuntu 18.04

**Web server:**: Apache/2.4.29 (Ubuntu)

**PHP version:** PHP 7.3.16-

**Nextcloud version:** latest - 18.0.3

**Updated from an older Nextcloud/ownCloud or fresh install:** updated from owncloud, OC 10->NC 12->13->14->15->16->17->18

**Where did you install Nextcloud from:** https://download.nextcloud.com/server/releases/latest-18.tar.bz2

Fluxenator:
Same issue here.

Operating System: Ubuntu 20.04

NC version: 18.0.8

drewjosh:
Same problem but with Nextcloud 20.0.6 without being in the maintenance mode.

pelMEN:
same problem 20.8
start scan then send serv to maintenance mode.

PR #37925, opened 2023-04-25
fix: change maintenance mode info wording

kesselb:
* Resolves: # 

## Summary

```
Nextcloud is in maintenance mode, hence the database isn't accessible.
Cannot perform any command except 'maintenance:mode --off'
```

A customer brought to our attention that the message is misleading.

In maintenance mode the database is accessible and it's possible to execute other commands then maintenance:mode

image

The reason we changed the message was https://github.com/nextcloud/server/issues/20515. 
I restored the old wording and added another line to make clear that some commands are unavailable.

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Console/Application.php
+++ b/lib/private/Console/Application.php
@@ -178,17 +178,13 @@ class Application {
 	 * for writing outputs.
 	 * @return void
 	 */
-	private function writeMaintenanceModeInfo(
-		InputInterface $input, ConsoleOutputInterface $output
-	) {
+	private function writeMaintenanceModeInfo(InputInterface $input, ConsoleOutputInterface $output): void {
 		if ($input->getArgument('command') !== '_completion'
 			&& $input->getArgument('command') !== 'maintenance:mode'
 			&& $input->getArgument('command') !== 'status') {
 			$errOutput = $output->getErrorOutput();
-			$errOutput->writeln(
-				'<comment>Nextcloud is in maintenance mode, hence the database isn\'t accessible.' . PHP_EOL .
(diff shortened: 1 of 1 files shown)


szaimen (approved):
That indeed looks like a better wording, thanks!

miaulalala (approved):

lib/private/Console/Application.php
```diff
@@ -178,17 +178,13 @@ public function loadCommands(
...
 		if ($input->getArgument('command') !== '_completion'
 			&& $input->getArgument('command') !== 'maintenance:mode'
 			&& $input->getArgument('command') !== 'status') {
 			$errOutput = $output->getErrorOutput();
-			$errOutput->writeln(
-				'<comment>Nextcloud is in maintenance mode, hence the database isn\'t accessible.' . PHP_EOL .
-				'Cannot perform any command except \'maintenance:mode --off\'</comment>' . PHP_EOL
-			);
+			$errOutput->writeln('<info>Nextcloud is in maintenance mode, no apps are loaded.</info>');
+			$errOutput->writeln('<info>Commands provided by apps are unavailable.</info>');
```
nickvergessen, line 187: info is green, comment is yellow colored:
https://github.com/nextcloud/3rdparty/blob/fcc80609824be396681d5a68ba495f9b85c9b0bb/symfony/console/Formatter/OutputFormatter.php#L74-L77

I think yellow matches better to the warning a like content.

kesselb: Done

nickvergessen (approved):

PR #37937, opened 2023-04-26
Remove job from oc_jobs when the file is not findable

artonge:
When an application is disabled, or when a background jobs is removed by the app developer, then the job won't be found at execution time.
In those cases, we return null, but it makes sense to remove those jobs from oc_job.

If the app was disabled, the job will be re-added if the app is enabled again.

### DIFF
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -307,7 +307,8 @@ class JobList implements IJobList {
 					$class = $row['class'];
 					$job = new $class();
 				} else {
-					// job from disabled app or old version of an app, no need to do anything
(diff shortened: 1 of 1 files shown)


lib/private/BackgroundJob/JobList.php
```diff
@@ -308,6 +308,7 @@ private function buildJob(array $row): ?IJob {
 					$job = new $class();
 				} else {
 					// job from disabled app or old version of an app, no need to do anything
```
solracsf, line 310: Maybe also remove this ", no need to do anything"?

icewind1991 (approved):

blizzz (approved):

PR #37609, opened 2023-04-06
Implement systemtags through webdav

tobiasKaminsky:
Exposes system tags under the `{http://nextcloud.org/ns}system-tags` property.

Issue #6424, opened 2017-09-08
Indicator for tagged files in list


hajika:
It looks like there is no hint or any sign, when a file has tags or comments. You first have to open the details page to check if there is something. It would much more useful, if an icon or anything else would indicate that there are (unread?) comments available for a specific file,

The same is for tags, it could be useful to have an optional column for them (thats maybe a new issue..?)

LukasReschke:
cc @nextcloud/designers

skjnldsv:
It would be a great addition indeed!

jancborchardt:
New comments are indicated in the file list by an icon on the respective file.

Yes, we could also add the tags a file has been tagged with if it has any:
- right next of the filename, like Gmail
- clicking the tag shows all files with this tag

cc @schiessle @blizzz @nextcloud/javascript

hajika:
> New comments are indicated in the file list by an icon on the respective file.

true! I didnt see it before, thats what I intended originally

![screenshot_2017-09-12_13-59-42](https://user-images.githubusercontent.com/27491258/30324837-b9ba62a4-97c2-11e7-9a2b-c141370c12d0.png)

It might be seen better, when the icon is not grey like the other ones but in a more attentive color.

jancborchardt:
True, we could experiment with making it full opacity. Want to contribute that as a pull request @hajika?

hajika:
I tried it with full opacity, Iam not fully convinced by the effect. And unfortunately css `fill` did not work on the `<img>` tag. Couldn't figure out yet why not.

rasos:
Could we even have the Details view open by default, or at least a config option to show the Details by default with Comments at all users? I find the Details view with Tags, Comments or photos with an immediate preview and the Metadata extremely useful- why should we always need to click on the three dots first (which are not very intuitive for e.g. leaving a comment).

![bildschirmfoto_2017-11-11_23-41-02](https://user-images.githubusercontent.com/485723/32694192-00fa3e72-c73a-11e7-977d-cdfba8553d18.png)

jancborchardt:
@rasos this is a separate discussion, please open a new issue. :)

We are talking about showing the sidebar by default on big screens: https://github.com/nextcloud/server/issues/7138

skjnldsv:
@jancborchardt
And how many tags should we display?
I don't think this is a good idea to use a popover if someone wanted to implement that. :)

hajika:
> to use a popover

[...]

PR #37938, opened 2023-04-26
feat(system-tags): show inline in files 

skjnldsv:
Rebase after https://github.com/nextcloud/server/pull/37609 is merged

### Light
localhost_8080_index php_apps_files__dir=_ fileid=2 (2)

### Dark
localhost_8080_index php_apps_files__dir=_ fileid=2 (1)


szaimen:
I've tested this and it seems to work in general :)

Some remarks:
- choosing a really long name might overlap things in certain resolutions:
  image
- newly added or removed labels via the sidebar will only be shown after a reload of the files app:
image

nimishavijay (commented):
> choosing a really long name might overlap things in certain resolutions:

Great catch, we can ellipsise for more than ~30 characters

No other comments from my side, it looks great :)

skjnldsv:
> newly added or removed labels via the sidebar will only be shown after a reload of the files app:

It depends how much time we actually want to spend on this.
It should be fixed with the new Files to Vue, but I'm not sure I should spend more time on this

jancborchardt (approved):
Looks awesome! :) I'd say the ellipsizing is good to fix as @nimishavijay said, but the refresh can be handled later.

skjnldsv:
image
Done

szaimen (approved):
> It depends how much time we actually want to spend on this.
> It should be fixed with the new Files to Vue, but I'm not sure I should spend more time on this

Fine by me then

fechnert:
This looks awesome, thank you for adding this! I'm very happy that it kinda looks like [my proposal](https://github.com/nextcloud/server/issues/6424#issuecomment-784198993). Awesome work!

skjnldsv:
Thanks for the positive feedback @fechnert :hugs:

jancborchardt:
@fechnert suggestions and contributions like that are always welcome! :)

gonzalo:
Awesome! They way that we use tags in our organization really would thank this new feature. We're still on 26, but planning to move to 27. Is this feature enabled by default in 27.0.5? or it will be in next releases?

jancborchardt:
@gonzalo if tags are enabled, this is how they will be shown by default from 27 on. :)

gonzalo:
Just looks awesome!!
image

Glandos:
This is really great, but I have one question before opening a new issue: is it normal that tags don't show up in a public link view? They are displayed in the list of a logged in user.

jancborchardt:
@Glandos at least in the current form and implementation, tags are considered information private to the instance. So they will indeed not show in the list, just like they didn’t show before in the link share view.

PR #37959, opened 2023-04-27
Check for open_basedir before reading /proc

### LINKED ISSUE #37921: [Bug]: flood of errors is_file(): open_basedir restriction in effect. File(/proc/cpuinfo) after 25 > 26 update

- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.

### Bug description

After updating Nextcloud server from 25.0.6 to 26.0 every activity results in an error in the log like:
```
[PHP] Error: is_file(): open_basedir restriction in effect. File(/proc/cpuinfo) is not within the allowed path(s): (/home/<account>/:/tmp:/var/tmp:/opt/alt/php74/usr/share/pear/:/dev/urandom:/usr/local/lib/php/:/usr/local/php74/lib/php/) at /home/<account>/domains/example.com/public_html/cloud/lib/private/Preview/Generator.php#351

GET /index.php/core/preview?fileId=464&c=c5edf881068c4f84bc6c9eb94f3cdcc9&x=375&y=375&forceIcon=0&a=1
from 2a10:3781:15:1:64d0:faf:896:88ea by <admin account> at 2023-04-23T20:44:04+00:00
```
This never occurred before the update (installed may 2022)

### Steps to reproduce

1. Install Nextcloud on shared hosting using the web installer
2. Use it for a year without notable issues
3. Update from 25.0.6 to 26.0
4. Look in Administration settings > Logging

### Expected behavior

Clean log or only warnings

### Installation method

Community Web installer on a VPS or web space

### Nextcloud Server version

26

### Operating system

RHEL/CentOS

### PHP engine version

PHP 8.1

### Web server

Other

### Database engine version

MySQL

Updated to a major version (ex. 22.2.3 to 23.0.1)

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
No SSH acces, but this is what the Support link provides:

Server configuration detail

Operating system: Linux 3.10.0-962.3.2.lve1.5.79.el7.x86_64 #1 SMP Wed Mar 15 09:10:44 UTC 2023 x86_64

Webserver: Apache/2 (litespeed)

Database: mysql 10.3.38

PHP version: 8.1.17

Modules loaded: Core, date, libxml, openssl, pcre, sqlite3, zlib, bz2, calendar, ctype, curl, hash, filter, ftp, gettext, json, iconv, SPL, pcntl, readline, Reflection, session, standard, mbstring, shmop, SimpleXML, tokenizer, xml, litespeed, bcmath, dom, fileinfo, gd, imagick, imap, intl, exif, mcrypt, mysqli, mysqlnd, PDO, pdo_mysql, pdo_sqlite, Phar, posix, soap, sockets, sodium, sysvsem, xmlreader, xmlwriter, xsl, zip, ionCube Loader, Zend OPcache

Nextcloud version: 26.0.1 - 26.0.1.1

{
    "instanceid": "***REMOVED SENSITIVE VALUE***",
    "trusted_domains": [
        "cloud.example.com"
    ],
    "datadirectory": "***REMOVED SENSITIVE VALUE***",
    "dbtype": "mysql",
    "version": "26.0.1.1",
    "overwrite.cli.url": "https:\/\/cloud.example.com",
    "dbname": "***REMOVED SENSITIVE VALUE***",
    "dbhost": "***REMOVED SENSITIVE VALUE***",
    "dbport": "",
    "dbtableprefix": "oc_",
    "mysql.utf8mb4": true,
    "dbuser": "***REMOVED SENSITIVE VALUE***",
    "dbpassword": "***REMOVED SENSITIVE VALUE***",
    "installed": true,
    "mail_smtpmode": "smtp",
    "mail_smtpsecure": "tls",
    "mail_sendmailmode": "smtp",
    "mail_smtphost": "***REMOVED SENSITIVE VALUE***",
    "mail_smtpport": "587",
    "mail_from_address": "***REMOVED SENSITIVE VALUE***",
    "mail_domain": "***REMOVED SENSITIVE VALUE***",
    "mail_smtpauthtype": "LOGIN",
    "mail_smtpauth": 1,
    "mail_smtpname": "***REMOVED SENSITIVE VALUE***",
    "mail_smtppassword": "***REMOVED SENSITIVE VALUE***",
    "maintenance": false,
    "theme": "",
    "loglevel": 2,
    "default_language": "nl",
    "default_locale": "nl_NL",
    "defaultapp": "files",
    "default_phone_region": "NL",
    "skeletondirectory": "\/home\/<account>\/domains\/example.com\/public_html\/clouddata\/skeleton",
    "templatedirectory": ""
}
```

### List of activated Apps

```shell
Enabled:
 - activity: 2.18.0
 - auto_groups: 1.5.1
 - bruteforcesettings: 2.6.0
 - circles: 26.0.0
 - cloud_federation_api: 1.9.0
 - comments: 1.16.0
 - contactsinteraction: 1.7.0
 - dashboard: 7.6.0
 - dav: 1.25.0
 - federatedfilesharing: 1.16.0
 - federation: 1.16.0
 - files: 1.21.1
 - files_external: 1.18.0
 - files_pdfviewer: 2.7.0
 - files_rightclick: 1.5.0
 - files_sharing: 1.18.0
 - files_trashbin: 1.16.0
 - files_versions: 1.19.1
 - firstrunwizard: 2.15.0
 - logreader: 2.11.0
 - lookup_server_connector: 1.14.0
 - nextcloud_announcements: 1.15.0
 - notifications: 2.14.0
 - oauth2: 1.14.0
 - password_policy: 1.16.0
 - photos: 2.2.0
 - privacy: 1.10.0
 - provisioning_api: 1.16.0
 - recommendations: 1.5.0
 - related_resources: 1.1.0-alpha1
 - settings: 1.8.0
 - sharebymail: 1.16.0
 - support: 1.9.0
 - survey_client: 1.14.0
 - systemtags: 1.16.0
 - text: 3.7.2
 - theming: 2.1.1
 - twofactor_backupcodes: 1.15.0
 - updatenotification: 1.16.0
 - user_status: 1.6.0
 - viewer: 1.10.0
 - workflowengine: 2.8.0
Disabled:
 - admin_audit
 - apporder: 0.15.0
 - backup: 1.2.0
 - encryption
 - groupfolders: 14.0.1
 - serverinfo: 1.14.0
 - suspicious_login
 - twofactor_totp
 - user_ldap
 - weather_status: 1.4.0
```

### Nextcloud Signing status

```shell
No errors have been found.
```

### Nextcloud Logs

```shell
Hundreds of:

[PHP] Error: is_file(): open_basedir restriction in effect. File(/proc/cpuinfo) is not within the allowed path(s): (/home/<account>/:/tmp:/var/tmp:/opt/alt/php74/usr/share/pear/:/dev/urandom:/usr/local/lib/php/:/usr/local/php74/lib/php/) at /home/<account>/domains/example.com/public_html/cloud/lib/private/Preview/Generator.php#351

GET /index.php/core/preview?fileId=464&c=c5edf881068c4f84bc6c9eb94f3cdcc9&x=375&y=375&forceIcon=0&a=1
from 2a10:3781:15:1:64d0:faf:896:88ea by <admin account> at 2023-04-23T20:44:04+00:00
```

### Additional info

See [this forum topic](https://help.nextcloud.com/t/open-basedir-restriction-after-update-to-26/158850). 

This has similarities with https://github.com/nextcloud/server/issues/27759 but that one is closed, was generated by a different .php and was about different directories so IMHO this a new bug.

Like on most shared hosting we cannot edit php.ini but the DirectAdmin panel allowed us to add /proc/cpuinfo to the open_basedir. I still believe this should not be needed.

Opinions I heard/read, but I'm no developer, so I cannot judge:

- Our provider: "open_basedir? Are they really using that?

In the forum topic mentioned above:

- "The problem was introduced by [this commit](https://github.com/nextcloud/server/commit/f9e9cd270dc1e6a7a6a7396b2f1af03512f81738#diff-1b6402d6a439b7e8fe8aa7289de2087cf0df27412f58c0fa39168a0f02592a37)"
- "cpuinfo is not the right wau to do this, [alternative suggested](https://help.nextcloud.com/t/open-basedir-restriction-after-update-to-26/158850/17)"

On php.net:

```
Caution

open_basedir is just an extra safety net, that is in no way comprehensive, and can therefore not be relied upon when security is needed.
```

solracsf:
* Resolves: https://github.com/nextcloud/server/issues/37921

## Summary

Check if `open_basedir` is set. If yes and is restrictive, or if can't be determined, set it to `0`.

## Checklist

### DIFF
--- a/lib/private/Preview/Generator.php
+++ b/lib/private/Preview/Generator.php
@@ -258,8 +258,20 @@ class Generator {
 	public static function getHardwareConcurrency(): int {
 		static $width;
 		if (!isset($width)) {
-			if (is_file("/proc/cpuinfo")) {
-				$width = substr_count(file_get_contents("/proc/cpuinfo"), "processor");
+			if (function_exists('ini_get')) {
+				$openBasedir = ini_get('open_basedir');
+				if ($openBasedir == '') {
+					$width = is_readable('/proc/cpuinfo') ? substr_count(file_get_contents('/proc/cpuinfo'), 'processor') : 0;
+				} else {
+					$openBasedirPaths = explode(':', $openBasedir);
+					foreach ($openBasedirPaths as $path) {
+						if (strpos($path, '/proc') === 0 || $path === '/proc/cpuinfo') {
+							$width = is_readable('/proc/cpuinfo') ? substr_count(file_get_contents('/proc/cpuinfo'), 'processor') : 0;
+						} else {
+							$width = 0;
+						}
+					}
+				}
 			} else {
 				$width = 0;
 			}
--- a/lib/private/Preview/Generator.php
+++ b/lib/private/Preview/Generator.php
@@ -257,20 +257,14 @@ class Generator {
 	 */
 	public static function getHardwareConcurrency(): int {
 		static $width;
+
 		if (!isset($width)) {
 			if (function_exists('ini_get')) {
 				$openBasedir = ini_get('open_basedir');
-				if ($openBasedir == '') {
+				if (empty($openBasedir) || strpos($openBasedir, '/proc/cpuinfo') !== false) {
 					$width = is_readable('/proc/cpuinfo') ? substr_count(file_get_contents('/proc/cpuinfo'), 'processor') : 0;
 				} else {
-					$openBasedirPaths = explode(':', $openBasedir);
-					foreach ($openBasedirPaths as $path) {
-						if (strpos($path, '/proc') === 0 || $path === '/proc/cpuinfo') {
-							$width = is_readable('/proc/cpuinfo') ? substr_count(file_get_contents('/proc/cpuinfo'), 'processor') : 0;
-						} else {
-							$width = 0;
-						}
-					}
+					$width = 0;
 				}
 			} else {
 				$width = 0;


lib/private/Preview/Generator.php
```diff
@@ -296,8 +296,20 @@ public static function unguardWithSemaphore($semId): bool {
...
 		if (!isset($width)) {
-			if (is_file("/proc/cpuinfo")) {
-				$width = substr_count(file_get_contents("/proc/cpuinfo"), "processor");
+			if (function_exists('ini_get')) {
+				$openBasedir = ini_get('open_basedir');
+				if ($openBasedir == '') {
+					$width = is_readable('/proc/cpuinfo') ? substr_count(file_get_contents('/proc/cpuinfo'), 'processor') : 0;
+				} else {
+					$openBasedirPaths = explode(':', $openBasedir);
+					foreach ($openBasedirPaths as $path) {
```
kesselb, line 305: It should also work to use strpos /proc and strpos /proc/cpuinfo on openbasedir without the explode :thinking:

kesselb:
I think we should remove getHardwareConcurrency and use a sane default.

A hint for `preview_concurrency_all` and `preview_concurrency_new` at https://docs.nextcloud.com/server/latest/admin_manual/installation/server_tuning.html would be nice. 

Every time a generator instance is loaded, we fetch /proc/cpuinfo to use the number of processor as default value. I think this is unnecessary.

We don't support FreeBSD. But TrueNAS which is based on FreeBSD. On FreeBSD /proc/cpuinfo is not available. We need a different way to retrieve the processors: https://github.com/nextcloud/serverinfo/blob/1146b21d13073c5e8799eb707096d12b5f682026/lib/OperatingSystems/FreeBSD.php#L81

My recommendation is to keep it simple.

joshtrichards:
This whole section of code `getHardwareConcurrency()` is semi-irrelevant anyhow since there are defaults defined and the admin can also define the best (preferred) parameters via `preview_concurrency_*` already. This is just fallback code and shouldn't be getting executed every time.

The overall generator CPU balancing feature is sound, but the code elsewhere within it could use some refactoring.

The only reason the open_basedir trigger in this issue can't easily be headed off by the admin simply configuring the `preview_concurrency_*` settings is because of the order of operations the fallback is implemented in.

I also wouldn't bother with trying to process the contents of the open_basedir if it's found. I'd just fall back to the sane defaults (which the admin can still override if they wish as currently coded). Plus even after parsing open_basedir the processing of /proc could still fail for other reasons (like different OSes as @kesselb noted).

That said, K.I.S.S. - and rather than submitting a competing PR - here's the quick and dirty path I suggest until a more extensive refactor can happen:

https://github.com/joshtrichards/nc-server/commit/30c5f2451eb0d5a018c580e8e7f9519d7d9ee6b3

solracsf:
I agree that simplication is (almost) ALWAYS the way to go, but this PR is only about open_basedir error suppression, nothing more. Feel free to submit a proper PR to simplify the whole thing.

joshtrichards (approved):
Seems okay to me.  Achieves the primary aim: allow some folks to add it to their open_basedir, but if they can't (or don't) it'll fall through to the defaults while still *also* permitting an explicit override to more appropriate local parameters by way of the `preview_concurrency_*` config parameters. And all without erroring out.

There are still some edge cases like if `open_basedir = "/tmp:/home/user:/proc"` which technically does permit `cpuinfo` access but would still fail this test, but given the above I don't think that matters nor is worth complicating the code further (plus it'd be sort of weird to lock things down with open_basedir yet then open up all of /proc IMO, but I've seen loads of weird things in the wild hah).

lib/private/Preview/Generator.php
kesselb: I was wrong. To check properly for /proc and/or /proc/cpuinfo we indeed need to iterate over the entries and make sure to detect if the whole folder is allowed.

icewind1991 (approved):

taziobaker:
Could this work?:

`public static function getHardwareConcurrency(): int {
    static $width;
    if (!isset($width)) {
        if (function_exists('nproc')) {
            // 'nproc' function is available, use it to determine the hardware concurrency
            $width = (int) nproc();
        } elseif (strtoupper(substr(PHP_OS, 0, 3)) === 'LIN') {
            // Use 'sysctl' to query the number of CPU cores on Linux
            $result = shell_exec('sysctl -n hw.ncpu');
            if ($result !== null) {
                $width = (int) trim($result);
            } else {
                $width = 0; // Handle the case when 'sysctl' is not available or fails.
            }
        } else {
            // Handle other platforms or cases
            $width = 0;
        }
    }
    return $width;
}`

This first checks if the nproc function is available. If it is, it uses nproc to determine the hardware concurrency. If nproc is not available and the operating system is Linux, it uses the sysctl approach to query the number of CPU cores. If neither method is available or if the code is running on a platform other than Linux, it sets the hardware concurrency to 0.

Just my ¢2

PR #37961, opened 2023-04-27
SystemTags endpoint to return tags used by a user with meta data

blizzz:
## Summary

Target case is photos app: when visiting the tags category, all systemtags of the whole cloud are retrieved. In subsequent steps the next tag is requested until the browser view is filled with tag tiles (i.e. previews are requested just as well).

With this approach, we incorporate the dav search and look for user related tags that are used by them, and already returns the statistics (number of files tagged with the respective tag) as well as a file id for the purpose to load the preview. This defaults to the file with the highest id.

Call:
```bash
curl -s -u 'user:password' \
  'https://my.nc.srv/remote.php/dav/systemtags-assigned' \
  -X PROPFIND -H 'Accept: text/plain' \
  -H 'Accept-Language: en-US,en;q=0.5'  -H 'Depth: 1' \
  -H 'Content-Type: text/plain;charset=UTF-8' \
  --data @/home/doe/request-systemtag-props.xml
```

With request-systemtag-props.xml:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<d:propfind xmlns:d="DAV:">
        <d:prop xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns">
                <oc:id/>
                <oc:display-name/>
                <oc:user-visible/>
                <oc:user-assignable/>
                <oc:can-assign/>
                <nc:files-assigned/>
                <nc:reference-fileid/>
        </d:prop>
</d:propfind>
```

Example output:
```xml
  …
  <d:response>
    <d:href>/master/remote.php/dav/systemtags/84</d:href>
    <d:propstat>
      <d:prop>
        <oc:id>84</oc:id>
        <oc:display-name>Computer</oc:display-name>
        <oc:user-visible>true</oc:user-visible>
        <oc:user-assignable>true</oc:user-assignable>
        <oc:can-assign>true</oc:can-assign>
        <nc:files-assigned>42</nc:files-assigned>
        <nc:reference-fileid>924022</nc:reference-fileid>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
  <d:response>
    <d:href>/remote.php/dav/systemtags/97</d:href>
    <d:propstat>
      <d:prop>
        <oc:id>97</oc:id>
        <oc:display-name>Bear</oc:display-name>
        <oc:user-visible>true</oc:user-visible>
        <oc:user-assignable>true</oc:user-assignable>
        <oc:can-assign>true</oc:can-assign>
        <nc:files-assigned>1</nc:files-assigned>
        <nc:reference-fileid>923422</nc:reference-fileid>
      </d:prop>
      <d:status>HTTP/1.1 200 OK</d:status>
    </d:propstat>
  </d:response>
  …
```

**Update 23/05/04**

Added was support to retrieve tags for a specific media type. This works with the URI. So, while the `systemtags-current` root provides tag information across all files, querying `/remote.php/dav/systemtags-current/image` will return tag information only for all image-typed files. 

In the first commit, the image-type condition was hard coded.

**/Update**

**Update 23/05/04**

The endpoint was renamed from `systemtags-current` to `systemtags-assigned`. The information at the top (before the updates) are adjusted.

**/Update**

## TODO

- [ ] Performance testing with a serious data set 
- [ ] Reconsider pragmatic architectual choices (quick PoC)

## Checklist

### DIFF
--- a/apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
+++ b/apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Arthur Schiwon <blizzz@arthur-schiwon.de>
+ *
+ * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\DAV\SystemTag;
+
+use OC\SystemTag\SystemTag;
+use OCP\Files\IRootFolder;
+use OCP\IUserSession;
+use OCP\SystemTag\ISystemTagManager;
+use Sabre\DAV\Exception\Forbidden;
+
+class SystemTagsInUseCollection extends \Sabre\DAV\SimpleCollection {
+	protected IUserSession $userSession;
+	protected IRootFolder $rootFolder;
+
+	public function __construct(IUserSession $userSession, IRootFolder $rootFolder) {
+		$this->userSession = $userSession;
+		$this->rootFolder = $rootFolder;
+		$this->name = 'systemtags-current';
+	}
+
+	public function setName($name): void {
+		throw new Forbidden('Permission denied to rename this collection');
+	}
+
+	public function getChildren() {
+		$user = $this->userSession->getUser();
+		if ($user === null) {
+			throw new Forbidden('Permission denied to read this collection');
+		}
+
+		$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+		$result = $userFolder->getSystemTags('image');
+		$children = [];
+		foreach ($result as $tagData) {
+			$tag = new SystemTag((string)$tagData['id'], $tagData['name'], (bool)$tagData['visibility'], (bool)$tagData['editable']);
+			$node = new SystemTagNode($tag, $user, false, \OCP\Server::get(ISystemTagManager::class));
+			$node->setNumberOfFiles($tagData['number_files']);
+			$node->setReferenceFileId($tagData['ref_file_id']);
+			$children[] = $node;
+		}
+		return $children;
+	}
+}
--- a/apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
+++ b/apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
@@ -42,7 +42,7 @@ class SystemTagsInUseCollection extends \Sabre\DAV\SimpleCollection {
 		$this->userSession = $userSession;
 		$this->rootFolder = $rootFolder;
 		$this->mediaType = $mediaType;
-		$this->name = 'systemtags-current';
+		$this->name = 'systemtags-assigned';
 		if ($this->mediaType != '') {
 			$this->name .= '/' . $this->mediaType;
 		}
--- a/apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
+++ b/apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace OCA\DAV\SystemTag;
 
 use OC\SystemTag\SystemTag;
+use OC\SystemTag\SystemTagsInFilesDetector;
 use OC\User\NoUserException;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotPermittedException;
@@ -40,19 +41,22 @@ class SystemTagsInUseCollection extends SimpleCollection {
 	protected IUserSession $userSession;
 	protected IRootFolder $rootFolder;
 	protected string $mediaType;
-	private ISystemTagManager $systemTagManager;
+	protected ISystemTagManager $systemTagManager;
+	protected SystemTagsInFilesDetector $systemTagsInFilesDetector;
 
 	/** @noinspection PhpMissingParentConstructorInspection */
 	public function __construct(
 		IUserSession $userSession,
 		IRootFolder $rootFolder,
 		ISystemTagManager $systemTagManager,
+		SystemTagsInFilesDetector $systemTagsInFilesDetector,
 		string $mediaType = ''
 	) {
 		$this->userSession = $userSession;
 		$this->rootFolder = $rootFolder;
 		$this->systemTagManager = $systemTagManager;
 		$this->mediaType = $mediaType;
+		$this->systemTagsInFilesDetector = $systemTagsInFilesDetector;
 		$this->name = 'systemtags-assigned';
 		if ($this->mediaType != '') {
 			$this->name .= '/' . $this->mediaType;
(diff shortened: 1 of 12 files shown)


apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	public function getChildren() {
+		$user = $this->userSession->getUser();
+		if ($user === null) {
+			throw new Forbidden('Permission denied to read this collection');
+		}
+
+		$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+		$result = $userFolder->getSystemTags('image');
```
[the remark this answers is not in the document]
blizzz: Accurate, the method is not included in the Interface so far.

lib/private/Files/Node/Folder.php
```diff
@@ -204,44 +204,25 @@
...
-
-		// search is handled by a single query covering all caches that this folder contains
-		// this is done by collect
-
-		$limitToHome = $query->limitToHome();
-		if ($limitToHome && count(explode('/', $this->path)) !== 3) {
-			throw new \InvalidArgumentException('searching by owner is only allowed in the users home folder');
-		}
-
+	protected function getCachesAndMountpointsForSearch(bool $limitToHome = false): array {
```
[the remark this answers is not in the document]
blizzz: https://psalm.dev/docs/annotating_code/type_syntax/array_types/#list-shapes
I must oversee something?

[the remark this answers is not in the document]
marcelklehr: I don't see it either :/

blizzz:
Added support for media types and updated the information in the summary. 

Compared to the first commit, it's a little breaking, for filtering image files was hardcoded. To achive the same results run the query against `/remote.php/dav/systemtags-current/image`, otherwise tags relevant to any file type are returned.

marcelklehr:
Can we call the endpoint something more meaningful than systemtags-current?

AndyScherzinger (approved):
fine by me except other peoples comments of course. And would also love to have this in for 27 even if we might optimize further in the future or have a complete new solution for certain things regarding tags.

Also, given the time I think Beta2 would be fine.

marcelklehr (approved):

blizzz:
> Can we call the endpoint something more meaningful than systemtags-current?

We can, but do you have suggestions? Tbh, I am not super happy about it either, but was so far the best I could think of:

- "-current" says it is about tags that are in use at this time

And i disregarded following ideas, because

- "-used" could be appropriate, but it has a touch of the past, or could be misinterpreted as something trashed
- "-recent" could suggest that valid, but older tags would not be considered
- "-active" could be misleading (what is inactive?)
- "-user" could be misleading to be user-only tags

blizzz:
> fine by me except other peoples comments of course. And would also love to have this in for 27 even if we might optimize further in the future or have a complete new solution for certain things regarding tags.
>
> Also, given the time I think Beta2 would be fine.

Beta2 is fine, just one thing that is missing is also that we extend `Folder` with public methods, and that should be reflected in the PHP API. We should refrain from touching `\OCP\Files\Folder` as it has at least 8 implementations, which it would break. So we'll have to add a public interface, but when we do it, we should be sure about it.

Thus said I am not one hundred percent convinced our additions should stay in the Folder implementation, however there is some of the logic we need. We could have a SytemTag specific collection that could wrap it and so have little changes there, but would have a hard dependency on the Folder implementation (if only on few specific parts).

Or we expose what is consolidated here as `getCachesAndMountpointsForSearch` as PHP API, that can be then consumed by new SystemTag specific classes and have better separation.

Would love to have @icewind1991​'s opinion on it.

marcelklehr:
> We can, but do you have suggestions?

How about

* `systemtags-in-use`
* `systemtags-assigned`

blizzz:
> > We can, but do you have suggestions?
> 
> How about
> 
>     * `systemtags-in-use`
> 
>     * `systemtags-assigned`

`systemtags-assigned` sounds good to me, and yes, clearer than `current`.

apps/dav/lib/SystemTag/SystemTagsInUseCollection.php
```diff
@@ -0,0 +1,80 @@
...
+		if ($user === null) {
+			throw new Forbidden('Permission denied to read this collection');
+		}
+
+		$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+		$result = $userFolder->getSystemTags($this->mediaType);
+		$children = [];
+		foreach ($result as $tagData) {
+			$tag = new SystemTag((string)$tagData['id'], $tagData['name'], (bool)$tagData['visibility'], (bool)$tagData['editable']);
+			$node = new SystemTagNode($tag, $user, false, \OCP\Server::get(ISystemTagManager::class));
```
icewind1991, line 73: inject `ISystemTagManager`?

lib/private/Files/Node/Folder.php
```diff
@@ -332,6 +337,24 @@ public function searchByTag($tag, $userId) {
 		return $this->search($query);
 	}
 
+	/**
+	 *
+	 * @return array<array-key, array{id: int, name: string, visibility: int, editable: int, ref_file_id: int, number_files: int}>
+	 */
+	public function getSystemTags(string $mediaType, int $limit = 0, int $offset = 0): array {
```
icewind1991, line 344: this could use a more descriptive name, something like `getUsedSystemTagsInFolder` or something

lib/private/Files/Node/Folder.php
```diff
@@ -332,6 +337,24 @@ public function searchByTag($tag, $userId) {
...
+		// Currently query has to have exactly one search condition. If no media type is provided,
+		// we fall back to the presence of a systemtag.
+		if (empty($mediaType)) {
+			$query = $this->queryFromOperator(new SearchComparison(ISearchComparison::COMPARE_LIKE, 'systemtag', '%'), null, $limit, $offset);
```
icewind1991, line 348: does adding this operator do anything? `LIKE '%'` seems to be always true to me

icewind1991 (changes requested):
Overall approach seems fine.

Not a big fan of adding it as a public method to folder but not sure if there is a better option

lib/private/Files/Node/Folder.php
blizzz: It's a shortcut to get a query to pass to `$searchHelper->findUsedTagsInCaches()`. Certainly this can be solved more elegantly. It was born out of the need to have something ready to bring Magenta back to a working state/environment.

blizzz:
> Not a big fan of adding it as a public method to folder but not sure if there is a better option

@icewind1991 yes feeling the same. One way to reduce the footprint would be to expose `getCachesAndMountpointsForSearch` only (new interface like  `ISearchAwareFolder` or something like that), then the rest  of the logic can be moved to classes withing the SystemTags namespace.

lib/private/Files/Node/Folder.php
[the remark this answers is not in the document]
blizzz: this psalm version does not know such composed lists

blizzz:
> > Not a big fan of adding it as a public method to folder but not sure if there is a better option
> 
> @icewind1991 yes feeling the same. One way to reduce the footprint would be to expose `getCachesAndMountpointsForSearch` only (new interface like `ISearchAwareFolder` or something like that), then the rest of the logic can be moved to classes withing the SystemTags namespace.

But of course that leaves the changes within `OC\Files\Cache\`.  I have repressed it somehow, and from the hip not a quick solution for it. I've got a commit ready to do the former that I can push, if it makes a big difference.

Or we indeed keep it provide as it is for now, just to have it merged. And have time to rethink internal structure and the PHP API, and be it for 28. I am a bit concerned though that the follow up would compete with other important tasks.

lib/private/Files/Node/Folder.php
icewind1991: I suspect that this might interfere with picking an index so it would be better to have another way to flag that the query should join on thags.

blizzz:
@icewind1991 moved it to the search helper as suggested

lib/private/Files/Node/Folder.php
blizzz: The behaviour appears same, however (checking with Postgrest) at some we'll have an `Index Scan using oc_systemtag_pkey on oc_systemtag systemtag  (cost=0.14..13.54 rows=69 width=20)` where the poential cost is slighlty lower when the comparison is not included. 13.54 instead of 13.72.

blizzz:
failing drone CI tests unrelated

blizzz:
Will need manual backports to 26 and 25

blizzz:
backports created manually

PR #37968, opened 2023-04-28
add npm-audit-fix

szaimen:


skjnldsv (approved):

skjnldsv:
Fails on all branches @szaimen https://github.com/nextcloud/server/actions/runs/4832831016/jobs/8612090288 :see_no_evil:

szaimen:
This is due to the command not able to fix all problems. The PR still gets created nonetheless.

> Maybe we should put the compiling in a separate step that runs always?

PR #37969, opened 2023-04-28
use more efficient tag retrieval on DAV report request

blizzz:
## Summary

- uses DAV search approach against valid files joined by systemtag selector
- reduced table join for tag/systemtag search
- supports pagination
- no changes to the output formats or similar

Example request body:

```xml
<?xml version="1.0"?>
<oc:filter-files xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns" xmlns:ocs="http://open-collaboration-services.org/ns">
  <d:prop>
    <d:getcontentlength/>
    <d:getcontenttype/>
    <d:getetag/>
    <d:getlastmodified/>
    <d:resourcetype/>
    <nc:face-detections/>
    <nc:file-metadata-size/>
    <nc:has-preview/>
    <nc:realpath/>
    <oc:favorite/>
    <oc:fileid/>
    <oc:permissions/>
    <nc:nbItems/>
  </d:prop>
  <oc:filter-rules>
    <oc:systemtag>32</oc:systemtag>
  </oc:filter-rules>
  <d:limit>
    <d:nresults>50</d:nresults>
    <nc:firstresult>0</nc:firstresult>
  </d:limit>
</oc:filter-files>
```

Example call:
 
```bash
curl -s -u 'user:password' 'https://my.nxt.cld/remote.php/dav/files/user/' -X REPORT  -H 'Accept: text/plain'  -H 'Content-Type: text/plain;charset=UTF-8' --data @/home/nextcloud/report-people-tag.xml
```

:information_source: the base branch is #37961 This PR is an addition to it, but I keep it seperated for smaller changesets.

## TODO

- [ ] Performance testing with a serious data set
- [ ] Reconsider pragmatic architectual choices (quick PoC)

## Checklist

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
@@ -338,9 +338,15 @@ class FilesReportPlugin extends ServerPlugin {
 		// exposed in API yet
 		if (!empty($systemTagIds)) {
 			$tags = $this->tagManager->getTagsByIds($systemTagIds, $this->userSession->getUser());
+
+			// For we run DB queries per tag and require intersection, we cannot apply limit and offset for DB queries on multi tag search.
+			$oneTagSearch = count($tags) === 1;
+			$dbLimit = $oneTagSearch ? $limit ?? 0 : 0;
+			$dbOffset = $oneTagSearch ? $offset ?? 0 : 0;
+
 			foreach ($tags as $tag) {
 				$tagName = $tag->getName();
-				$tmpNodes = $this->userFolder->searchBySystemTag($tagName, $this->userSession->getUser()->getUID(), $limit ?? 0, $offset ?? 0);
+				$tmpNodes = $this->userFolder->searchBySystemTag($tagName, $this->userSession->getUser()->getUID(), $dbLimit, $dbOffset);
 				if (count($nodes) === 0) {
 					$nodes = $tmpNodes;
 				} else {
@@ -352,6 +358,10 @@ class FilesReportPlugin extends ServerPlugin {
 					// there cannot be a common match when nodes are empty early.
 					return $nodes;
 				}
+			}
+
+			if (!$oneTagSearch && ($limit !== null || $offset !== null)) {
+				$nodes = array_slice($nodes, $offset, $limit);
 			}
 		}
 
--- a/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
@@ -219,6 +219,10 @@ class FilesReportPlugin extends ServerPlugin {
 		// gather all file ids matching filter
 		try {
 			$resultFileIds = $this->processFilterRulesForFileIDs($filterRules);
+			// no logic in circles and favorites for paging, we always have all results, and slice later on
+			$resultFileIds = array_slice($resultFileIds, $offset ?? 0, $limit ?? null);
+			// fetching nodes has paging on DB level – therefore we cannot mix and slice the results, similar
+			// to user backends. I.e. the final result may return more results than requested.
 			$resultNodes = $this->processFilterRulesForFileNodes($filterRules, $limit ?? null, $offset ?? null);
 		} catch (TagNotFoundException $e) {
 			throw new PreconditionFailed('Cannot filter by non-existing tag', 0, $e);
--- a/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
@@ -27,7 +27,6 @@
  */
 namespace OCA\DAV\Connector\Sabre;
 
-use OC\Files\Node\LazyFolder;
 use OC\Files\View;
 use OCP\App\IAppManager;
 use OCP\Files\Folder;
@@ -46,7 +45,6 @@ use Sabre\DAV\Xml\Element\Response;
 use Sabre\DAV\Xml\Response\MultiStatus;
 
 class FilesReportPlugin extends ServerPlugin {
-
 	// namespace
 	public const NS_OWNCLOUD = 'http://owncloud.org/ns';
 	public const NS_NEXTCLOUD = 'http://nextcloud.org/ns';
@@ -335,12 +333,28 @@ class FilesReportPlugin extends ServerPlugin {
 		// exposed in API yet
 		if (
 			!empty($systemTagIds)
-			&& ($this->userFolder instanceof \OC\Files\Node\Folder
-				|| $this->userFolder instanceof LazyFolder)
+			&& (method_exists($this->userFolder, 'searchBySystemTag'))
 		) {
 			$tags = $this->tagManager->getTagsByIds($systemTagIds);
-			$tagName = (current($tags))->getName();
-			$nodes = $this->userFolder->searchBySystemTag($tagName, $this->userSession->getUser()->getUID(), $limit ?? 0, $offset ?? 0);
+			foreach ($tags as $tag) {
+				if (!$tag->isUserVisible()) {
+					// searchBySystemTag() also has the criteria to include only user visible tags. They can be skipped early nevertheless.
+					continue;
+				}
+				$tagName = $tag->getName();
+				$tmpNodes = $this->userFolder->searchBySystemTag($tagName, $this->userSession->getUser()->getUID(), $limit ?? 0, $offset ?? 0);
+				if (count($nodes) === 0) {
+					$nodes = $tmpNodes;
+				} else {
+					$nodes = array_uintersect($nodes, $tmpNodes, function (\OCP\Files\Node $a, \OCP\Files\Node $b): int {
+						return $a->getId() - $b->getId();
+					});
+				}
+				if ($nodes === []) {
+					// there cannot be a common match when nodes are empty early.
+					return $nodes;
+				}
+			}
 		}
 
 		return $nodes;
--- a/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
@@ -30,6 +30,7 @@ namespace OCA\DAV\Connector\Sabre;
 use OC\Files\View;
 use OCP\App\IAppManager;
 use OCP\Files\Folder;
+use OCP\Files\Node as INode;
 use OCP\IGroupManager;
 use OCP\ITagManager;
 use OCP\IUserSession;
@@ -238,7 +239,9 @@ class FilesReportPlugin extends ServerPlugin {
 		// find sabre nodes by file id, restricted to the root node path
 		$additionalNodes = $this->findNodesByFileIds($reportTargetNode, $resultFileIds);
 		if ($additionalNodes && $results) {
-			$results = array_intersect($results, $additionalNodes);
+			$results = array_uintersect($results, $additionalNodes, function (Node $a, Node $b): int {
+				return $a->getId() - $b->getId();
+			});
 		} elseif (!$results && $additionalNodes) {
 			$results = $additionalNodes;
 		}
@@ -344,7 +347,7 @@ class FilesReportPlugin extends ServerPlugin {
 				if (count($nodes) === 0) {
 					$nodes = $tmpNodes;
 				} else {
-					$nodes = array_uintersect($nodes, $tmpNodes, function (\OCP\Files\Node $a, \OCP\Files\Node $b): int {
+					$nodes = array_uintersect($nodes, $tmpNodes, function (INode $a, INode $b): int {
 						return $a->getId() - $b->getId();
 					});
 				}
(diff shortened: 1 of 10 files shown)


AndyScherzinger (approved):

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -209,13 +219,21 @@ public function onReport($reportName, $report, $uri) {
...
+		$results = [];
+		foreach ($resultNodes as $entry) {
+			if ($entry) {
+				$results[] = $this->wrapNode($entry);
+			}
+		}
+
 		// find sabre nodes by file id, restricted to the root node path
-		$results = $this->findNodesByFileIds($reportTargetNode, $resultFileIds);
+		$results = array_merge($this->findNodesByFileIds($reportTargetNode, $resultFileIds), $results);
```
marcelklehr, line 236: This changes the filter logic from an AND to an OR, afaict. (processFilterRules was using array_intersect before)

blizzz: You're right, good catch!

blizzz:
* [ ] breaks universal search by tags

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -209,13 +219,26 @@ public function onReport($reportName, $report, $uri) {
 
 		// gather all file ids matching filter
 		try {
-			$resultFileIds = $this->processFilterRules($filterRules);
+			$resultFileIds = $this->processFilterRulesForFileIDs($filterRules);
+			$resultNodes = $this->processFilterRulesForFileNodes($filterRules, $limit ?? null, $offset ?? null);
```
marcelklehr, line 223: Does this mean we only apply the limit and offset params to the tag search and not to e.g. results from the favorite property?

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -209,13 +219,26 @@ public function onReport($reportName, $report, $uri) {
...
+			if ($entry) {
+				$results[] = $this->wrapNode($entry);
+			}
+		}
+
 		// find sabre nodes by file id, restricted to the root node path
-		$results = $this->findNodesByFileIds($reportTargetNode, $resultFileIds);
+		$additionalNodes = $this->findNodesByFileIds($reportTargetNode, $resultFileIds);
+		if ($additionalNodes && $results) {
+			$results = array_intersect($results, $additionalNodes);
```
marcelklehr, line 238: I'm skeptical if this works. PHP docs say ["Two elements are considered equal if and only if (string) $elem1 === (string) $elem2. In words: when the string representation is the same. "](https://www.php.net/manual/en/function.array-intersect.php) The Sabre Node class doesn't seem to have a toString method, so (without having tested this) this might always reduce to one result or yield other unexpected results?

lib/private/Files/Cache/QuerySearchHelper.php
```diff
@@ -152,22 +177,20 @@ public function searchInCaches(ISearchQuery $searchQuery, array $caches): array
...
+						break;
+					case 'tagname':
+					case 'favorite':
+						$this->equipQueryForDavTags($query, $user);
+						break;
+				}
+			} elseif ($searchQuery->getSearchOperation() instanceof SearchBinaryOperator) {
+				$this->equipQueryForSystemTags($query);
+				$this->equipQueryForDavTags($query, $user);
+			}
```
marcelklehr, line 193: Can we, instead of giving up after one level, perhaps reuse the logic in shouldJoinTags, but have one for system tags and one for dav tags?

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
blizzz: Indeed. Method for Tags or Circles do not offer paging. So, the inconsistency is bad, but at least we should slice it  – by source. I consider it out of scope to touch favorites and circles handling.

But I had a closer look at the circles logic. It calls `\OCA\Circles\Api\v1\Circles::getFilesForCircles()` which is marked as deprecated and only throws a `BadMethodCallException`. This is an issue of its own.

lib/private/Files/Cache/QuerySearchHelper.php
blizzz: it stems from universal search where this is combined, but of course you might run other queries.

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -310,47 +321,29 @@ protected function processFilterRules($filterRules) {
...
-			}
+		// type check to ensure searchBySystemTag is available, it is not
+		// exposed in API yet
+		if (
+			!empty($systemTagIds)
+			&& ($this->userFolder instanceof \OC\Files\Node\Folder
+				|| $this->userFolder instanceof LazyFolder)
+		) {
+			$tags = $this->tagManager->getTagsByIds($systemTagIds);
+			$tagName = (current($tags))->getName();
```
blizzz, line 342: noticed, it is also legit to have a filter with multiple systemtags

blizzz: Current approach to solve this is very simple and linear: a query per each tag and intersect the results. this is not particularly satisfying, and also does not work with paging on DB level. 

When trying to use one search comparison per tag, the query result will always be empty as  there is just one systemtag per row after all the joins. 

It would work when adding the tag names into the join constraints, ruling out rows where the tag is null, grouping by the file id and having a condition on the number of tags, like
```sql
SELECT …
…
LEFT JOIN oc_systemtag systemtag ON (systemtag.id = systemtagmap.systemtagid) AND (systemtag.visibility = 1) AND ((systemtag.name = 'Foo') OR (systemtag.name = 'Bar'))
…

WHERE … AND systemtag.name IS NOT NULL 
GROUP BY file.fileid 
HAVING count(systemtag.name) = 2;
```

This might be more feasible to do, but also need to check performance impact.

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
blizzz: you're right, an `array_uintersect` it shall be.

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
blizzz: Going for the simple approach first after discussion with @juliushaertl

blizzz: Going for the simple approach first after discussion with @juliushaertl

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
blizzz: That was fixed recently in https://github.com/nextcloud/circles/pull/1318

blizzz: and the slicing is in.

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -310,47 +319,48 @@
...
+		// type check to ensure searchBySystemTag is available, it is not
+		// exposed in API yet
+		if (
+			!empty($systemTagIds)
+			&& (method_exists($this->userFolder, 'searchBySystemTag'))
+		) {
+			$tags = $this->tagManager->getTagsByIds($systemTagIds, $this->userSession->getUser());
+			foreach ($tags as $tag) {
+				$tagName = $tag->getName();
+				$tmpNodes = $this->userFolder->searchBySystemTag($tagName, $this->userSession->getUser()->getUID(), $limit ?? 0, $offset ?? 0);
```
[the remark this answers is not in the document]
marcelklehr: Can we fix this psalm error?

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -310,47 +324,38 @@ protected function processFilterRules($filterRules) {
...
-				// Empty intersection, nothing can show up anymore
-				return [];
+		// type check to ensure searchBySystemTag is available, it is not
+		// exposed in API yet
+		if (!empty($systemTagIds)) {
+			$tags = $this->tagManager->getTagsByIds($systemTagIds, $this->userSession->getUser());
+			foreach ($tags as $tag) {
+				$tagName = $tag->getName();
+				$tmpNodes = $this->userFolder->searchBySystemTag($tagName, $this->userSession->getUser()->getUID(), $limit ?? 0, $offset ?? 0);
+				if (count($nodes) === 0) {
```
marcelklehr, line 344: Isn't this always true? Since we have `$nodes = []` above?

lib/private/Files/Cache/QuerySearchHelper.php
```diff
@@ -147,27 +175,12 @@ public function searchInCaches(ISearchQuery $searchQuery, array $caches): array
...
-				->leftJoin('systemtagmap', 'systemtag', 'systemtag', $builder->expr()->andX(
-					$builder->expr()->eq('systemtag.id', 'systemtagmap.systemtagid'),
-					$builder->expr()->eq('systemtag.visibility', $builder->createNamedParameter(true))
-				));
+		$requestedFields = $this->searchBuilder->extractRequestedFields($searchQuery->getSearchOperation());
+		if (in_array('systemtag', $requestedFields)) {
+			$this->equipQueryForSystemTags($query, $this->requireUser($searchQuery));
+		}
+		if (in_array('tagname', $requestedFields) || in_array('favorite', $requestedFields)) {
+			$this->equipQueryForDavTags($query, $this->requireUser($searchQuery));
```
marcelklehr, line 183: nice!

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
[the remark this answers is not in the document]
blizzz: Is there a scenario where we can reach this point without being authed?

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
blizzz: only for the first tag

marcelklehr: AH, I get it now :)

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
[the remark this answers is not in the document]
marcelklehr: I don't know. Mh, it's only Github's psalm analysis that finds this, not our own, so probably good to go.

marcelklehr (approved):

juliushaertl:
To adapt
- [ ] group folders: PHP Fatal error:  Class OCA\GroupFolders\Helper\LazyFolder contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (OCP\Files\Folder::searchBySystemTag) in /var/www/html/apps-extra/groupfolders/lib/Helper/LazyFolder.php on line 32

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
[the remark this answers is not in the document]
marcelklehr: Ah, but applying $limit and $offset within this loop doesn't make much sense, I think, since the results are intersected in the end, that will cause weird results, no?

juliusknorr (changes requested):
Changes look good and seem do the trick

Blocking for missing docs as the public API is changed:

@blizzz Can you prepare a companion PR for the documentation update on the API changes and maybe also with the short doc section about the system tag retrieval that you mentioned? Then we can merge them in sync ;)

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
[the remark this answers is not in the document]
blizzz: You are right about this. In worst case no results are returned, while some results may be returned, if we didn't to the limitation early. 

To keep it simple we can keep the limit of only one tag was provided for the search, and otherwise slice it later on. This would fit with the shortpath on mutli-tag-search support we decided upon.

blizzz:
> Changes look good and seem do the trick
> 
> Blocking for missing docs as the public API is changed:
> 
> @blizzz Can you prepare a companion PR for the documentation update on the API changes and maybe also with the short doc section about the system tag retrieval that you mentioned? Then we can merge them in sync ;)

The API Changes at: https://github.com/nextcloud/documentation/pull/10678

The other thing you refer to, that would be the DAV endpoint extension from the first PR?

juliusknorr (approved):

juliushaertl:
Thanks, a right that was the other one. Yes, that would be nice to have as well :)

PR #38003, opened 2023-05-01
Improve translations api with detecting languages

nickvergessen:
## Summary

1. Fix a bug that when the first translation provider was unable to detect a language translations would not work
2. Return the from language (on success and error) so that clients can select the proper option and/or give hints to the user how they can get closer to a translation

## Checklist

### DIFF
--- a/core/Controller/TranslationApiController.php
+++ b/core/Controller/TranslationApiController.php
@@ -32,8 +32,8 @@ use OCP\AppFramework\Http\DataResponse;
 use OCP\IL10N;
 use OCP\IRequest;
 use OCP\PreConditionNotMetException;
+use OCP\Translation\CouldNotTranslateException;
 use OCP\Translation\ITranslationManager;
-use RuntimeException;
 
 class TranslationApiController extends \OCP\AppFramework\OCSController {
 	private ITranslationManager $translationManager;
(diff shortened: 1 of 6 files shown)


lib/private/Translation/TranslationManager.php
```diff
@@ -58,28 +59,36 @@ public function getLanguages(): array {
...
+		if ($fromLanguage === null) {
+			foreach ($this->getProviders() as $provider) {
+				if ($provider instanceof IDetectLanguageProvider) {
+					$fromLanguage = $provider->detectLanguage($text);
+				}
+
+				if ($fromLanguage !== null) {
+					break;
+				}
 			}
```
vitormattos, line 76: Did we have unit test to cover this?

nickvergessen: Unluckily the class was written without unit tests.

lib/private/Translation/TranslationManager.php
```diff
@@ -58,28 +59,36 @@ public function getLanguages(): array {
 		return $languages;
 	}
 
-	public function translate(string $text, ?string $fromLanguage, string $toLanguage): string {
+	public function translate(string $text, ?string &$fromLanguage, string $toLanguage): string {
```
marcelklehr, line 62: Nice!

core/Controller/TranslationApiController.php
```diff
@@ -68,15 +68,19 @@ public function languages(): DataResponse {
...
+
 			]);
 		} catch (PreConditionNotMetException) {
 			return new DataResponse(['message' => $this->l->t('No translation provider available')], Http::STATUS_PRECONDITION_FAILED);
 		} catch (InvalidArgumentException) {
 			return new DataResponse(['message' => $this->l->t('Could not detect language')], Http::STATUS_BAD_REQUEST);
-		} catch (RuntimeException) {
-			return new DataResponse(['message' => $this->l->t('Unable to translate')], Http::STATUS_BAD_REQUEST);
+		} catch (CouldNotTranslateException) {
+			return new DataResponse(['message' => $this->l->t('Unable to translate'), 'from' => $fromLanguage], Http::STATUS_BAD_REQUEST);
```
marcelklehr, line 83: will the variable be set here if it was null?

lib/private/Translation/TranslationManager.php
vitormattos: Isn't a good moment to create or create a fallback issue to implement tests on this class?

vitormattos (approved):

marcelklehr (approved):

PR #38004, opened 2023-05-01
Nextcloud 25 design bug

VladlenO:
guys, i respect you.  you are doing a great thing.  but my God, what kind of drug addicts did the design in the 25th version?  corny even there is no adaptation to phones.  Are we back in 2007?  what the hell is with the icons?  if the menu is narrowed even on a PC, then the icons are either barely noticeable, or they are not there!  why the background is always loading (all pages), which is why even 50 megabit internet is not enough.  there was no such practice before, or rather it was in the dashboard, which was good.  Don't let addicts design anymore!  return it as it was, please!
Screenshot_20230501_161536_com huawei browser
Screenshot_20230501_161552_com huawei browser
Screenshot_20230501_161603_com huawei browser


VladlenO:
do you really not see the problem here?

nickvergessen:
You send a pull request from stable25 to master, instead of creating an issue at: https://github.com/nextcloud/server/issues/new?assignees=&labels=bug%2C0.+Needs+triage&template=BUG_REPORT.yml&title=%5BBug%5D%3A+

jancborchardt:
Hi @VladlenO 

> guys, i respect you. you are doing a great thing. but my God, what kind of drug addicts did the design in the 25th version?

If you respect the community, then please be civil at all times and refrain from name calling. We have a code of conduct: https://nextcloud.com/contribute/code-of-conduct/

> corny even there is no adaptation to phones. Are we back in 2007? what the hell is with the icons? if the menu is narrowed even on a PC, then the icons are either barely noticeable, or they are not there! why the background is always loading (all pages), which is why even 50 megabit internet is not enough. there was no such practice before, or rather it was in the dashboard, which was good. Don't let addicts design anymore! return it as it was, please!

There seem to be a lot of different things in here. The icons not being visible in the dropdown menu looks like a bug, please open a separate issue for that, thank you! :)

The background should get cached and not load newly every time. This might be an issue with your webserver setup as well. If you can reproduce it, please open a _separate_ issue for that too.

Note that to save resources, you can also set the background to a plain color, then no image is loaded.

PR #38024, opened 2023-05-02
Fix 32bits ci

come-nc:
## Summary

Fix 32bit CI which was failing with:
```
There was 1 error:

1) OCA\Files_Sharing\Tests\UpdaterTest::testDeleteParentFolder
Exception: Could not download app files_trashbin

/__w/server/server/lib/private/Installer.php:387
/__w/server/server/lib/private/legacy/OC_App.php:275
/__w/server/server/apps/files_sharing/tests/UpdaterTest.php:79
```

## Checklist

### DIFF
--- a/tests/lib/Template/JSResourceLocatorTest.php
+++ b/tests/lib/Template/JSResourceLocatorTest.php
@@ -135,7 +135,6 @@ class JSResourceLocatorTest extends \Test\TestCase {
 		$this->assertEquals($expectedWebRoot, $webRoot);
 		$this->assertEquals($expectedFile, $file);
 
-		array_pop(\OC::$APPSROOTS);
 		//unlink($new_apps_path_symlink);
 		//$this->rrmdir($new_apps_path);
 	}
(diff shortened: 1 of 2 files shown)


tests/lib/Template/JSResourceLocatorTest.php
```diff
@@ -135,7 +135,6 @@ public function testFindWithAppPathSymlink() {
 		$this->assertEquals($expectedWebRoot, $webRoot);
 		$this->assertEquals($expectedFile, $file);
 
-		array_pop(\OC::$APPSROOTS);
 		//unlink($new_apps_path_symlink);
 		//$this->rrmdir($new_apps_path);
```
come-nc, line 139: @susnux Should this be uncommented?

susnux: Yes I think it is safe to remove that files / symlink here.

Additonally I would also remove line (`expectedRoot` variable)
https://github.com/nextcloud/server/blob/95d98b57ec2b35f6e10ef3cd100b6287a9031d38/tests/lib/Template/JSResourceLocatorTest.php#L167

as this is dead code and probably copied from the other testcase above.

susnux (approved):
:+1:

artonge (approved):

PR #38026, opened 2023-05-02
Use a Generator for users in searchInGroups, type it as iterable

come-nc:
## Summary

Since this is a new method in 27 it’s the occasion to make use of a generator instead of an array

## Checklist

### DIFF
--- a/lib/private/Group/Database.php
+++ b/lib/private/Group/Database.php
@@ -336,10 +336,15 @@ class Database extends ABackend implements
 	 * @return array<int,string> an array of user ids
 	 */
 	public function usersInGroup($gid, $search = '', $limit = -1, $offset = 0): array {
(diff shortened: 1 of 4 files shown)


lib/private/Group/Database.php
```diff
@@ -380,11 +385,9 @@ public function searchInGroup(string $gid, string $search = '', int $limit = -1,
 		$users = [];
 		$userManager = \OCP\Server::get(IUserManager::class);
 		while ($row = $result->fetch()) {
-			$users[$row['uid']] = new LazyUser($row['uid'], $userManager, $row['displayname'] ?? null);
+			yield $row['uid'] => new LazyUser($row['uid'], $userManager, $row['displayname'] ?? null);
 		}
 		$result->closeCursor();
```
come-nc, line 390: This will not get called if the Generator is not exhausted by the caller, so this approach may be problematic.
Anyone have a better idea on how to do that?

Issue #37993, opened 2023-04-29
Imaginary url-signature-key missing and webp and avif support


JanisPlayer:
https://docs.nextcloud.com/server/24/admin_manual/installation/server_tuning.html

Would it maybe be possible to add a key for the API in the nextcloud?
-url-signature-key secret
Currently I have solved this with iptables.
But it would be nice if it were easier to protect against unauthorized access.

I have another idea too.
Generating WepP images via Imaginary would probably be significantly more performant.
I know that there is still no official WebP support for the PHP ImageMagick.
But I think Imaginary is perfect for this, I can generate very small images with great quality on a second server without it greatly affecting the performance of my Nextcloud.
Additional Avif support would of course be even nicer.
I mean Imaginary is a very underused option and can save a lot of memory and JPG can be replaced especially for thumbnails.
https://github.com/nextcloud/server/issues/13552
And yes, I know this topic has been discussed many times, but I think it's different with imaginary.

What I also find a pity about Nextcloud is that there is no button with which I can display the original image without downloading it.
Or maybe a short-term stored version (cache) that can be set, for example with 1080p and 90% quality, is supposed to save memory at the expense of performance.

I don't have much memory and can therefore only create JPG images with 30% quality and a maximum of 1024px.
So if you have more than 17K images, that's a problem as far as storage is concerned.
And with WebP I might even be able to specify a target value and not have to worry so much about the quality.

That might be a few ideas to make Nextcloud even better or more usable, because the way it is today, it can't keep up with Google Photos.
PHP alone is simply unsuitable and requires an external service here and it definitely has the ability to create the images quickly enough if there is enough power and I think with Imaginary and a few adjustments Nextcloud can finally keep up with Google without too much storage space waste.

I also have to say that my cell phone gallery app urgently needs a new version, which smears when scrolling and also when too many pictures have to be compared when uploading.
But uploading has gotten better, but you can't show any warnings, otherwise it just spams and crashes.

szaimen:
Feel free to create a PR for discussion

JanisPlayer:
I have another question, where can I find the instructions or a template on how to develop Nextcloud apps?

In the ticket, I had an idea about HQ previews and displaying images in their original quality on demand. The HQ previews should be temporarily created with a parameter for the preview generator, and optionally stored in a cache folder. This could be helpful for users with limited storage but sufficient bandwidth or computing power.

[...]

PR #38032, opened 2023-05-03
Imaginary WebP support

JanisPlayer:
* Resolves: # https://github.com/nextcloud/server/issues/37993)

## Summary
Hello, is my first pull, it was suggested to me in a comment and I must admit I don't really know how to do something like this.
Anyway, this code adds Imaginary WebP support and allows to add an API key.

In theory you could also extend this with e.g. Imaginary_query in the config, where you could just add something like that in the config itself.

However, until now I only wanted to add these two options.

```
  'preview_format' => 'webp',
  occ config:app:set preview webp_quality --value="30"
```

If preview_format or webp_quality doesn't exist the default will be used again.

So far WebP works exclusively for Imaginary, if it should work for the other generations I might need some help to find my way around the code.

As a small warning, I am not a skilled application developer only a hobby developer, because it has bothered that WebP does not work.

But I wanted to contribute something to the project and make it work faster.

Hope I did this right.

## TODO

- [x] for image/png application/illustrator you could use lossless or close to lossless WebP, because i don't know now how to do that lossless with imaginary. image/heic you could actually add in WebP support.

## Checklist

- Code is properly formatted
- [x] Tested
On web server, android

### DIFF
--- a/lib/private/Preview/Imaginary.php
+++ b/lib/private/Preview/Imaginary.php
@@ -106,6 +106,15 @@ class Imaginary extends ProviderV2 {
 				$mimeType = 'jpeg';
 		}
 
+		$preview_format = $this->config->getSystemValueString('preview_format', 'jpeg');
+
+		switch ($preview_format) { // Change the format to the correct one
+			case 'webp':
+				$mimeType = 'webp';
+				break;
+			default:
+		}
+
 		$operations = [];
 
 		if ($convert) {
@@ -121,7 +130,16 @@ class Imaginary extends ProviderV2 {
 			];
 		}
 
-		$quality = $this->config->getAppValue('preview', 'jpeg_quality', '80');
+		switch ($mimeType) {
+			case 'jpeg':
+				$quality = $this->config->getAppValue('preview', 'jpeg_quality', '80');
+				break;
+			case 'webp':
+				$quality = $this->config->getAppValue('preview', 'webp_quality', '80');
+				break;
+			default:
+				$quality = $this->config->getAppValue('preview', 'jpeg_quality', '80');
+		}
 
 		$operations[] = [
 			'operation' => ($crop ? 'smartcrop' : 'fit'),
--- a/lib/private/Preview/Generator.php
+++ b/lib/private/Preview/Generator.php
@@ -651,6 +651,8 @@ class Generator {
 				return 'png';
 			case 'image/jpeg':
 				return 'jpg';
+			case 'image/webp':
+				return 'webp';
 			case 'image/gif':
 				return 'gif';
 			default:


lib/private/Preview/Imaginary.php
```diff
@@ -136,9 +155,10 @@ public function getCroppedThumbnail(File $file, int $maxX, int $maxY, bool $crop
 		];
 
 		try {
+			$imaginary_key = $this->config->getSystemValueString('imaginary_key', 'invalid');
 			$response = $httpClient->post(
 				$imaginaryUrl . '/pipeline', [
-					'query' => ['operations' => json_encode($operations)],
+					'query' => ['operations' => json_encode($operations),'key' => $imaginary_key],
```
solracsf, line 161:
```suggestion
					'query' => ['operations' => json_encode($operations), 'key' => $imaginary_key],
```

lib/private/Preview/Imaginary.php
```diff
@@ -136,9 +155,10 @@ public function getCroppedThumbnail(File $file, int $maxX, int $maxY, bool $crop
 		];
 
 		try {
+			$imaginary_key = $this->config->getSystemValueString('imaginary_key', 'invalid');
```
artonge, line 158: I would default to an empty string, or is there any reason to use `invalid`?

artonge (approved):

artonge:
This looks good IMO.
Some documentation will be needed, can you take care of that @JanisPlayer ?

lib/private/Preview/Imaginary.php
JanisPlayer: Yes, that's right, that doesn't make sense what I did there.
I wanted to start with an if query whether to send key at all, for that reason "invalid" was the default return.
I actually had more changes in mind, but didn't want to blow up all the code with lots of queries now, so I submitted the smallest variant.
When I figure out how to change it, I'll do it, ty. :D

JanisPlayer (commented):
The default return of 'invalid' is unnecessary.

lib/private/Preview/Imaginary.php
```diff
@@ -136,9 +155,10 @@ public function getCroppedThumbnail(File $file, int $maxX, int $maxY, bool $crop
 		];
 
 		try {
+			$imaginary_key = $this->config->getSystemValueString('imaginary_key', '');
 			$response = $httpClient->post(
 				$imaginaryUrl . '/pipeline', [
-					'query' => ['operations' => json_encode($operations)],
+					'query' => ['operations' => json_encode($operations), 'key' => $imaginary_key],
```
szaimen, line 161: will this break on instances without key if an empty key was provided?

lib/private/Preview/Imaginary.php
```diff
@@ -103,7 +103,17 @@ public function getCroppedThumbnail(File $file, int $maxX, int $maxY, bool $crop
 				$mimeType = 'png';
 				break;
 			default:
-				$mimeType = 'jpeg';
+				$preview_format = $this->config->getSystemValueString('preview_format', 'jpeg');
```
szaimen, line 106: should also be added to the sample.config.php

lib/private/Preview/Imaginary.php
```diff
@@ -103,7 +103,17 @@ public function getCroppedThumbnail(File $file, int $maxX, int $maxY, bool $crop
...
+				switch ($preview_format) {
+					case 'jpeg':
+						$mimeType = 'jpeg';
+					    break;
+					case 'webp':
+						$mimeType = 'webp';
```
szaimen, line 112: what I wonder about is if webp previews would need to be explicitely enabled in the client apps. cc @tobiasKaminsky on this

lib/private/Preview/Imaginary.php
JanisPlayer: Seems Imaginary just ignores when an empty &key= is sent.
Tested it via GET and POST.

config/config.sample.php
```diff
@@ -1217,6 +1217,23 @@
...
+ * If you want set the preview format for imaginary.
+ * You can change between jpeg and webp, the default is jpeg.
+ */
+'preview_format' => 'webp',
+
+/**
+ * WebP quality setting for imaginary.
+ * Defaults to '80' quality
+ */
+'webp_quality' => '80',
```
szaimen, line 1235: this is the wrong place for these as they are include din the preview app and will not be read from config.php

lib/private/Preview/Imaginary.php
szaimen: okay

config/config.sample.php
JanisPlayer: Ah now I understand the difference between config->getSystem* and config->getApp*. :D
Should I use getAppValueString instead of `$preview_format = $this->config->getSystemValueString('preview_format', 'jpeg');`?
[Code](https://github.com/JanisPlayer/server/blob/930516e20556addf1c44a588ed7eac6272a5a0c1/lib/private/Preview/Imaginary.php#LL106C1-L106C1) (I don't know how to mark the code.)

szaimen: I am not sure if it should be better stored in the preview app. cc @skjnldsv on this

lib/private/Preview/Imaginary.php
tobiasKaminsky: Clients ask via thumbnail api for any image type, as long as mimetype is "image/*" or "video/*".

szaimen: so the filetype is set by the server? sounds nice!

config/config.sample.php
JanisPlayer: I just realized by accident while testing where the difference is.
App Config can only be changed with this command, what is in the config is completely irrelevant.
So I used to have this entry inside the config completely for nothing and it probably never had any effect, fail. :D
Luckily I also used this command.
occ config:app:set preview webp_quality --value="30"
occ config:app:get preview webp_quality
Means for me, my entry in the documentation is wrong.
But I think you will change your mind anyway, which config is used for which setting?
But I will correct the documentation already.

lib/private/Preview/Imaginary.php
```diff
@@ -84,26 +84,54 @@ public function getCroppedThumbnail(File $file, int $maxX, int $maxY, bool $crop
...
+				switch ($preview_format) {
+					case 'jpeg':
+						$mimeType = 'jpeg';
+						break;
+					case 'webp':
+						$mimeType = 'webp';
+						break;
+					default:
+						$mimeType = 'jpeg';
+				}
```
artonge, line 134: Am I right that we are adding a new switch to every case of the existing switch?
Couldn't we add a single switch after the original switch?

lib/private/Preview/Imaginary.php
tobiasKaminsky: In general yes. I do not know nothing about this imaginary.
But in e.g. NC26 I can upload any image format and ask for a thumbnail.
Either server can handle it, then I receive a thumbnail or it cannot, then clients show placeholder.

API is: /index.php/apps/files/api/v1/thumbnail/128/128/test/image.tiff

This also works for office files, etc.

szaimen: What I was wondering about was if it is a problem if at this endpoint a webp is returner?

lib/private/Preview/Imaginary.php
JanisPlayer: It would also be possible to ask for the switch whether $preview_format = 'webp' is set, and if it is set, simply set $mimeType = 'webp'. And above, only the formats PNG or JPEG should be set if WebP is not set. (This is not necessary since the variable is corrected further down anyway if WebP is set.)
Since WebP is selected in any case if JPEG is not set, this of course works as well.
However, I did not do it this way because Imaginary (Nextcloud version) may add more formats to its support, but all modern formats actually have an alpha channel.
So it can actually be corrected further down to save code.
So both would have the same effect, it's a matter of preference.
There may be a better way, but that's what comes to mind right now.

artonge: This seems indeed more readable to me :)

JanisPlayer: Have tested it, goes and changed.
It definitely saves a lot of space that way.
https://github.com/nextcloud/server/pull/38032/commits/f926d549bfc6f584fe27f909702d62788b919e4d

lib/private/Preview/Imaginary.php
tobiasKaminsky: Not sure if it is somewhere stated, or just done all the time, but thumbnails are all pngs.
This should be kept this way.

So for thumbnails, above endpoint, all must convert to png.

JanisPlayer:
@szaimen 
Yes can try.
The problem I can not create another fork then I have to make a new branch?
And if I take out the function again it must be reset to the gefork version or to the current version?
I think the [fork](https://github.com/JanisPlayer/server/blob/4f24fb9ab042b88b9319e64f14a2f5128a0c37eb/lib/private/Preview/Imaginary.php) version makes sense, because that's what I selected it for.
But I'm not sure how GitHub manages that.
But I think that's what they meant by the issue, because I see they made a change here for the timeout.
And the connect_timeout is nice, to probably prevent Nextcloud from being slowed down when the server has connection problems, the problem had also noticed once where my internet provider at home had problems.

szaimen:
I think the easiest way is probably to start from here: https://github.com/nextcloud/server/edit/master/lib/private/Preview/Imaginary.php and be guided throught the process by github automatically.

nfebe (approved):

artonge:
Still looks good. Rebased to have the performance test green.
Ideally, could you squash your commit into one ?

JanisPlayer:
Apologies, I think I just did something completely wrong.
I have reverted it.
I used these commands:
`git reset --soft HEAD~17`
This command to reset the last 17 commits.
`git commit -am "Add imaginary WebP support"`
This command to consolidate all changes into one commit.
`git push --force origin master`
This command to apply the changes, but it seems it didn't only affect my last 17 commits and caused a huge mess.
I'm trying to figure out shortly what I did wrong.
However, I have reverted it to `b83bd552b9d` to clear the confusion, but unfortunately, I realized it too late, sorry.

miaulalala:
@JanisPlayer if you want some help rebasing your PR, join us here https://cloud.nextcloud.com/call/xs25tz5y and we can chat / have a call

JanisPlayer:
I tried it with "-fixup sha," as someone recommended to me. While I saw the changes it made in the console and then in GitKraken, I thought it would apply the "Merge branch 'revert'" and that it would be fine in this case. However, it seems that this way, all changes are overwritten for the PR back to the current point in time.

I believe this might not work after a "Merge branch" as I intend to. I'm now frustrated because I pressed that button, and I can't correct my mistake. I'm even considering just deleting everything and starting over because I know that will definitely work.

skjnldsv:
Rebased for you here is how I did it
1. `git reset --hard 92477bc018785eea473cf0e471437add09e3bc3e`, to reset to the last commit before the last three merge you had
2. `git rebase -i --autosquash HEAD~14` and then edit all but the first to add `f` instead of `pick` and squash all commits into one
3. `git fetch origin master:master`
4. `git rebase master`
5. Then force push

invario:
Wondering if anyone has experienced what I detailed in #43878 ?

PR #38047, opened 2023-05-03
chore: update HowToApplyALicense.md

ShGKme:
## Summary

I'd like to improve `HowToApplyALicense.md`

1. Update **applying to an existing file** with a diff
2. Add different files for different file types (add html-like for `.vue`)
3. Add different samples for different types of source:

The current example doesn't work on frontend. It is blocked by [@nextcloud/eslint-config](https://github.com/nextcloud/eslint-config/). Which inherits `eslint-plugin-jsdoc` and checks the license by [spdx-expression-parse](https://github.com/jslicense/spdx-expression-parse.js). So all JS files have `AGPL-3.0-or-later` instead of recommended by this doc `GNU AGPL version 3 or any later version`. 

## Checklist

### DIFF
--- a/contribute/HowToApplyALicense.md
+++ b/contribute/HowToApplyALicense.md
@@ -9,16 +9,60 @@ Additionally Nextcloud doesn't require a CLA (Contributor License
 Agreement). The copyright belongs to all the individual
 contributors.
 
-If you modify an existing file, please keep the existing license header as
-it is and just add your copyright notice:
+## Apply a license to a new file
 
-````
-@copyright Copyright (c) <year>, <your name> (<your email address>)
+If you create a new file please use a license header
+
+#### Frontend source (`.js`, `.ts`, `.css` and etc)
+
+```js
+/**
+ * @copyright Copyright (c) <year>, <your name> (<your email address>)
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
 ````
 
-If you create a new file please use this license header:
+or `.vue` files
 
-````
+```html
+<!--
+  - @copyright Copyright (c) <year>, <your name> (<your email address>)
+  -
+  - @license AGPL-3.0-or-later
+  -
+  - This program is free software: you can redistribute it and/or modify
+  - it under the terms of the GNU Affero General Public License as
+  - published by the Free Software Foundation, either version 3 of the
+  - License, or (at your option) any later version.
+  -
+  - This program is distributed in the hope that it will be useful,
+  - but WITHOUT ANY WARRANTY; without even the implied warranty of
+  - MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+  - GNU Affero General Public License for more details.
+  -
+  - You should have received a copy of the GNU Affero General Public License
+  - along with this program. If not, see <http://www.gnu.org/licenses/>.
+-->
+```
+
+#### Backend source (`.php`)
+
+```php
 /**
  * @copyright Copyright (c) <year>, <your name> (<your email address>)
  *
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

contribute/HowToApplyALicense.md
```diff
@@ -38,8 +82,39 @@ If you create a new file please use this license header:
...
+## Apply a licence to an existing file
+
+If you modify an existing file, please keep the existing license header as
+it is and just add your copyright notice, for example:
+
+````diff
+/**
+ * @copyright Copyright (c) 2022, Alice (alice@nextcloud.local)
+ * @copyright Copyright (c) 2023, Bob (bob@nextcloud.local)
++* @copyright Copyright (c) <year>, <your name> (<your email address>)
```
skjnldsv, line 96: We lost the `@author` section now?
We used to not update the copyright, but add ourselves to the author section

```php
/**
 * @copyright Copyright (c) 2018 Robin Appelman <robin@icewind.nl>
 *
 * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
 * @author Christoph Wurst <christoph@winzerhof-wurst.at>
 * @author Robin Appelman <robin@icewind.nl>
 *
 * @license AGPL-3.0-or-later
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU Affero General Public License as
 * published by the Free Software Foundation, either version 3 of the
 * License, or (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU Affero General Public License for more details.
 *
 * You should have received a copy of the GNU Affero General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 *
 */```

skjnldsv (dismissed):

contribute/HowToApplyALicense.md
skjnldsv: I have no idea what is the proper way btw, maybe yours is what we should do from now on :see_no_evil:
Just saying what I've seen everyone do in the past :grin:

ShGKme: > We lost the @author section now?

`@author` jsdoc/phpdoc has no legal meaning (C) @nickvergessen

> We used to not update the copyright, but add ourselves to the author section

It is always different. For example, in these files, there are several copyrights. And dozens other files in server (and other apps too).

https://github.com/nextcloud/server/blob/6f9afa77b34e2242583e71c8d31a84e499cbf092/apps/dav/lib/CalDAV/Reminder/Backend.php#L5-L7

https://github.com/nextcloud/server/blob/54b6d0708bcf389857fdc7e0afe8f0edcf8ea3f3/lib/private/Metadata/Provider/ExifProvider.php#L4-L7

https://github.com/nextcloud/server/blob/ede9ac2a752f5268163688e9761331a168a5a90e/tests/acceptance/features/bootstrap/AppNavigationContext.php#L3-L8

Also, in the **HowToApplyALicense** it is said to add a copyright

https://github.com/nextcloud/server/blob/e5ba7eb3d574198b096fa8ed098b7838b208a32b/contribute/HowToApplyALicense.md?plain=1#L12-L17

nickvergessen: > We used to not update the copyright, but add ourselves to the author section

This origins from oC times, where the company held the copyright and the commiter was just the author. So we can do copyrights only now, from my pov

skjnldsv: That seems like something far out of my league and that should be raised from a legal perspective then :)

nickvergessen: cc @AndyScherzinger  for confirmation

AndyScherzinger: Confirming, yes 👍 

It is btw also how we do it for the Android apps' code - where we actually maintain both authorship and copyright.
At least for java/kotlin it makes sense to do both since `@author` is a javadoc recognized annotation.

AndyScherzinger (approved):

contribute/HowToApplyALicense.md
ShGKme: > It is btw also how we do it for the Android apps' code - where we actually maintain both authorship and copyright.
At least for java/kotlin it makes sense to do both since @author is a javadoc recognized annotation.

Should it be mentioned in the doc?

skjnldsv: And from a legal perspective?
Because this is not a light change I think, no?

AndyScherzinger: I can check with @karlitschek - however any contributor would own the copyright of their contributions since we do not have a CLA, so the copyright is not transfered. For employees however one could argue that the company owns them while that depends on the specifics in our contract.

What do you think Frank? I tend to say the for `server` or published apps the copyright can always linked to the specific person contributing.

ChristophWurst: As far as I know it depends on your country. Here (and apparently in Germany too) the copyright stays with the author. Employers have a right to use the work, but the copyright doesn't transfer.

skjnldsv: Ping? Close or merge @AndyScherzinger

PR #38048, opened 2023-05-03
feat(dav): store scopes for properties and filter locally scoped properties for federated address book sync

### LINKED ISSUE #38021: Store local properties in system addressbook too
### How to use GitHub

The system address book only contains contacts with published or federated properties. Other properties are not written.

1) Write local properties too. Write contacts with only local properties. Only discard properties scoped as private.
2) Filter contacts when syncing to federated instances to remove all local properties.

**Additional context**

https://github.com/nextcloud/server/issues/19575#issuecomment-1527375723

miaulalala:
* Resolves: https://github.com/nextcloud/server/issues/38021

## Summary

Adds a users property scopes to the system address book. Private properties are not written at all.
When syncing with a federated share, the locally scoped properties should not be shared with the remote host.

## Todo
- [x] Store scopes for properties
- [x] Write local properties to system address book too
- [x] Filter locally scoped properties for federated address book sync
- [x] Repair step to add scopes to existing card data
- [x] CardDav sync token adjustment for cards inaccessible to federated instances

## How to test
Needs a federated instance setup
1. Change a property on a local user to be scope local only
2. trigger address book sync from remote host
3. check the db for the changed contact by looking into the second system address book.
Before
```
BEGIN:VCARD
VERSION:3.0
PRODID:-//Sabre//Sabre VObject 4.4.2//EN
UID:admin
FN;X-NC-SCOPE=v2-federated:admin
N;X-NC-SCOPE=v2-federated:admin;;;;
ADR;TYPE=OTHER;X-NC-SCOPE=v2-local:Testing test test test;;;;;;
EMAIL;TYPE=OTHER;X-NC-SCOPE=v2-federated:miau_lalala@gmx.net
TEL;TYPE=OTHER;X-NC-SCOPE=v2-local:+435454454544
CLOUD:admin@http://localhost
END:VCARD
```

After
```
BEGIN:VCARD
VERSION:3.0
PRODID:-//Sabre//Sabre VObject 4.4.2//EN
UID:admin
FN;X-NC-SCOPE=v2-federated:admin
N;X-NC-SCOPE=v2-federated:admin;;;;
EMAIL;TYPE=OTHER;X-NC-SCOPE=v2-federated:miau_lalala@gmx.net
CLOUD:admin@http://localhost
END:VCARD
```

## Checklist

### DIFF
--- a/apps/dav/lib/CardDAV/Converter.php
+++ b/apps/dav/lib/CardDAV/Converter.php
@@ -29,15 +29,12 @@ namespace OCA\DAV\CardDAV;
 
 use Exception;
 use OCP\Accounts\IAccountManager;
-use OCP\Accounts\PropertyDoesNotExistException;
 use OCP\IImage;
 use OCP\IUser;
 use Sabre\VObject\Component\VCard;
 use Sabre\VObject\Property\Text;
-use function array_merge;
 
 class Converter {
-
 	/** @var IAccountManager */
 	private $accountManager;
 
@@ -46,13 +43,7 @@ class Converter {
 	}
 
 	public function createCardFromUser(IUser $user): ?VCard {
-		$account = $this->accountManager->getAccount($user);
-		$userProperties = $account->getProperties();
-		try {
-			$additionalEmailsCollection = $account->getPropertyCollection(IAccountManager::COLLECTION_EMAIL);
-			$userProperties = array_merge($userProperties, $additionalEmailsCollection->getProperties());
-		} catch (PropertyDoesNotExistException $e) {
-		}
+		$userProperties = $this->accountManager->getAccount($user)->getAllProperties();
 
 		$uid = $user->getUID();
 		$cloudId = $user->getCloudId();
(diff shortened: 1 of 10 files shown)


ChristophWurst (commented):
:+1: code looks very good

ChristophWurst:
Only one remark. Existing users with local properties don't get a trigger to update their vcards, right? Only when a user property changes we regenerate the system card

ChristophWurst (approved):
Tested with my database user. Set all props to local and now I suddenly have an entry in the system addres book. All props are scoped to v2-local :+1: 

Didn't test federation

miaulalala:
> Only one remark. Existing users with local properties don't get a trigger to update their vcards, right? Only when a user property changes we regenerate the system card

Yes that is indeed an issue, everything that already exists and doesn't update is gonna stay the same. Maybe a repair step could help?

ChristophWurst:
`\OCA\DAV\CardDAV\SyncService::syncInstance` can do the full instance update. Repair step would run for every future migration. This is a one-time thing. A migration without db schema feels wrong too. So yeah, repair step, but you have to add a flag to oc_preferences whether users were updated or not.

apps/dav/lib/Migration/Version1027Date20230504122946.php
```diff
@@ -0,0 +1,56 @@
...
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 */
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+		/** @var SyncService $syncService */
+		try {
+			$syncService = \OC::$server->get(SyncService::class);
+		} catch (NotFoundExceptionInterface | ContainerExceptionInterface $e) {
+			/** @var LoggerInterface $logger */
+			$logger = \OC::$server->get(LoggerInterface::class);
```
ChristophWurst, line 50: inject the service and the logger please

ChristophWurst (approved):

ChristophWurst:
Is this generally ready to review?

apps/dav/lib/CardDAV/Converter.php
```diff
@@ -44,7 +43,7 @@ public function __construct(IAccountManager $accountManager) {
 	}
 
 	public function createCardFromUser(IUser $user): ?VCard {
-		$userProperties = $this->accountManager->getAccount($user)->getProperties();
+		$userProperties = $this->accountManager->getAccount($user)->getAllProperties();
```
ChristophWurst, line 46: I guess this obsoletes https://github.com/nextcloud/server/pull/38073

JohannesGGE (approved):

tcitworld:
At this point there's hopefully a way to factorize things in `SystemAddressbook.php`.

apps/dav/lib/CardDAV/SystemAddressbook.php
```diff
@@ -50,4 +64,193 @@ public function getChildren() {
...
+		}
+
+		$changed = $this->carddavBackend->getChangesForAddressBook(
+			$this->addressBookInfo['id'],
+			$syncToken,
+			$syncLevel,
+			$limit
+		);
+
+		// If it's ugly but it works, is it still ugly?
```
ChristophWurst, line 232: yes

apps/dav/lib/CardDAV/Converter.php
miaulalala: https://github.com/nextcloud/server/pull/38073 was merged, conflict needs to be resolved

ChristophWurst:
To be sure the filtered sync changes work as expected I'd like to throw DAVxc5 and Thunderbird against the federated address book using system:sharedsecret. That should work, right?

miaulalala:
> To be sure the filtered sync changes work as expected I'd like to throw DAVxc5 and Thunderbird against the federated address book using system:sharedsecret. That should work, right?

If you know the shared secret that should work, yes.

ChristophWurst:
I'm struggling to connect to the collection with the `system` user. I have a debugger attached at `\OCA\Federation\DAV\FedAuth::validateUserPass` but it doesn't go there apparently.

miaulalala:
drone fails because of merge conflict

ChristophWurst:
> Listing collection `/remote.php/dav/addressbooks/system/system/system/': collection is empty.

Unintentionally tested the sharing enumeration restriction. Works :+1: 

Without it, I finally get my system contacts:

```
dav:/remote.php/dav/addressbooks/system/system/system/> ls
Listing collection `/remote.php/dav/addressbooks/system/system/system/': succeeded.
        Database:admin.vcf                   515  May  4 20:01
        Database:asdfasdfdasfsdfdsfasd.vcf        248  May  4 19:46
        Database:asdfasfasfasdf.vcf          259  May  4 20:26

```

My user's location is only shown in the downloaded .vcf if set to federated or published. If set to local it's not readable for the remote instance :+1:

ChristophWurst:
davx5fed

davx5 can connect but doesn't fetch *any* contacts. Same issue on master. `\OCA\DAV\CardDAV\SystemAddressbook::getChildren` is not called.

miaulalala:
> davx5fed
> 
> davx5 can connect but doesn't fetch _any_ contacts. Same issue on master. `\OCA\DAV\CardDAV\SystemAddressbook::getChildren` is not called.

I think the array_map function in https://github.com/nextcloud/server/blob/0f0c67c88f4773923e1995f639481e4848c594c9/apps/dav/lib/CardDAV/UserAddressBooks.php#L79 might be the cause.

Can you add a debugger and check if the systemaddressbook is actually instantiated?

ChristophWurst:
> Can you add a debugger and check if the systemaddressbook is actually instantiated?

Bildschirmfoto vom 2023-05-08 13-37-00

it is

ChristophWurst (approved):
I let davx5 play federated instance and sync the addressbook. Works on master, then switched to this branch, rebuilt the SAB and synced again. No local props show up on the phone :+1:

ChristophWurst:
Fixed the static analysis issue with nullable changes. Even tried to fix upstream but the usual suspect was faster: https://github.com/sabre-io/dav/commit/9b10ec69d8fa56a1920c9dc4ad6ef131aa8cf0eb

ChristophWurst:
contacts menu integration test fails reliably -> merge

PR #38051, opened 2023-05-03
fix(weather_status): Override the default `NcActions` aria-label

### LINKED ISSUE #37018: [BITV] 9.2.5.3/2.1 - The visible label "Set location for weather" does not appear in the accessible name of the button in the same form as it is overridden by the use of aria-label="Actions". In this case voice activation would fail as visible text is not in accessible name. (1)
- [ ] Fix
- [ ] Backport

51e36c48b52ba8aa27249843b98fe342

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-c6347c127a-v2-n1

</details>

susnux:
* Resolves: #37018

## Summary
Currently the default `aria-label` of `NcActions` hides the menu text, this is fixed by replacing the `aria-label` with the menu text.

Note: I first thought about removing the aria label and let the browser fallback to the text, but the property requires a string so `:aria-label="null"` works but a future change might replace all falsy values with the default *Actions* text.

## Checklist


JuliaKirschenheuter (approved):

artonge (approved):
Nice !
Can you add some before/after screenshots?
I could not reproduce the overlapping of the aria-label with the text.

susnux:
> Nice ! Can you add some before/after screenshots? I could not reproduce the overlapping of the aria-label with the text.

There is no visual overlapping, but the `aria-label` is overriding the inner text so the visual label does not appear in the names:

before | after
---|---
before the label only says ACTIONS|after the label is the same as the visual one

PR #38065, opened 2023-05-04
fix(mimetype): Fix returning value when finding existing mimetype in MimeType Loader

tcitworld:
Fixes https://github.com/nextcloud/server/pull/35744/files#r1184644610

## Checklist

### DIFF
--- a/lib/private/Files/Type/Loader.php
+++ b/lib/private/Files/Type/Loader.php
@@ -116,7 +116,7 @@ class Loader implements IMimeTypeLoader {
 	 * @return int inserted ID
 	 */
 	protected function store($mimetype) {
-		$row = $this->atomic(function () use ($mimetype) {
+		$mimetypeId = $this->atomic(function () use ($mimetype) {
 			try {
 				$insert = $this->dbConnection->getQueryBuilder();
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

lib/private/Files/Type/Loader.php
```diff
@@ -116,42 +116,36 @@ public function reset() {
...
-				if ($row) {
-					return [
-						'mimetype' => $mimetype,
-						'id' => $row['id'],
-					];
+					->where($qb->expr()->eq('mimetype', $qb->createNamedParameter($mimetype)));
+				$result = $qb->executeQuery();
+				$id = (int) $result->fetchOne();
+				$result->closeCursor();
+				if ($id) {
```
come-nc, line 139: Is it known for sure that `0` is not a valid id?

come-nc (approved):

lib/private/Files/Type/Loader.php
tcitworld: Moved the cast below and added a strict check.

come-nc (approved):

PR #38067, opened 2023-05-04
fix: sort tags by name

blizzz:
Alternative to #38066

@tobiasKaminsky please test, i did only superficially.

### DIFF
--- a/apps/dav/lib/SystemTag/SystemTagPlugin.php
+++ b/apps/dav/lib/SystemTag/SystemTagPlugin.php
@@ -25,6 +25,7 @@
  */
 namespace OCA\DAV\SystemTag;
 
+use OC\SystemTag\SystemTag;
 use OCA\DAV\Connector\Sabre\Directory;
 use OCA\DAV\Connector\Sabre\Node;
 use OCP\IGroupManager;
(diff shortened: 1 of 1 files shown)


icewind1991 (approved):

skjnldsv (approved):

AndyScherzinger (approved):

tobiasKaminsky:
Tested and works, api wise.

apps/dav/lib/SystemTag/SystemTagPlugin.php
```diff
@@ -306,8 +308,11 @@ private function propfindForFile(PropFind $propFind, Node $node): void {
 			if ($user === null) {
 				return;
 			}
-	
+
 			$tags = $this->getTagsForFile($node->getId(), $user);
+			usort($tags, function (SystemTag $tagA, SystemTag $tagB): int {
```
come-nc, line 313:
```suggestion
			usort($tags, function (ISystemTag $tagA, ISystemTag $tagB): int {
```
The interface has the `getName` method, why type this to the internal implementation?

apps/dav/lib/SystemTag/SystemTagPlugin.php
```diff
@@ -25,6 +25,7 @@
  */
 namespace OCA\DAV\SystemTag;
 
+use OC\SystemTag\SystemTag;
```
come-nc, line 28:
```suggestion
```
Apps should not use `OC`. I do not see the need here.

come-nc (dismissed):

tobiasKaminsky:
When adding a tag "0123", then it first is shown correctly.
image

But on page refresh it is like this:
image

tobiasKaminsky (dismissed):
see comment about "0123"

blizzz:
@tobiasKaminsky I can reproduce this behaviour with current master, so these changes are not causing it. Please open a separate issue.

PR #38077, opened 2023-05-04
Remove version check for ancient libxml version

provokateurin:
## Summary

New error with psalm 5.

With `LIBXML_LOADED_VERSION` I get `Cannot perform a numeric operation with a non-numeric type '21004-GITv2.10.4'`.
Looking at https://www.php.net/manual/en/libxml.constants.php it's not even documented...

EDIT: This version is so old that we can just remove this check.

## Checklist

### DIFF
--- a/lib/private/legacy/OC_Util.php
+++ b/lib/private/legacy/OC_Util.php
@@ -692,20 +692,6 @@ class OC_Util {
 			];
 		}
 
-		if (function_exists('xml_parser_create') &&
-			LIBXML_LOADED_VERSION < 20700) {
-			$version = LIBXML_LOADED_VERSION;
-			$major = floor($version / 10000);
-			$version -= ($major * 10000);
-			$minor = floor($version / 100);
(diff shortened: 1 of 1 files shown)


provokateurin:
Looking at the code I was able to make it even easier since the version was used to parse and format it, but there is already a variable for the formatted version.

nickvergessen:
https://github.com/php/php-src/blob/master/ext/libxml/libxml.c#LL45C13-L45C34 hmm strange it's missing from the docs

kesselb:
Good finding :+1: 

I think it should also work to remove the check for libxml 2.7.0. 
It was released in 2008 (https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.7.4) so everyone should have it by now ;)

kesselb (approved):

provokateurin:
> It was released in 2008

Haha ok then it's even easier :D

blizzz (approved):

PR #38093, opened 2023-05-05
Allow inserting HTML fragments in email templates

tcitworld:
**This is a proposal, this PR is here for discussions.**

Email templates currently only allow to add text bodies, lists and buttons. It's not possible to add formatting or other elements. Also, text bodies are always centered, which can be hard to read when there a lot of text. This provides a way to inject arbitrary HTML in emails, which can be useful in some cases.

One of these cases I have is https://github.com/nextcloud/announcementcenter/issues/644, which would be achieved otherwise by creating an `IMessage` from scratch, but that prevents consistency and theming.

Right now the HTML fragment is really just injected in the HTML body, but we could validate (very basically, at a XML level) and sanitize it (using [SF's HTML Sanitizer](https://symfony.com/doc/current/html_sanitizer.html)).

Another option is to allow for Markdown fragments directly, and do the rendering here (brings another specific dependency though).

## TODO

- [ ] Add tests to `tests/lib/Mail/EMailTemplateTest.php`
- [ ] Add docs for the new feature

## Checklist

### DIFF
--- a/lib/private/Mail/EMailTemplate.php
+++ b/lib/private/Mail/EMailTemplate.php
@@ -598,6 +598,25 @@ EOF;
 	}
 
 	/**
+	 * Adds a HTML fragment to the body of the email. Use only for basic HTML fragments and make sure the given HTML is valid, as this can affect the whole email HTML body
+	 *
+	 * @param string $html The HTML fragment to add to the body of the email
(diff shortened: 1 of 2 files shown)


come-nc:
I don’t have anything against it, but it does feel like there should be some basic safety check.

PR #38123, opened 2023-05-08
NMC-127 - File action button behaviour for different viewport

TSI-kavitasonawane:
This PR contains the following changes:
File action button behaviour changes:

- We now show more single actions depending on the viewport size in the table header beginning from the first option in the dropdown.
- We add a new option called "Cancel"  which will remove any selection (just remove selection, not the files of course).
- We right align the Actions button.
- We will show every option in the action button dropdown. It doesnt matter which option is also shown in the table header.
- We add a new label called "All".
- on S: we hide the label "All" according to the screen design

Screenshot after changes:

filesaction
actionchange


skjnldsv:
Not sure if worth having considering 28 and Files to vue
cc @AndyScherzinger

skjnldsv:
FIxed in 28

PR #38133, opened 2023-05-09
feat: Add setup check warning if instance is in debug mode

juliusknorr:
I've seen a couple of instances running in debug mode in production recently and thought it might be a good idea to add a setup check for this since we only recommend it for development purposes as it may widen security measurements and also could introduce unexpected behaviour for regular users where the js assets are not automatically invalidated in the browser cache.

## Checklist

### DIFF
--- a/core/js/setupchecks.js
+++ b/core/js/setupchecks.js
@@ -521,6 +521,12 @@
 							type: OC.SetupChecks.MESSAGE_TYPE_WARNING
 						})
 					}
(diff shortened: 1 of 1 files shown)


solracsf:
> as it may widen security measurements and also could introduce unexpected behaviour for regular users.

Maybe add it to the warning too?

szaimen (approved):
Good idea!

skjnldsv (approved):

PR #38150, opened 2023-05-09
make Node::getParent lazy

icewind1991:
Since we know the parent exists and we have some basic metadata already, a lazy folder can be enough is some cases.

Main motivation is allow getting the parent it without extra db queries using `$node->getParent()->getId()`.

### DIFF
--- a/lib/private/Files/Node/LazyFolder.php
+++ b/lib/private/Files/Node/LazyFolder.php
@@ -29,7 +29,9 @@ namespace OC\Files\Node;
 use OC\Files\Utils\PathHelper;
 use OCP\Files\Folder;
 use OCP\Constants;
+use OCP\Files\IRootFolder;
 use OCP\Files\Mount\IMountPoint;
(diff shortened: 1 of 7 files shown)


icewind1991:
Also made `LazyFolder::get` not get the real folder if the path is known

lib/private/Files/Node/LazyFolder.php
```diff
@@ -390,9 +416,23 @@ public function getExtension(): string {
...
+		}
 		return $this->__call(__FUNCTION__, func_get_args());
 	}
 
+	public function isValidPath($path) {
+		if (!str_starts_with($path, '/')) {
+			$path = '/' . $path;
+		}
+		return !(str_contains($path, '/../') || strrchr($path, '/') === '/..');
+	}
```
come-nc, line 434: We do not have a method for this already? Manual checks like this always scare me.

skjnldsv (approved):

come-nc (approved):

PR #38159, opened 2023-05-09
fix too dark or bright primary element color

szaimen:
To reproduce: choose a very dark primary color and select the dark mode.

| Before | After |
|---|---|
| image | image |

### DIFF
--- a/apps/theming/tests/UtilTest.php
+++ b/apps/theming/tests/UtilTest.php
@@ -105,19 +105,34 @@ class UtilTest extends TestCase {
 		$this->util->invertTextColor('');
 	}
 
-	public function testElementColorDefault() {
+	public function testElementColorDefaultBlack() {
 		$elementColor = $this->util->elementColor("#000000");
+		$this->assertEquals('#4d4d4d', $elementColor);
+	}
+
+	public function testElementColorDefaultWhite() {
+		$elementColor = $this->util->elementColor("#ffffff");
+		$this->assertEquals('#b3b3b3', $elementColor);
+	}
+
+	public function testElementColorBlackOnDarkBackground() {
+		$elementColor = $this->util->elementColor("#000000", false);
+		$this->assertEquals('#4d4d4d', $elementColor);
+	}
+
+	public function testElementColorBlackOnBrightBackground() {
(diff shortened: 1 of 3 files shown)


nimishavijay (commented):
Looks good in the screenshot 👍 as long as it is not affected in light mode it is good to go

apps/theming/lib/IconBuilder.php
```diff
@@ -234,7 +234,13 @@ public function colorSvg($app, $image) {
 		}
 		$svg = file_get_contents($imageFile);
 		if ($svg !== false && $svg !== "") {
-			$color = $this->util->elementColor($this->themingDefaults->getColorPrimary());
+			$primaryColor = $this->themingDefaults->getColorPrimary();
+			$luminance = $this->util->calculateLuminance($primaryColor);
+			$isBrightColor = false;
+			if ($luminance > 0.8) {
```
nickvergessen, line 240: and cut off into the other direction for bright mode?

szaimen: The cut-off for dark-mode is already handled in elementColor so both will be fine. I verified this already

szaimen:
Can someone please restart drone for me? Thanks a lot!

apps/theming/lib/IconBuilder.php
szaimen: good point

apps/theming/tests/UtilTest.php
```diff
@@ -107,17 +107,17 @@ public function testInvertTextColorEmpty() {
 
 	public function testElementColorDefault() {
 		$elementColor = $this->util->elementColor("#000000");
```
nickvergessen, line 109: also test the function with true as second parameter with `#000000` and `#ffffff`?

szaimen: done

nickvergessen (approved):

susnux (approved):
Works, very nice!

jancborchardt (approved):
Looks good! :)

szaimen:
CI failure unrelated

szaimen:
I wonder should we backport this or rather not?

PR #38174, opened 2023-05-10
feat: Enhance template picker size and layout depending on aspect ratio

juliusknorr:
Adapt the template picker modal size to large to take more use of screen when available and adapt the sizing of the previews to be

- maximum 3 per line for landscape aspect ratio, size has been increased
- maximum 4 per line for portrait/square aspect ratio

## Summary

| Before | After |
|---|---|
| Screenshot 2023-05-10 at 12 18 21 | Screenshot 2023-05-10 at 12 11 02 |
| Screenshot 2023-05-10 at 12 18 48 | Screenshot 2023-05-10 at 12 11 15 |
| Screenshot 2023-05-10 at 12 18 37 | Screenshot 2023-05-10 at 12 11 44 |
| Screenshot 2023-05-10 at 12 18 30 | Screenshot 2023-05-10 at 12 11 29 |

## Checklist


apps/files/src/views/TemplatePicker.vue
```diff
@@ -136,6 +135,11 @@ export default {
 		 * @return {object}
 		 */
 		style() {
+			// Fallback to 16:9 landscape ratio
+			const ratio = this.provider.ratio ? this.provider.ratio : 1.77
+			// Landscape templates should be wider than tall ones
+			// We fit 3 templates per row at max for landscape and 4 for portrait
+			const width = ratio > 1 ? margin * 30 : margin * 20
```
szaimen, line 142: Shouldnt the layout just change from normal to large maybe depending on the screen size? This would feel less like a hack to me

juliushaertl: The size change was already discussed with @jancborchardt as a good improvement to make the previews more useful with larger size ;)

juliushaertl:
cc @nickvergessen As i noticed that talk has a similar picker implementation on its own

https://github.com/nextcloud/spreed/blob/master/src/components/NewMessageForm/NewMessageForm.vue#L489

szaimen (approved):
I see. Fine by me then but didnt test

szaimen:
However the buttons to confirm and cancel being so far apart on the left and right looks kind of strange to me...

nickvergessen:
Thanks for the ping, added a talk issue for tracking

jancborchardt (approved):
Looks great, very good idea @juliushaertl! :+1: as discussed! :)

> However the buttons to confirm and cancel being so far apart on the left and right looks kind of strange to me...

@szaimen yeah, although I wouldn’t put them directly next to each other either. Rather we could remove the Cancel button since there is already an X close on the top right. We also don’t have a "Cancel" in the Move/Copy dialog for example.

szaimen:
> Rather we could remove the Cancel button since there is already an X close on the top right. We also don’t have a "Cancel" in the Move/Copy dialog for example.

Good idea

juliushaertl:
Done:
Screenshot 2023-05-10 at 15 30 52

PR #1606, opened 2016-10-01
search design details

jancborchardt:
Some search design details I did a while ago. Including:
- fixing wording
- use correct styling/elements
- cut off leading slash

Please review @nextcloud/designers @nextcloud/javascript :)

PR #38179, opened 2023-05-10
fix share roots always being marked as writable

icewind1991:
Since movable mounts are always marked as "updatable" some extra logic is needed to split "updatable" into "renamable" and "writable"

Fixes https://github.com/nextcloud/viewer/issues/1606

### DIFF
--- a/lib/public/Files/DavUtil.php
+++ b/lib/public/Files/DavUtil.php
@@ -32,6 +32,9 @@
 
 namespace OCP\Files;
 
+use OCP\Constants;
+use OCP\Files\Mount\IMovableMount;
+
 /**
  * This class provides different helper functions related to WebDAV protocol
  *
@@ -73,10 +76,21 @@ class DavUtil {
 			$p .= 'D';
 		}
 		if ($info->isUpdateable()) {
-			$p .= 'NV'; // Renameable, Moveable
+			$p .= 'NV'; // Renameable, Movable
 		}
+
+		// since we always add update permissions for the root of movable mounts
+		// we need to check the shared cache item directly to determine if it's writable
+		$storage = $info->getStorage();
+		if ($info->getInternalPath() === '' && $info->getMountPoint() instanceof IMovableMount) {
+			$rootEntry = $storage->getCache()->get('');
+			$isWritable = $rootEntry->getPermissions() & Constants::PERMISSION_UPDATE;
+		} else {
+			$isWritable = $info->isUpdateable();
+		}
+
 		if ($info->getType() === FileInfo::TYPE_FILE) {
-			if ($info->isUpdateable()) {
+			if ($isWritable) {
 				$p .= 'W';
 			}
 		} else {
--- a/apps/dav/tests/unit/Connector/Sabre/NodeTest.php
+++ b/apps/dav/tests/unit/Connector/Sabre/NodeTest.php
@@ -25,14 +25,20 @@
  * along with this program. If not, see <http://www.gnu.org/licenses/>
  *
  */
+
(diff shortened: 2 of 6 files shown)


lib/public/Files/DavUtil.php
```diff
@@ -75,12 +80,23 @@ public static function getDavPermissions(FileInfo $info): string {
 		if ($info->isUpdateable()) {
 			$p .= 'NV'; // Renameable, Moveable
 		}
+
+		// since we always add update permissions for the root of movable mounts (and thus shares)
+		// we need to check the shared cache item directly to determine if it's writable
+		$storage = $info->getStorage();
+		if ($info->getInternalPath() === '' && $storage->instanceOfStorage(SharedStorage::class)) {
```
skjnldsv, line 87: Weird psalm error :thinking:

come-nc: @icewind1991 Can this be put into `isUpdatable` instead?
Or in new methods `isWritable`/…?

icewind1991: Can't be put into `isUpdatable` because that is also used to check for rename permissions.

Adding a new method to the interface is wrought with compatibility issues

marcelklehr: > is wrought with compatibility pitfalls

what would those be?

icewind1991: It means that any (possibly 3rdparty) implementation of the interface needs to be updated as well

come-nc:
> Any update in this? Would be cool if we could solve #1606 soon

Wrong ticket number I suppose?

come-nc:
> > Any update in this? Would be cool if we could solve #1606 soon
> 
> Wrong ticket number I suppose?

Oh okay wrong repository, it’s https://github.com/nextcloud/viewer/issues/1606

lib/public/Files/DavUtil.php
```diff
@@ -75,8 +78,19 @@ public static function getDavPermissions(FileInfo $info): string {
...
```
come-nc, line 85:
```suggestion
		/** @psalm-suppress UndefinedClass files_sharing may be disabled but this will still work as expected */
		if ($info->getInternalPath() === '' && $storage->instanceOfStorage(SharedStorage::class)) {
```

Add a suppress like this then, I do not see a better way. And `::class` does not actually complains on non-existing classes so it should be fine. Maybe still test this with files_sharing disabled to be sure.

A clean fix would be to add a new interface for this, like IWritableAwareStorage, and change the code for:
```suggestion
		if ($storage->instanceOfStorage(IWritableAwareStorage::class)) {
			$isWritable = $storage->isWritable($info);
```

marcelklehr:
Hello there!

What is the current status on this? Assume another review round is needed? Or is there still a disagreement that we need to resolve on how to solve this issue? Would be cool if we could cross this issue off soon :blue_heart: 

Cheers

marcelklehr (approved):

come-nc (approved):

marcelklehr:
Some CI runs are failling.

marcelklehr:
image

nfebe (approved):
Looks good

ArtificialOwl (approved):

marcelklehr:
More approvals will not help, I'm afraid. the tests and CI failures need to be fixed cc @icewind1991

GretaD:
I need this to be merged as soon as possible for a customer

icewind1991:
https://github.com/nextcloud/server/pull/39614 for the backport

PR #38180, opened 2023-05-10
fix(user_status): Also delete backup status when reverting a hard-cod…

nickvergessen:
…ed status

* Resolves: #38171

## Summary

I see how this could help, but while trying to test it, I was unable to reproduce the original issue and when digging further I found: 
https://github.com/nextcloud/server/blob/c3c3dcbcd9ff9b8aa9c68a5b7b19507c2c334e40/apps/dav/lib/BackgroundJob/UserStatusAutomation.php#L164-L165
which should prevent the original issue already.
So at the moment I'm unsure if we should apply the patch or first wait until we know how it happens so we can write a test for it.

But I didn't want to loose the idea, so here is a draft PR never the less.

## Checklist

### DIFF
--- a/apps/user_status/lib/Service/StatusService.php
+++ b/apps/user_status/lib/Service/StatusService.php
@@ -500,14 +500,19 @@ class StatusService {
 		try {
 			/** @var UserStatus $userStatus */
 			$backupUserStatus = $this->mapper->findByUserId($userId, true);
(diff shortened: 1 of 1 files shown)


nickvergessen:
cc @miaulalala thoughts?

miaulalala (approved):
Logic looks sound to me.

PR #38190, opened 2023-05-10
chore(php): Refactor dav app empty array check

ChristophWurst:
* Resolves: # 

## Summary

https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/

## TODO

- [x] Refactor

## Checklist


apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -1154,7 +1154,7 @@ public function getCalendarObject($calendarId, $objectUri, int $calendarType = s
 	 * @return array
 	 */
 	public function getMultipleCalendarObjects($calendarId, array $uris, $calendarType = self::CALENDAR_TYPE_CALENDAR):array {
-		if (empty($uris)) {
+		if ($uris === []) {
```
come-nc, line 1157: I really do not see the point of such a change when the var is clearly typed as an array. `empty` works and is easy to read.

ChristophWurst: Alright

PR #38195, opened 2023-05-11
Update icewind/streams to 0.7.7 in files_external

come-nc:
## Summary

Solves psalm issues about return types in files_external.
A bit weird that we have icewind/streams is two places but it’s a dependency of icewind/smb which is only pulled in files_external, so not sure this is fixable.

## Checklist


apps/files_external/3rdparty/composer.json
```diff
@@ -8,7 +8,6 @@
 		"classmap-authoritative": true
 	},
 	"require": {
-		"icewind/streams": "0.7.4",
```
icewind1991, line 11: iirc. the explicit requirement was added to ensure that both copies of the library are at the same version so that no matter which of the copies wins the autoload race we get the same code.

icewind1991 (changes requested):

provokateurin (approved):

icewind1991 (approved):

PR #38216, opened 2023-05-12
chore: Update Nextcloud 27 authors

ChristophWurst:
## Summary

Updates header info and AUTHORS.

The diff is very big because we forgot to update the license info for two years.

## Checklist


apps/comments/src/services/EditComment.js
```diff
@@ -21,15 +21,6 @@
  */
 
 import client from './DavClient.js'
-
-/**
- * Edit an existing comment
- *
- * @param {string} commentsType the ressource type
- * @param {number} ressourceId the ressource ID
```
ChristophWurst, line 29: this is very unfortunate. the license script must be buggy.

nickvergessen:
I'm still not sure if removing `@author` is legit, so maybe we only bump the AUTHORS file with additions?

PR #38223, opened 2023-05-12
Use the color-primary-element* variables

szaimen:
Explanation: the color-primary variables are not to be used in components because the introduce problems with high-contrast primary colors. Fix this by using the primary-element variables instead.


skjnldsv (approved):

Pytal (approved):

szaimen:
Integration-contacts-menu failure unrelated

PR #38225, opened 2023-05-12
disable pull-down-to-refresh for mobile

szaimen:
Disable pull-down-to-refresh for mobile since this sometimes causes unintended refreshs on mobile when scrolling up.

The solution is based on https://thewebdev.info/2021/10/09/how-to-prevent-pull-down-to-refresh-in-mobile-chrome/


jancborchardt (approved):
Seems good!

Just to clarify as I was a bit confused initially, this disables the browser native refresh, not anything from Nextcloud. (Which Firefox mobile also has since a recent update.)

miaulalala (approved):

susnux (approved):

PR #38226, opened 2023-05-12
add utility command for object store objects

icewind1991:
Adds some basic utility command for working directly with objects on the object store.

While admins should usually go through the regular file utilities instead of touching the object store directly. In some cases, such as if there is a desync between the object store and filecache, this isn't an option.
If we detect that the admin is trying to touch an object that belongs to an existing file, a warning is shown.

By having an option to do this directly trough `occ` it removes the requirement for admins to install 3rdparty tools such as `s3cmd` and the object store configuration can be infered from the nextcloud config.


apps/files/lib/Command/Object/Get.php
```diff
@@ -0,0 +1,78 @@
...
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('files:object:get')
+			->setDescription('Get the contents of an object')
+			->addArgument('object', InputArgument::REQUIRED, "Object to get")
+			->addArgument('output', InputArgument::REQUIRED, "Target local file to output to")
+			->addOption('bucket', 'b', InputOption::VALUE_REQUIRED, "Bucket to delete the object from, only required in cases where it can't be determined from the config");
```
artonge, line 47:
```suggestion
			->addOption('bucket', 'b', InputOption::VALUE_REQUIRED, "Bucket to get the object from, only required in cases where it can't be determined from the config");
```

apps/files/lib/Command/Object/Get.php
```diff
@@ -0,0 +1,78 @@
...
+		$this->objectUtils = $objectUtils;
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('files:object:get')
+			->setDescription('Get the contents of an object')
+			->addArgument('object', InputArgument::REQUIRED, "Object to get")
+			->addArgument('output', InputArgument::REQUIRED, "Target local file to output to")
```
artonge, line 46:
```suggestion
			->addArgument('output', InputArgument::REQUIRED, "Target local file to output to. Default to stdout")
```

artonge (approved):
Some copy-pasting errors, but fine :+1:

apps/files/lib/Command/Object/Get.php
```diff
@@ -0,0 +1,78 @@
...
+		if (!$objectStore) {
+			return 1;
+		}
+
+		if (!$objectStore->objectExists($object)) {
+			$output->writeln("<error>Object $object does not exist</error>");
+			return 1;
+		} else {
+			$source = $objectStore->readObject($object);
+			if (!$source) {
```
come-nc, line 63: Psalm say this is not possible and phpdoc in OCP agrees. Not sure if true or not.

apps/files/lib/Command/Object/ObjectUtil.php
```diff
@@ -0,0 +1,99 @@
...
+	public function objectExistsInDb(string $object): int|false {
+		if (str_starts_with($object, 'urn:oid:')) {
+			$fileId = (int)substr($object, strlen('urn:oid:'));
+			$query = $this->connection->getQueryBuilder();
+			$query->select('fileid')
+				->from('filecache')
+				->where($query->expr()->eq('fileid', $query->createNamedParameter($fileId, IQueryBuilder::PARAM_INT)));
+			$result = $query->executeQuery();
+			if ($result->fetchOne() !== false) {
+				return $fileId;
```
come-nc, line 91: is 0 a valid value for fileId here? If yes callers need to be adapted to explicitely check against false.

apps/files/lib/Command/Object/Get.php
```diff
@@ -0,0 +1,78 @@
...
+				$output->writeln("<error>Failed to read $object from object store</error>");
+				return 1;
+			}
+			$target = ($outputName === null || $outputName === '-') ? STDOUT : fopen($outputName, 'w');
```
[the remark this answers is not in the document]
icewind1991: correct

apps/files/lib/Command/Object/Get.php
icewind1991: Yes, unlike the `files:get`, determining the mimetype here isn't reliable so I don't think it can do a good enough "don't write binary data to the terminal" safety.

come-nc (approved):

PR #38240, opened 2023-05-13
Use aria-expanded correctly on toggle user actions

nfebe:
The aria-expanded attribute was added on the expandable item rather Than the button that triggers the expand.

Resolves : https://github.com/nextcloud/server/issues/37137


artonge (approved):

artonge:
Ah, looks like some tests are failing:

```
       ✗ shows conflict dialog when no client side conflict
	Expected false to equal true.
	    at apps/files/tests/js/fileUploadSpec.js:224:42
```

Fenn-CS:
Messed this up and for some reason forcing pushing the commits don't seem to fix it, so open a new PR see : #38294

PR #38256, opened 2023-05-15
Adjust file custom picker component padding

julien-nc:
Because the smart picker does not have padding anymore.


juliusknorr (approved):

szaimen (approved):
LGTM but didnt test

susnux (approved):

PR #38261, opened 2023-05-15
Refactors "strpos" calls in  lib/private to improve code readability.

fsamapoor:

### DIFF
--- a/lib/private/Files/Type/Detection.php
+++ b/lib/private/Files/Type/Detection.php
@@ -116,7 +116,7 @@ class Detection implements IMimeTypeDetector {
 
 		// Update the alternative mimetypes to avoid having to look them up each time.
 		foreach ($this->mimetypes as $extension => $mimeType) {
-			if (strpos($extension, '_comment') === 0) {
+			if (str_starts_with($extension, '_comment')) {
 				continue;
 			}
 			$this->secureMimeTypes[$mimeType[0]] = $mimeType[1] ?? $mimeType[0];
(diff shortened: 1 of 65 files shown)


nfebe (approved):
LGTM

lib/private/Files/Type/Detection.php
```diff
@@ -238,15 +238,15 @@ public function detectContent(string $path): string {
...
+				$mimeType = str_contains($info, ';') ? substr($info, 0, strpos($info, ';')) : $info;
 				$mimeType = $this->getSecureMimeType($mimeType);
 				if ($mimeType !== 'application/octet-stream') {
 					return $mimeType;
 				}
 			}
 		}
 
-		if (strpos($path, '://') !== false && strpos($path, 'file://') === 0) {
+		if (str_contains($path, '://') && str_starts_with($path, 'file://')) {
```
icewind1991, line 249: these 2 can be combined as the latter implies the former

lib/private/TemplateLayout.php
```diff
@@ -284,7 +284,7 @@ public function __construct($renderAs, $appId = '') {
 			} else {
 				$suffix = $this->getVersionHashSuffix($web, $file);
 
-				if (strpos($file, '?v=') == false) {
+				if (!strpos($file, '?v=')) {
```
icewind1991, line 287: `!str_contains`

icewind1991 (approved):

lib/private/TemplateLayout.php
fsamapoor: Thanks. Done.

PR #38274, opened 2023-05-15
fix(middleware): Also abort the request when reaching max delay in af…

nickvergessen:
…terController

Best reviewed with ?w=1


come-nc (commented):
Why is the diff much shorter in backports?

come-nc (approved):
Understood, it’s because attribute are supported on master and not on previous versions

miaulalala (approved):

PR #38300, opened 2023-05-16
fix: better error message on missing user files dir

skjnldsv:
When the user directory is missing,t he error thrown is very cryptic. 
Now it's a bit better and we know the error.

I hesitated to to have a real error page, but I figured this is more like an admin info to have.

image

### DIFF
--- a/lib/private/legacy/OC_Helper.php
+++ b/lib/private/legacy/OC_Helper.php
@@ -487,7 +487,7 @@ class OC_Helper {
 			$rootInfo = \OC\Files\Filesystem::getFileInfo($path, $includeExtStorage ? 'ext' : false);
 		}
 		if (!$rootInfo instanceof \OCP\Files\FileInfo) {
-			throw new \OCP\Files\NotFoundException();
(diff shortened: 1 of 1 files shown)


apps/files/lib/Controller/ViewController.php
```diff
@@ -246,7 +246,11 @@ public function index($dir = '', $view = '', $fileid = null, $fileNotFound = fal
 			// If view is files, we use the directory, otherwise we use the root storage
 			$storageInfo =  $this->getStorageInfo(($view === 'files' && $dir) ? $dir : '/');
 		} catch(\Exception $e) {
-			$storageInfo = $this->getStorageInfo();
+			try {
+				$storageInfo = $this->getStorageInfo();
+			} catch (\Exception $e) {
+				throw new \Exception('The user files directory is missing');
```
skjnldsv, line 252: To be fair, we should maybe directly throw this at the source.

artonge (approved):

icewind1991 (approved):

PR #38301, opened 2023-05-16
fix(caldav): Ignore invalid events for reminder generation

ChristophWurst:
## Summary

Fixes ``Call to a member function serialize() on null at apps/dav/lib/CalDAV/Reminder/ReminderService.php:261`` with 25.0.4.

## TODO

- [x] Fix

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/Reminder/ReminderService.php
+++ b/apps/dav/lib/CalDAV/Reminder/ReminderService.php
@@ -793,6 +793,10 @@ class ReminderService {
 			if ($child->name !== 'VEVENT') {
 				continue;
 			}
+			// Ignore invalid events with no DTSTART
+			if ($child->DTSTART === null) {
+				continue;
+			}
 
 			$vevents[] = $child;
 		}
--- a/apps/dav/tests/unit/CalDAV/Reminder/ReminderServiceTest.php
+++ b/apps/dav/tests/unit/CalDAV/Reminder/ReminderServiceTest.php
@@ -342,6 +342,53 @@ ICS;
 		$this->reminderService->onCalendarObjectCreate($objectData);
 	}
 
+	/**
+	 * RFC5545 says DTSTART is REQUIRED, but we have seen event without the prop
+	 */
+	public function testOnCalendarObjectCreateNoDtstart(): void {
+		$calendarData = <<<EOD
+BEGIN:VCALENDAR
+PRODID:-//Nextcloud calendar v1.6.4
+BEGIN:VEVENT
+CREATED:20160602T133732
+DTSTAMP:20160602T133732
+LAST-MODIFIED:20160602T133732
+UID:wej2z68l9h
+SUMMARY:Test Event
+BEGIN:VALARM
+ACTION:EMAIL
+TRIGGER:-PT15M
+END:VALARM
+BEGIN:VALARM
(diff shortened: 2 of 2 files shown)


apps/dav/lib/CalDAV/Reminder/ReminderService.php
```diff
@@ -267,7 +275,14 @@ public function onCalendarObjectCreate(array $objectData):void {
 		if ($masterItem) {
 			$processedAlarms = [];
 			$masterAlarms = [];
-			$masterHash = $this->getEventHash($masterItem);
+			try {
+				$masterHash = $this->getEventHash($recurrenceException);
```
tcitworld, line 279: **nitpick:** 
```suggestion
				$masterHash = $this->getEventHash($masterItem);
```

apps/dav/lib/CalDAV/Reminder/ReminderService.php
```diff
@@ -243,7 +244,14 @@ public function onCalendarObjectCreate(array $objectData):void {
 		$isRecurring = $masterItem ? $this->isRecurring($masterItem) : false;
 
 		foreach ($recurrenceExceptions as $recurrenceException) {
-			$eventHash = $this->getEventHash($recurrenceException);
+			try {
+				$eventHash = $this->getEventHash($recurrenceException);
+			} catch (Throwable $e) {
```
tcitworld, line 249: **question:** Which exceptions do you except to be sent here? From what I can see serializing properties, using `implode` and `md5` shouldn't throw anything.

ChristophWurst: ``Call to a member function serialize() on null at apps/dav/lib/CalDAV/Reminder/ReminderService.php:261``

Added it to the PR description too. We saw this on a production instance, during the upgrade from an old version. Apparently there can be events without a DTSTART.

ChristophWurst: Do you have any ideas for a better place to catch this? We might be able to skip invalid events all together for the reminder processing.

tcitworld: Not catching, but we can simply validate events in `getAllVEventsFromVCalendar`.

Something like:
```php
if (!$child->DTSTART) {
  continue
}
```

apps/dav/tests/unit/CalDAV/Reminder/ReminderServiceTest.php
```diff
@@ -342,6 +342,53 @@ public function testOnCalendarObjectCreateSingleEntry():void {
...
+END:VCALENDAR
+EOD;
+		$objectData = [
+			'calendardata' => $calendarData,
+			'id' => '42',
+			'calendarid' => '1337',
+			'component' => 'vevent',
+		];
+
+		$this->backend->expects($this->exactly(2))
```
tcitworld, line 376: **question:** Shouldn't the method never be called?

tcitworld (changes requested):

tcitworld (approved):

kesselb (approved):

hamza221 (approved):

ChristophWurst:
the backport worked locally

PR #38321, opened 2023-05-16
docs: add dbtableprefix to sample config

pulsejet:
Unfortunately (and naturally), people copy values from the sample config and end up with databases with no prefix. This causes issues for Doctrine in certain corner cases (related https://github.com/pulsejet/memories/issues/648).

So this is a simple proposal to set the sample config value to the default value.

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -136,7 +136,7 @@ $CONFIG = [
  *
  * Default to ``oc_``
  */
-'dbtableprefix' => '',
+'dbtableprefix' => 'oc_',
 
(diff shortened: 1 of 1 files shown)


szaimen (approved):

nickvergessen (approved):
Fine by me, but as far as I'm aware none of the installation options allow setting it anymore, so it should always be oc_
It was removed for the same reason you outlined, it can break without a prefix in certain corner cases.

szaimen:
CI failure unrelated

PR #38322, opened 2023-05-16
fix: Throw early when a service cannot be found in the app container

### LINKED ISSUE #10263: Inner exception is lost during class injection creating a misleading error message
During instantiation of a class during injection if the queried class (or a class inside it) throws an exception this exception is not propagated up and lost.  Instead a misleading error message similar to 

>{index} Exception: {"Exception":"OCP\\AppFramework\\QueryException","Message":"Could not resolve db! Class db does not exist","Code":0,"Trace":"#0 \/var│
Ju│\/www\/nextcloud\/lib\/private\/AppFramework\/Utility\/SimpleContainer.php(118): OC\\AppFramework\\Utility\\SimpleContainer->resolve('db')\n#1 \/var\/www\/nextcloud\/lib\/private\/ServerConta│
Ju│iner.php(132): OC\\AppFramework\\Utility\\SimpleContainer->query('db')\n#2 \/var\/www\/nextcloud\/lib\/private\/AppFramework\/Utility\/SimpleContainer.php(74): OC\\ServerContainer->query('db'│
Ju│)\n#3 \/var\/www\/nextcloud\/lib\/private\/AppFramework\/Utility\/SimpleContainer.php(97): OC\\AppFramework\\Utility\\SimpleContainer->buildClass(Object(ReflectionClass))\n#4 \/var\/www\/next│
Ju│cloud\/lib\/private\/AppFramework\/Utility\/SimpleContainer.php(118): OC\\AppFramework\\Utility\\SimpleContainer->resolve('OCA\\\\UserBackend...')\n#5 \/var\/www\/nextcloud\/lib\/private\/Ser│
Ju│verContainer.php(132): OC\\AppFramework\\Utility\\SimpleContainer->query('OCA\\\\UserBackend...')\n#6 \/var\/www\/nextcloud\/apps\/user_backend_sql_raw\/lib\/AppInfo\/Application.php(56): OC\│
Ju│\ServerContainer->query('OCA\\\\UserBackend...')\n#7 \/var\/www\/nextcloud\/apps\/user_backend_sql_raw\/appinfo\/app.php(21): OCA\\UserBackendSqlRaw\\AppInfo\\Application->__construct()\n#8 \│
Ju│/var\/www\/nextcloud\/lib\/private\/legacy\/app.php(245): require_once('\/var\/www\/nextcl...')\n#9 \/var\/www\/nextcloud\/lib\/private\/legacy\/app.php(154): OC_App::requireAppFile('user_bac│
Ju│kend_sq...')\n#10 \/var\/www\/nextcloud\/lib\/private\/legacy\/app.php(127): OC_App::loadApp('user_backend_sq...')\n#11 \/var\/www\/nextcloud\/lib\/private\/legacy\/user.php(131): OC_App::loa│
Ju│dApps(Array)\n#12 \/var\/www\/nextcloud\/lib\/base.php(729): OC_User::setupBackends()\n#13 \/var\/www\/nextcloud\/lib\/base.php(1081): OC::init()\n#14 \/var\/www\/nextcloud\/index.php(40): re│
Ju│quire_once('\/var\/www\/nextcl...')\n#15 {main}","File":"\/var\/www\/nextcloud\/lib\/private\/AppFramework\/Utility\/SimpleContainer.php","Line":103}

is thrown. The class `Db`exists (not `db` as the error message erroneously reports). This happens [in the second branch](https://github.com/nextcloud/server/blob/23785d108d6f1a5641bad2cf91ae3c696aac0ef0/lib/private/AppFramework/Utility/SimpleContainer.php#L81) of `SimpleContainer.php` in the `query()` method. Imho the exception should be rethrown.

# Screenshot of debugger state
screenshot_20180717_033324

On line 68 you can see the thrown exception, but still on line 72 the second branch is chosen just because the class name (Db) is known. But returning the actual exception message is much more useful than simply returning the name of the immediate class name.

# Steps to reproduce
I don't have a minimal example but one way would be to install the [User Backend Raw SQL app](https://apps.nextcloud.com/apps/user_backend_sql_raw) and set `db_type` in the [configuration](https://github.com/PanCakeConnaisseur/user_backend_sql_raw#configuration) to e.g. `mysql` which is not a supported string, which leads to the above error.

# Versions
Tested with Nextcloud 13.0.4 and 13.0.5 RC 2 Build:2018-07-16T23:01:35+00:00 43785cf1a80892e35a133b76a087eb52d19d4060.

# Proposed solution
I don't understand the implementation of dependency injection in Nextcloud completely, but I think all  thrown exceptions should be rethrown.

juliusknorr:
## Summary

I noticed that when an app has a wrongly named file for the matching class name like a class TestService being in a file CoreService.php and then being used through dependency injection the error message was not very useful. This PR aims to make error messages more useful in case loading a class within an app fails.

The reason for that is that afterwards the server container is queried which might fail due to other missing dependencies anyways. Since we already check and only run this code path for getting services within the OCA\ namespace we can probably just throw directly as I'm not aware of any OCA classes that could land in the server container.

Would merge for 28 only as it has a bit of a risk of breaking things.

fix https://github.com/nextcloud/server/issues/10263

## Checklist

### DIFF
--- a/lib/private/ServerContainer.php
+++ b/lib/private/ServerContainer.php
@@ -154,8 +154,10 @@ class ServerContainer extends SimpleContainer {
 			try {
 				return $appContainer->queryNoFallback($name);
 			} catch (QueryException $e) {
-				// Didn't find the service or the respective app container,
-				// ignore it and fall back to the core container.
+				// Didn't find the service or the respective app container
+				// In this case the service won't be part of the core container,
+				// so we can throw directly
+				throw $e;
 			}
 		} elseif (strpos($name, 'OC\\Settings\\') === 0 && substr_count($name, '\\') >= 3) {
 			$segments = explode('\\', $name);


lib/private/ServerContainer.php
```diff
@@ -154,8 +154,10 @@ public function query(string $name, bool $autoload = true) {
 			try {
 				return $appContainer->queryNoFallback($name);
 			} catch (QueryException $e) {
-				// Didn't find the service or the respective app container,
-				// ignore it and fall back to the core container.
+				// Didn't find the service or the respective app container
+				// In this case the service won't be part of the core container,
+				// so we can throw directly
+				throw $e;
 			}
```
come-nc, line 161: The try/catch can be removed, no?

come-nc (approved):

lib/private/ServerContainer.php
juliushaertl: I felt it makes reading the methods code a bit easier knowing explicitly that it may throw at this place

juliushaertl:
🏓 for reviews :)

marcelklehr (approved):

ChristophWurst (approved):

PR #35432, opened 2022-11-25
Prune core-common js and lazy load modules

pulsejet:
Enables lazy loading of libraries in the JS bundle. This improves perceivable UI loading time ~~by almost 2x~~ (I realised my dev/test setup isn't ideal for testing this) in my test environment (to load the Files app) and reduces the transferred data from ~5MB to ~3MB. These results are shown beside the screenshots.

The most painful things loaded in `core-common.js` (now separate):
1. Nextcloud vue components. These are huge and not always required.
2. The password confirmation dialog. This one is the biggest.
3. Emoji and moment JSON

Caveat:
1. The total size of everything is slightly larger, but this is fine, since not everything is loaded (much less, actually)
2. Slightly increased complexity. I doubt it's significant.

**Production bundle before optimization:**
before

**After pruning**:
after

PR #38329, opened 2023-05-17
Optimize webpack chunking

pulsejet:
Supercedes and closes #35432

This takes a much simpler approach for achieving something similar to the linked PR. Instead of lazy loading every component, webpack directly allows only splitting components used in more than `n` modules. By tweaking this parameter (3 seems to have great results), we can split the chunks efficiently achieving similar performance gains. This is also much less invasive and more semantically meaningful.

Some amount of async loading is still needed since apps greedily attempt to load a lot of things.

|  wp_before | wp_after |

Files app (no cache):

| 135 requests             | 138 requests             |
| 5.3 MB transferred       | 3.7 MB transferred       |
| 21.0 MB resources        | 12.2 MB resources        |
| DOMContentLoaded: 4.19 s | DOMContentLoaded: 3.02 s |

Files app (with cache):

| 133 requests             | 133 requests             |
| 72.4 kB transferred      | 72.4 kB transferred      |
| 21.0 MB resources        | 12.2 MB resources        |
| DOMContentLoaded: 2.60 s | DOMContentLoaded: 1.39 s |

Memories (no cache):

| 83 requests              | 83 requests              |
| 5.2 MB transferred       | 3.3 MB transferred       |
| 21.8 MB resources        | 12.3 MB resources        |
| DOMContentLoaded: 3.15 s | DOMContentLoaded: 2.25 s |

Video of load performance (before on top, after on bottom)
https://github.com/nextcloud/server/assets/10709794/5ab1144e-c811-4b8d-9798-463982b46cb7


juliushaertl:
I think this is a great initiative. I'd have one concern which is due to the fact that we currently still require to have the bundles committed in the repository itself. The dynamic naming with numeric identifiers has the risk that we end up having leftover chunks in the dist folder if they are not cleaned up before each production build, so we need to double check that this is the case.

juliushaertl:
Ok, might actually be the case already with https://github.com/nextcloud/server/blob/pulsejet/patch-webpack/webpack.common.js#L59-L61

juliusknorr (commented):
Eslint is not happy, otherwise this seems good from my perspective.

artonge (approved):

marcelklehr:
Thank you @pulsejet for not giving up on this :blue_heart:

pulsejet:
> Ok, might actually be the case already with https://github.com/nextcloud/server/blob/pulsejet/patch-webpack/webpack.common.js#L59-L61

Never knew about this, but seems it'll do the trick: https://webpack.js.org/guides/output-management/#cleaning-up-the-dist-folder.

ESLint should pass now.

pulsejet:
Just noticed something funny: this is actually also a _bug fix_ at certain points. For instance, we explicitly _want_ to do lazy loading (before this PR) at this line for `caniuse-lite`. But as is evident from the bundle, the library was present in `core-common.js` all along, so the lazy loading never really worked.
https://github.com/nextcloud/server/blob/fa3bbb26538f8b56b2ab91dcfe3972a27a550d19/core/src/views/UnsupportedBrowser.vue#L125-L130

pulsejet:
Short before (top) and after (bottom) video for this (warm cache for both):

https://github.com/nextcloud/server/assets/10709794/5ab1144e-c811-4b8d-9798-463982b46cb7

pulsejet:
Bump? I really don't understand what keeps blocking this...

artonge:
@nextcloud/server-frontend could you review/approve ?

kesselb:
@pulsejet rebased, compiled and force-pushed your branch. 

Used node 18.7 and npm 8.15 to build locally.

pulsejet:
Thanks @kesselb. Just curious: doesn't the bot do that usually or I should be pushing compiled assets to the PR?

skjnldsv (approved):

skjnldsv:
> Bump? I really don't understand what keeps blocking this...

27 being released and master being in feature freeze.
This is now branched-off and we can resume

skjnldsv:
> Thanks @kesselb. Just curious: doesn't the bot do that usually or I should be pushing compiled assets to the PR?

the bot needs to be called for it :)
`/compile amend /`

major-mayer:
Now that I finally upgraded my instance to 28, I want to say thank you for this PR again @pulsejet 
This increases the performance noticeably!

PR #38333, opened 2023-05-17
weather_status: Improve contrast of favorite icon

### LINKED ISSUE #36969: [BITV] 9.1.4.11/2.3 - "Weather" menu - The "Favorites" button does not meet the contrast requirements for graphical controls, whether it is active or inactive. (1)
- [ ] Fix
- [ ] Backport

20ff47cad6376bf69aae1fdaf0a50b3e

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-3dbfb7615a-v2-n3

</details>

susnux:
* Resolves: #36969

## Summary
Do not use a yellow icon but the default colored one and use the `StarOff` icon for removing a favorite instead of using `StarOutline`.

This also adds a feature: The current selected favorite is highlighted by setting the icon color to the primary color.

## screenshots
before | after
---|---
image|image
image|image
image|image

## Checklist


szaimen:
@nimishavijay @jancborchardt I think we need to think here how to properly display the current state: enabled <-> disabled

susnux:
> @nimishavijay @jancborchardt I think we need to think here how to properly display the current state: enabled <-> disabled

Yes, mainly we should  decide whether we want to keep the yellow star or not, because we have the same problem on the files list (yellow star on white background => too low contrast).

jancborchardt (changes requested):
StarOff looks strange, cause it looks as if a click on it would remove the star.

Instead we should use the "non-filled" star icon for the state when it's not a favorite, that is more common.

Regarding color, just dark is fine, or do we find maybe a dark yellow/gold that still looks like a star color? In dark mode it can/must of course be light yellow.

susnux:
> do we find maybe a dark yellow/gold that still looks like a star color

@jancborchardt yellow will not work on white background, but dark golden at least fulfills the AA criteria. Will look like this:
image
image

Do you think this looks better than the default color?

jancborchardt:
@susnux yep, I would say that dark gold is good! It at least conveys a little bit of a star color, otherwise favorite gets lost in the sea of other functions.

(Different issue, but we should probably also make sure that the "Delete" icon in all the action menus is colored accordingly – also something for the Penpot library @nimishavijay )

jancborchardt (approved):

apps/weather_status/src/App.vue
```diff
@@ -602,6 +609,17 @@ export default {
     min-height: 44px !important;
 }
 
+// Set color to primary element for current / active favorite address
+.current-favorite {
+	//color: var(--color-primary-element)!important;
```
Pytal, line 614:
```suggestion
```

Pytal (approved):

susnux:
I had to adjust the color slightly to `#a08b00` to guarantee the contrast even when hovered.

nfebe (approved):

szaimen:
Conflicts

susnux:
> Conflicts

Resolved

PR #38367, opened 2023-05-19
list reserved date and allow filtering by reservered for background job list command

icewind1991:
- Add "reserved at" to the output of `occ background-job:list`
- Add option to only list reserved jobs with `occ background-job:list --reserved`

Should be useful when investigating cron issues

### DIFF
--- a/lib/private/BackgroundJob/Job.php
+++ b/lib/private/BackgroundJob/Job.php
@@ -33,11 +33,9 @@ use OCP\ILogger;
  * @deprecated internal class, use \OCP\BackgroundJob\Job
  */
 abstract class Job implements IJob {
-	/** @var int */
-	protected $id;
-
-	/** @var int */
-	protected $lastRun;
+	protected int $id;
+	protected int $lastRun;
+	protected int $reservedAt;
 
 	/** @var mixed */
 	protected $argument;
@@ -80,6 +78,10 @@ abstract class Job implements IJob {
 		$this->lastRun = $lastRun;
 	}
 
+	public function setReservedAt(int $reservedAt): void {
+		$this->reservedAt = $reservedAt;
+	}
+
 	public function setArgument($argument) {
 		$this->argument = $argument;
 	}
@@ -94,5 +96,9 @@ abstract class Job implements IJob {
 
 	public function getArgument() {
 		return $this->argument;
+	}
+
+	public function getReservedAt(): int {
+		return $this->reservedAt;
 	}
 }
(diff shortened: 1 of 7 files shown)


lib/private/BackgroundJob/Job.php
```diff
@@ -80,6 +78,10 @@ public function setLastRun(int $lastRun) {
 		$this->lastRun = $lastRun;
 	}
 
+	public function setReservedAt(int $reservedAt) {
```
come-nc, line 81:
```suggestion
	public function setReservedAt(int $reservedAt): void {
```

lib/public/BackgroundJob/IJob.php
```diff
@@ -87,6 +87,12 @@ public function setLastRun(int $lastRun);
 	 */
 	public function setArgument($argument);
 
+	/**
+	 * @param int $reservedAt
+	 * @since 28.0.0
+	 */
+	public function setReservedAt(int $reservedAt);
```
come-nc, line 94:
```suggestion
	public function setReservedAt(int $reservedAt): void;
```

lib/public/BackgroundJob/Job.php
```diff
@@ -119,6 +120,10 @@ public function setArgument($argument) {
 		$this->argument = $argument;
 	}
 
+	public function setReservedAt(int $reservedAt) {
```
come-nc, line 123:
```suggestion
	public function setReservedAt(int $reservedAt): void {
```

lib/private/BackgroundJob/JobList.php
```diff
@@ -190,6 +190,10 @@ public function getJobsIterator($job, ?int $limit, int $offset): iterable {
 			$query->where($query->expr()->eq('class', $query->createNamedParameter($class)));
 		}
 
+		if ($reservedOnly) {
+			$query->where($query->expr()->gt('reserved_at', $query->createNamedParameter($this->timeFactory->getTime() - 12 * 3600, IQueryBuilder::PARAM_INT)));
```
come-nc, line 194: Why is that a comparison with a timestamp and not just a test if it is set?
It means some job can have a value in "reserved at" column and not appear when using `--reserved`.

lib/private/BackgroundJob/JobList.php
```diff
@@ -320,6 +324,7 @@
 			$job->setId((int) $row['id']);
 			$job->setLastRun((int) $row['last_run']);
 			$job->setArgument(json_decode($row['argument'], true));
+			$job->setReservedAt((int) $row['reserved_at']);
```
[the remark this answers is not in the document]
come-nc: You need to add the new column in phpdoc 	 `* @param array{class:class-string<IJob>, id:mixed, last_run:mixed, argument:string} $row`

lib/private/BackgroundJob/JobList.php
icewind1991: The reservation expires after 12h (to deal with cron crashing and not releasing it's reservation).

So not filtering would lead to showing items that aren't reserved anymore according to the joblist logic

come-nc (approved):

PR #38386, opened 2023-05-21
refactor: migrate OC_EventSource to dependency injection

kesselb:
* Resolves: # 

## Summary

Test: Increase the version number in version.php and visit Nextcloud. 
The update should work as before.

Tip: Set a breakpoint in OC_EventSource.send and watch the request in your browser.

image

## TODO

- [ ] ...

## Checklist


lib/public/IServerContainer.php
```diff
@@ -395,15 +395,6 @@ public function getSearch();
...
 	public function getCertificateManager();
 
-	/**
-	 * Create a new event source
-	 *
-	 * @return \OCP\IEventSource
-	 * @since 8.0.0
-	 * @deprecated 20.0.0 have it injected or fetch it through \Psr\Container\ContainerInterface::get
-	 */
-	public function createEventSource();
```
ChristophWurst, line 405: Fine to be removed because deprecated but needs docs

ChristophWurst (approved):

kesselb:
I overlooked that createEventSource would always create a new object. 
Updated the container definition and set $shared = false for the server. 

Can't say if a new instance makes sense, but that was the old behavior.

kesselb:
Example: https://github.com/Rello/audioplayer/pull/586

ChristophWurst (approved):

come-nc (commented):
In this instance I find it clearer to use createEventSource, as it makes it clear this is not a singleton. What are the pros of DI for non-singletons?

kesselb:
> In this instance I find it clearer to use createEventSource, as it makes it clear this is not a singleton. What are the pros of DI for non-singletons?

I find that a valid point. My motivation was to inject OC::$server->getRequest() via DI. 

@ChristophWurst you deprecated it a while ago. Are you okay to keep `createEventSource`?

Should we also add createEventSource to https://github.com/nextcloud/server/blob/master/lib/public/Server.php so apps can use it without using a private api?

icewind1991:
Might be worth adding an `EventSourceFactory` that can be injected. Same pattern as we use for http clients and others

ChristophWurst:
Yes let's go with a injectable and mockable factory instead of a static function

ChristophWurst (approved):
Clean

icewind1991 (approved):

PR #38393, opened 2023-05-22
bug-report-template - move `Official All-in-One appliance` to the bottom

szaimen:
Motivation: Most people seem to just use the top one which is incorrect in most cases since most reports come up on different systems than AIO.


artonge (commented):
Then people will default to use the new top one `"Community Web installer on a VPS or web space"`, no?
Or is AIO being mistaken? If so, we should maybe rename it to something like `Official All-in-One docker appliance`?

come-nc (approved):

skjnldsv (approved):

PR #38397, opened 2023-05-22
enh(contacts): show/hide addressbooks for all

### LINKED ISSUE #479: Add new option to hide an address book's contacts without disabling the address book
This issue results from dialog in [Enable/disable address books #412](https://github.com/nextcloud/contacts/pull/412).

The new option already implemented to disable an address book seems to do that on such a low level that it is then disabled per se - for all purposes. For this reason, it is not possible to disable an address book which has been shared with you rather than belonging to you.

In many use cases - team work, small to medium businesses, large contact archives used only seldom - we need a different function: to HIDE the contacts of a particular address book from the VIEW of the logged-in user (available also when not the owner). This is especially true as the historical design of the contacts app is to load all the visible contacts to the front end and then do all the nifty things like selection by category in JS. Which is great with up to (nowadays) about 1000 contacts, but becomes sluggish to unbearable the more you go over the top of that limit.

I would personally prefer that hidden contacts would still be searchable when I type in part of a name or email in the Contacts app search box. And I would certainly want that contacts hidden in the Contacts app (but not disabled) be active for the mail app, e.g. to make sure that mails from contacts known to me are not put into the spam folder.

Finally, in the settings area, if the little "dotdotdot" dropdown for each address book gets a further option "hide/show", then we should really try to explain the difference between Disable and Hide right there where the user has that choice - are tooltips possible there?

Some remote part of me is still dreaming about a backend-based variant of this app - termed by yours truly "Big Contacts" - which would allow working with many thousands of contacts, with the frontend fetching them piece by piece from the backend via traditional lazy loading / infinite scrolling, [see this issue](https://help.nextcloud.com/t/handling-a-large-number-1000-of-contacts-in-the-web-ui/22064).

Be that as it may, the abilty to hide an address book would be a great step on - so thanks John, thanks Henni, thanks Alexander!

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/55015861-add-new-option-to-hide-an-address-book-s-contacts-without-disabling-the-address-book?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F46751899&utm_medium=issues&utm_source=github).
</bountysource-plugin>

JohannesGGE:
* Fixes https://github.com/nextcloud/contacts/issues/479

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/CardDAV/AddressBook.php
+++ b/apps/dav/lib/CardDAV/AddressBook.php
@@ -46,7 +46,6 @@ use Sabre\DAV\PropPatch;
  * @property CardDavBackend $carddavBackend
  */
 class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMoveTarget {
-
 	/**
 	 * AddressBook constructor.
 	 *
@@ -116,7 +115,12 @@ class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMov
 				'privilege' => '{DAV:}write',
 				'principal' => $this->getOwner(),
 				'protected' => true,
-			]
+			],
+			[
+				'privilege' => '{DAV:}write-properties',
+				'principal' => '{DAV:}authenticated',
+				'protected' => true,
+			],
 		];
 
 		if ($this->getOwner() === 'principals/system/system') {
@@ -147,7 +151,7 @@ class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMov
 		}
 
 		$acl = $this->carddavBackend->applyShareAcl($this->getResourceId(), $acl);
-		$allowedPrincipals = [$this->getOwner(), parent::getOwner(), 'principals/system/system'];
+		$allowedPrincipals = [$this->getOwner(), parent::getOwner(), 'principals/system/system', '{DAV:}authenticated'];
 		return array_filter($acl, function ($rule) use ($allowedPrincipals) {
 			return \in_array($rule['principal'], $allowedPrincipals, true);
 		});
@@ -166,8 +170,7 @@ class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMov
 		return new Card($this->carddavBackend, $this->addressBookInfo, $obj);
 	}
 
-	public function getChildren()
-	{
+	public function getChildren() {
 		$objs = $this->carddavBackend->getCards($this->addressBookInfo['id']);
 		$children = [];
 		foreach ($objs as $obj) {
@@ -178,8 +181,7 @@ class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMov
 		return $children;
 	}
 
-	public function getMultipleChildren(array $paths)
-	{
+	public function getMultipleChildren(array $paths) {
 		$objs = $this->carddavBackend->getMultipleCards($this->addressBookInfo['id'], $paths);
 		$children = [];
 		foreach ($objs as $obj) {
@@ -221,10 +223,12 @@ class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMov
 	}
 
 	public function propPatch(PropPatch $propPatch) {
-		if (isset($this->addressBookInfo['{http://owncloud.org/ns}owner-principal'])) {
-			throw new Forbidden();
+		// shared address books will be handled by
+		// \OCA\DAV\DAV\CustomPropertiesBackend::propPatch
+		// to save values in db table instead of dav object
+		if (!$this->isShared()) {
+			parent::propPatch($propPatch);
 		}
-		parent::propPatch($propPatch);
 	}
 
 	public function getContactsGroups() {
--- a/apps/dav/lib/DAV/CustomPropertiesBackend.php
+++ b/apps/dav/lib/DAV/CustomPropertiesBackend.php
@@ -176,6 +176,21 @@ class CustomPropertiesBackend implements BackendInterface {
 			}
 		}
 
+		// substr of addressbooks/ => path is inside the CardDAV component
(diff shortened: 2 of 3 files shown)


JohannesGGE:
Due to the different implementation of the recently contacted addressbook, the call for changing the enabled property can't be executed, and an "addressbook is immutable" is thrown in `apps/contactsinteraction/lib/AddressBook.php::propPatch()`.

Other addressbooks call `apps/dav/lib/CardDAV/AddressBook.php::propPatch()`

As far as I can tell the only needed code is the DB update of `apps/dav/lib/DAV/CustomPropertiesBackend.php::updateProperties()` to have a working enable/disable for the recently contacted addressbook.

JohannesGGE:
Recently contacted address book will be tackled in another issue.

apps/dav/lib/CardDAV/AddressBook.php
```diff
@@ -221,9 +226,6 @@ public function delete() {
 	}
 
 	public function propPatch(PropPatch $propPatch) {
-		if (isset($this->addressBookInfo['{http://owncloud.org/ns}owner-principal'])) {
```
ChristophWurst, line 224: why is this dropped?

apps/dav/lib/CardDAV/AddressBook.php
```diff
@@ -116,7 +116,12 @@ public function getACL() {
 				'privilege' => '{DAV:}write',
 				'principal' => $this->getOwner(),
 				'protected' => true,
-			]
+			],
+            [
+                'privilege' => '{DAV:}write-properties',
```
ChristophWurst, line 121: so this means users are now allowed to update properties of all address books, right? are there any property changes we might not want? e.g. rename.

apps/dav/lib/CardDAV/AddressBook.php
JohannesGGE: owner-principal seems to be set if the address book is shared, system or recently contacted. Own address books don't have this prop.

apps/dav/lib/CardDAV/AddressBook.php
JohannesGGE: just tested it with renaming, and it turns out the change of the name is not stored as a user prop but changes the real name, so the owner sees the change as well. I'll have a look in the code again to check where the distinction between the props is made.

JohannesGGE: @tcitworld As far as I understand it, the propPatch() in CustomPropertiesBackend.php is just called if the prop is a user prop. So the distinction is done somewhere before

JohannesGGE:
The addressbook is now checked for sharing. If it is shared the props will be saved as userProps. So even if the addressbook is renamed by user that is not the owner (not possible in the contacts app yet), the owners addressbook name isn't changed. For the case the addressbook has a userProp name, this name is shown now.

apps/dav/lib/DAV/CustomPropertiesBackend.php
```diff
@@ -176,6 +176,21 @@ public function propFind($path, PropFind $propFind) {
...
+				'{DAV:}displayname',
+			];
+
+			foreach ($customPropertiesForShares as $customPropertyForShares) {
+				if (in_array($customPropertyForShares, $allRequestedProps)) {
+					$requestedProps[] = $customPropertyForShares;
+				}
+			}
+		}
+
```
tcitworld, line 193: **suggestion:** I think there's a way to factorize this block with the block above in a single function which takes calendars and addressbooks specifics as parameters.

apps/dav/lib/DAV/CustomPropertiesBackend.php
```diff
@@ -176,6 +176,21 @@ public function propFind($path, PropFind $propFind) {
 			}
 		}
 
+		// substr of addressbooks/ => path is inside the CardDAV component
+		// three '/' => this a addressbook (no addressbook-home nor contact object)
+		if (substr($path, 0, 13) === 'addressbooks/' && substr_count($path, '/') === 3) {
```
ChristophWurst, line 181: server is PHP8.0+ so you can use https://www.php.net/manual/en/function.str-starts-with.php

apps/dav/lib/DAV/CustomPropertiesBackend.php
```diff
@@ -176,6 +176,21 @@ public function propFind($path, PropFind $propFind) {
...
+			$allRequestedProps = $propFind->getRequestedProperties();
+			$customPropertiesForShares = [
+				'{DAV:}displayname',
+			];
+
+			foreach ($customPropertiesForShares as $customPropertyForShares) {
+				if (in_array($customPropertyForShares, $allRequestedProps)) {
```
ChristophWurst, line 188:
```suggestion
				if (in_array($customPropertyForShares, $allRequestedProps, true)) {
```

for strict comparison

miaulalala (approved):

kesselb:
nitpick: enh(contacts) should be feat(contacts) according to https://www.conventionalcommits.org/en/v1.0.0/

apps/dav/lib/CardDAV/AddressBook.php
kesselb: The change also surprised me. 

Apparently, this was added before the CustomPropertiesBackend: https://github.com/nextcloud/server/commit/95e218b00c9526bf3cba8ce719b09b1ea9d49ba9

If we throw forbidden here, the CustomPropertiesBackend::propPatch is not executed?

kesselb (approved):
Well done :tada:

PR #38398, opened 2023-05-22
Store encrypted OAuth2 client secrets

julien-nc:
This PR includes:
* A migration step that encrypts existing client secrets and changes the length of the `secret` column in `oc_oauth2_clients`
* 2 new methods (`OCA\OAuth2\Db\Client::getRawSecret` and `OCA\OAuth2\Db\Client::setRawSecret`) in the Client DB entity which are used in the `oauth2` app instead of `OCA\OAuth2\Db\Client::getSecret` and `OCA\OAuth2\Db\Client::setSecret`

Maybe there's a more straightforward way to implement this.
I gave up on trying to redefine `OCA\OAuth2\Db\Client::getSecret` and `OCA\OAuth2\Db\Client::setSecret` because `setSecret` is used to build the Client entity so the value we get from the DB is encrypted again by `setSecret` and then stored in the entity attribute :grin:.

Remaining doubts/questions:
* Is 256 chars long enough for an encrypted 64 chars string?
* Is it fine that this breaks any other apps directly using the `OCA\OAuth2\Db\Client::getSecret` and `OCA\OAuth2\Db\Client::setSecret` methods?


marcelklehr:
> Is it fine that this breaks any other apps directly using the OCA\OAuth2\Db\Client::getSecret and OCA\OAuth2\Db\Client::setSecret methods?

Quick search in nextcloud org only reveals: https://github.com/nextcloud/integration_openproject

julien-nc:
> Quick search in nextcloud org only reveals: https://github.com/nextcloud/integration_openproject

Yeah that's the one I found by searching for `getSecret` usages in my IDE.

ChristophWurst:
For the software architecture I would suggest to leave the db layer like before and do the encryption/description in the service layer (controller in the oauth app case). That avoids adding logic and a non-testable service locator to the entity class.

ChristophWurst:
> Is it fine that this breaks any other apps directly using the `OCA\OAuth2\Db\Client::getSecret` and `OCA\OAuth2\Db\Client::setSecret` methods?

OCA is a private app namespace. We can change it at any time. Other apps are not supposed to access other apps this way but go through OCP

apps/oauth2/lib/Db/Client.php
```diff
@@ -43,12 +44,18 @@ class Client extends Entity {
 	protected $clientIdentifier;
 	/** @var string */
 	protected $secret;
+	/**
+	 * @var ICrypto
+	 */
+	private mixed $crypto;
```
ChristophWurst, line 50: unused -> drop

apps/oauth2/lib/Migration/Version011601Date20230522143227.php
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options) {
+		$qb = $this->connection->getQueryBuilder();
+		$secrets = [];
+		$qb->select('id', 'secret')
+			->from('oauth2_clients');
+		$req = $qb->executeQuery();
+		while ($row = $req->fetch()) {
+			$secrets[$row['id']] = $row['secret'];
```
ChristophWurst, line 67: this can exhaust memory if there are lots of rows. you should be able to update rows in the while loop.

apps/oauth2/lib/Migration/Version011601Date20230522143227.php
```diff
@@ -0,0 +1,83 @@
...
+		}
+		$req->closeCursor();
+		$qb = $qb->resetQueryParts();
+
+		foreach ($secrets as $id => $secret) {
+			$encryptedSecret = $this->crypto->encrypt($secret);
+			$qb->update('oauth2_clients')
```
ChristophWurst, line 74: build the query once with a parameter and only set the parameter in the loop. building a query is expensive.

ChristophWurst (commented):
performance nitpicks

apps/oauth2/lib/Migration/Version011601Date20230522143227.php
```diff
@@ -0,0 +1,80 @@
...
+
+		$qbSelect = $this->connection->getQueryBuilder();
+		$qbSelect->select('id', 'secret')
+			->from('oauth2_clients');
+		$req = $qbSelect->executeQuery();
+		while ($row = $req->fetch()) {
+			$id = $row['id'];
+			$secret = $row['secret'];
+			$encryptedSecret = $this->crypto->encrypt($secret);
+			$qbUpdate->set('secret', $qbUpdate->createNamedParameter($encryptedSecret, IQueryBuilder::PARAM_STR))
```
ChristophWurst, line 72: I think this still rebuilds the query. Use two parameters like here https://github.com/nextcloud/mail/blob/8f3ea1f6d478de41aef9912877e76da598c02450/lib/Db/MessageMapper.php#L377-L417

apps/oauth2/lib/Migration/Version011601Date20230522143227.php
julien-nc: Done :+1:

apps/oauth2/lib/Db/Client.php
julien-nc: Yeah this is obviously not intentional :grin:

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -125,7 +125,8 @@ public function getToken($grant_type, $code, $refresh_token, $client_id, $client
 		}
 
 		// The client id and secret must match. Else we don't provide an access token!
-		if ($client->getClientIdentifier() !== $client_id || $client->getSecret() !== $client_secret) {
+		$storedClientSecret = $this->crypto->decrypt($client->getSecret());
```
nickvergessen, line 128: Catch exceptions of failing decrypt so it still returns:
```
			return new JSONResponse([
				'error' => 'invalid_client',
			], Http::STATUS_BAD_REQUEST);
```
And log an error?

julien-nc: :+1:

apps/oauth2/lib/Migration/Version011601Date20230522143227.php
```diff
@@ -0,0 +1,82 @@
...
+
+class Version011601Date20230522143227 extends SimpleMigrationStep {
+
+	public function __construct(
+		private IDBConnection $connection,
+		private ICrypto $crypto,
+	) {
+	}
+
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
```
ChristophWurst, line 44: info.xml bump missing to trigger the execution of this migration

apps/oauth2/lib/Settings/Admin.php
```diff
@@ -29,36 +29,39 @@
...
+				$secret = $this->crypto->decrypt($client->getSecret());
+				$result[] = [
+					'id' => $client->getId(),
+					'name' => $client->getName(),
+					'redirectUri' => $client->getRedirectUri(),
+					'clientId' => $client->getClientIdentifier(),
+					'clientSecret' => $secret,
+				];
+			} catch (\Exception $e) {
+				$this->logger->info('[Settings] OAuth client secret decryption error', ['exception' => $e]);
```
nickvergessen, line 63: Would go with error because the admin settings are empty now and then default log level should show the info why

```suggestion
				$this->logger->error('[Settings] OAuth client secret decryption error', ['exception' => $e]);
```

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -124,8 +107,16 @@ public function getToken($grant_type, $code, $refresh_token, $client_id, $client
 			$client_secret = $this->request->server['PHP_AUTH_PW'];
 		}
 
+		try {
+			$storedClientSecret = $this->crypto->decrypt($client->getSecret());
+		} catch (\Exception $e) {
+			$this->logger->info('OAuth client secret decryption error', ['exception' => $e]);
```
nickvergessen, line 113:
```suggestion
			$this->logger->error('OAuth client secret decryption error', ['exception' => $e]);
```

error or warning?

nickvergessen (approved):

rullzer (approved):

julien-nc:
Rebased and fixed the failing SettingsControllerTest.
CI failures are not related IMO.

szaimen:
@julien-nc 
I think these are related:
Screenshot_20230607_184436_Brave

julien-nc:
@szaimen Thanks, I'm on it.

apps/oauth2/lib/Migration/Version011601Date20230522143227.php
```diff
@@ -0,0 +1,82 @@
...
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('oauth2_clients')) {
+			$table = $schema->getTable('oauth2_clients');
+			if ($table->hasColumn('secret')) {
+				$column = $table->getColumn('secret');
+				$column->setLength(256);
```
ChristophWurst, line 52: let's widen to 512?

ChristophWurst:
My application secret is 48 chars long

image

PR #38429, opened 2023-05-24
Fix deletion of User when system link are used in him user folder

tanganellilore:
Signed-off-by: Lorenzo Tanganelli <lorenzo.tanganelli@hotmail.it>

* Resolves: # 

## Summary
In a localstorage with `localstorage.allowsymlinks` enabled, if we add a symlink in a folder under `<user_id>` data and then we try to delete this `<user_id>`, we receive a delete error and user is not deleted completly because these folder are recognized as a dir and not symlink.

This beacase in the `lib/private/Files/Storage/Local.php` at line 140 it's used the function `isDir()` that , as per php documentation, `->isDir()` return TRUE for symlinks of directories. Better use `getType()` method instead, which returns 'link' for symlinks.`

To easy fix it, without rewrite function with getType(), we can simply invert `elseif` in that function, checking dir after the check of `isFile` or `isLink`.

In that way we check if is file or link before and unlink it.

## TODO

N/A

## Checklist

### DIFF
--- a/lib/private/Files/Storage/Local.php
+++ b/lib/private/Files/Storage/Local.php
@@ -136,10 +136,10 @@ class Local extends \OC\Files\Storage\Common {
 				if (in_array($file->getBasename(), ['.', '..'])) {
 					$it->next();
 					continue;
-				} elseif ($file->isDir()) {
-					rmdir($file->getPathname());
 				} elseif ($file->isFile() || $file->isLink()) {
 					unlink($file->getPathname());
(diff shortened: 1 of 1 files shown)


solracsf (approved):
Makes sense 👍

nfebe (approved):
Looks good , thanks! Is there an issue for this? If so, please link it!

tanganellilore:
I haven't open it, just put a PR. Lett me know if issue is required.

PR #26182, opened 2021-03-17
Throw our own exception in \OCP\DB\QueryBuilder\IQueryBuilder::execute

ChristophWurst:
Just as documented. This seems to be a leftover form the dbal changes in
21. We noticed that `execute` still throws the raw dbal exception and
catches for the documented execption never trigger.

Follow-up to https://github.com/nextcloud/server/pull/24948

PR #38439, opened 2023-05-24
Wrap doctrine exceptions on IQueryBuilder::execute

juliusknorr:
While the method is deprecated it is still used in tons of places especially in server, and the annotation for throwing a `OCP\DB\Exception` was wrong for a long time. This PR aims to wrap the `\Doctrine\DBAL\Exception` that is actually thrown in a `OCP\DB\Exception` as it is already done for executeQuery and executeStatement.

Since this is bringing the implementation up to date with the alerady existing php doc comments I wouldn't consider this breaking but I could imagine all kind of issues showing up, therefore I opened as a backport fix for 27 and below https://github.com/nextcloud/server/pull/38440

Let's see what CI thinks about this

### DIFF
--- a/lib/private/DB/QueryBuilder/QueryBuilder.php
+++ b/lib/private/DB/QueryBuilder/QueryBuilder.php
@@ -277,7 +277,12 @@ class QueryBuilder implements IQueryBuilder {
 			]);
 		}
 
-		$result = $this->queryBuilder->execute();
+		try {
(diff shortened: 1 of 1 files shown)


ChristophWurst:
I think I did that but had to revert. This is tricky.

ChristophWurst:
https://github.com/nextcloud/server/pull/26182

PR #38440, opened 2023-05-24
[stable27] fix: Catch Deadlock properly as execute throws Doctrine exceptions not our wrapped ones

juliusknorr:
Followup fix to https://github.com/nextcloud/server/pull/37820 since the Cache update method uses IQueryBuilder::execute which does not wrap the Doctrine exception as the phpdoc annotation wrongly defines.

Fix only against stable27 as for master i think we should fix the wrapping instead: https://github.com/nextcloud/server/pull/38439

### DIFF
--- a/lib/private/Files/Cache/Updater.php
+++ b/lib/private/Files/Cache/Updater.php
@@ -27,7 +27,7 @@
  */
 namespace OC\Files\Cache;
 
-use OC\DB\Exceptions\DbalException;
+use Doctrine\DBAL\Exception\DeadlockException;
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

szaimen:
@juliushaertl shouldnt we also push this to master in case https://github.com/nextcloud/server/pull/38439 does not get merged until the next branch-off?

blizzz (approved):
OK with the outlook of a proper wrapping fix :)

PR #38445, opened 2023-05-24
Outsource CSRF validation

kesselb:
## Summary

To validate a CSRF token the Request object needs CsrfTokenManager. 

CsrfTokenManager is a heavy dependency.

```mermaid
flowchart TD
    CsrfTokenManager
    CsrfTokenManager-->CsrfTokenGenerator-->ISecureRandom
    CsrfTokenManager-->SessionStorage-->ISession-->IUserSession
    IUserSession-->OC\User\Session
    OC\User\Session-->OC\User\Manager
    OC\User\Session-->OCP\ISession
    OC\User\Session-->ISecureRandom
    OC\User\Session-->LoggerInterface
    OC\User\Session-->IEventDispatcher
    OC\User\Manager-->ICacheFactory-->ICache
```

## TODO

- [x] Merge #38386 

## Checklist


tcitworld (commented):
Would refactoring `OC\AppFramework\Http\Request::passesCSRFCheck` to use the new `CsrfValidator` create more dependency hell? It would avoid to have the logic in two different places.

kesselb:
> It would avoid to have the logic in two different places.

I agree, having the logic in two different places it not an improvement. 

The goal is to remove passesCSRFCheck and the CSRFTokenManager from IRequest. 
I deprecated it, but it's public api and has to stay for 3 years.

lib/private/Security/CSRF/CsrfValidator.php
```diff
@@ -0,0 +1,52 @@
...
+namespace OC\Security\CSRF;
+
+use OCP\IRequest;
+
+class CsrfValidator {
+	public function __construct(
+		private CsrfTokenManager $csrfTokenManager) {
+	}
+
+	public function validate(IRequest $request) {
```
ChristophWurst, line 35:
```suggestion
	public function validate(IRequest $request): bool {
```

nickvergessen:
But now all the classes need CsrfValidator and that still needs CsrfTokenManager?
Or did I miss the point?

kesselb:
> But now all the classes need CsrfValidator and that still needs CsrfTokenManager?

That's correct. 

IRequest.passesCSRFCheck is only needed in a couple of places, but IRequest is injected in many more classes. 
For Nextcloud 30, we can remove passesCSRFCheck and CsrfTokenManager from Request.

nickvergessen:
> For Nextcloud 30, we can remove passesCSRFCheck and CsrfTokenManager from Request.

3 years, not 3 versions xP so 36 😔

nickvergessen:
But okay, got it now and see how it improves.

Could also remove the dependency already and just depend on `\OC::$server->get()` for the time being? 😅

kesselb:
> 3 years, not 3 versions xP so 36 😔

:sob: 

> Could also remove the dependency already and just depend on \OC::$server->get() for the time being? 😅

Good idea :+1: 
I would like to do this as a follow-up.

apps/dav/lib/Server.php
```diff
@@ -120,7 +122,8 @@ public function __construct(IRequest $request, string $baseUri) {
 			\OC::$server->getUserSession(),
 			\OC::$server->getRequest(),
 			\OC::$server->getTwoFactorAuthManager(),
-			\OC::$server->getBruteForceThrottler()
+			\OC::$server->get(Throttler::class),
```
nickvergessen, line 125: Please use the OCP interface

lib/private/AppFramework/DependencyInjection/DIContainer.php
```diff
@@ -233,7 +233,8 @@ public function __construct(string $appName, array $urlParams = [], ServerContai
 					$c->get(IRequest::class),
 					$c->get(IControllerMethodReflector::class),
 					$c->get(IUserSession::class),
-					$c->get(OC\Security\Bruteforce\Throttler::class)
+					$c->get(OC\Security\Bruteforce\Throttler::class),
```
nickvergessen, line 236: OCP interface?

kesselb:
> OCP interface

Now that you mentioned it, CsrfValidator needs an interface as well for forms[^1] and user_saml[^2]. 

[^1]: https://github.com/nextcloud/forms/blob/ef5fe2095b764b0dbd32fe64aa7d013029850525/lib/Middleware/PublicCorsMiddleware.php#L75
[^2]: https://github.com/nextcloud/user_saml/blob/56f3fcd2718ef40800ffd96bf5d4df21ef12b831/lib/Controller/SAMLController.php#L476

lib/public/IRequest.php
```diff
@@ -201,6 +201,7 @@ public function getCookie(string $key);
 	 *
 	 * @return bool true if CSRF check passed
 	 * @since 6.0.0
+	 * @deprecated 28.0.0 use CsrfValidator.validate instead
```
ChristophWurst, line 204:
```suggestion
	 * @deprecated 28.0.0 use \OCP\Security\CSRF\ICsrfValidator::validate instead
```

lib/public/Security/CSRF/ICsrfValidator.php
```diff
@@ -0,0 +1,38 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+/**
+ * @since 28.0.0
+ */
+interface ICsrfValidator {
+	/**
+	 * @since 28.0.0
```
ChristophWurst, line 35: description?

lib/public/Security/CSRF/ICsrfValidator.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+namespace OCP\Security\CSRF;
```
ChristophWurst, line 5:
```suggestion
namespace OCP\Security\Csrf;
```

maybe

https://stackoverflow.com/questions/2236807/java-naming-convention-with-acronyms

ChristophWurst (approved):

lib/public/Security/CSRF/ICsrfValidator.php
kesselb: Using `Csrf` where possible, renaming the namespace with the existing classes is too risky :worried:

PR #38457, opened 2023-05-25
Use default page size for jumping to desired offset

### LINKED ISSUE #38417: Improve support for offset in ldap searches
When searching for users in LDAP backend and using an offset, the LDAP backend uses a paged search, and loops until the offset if needed.
For this loop it uses the limit asked in the request as page size, I think because some server may not handle correctly switching page size in the middle of a paged search (would need to confirm that).

Because of that, using a big offset with a small page size will result in a ton of small searches.
`occ ldap:search --limit=2 --offset=100 ""` will do 50 requests, while `occ ldap:search --limit=10 --offset=100 ""` will do 10, and return faster despite having more data to fetch.
This is triggered by the disabled user list, because it loops through all users by pages of 25, searching for disabled users.

A request to offset 1925 on my test setup triggered an XDebug error because it was nesting more than 250 function calls.

Ideas for solutions:
- Use maximum (default) page size always and cut results after
- Use maximum page size until desired offset and target page size for final call (but is that supported?)
- Make sure to reuse paged searches cookies when multiple calls are done after one another (will not fix all but will help at least the disabled user list thing)
- For disabled user list, we should have an API endpoint to get this list that does the computing on server side and avoid this looping through all users in frontend

come-nc:
* Resolves: #38417 

## Summary

Instead of using the final limit as page size, use the default one, which may be higher.
For instance `occ ldap:search --limit=2 --offset=100 ""` will now do 2 requests instead of 50.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Access.php
+++ b/apps/user_ldap/lib/Access.php
@@ -1994,8 +1994,14 @@ class Access extends LDAPUtility {
 				// to start from 0 to come to the desired page. cookie value
 				// of '0' is valid, because 389ds
 				$defaultPageSize = (int)$this->connection->ldapPagingSize;
-				$reOffset = ($offset - $defaultPageSize) < 0 ? 0 : $offset - $defaultPageSize;
-				$this->search($filter, $base, $attr, $defaultPageSize, $reOffset, true);
+				if ($offset < $defaultPageSize) {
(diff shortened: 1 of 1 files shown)


susnux (approved):
Code looks good to me

blizzz (approved):
makes sense :+1:

blizzz:
also tested now successfully against Samba4.

PR #38467, opened 2023-05-25
Add fallback when a there is no preview for a version

### LINKED ISSUE #37067: Versions tab shows broken image for empty files
When a version file is empty or the preview could not be generated it currently loads a 404 image which some browsers render as a broken preview:

Screenshot 2023-03-07 at 08 53 44

Best would be to have a fallback image for empty files or if the onerror handler for the image tag triggers.

cc @artonge since you worked on the vue rewrite if i remember correctly

artonge:
Fix https://github.com/nextcloud/server/issues/37067

Before (Chrome) | Before (Firefox) | After
-- | -- | --
Screenshot from 2023-05-25 15-43-15 | Screenshot from 2023-05-25 15-43-06 | Screenshot from 2023-05-25 15-42-08

@nextcloud/designers Unsure about the new icon. Do you prefer to go with the "After" version, or the "Firefox" one?
No icon could be confusing because the user might think that the file is empty, even though it might not.


apps/files_versions/src/components/Version.vue
```diff
@@ -281,6 +289,11 @@ export default {
 		height: 3rem;
 		border: 1px solid var(--color-border);
 		border-radius: var(--border-radius-large);
+
+		// Useful to display no preview icon.
+		display: flex;
+		justify-content: center;
+		color: var(--color-main-text);
```
szaimen, line 296:
```suggestion
		color: var(--color-text-light);
```

szaimen (approved):
One small suggestion, apart from that tested and works and LGTM!

nfebe (approved):

Pytal (approved):

nimishavijay (commented):
Good call! The "After" version definitely looks better as you said, and agreed with @szaimen that it can be a lighter color :)

solracsf:
Must be backported manually if needed, it seems 😞

PR #38468, opened 2023-05-25
log failures to read certificates during listing

icewind1991:
And check that we could actually read the file

### DIFF
--- a/lib/private/Security/CertificateManager.php
+++ b/lib/private/Security/CertificateManager.php
@@ -83,8 +83,14 @@ class CertificateManager implements ICertificateManager {
 		while (false !== ($file = readdir($handle))) {
 			if ($file != '.' && $file != '..') {
 				try {
-					$result[] = new Certificate($this->view->file_get_contents($path . $file), $file);
+					$content = $this->view->file_get_contents($path . $file);
+					if ($content !== false) {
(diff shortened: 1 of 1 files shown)


kesselb (approved):

lib/private/Security/CertificateManager.php
```diff
@@ -83,8 +83,14 @@ public function listCertificates(): array {
 		while (false !== ($file = readdir($handle))) {
 			if ($file != '.' && $file != '..') {
 				try {
-					$result[] = new Certificate($this->view->file_get_contents($path . $file), $file);
+					$content = $this->view->file_get_contents($path . $file);
+					if ($content) {
+						$result[] = new Certificate($content, $file);
+					} else {
+						$this->logger->error("Failed to read certificate from $path");
```
Fenn-CS, line 90: Is this really necessary? file contents can return a non-boolean value that evaluates to false, would this statement still be correct?

nfebe (approved):
I left an inline comment/question, otherwise it looks good. Thanks!

lib/private/Security/CertificateManager.php
icewind1991: Switched to a strict check

PR #36690, opened 2023-02-13
fix: Make sure that rollback hook is triggered on all version backends

juliusknorr:
## Summary

Alternative to https://github.com/nextcloud/groupfolders/pull/2260 to apply to any version backend.

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #38469, opened 2023-05-25
Fix initialisation of versions in the DB

artonge:
Broken after https://github.com/nextcloud/server/pull/36690

`$userFolder` was only set for `SharedStorage`
Also took the opportunity to fix some psalm warnings.


juliusknorr (approved):

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
```diff
@@ -66,17 +66,35 @@ public function useBackendForStorage(IStorage $storage): bool {
...
 				throw new NotFoundException("version file not found for share owner");
 			}
+		} else {
+			$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+		}
+
+		$fileId = $file->getId();
+		if ($fileId === null) {
+			throw new NotFoundException("File not found ($fileId)");
 		}
```
come-nc, line 98: Move before if/else and remove duplicate?

artonge: No possible as the check is needed at both places. `$file` is redeclared and used inside the `if`.

come-nc (approved):

come-nc (approved):

PR #38470, opened 2023-05-25
add command to migrate from owncloud wnd external storage

icewind1991:
- [x] update backend type
- [x] update storage id
- [x] reencode password
- [x] handle non-basic auth providers
	- [x] credentials are stored in `oc_external_config`, need to re-encrypte the password
		- [x] `password::password`
	- [x] credentials are stored in `oc_credentials`, need to migrate them to `oc_storages_credentials` and re-encrypte the password
		- [x] `password::logincredentials`
		- [x] `password::global`
	- [x] Stored in memory so nothing to do
		- [x] `password::sessioncredentials`
		- [x] `kerberos::kerberos`
	- [x] Not supported by Nextcloud
		- [x] `password::hardcodedconfigcredentials`


artonge:
I also pushed more changes here: https://github.com/nextcloud/server/pull/39593/commits

PR #38477, opened 2023-05-26
Use a11y activation util

Pytal:
## Summary

Use util from `@nextcloud/vue` and drop core util

## Checklist


core/src/OC/dialogs.js
```diff
@@ -49,7 +49,7 @@ import $ from 'jquery'
 
 import OC from './index.js'
 import OCA from '../OCA/index.js'
-import { isA11yActivation } from '../Util/a11y.js'
+import { isA11yActivation } from '@nextcloud/vue'
```
susnux, line 52: This will increase the dist bundle size by about 5MB.
Webpack is not good at tree shaking, so the best would be to use  `import ... from '@nextcloud/vue/dist/.../'` but the a11y is currently not exported like this.

BTW: The other option to use the named import schema in every core file, is even worse...

Before this | With this | Everything as named import
---|---|---
57872356 | 62524760 | 73743763

(bytes)

JuliaKirschenheuter: 5MB seems to be a lot

JuliaKirschenheuter (approved):
I'm voting for usage of `import ... from '@nextcloud/vue/dist/.../'` too

JuliaKirschenheuter (commented):
I'm voting for usage of import ... from '@nextcloud/vue/dist/.../' too

core/src/OC/dialogs.js
susnux: Yeah this already known (one of the reasons I am pushing ESM + vite).

But for now we need to add an entry point for this to the `@nextcloud/vue` library and stick with the explicit default import.

PR #38489, opened 2023-05-26
add commands to list shares and set share owner

icewind1991:
- Add command to list shares with filter option
- Add command to set share owner

Use case is for setups where multiple users have access to the same set of files (group folders/external storage/etc) and the admins wants to transfer the shares from a user that is being deleted to another user that still has access.

### DIFF
--- a/apps/files_sharing/lib/Command/ListCommand.php
+++ b/apps/files_sharing/lib/Command/ListCommand.php
@@ -1,0 +1,197 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2023 Robin Appelman <robin@icewind.nl>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Files_Sharing\Command;
+
+use OC\Core\Command\Base;
+use OCP\IUserManager;
+use OCP\Share\IManager;
+use OCP\Share\IShare;
+use Symfony\Component\Console\Helper\Table;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class ListCommand extends Base {
+	private IManager $shareManager;
+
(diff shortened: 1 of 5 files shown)


apps/files_sharing/lib/Command/ListCommand.php
```diff
@@ -0,0 +1,197 @@
...
+		}
+
+		$allShares = $this->shareManager->getAllShares();
+
+		$filteredShares = new \CallbackFilterIterator($allShares, function(IShare $share) use ($ownerInput, $shareType, $sharedByInput, $sharedWithInput, $fileInput) {
+			return $this->filterShare($share, $ownerInput, $sharedByInput, $sharedWithInput, $shareType, $fileInput);
+		});
+
+		$shareData = [];
+		foreach ($filteredShares as $share) {
```
come-nc, line 84: I think the easiest to avoid type problems with the iterable is to just call filterShare in an `if` there and `continue`.

apps/files_sharing/lib/Command/ListCommand.php
```diff
@@ -0,0 +1,197 @@
...
+		return 0;
+	}
+
+	private function filterShare(
+		IShare $share,
+		string $ownerInput = null,
+		string $sharedByInput = null,
+		string $sharedWithInput = null,
+		int $shareType = null,
+		int $fileInput = null
```
come-nc, line 116: Why the default values?
If it’s just to get nullable types prefix type with `?` instead.

apps/files_sharing/lib/Command/SetOwner.php
```diff
@@ -0,0 +1,127 @@
...
+		$targetId = $input->getArgument('new-owner');
+
+		$target = $this->userManager->get($targetId);
+		if (!$target) {
+			$output->writeln("<error>User $targetId not found</error>");
+			return 1;
+		}
+
+		$share = $this->shareManager->getShareById($shareId);
+		if (!$share) {
```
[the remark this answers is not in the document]
come-nc: Replace by try/catch or remove

apps/files_sharing/lib/Command/SetOwner.php
```diff
@@ -0,0 +1,127 @@
...
+			$output->writeln("<error>Share $shareId not found</error>");
+			return 1;
+		}
+
+		$sourceFile = $share->getNode();
+		$usersWithAccessToSource = $this->fileUtils->getFilesByUser($sourceFile);
+
+		$targetHasNonSharedAccess = false;
+		$targetHasSharedAccess = false;
+		$fileOrFolder = ($sourceFile instanceof File) ? "file" : "folder";;
```
come-nc, line 89:
```suggestion
		$fileOrFolder = ($sourceFile instanceof File) ? "file" : "folder";
```

apps/files_sharing/lib/Command/SetOwner.php
```diff
@@ -0,0 +1,127 @@
...
+				$targetHasNonSharedAccess = true;
+			}
+		}
+
+		if (!$targetHasSharedAccess && !$targetHasNonSharedAccess) {
+			$output->writeln("<error>$targetId has no access to the $fileOrFolder $sourceName shared by $shareId</error>");
+			return 1;
+		}
+
+		if (!$targetHasNonSharedAccess && $targetHasSharedAccess) {
```
come-nc, line 113:
```suggestion
		if (!$targetHasNonSharedAccess) {
```

PR #38490, opened 2023-05-26
emit an event when a message is logged

icewind1991:
Allow apps to do their own handling, notably this is emitted even if it would be normally filtered by log level

Needed for https://github.com/nextcloud/logreader/pull/873

### DIFF
--- a/lib/private/EventDispatcher/EventDispatcher.php
+++ b/lib/private/EventDispatcher/EventDispatcher.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
  */
 namespace OC\EventDispatcher;
 
+use OC\Log;
 use Psr\Log\LoggerInterface;
 use function get_class;
 use OC\Broadcast\Events\BroadcastEvent;
@@ -54,6 +55,12 @@ class EventDispatcher implements IEventDispatcher {
 		$this->dispatcher = $dispatcher;
 		$this->container = $container;
 		$this->logger = $logger;
+
+		// inject the event dispatcher into the logger
+		// this is done here because there is a cyclic dependency between the event dispatcher and logger
+		if ($this->logger instanceof Log or $this->logger instanceof Log\PsrLoggerAdapter) {
+			$this->logger->setEventDispatcher($this);
+		}
 	}
 
 	public function addListener(string $eventName,
--- a/lib/public/Log/BeforeMessageLoggedEvent.php
+++ b/lib/public/Log/BeforeMessageLoggedEvent.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+/**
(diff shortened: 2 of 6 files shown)


lib/private/EventDispatcher/EventDispatcher.php
```diff
@@ -54,6 +55,12 @@ public function __construct(SymfonyDispatcher $dispatcher,
 		$this->dispatcher = $dispatcher;
 		$this->container = $container;
 		$this->logger = $logger;
+
+		// inject the event dispatcher into the logger
+		// this is done here because there is a cyclic dependency between the event dispatcher and logger
+		if ($this->logger instanceof Log or $this->logger instanceof Log\PsrLoggerAdapter) {
```
come-nc, line 61:
```suggestion
		if ($this->logger instanceof LoggerInterface) {
```
This is not enough?

lib/private/Files/Cache/Scanner.php
```diff
@@ -138,6 +138,7 @@ protected function getData($path) {
 	 * @throws \OCP\Lock\LockedException
 	 */
 	public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData = null, $lock = true, $data = null) {
+		\OC::$server->getLogger()->logException(new \Exception("scanning $file"), ['app' => 'files', 'level' => 2]);
```
come-nc, line 141: Both getLogger and logException are deprecated.

come-nc: Also this is not related to this PR, forgotten debug?

lib/public/Log/BeforeMessageLoggedEvent.php
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+
+	/**
+	 * Get the message of the log item
+	 *
+	 * @return string
+	 * @since 28.0.0
+	 */
+	public function getMessage() {
```
come-nc, line 75:
```suggestion
	public function getMessage(): string {
```

come-nc (commented):
Performance-wise, is that not too much overhead?

lib/private/Files/Cache/Scanner.php
icewind1991: :see_no_evil:

lib/private/EventDispatcher/EventDispatcher.php
icewind1991: `setEventDispatcher` is not part of the interface, only the 2 implementations.

icewind1991:
> Performance-wise, is that not too much overhead?

From my understanding the event dispatcher is well optimized for the "no listeners" case.

come-nc (approved):

nfebe (approved):

ChristophWurst:
New API -> docs

PR #38506, opened 2023-05-30
Add uploadsdirectory param

### LINKED ISSUE #38505: Add support to custom user uploads dir
Hi Team, I notice that actually is not possible customize the path where temporary uploads ,`<user_id>/uploads`, are stored. This happen for example with Chunk file.

It's possible customize the temp path for PHP and so on, but ufter tmp path, chunks go into `<user_id>/uploads` and then, after assebling, will go to the standar data folder for user.

It will be cool, have a settings that will allow us to "link" `<user_id>/uploads` to specific path, where we can mount, for example, local storage, that in general is much faster than long term storage.

Add a option to customize `<user_id>/uploads` or adding a symlink to it.

Actually there is no alternative apart disable chunking.

**Additional context**
I will send you a PR that will address this feature with usage of symlink and adding of new config parameter called `uploadsdirecotry`.
More info in the PR.

Thanks

tanganellilore:
* Resolves: #38505
* 
## Summary
This PR will address the feature above, adding new config parameter and manage it.

The "smart" solution adopted is to use the conjunction of param that already exists `localstorage.allowsymlinks` and new config param `uploadsdirectory`.
The idea is to use a symlink to avoid the rework in all files/functions where `<user_id>/uploads` it's used. 

The only piece of code that require changes is `apps/dav/lib/Upload/UploadHome.php` that check if `localstorage.allowsymlinks` is true and `uploadsdirectory` is setted, if yes we create symlink (if not already exists), if not, we continue to use the current implementation.

In case of user changes `localstorage.allowsymlinks` or `uploadsdirectory`, we will delete current link (or directory) and recreate it (link or dir based on settings).

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/Upload/UploadHome.php
+++ b/apps/dav/lib/Upload/UploadHome.php
@@ -105,7 +105,7 @@ class UploadHome implements ICollection {
 		$user_path = '/' . $user->getUID() . '/uploads';
         $absoluteUserPath = $rootView->getLocalFile($user_path);
 
-		if ($allowSymlinks && $uploadsDirectory != '' && $absoluteUserPath) {
+		if ($allowSymlinks && $uploadsDirectory !== '' && $absoluteUserPath) {
 			$upload_user_path = $uploadsDirectory . $user_path;
 
 			if (!$rootView->file_exists($user_path) || !is_link($absoluteUserPath) || ($upload_user_path != realpath($absoluteUserPath)) ) {
--- a/apps/dav/lib/Upload/UploadHome.php
+++ b/apps/dav/lib/Upload/UploadHome.php
@@ -98,35 +98,35 @@ class UploadHome implements ICollection {
 		$user = \OC::$server->getUserSession()->getUser();
 		Filesystem::initMountPoints($user->getUID());
 
-		$config = \OC::$server->getConfig();
+		$config = \OC::$server->get(IConfig::class);
 
 		$allowSymlinks = $config->getSystemValueBool('localstorage.allowsymlinks', false);
-		$uploadsDirectory = $config->getSystemValue('uploadsdirectory', '');
+		$uploadsDirectory = $config->getSystemValueString('uploadsdirectory', '');
 		$user_path = '/' . $user->getUID() . '/uploads';
-        $absolute_user_path = $rootView->getLocalFile($user_path);
+        $absoluteUserPath = $rootView->getLocalFile($user_path);
 
-		if ($allowSymlinks && $uploadsDirectory != '' && $absolute_user_path) {
+		if ($allowSymlinks && $uploadsDirectory != '' && $absoluteUserPath) {
 			$upload_user_path = $uploadsDirectory . $user_path;
 
-			if (!$rootView->file_exists($user_path) || !is_link($absolute_user_path) || ($upload_user_path != realpath($absolute_user_path)) ) {
+			if (!$rootView->file_exists($user_path) || !is_link($absoluteUserPath) || ($upload_user_path != realpath($absoluteUserPath)) ) {
 
                 if (!is_dir($upload_user_path)) {
                     mkdir($upload_user_path, 0750, true);
                 }
 
 				// useful if link is broken due to $upload_user_path changes
-                if (is_link($absolute_user_path)) {
-                    unlink($absolute_user_path);
-                } else if (is_dir($absolute_user_path)) {
+                if (is_link($absoluteUserPath)) {
+                    unlink($absoluteUserPath);
+                } elseif (is_dir($absoluteUserPath)) {
                     $rootView->rmdir($user_path);
                 }
-                symlink($upload_user_path, $absolute_user_path);
+                symlink($upload_user_path, $absoluteUserPath);
 
 			}
 
 		} else {
-            if ($absolute_user_path && is_link($absolute_user_path)) {
-                unlink($absolute_user_path);
+            if ($absoluteUserPath && is_link($absoluteUserPath)) {
+                unlink($absoluteUserPath);
             }
 			if (!$rootView->file_exists($user_path)) {
 				$rootView->mkdir($user_path);
--- a/apps/dav/lib/Upload/UploadHome.php
+++ b/apps/dav/lib/Upload/UploadHome.php
@@ -28,6 +28,7 @@ namespace OCA\DAV\Upload;
 use OC\Files\Filesystem;
 use OC\Files\View;
 use OCA\DAV\Connector\Sabre\Directory;
+use OCP\IConfig;
 use Sabre\DAV\Exception\Forbidden;
 use Sabre\DAV\ICollection;
 
--- a/apps/dav/lib/Upload/UploadHome.php
+++ b/apps/dav/lib/Upload/UploadHome.php
@@ -105,10 +105,10 @@ class UploadHome implements ICollection {
 		$user_path = '/' . $user->getUID() . '/uploads';
         $absolute_user_path = $rootView->getLocalFile($user_path);
 
-		if ($allowSymlinks && $uploadsDirectory != '') {
+		if ($allowSymlinks && $uploadsDirectory != '' && $absolute_user_path) {
 			$upload_user_path = $uploadsDirectory . $user_path;
 
-			if (!$rootView->file_exists($user_path) || !is_link($link_user_path) || ($upload_user_path != readlink($absolute_user_path)) ) {
+			if (!$rootView->file_exists($user_path) || !is_link($absolute_user_path) || ($upload_user_path != realpath($absolute_user_path)) ) {
 
                 if (!is_dir($upload_user_path)) {
                     mkdir($upload_user_path, 0750, true);
@@ -125,7 +125,7 @@ class UploadHome implements ICollection {
 			}
 
 		} else {
-            if (is_link($absolute_user_path)) {
+            if ($absolute_user_path && is_link($absolute_user_path)) {
                 unlink($absolute_user_path);
             }
 			if (!$rootView->file_exists($user_path)) {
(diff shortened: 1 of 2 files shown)


apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +97,43 @@ private function getView() {
 		$rootView = new View();
 		$user = \OC::$server->getUserSession()->getUser();
 		Filesystem::initMountPoints($user->getUID());
-		if (!$rootView->file_exists('/' . $user->getUID() . '/uploads')) {
-			$rootView->mkdir('/' . $user->getUID() . '/uploads');
+
+		$config = \OC::$server->getConfig();
+
+		$allowSymlinks = $config->getSystemValueBool('localstorage.allowsymlinks', false);
+		$uploadsDirectory = $config->getSystemValue('uploadsdirectory', '');
```
come-nc, line 104:
```suggestion
		$uploadsDirectory = $config->getSystemValueString('uploadsdirectory', '');
```

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +97,43 @@ private function getView() {
...
+		$user_path = '/' . $user->getUID() . '/uploads';
+        $absolute_user_path = $rootView->getLocalFile($user_path);
```
come-nc, line 106:
```suggestion
        $absoluteUserPath = $rootView->getLocalFile($user_path);
```
camelCase please.

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +97,43 @@ private function getView() {
...
+
+		if ($allowSymlinks && $uploadsDirectory != '') {
```
come-nc, line 108:
```suggestion
		if ($allowSymlinks && $uploadsDirectory !== '') {
```

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +97,43 @@
...
+			$upload_user_path = $uploadsDirectory . $user_path;
+
+			if (!$rootView->file_exists($user_path) || !is_link($link_user_path) || ($upload_user_path != readlink($absolute_user_path)) ) {
```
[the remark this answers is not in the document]
come-nc: I agree with psalm here, $link_user_psalm is not defined.

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +97,43 @@ private function getView() {
...
+
+                if (!is_dir($upload_user_path)) {
+                    mkdir($upload_user_path, 0750, true);
+                }
+
+				// useful if link is broken due to $upload_user_path changes
+                if (is_link($absolute_user_path)) {
+                    unlink($absolute_user_path);
+                } else if (is_dir($absolute_user_path)) {
```
come-nc, line 120:
```suggestion
                } elseif (is_dir($absolute_user_path)) {
```
Same for other else ifs

come-nc (commented):
Comments on code

config/config.sample.php
```diff
@@ -1897,6 +1897,14 @@
  */
 'updatedirectory' => '',
 
+/**
+ * Override where Nextcloud stores uploads user files while uploading (chunks). Useful in situations
+ * where the default `<user_id>/uploads` is on network disk like NFS.
+ * Defaults to the value of `` if unset. Require `localstorage.allowsymlinks` set to `true` because
+ * impementation require symlink to works.
+ */
+'uploadsdirectory' => '',
```
solracsf, line 1906: Maybe renaming to `uploads_path` instead to be similar to `cache_path` ?
https://github.com/nextcloud/server/blob/1e1e2fec64aa7dc5e8656bbee9c09494402420af/config/config.sample.php#L1527-L1535

tanganellilore: I'm agree, i followed the other keys like `updatedirectory`.
But i prefer uploads_path .

let me know if i can change it

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +98,43 @@ private function getView() {
...
 		$user = \OC::$server->getUserSession()->getUser();
 		Filesystem::initMountPoints($user->getUID());
-		if (!$rootView->file_exists('/' . $user->getUID() . '/uploads')) {
-			$rootView->mkdir('/' . $user->getUID() . '/uploads');
+
+		$config = \OC::$server->get(IConfig::class);
+
+		$allowSymlinks = $config->getSystemValueBool('localstorage.allowsymlinks', false);
+		$uploadsDirectory = $config->getSystemValueString('uploadsdirectory', '');
+		$user_path = '/' . $user->getUID() . '/uploads';
```
come-nc, line 106:
```suggestion
		$userPath = '/' . $user->getUID() . '/uploads';
```
camelCase

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +98,43 @@ private function getView() {
...
+        $absoluteUserPath = $rootView->getLocalFile($user_path);
+
+		if ($allowSymlinks && $uploadsDirectory !== '' && $absoluteUserPath) {
+			$upload_user_path = $uploadsDirectory . $user_path;
```
come-nc, line 110:
```suggestion
			$uploadUserPath = $uploadsDirectory . $user_path;
```

config/config.sample.php
```diff
@@ -1897,6 +1897,14 @@
  */
 'updatedirectory' => '',
 
+/**
+ * Override where Nextcloud stores uploads user files while uploading (chunks). Useful in situations
```
come-nc, line 1901:
```suggestion
 * Override where Nextcloud stores uploaded user files while uploading (chunks). Useful in situations
```

config/config.sample.php
```diff
@@ -1897,6 +1897,14 @@
...
+ * where the default `<user_id>/uploads` is on network disk like NFS.
+ * Defaults to the value of `` if unset. Require `localstorage.allowsymlinks` set to `true` because
```
come-nc, line 1903:
```suggestion
 * Defaults to the value of '' if unset. Require `localstorage.allowsymlinks` set to `true` because
```

config/config.sample.php
come-nc: No strong opinion on the name, use whatever seems best to you.

come-nc (commented):
Small comments on codestyle.
Codewise this looks good to me. I’d just like other opinion on the chosen solution to rely on symlink for this, does anyone see this causing trouble in the future?

config/config.sample.php
tanganellilore: I will use `uploads_path`

tanganellilore:
Thnaks @come-nc ,

the other alternative, is to rework where chunk uploads are store, without usage of symlink, but I don't have a lot fo experience on nextcloud codebase to identfy where uploads folder is used (i suppose on ChunkPlugin, but i don't know if somewhere else)

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +98,43 @@ private function getView() {
...
+
+		$allowSymlinks = $config->getSystemValueBool('localstorage.allowsymlinks', false);
+		$uploadsPath = $config->getSystemValueString('uploads_path', '');
+		$userPath = '/' . $user->getUID() . '/uploads';
+        $absoluteUserPath = $rootView->getLocalFile($userPath);
+
+		if ($allowSymlinks && $uploadsPath !== '' && $absoluteUserPath) {
+			$uploadUserPath = $uploadsPath . $userPath;
+
+			if (!$rootView->file_exists($userPath) || !is_link($absoluteUserPath) || ($uploadUserPath != realpath($absoluteUserPath)) ) {
```
artonge, line 112: We do we check for `!$rootView->file_exists($userPath)`?

artonge: I would invert the if/else statement to have a more readable condition.

```suggestion
			if ($rootView->file_exists($userPath) && is_link($absoluteUserPath) && ($uploadUserPath === realpath($absoluteUserPath)) ) {
```

apps/dav/lib/Upload/UploadHome.php
```diff
@@ -97,10 +98,43 @@ private function getView() {
...
+                    unlink($absoluteUserPath);
+                } elseif (is_dir($absoluteUserPath)) {
+                    $rootView->rmdir($userPath);
+                }
+                symlink($uploadUserPath, $absoluteUserPath);
+
+			}
+
+		} else {
+            if ($absoluteUserPath && is_link($absoluteUserPath)) {
```
artonge, line 129: More explicit check?
```suggestion
            if ($absoluteUserPath !== null && is_link($absoluteUserPath)) {
```

artonge (commented):
Not 100% sure about the current logic, I left a few comments to make is more readable.

Also, did you consider changing the code to directly write files to the PHP tmp folder? Probably a bigger change, but it would remove the need of creating and maintaining symlinks. 
Or we could also append each successive chunks to the same final file, which would also reduce the read/write to 1. But chunks are probably coming unordered.

tanganellilore:
As already commented, we can think to use directly a path like tmp for php, but i don't know which function need to be rewrite, in each piece/application.

If everything is under this repo, i can check it, but i don't know all dependecies

tanganellilore:
@come-nc  and @artonge I proposed new PR that should be address removal of symlink.

Let me know if I need to fix this one or we move to the "new" approach.

Thanks

icewind1991:
The use case of having the uploads go somewhere else is already supported by setting the `cache_path` config value:

https://github.com/nextcloud/server/blob/426c0341ffff262f58d1b7f031de4f0c53c8bec5/lib/private/Files/Mount/CacheMountProvider.php#L54-L70

tanganellilore:
Thanks @icewind1991, Yes i notice it when i work in the new PR, but this means that one config cover two different path.
I think to maintain 2 path separated with 2 different config.

What so you think? If yes we can consider only my last PR and close this one.

tanganellilore:
Close in favour of https://github.com/nextcloud/server/pull/38587

PR #38528, opened 2023-05-30
Prevent keyboard enter on sharing dialog

solracsf:
* Resolves: https://github.com/nextcloud/server/issues/38527

## Summary

Motivation: https://github.com/nextcloud/server/issues/38527

## Checklist

### DIFF
--- a/apps/files_sharing/src/components/SharingInput.vue
+++ b/apps/files_sharing/src/components/SharingInput.vue
@@ -34,6 +34,7 @@
 			:user-select="true"
 			:options="options"
 			v-model="value"
+			@keydown.enter.prevent
 			@open="handleOpen"
 			@search="asyncFind"
 			@option:selected="addShare">


Pytal:
How can keyboard users share files then?

solracsf:
@Pytal i must admit i didn't think about it 🤔 
Is there another option here, as this is a problem? Like disallow the enter key only before anything is selected?

Pytal:
Many popups, dialogs, and other inputs function the same way by focusing the first element on open which would also be selected if enter was hit accidently

Could you elaborate on what parts of the UX make this element more susceptible to accidental enters @solracsf?

solracsf:
As I've stated at https://github.com/nextcloud/server/issues/38527 say you type "a" and hit Enter (many users think this work like a search box where you hit Enter to start searching). The first action (or user) seleted by that "a" will be performed.

This is the main reason of this PR.

Pytal:
Should we change the UX to not auto-focus the first option in search contexts @nimishavijay?

image
image

szaimen:
This is a tough issue. I can understand both sides. If the first result does not get pre-selected, we would somehow need a way to make it discoverable that the arrows can be used to select the options. WDYT @nextcloud/designers ? :)

Pytal:
So with https://github.com/nextcloud/nextcloud-vue/pull/4209 the dropdown will stay open but the actual results are dependent on network and server latency

For highlighting the first option more prominently I would not opt for a border as we typically reserve that for explicit keyboard focus, what could we do here @nimishavijay?

nimishavijay:
> For highlighting the first option more prominently I would not opt for a border 

That sounds good to me! With these 2 fixes it should be improved already while maintaining consistency in interactions. @solracsf would that work for you? We can see if this improves the experience for a few weeks and then take further steps if required, what do you think? :)

solracsf:
It's a good first step. Fine by me.
@Pytal feel free to amend this PR, or close and styart a new one if needed. 👍

PR #38554, opened 2023-05-31
Refactors ChangePasswordController to improve code readability.

fsamapoor:
I'm always having a hard time reading nested code blocks.
I have refactored "changeUserPassword" method by using early returns and reducing indentations to improve code readability.

### DIFF
--- a/apps/settings/lib/Controller/ChangePasswordController.php
+++ b/apps/settings/lib/Controller/ChangePasswordController.php
@@ -188,6 +188,13 @@ class ChangePasswordController extends Controller {
 					],
 				]);
 			}
+
+			return new JSONResponse([
+				'status' => 'success',
+				'data' => [
+					'username' => $username,
+				],
+			]);
 		}
 
 		//handle the recovery case
--- a/apps/settings/lib/Controller/ChangePasswordController.php
+++ b/apps/settings/lib/Controller/ChangePasswordController.php
@@ -169,90 +169,7 @@ class ChangePasswordController extends Controller {
 			]);
 		}
 
-		if ($this->appManager->isEnabledForUser('encryption')) {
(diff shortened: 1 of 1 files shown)


fsamapoor:
@szaimen Sorry to bother, but it's been three days and I think this PR has been skipped and not being labeled.

apps/settings/lib/Controller/ChangePasswordController.php
```diff
@@ -169,90 +169,7 @@ public function changeUserPassword(string $username = null, string $password = n
...
-						'data' => [
-							'message' => $this->l->t('Unable to change password'),
-						]
-					]);
-				}
-			}
-		} else {
+		if (!$this->appManager->isEnabledForUser('encryption')) {
 			try {
 				if ($targetUser->setPassword($password) === false) {
```
come-nc, line 174: But if this does not return false it will try to build encryption classes and it will crash, no?

fsamapoor: Yes, it seems like I haven't done it completely right. Let me revise this and get back to you again.

fsamapoor: Sorry about the delay. I don't think the method is "that much" easier to read, but I thought I'd leave the decision up to you. If you feel like it is not worth it, we can close the PR. No problem at all.

Or maybe we can go a few steps further and do some actual cleanup to improve the readability of the class. If this was my code, I'd probably break that 170-line method into smaller pieces. What do you think?

come-nc: Yeah quite frankly I prefer the original with the if/else to having twice the return success.

PR #38556, opened 2023-05-31
chore(deps): Bump doctrine/dbal to 3.7.x

nickvergessen:
## TODO

- [x] https://github.com/nextcloud/3rdparty/pull/1424

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required https://github.com/nextcloud/documentation/pull/10712


apps/files_sharing/tests/Command/CleanupRemoteStoragesTest.php
```diff
@@ -79,11 +79,11 @@ protected function setUp(): void {
 		$shareExternalQuery->insert('share_external')
 			->setValue('share_token', '?')
 			->setValue('remote', '?')
-			->setValue('name', '?')->setParameter(2, 'irrelevant')
```
nickvergessen, line 82: This test failed before. It's a bit strange, but basically positional parameters get confused when executed multiple times.
Simply assigning them **all** again in the loop makes it work properly.

nickvergessen: I searched for `->setParameter(0` within our set of shipped apps and it seems to be the only occurrence.
Should still document in case an app uses it.

nickvergessen:
Forgot the schema migration thingy. Back to draft.

nickvergessen:
- [x] OC\Repair\SqliteAutoincrement
    - I installed current master and exported the schema as SQL
    - Then I went to this branch, removed the content of the run() method (so made it no-op)
    - I installed again and exported the schema as SQL
    - The files are exactly the same, so whatever we tried to fix was fixed since 2015 in doctrine dbal :-O
    -> Removed

lib/private/DB/Migrator.php
```diff
@@ -95,18 +94,18 @@ public function createSchema() {
...
 	 * @return SchemaDiff
 	 */
 	protected function getDiff(Schema $targetSchema, Connection $connection) {
-		// adjust varchar columns with a length higher then getVarcharMaxLength to clob
+		// adjust varchar columns with a length higher than getVarcharMaxLength to clob
 		foreach ($targetSchema->getTables() as $table) {
 			foreach ($table->getColumns() as $column) {
 				if ($column->getType() instanceof StringType) {
-					if ($column->getLength() > $connection->getDatabasePlatform()->getVarcharMaxLength()) {
+					if ($column->getLength() > 4000) {
```
come-nc, line 108: :monocle_face: No more property/function for this?
Should we add a constant on our side?

lib/private/DB/Migrator.php
```diff
@@ -95,18 +94,18 @@ public function createSchema() {
...
-		return $this->connection->getSchemaManager()->createSchema();
+		return $this->connection->createSchemaManager()->introspectSchema();
 	}
 
 	/**
 	 * @return SchemaDiff
 	 */
 	protected function getDiff(Schema $targetSchema, Connection $connection) {
-		// adjust varchar columns with a length higher then getVarcharMaxLength to clob
+		// adjust varchar columns with a length higher than getVarcharMaxLength to clob
```
come-nc, line 104: This comment also references getVarcharMaxLength

lib/private/DB/OracleMigrator.php
```diff
@@ -29,176 +32,114 @@
...
-			$tableDiff->removedIndexes = array_map(function (Index $index) {
-				return $this->quoteIndex($index);
-			}, $tableDiff->removedIndexes);
+			foreach ($table->getIndexes() as $index) {
+				if ($index->isPrimary()) {
+					$quotedTable->setPrimaryKey(
+						array_map(function ($columnName) {
+							return $this->connection->quoteIdentifier($columnName);
+						}, $index->getColumns()),
+						//TODO migrate existing uppercase indexes, then $this->connection->quoteIdentifier($index->getName()),
```
come-nc, line 76: Is this TODO blocking?

lib/private/DB/Migrator.php
Altahrim, line 108: Why 4000?
Need to adjust the comment above too

Altahrim (approved):

lib/private/DB/OracleMigrator.php
nickvergessen: No, it's actually from the old code. Just was moved here.

nickvergessen: Resolving it is actually breaking, so not sure we would ever resolve it. But it's important to remember when the migration code needs adjusting, so I kept it

lib/private/DB/Migrator.php
nickvergessen: yeah https://github.com/doctrine/dbal/blob/3.7.x/src/Platforms/AbstractPlatform.php#L666-L682

They actually dropped the functionality. They don't care about the final type being different in different databases. But we prefer consistency across all DBs and therefore need to continue changing the type. Otherwise new installs would be STRING on some DBs and installations, while being TEXT on others, and that would be way to confusing later on and when giving support.

nickvergessen: Adjusted the comment. Can you check again?

lib/private/DB/Migrator.php
```diff
@@ -178,7 +178,7 @@ protected function convertStatementToScript($statement) {
 	}
 
 	protected function getFilterExpression() {
-		return '/^' . preg_quote($this->config->getSystemValueString('dbtableprefix', 'oc_')) . '/';
+		return '/^' . preg_quote($this->config->getSystemValueString('dbtableprefix', 'oc_'), '/') . '/';
```
blizzz, line 181: :shrug:

blizzz (approved):

PR #38561, opened 2023-05-31
feat(request): Allow to match the client version with the IRequest::U…

nickvergessen:
…SER_AGENT_* regex

## Summary

* The idea is to allow APIs to "block" clients based on their versions, similar to:
https://github.com/nextcloud/server/blob/c63ff42410107570eef49cd0deea75821c5aeae1/apps/dav/lib/Connector/Sabre/BlockLegacyClientPlugin.php#L69-L71

    - [x] Need to clarify when the desktop client started to use a correct user agent and whether we can switch to that const there as well. If that is still a thing we could extend the regex to `/^(?:Mozilla\/5\.0 \([A-Za-z ]+\) )?(?:mirall|csyncoC)\/([^ ]*).*$/` cc @tobiasKaminsky @mgallien 

* Similar plan is developed in Talk and we would like to avoid duplicating the constants:
https://github.com/nextcloud/spreed/pull/9632

## Checklist


mgallien:
Starting with version 1.3.0 the user agent was following the pattern `/^(Mozilla\/5\.0 \([A-Za-z ]+\) )(mirall)\/([^ ]*).*$/`

nickvergessen:
That 1.3.0 version is from 2013, so that is save enough. Thanks for checking!

ChristophWurst (approved):
Nice

nfebe (approved):
LGTM :)

PR #38562, opened 2023-05-31
set a default and max ttl for redis keys

icewind1991:
having infinite TTL can lead to leaked keys as the prefix changes with version upgrades

### DIFF
--- a/lib/private/Memcache/Redis.php
+++ b/lib/private/Memcache/Redis.php
@@ -52,6 +52,9 @@ class Redis extends Cache implements IMemcacheTTL {
 		],
 	];
 
+	private const DEFAULT_TTL = 24 * 60 * 60; // 1 day
+	private const MAX_TTL = 30 * 24 * 60 * 60; // 1 month
+
 	/**
 	 * @var \Redis|\RedisCluster $cache
 	 */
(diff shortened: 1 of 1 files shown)


blizzz (approved):
Should be mentioned in developer documentation as well.

ChristophWurst:
``\OCP\ICache::set`` says `int $ttl Time To Live in seconds. Defaults to 60*60*24`

ChristophWurst:
I think it would be reasonable to also define an upper bound for TTL. Like a cache key can only exist for three months.

icewind1991:
Switched the default to 1 day as per the interface docs and added a max of 1 month

lib/private/Memcache/Redis.php
```diff
@@ -79,11 +82,12 @@ public function get($key) {
 
 	public function set($key, $value, $ttl = 0) {
 		$value = self::encodeValue($value);
-		if ($ttl > 0) {
-			return $this->getCache()->setex($this->getPrefix() . $key, $ttl, $value);
-		} else {
-			return $this->getCache()->set($this->getPrefix() . $key, $value);
+		if ($ttl === 0) {
```
come-nc, line 85: I’m a tiny bit worried that we may have code path not sending an int in $ttl since it is not strong typed, so here passing '0' would still result in infinite TTL I think?
Not sure if this worry is justified.

skjnldsv (approved):

lib/private/Memcache/Redis.php
skjnldsv: We were already checking for an int, I'd say LGTM

PR #38568, opened 2023-06-01
redis: Do not try to authenticate with non-string password/user

rawtaz:
* Resolves: The never ending `ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?` errors users encounter (usually when upgrading).

## Summary

If the configuration for Redis in Nextcloud has a password that is false (e.g. `'password' => false` or `'password' => getenv('REDIS_HOST_PASSWORD')` with no environment variable set), `RedisFactory.php`'s test `if (isset($config['password']) && $config['password'] !== '')` will evaluate to true, and due to that Nextcloud will try to *authenticate* with Redis even though there is no password actually set.

Example config from an old (v19) installation that was updated where the problem triggers:
~~~
<?php
if (getenv('REDIS_HOST')) {    // <-- Set.
  $CONFIG = array (
    'memcache.distributed' => '\OC\Memcache\Redis',
    'memcache.locking' => '\OC\Memcache\Redis',
    'redis' => array(
      'host' => getenv('REDIS_HOST'),
      'password' => getenv('REDIS_HOST_PASSWORD'),    // <-- Not set.
    ),
  );

  if (getenv('REDIS_HOST_PORT') !== false) {
    $CONFIG['redis']['port'] = (int) getenv('REDIS_HOST_PORT');
  } elseif (getenv('REDIS_HOST')[0] != '/') {
    $CONFIG['redis']['port'] = 6379;
  }
}  
~~~

This in turn has caused the so many times encountered `ERR AUTH <password> called without any password configured for the default user. Are you sure your configuration is correct?` error when using Nextcloud with Redis. Many users have encountered this over many years, and the usual "solution" to that is to set a password for Redis directly or using an environment variable. However, that is not an actual fix for the actual cause of this error. In particular for environments where the Redis instance is only reachable by Nextcloud alone - encouraging people to use passwords with Redis is something that should be done by other means than having code that does not work as it should when a password is not set.

So, if there is no password set in such a way that it evaluates to false in the Redis config in Nextcloud, then Nextcloud should not attempt to authenticate when contacting Redis. This PR fixes that, it only tries to authenticate if there is any content in the password. The same logic is also applied to the `user` setting in the configuration for Redis in Nextcloud.

## TODO

- Merge it :-) I've tested this PR with the latest release of NC which is v26.0.2.

## Checklist

### DIFF
--- a/lib/private/RedisFactory.php
+++ b/lib/private/RedisFactory.php
@@ -72,8 +72,8 @@ class RedisFactory {
 		}
 
 		$auth = null;
-		if (isset($config['password']) && $config['password'] !== '') {
-			if (isset($config['user']) && $config['user'] !== '') {
+		if (isset($config['password']) && (string)$config['password'] !== '') {
+			if (isset($config['user']) && (string)$config['user'] !== '') {
 				$auth = [$config['user'], $config['password']];
 			} else {
 				$auth = $config['password'];


solracsf (approved):
Makes sense 🆗

kesselb:
Your pull request title and description is highly confusing.

`password => ''` already works. 
`password => false` do not work. 

Where did we document that false is a valid value for password or user?

come-nc (changes requested):
Nope.
If we want to accept non-string values, cast to string, but do not call strlen with something that may not be a string, it will cause type errors.

come-nc:
Nothing in https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#memory-caching-backend-configuration or config.sample.php suggests that any other type than a string may be supported for password and username.

"Many users have encountered this over many years", could you reference to these users? Was that really related to having false in this config option? That would indicate that there is a bad configuration or documentation somewhere if many user end up with false in there, I’m curious.

rawtaz:
Good to have you chaps' eyes on this.

First background, I am using the official nextcloud:26.0.2-fpm-alpine Docker image. No modifications to it, but it has been upgraded from a v19 image which did have `'password' => false` in `config/config.php` and `'password' => getenv('REDIS_HOST_PASSWORD')` in `config/redis.config.php`. Neither of these has been manually put there AFAIK, both should have been populated by the older image.

> If we want to accept non-string values, cast to string, but do not call strlen with something that may not be a string, it will cause type errors.

I am not getting any type errors or warnings with the `strlen()` using this official Docker image, at least not in the PHP error log and with `error_reporting` set to `E_ALL`, and nothing on the website. Nonetheless, I changed it to cast to string instead of using `strlen()`.

> Your pull request title and description is highly confusing.

I'm inclined to agree. I had not noticed the `config.redis.php` that was using `getenv()` which returned the `false` (there should never have been any duplicate configurations created in the first place by the older image), but this clears it up. The only problem to address is that some old configs yield a `false` value for the Redis password, and this causes an internal server error.

> `password => ''` already works.

You are correct, I made changes to the title, description and commit to reflect this. This PR is now only about handling non-string values from the config properly to avoid the ISE.

> password => false do not work.
> Where did we document that false is a valid value for password or user?

It's not documented anywhere AFAIK, but several users have encountered this problem when their config has whatever results in a `false` value for the password, which mainly happens when they have the default-populated `getenv()` call in place and there's no environment variable set. It's unfortunate if they then have to resort to setting a password using the env var or make changes to the `redis.config.php` such that it does not return `false` when `getenv()` doesn't find an env var, it's just messy. Better that the Nextcloud code handles it if it isn't a big deal (in this PR it's just casting to `string` and nothing more).

> "Many users have encountered this over many years", could you reference to these users?

If you search for the first part of the error message you will find several Nextcloud users who encountered this problem. It seems to me that most of them resorted to setting a password or an environment variable, even though this was not their desire in the first place (the problem started happening when upgrading).

Indeed, once could argue "fix your config". But since there's a value in having e.g. the `getenv()` there I think it's better that the Redis factory handles non-string and non-zero-length values, otherwise people who are not technical will need to modify their config file in a way that equates to writing code (at least for old configs, we could make new auto-generated configs generate more suitable code that does not possibly produce `false` for the Redis password but still takes into account any env vars).

come-nc:
> I am not getting any type errors or warnings with the `strlen()` using this official Docker image, at least not in the PHP error log and with `error_reporting` set to `E_ALL`, and nothing on the website. Nonetheless, I changed it to cast to string instead of using `strlen()`.

Yeah but this may be dependent on PHP version, and on `strict_types`. We tend to activate strict types more and more, so best to avoid code that will break under it.

Being able to use `getenv` is a valid usecase. It would makemore sense to cast in the config.php but it does not harm to cast what we get from there to avoid surprises.

come-nc (approved):

kesselb:
I think this should be solved here: https://github.com/nextcloud/docker/blob/22d28e45cdcdb553239af45c7b1566d628f72bc6/.config/redis.config.php#L8

PR #30276, opened 2021-12-15
Add oauth2_clients migration for Owncloud

artonge:
Handy script to test the migration. Use it from a nextcloud root directory containing the migration.

```shell
#!/bin/bash

set -eu

# Serving ownCloud then Nextcloud at localhost:8080
# Username: admin
# Password: admin

docker kill oc || true
docker rm oc || true

docker run \
	--rm \
	--name oc \
	--detach \
	--env OWNCLOUD_DOMAIN=localhost:8080 \
	--publish 8080:8080 \
	--volume /var/www/owncloud \
	--volume "$PWD":/mnt/local \
	owncloud/server:10.5

docker exec -it oc apt update
docker exec -it oc apt install -y nano rsync

docker exec -it oc sed -i '9,14d' /var/www/owncloud/config/config.php
docker exec -it oc occ app:enable oauth2
docker exec -it oc occ market:install calendar
docker exec -it oc occ app:enable configreport
docker exec -it oc cp /var/www/owncloud/config/config.php /var/www/config.php

instance=""
read -rp "> Double press enter to start the migration." instance

docker exec -it oc rsync --delete --recursive --human-readable --exclude .git --exclude node_modules /mnt/local/ /var/www/owncloud
docker exec -it oc cp /var/www/config.php /var/www/owncloud/config/config.php
docker exec -it oc chown -R www-data:www-data /var/www/owncloud

docker exec -it oc occ upgrade -vvv || true

occ db:convert-filecache-bigint
occ db:add-missing-columns
occ db:add-missing-indices
occ db:add-missing-primary-keys

# Use this command to copy your change in MigrateOauthTables.php to the docker container.
# docker exec -it oc cp /mnt/local/lib/private/Repair/Owncloud/MigrateOauthTables.php /var/www/owncloud/lib/private/Repair/Owncloud/MigrateOauthTables.php
```

PR #38577, opened 2023-06-01
Improve oauth2 database migration from ownCloud

julien-nc:
* Drop the oauth2_clients trusted column
* Delete unsupported clients and their access tokens
  The redirect_uri set by OC can be like `oc://android.owncloud.com` or `http://host/*` and we don't support those.

I tested it by first putting those changes in a migration step in the oauth2 app. Worked fine.

### DIFF
--- a/lib/private/Repair/Owncloud/MigrateOauthTables.php
+++ b/lib/private/Repair/Owncloud/MigrateOauthTables.php
@@ -82,6 +82,9 @@ class MigrateOauthTables implements IRepairStep {
 		if ($table->hasColumn('allow_subdomains')) {
 			$table->dropColumn('allow_subdomains');
 		}
+		if ($table->hasColumn('trusted')) {
+			$table->dropColumn('trusted');
+		}
 
 		if (!$schema->getTable('oauth2_clients')->hasColumn('client_identifier')) {
 			$table->addColumn('client_identifier', 'string', [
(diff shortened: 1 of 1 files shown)


lib/private/Repair/Owncloud/MigrateOauthTables.php
```diff
@@ -119,5 +122,36 @@ public function run(IOutput $output) {
...
+			->where(
+				$qbSelectClientId->expr()->iLike('redirect_uri', $qbDeleteAccessTokens->createNamedParameter('oc://%', IQueryBuilder::PARAM_STR))
+			)
+			->orWhere(
+				$qbSelectClientId->expr()->iLike('redirect_uri', $qbDeleteAccessTokens->createNamedParameter('%*', IQueryBuilder::PARAM_STR))
+			);
+
+		$qbDeleteAccessTokens->delete('oauth2_access_tokens')
+			->where(
+				$qbSelectClientId->expr()->in('client_id', $qbDeleteAccessTokens->createFunction($qbSelectClientId->getSQL()), IQueryBuilder::PARAM_STR_ARRAY)
```
artonge, line 142: Does this work? I though `getSQL` didn't replace the parameters.

artonge:
If this can help, I kept a script to test the initial version of the migration: https://github.com/nextcloud/server/pull/30276

lib/private/Repair/Owncloud/MigrateOauthTables.php
```diff
@@ -119,5 +122,36 @@ public function run(IOutput $output) {
...
+
+		$output->info('Delete clients (and their related access tokens) with the redirect_uri starting with oc:// or ending with *');
+		// delete the access tokens
+		$qbDeleteAccessTokens = $this->db->getQueryBuilder();
+
+		$qbSelectClientId = $this->db->getQueryBuilder();
+		$qbSelectClientId->select('id')
+			->from('oauth2_clients')
+			->where(
+				$qbSelectClientId->expr()->iLike('redirect_uri', $qbDeleteAccessTokens->createNamedParameter('oc://%', IQueryBuilder::PARAM_STR))
```
juliushaertl, line 134: Do we know if the column is indexed on the owncloud side?

AndyScherzinger: We don't but I guess @pmarini-nc can get that information for us

pmarini-nc: I understand the column that we want to know whether it is indexed is `oauth2_clients.id`. I asked the prospect

lib/private/Repair/Owncloud/MigrateOauthTables.php
julien-nc: That works. Maybe because the expression is built with the same query builder.
`$qbSelectClientId->expr()` and `$qbSelectClientId->getSQL()`

lib/private/Repair/Owncloud/MigrateOauthTables.php
julien-nc: I installed ownCloud 10.12.1. The only oauth2 related index is on the `token` column of `oauth2_access_tokens`.

julien-nc:
Potential problem: The `name` column of `oauth2_clients` is a `VARCHAR(80)` in ownCloud and a `VARCHAR(64)` in Nextcloud.
No problem for the migration, the DB will keep the column with a length of 80.
But what if an existing stored value's length is more than 64 and we, for some reason, insert it somewhere else, expecting it to be 64 max?

Do you think we should cut the values during the migration?

AndyScherzinger:
> Do you think we should cut the values during the migration?

I think we should just bump the field length to `VARCHAR(64)` during/after migration, so it matches our Nextcloud-length. We don't need to cut anything during migration since the original oC length is shorter than ours however any newly added row after the migration to Nextcloud would expect 64 and might fail. Hence the DB structure needs to be updated to `VARCHAR(64)`.

To support future migrations it be code if this is done in code by us and not purely via DBA cli else we might miss it in future migration scenarios.

Wat do you think @julien-nc ?

julien-nc:
@AndyScherzinger Actually we already set the `name` column length to 64 in this ownCloud oauth2 repair step: https://github.com/nextcloud/server/blob/master/lib/private/Repair/Owncloud/MigrateOauthTables.php#L79-L81

I wonder what happens for rows having a name which is longer than 64.
I think running `UPDATE oc_oauth2_clients SET name=substr(name, 1, 64);` before changing the column size is safer.

julien-nc:
@AndyScherzinger https://github.com/nextcloud/server/pull/38577/commits/aa3f6b1701e5833682337b02805b99dcc7b6c880 shortens the values before resizing the column.

AndyScherzinger:
Sounds/looks good to me @julien-nc 👍

lib/private/Repair/Owncloud/MigrateOauthTables.php
juliushaertl: All right, then this would require a full table scan for the like statements, which could take longer depending on the size of the table, but I'd say since this is a onetime migration this should be fine.

juliusknorr (approved):

AndyScherzinger (approved):
LGTM 👍 
...as per our discussions within the PR :)

julien-nc:
CI failures are not related (Cypress jobs)

PR #38591, opened 2023-06-01
fix(caching): Avoid checking existence before fetching

ChristophWurst:
## Summary

Following https://github.com/nextcloud/server/blob/ff661045edd3a796d2a555e5a8d9762eb8ea30fb/lib/public/ICache.php#L59.

The cache might expire between checking for key existence and fetching the value. In this rare case the code continues with a null value when it doesn't expect one.

## Checklist


nickvergessen (approved):

pulsejet (approved):
The doc here should be updated to say that `null` must be returned if the key doesn't exist. Implementions might end up returning something else, e.g. `false`

https://github.com/nextcloud/server/blob/ff661045edd3a796d2a555e5a8d9762eb8ea30fb/lib/public/ICache.php#L36-L42

Pytal (approved):

PR #38613, opened 2023-06-02
feat(HTTPClient): Provide wrapped access to Guzzle's asyncRequest()

### LINKED ISSUE #35959: Exposing GuzzleHttp\Client::asyncRequest()
Since quite some time Guzzle added support for async requests:
https://docs.guzzlephp.org/en/stable/quickstart.html#async-requests

It is also available in the code base we ship, but currently not accessible as we wrap away guzzle with our 
https://github.com/nextcloud/server/blob/8aea25b5b92dac105f7e862470ee0dcf0e876615/lib/private/Http/Client/Client.php#L51

Async requests would come in very handy and could help to speed up several requests were we don't care about the outcome, e.g. potential webhooks.

A method like the following would work
```php
	public function postAsync(string $uri, array $options = []): void {
		$this->preventLocalAddress($uri, $options);

		if (isset($options['body']) && is_array($options['body'])) {
			$options['form_params'] = $options['body'];
			unset($options['body']);
		}

		$this->client->requestAsync('post', $uri, $this->buildRequestOptions($options));
	}
```

but it hides away the Promise option. But if we'd expose that, it would mean more wrapping or exposing Guzzle.
Unluckily the Async stuff is not part of the PSR, so we can't write a generic wrapper for that.

Other current solutions:
* Make the app/server depend on Guzzle directly
* Ignoring user cancel and send the output while the process continues afterwards: https://github.com/nextcloud/circles/blob/96447086eca1b4dfba21b05f750de3f9c13e605e/lib/Service/MiscService.php#L358-L370

Any preferences?
cc @come-nc @ChristophWurst

nickvergessen:
* Resolves: #35959 

## Summary

Following the approach discussed in https://github.com/nextcloud/server/issues/35959 to allow making async requests, e.g. required for interaction with webhooks in a non-blocking way.

## Sample usage

```php
$client = $this->clientService->newClient();
$promise = $client->postAsync($webhook->getUrl(), $data);

$promise->then(function (IResponse $response) use ($webhook) {
	if ($response->getStatusCode() !== Http::STATUS_OK && $response->getStatusCode() !== Http::STATUS_ACCEPTED) {
		$this->logger->error('Webhook responded with unexpected status code, increasing error count');
	}
	$this->logger->debug('Webhook successfully executed');
}, function (RequestException $exception) {
	$this->logger->error('Webhook error occurred, increasing error count', ['exception' => $exception]);
});
```

## Checklist


nickvergessen:
Tests and docs will follow after my vacation.
But could get the first round of reviews on the approach in the meantime 😎

lib/private/Http/Client/ClientService.php
```diff
@@ -59,6 +60,7 @@ public function __construct(
 		DnsPinMiddleware $dnsPinMiddleware,
 		IRemoteHostValidator $remoteHostValidator,
 		IEventLogger $eventLogger,
+		protected LoggerInterface $logger,
```
blizzz, line 63: is it used anywhere?

nickvergessen: right, removed it from the Client calls again

come-nc: I think it is used, it’s passed to `Client` which passes it to `GuzzlePromiseAdapter` which uses it.

lib/public/Http/Client/IPromise.php
```diff
@@ -0,0 +1,97 @@
...
+	 *
+	 * @param bool $unwrap
+	 *
+	 * @return mixed
+	 *
+	 * @throws \LogicException if the promise has no wait function or if the
+	 *                         promise does not settle after waiting.
+	 * @since 28.0.0
+	 */
+	public function wait($unwrap = true);
```
ChristophWurst, line 96:
```suggestion
	public function wait(bool $unwrap = true): mixed;
```

lib/public/Http/Client/IPromise.php
```diff
@@ -0,0 +1,97 @@
...
+		callable $onRejected = null,
+	): IPromise;
+
+	/**
+	 * Get the state of the promise ("pending", "rejected", or "fulfilled").
+	 *
+	 * The three states can be checked against the constants defined on
+	 * PromiseInterface: PENDING, FULFILLED, and REJECTED.
+	 *
+	 * @return string
```
ChristophWurst, line 67:
```suggestion
	 * @return self::STATE_*
```

if you rename the consts

lib/public/Http/Client/IPromise.php
```diff
@@ -0,0 +1,97 @@
...
+	/**
+	 * @since 28.0.0
+	 */
+	public const REJECTED = 'rejected';
+
+	/**
+	 * Appends fulfillment and rejection handlers to the promise, and returns
+	 * a new promise resolving to the return value of the called handler.
+	 *
+	 * @param callable $onFulfilled Invoked when the promise fulfills. Gets an \OCP\Http\Client\IResponse passed in as argument
```
ChristophWurst, line 50: can we type the two callables?

ChristophWurst (approved):
Looks good

nickvergessen:
@come-nc @blizzz second review?

lib/private/Http/Client/GuzzlePromiseAdapter.php
```diff
@@ -0,0 +1,139 @@
...
+		?callable $onFulfilled = null,
+		?callable $onRejected = null,
+	): IPromise {
+		if ($onFulfilled !== null) {
+			$wrappedOnFulfilled = static function (ResponseInterface $response) use ($onFulfilled) {
+				$onFulfilled(new Response($response));
+			};
+		} else {
+			$wrappedOnFulfilled = null;
+		}
```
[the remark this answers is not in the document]
come-nc: Oh it wraps the internal $response object in something from OCP, right? Then I get it, but I do not get the one for onRejected.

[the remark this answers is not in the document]
nickvergessen: yeah, it wraps into an OCP interface.
The onRejected I didn't manipulate, as the Guzzle interface just says it throws \Exception, so not sure it makes too much sense to wrap it in yet another exception. I had half of the code written to wrap guzzle ClientException and ServerException, but since that does not cover all of them I kind of gave up, especially since the same exceptions are leaked/handed out by the normal IClient methods.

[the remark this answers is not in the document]
nickvergessen: yeah it's about only exposing OCP.
Same basically on the rejection, we are removing the Guzzle "type hint" and just require a `\Exception` eating closure instead of a `GuzzleHttp\Exception\RequestException` one.

come-nc (approved):

PR #38260, opened 2023-05-15
Refactors "strpos" calls in  lib/public to improve code readability.

fsamapoor:
- [x] Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)

- [x] [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits

PR #38619, opened 2023-06-02
Refactors "strpos" calls in /apps/files_external

fsamapoor:
## Summary
Following https://github.com/nextcloud/server/pull/38261 and https://github.com/nextcloud/server/pull/38260, I have replaced `strpos` calls in `/apps/files_external` namespace as well to improve code readability.

### DIFF
--- a/apps/files_external/lib/Lib/Storage/SMB.php
+++ b/apps/files_external/lib/Lib/Storage/SMB.php
@@ -141,9 +141,7 @@ class SMB extends Common implements INotifyStorage {
 	private function splitUser($user) {
 		if (str_contains($user, '/')) {
 			return explode('/', $user, 2);
-		}
-
-		if (str_contains($user, '\\')) {
+		} elseif (str_contains($user, '\\')) {
 			return explode('\\', $user);
 		}
 
--- a/apps/files_external/lib/Lib/Storage/SMB.php
+++ b/apps/files_external/lib/Lib/Storage/SMB.php
@@ -139,13 +139,15 @@ class SMB extends Common implements INotifyStorage {
 	}
 
 	private function splitUser($user) {
-		if (strpos($user, '/')) {
+		if (str_contains($user, '/')) {
 			return explode('/', $user, 2);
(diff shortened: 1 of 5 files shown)


icewind1991 (approved):

apps/files_external/lib/Lib/Storage/SMB.php
```diff
@@ -139,13 +139,15 @@ public function __construct($params) {
 	}
 
 	private function splitUser($user) {
-		if (strpos($user, '/')) {
+		if (str_contains($user, '/')) {
 			return explode('/', $user, 2);
-		} elseif (strpos($user, '\\')) {
+		}
+
+		if (str_contains($user, '\\')) {
```
come-nc, line 146:
```suggestion
		} elseif (str_contains($user, '\\')) {
```

fsamapoor: Hello @come-nc 
Is there any benefit to use the `elseif` when the previous `if` block returns the value?
Basically, there is no difference between having `if` instead of `elseif` in this case, but I'm my opinion, the more we can reduce the code indentation, the more readable the codebase becomes.

come-nc: The `elseif` is easier to read, you see right away that there are two cases handled differently. It also avoids having to refactor if the code changes in the future.

I would also prefer to keep the `else`, otherwise I have to read all paths above to check that they all returns to understand that the last part is an `else` block. On the other end having it without `else` makes it clearer that the function always returns, so I’m less sure for this one.

fsamapoor: I'm used to avoiding the `else` conditions and structuring the logic with guard clauses, but that's a personal preference of mine and it is not the point of this PR. Thanks for the clarification and sharing your point of view. The suggested change has been committed.

nfebe (approved):

come-nc (approved):

szaimen:
CI failure unrelated

PR #38636, opened 2023-06-04
[1/3] Refactors /core controllers using constructor property promotion.

fsamapoor:
This PR is the first part of the three-part refactoring of the `/core` controllers by using PHP8's constructor property promotion to remove redundant lines and make the code cleaner.

### DIFF
--- a/core/Controller/AppPasswordController.php
+++ b/core/Controller/AppPasswordController.php
@@ -45,11 +45,12 @@ class AppPasswordController extends \OCP\AppFramework\OCSController {
 	public function __construct(
 		string $appName,
 		IRequest $request,
-								private ISession $session,
-								private ISecureRandom $random,
-								private IProvider $tokenProvider,
-								private IStore $credentialStore,
-								private IEventDispatcher $eventDispatcher) {
+		private ISession $session,
+		private ISecureRandom $random,
+		private IProvider $tokenProvider,
+		private IStore $credentialStore,
+		private IEventDispatcher $eventDispatcher,
+	) {
 		parent::__construct($appName, $request);
 	}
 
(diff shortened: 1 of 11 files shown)


core/Controller/AppPasswordController.php
```diff
@@ -42,26 +42,14 @@
...
 
 class AppPasswordController extends \OCP\AppFramework\OCSController {
-	private ISession $session;
-	private ISecureRandom $random;
-	private IProvider $tokenProvider;
-	private IStore $credentialStore;
-	private IEventDispatcher $eventDispatcher;
-
 	public function __construct(string $appName,
 								IRequest $request,
```
come-nc, line 46:
```suggestion
	public function __construct(
		string $appName,
		IRequest $request,
```
If we go with this refactor I’d prefer if we use this indentation method to avoid crazy empty spaces.
Also the last one can have a comma as well, resulting in clearer diff when we add a property later.

fsamapoor: I think I might have messed up something in this PR regarding the spaces. Sorry about that.

Although I like the idea of always having a last comma, I just tried to make them consistent with the existing conventions in the controller classes. No personal preference intended. I appreciate your attention to this matter.

come-nc (approved):

core/Controller/AvatarController.php
```diff
@@ -52,36 +51,19 @@
  * @package OC\Core\Controller
  */
 class AvatarController extends Controller {
-	protected IAvatarManager $avatarManager;
-	protected ICache $cache;
-	protected IL10N $l;
-	protected IUserManager $userManager;
-	protected IUserSession $userSession;
```
artonge, line 59: No need for `$userSession` anymore?

artonge (approved):
Looks good

core/Controller/AvatarController.php
fsamapoor: There was no usage of the `$userSession` property in the class and my IDE found no usage in all places. So it's probably redundant; if not, I thought the test suite would show something has broken somewhere else.

PR #38637, opened 2023-06-04
[2/3] Refactors /core controllers using constructor property promotion.

fsamapoor:
This PR is the second part of the three-part refactoring of the `/core` controllers by using PHP8's constructor property promotion to remove redundant lines and make the code cleaner.

### DIFF
--- a/core/Controller/JsController.php
+++ b/core/Controller/JsController.php
@@ -45,13 +45,14 @@ use OCP\IRequest;
 
 class JsController extends Controller {
 	protected IAppData $appData;
-	protected ITimeFactory $timeFactory;
 
-	public function __construct($appName, IRequest $request, Factory $appDataFactory, ITimeFactory $timeFactory) {
+	public function __construct(string $appName,
+								IRequest $request,
+								Factory $appDataFactory,
+								protected ITimeFactory $timeFactory) {
 		parent::__construct($appName, $request);
 
 		$this->appData = $appDataFactory->get('js');
-		$this->timeFactory = $timeFactory;
 	}
 
 	/**
--- a/core/Controller/JsController.php
+++ b/core/Controller/JsController.php
@@ -46,10 +46,12 @@ use OCP\IRequest;
 class JsController extends Controller {
 	protected IAppData $appData;
 
-	public function __construct(string $appName,
-								IRequest $request,
-								Factory $appDataFactory,
-								protected ITimeFactory $timeFactory) {
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		Factory $appDataFactory,
+		protected ITimeFactory $timeFactory,
+	) {
 		parent::__construct($appName, $request);
 
 		$this->appData = $appDataFactory->get('js');
--- a/core/Controller/LostController.php
+++ b/core/Controller/LostController.php
@@ -73,54 +73,26 @@ use function reset;
  * @package OC\Core\Controller
  */
 class LostController extends Controller {
-	protected IURLGenerator $urlGenerator;
-	protected IUserManager $userManager;
-	protected Defaults $defaults;
-	protected IL10N $l10n;
 	protected string $from;
-	protected IManager $encryptionManager;
-	protected IConfig $config;
-	protected IMailer $mailer;
-	private LoggerInterface $logger;
-	private Manager $twoFactorManager;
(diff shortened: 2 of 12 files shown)


core/Controller/JsController.php
```diff
@@ -45,13 +45,16 @@
 
 class JsController extends Controller {
 	protected IAppData $appData;
-	protected ITimeFactory $timeFactory;
 
-	public function __construct($appName, IRequest $request, Factory $appDataFactory, ITimeFactory $timeFactory) {
+	public function __construct(
+		string $appName,
```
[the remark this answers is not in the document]
fsamapoor: I just double-checked and in all other controllers where the `$appName` was being typed, it was `string $appName`.
The only exception is `LoginController` where it is `?string` as you mentioned.
How could we make sure what's the right type here?
should I just skip the typing for this variable just to be sure that nothing is broken?

core/Controller/LostController.php
```diff
@@ -73,54 +73,28 @@
...
-		IUserManager $userManager,
-		Defaults $defaults,
-		IL10N $l10n,
-		IConfig $config,
+		private IURLGenerator $urlGenerator,
+		private IUserManager $userManager,
+		private Defaults $defaults,
+		private IL10N $l10n,
+		private IConfig $config,
 		string $defaultMailAddress,
```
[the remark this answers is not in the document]
come-nc: This is the from address, not the default one, I do not get the "default" naming for the constructor parameter. Maybe I missed something.

My proposition was to rename both parameter and property to `$fromMailAddress` to be able to promote it with the rest.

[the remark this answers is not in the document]
fsamapoor: Done.

come-nc (approved):

core/Controller/LostController.php
```diff
@@ -73,54 +73,25 @@
...
-		IInitialState $initialState,
-		IVerificationToken $verificationToken,
-		IEventDispatcher $eventDispatcher,
-		Limiter $limiter
+		private IURLGenerator $urlGenerator,
+		private IUserManager $userManager,
+		private Defaults $defaults,
+		private IL10N $l10n,
+		private IConfig $config,
+		protected string $fromMailAddress,
```
artonge, line 84: Why the rename? Wouldn't this break DI?

fsamapoor: This was a suggestion from @come-nc which we discussed earlier:
https://github.com/nextcloud/server/pull/38637#discussion_r1222822079
I don't think it would break anything. Any suggestion about how to make sure of it?

artonge: `$defaultMailAddress` is set here: https://github.com/nextcloud/server/blob/1001adc390ce873a0bf9b2e559853702f2b02ea7/core/Application.php#L79
So if we rename, we have to rename there too.

I think the original name was ok. @come-nc Any reason you want to rename it?
We can do a middle ground with `defaultFromMailAddress`.


FYI, you can test be trying to reset a password from the login page.

come-nc: I had no idea this came from DI.

@fsamapoor Sorry about that, your original solution was the best one then.

fsamapoor: Thank you both.
I have reverted the latest commit regarding the renaming of the variable and now it's back to the prior state.

artonge: I'll be picky, but can we cut the middle man and remove `$this->from` and use `$defaultMailAddress` directly?
It's only used once.

fsamapoor: Would that be possible to replace `$this->from` with `$defaultMailAddress`?

If I understand correctly, I should promote the `$defaultMailAddress` property so that I can remove the `$from` property altogether and use the `$this->defaultMailAddress` instead, without the redundant `$from` proxy. Is that right?

artonge: Ah, but it's protected. Why is that? Can you check if the `LockControler` is being extended?

fsamapoor: I think you meant `LostController`. It is not.
I looked into the git history of the class and it seems to me that all of the properties had been defined as `protected` in the first place and eventually have been changed to `private`. `$from` is the only protected property right now.

artonge (approved):

PR #38638, opened 2023-06-04
[3/3] Refactors /core controllers using constructor property promotion.

fsamapoor:
This PR is the third part of the three-part refactoring of the `/core` controllers by using PHP8's constructor property promotion to remove redundant lines and make the code cleaner.

### DIFF
--- a/core/Controller/SearchController.php
+++ b/core/Controller/SearchController.php
@@ -34,10 +34,12 @@ use OCP\Search\Result;
 use Psr\Log\LoggerInterface;
 
 class SearchController extends Controller {
-	public function __construct(string $appName,
-								IRequest $request,
-								private ISearch $searcher,
-								private LoggerInterface $logger) {
+	public function __construct(
+		string $appName,
+		IRequest $request,
+		private ISearch $searcher,
+		private LoggerInterface $logger,
+	) {
 		parent::__construct($appName, $request);
 	}
 
--- a/core/Controller/SearchController.php
+++ b/core/Controller/SearchController.php
@@ -34,19 +34,11 @@ use OCP\Search\Result;
 use Psr\Log\LoggerInterface;
 
 class SearchController extends Controller {
-	private ISearch $searcher;
-	private LoggerInterface $logger;
-
-	public function __construct(
-		string $appName,
-		IRequest $request,
-		ISearch $search,
-		LoggerInterface $logger
-	) {
+	public function __construct(string $appName,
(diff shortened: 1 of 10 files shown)


core/Controller/SearchController.php
```diff
@@ -34,19 +34,13 @@
...
 class SearchController extends Controller {
-	private ISearch $searcher;
-	private LoggerInterface $logger;
-
 	public function __construct(
 		string $appName,
 		IRequest $request,
-		ISearch $search,
-		LoggerInterface $logger
+		private ISearch $searcher,
```
artonge, line 40: The rename might break DI, can you check that this does not change anything?

core/Controller/WellKnownController.php
```diff
@@ -33,13 +33,11 @@
...
 
 class WellKnownController extends Controller {
-	/** @var RequestManager */
-	private $requestManager;
-
-	public function __construct(IRequest $request,
-								RequestManager $wellKnownManager) {
+	public function __construct(
+		IRequest $request,
+		private RequestManager $requestManager,
```
artonge, line 38: Same, the rename might break DI.

core/Controller/SearchController.php
fsamapoor: Sure. Let me investigate and I'll keep you updated.

fsamapoor: I injected it as it is to a controller and it worked without any problems.

Furthermore, `ISearch` is [one of the documented type hints](https://docs.nextcloud.com/server/latest/developer_manual/basics/dependency_injection.html#predefined-core-services) that can be used to inject core services.

However, I'm totally fine with reverting the renamings. Honestly, this was not a goal of mine and I just renamed the parameters thinking the test suit will show if I have broken anything in the process.

artonge (approved):
LGTM

core/Controller/SetupController.php
```diff
@@ -35,15 +35,10 @@
...
 
 class SetupController {
-	protected Setup $setupHelper;
 	private string $autoConfigFile;
 
-	/**
-	 * @param Setup $setupHelper
-	 */
-	public function __construct(Setup $setupHelper) {
+	public function __construct(protected Setup $setupHelper) {
```
come-nc, line 40:
```suggestion
	public function __construct(
		protected Setup $setupHelper,
	) {
```

come-nc (approved):

PR #38645, opened 2023-06-05
Avoid requests to opengraph image if no host detected

julien-nc:
When using the `LinkReferenceProvider` to resolve `https://www.compart.com/en/unicode/U+033C`, for example, the OpenGraph image URI is `/en/unicode/images/icon/icon-compart.svg`. It is kind of fine because our HTTP client refuses to make requests when there is no host.

We might want to improve this.
With this PR, we don't make a request if no host is detected.

Maybe we even want to forbid local addresses regardless of the `allow_local_remote_servers` system config. What do you think?

PS: The "previews and thumbnails" label is not exactly appropriate but it's the most fitting :grin:

### DIFF
--- a/lib/private/Collaboration/Reference/LinkReferenceProvider.php
+++ b/lib/private/Collaboration/Reference/LinkReferenceProvider.php
@@ -151,22 +151,27 @@ class LinkReferenceProvider implements IReferenceProvider {
 
 		if ($object->images) {
 			try {
-				$appData = $this->appDataFactory->get('core');
-				try {
-					$folder = $appData->getFolder('opengraph');
-				} catch (NotFoundException $e) {
-					$folder = $appData->newFolder('opengraph');
-				}
-				$response = $client->get($object->images[0]->url, [ 'timeout' => 10 ]);
-				$contentType = $response->getHeader('Content-Type');
-				$contentLength = $response->getHeader('Content-Length');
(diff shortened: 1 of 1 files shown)


lib/private/Collaboration/Reference/LinkReferenceProvider.php
```diff
@@ -151,22 +151,27 @@ private function fetchReference(Reference $reference): void {
...
-				$contentLength = $response->getHeader('Content-Length');
-
-				if (in_array($contentType, self::ALLOWED_CONTENT_TYPES, true) && $contentLength < self::MAX_PREVIEW_SIZE) {
-					$stream = Utils::streamFor($response->getBody());
-					$bodyStream = new LimitStream($stream, self::MAX_PREVIEW_SIZE, 0);
-					$reference->setImageContentType($contentType);
-					$folder->newFile(md5($reference->getId()), $bodyStream->getContents());
-					$reference->setImageUrl($this->urlGenerator->linkToRouteAbsolute('core.Reference.preview', ['referenceId' => md5($reference->getId())]));
+				$host = parse_url($object->images[0]->url, PHP_URL_HOST);
+				if ($host === false || $host === null) {
```
ChristophWurst, line 155: nit: return early  to avoid indentation

ChristophWurst (approved):

juliusknorr (approved):

PR #38665, opened 2023-06-06
fix(weather_status): Pass address as param to OSM API

susnux:
## Summary
For some addresses the location can not be fetched although the address is correct.
For example `Bruxelles` will fetch the forecast initially but when adding it as a favorite and selecting it, it does not work because the address is saved as `Ville de Bruxelles - Stad Brussel, België / Belgique / Belgien` which contains the `/` character.

This is simply solved by passing the address as the query parameter (as the OSM documentation states).

### Screenshots (screen cast gifs)
before | after
---|---
before|after

## Checklist

### DIFF
--- a/apps/weather_status/lib/Service/WeatherStatusService.php
+++ b/apps/weather_status/lib/Service/WeatherStatusService.php
@@ -8,7 +8,7 @@ declare(strict_types=1);
  * @author Julien Veyssier <eneiluj@posteo.net>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
  *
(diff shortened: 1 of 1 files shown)


blizzz (approved):

nfebe (approved):
Cool!

tcitworld:
This also fixes another issue caused by Nominatim changing their API, see https://github.com/osm-search/Nominatim/issues/3134 and https://github.com/nextcloud/maps/issues/1111

Can we backport to stable26 and 25?

PR #38679, opened 2023-06-06
feat: add width, height, crop and mode to BeforePreviewFetchedEvent

kesselb:
## Summary

user_usage_report keeps a counter for files read.[^1]

The app uses

- OC_Filesystem::read hook
- Event listener (the legacy one) for IPreview::EVENT

Request for previews do not trigger the hook, hence the additional event listener. 
The thumbnails for a list or grid view should not count, so we need the width and height.

[^1]: https://github.com/nextcloud/user_usage_report/blob/c12e97d84033818af27ae5cf659c88f743137be7/lib/AppInfo/Application.php#L51

## TODO

- [x] CI

## Checklist

### DIFF
--- a/lib/public/Preview/BeforePreviewFetchedEvent.php
+++ b/lib/public/Preview/BeforePreviewFetchedEvent.php
@@ -27,19 +27,27 @@ declare(strict_types=1);
 namespace OCP\Preview;
 
 use OCP\Files\Node;
+use OCP\IPreview;
 
 /**
  * @since 25.0.1
  */
 class BeforePreviewFetchedEvent extends \OCP\EventDispatcher\Event {
-	private Node $node;
-
 	/**
 	 * @since 25.0.1
 	 */
-	public function __construct(Node $node) {
+	public function __construct(
+		private Node $node,
+		/** @deprecated 28.0.0 null deprecated **/
+		private ?int $width = null,
+		/** @deprecated 28.0.0 null deprecated **/
+		private ?int $height = null,
+		/** @deprecated 28.0.0 null deprecated **/
+		private ?bool $crop = null,
+		/** @deprecated 28.0.0 null deprecated **/
+		private ?string $mode = null,
+	) {
 		parent::__construct();
-		$this->node = $node;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


lib/public/Preview/BeforePreviewFetchedEvent.php
```diff
@@ -32,14 +32,14 @@
...
-	private Node $node;
-
 	/**
 	 * @since 25.0.1
 	 */
-	public function __construct(Node $node) {
+	public function __construct(
+		private Node $node,
+		private int $width,
+		private int $height) {
```
ChristophWurst, line 41: The two new required constructor arguments form a breaking change for apps creating an instance, e.g. to dispatch their own event.

Can we assume that apps don't do that? Then the change could be ok.

Otherwise I would prefer to make the two new properties optional and nullable, document that in three versions (nextcloud 31) the arguments are required and only then make them non-null.

ChristophWurst (changes requested):

tcitworld:
Would you mind adding more details while we're at it? What I added in https://github.com/nextcloud/server/pull/37193 for admin_audit (`crop`, `mode` and `mimetype`).

kesselb:
@tcitworld added crop and mode.

ChristophWurst (approved):

tcitworld (approved):

tcitworld (approved):

ChristophWurst (approved):

lib/public/Preview/BeforePreviewFetchedEvent.php
```diff
@@ -32,14 +32,21 @@
...
-	private Node $node;
-
 	/**
 	 * @since 25.0.1
 	 */
-	public function __construct(Node $node) {
+	public function __construct(
+		private Node $node,
+		/** @deprecated 28.0.0 default value is deprecated **/
+		private int $width = 0,
```
ChristophWurst, line 41: I wonder if it's better to have wrong data (0x0px image) or null values when the emitter doesn't pass values

kesselb: True True

ChristophWurst (approved):

lib/public/Preview/BeforePreviewFetchedEvent.php
```diff
@@ -48,4 +55,32 @@ public function __construct(Node $node) {
...
+	 * @since 28.0.0
+	 */
+	public function isCrop(): ?bool {
+		return $this->crop;
+	}
+
+	/**
+	 * @since 28.0.0
+	 */
+	public function getMode(): ?string {
```
ChristophWurst, line 83:
```suggestion
	/**
	 * @since 28.0.0
	 * @return null|IPreview::MODE_*
	 */
	public function getMode(): ?string {
```

lib/public/Preview/BeforePreviewFetchedEvent.php
```diff
@@ -32,14 +32,21 @@
  * @since 25.0.1
  */
 class BeforePreviewFetchedEvent extends \OCP\EventDispatcher\Event {
-	private Node $node;
-
 	/**
 	 * @since 25.0.1
 	 */
```
ChristophWurst, line 37:
```suggestion
	 * @param string|IPreview::MODE_*
	 */
```

ChristophWurst (approved):
Psalm sugar

lib/public/Preview/BeforePreviewFetchedEvent.php
kesselb: Good idea

st3iny (approved):

kesselb:
Documentation: https://github.com/nextcloud/documentation/pull/10674

PR #38701, opened 2023-06-08
Fix oauth2 tests

julien-nc:
that were broken by #38398


szaimen (approved):
Seems to fix the nodb tests. Thanks a lot! :)

kesselb (approved):

PR #38722, opened 2023-06-09
fix: Remove duplicate memory session creation on cron jobs

juliusknorr:
## Summary

There might be scenarios (especially on previous versions where session writes do not reopen) where application code lie the encryption KeyManager still holds the previously initialized Memory session which they obtained through DI. There should be no need to close the existing one and create a new one and crypto wrappers should also be unneeded as we are storing in memory only anyways.

There is already a memory session by default which is available if no session is set manually through `OC::initSession` or the calls removed in ths PR.

https://github.com/nextcloud/server/blob/a2afc7b6a9b42598cadcc3d3e9dde80e0a8a9ce4/lib/private/Server.php#L561

## TODO

- [x] Waiting for CI

## Checklist

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -54,14 +54,6 @@ try {
 	// load all apps to get all api routes properly setup
 	OC_App::loadApps();
 
-	\OC::$server->getSession()->close();
-
-	// initialize a dummy memory session
-	$session = new \OC\Session\Memory('');
-	$cryptoWrapper = \OC::$server->getSessionCryptoWrapper();
-	$session = $cryptoWrapper->wrapSession($session);
-	\OC::$server->setSession($session);
-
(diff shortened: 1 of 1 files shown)


nfebe (approved):
Makes sense!

come-nc (commented):
I do not grasp what this is about…
git says @nickvergessen touched these lines, do you know more?

nickvergessen:
PR was https://github.com/owncloud/core/pull/18482
🤔

icewind1991:
Would it make more sense to move the explicit memory session creation to before the app loading>

juliushaertl:
Closing as I cannot reproduce the underlying issue anymore.

PR #38737, opened 2023-06-10
Add a Dependabot configuration to autoupdate GitHub action versions

kurtmckee:
* Resolves: No issue filed

## Summary

GitHub workflow logs show that some of the actions are running on Node 12, which is deprecated and will be removed soon. [[example from a recent run](https://github.com/nextcloud/server/actions/runs/5231339838)]

This can be addressed by adding a Dependabot configuration for GitHub action versions. Therefore this PR targets a Dependabot configuration for ongoing updates, rather than updating the action versions as a one-off.

I've added sign-off signatures to my git commits. Please let me know if I overlooked anything that needs to be addressed.

Thanks for your work on NextCloud!

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -171,3 +171,12 @@ updates:
     # ignore all GitHub linguist patch updates
     - dependency-name: "*"
       update-types: ["version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]
+
+# GitHub Actions
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: weekly
+    day: saturday
+    time: "03:00"
+    timezone: Europe/Paris


.github/dependabot.yml
```diff
@@ -171,3 +171,12 @@ updates:
...
       update-types: ["version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]
+
+# GitHub Actions
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    interval: weekly
+    day: saturday
+    time: "03:00"
+    timezone: Europe/Paris
```
Fenn-CS, line 182:
```suggestion
```

nfebe (changes requested):
Thank you! This looks great according to official [GitHub notes.](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot)

One minor change request, can we just leave this as weekly? And leave the default `Once-a-week` timing for Github (Scheduled on Mondays?)

_Consideration: We could also make this monthly, I don't think actions change so fast._

.github/dependabot.yml
kurtmckee: I don't have any preference at all, so I just want to confirm before signing off and committing your suggestion:

All other Dependabot updates are scheduled for "Saturdays at 3:00am Europe/Paris", so I simply copied that schedule. If I make this change, I think that it will introduce Dependebato PRs on a noticeably different schedule than the others.

Please confirm that's acceptable and intended, and I'll take your suggestion and add that new commit. 👍

susnux (approved):

.github/dependabot.yml
kurtmckee: @Fenn-CS Please let me know if this change is required, or if following the other Dependabot schedules in the file is acceptable.

I think you'll need to update your review to "Approve" for this to be eligible for merging.

Fenn-CS: @kurtmckee Sorry for the delay! Too many notifications....

nfebe (approved):
Thank you!

Fenn-CS:
We would then wait for CI to get green :)

kurtmckee:
Thanks, @Fenn-CS! I want CI to go green, too.

It appears the failures are unrelated to this change, however, and that many recent commits to the default git branch have failed for varying reasons. Would it be an option to re-run the failed CI jobs to see if they clear up on a second run?

Fenn-CS:
You could rerun, but usually I just wait when I know it's unrelated.

It would magically become green if it can :)

kurtmckee:
I'm not able to independently re-run the failed jobs. Is that something you're able to re-run?

szaimen:
CI failure unrelated

Fenn-CS:
@skjnldsv Not a big deal, so if we are having too many depenabot pr's to review that no one has time to look at, it makes sense to revert.

Fenn-CS:
>     The update cadence seems aggressive; it could be tuned to "monthly" instead of "weekly"

I am also fine with this.

kurtmckee:
Dependabot supports batched updates. [Here's an example PR] from one of my repositories, with all available GitHub actions updates rolled into a single PR.

skjnldsv:
> The update cadence seems aggressive; it could be tuned to "monthly" instead of "weekly"

good for me too

Issue #33330, opened 2022-07-24
[Bug]: Provisionning API /ocs/v1.php/cloud/apps/<app-id> doesn't return anything in XML format


MercierCorentin:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Following these instruction : https://docs.nextcloud.com/server/latest/admin_manual/configuration_user/instruction_set_for_apps.html#get-app-info

I get no result when trying to get an app metadata with the provisioning API xml format.

When JSON formatting is requested, I get a valid result.
Others app routes are working, I can disable and enable an app. 

### Steps to reproduce

1. Create a fresh install from the official docker image
2. Use the following command:
```bash
curl -u {admin}:{password -X GET 'http://{nc-instance-domain}/ocs/v1.php/cloud/apps/files' -H "OCS-APIRequest: true"
```
4. You get a HTTP 200 OK response with a content length of 0. 

### Expected behavior

I would expect the documented response ([here](https://docs.nextcloud.com/server/latest/admin_manual/configuration_user/instruction_set_for_apps.html#get-app-info)):
```xml
<?xml version="1.0"?>
<ocs>
  <meta>
    <statuscode>100</statuscode>
    <status>ok</status>
  </meta>
  <data>
    <info/>
    <remote>
      <files>appinfo/remote.php</files>
      <webdav>appinfo/remote.php</webdav>
      <filesync>appinfo/filesync.php</filesync>
    </remote>
    <public/>
    <id>files</id>
    <name>Files</name>
    <description>File Management</description>
    <licence>AGPL</licence>
    <author>Robin Appelman</author>
    <require>4.9</require>
    <shipped>true</shipped>
    <standalone></standalone>
    <default_enable></default_enable>
    <types>
      <element>filesystem</element>
    </types>
  </data>
</ocs>
```

### Installation method

Official Docker image

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.0

[...]

PR #38745, opened 2023-06-12
Allow stdClass in XML responses

provokateurin:
## Summary

Allows returning `\stdClass()` in DataResponse that are encoded to XML.
JSON can already to it and converts it to `{}`, but for XML special handling is required.

This change is required in order to properly describe APIs using empty JSON objects (`[]` gets converted to `[]` in JSON instead of `{}`). XML just also needs to support it.

## Checklist

### DIFF
--- a/lib/private/AppFramework/OCS/BaseResponse.php
+++ b/lib/private/AppFramework/OCS/BaseResponse.php
@@ -143,6 +143,10 @@ abstract class BaseResponse extends Response {
 				$k = 'element';
 			}
 
+			if ($v instanceof \stdClass) {
+				$v = [];
+			}
+
 			if (\is_array($v)) {
 				$writer->startElement($k);
 				$this->toXML($v, $writer);
--- a/tests/lib/AppFramework/OCS/BaseResponseTest.php
+++ b/tests/lib/AppFramework/OCS/BaseResponseTest.php
@@ -45,13 +45,14 @@ class BaseResponseTest extends \Test\TestCase {
 				'someElement' => 'withAttribute',
 			],
 			'value without key',
+			'object' => new \stdClass(),
 		];
 
 		$this->invokePrivate($response, 'toXml', [$data, $writer]);
 		$writer->endDocument();
 
 		$this->assertEquals(
-			"<?xml version=\"1.0\"?>\n<hello>hello</hello><information test=\"some data\"><someElement>withAttribute</someElement></information><element>value without key</element>\n",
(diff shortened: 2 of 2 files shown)


come-nc (approved):

ChristophWurst (commented):
Could this be tested in ``\Test\AppFramework\Middleware\BaseResponseTest``?

ChristophWurst (approved):

AlexM4H:
Many thanks for the PR. A backport to Nextcloud 27.1.0 would be desirable.

provokateurin:
@AlexM4H does it fix a bug for you? This was only meant as a feature, so it doesn't get any backport

AlexM4H:
It fixes #33330  for me.

provokateurin:
On a technical level this is still a bug with the respective apps and not the core as this was never supported. @ChristophWurst @come-nc shall we backport it anyway?

come-nc:
Hard to say. Backporting API change is always dangerous as it may trigger subtle sideeffects.
On the other hand if it fixes bugs…

Without this an stdClass is just stripped from XML output?

@provokateurin I think you’re more knowledgeable than me to decide whether to backport.

provokateurin:
It's not breaking at all, but if an app returns stdClass the output without this patch would just be empty thus broken.

I still feel like the apps should fix it themselves as any version before this patch is broken. That would mean apps can target <28 if they rely on this.

AlexM4H:
Please keep in mind that Forms is a featured app. "Featured apps are developed by and in the community. They provide core functionality and are ready for production use."

The fix would be necessary for some app developers for NC 26 / 27, as NC 28 already includes the fix as a new feature.

AlexM4H:
The corresponding issue is still open 

https://github.com/nextcloud/forms/issues/1469

Chartman123:
@AlexM4H yes, I know. We're waiting for a server fix there 😆 

My question was for @provokateurin 😉

provokateurin:
> Please keep in mind that Forms is a featured app. "Featured apps are developed by and in the community. They provide core functionality and are ready for production use."

Sorry, this is not helpful. This is a bug with the app and not the core.

As I said the behavior the forms app is relying on was never intended to work (in previous versions). I added this as a feature but it is only available in >=28. If you target 28 and higher you can surely rely on it.

For example Talk also had to work around it like this: https://github.com/nextcloud/spreed/blob/eaf04bbe5a7626cc1565cc455124f8a10b2ff078/lib/Model/Message.php#L163

cc @DaphneMuller

Chartman123:
> For example Talk also had to work around it like this

Thanks I'll have a look at it :)

provokateurin:
In the future when you target >=28 you can of course drop the workaround.

PR #38748, opened 2023-06-12
fix(caldav): log serialised iTip message and method for schedule processing

miaulalala:
## Summary
Log the serialised data that gets passed to Sabre


miaulalala:
Wrong base branch

PR #34232, opened 2022-09-24
compare cached filesize on download

ArtificialOwl:
- need more test on specific storages (external, s3, ...)
- might need a refresh of parent[s] folder cached data

- [ ] TEST: with master-key encryption

PR #38763, opened 2023-06-12
[24] set stream size for SeekableHttpStream

icewind1991:
Partial backport of https://github.com/nextcloud/server/pull/34232 to allow other code to use the stream size


blizzz:
You added the review label, but it is still in draft state?

PR #38770, opened 2023-06-12
Make oauth2 client secret column larger

julien-nc:
refs https://github.com/nextcloud/server/pull/38398#discussion_r1226617872

This also adds a simple test inserting a client with a long secret.

@kesselb Any idea how much chars we can expect at most by giving a 64 chars to https://github.com/nextcloud/server/blob/master/lib/private/Security/Crypto.php#L93 ?

### DIFF
--- a/apps/oauth2/lib/Migration/Version011601Date20230522143227.php
+++ b/apps/oauth2/lib/Migration/Version011601Date20230522143227.php
@@ -49,7 +49,7 @@ class Version011601Date20230522143227 extends SimpleMigrationStep {
 			$table = $schema->getTable('oauth2_clients');
 			if ($table->hasColumn('secret')) {
 				$column = $table->getColumn('secret');
-				$column->setLength(256);
+				$column->setLength(512);
 				return $schema;
 			}
 		}
--- a/apps/oauth2/tests/Db/ClientMapperTest.php
+++ b/apps/oauth2/tests/Db/ClientMapperTest.php
@@ -84,4 +84,14 @@ class ClientMapperTest extends TestCase {
 	public function testGetClients() {
 		$this->assertSame('array', gettype($this->clientMapper->getClients()));
 	}
+
+	public function testInsertLongEncryptedSecret(): void {
+		$client = new Client();
+		$client->setClientIdentifier('MyNewClient');
+		$client->setName('Client Name');
+		$client->setRedirectUri('https://example.com/');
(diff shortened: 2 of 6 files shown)


apps/oauth2/tests/Db/ClientMapperTest.php
```diff
@@ -84,4 +84,13 @@ public function testGetByUidNotExisting() {
...
 	}
+
+	public function testInsertLongEncryptedSecret(): void {
+		$client = new Client();
+		$client->setClientIdentifier('MyNewClient');
+		$client->setName('Client Name');
+		$client->setRedirectUri('https://example.com/');
+		$client->setSecret('b81dc8e2dc178817bf28ca7b37265aa96559ca02e6dcdeb74b42221d096ed5ef63681e836ae0ba1077b5fb5e6c2fa7748c78463f66fe0110c8dcb8dd7eb0305b16d0cd993e2ae275879994a2abf88c68|e466d9befa6b0102341458e45ecd551a|013af9e277374483123437f180a3b0371a411ad4f34c451547909769181a7d7cc191f0f5c2de78376d124dd7751b8c9660aabdd913f5e071fc6b819ba2e3d919|3');
+		$this->clientMapper->insert($client);
+	}
```
ChristophWurst, line 95: forgot a addToAssertCount(1) or assertTrue(true). I think phpunit will warn that the test doesn't assert anything

apps/oauth2/lib/Migration/Version011601Date20230522143227.php
```diff
@@ -49,7 +49,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$table = $schema->getTable('oauth2_clients');
 			if ($table->hasColumn('secret')) {
 				$column = $table->getColumn('secret');
-				$column->setLength(256);
+				$column->setLength(512);
```
rullzer, line 52: Changing migrations later on feels iffy

rullzer: Yeah I see the trouble 
But you need indeed another one for people that did already run the migration

apps/oauth2/tests/Db/ClientMapperTest.php
julien-nc: :+1: Done.

julien-nc:
> nit: reorder your commits and push the test only. we should see failing CI. Then you push the fix and we should see green ci.

Test only: running.

julien-nc:
The new test failed before changing the column size: https://github.com/nextcloud/server/actions/runs/5264783429/jobs/9516557792?pr=38770

apps/oauth2/tests/Db/ClientMapperTest.php
ChristophWurst: $this->

julien-nc:
And tests are passing now.

@rullzer So the length has been changed in the previous migration and a new step that changes the length has also been added. This way we make sure:
* people who already ran the `Version011601Date20230522143227` step will have the size changed by the new step (this can only happen by using a git working copy of server)
* people who didn't run `Version011601Date20230522143227` will have the correct column size before encrypting existing values

rullzer (approved):

ChristophWurst (approved):

PR #38780, opened 2023-06-12
ci: try setup-php v2.25.2

blizzz:
Try setup-php v2.25.2 to see whether installing phpunit works.


blizzz:
verbose is unfortunately not very verbose around the PHP tools parts.

PR #38793, opened 2023-06-13
align subadmins css class name

Jenandra:
## Summary

The actually used class name in apps/settings/src/components/UserList/UserRowSimple.vue is subAdminsGroups. Because of this, css for the subadmins groups column in the user list is missing, causing formatting to break when users are admins of many (10+) groups.

See also: https://help.nextcloud.com/t/ansicht-von-user-mit-vielen-gruppen-ist-unubersichtlich/72149/3


artonge (changes requested):
What about
https://github.com/nextcloud/server/blob/c12a64fa526cc6c575eeb3acdf3e8713e853c32f/apps/settings/src/components/UserList.vue#L196
https://github.com/nextcloud/server/blob/c12a64fa526cc6c575eeb3acdf3e8713e853c32f/apps/settings/src/components/UserList/UserRow.vue#L163
?

Don't `.subadmin` target other elements?
Why don't the component declare its own css?

artonge (approved):
Looks good

susnux (approved):
looks good now

JuliaKirschenheuter (approved):

artonge:
Thanks @Jenandra !

PR #38802, opened 2023-06-14
Add types to responses

provokateurin:
From https://github.com/nextcloud/server/pull/36666

## Summary

Sadly the methods have to be overridden in all the subclasses, but the code for that is always mostly the same.

## Checklist

### DIFF
--- a/lib/private/AppFramework/OCS/V1Response.php
+++ b/lib/private/AppFramework/OCS/V1Response.php
@@ -5,6 +5,7 @@
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Joas Schilling <coding@schilljs.com>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -25,8 +26,16 @@
 namespace OC\AppFramework\OCS;
 
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
 
+/**
+ * @psalm-import-type DataResponseType from DataResponse
+ * @template S of int
+ * @template-covariant T of DataResponseType
+ * @template H of array<string, mixed>
+ * @template-extends BaseResponse<int, DataResponseType, array<string, mixed>>
+ */
 class V1Response extends BaseResponse {
 	/**
 	 * The V1 endpoint has very limited http status codes basically everything
--- a/lib/public/AppFramework/Http/DataDisplayResponse.php
+++ b/lib/public/AppFramework/Http/DataDisplayResponse.php
@@ -6,6 +6,7 @@
  * @author Julius Härtl <jus@bitgrid.net>
  * @author Morris Jobke <hey@morrisjobke.de>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license AGPL-3.0
  *
@@ -30,6 +31,9 @@ use OCP\AppFramework\Http;
  * Class DataDisplayResponse
  *
  * @since 8.1.0
+ * @template S of int
+ * @template H of array<string, mixed>
+ * @template-extends Response<int, array<string, mixed>>
  */
 class DataDisplayResponse extends Response {
 	/**
@@ -41,17 +45,14 @@ class DataDisplayResponse extends Response {
 
 	/**
 	 * @param string $data the data to display
-	 * @param int $statusCode the Http status code, defaults to 200
-	 * @param array $headers additional key value based headers
+	 * @param S $statusCode the Http status code, defaults to 200
+	 * @param H $headers additional key value based headers
 	 * @since 8.1.0
 	 */
-	public function __construct($data = '', $statusCode = Http::STATUS_OK,
-								$headers = []) {
-		parent::__construct();
+	public function __construct(string $data = '', int $statusCode = Http::STATUS_OK, array $headers = []) {
+		parent::__construct($statusCode, $headers);
 
 		$this->data = $data;
-		$this->setStatus($statusCode);
-		$this->setHeaders(array_merge($this->getHeaders(), $headers));
 		$this->addHeader('Content-Disposition', 'inline; filename=""');
 	}
 
--- a/lib/public/AppFramework/Http/DataResponse.php
+++ b/lib/public/AppFramework/Http/DataResponse.php
@@ -6,6 +6,7 @@
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Morris Jobke <hey@morrisjobke.de>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license AGPL-3.0
  *
@@ -30,34 +31,37 @@ use OCP\AppFramework\Http;
  * A generic DataResponse class that is used to return generic data responses
  * for responders to transform
  * @since 8.0.0
+ * @psalm-type DataResponseType = array|int|float|string|bool|object|null|\stdClass|\JsonSerializable
+ * @template S of int
+ * @template-covariant T of DataResponseType
+ * @template H of array<string, mixed>
+ * @template-extends Response<int, array<string, mixed>>
  */
 class DataResponse extends Response {
 	/**
 	 * response data
-	 * @var array|int|float|string|bool|object
+	 * @var T
 	 */
 	protected $data;
 
 
 	/**
-	 * @param array|int|float|string|bool|object $data the object or array that should be transformed
-	 * @param int $statusCode the Http status code, defaults to 200
-	 * @param array $headers additional key value based headers
+	 * @param T $data the object or array that should be transformed
+	 * @param S $statusCode the Http status code, defaults to 200
+	 * @param H $headers additional key value based headers
 	 * @since 8.0.0
 	 */
-	public function __construct($data = [], $statusCode = Http::STATUS_OK,
-								array $headers = []) {
-		parent::__construct();
+	public function __construct(mixed $data = [], int $statusCode = Http::STATUS_OK, array $headers = []) {
+		parent::__construct($statusCode, $headers);
 
 		$this->data = $data;
-		$this->setStatus($statusCode);
-		$this->setHeaders(array_merge($this->getHeaders(), $headers));
 	}
 
 
 	/**
 	 * Sets values in the data json array
-	 * @param array|int|float|string|object $data an array or object which will be transformed
+	 * @psalm-suppress InvalidTemplateParam
+	 * @param T $data an array or object which will be transformed
 	 * @return DataResponse Reference to this object
 	 * @since 8.0.0
 	 */
@@ -70,7 +74,7 @@ class DataResponse extends Response {
 
 	/**
 	 * Used to get the set parameters
-	 * @return array|int|float|string|bool|object the data
+	 * @return T the data
 	 * @since 8.0.0
 	 */
 	public function getData() {
--- a/lib/public/AppFramework/Http/FileDisplayResponse.php
+++ b/lib/public/AppFramework/Http/FileDisplayResponse.php
@@ -4,6 +4,7 @@
  *
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -24,31 +25,33 @@
 namespace OCP\AppFramework\Http;
 
 use OCP\AppFramework\Http;
+use OCP\Files\File;
+use OCP\Files\SimpleFS\ISimpleFile;
 
 /**
  * Class FileDisplayResponse
  *
  * @since 11.0.0
+ * @template S of int
+ * @template H of array<string, mixed>
+ * @template-extends Response<int, array<string, mixed>>
  */
 class FileDisplayResponse extends Response implements ICallbackResponse {
-	/** @var \OCP\Files\File|\OCP\Files\SimpleFS\ISimpleFile */
+	/** @var File|ISimpleFile */
 	private $file;
 
 	/**
 	 * FileDisplayResponse constructor.
 	 *
-	 * @param \OCP\Files\File|\OCP\Files\SimpleFS\ISimpleFile $file
-	 * @param int $statusCode
-	 * @param array $headers
+	 * @param File|ISimpleFile $file
+	 * @param S $statusCode
+	 * @param H $headers
 	 * @since 11.0.0
 	 */
-	public function __construct($file, $statusCode = Http::STATUS_OK,
-								$headers = []) {
-		parent::__construct();
+	public function __construct(File|ISimpleFile $file, int $statusCode = Http::STATUS_OK, array $headers = []) {
+		parent::__construct($statusCode, $headers);
 
 		$this->file = $file;
-		$this->setStatus($statusCode);
-		$this->setHeaders(array_merge($this->getHeaders(), $headers));
 		$this->addHeader('Content-Disposition', 'inline; filename="' . rawurldecode($file->getName()) . '"');
 
 		$this->setETag($file->getEtag());
--- a/lib/public/AppFramework/Http/NotFoundResponse.php
+++ b/lib/public/AppFramework/Http/NotFoundResponse.php
@@ -6,6 +6,7 @@
  * @author Lukas Reschke <lukas@statuscode.ch>
  * @author Morris Jobke <hey@morrisjobke.de>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license AGPL-3.0
  *
@@ -24,18 +25,24 @@
  */
 namespace OCP\AppFramework\Http;
 
+use OCP\AppFramework\Http;
+
 /**
  * A generic 404 response showing an 404 error page as well to the end-user
  * @since 8.1.0
+ * @template S of int
+ * @template H of array<string, mixed>
+ * @template-extends TemplateResponse<int, array<string, mixed>>
  */
 class NotFoundResponse extends TemplateResponse {
 	/**
+	 * @param S $status
+	 * @param H $headers
 	 * @since 8.1.0
 	 */
-	public function __construct() {
-		parent::__construct('core', '404', [], 'guest');
+	public function __construct(int $status = Http::STATUS_NOT_FOUND, array $headers = []) {
+		parent::__construct('core', '404', [], 'guest', $status, $headers);
 
 		$this->setContentSecurityPolicy(new ContentSecurityPolicy());
-		$this->setStatus(404);
 	}
 }
--- a/lib/public/AppFramework/Http/RedirectResponse.php
+++ b/lib/public/AppFramework/Http/RedirectResponse.php
@@ -7,6 +7,7 @@
  * @author Roeland Jago Douma <roeland@famdouma.nl>
  * @author Thomas Müller <thomas.mueller@tmit.eu>
  * @author v1r0x <vinzenz.rosenkranz@gmail.com>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license AGPL-3.0
  *
@@ -30,6 +31,9 @@ use OCP\AppFramework\Http;
 /**
  * Redirects to a different URL
  * @since 7.0.0
+ * @template S of int
+ * @template H of array<string, mixed>
+ * @template-extends Response<int, array<string, mixed>>
  */
 class RedirectResponse extends Response {
 	private $redirectURL;
@@ -37,13 +41,14 @@ class RedirectResponse extends Response {
 	/**
 	 * Creates a response that redirects to a url
 	 * @param string $redirectURL the url to redirect to
+	 * @param S $status
+	 * @param H $headers
 	 * @since 7.0.0
 	 */
-	public function __construct($redirectURL) {
-		parent::__construct();
+	public function __construct(string $redirectURL, int $status = Http::STATUS_SEE_OTHER, array $headers = []) {
+		parent::__construct($status, $headers);
 
 		$this->redirectURL = $redirectURL;
-		$this->setStatus(Http::STATUS_SEE_OTHER);
 		$this->addHeader('Location', $redirectURL);
 	}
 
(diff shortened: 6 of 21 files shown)


provokateurin:
The `addHeader` is not being override due to the lack of one feature and the existence of one (might be even two) bug(s) in psalm. The method is not used much in the server code where it would be relevant for the OpenAPI changes. This PR should be merged regardless of this missing piece.

provokateurin:
> I want to remove the `@since` tag because it would be inherited with the `@inheritDoc` tag

I opened https://github.com/nextcloud/server/pull/38810 for that

lib/private/AppFramework/Middleware/OCSMiddleware.php
```diff
@@ -90,7 +91,7 @@ public function afterException($controller, $methodName, \Exception $exception)
 	/**
 	 * @param Controller $controller
 	 * @param string $methodName
-	 * @param Response $response
+	 * @param Response|JSONResponse $response
```
come-nc, line 94: How come this is needed, JSONResponse extends Response no?

apps/settings/lib/WellKnown/ChangePasswordHandler.php
```diff
@@ -45,6 +46,7 @@ public function handle(string $service, IRequestContext $context, ?IResponse $pr
 			return $previousResponse;
 		}
 
+		/** @var Response $response */
 		$response = new RedirectResponse($this->urlGenerator->linkToRouteAbsolute('settings.PersonalSettings.index', ['section' => 'security']));
```
come-nc, line 50: RedirectResponse extends Response, so the `@var` annotation seems useless?

lib/private/AppFramework/OCS/V1Response.php
```diff
@@ -25,13 +26,22 @@
...
+ * @template-covariant T of DataResponseType
+ * @template H of array<string, mixed>
+ * @template-extends BaseResponse<S, DataResponseType, H>
+ */
 class V1Response extends BaseResponse {
 	/**
 	 * The V1 endpoint has very limited http status codes basically everything
 	 * is status 200 except 401
 	 *
+	 * @psalm-suppress ImplementedReturnTypeMismatch Expected because we match some status codes
```
come-nc, line 44: Can you explain this?

lib/private/AppFramework/Middleware/OCSMiddleware.php
provokateurin: Yes, but this is due to typing. Psalm makes a difference between these subtypes. To solve this properly something like `@template-covariant` but for `@param` would be needed.

apps/settings/lib/WellKnown/ChangePasswordHandler.php
provokateurin: Same as above

nickvergessen:
> I think the addHeader method should be deprecated to solve the issue

Deprecation means we can remove it in 3 years. So it won't solve your issue any time soon.

provokateurin:
I know, I will have to add some comments about this in the OpenAPI tutorial/guidelines.

lib/public/AppFramework/Http/DataDisplayResponse.php
```diff
@@ -86,4 +90,30 @@ public function setData($data) {
...
+
+	/**
+	 * @inheritDoc
+	 * @template NewH as array<string, mixed>
+	 * @param NewH $headers value header pairs
+	 * @psalm-this-out self<S, NewH>
+	 * @return $this
+	 * @since 8.0.0
+	 */
+	public function setHeaders(array $headers) {
```
nickvergessen, line 102: Any reason all those duplicates are needed?
Is it the `@psalm-this-out self` part?

provokateurin: It is required to override the method in order to let psalm know that the template types have changed. I wish psalm was smart enough to infer this from the parent class and its template parameters, but sadly it doesn't do that.

ChristophWurst:
Is there any mechanism is place to ensure this doesn't fall apart with future modifications of the code? How does a developer check if they need to annotate changed/new APIs?

provokateurin:
> Is there any mechanism is place to ensure this doesn't fall apart with future modifications of the code?

Do you mean the code of the response classes? Those are validated by psalm

> How does a developer check if they need to annotate changed/new APIs?

Developers don't need to add any new annotations because of these changes.

lib/public/AppFramework/Http/DataDisplayResponse.php
ChristophWurst: Isn't that because ``Response::setHeader`` returns `self` and not `static`?

https://psalm.dev/r/70a344550d vs https://psalm.dev/r/5cf059ceb8

provokateurin: I will try it, I never knew this is possible

nickvergessen: I think static is PHP 8.0 or 8.1+
So can be used in master now

provokateurin: I tried this, but it doesn't work. Afterwards psalm could not detect anymore that I set some headers and just ignored whatever I set as the new headers :disappointed:

lib/private/AppFramework/Middleware/OCSMiddleware.php
provokateurin: Fixed

lib/private/AppFramework/OCS/V1Response.php
provokateurin: Fixed

apps/settings/lib/WellKnown/ChangePasswordHandler.php
provokateurin: Fixed

lib/public/AppFramework/Http/DataDisplayResponse.php
ChristophWurst: Can you isolate the issue in a minimal https://psalm.dev/r example?

I wonder if setHeader is problematic if you don't continue with its result. Only the return value changes its type, not the original object.

provokateurin: sorry i wasn't clear about this:

i set the return type to static and removed the overrides, but then psalm didn't know that i changed the type after calling setHeaders.

it kind of makes sense because it's not about calling the right function or returning the right value, but about telling psalm that the template parameter types have changed. i guess it just doesn't work to update the parent class and expect the change to propagate to the child class.

i can try isolating the problem, but i'm pretty sure it won't work and is not intended to work.

lib/public/AppFramework/Http/Response.php
```diff
@@ -231,7 +236,9 @@ public function addHeader($name, $value) {
 
 	/**
 	 * Set the headers
-	 * @param array $headers value header pairs
+	 * @template NewH as array<string, mixed>
+	 * @param NewH $headers value header pairs
+	 * @psalm-this-out self<S, NewH>
```
come-nc, line 241:
```suggestion
	 * @psalm-this-out static<S, NewH>
```
Using that does not avoid the need for the duplication in subclasses?

lib/public/AppFramework/Http/Response.php
```diff
@@ -201,6 +205,7 @@ public function getCookies() {
 	 * @param string $value The value, null will delete it
 	 * @return $this
 	 * @since 6.0.0 - return value was added in 7.0.0
+	 * @depreacted Use setHeaders instead
```
come-nc, line 208:
```suggestion
	 * @deprecated Use setHeaders instead
```
The same typo is present several times in the PR

lib/public/AppFramework/Http/Response.php
provokateurin: That's actually something I oversaw, maybe this can make it work.

provokateurin: Yep this works! No need to override the methods (except for deprecating setStatus in some places).

lib/public/AppFramework/Http/RedirectResponse.php
```diff
@@ -55,4 +58,14 @@ public function __construct($redirectURL) {
 	public function getRedirectURL() {
 		return $this->redirectURL;
 	}
+
+	/**
+	 * @inheritDoc
+	 * @deprecated Do not use this method. It modifies the status code which you are not supposed to do on a RedirectResponse
```
nickvergessen, line 64:
```suggestion
	 * @internal Do not use this method. It modifies the status code which you are not supposed to do on a RedirectResponse
```

lib/public/AppFramework/Http/TooManyRequestsResponse.php
```diff
@@ -50,4 +54,14 @@ public function render() {
 		$template = new Template('core', '429', 'blank');
 		return $template->fetchPage();
 	}
+
+	/**
+	 * @inheritDoc
+	 * @deprecated Do not use this method. It modifies the status code which you are not supposed to do on a TooManyRequestsResponse
```
nickvergessen, line 60:
```suggestion
	 * @internal Do not use this method. It modifies the status code which you are not supposed to do on a TooManyRequestsResponse
```

nickvergessen (commented):
You mean internal, right? We are not planning to remove the function

kesselb (approved):

lib/public/AppFramework/Http/NotFoundResponse.php
```diff
@@ -36,6 +41,16 @@ public function __construct() {
 		parent::__construct('core', '404', [], 'guest');
 
 		$this->setContentSecurityPolicy(new ContentSecurityPolicy());
-		$this->setStatus(404);
+		$this->setStatus(Http::STATUS_NOT_FOUND);
+	}
+
+	/**
+	 * @inheritDoc
+	 * @deprecated Do not use this method. It modifies the status code which you are not supposed to do on a NotFoundResponse
```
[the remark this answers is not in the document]
provokateurin: https://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.internal.pkg.html I'm not sure if internal actually makes sense. It seems it's only used for having different sets of documentation?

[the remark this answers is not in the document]
nickvergessen: > What implications does marking as internal have?

It crosses it out on IDEs indicating that it shouldn't be used.

If you stay with deprecated the next thing after the keyword should be the version number.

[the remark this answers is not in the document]
provokateurin: But is using internal better than deprecated? Especially when I want to make this method fail in a later version?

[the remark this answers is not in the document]
nickvergessen: We can deprecate it and then remove it in 3 years.
Until then I would additionally mark it internal to indicate that it was never meant to be available.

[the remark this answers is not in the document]
provokateurin: Ok, adding both sounds good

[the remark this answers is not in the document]
ChristophWurst: https://en.wikipedia.org/wiki/Liskov_substitution_principle :wave: 

I don't think overwriting the method and saying it can't be used is a good idea.

If we want to enforce these sematics by syntax there could be an interface `IResponseMutableStatus` with a `::setStatus` that is only implemented by selected responses. Status-immutable responses will simply not have that method.

lib/public/AppFramework/Http/RedirectResponse.php
ChristophWurst, line 64: I'm not sure if this is really true. The redirect could be 301 or 302

provokateurin:
> https://en.wikipedia.org/wiki/Liskov_substitution_principle wave
> 
> I don't think overwriting the method and saying it can't be used is a good idea.
> 
> If we want to enforce these sematics by syntax there could be an interface IResponseMutableStatus with a ::setStatus that is only implemented by selected responses. Status-immutable responses will simply not have that method.

@ChristophWurst for some reason I can't comment below your comment.
i think this is a good idea, but right now i can only deprecate it, right?
and then in 3 years someone can implement this interface to make the setStatus method unavailable.

lib/public/AppFramework/Http/RedirectResponse.php
provokateurin: Right, but I couldn't find any usage anywhere in the Nextcloud org. Should we still allow it?

lib/public/AppFramework/Http/NotFoundResponse.php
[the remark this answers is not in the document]
provokateurin: the interface is a good idea, but right now i can only deprecate the method, no?
in 3 years someone can then implement this interface to make the setStatus method unavailable.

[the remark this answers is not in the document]
provokateurin: the other solution would be to just allow any status for these responses which is fine, but doesn't make a lot of sense for e.g. NotFoundResponse

lib/public/AppFramework/Http/RedirectResponse.php
provokateurin: (Could also solve this by just allowing any status code for any response like suggested above)

lib/public/AppFramework/Http/Response.php
```diff
@@ -201,6 +203,7 @@ public function getCookies() {
 	 * @param string $value The value, null will delete it
 	 * @return $this
 	 * @since 6.0.0 - return value was added in 7.0.0
+	 * @deprecated 28.0.0 Use setHeaders instead
```
nickvergessen, line 206: setHeaders() is quite cumbersome because you first need to get all existing headers, add your own header to the array and then write again.

I found 4 apps (richdocuments, photos, onlineoffice, files) and all don't do that.

Also all the additional checks this method has (trim, status code) are missing there.

Is there a strong need to deprecate this method?

provokateurin: With the current state of psalm it is not possible to annotate this method in a way that lets psalm understand that a new key-value pair has been added to the headers. You can still use this method, but it won't help you with psalm and you might forget some headers. We can keep the method, but I would still advise against using it in the OpenAPI tutorial. That would also be fine for me.

lib/public/AppFramework/Http/NotFoundResponse.php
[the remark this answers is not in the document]
ChristophWurst: Adding the method today makes the migration smoother because you can already write (app) code that only works with responses that implement IResponseMutableStatus.
If you add the interface only in the release that drops the methods for some sub classes it is a hard cut where apps can only support that one server version and no previous releases.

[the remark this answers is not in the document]
provokateurin: I see. I though a bit about it and I will allow setting the status code for any response. This is easier as it doesn't need any deprecation or overrides and there is no documentation about the intention of these responses that I assumed to only have one status code ever, so it won't break something that maybe was intended to work.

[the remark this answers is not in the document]
ChristophWurst: Sounds good!

lib/public/AppFramework/Http/Response.php
provokateurin: Removed deprecation

provokateurin:
I added the status and headers parameters everywhere to let psalm know the default value. Otherwise it turned it into generic int which stopped some type validations and made it possible to set a wrong return type for a controller method.

ChristophWurst (approved):

lib/public/AppFramework/Http/TextPlainResponse.php
```diff
@@ -38,14 +42,14 @@ class TextPlainResponse extends Response {
...
 	 * constructor of TextPlainResponse
 	 * @param string $text The text body
-	 * @param int $statusCode the Http status code, defaults to 200
+	 * @param S $statusCode the Http status code, defaults to 200
+	 * @param H $headers
 	 * @since 22.0.0
 	 */
-	public function __construct(string $text = '', int $statusCode = Http::STATUS_OK) {
-		parent::__construct();
+	public function __construct(string $text = '', int $statusCode = Http::STATUS_OK, $headers = []) {
```
nickvergessen, line 49:
```suggestion
	public function __construct(string $text = '', int $statusCode = Http::STATUS_OK, array $headers = []) {
```

Can we type those new parameters or will that confuse Psalm again?

provokateurin: Fixed

provokateurin: (also for the $status)

lib/public/AppFramework/Http/DataDisplayResponse.php
```diff
@@ -41,17 +45,14 @@ class DataDisplayResponse extends Response {
...
-	 * @param int $statusCode the Http status code, defaults to 200
-	 * @param array $headers additional key value based headers
+	 * @param S $statusCode the Http status code, defaults to 200
+	 * @param H $headers additional key value based headers
 	 * @since 8.1.0
 	 */
-	public function __construct($data = '', $statusCode = Http::STATUS_OK,
-								$headers = []) {
-		parent::__construct();
+	public function __construct($data = '', int $statusCode = Http::STATUS_OK, array $headers = []) {
```
nickvergessen, line 52: Either touching is okay, then:
```suggestion
	public function __construct(string $data = '', int $statusCode = Http::STATUS_OK, array $headers = []) {
```

otherwise since the parameters existed here already:
```suggestion
	public function __construct($data = '', $statusCode = Http::STATUS_OK, $headers = []) {
```

lib/public/AppFramework/Http/DataResponse.php
```diff
@@ -30,34 +31,37 @@
...
-	 * @param array $headers additional key value based headers
+	 * @param T $data the object or array that should be transformed
+	 * @param S $statusCode the Http status code, defaults to 200
+	 * @param H $headers additional key value based headers
 	 * @since 8.0.0
 	 */
-	public function __construct($data = [], $statusCode = Http::STATUS_OK,
-								array $headers = []) {
-		parent::__construct();
+	public function __construct($data = [], int $statusCode = Http::STATUS_OK, array $headers = []) {
```
nickvergessen, line 54: was also changed here

lib/public/AppFramework/Http/FileDisplayResponse.php
```diff
@@ -38,17 +42,14 @@ class FileDisplayResponse extends Response implements ICallbackResponse {
...
-	 * @param int $statusCode
-	 * @param array $headers
+	 * @param S $statusCode
+	 * @param H $headers
 	 * @since 11.0.0
 	 */
-	public function __construct($file, $statusCode = Http::STATUS_OK,
-								$headers = []) {
-		parent::__construct();
+	public function __construct($file, int $statusCode = Http::STATUS_OK, array $headers = []) {
```
nickvergessen, line 49: and here

nickvergessen (commented):
but fine either way

lib/public/AppFramework/Http/DataDisplayResponse.php
provokateurin: Fixed

lib/public/AppFramework/Http/TemplateResponse.php
```diff
@@ -98,11 +105,12 @@ class TemplateResponse extends Response {
...
 	 * template
 	 * @param string $renderAs how the page should be rendered, defaults to user
+	 * @param S $status
+	 * @param H $headers
 	 * @since 6.0.0 - parameters $params and $renderAs were added in 7.0.0
 	 */
-	public function __construct($appName, $templateName, array $params = [],
-								$renderAs = self::RENDER_AS_USER) {
-		parent::__construct();
+	public function __construct(string $appName, string $templateName, array $params = [], $renderAs = self::RENDER_AS_USER, int $status = Http::STATUS_OK, array $headers = []) {
```
nickvergessen, line 112:
```suggestion
	public function __construct(string $appName, string $templateName, array $params = [], string $renderAs = self::RENDER_AS_USER, int $status = Http::STATUS_OK, array $headers = []) {
```

nickvergessen (approved):
Missed one last type, but fine by me

lib/public/AppFramework/Http/Response.php
```diff
@@ -91,15 +92,19 @@ class Response {
...
-	public function __construct() {
+	public function __construct(int $status = Http::STATUS_OK, array $headers = []) {
 		/** @var IRequest $request */
 		/**
 		 * @psalm-suppress UndefinedClass
 		 */
 		$request = \OC::$server->get(IRequest::class);
 		$this->addHeader("X-Request-Id", $request->getId());
+		$this->setStatus($status);
+		$this->setHeaders($headers);
```
provokateurin, line 107: This overrides the X-Request-Id header

lib/public/AppFramework/Http/Response.php
```diff
@@ -244,21 +252,21 @@ public function setHeaders(array $headers) {
...
 	/**
 	 * Returns the set headers
-	 * @return array the headers
+	 * @return array{Cache-Control: string, Content-Security-Policy: string, Feature-Policy: string, X-Robots-Tag: string, Last-Modified?: string, ETag?: string, ...H} the headers
 	 * @since 6.0.0
 	 */
 	public function getHeaders() {
-		$mergeWith = [];
+		$mergeWith = [
+			'Cache-Control' => 'no-cache, no-store, must-revalidate',
```
come-nc, line 260: You added Cache-control header here? Is it refactor from somewhere else?

provokateurin: It was previously the initialization value. See the top of the file

come-nc (approved):

provokateurin:
:tada:

ChristophWurst:
Mail's Psalm shows that **all** responses have to be type-annotated now due to https://github.com/vimeo/psalm/issues/6870#issuecomment-1334918110. Example: https://github.com/nextcloud/mail/blob/b7b84f452ac6951d0d2e6d96e33b9eec91ac6098/lib/Http/AttachmentDownloadResponse.php https://psalm.dev/r/92dc4c7871

```
ERROR: MissingTemplateParam - lib/Http/AttachmentDownloadResponse.php:29:7 - OCA\Mail\Http\AttachmentDownloadResponse has missing template params when extending OCP\AppFramework\Http\DownloadResponse, expecting 3 (see https://psalm.dev/182)
class AttachmentDownloadResponse extends DownloadResponse {
```

It means those responses have to specify a type for the template (for static headers and status code) https://psalm.dev/r/68242bdab6.

If I spec the type for server master I'll get an error for stable27 and older: https://psalm.dev/r/7038ef2ed1

Could there be any trick to allow responses to have types but with a default?

provokateurin:
For subclasses you either need to implement the template parameters which should be quite easy or you just supress the error. Implementing the template parameters should be quite easy and afterwards you don't need to add the parameters when you use the type (like in return annotations), you can just leave them empty and it will use the "default" parameters.

ChristophWurst:
If I spec the type for server master I'll get an error for stable27 and older: https://psalm.dev/r/7038ef2ed1

provokateurin:
Well, then suppress until you support only 28+?

PR #38808, opened 2023-06-14
Fix issues where unencrypted_size was being falsely used for non-encrypted home folders

juliusknorr:
This fixes issues that occur when encryption is enabled but the encryption of home folders is disabled on object storage. 

Before this change the unencrypted_size might have been set but not always updated (e.g. when writing to a file through fopen). This lead to issues later on when trying to download the file but the size returned from the filecache didn't match with what the storage returns. 

#34232 didn't help here since the scan command it uses is not available for object storage.

Steps to reproduce
- Enable encryption
- Disable home folder encryption
- Upload a docx file
- Open the docx file in Collabora, change something and save (this uses fopen)

Before:
- The unencrypted_size column wasn't updated and the file was no longer downloadable

After:
- The file is downloadable even if the unencrypted size has a wrong data, since the file isn't encrypted we should just stick to the regular size column
- The unencrypted_size column will be properly updated by the ObjectStoreStorage already, for cases when encryption is enabled the Encryption Wrapper will anyways update the value afterwards

## Checklist

### DIFF
--- a/lib/private/Files/Cache/CacheEntry.php
+++ b/lib/private/Files/Cache/CacheEntry.php
@@ -134,7 +134,7 @@ class CacheEntry implements ICacheEntry {
 	}
 
 	public function getUnencryptedSize(): int {
-		if (isset($this->data['unencrypted_size']) && $this->data['unencrypted_size'] > 0) {
+		if ($this->data['encrypted'] && isset($this->data['unencrypted_size']) && $this->data['unencrypted_size'] > 0) {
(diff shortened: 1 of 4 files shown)


juliushaertl:
🏓 For reviews ... confirmed to work on the affected instance already

artonge (approved):

icewind1991 (approved):

artonge (approved):

PR #38810, opened 2023-06-14
Don't required since tag in OCP for inherited docs

provokateurin:
## Summary

This allows to override methods and classes without having to specify the since tag again. It will be inherited from the parent method or class because the whole doc is inherited. The parent method or class doesn't need to be checked when the inherited doc is encountered, because it will already get checked by the same checker.

## Checklist

### DIFF
--- a/build/psalm/OcpSinceChecker.php
+++ b/build/psalm/OcpSinceChecker.php
@@ -68,7 +68,7 @@ class OcpSinceChecker implements Psalm\Plugin\EventHandler\AfterClassLikeVisitIn
 			return;
 		}
 
-		if (!isset($parsedDocblock->tags['since'])) {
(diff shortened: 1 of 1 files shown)


kesselb (approved):
Good change :+1:

ChristophWurst:
> This allows to override methods and classes without having to specify the since tag again.

Do you have an example?

provokateurin:
@ChristophWurst https://github.com/nextcloud/server/pull/38802/commits/8b1d84916c1602090bdbbf86ea43555f3bc3150b

ChristophWurst:
Shouldn't it be `{@inheritDoc}`? https://docs.phpdoc.org/guide/guides/inheritance.html

provokateurin:
The `{@` and `}` parts are automatically removed, but the casing of the tags is not touched. A lot of methods in OCP use the lowercase `inheritdoc` which I assume to work. Might be worth fixing separately

provokateurin:
> https://docs.phpdoc.org/guide/guides/inheritance.html

Oh that page says the since tag is not inherited :/
Well then this PR is useless.

PR #38823, opened 2023-06-14
generate user themed favicon and touchicon

szaimen:
The problem was that the color was not taken into account by the logic. So as soon as the icon is created and put into place in the theming directory, it gets never re-created by changing the color. My PR now takes the color of the icon into account by putting it into the file name. The same logic is alreaddy applied and funcitoning here: https://github.com/nextcloud/server/blob/7afcc44827333683dc104cc9701963151d22946a/apps/theming/lib/Controller/IconController.php#L89-L97

### DIFF
--- a/apps/theming/lib/Controller/IconController.php
+++ b/apps/theming/lib/Controller/IconController.php
@@ -120,14 +120,15 @@ class IconController extends Controller {
 		} catch (NotFoundException $e) {
 		}
 		if ($iconFile === null && $this->imageManager->shouldReplaceIcons()) {
+			$color = $this->themingDefaults->getColorPrimary();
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Could you describe why it was broken before and how your changes fix the problem? I am not familiar with this area of the server code

juliusknorr (approved):

szaimen:
> Could you describe why it was broken before and how your changes fix the problem? I am not familiar with this area of the server code

I adjusted the description

ChristophWurst (approved):
Makes sense

szaimen:
Node failure unrelated

Issue #37091, opened 2023-03-07
[BITV] 9.3.3.2/8.3 - All the form fields in the "New user" dialog lack visual labels. Additionally, the required fields are not marked as such. (1)


AndyScherzinger:
See also https://webaim.org/articles/label-name/

- [x] https://github.com/nextcloud/server/pull/38826

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-9f92015c98-v8-n3

</details>

PR #35898, opened 2022-12-27
Fix add user modal

szaimen:
Supersedes https://github.com/nextcloud/server/pull/33977

PR #38826, opened 2023-06-15
enh(a11y): New user modal

Pytal:
- For https://github.com/nextcloud/server/issues/37091
- Supersede https://github.com/nextcloud/server/pull/35898

## Summary

| | Screenshot |
--- | ---
**Before** | image
**After** | image

Before | After
--- | ---
image | image

## Checklist


artonge (approved):
Fancy.
Not tested, and probably need a double check as there is an eslint error.

szaimen (approved):
Tested and seems to work but didnt review the code. 

Looks much better now! :)

nimishavijay (dismissed):
Nice! Great improvement :) Some points regarding the "One of either password or email..." message, no other feedback, looks great :)

- [ ] We could show it as a subline under both the password and email fields to avoid confusion? Then the "(required)" next to the labels can also be removed. This way if the form is submitted without either of those fields filled, it's easy to indicate the error as the field already has the sublines for info. What do you think? cc @nextcloud/designers
- [ ] If we are not doing that, the message should be left aligned with the rest of the labels
- [ ] Would the wording "Either password or email is required" be better? @jancborchardt and @Pytal

marcoambrosini (approved):

szaimen:
@Pytal drone failure apparently related

Pytal:
> Nice! Great improvement :) Some points regarding the "One of either password or email..." message, no other feedback, looks great :)
> 
> * [ ]  We could show it as a subline under both the password and email fields to avoid confusion? Then the "(required)" next to the labels can also be removed. This way if the form is submitted without either of those fields filled, it's easy to indicate the error as the field already has the sublines for info. What do you think? cc @nextcloud/designers

Duplicating the text as sublines seems strange 🤔

For invalid input the native form validation should suffice
image

> * [x]  If we are not doing that, the message should be left aligned with the rest of the labels
> * [x]  Would the wording "Either password or email is required" be better? @jancborchardt and @Pytal

JuliaKirschenheuter (approved):
Tested ;) Works, code looks good

apps/settings/src/components/Users/NewUserModal.vue
```diff
@@ -0,0 +1,461 @@
...
+	computed: {
+		settings() {
+			return this.$store.getters.getServerData
+		},
+
+		usernameLabel() {
+			if (this.settings.newUserGenerateUserID) {
+				return t('settings', 'Username will be autogenerated')
+			}
+			return t('settings', 'Username (required)')
```
szaimen, line 238: Small suggestion (not sure if we use that somewhere but would possibly make sense to make clear that it cannot be changed afterwards?):

```suggestion
			return t('settings', 'User id (required)')
```

apps/settings/src/components/Users/NewUserModal.vue
```diff
@@ -0,0 +1,461 @@
...
+				:label-visible="true"
+				autocapitalize="none"
+				autocomplete="off"
+				autocorrect="off"
+				pattern="[a-zA-Z0-9 _\.@\-']+"
+				required />
+			<NcTextField class="modal__item"
+				data-test="displayName"
+				:value.sync="newUser.displayName"
+				:label="t('settings', 'Display name')"
```
szaimen, line 47: Maybe?

```suggestion
				:label="t('settings', 'Display name of user')"
```

apps/settings/src/components/Users/NewUserModal.vue
Pytal: @nimishavijay?

apps/settings/src/components/Users/NewUserModal.vue
Pytal: @nimishavijay?

Pytal:
What do you think about @szaimen's wording suggestions for follow-up @jancborchardt?

PR #38839, opened 2023-06-15
Replace plain input fields with NcTextField fields and NcMultiSelect …

### LINKED ISSUE #36976: [BITV] 9.1.4.11/8.2 - The frames of the input fields do not meet the contrast requirements for graphical user interface elements. (1)
1a8091c4ec3723022bc422c053334fda

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-3dbfb7615a-v8-n2

</details>

### LINKED ISSUE #38145: [BITV]  Replace all `NcMultiselect` with `NcSelect` for UserRow
This is improvement. 

Replace all `NcMultiselect` with `NcSelect` for https://github.com/nextcloud/server/blob/master/apps/settings/src/components/UserList/UserRow.vue#L137

JuliaKirschenheuter:
Replace plain input fields with NcTextField fields and NcMultiSelect fields with NcSelect fields

* Resolves: https://github.com/nextcloud/server/issues/38145
* Resolves: https://github.com/nextcloud/server/issues/36976

## Summary

Before: 
old1

After:
new

Also added notifications after `displaty name`, `password` and `email` were changed.

## TODO
- [x] **Make possible to create a new group for availableGroups**
- [x] Create item limitation for subAdminsGroups and availableGroups NcSelect (`:limit` prop on `NcSelect` is just not working)

## Checklist


JuliaKirschenheuter:
@Pytal @susnux @artonge 
how can i resolve this todo point: Make possible to create a new group for availableGroups?

JuliaKirschenheuter:
@jancborchardt @nimishavijay
is it ok design-wise?

Pytal:
> @Pytal @susnux @artonge how can i resolve this todo point: Make possible to create a new group for availableGroups? 

I believe the name needs to be destructured with something like 👇
```js
async createGroup({ name: gid }) {
```

> and this one: Create item limitation for subAdminsGroups and availableGroups NcSelect (:limit prop on NcSelect is just not working)?

Since limit for NcSelect applies to dropdown options rather than selected options, as seen in `vue-multiselect`, you may try https://vue-select.org/guide/selectable.html#limiting-the-number-of-selections

JuliaKirschenheuter:
> I believe the name needs to be destructured with something like
async createGroup({ name: gid }) 

Thank you @Pytal! Group will be added, but i still have an error
`{"ocs":{"meta":{"status":"failure","statuscode":101,"message":""},"data":[]}}`
How could i fix it?

JuliaKirschenheuter:
> Since limit for NcSelect applies to dropdown options rather than selected options, as seen in `vue-multiselect`, you may try https://vue-select.org/guide/selectable.html#limiting-the-number-of-selections

Thanks! I've tried it before too, but used wrong attr: options instead of value. It works now, but we probably should do something with styles (but not in this PR):
Screenshot from 2023-06-16 08-40-13

JuliaKirschenheuter:
Regarding adding a new group there are several problems:

- `option:selected` returns array of all selected groups. Each new selected group - is a last element of an array.
- `option:created` doesn't allow to use a result like a value of new option. Based on this event, it allows creating a new group on the server, but it won't be added to the list of group options from the server itself. Just `{ name }` will be added.
- `create-option` is used for creating new options but that works only for sync input of each symbol. It is not possible to use it for async adding a group on the server.
- After `option:created` has been fired an `option:selected` event will follow. But the result of `option:selected` would have a value like `{ name }` even if we would put manually to event `option:created` a list of a groups.

Thanks @ShGKme for investigation!

But still have no idea how to fix that problem ;(

ShGKme:
Some notes:

**1 - `option:selected` returns array of all selected groups**

Not a big problem, just a difference with `NcSelect`.

We can either call `addUserGroup` with the last element in the template: `@option:selected="addUserGroup($event.at(-1))"`...
... or update `addUserGroup`/create a new method to handle an array.

**2 - `option:created` doesn't allow to use a result like a value of new option**

> But it won't be added to the list of group options from the server itself

Sorry, I was a bit not accurate. I meant, the interval selected list (selected groups) in `NcSelect` has an internal `{ name: <inputted> }` value.

**Combination of 2 + 3 + 4 - not support of async tag creation**

The problem is that `@option:selected` and its handler `addUserGroup` is triggered also for a new group before it was created on the server with `NcSelect`'s internal `{ name: <inputted> }` value.

---

So if don't use `v-model`, all we need is to ignore this being created value in the `option:selected` handler.

The simplest solution is to check if it is a real group object, aka
```js
async addUserGroup(group) {
  if (group.id === undefined) {
    // This is NcSelect's internal value for a new inputted group name
    // Ignore
    return
  }
  // Add the user to the group
}
```

This solution should work but looks dirty and implicit. I'd explicitly mark the being inputted option as temp/being created using `create-option`.

```html
<NcSelect
  :create-option="(value) => ({ name: value, isCreating: true })"
  @option:selected="addGroup($event.at(-1))"
/>
```

```js
async addUserGroup(group) {
  if (group.isCreating) {
    // This is NcSelect's internal value for a new inputted group name
    // Ignore
    return
  }
  // Add the user to the group
}
```

An alternative solution could be: to drop using `option:created` and instead of ignorance create the group in the `option:selected` handler. But the core idea is the same.

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +96,137 @@
...
+				:show-trailing-button="true"
+				class="user-row-text-field"
+				:class="{'icon-loading-small': loading.mailAddress}"
 				:disabled="loading.mailAddress||loading.all"
-				:placeholder="t('settings', 'Add new email address')"
-				:value="user.email"
-				autocapitalize="off"
-				autocomplete="new-password"
-				autocorrect="off"
-				spellcheck="false"
```
artonge, line 135: Not possible to keep those attributes?

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -73,23 +73,20 @@
...
-				@submit.prevent="updateDisplayName">
-				<label class="hidden-visually" :for="'displayName'+user.id+rand">{{ t('settings', 'Edit display name') }}</label>
-				<input :id="'displayName'+user.id+rand"
-					ref="displayName"
-					:disabled="loading.displayName||loading.all"
-					:value="user.displayname"
-					autocapitalize="off"
-					autocomplete="off"
-					autocorrect="off"
-					spellcheck="false"
```
artonge, line 87: Not possible to keep those attributes?

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +96,137 @@
...
 		</div>
 		<div v-if="subAdminsGroups.length>0 && settings.isAdmin"
 			:class="{'icon-loading-small': loading.subadmins}"
 			class="subadmins">
 			<label class="hidden-visually" :for="'subadmins'+user.id+rand">{{ t('settings', 'Set user as admin for') }}</label>
-			<NcMultiselect :id="'subadmins'+user.id+rand"
+			<NcSelect :id="'subadmins'+user.id+rand"
 				:close-on-select="false"
 				:disabled="loading.subadmins||loading.all"
-				:limit="2"
```
artonge, line 168: Was it not useful?

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +96,137 @@
...
+				class="select-vue"
+				@search="searchUserManager"
+				@option:selected="updateUserManager"
+				@input="updateUserManager" />
 		</div>
 
-		<!-- don't show this on edit mode -->
-		<div v-if="showConfig.showStoragePath || showConfig.showUserBackend"
-			class="storageLocation" />
-		<div v-if="showConfig.showLastLogin" />
```
artonge, line 237: Not used?

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -444,14 +460,21 @@ export default {
...
-				value: this.currentManager ? this.currentManager.id : '',
-			}).then(() => {
+			try {
+				this.$store.dispatch('setUserData', {
+					userid: this.user.id,
+					key: 'manager',
+					value: this.currentManager ? this.currentManager.id : '',
+				})
+			} catch (error) {
+				console.error(error)
```
artonge, line 474: Should we tell the user that the update failed?

nimishavijay:
@JuliaKirschenheuter it looks great! Much better now :) 

Some minor enhancements:
- [ ] decrease the width of the "Quota", "Language", and "Manager" columns (by around 3-4px each), and increase the width of the email column (by around 12-14px)
- [ ] Show the read only items even in edit mode instead of hiding them. They can be shown in the same way as they are shown in view mode, so just as plain text.

They can also be in a follow up PR! The changes here look good! :)

apps/settings/src/components/UserList/UserRow.vue
JuliaKirschenheuter: was replaced via `:selectable="() => userSubAdminsGroups.length < 2"`

apps/settings/src/components/UserList/UserRow.vue
JuliaKirschenheuter: i think that was wrong implemented. I've corrected manager column in edit mode that it is now in a right column.

apps/settings/src/components/UserList/UserRow.vue
JuliaKirschenheuter: you mean besides console.error (which doesn't tell to user anything)?

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -73,23 +73,24 @@
...
+				:value.sync="editedDisplayName"
+				autocapitalize="off"
+				autocomplete="off"
+				autocorrect="off"
+				spellcheck="false"
+				type="text"
+				@trailing-button-click="updateDisplayName">
+				<template #trailing-button-icon>
+					<ArrowRight :size="20" />
+				</template>
```
Pytal, line 92:
```suggestion
```
No need to duplicate as we have
```vue
trailing-button-icon="arrowRight"
```
above

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +100,146 @@
...
 				required
 				spellcheck="false"
 				type="password"
-				value="">
-			<input class="icon-confirm" type="submit" value="">
-		</form>
+				@trailing-button-click="updatePassword">
+				<template #trailing-button-icon>
+					<ArrowRight :size="20" />
+				</template>
```
Pytal, line 124:
```suggestion
```

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +100,146 @@
...
 				autocorrect="off"
 				spellcheck="false"
-				type="email">
-			<input class="icon-confirm" type="submit" value="">
-		</form>
+				type="email"
+				@trailing-button-click="updateEmail">
+				<template #trailing-button-icon>
+					<ArrowRight :size="20" />
+				</template>
```
Pytal, line 147:
```suggestion
```

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +100,146 @@
...
-				class="multiselect-vue"
-				label="name"
-				track-by="id"
-				@remove="removeUserSubAdmin"
-				@select="addUserSubAdmin">
-				<span slot="noResult">{{ t('settings', 'No results') }}</span>
-			</NcMultiselect>
+				class="select-vue"
+				@option:deselected="removeUserSubAdmin"
+				@option:selected="addUserSubAdmin($event.at(-1))" />
```
Pytal, line 185:
```suggestion
				@option:selected="options => addUserSubAdmin(options.at(-1))" />
```
Seems a bit obtuse, let's make it more explicit

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +100,146 @@
...
-				@remove="removeUserGroup"
-				@select="addUserGroup"
-				@tag="createGroup">
-				<span slot="noResult">{{ t('settings', 'No results') }}</span>
-			</NcMultiselect>
+				:no-wrap="true"
+				:selectable="() => userGroups.length < 2"
+				:create-option="(value) => ({ name: value, isCreating: true })"
+				@option:created="createGroup"
+				@option:selected="addUserGroup($event.at(-1))"
```
Pytal, line 166:
```suggestion
				@option:selected="options => addUserGroup(options.at(-1))"
```
Seems a bit obtuse, let's make it more explicit

Pytal (approved):

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -444,14 +462,22 @@ export default {
 		},
 
 		updateUserManager(manager) {
+			if (manager === null) {
+				this.currentManager = ''
+			}
```
Fenn-CS, line 467:
```suggestion
		updateUserManager(manager = null) {
			this.currentManager = ''
		
```

nfebe (approved):

apps/settings/src/components/UserList/UserRow.vue
```diff
@@ -99,151 +97,140 @@
...
+		<div class="mailAddress">
 			<label class="hidden-visually" :for="'mailAddress'+user.id+rand">{{ t('settings', 'Add new email address') }}</label>
-			<input :id="'mailAddress'+user.id+rand"
-				ref="mailAddress"
+			<NcTextField :id="'mailAddress'+user.id+rand"
+				:show-trailing-button="true"
+				class="user-row-text-field"
+				:class="{'icon-loading-small': loading.mailAddress}"
 				:disabled="loading.mailAddress||loading.all"
-				:placeholder="t('settings', 'Add new email address')"
```
artonge, line 130: No more placeholder?

artonge (approved):
Nitpick, but good otherwise.

apps/settings/src/components/Users/UserRow.vue
```diff
@@ -755,4 +812,19 @@ export default {
 	.row::v-deep .multiselect__single {
```
susnux, line 812: This rule is not required anymore, as you replaced all `NcMultiselect` components.

apps/settings/src/components/Users/UserRow.vue
```diff
@@ -351,6 +346,9 @@ export default {
 				wipe: false,
 				manager: false,
 			},
+			editedDisplayName: this.user.displayname,
+			editedPassword: '',
+			editedMail: this.user.email,
```
susnux, line 351:
```suggestion
			editedMail: this.user.email ?? '',
```

Ensure the email is not null, to prevent this:
image

susnux (approved):
:+1:

artonge (commented):
This does not look right, is this known?
Probably a matter of `z-index` somewhere.

Screenshot from 2023-06-27 10-39-51

susnux:
Drone error *is* related because it looks for the multiselect instead of the select, so the acceptance tests need to be adjusted, this needs to be ported to NcSelect:

https://github.com/nextcloud/server/blob/4cfab4b838ed40dec200f7673992009896c69f16/tests/acceptance/features/bootstrap/UsersSettingsContext.php#L155

https://github.com/nextcloud/server/blob/4cfab4b838ed40dec200f7673992009896c69f16/tests/acceptance/features/bootstrap/UsersSettingsContext.php#L120

JuliaKirschenheuter:
> This does not look right, is this known?

yes, it is known. I've tried to fix that problem here https://github.com/nextcloud/server/pull/37870, but still had problems with a table itself and had to revert this PR. But that is known and will be fixed.

susnux:
> This does not look right, is this known? Probably a matter of `z-index` somewhere.

I guess this rule:
https://github.com/nextcloud/server/blob/cc1498b225bf1d70bd81854f8941f333fcd61209/apps/settings/src/components/Users/UserRow.vue#L807

JuliaKirschenheuter:
> > This does not look right, is this known? Probably a matter of `z-index` somewhere.
> 
> I guess this rule:
> 
> https://github.com/nextcloud/server/blob/cc1498b225bf1d70bd81854f8941f333fcd61209/apps/settings/src/components/Users/UserRow.vue#L807

no, unfortunately not. Please look in master, there is a same problem.

JuliaKirschenheuter:
@susnux, @artonge could you please help me to fix integration tests?

tests/acceptance/features/bootstrap/UsersSettingsContext.php
```diff
@@ -151,7 +152,7 @@ public static function theColumn($column) {
 	 * @return Locator
 	 */
 	public static function selectedSelectOption($cell, $user) {
-		return Locator::forThe()->css(".vs--single .name-parts__first")->
+		return Locator::forThe()->css(".multiselect__single")->
```
JuliaKirschenheuter: but i have no multiselect anymore. why should i take it?

tests/acceptance/features/bootstrap/UsersSettingsContext.php
```diff
@@ -117,7 +117,8 @@ public static function displayNameCellForUser($user) {
 	 * @return Locator
 	 */
 	public static function optionInInputForUser($cell, $user) {
-		return Locator::forThe()->css(".vs__dropdown-option--highlight")->
+		return Locator::forThe()->css(".multiselect__option--highlight")->
```
JuliaKirschenheuter: but i have no multiselect anymore. why should i take it?

tests/acceptance/features/bootstrap/UsersSettingsContext.php
susnux: Sorry rebase gone wrong, should be fixed again.

susnux:
There is still the issue with the password change, I verified it works locally.
I think the issue is that it does not wait for the password to change:
> The loading icon for user user0 was not found after 1 seconds, assumming that it was shown and hidden again before the check started and continuing

Maybe we should switch to cypress, as it seems to be easier to adjust.

Pytal (approved):

PR #38852, opened 2023-06-16
Move deprecation warnings to debug

szaimen:
Same as https://github.com/nextcloud/server/pull/37816 but for js


szaimen (commented):
Alternatively we might want to check for window.oc_debug and only show the warning in that case. WDYT @skjnldsv @ChristophWurst ?

ChristophWurst:
Yes, showing this only to devs would be preferred

szaimen:
> Yes, showing this only to devs would be preferred

@ChristophWurst so which version shall we choose? the current one by moving the warnings to debug or checking for window_oc_debug and then printing a warning?

ChristophWurst:
Debug logs and debug mode only

susnux (approved):

szaimen:
@ChristophWurst addressed

szaimen:
> And there will be no difference between important deprecation warning messages from the libs and verbose debug logs for app/lib in terms of log levels.

that is what I was wondering about as well. So my suggestion would be to show these as warnings if debug mode is enabled. WDYT?

ShGKme:
> that is what I was wondering about as well. So my suggestion would be to show these as warnings if debug mode is enabled. WDYT?

I'd either keep `console.warn` for deprecation warnings or make something more high-level.

szaimen:
> I'd either keep `console.warn` for deprecation warnings or make something more high-level.

please check out https://github.com/nextcloud/server/pull/38852/commits/6dd433a8f6b8166131ca004ccebaaf99e2595cee

ShGKme (approved):

skjnldsv (approved):

szaimen:
https://github.com/nextcloud/server/pull/38980

PR #38854, opened 2023-06-16
Introduce LanguageModel/TextProcessing OCP API

### LINKED ISSUE #38059: Localhosted language model (like llama.cpp)
I don't like that you have to pay openai and send them the data to use the new AI smart picker tools.

I think it'd be great if we could use localhosted models like llama.cpp.

### LINKED ISSUE #38578: Add API to summarize text
### How to use GitHub

As a Nextcloud developer I want to be able to summarize text for the users of the app. E.g. to get the gist of an email thread.

Provide an OCP API and optionally an OCS API where an app can send long text and receive a short summary text. The API has to be optional because not every installation will have a backend for language processing.

N/a

**Additional context**

cc @DaphneMuller

marcelklehr:
* Resolves: #38578 
* Closes: #38059

## Summary
~~This adds a LanguageModel namespace with an `ILanguageModelManager` that queries registered `ILanguageModelProvider`s to either synchronously run or asynchronously schedule  `ILanguageModelTask`s. When scheduled asynchronously, the tasks will be run in a background job and either a `TaskFailedEvent` or a `TaskSuccessfulEvent` is emitted with error message or the output, respectively. Currently implemented are the following task types: `SummarizeTask`, `HeadlineTask`, `TopicsTask` and `FreePromptTask`.~~

In coordination with #julien-nc  we reworked this PR to merge it with another upcoming API. Now, this adds a TextProcessing namespace with an `IManager` that queries registered `IProvider`s to either synchronously run or asynchronously schedule  `Task`s. When scheduled asynchronously, the tasks will be run in a background job and either a `TaskFailedEvent` or a `TaskSuccessfulEvent` is emitted with error message or the output, respectively. Currently implemented are the following task types: `SummaryTaskType`, `HeadlineTaskType`, `TopicsTaskType` and `FreePromptTaskType`. Apps can register providers for existing task types in core or use their own task types, which we may pull into core at some point, if we want. 

## TODO

- [x] ocs API with a polling endpoint to query the result
  - [x] db needs an output column and ILanguageModelTask probably also needs getters and setters for that 
- [x] make linters happy
- [x] automatically expire tasks in db

## Checklist


julien-nc (commented):
Noice!

> * [ ] Setting default provider per task type

Any idea on how this would be done? We could start and avoid thinking about making this appear in the UI. How would the information be set and stored? occ command, OCS endpoint? Stored in a dedicated table? Consumed by the manager's runTask method so it would really reflect the current value?

lib/private/LanguageModel/LanguageModelManager.php
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	public function canHandleTask(ILanguageModelTask $task): bool {
+		return !empty(array_filter($this->getAvailableTasks(), fn ($class) => $task instanceof $class));
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function runTask(ILanguageModelTask $task): string {
```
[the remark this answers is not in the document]
marcelklehr: Consumers can call `runTask` to run it synchronously or `scheduleTask` to run it asynchronously.

> Some tasks might be short enough to run synchronously.

Genereally I'd only recommend calling runTask in background tasks.

> How would the consumers deal with asynchronicity? The UI would poll until the task has run?

Yes. `scheduleTask` sets the task id which allows to poll using `getTask` of the Manager or the Controller.

marcelklehr:
>  > Setting default provider per task type
>
> Any idea on how this would be done?

I think it would make sense to do this in a new PR together with the settings for translations and STT.

lib/public/AppFramework/Bootstrap/IRegistrationContext.php
```diff
@@ -219,6 +220,16 @@ public function registerWellKnownHandler(string $class): void;
 	 */
 	public function registerSpeechToTextProvider(string $providerClass): void;
 
+	/**
+	 * Register a custom LanguageModel provider class that provides a promptable language model
+	 * through the OCP\LanguageModel APIs
+	 *
+	 * @param string $providerClass
+	 * @psalm-param class-string<ILanguageModelProvider> $providerClass
+	 * @since 27.0.0
```
kesselb, line 229:
```suggestion
	 * @since 28.0.0
```

lib/private/LanguageModel/LanguageModelManager.php
```diff
@@ -0,0 +1,169 @@
...
+
+		return $this->providers;
+	}
+
+	public function hasProviders(): bool {
+		$context = $this->coordinator->getRegistrationContext();
+		if ($context === null) {
+			return false;
+		}
+		return !empty($context->getSpeechToTextProviders());
```
kesselb, line 71:
```suggestion
		return count($context->getSpeechToTextProviders()) > 0;
```

kesselb: or `$context->getSpeechToTextProviders() !== []`;

lib/private/LanguageModel/LanguageModelManager.php
```diff
@@ -0,0 +1,169 @@
...
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getAvailableTaskTypes(): array {
+		return array_map(fn ($taskClass) => $taskClass::TYPE, $this->getAvailableTasks());
+	}
+
+	public function canHandleTask(ILanguageModelTask $task): bool {
+		return !empty(array_filter($this->getAvailableTasks(), fn ($class) => $task instanceof $class));
```
kesselb, line 96:
```suggestion
		return count(array_filter($this->getAvailableTasks(), fn ($class) => $task instanceof $class)) > 0;
```

kesselb:
Looks nice, thank you!

I've suggested replacing !empty with count() > 0. Backstory is https://localheinz.com/articles/2023/05/10/avoiding-empty-in-php/#content-array which advises against using empty, when the code is properly typed.

lib/public/LanguageModel/ILanguageModelManager.php
```diff
@@ -0,0 +1,77 @@
...
+	 * @throws PreConditionNotMetException If no or not the requested provider was registered but this method was still called
+	 * @since 28.0.0
+	 */
+	public function scheduleTask(ILanguageModelTask $task) : void;
+
+	/**
+	 * @param int $id The id of the task
+	 * @return ILanguageModelTask
+	 * @throws RuntimeException If the query failed
+	 * @throws \ValueError If the task could not be found
```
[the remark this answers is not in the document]
marcelklehr: yeah, didn't want to create a new class. Do you have a better fitting class at hand? Of should I just create a new class?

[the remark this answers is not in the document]
ChristophWurst: I've seen other cases where a generic *NotFoundException* could be handy. Right now we only have one for DB operations. Using that one here feels wrong. We could add a generic one to OCP, like `\OCP\Common\Exception\NotFoundException`

lib/public/LanguageModel/TopicsTask.php
```diff
@@ -0,0 +1,42 @@
...
+	public const TYPE = 'topics';
+
+	/**
+	 * @inheritDoc
+	 * @since 28.0.0
+	 */
+	public function visitProvider(ILanguageModelProvider $provider): string {
+		if (!$provider instanceof ITopicsProvider) {
+			throw new \RuntimeException('SummaryTask#visitProvider expects IHeadlineProvider');
+		}
```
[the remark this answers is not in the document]
marcelklehr: Good idea! I managed to fix it :)

lib/public/LanguageModel/FreePromptTask.php
```diff
@@ -15,15 +38,15 @@ final class FreePromptTask extends AbstractLanguageModelTask {
 	 * @inheritDoc
 	 * @since 28.0.0
 	 */
-	public function visitProvider(ILanguageModelProvider $provider): string {
```
ChristophWurst: nit: you can leave the native type in addition so PHP does a runtime check

marcelklehr: I think I tried that but psalm wouldn't have it

lib/public/LanguageModel/ILanguageModelProvider.php
```diff
@@ -0,0 +1,47 @@
...
+	 */
+	public function getName(): string;
+
+	/**
+	 * @param string $prompt The prompt to call the model with
+	 * @return string the output
+	 * @since 28.0.0
+	 * @throws RuntimeException If the text could not be transcribed
+	 */
+	public function prompt(string $prompt): string;
```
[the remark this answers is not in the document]
ChristophWurst: > There might be tasks in the future that require more params, but we can add those as normal params then

That will be a breaking change then because implementers of the OCP API have to adjust. If you use an object you can add additional methods without adjusting implementers of an interface.

We used the same trick for the bootstrap context for example.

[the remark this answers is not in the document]
marcelklehr: > That will be a breaking change then

Mh, I don't think it would, as new tasks will get a new interface, like ISummaryProvider here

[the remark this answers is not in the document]
marcelklehr: So, I think we might add new tasks (as new interfaces) but probably not new params for existing tasks.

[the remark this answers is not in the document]
ChristophWurst: Alright

lib/private/LanguageModel/Db/TaskMapper.php
```diff
@@ -30,4 +31,21 @@ public function find(int $id): Task {
...
+	 */
+	public function deleteOlderThan(int $timeout): int {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->tableName)
+			->where($qb->expr()->lt('last_updated', $qb->createPositionalParameter(time() - $timeout)));
+		return $qb->executeStatement();
+	}
+
+	public function update(Entity $entity): Entity {
+		$entity->setLastUpdated(time());
```
ChristophWurst, line 48: not testable (reminds me that this PR doesn't have unit tests)

inject the time factory

lib/public/LanguageModel/FreePromptTask.php
marcelklehr: Ah, I have to use psalm-param. got it.

marcelklehr:
I've never written tests for server before, so please bear with me. I hope this first commit goes in the right direction :)

tests/lib/LanguageModel/LanguageModelManagerTest.php
```diff
@@ -0,0 +1,269 @@
...
+
+class LanguageModelManagerTest extends \Test\TestCase {
+	private ILanguageModelManager $languageModelManager;
+	private Coordinator $coordinator;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->languageModelManager = new LanguageModelManager(
+			\OC::$server->get(IServerContainer::class),
```
ChristophWurst, line 86: inject mocks please

marcelklehr:
Tests are in :)

lib/public/LanguageModel/ILanguageModelManager.php
```diff
@@ -0,0 +1,80 @@
...
+	/**
+	 * @throws PreConditionNotMetException If no or not the requested provider was registered but this method was still called
+	 * @throws RuntimeException If something else failed
+	 * @since 28.0.0
+	 */
+	public function runTask(ILanguageModelTask $task): string;
+
+	/**
+	 * Will schedule an LLM inference process in the background. The result will become available
+	 * with the \OCP\LanguageModel\Events\TaskFinishedEvent
```
ChristophWurst, line 65: how can an app associate the task dispatched here when `TaskSuccessfulEvent` is emitted?

ChristophWurst: specifically for https://github.com/nextcloud/mail/issues/8508 I need to know which user and which thread has been summarized. Would it make sense if apps could pass an identifier when the task is scheduled that is exposed as getter on the event?

ChristophWurst:
> I've never written tests for server before

Are you familiar with the concept of unit tests, integration tests and their differences? Asking because right now the tests do *a lot* and are not isolated and therefore probably not the fastest. In my opinion there should be full coverage with real unit tests and integration tests for selected paths that are most important.

lib/public/LanguageModel/ILanguageModelManager.php
marcelklehr: In the current state this method sets the task ID on the task and people can set userId and appId on the task object which will be persisted.

> Would it make sense if apps could pass an identifier when the task is scheduled that is exposed as getter on the event?

We can do that, yeah. What were you thinking? A string? Or an int? Or something more complex?

marcelklehr:
Hey :)

> Are you familiar with the concept of unit tests, integration tests and their differences?

I'm somewhat familiar with unit and integration tests. I tend to not buy into unit tests and rather go with integration tests, as I feel unit tests have a low return on investment.

> Asking because right now the tests do a lot and are not isolated and therefore probably not the fastest.

I did mock most if not all of the unrelated classes. Currently it takes 6s for me.

> In my opinion there should be full coverage with real unit tests and integration tests for selected paths that are most important.

I will get to writing unit tests, then :) Can we finalize the API first, then? That would avoid blocking you for implementing the downstream stuff and me for the docs.

ChristophWurst:
Done right, unit tests will make you structure code into easier chunks and they can test all the nifty edge cases without the heavy setup cost.

> I did mock most if not all of the unrelated classes. Currently it takes 6s for me.

That's a lot IMO. The server has thousands of tests. If every feature is tested like this we make test execution unreasonable for local changes.

Mail's full unit test suitefor comparison:

```
PHPUnit 9.5.28 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.7
Configuration: tests/phpunit.unit.xml

...............................................................  63 / 595 ( 10%)
............................................................... 126 / 595 ( 21%)
............................................................... 189 / 595 ( 31%)
............................................................... 252 / 595 ( 42%)
............................................................... 315 / 595 ( 52%)
............................................................... 378 / 595 ( 63%)
............................................................... 441 / 595 ( 74%)
............................................................... 504 / 595 ( 84%)
............................................................... 567 / 595 ( 95%)
............................                                    595 / 595 (100%)

Time: 00:01.936, Memory: 26.00 MB

OK (595 tests, 1537 assertions)

real    0m2,958s
user    0m2,653s
sys     0m0,218s
```

But let's not get sidetracked. I'm not directly responsible for this repo so I'll leave the decisions to the server devs.

I'll put my focus back on the usability of the API for https://github.com/nextcloud/mail/issues/8508 :+1:

lib/public/LanguageModel/ILanguageModelManager.php
ChristophWurst: I think a string would be most flexible. Apps can come up with their own format for it, be it serialized json or something else.

lib/public/LanguageModel/ILanguageModelTask.php
```diff
@@ -0,0 +1,140 @@
...
+	 * @return string
+	 * @since 28.0.0
+	 */
+	public function getType(): string;
+
+	/**
+	 * @return int
+	 * @since 28.0.0
+	 */
+	public function getStatus(): int;
```
ChristophWurst, line 91:
```suggestion
	 * @return ILanguageModelTask::STATUS_*
	 * @since 28.0.0
	 */
	public function getStatus(): int;
```

lib/public/LanguageModel/ILanguageModelTask.php
```diff
@@ -0,0 +1,140 @@
...
+
+	/**
+	 * @param int $status
```
ChristophWurst, line 94:
```suggestion
	 * @param ILanguageModelTask::STATUS_* $status
```

ChristophWurst (commented):
:+1: Looks usable for the Mail thread summarization once we have a generic identification property to map successful summaries back to their corresponding threads. App ID and user ID is not sufficient.

core/Controller/LanguageModelApiController.php
```diff
@@ -0,0 +1,97 @@
...
+			]);
+		} catch (PreConditionNotMetException) {
+			return new DataResponse(['message' => $this->l->t('Necessary language model provider is not available')], Http::STATUS_PRECONDITION_FAILED);
+		}
+	}
+
+	/**
+	 * @PublicPage
+	 */
+	public function getTask(int $id): DataResponse {
```
marcelklehr, line 84: I just realized this may not be ideal from a security perspective. There is no authentication in any way here so people can just guess an id and leak private data :/

ChristophWurst: Drop `@PublicPage` and compare the task's user ID with the current user ID?

marcelklehr: The idea was for guests to also be able to call this API, though. For this we'd need PublicPage. However, any guest could then request any other guest's tasks. :/

lib/private/LanguageModel/Db/Task.php
```diff
@@ -0,0 +1,74 @@
...
+		return Task::fromParams([
+			'id' => $task->getId(),
+			'type' => $task->getType(),
+			'lastUpdated' => time(),
+			'status' => $task->getStatus(),
+			'input' => $task->getInput(),
+			'output' => $task->getOutput(),
+			'userId' => $task->getUserId(),
+			'appId' => $task->getAppId(),
+		]);
```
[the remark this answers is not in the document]
marcelklehr: Not sure how I can avoid this

core/Controller/LanguageModelApiController.php
ChristophWurst: With guest you mean anonymous users, not a user generated by https://github.com/nextcloud/guests/, right? Because the latter does have a session and is logged in.

marcelklehr: Right, yes.

ChristophWurst: You could add a flag for tasks and results that are considered public and only allow access to those

marcelklehr: Mh, thinking about this more, it's probably unlikely that anonymous users will schedule tasks with private data

marcelklehr: I added the userId === task->userId check now, I think this should be good enough.

lib/private/LanguageModel/Db/Task.php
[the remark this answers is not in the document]
come-nc: Not clear to me. What is the return type of Task::fromParams seen by psalm?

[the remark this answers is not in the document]
marcelklehr: OCP\AppFramework\Db\Entity::fromParams declares `static` as its return type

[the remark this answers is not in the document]
come-nc: I’m lost then, Task::fromParams has to return a Task. Not sure what bugs psalm.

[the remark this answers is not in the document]
ChristophWurst: This could be a Psalm bug https://psalm.dev/r/f5530d3538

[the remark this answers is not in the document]
ChristophWurst: https://psalm.dev/r/6f8dde5356 this is very strange. Note how `Task::fromParams` results in different types for $t1 and $temp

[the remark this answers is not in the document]
ChristophWurst: You can help Psalm by saving the new task into a variable and spec'ing a type: https://psalm.dev/r/e2837a5e3e

marcelklehr:
I added the 'identifier' prop

lib/private/LanguageModel/Db/Task.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
```
come-nc, line 2: copyright and strict_types missing, same for TaskMapper

lib/private/LanguageModel/LanguageModelManager.php
```diff
@@ -0,0 +1,185 @@
...
+		if ($context === null) {
+			return false;
+		}
+		return count($context->getLanguageModelProviders()) > 0;
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getAvailableTasks(): array {
```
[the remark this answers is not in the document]
come-nc: Or 
```suggestion
	public function getAvailableTaskClasses(): array {
```
?

lib/private/LanguageModel/LanguageModelManager.php
```diff
@@ -0,0 +1,185 @@
...
+
+	/**
+	 * @inheritDoc
+	 */
+	public function getAvailableTaskTypes(): array {
+		return array_map(fn ($taskClass) => $taskClass::TYPE, $this->getAvailableTasks());
+	}
+
+	public function canHandleTask(ILanguageModelTask $task): bool {
+		return count(array_filter($this->getAvailableTasks(), fn ($class) => $task instanceof $class)) > 0;
```
come-nc, line 107: Hard to read and suboptimal, use a foreach and stop at first true?

lib/private/LanguageModel/TaskBackgroundJob.php
```diff
@@ -0,0 +1,62 @@
...
+use OCP\PreConditionNotMetException;
+
+class TaskBackgroundJob extends QueuedJob {
+	public function __construct(
+		ITimeFactory $timeFactory,
+		private ILanguageModelManager $languageModelManager,
+		private IEventDispatcher $eventDispatcher,
+	) {
+		parent::__construct($timeFactory);
+		$this->setAllowParallelRuns(false);
```
come-nc, line 44: Maybe add a comment to explain why

lib/public/Common/Exception/NotFoundException.php
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * @since 28.0.0
+ */
+class NotFoundException extends \Exception {
+	/**
+	 * Constructor
+	 * @param string $msg the error message
+	 * @since 28.0.0
+	 */
+	public function __construct($msg) {
```
come-nc, line 38:
```suggestion
	public function __construct(string $msg) {
```

lib/public/LanguageModel/AbstractLanguageModelTask.php
```diff
@@ -0,0 +1,194 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OCP\LanguageModel;
+
+use OC\LanguageModel\Db\Task;
```
come-nc, line 28: Nope. Cannot use `OC` from `OCP`.

lib/public/LanguageModel/Events/TaskFailedEvent.php
```diff
@@ -0,0 +1,28 @@
...
+ * @since 28.0.0
+ */
+class TaskFailedEvent extends AbstractLanguageModelEvent {
+	/**
+	 * @param ILanguageModelTask $task
+	 * @param string $errorMessage
+	 * @since 28.0.0
+	 */
+	public function __construct(ILanguageModelTask $task,
+		private string $errorMessage) {
```
come-nc, line 17:
```suggestion
	public function __construct(
		ILanguageModelTask $task,
		private string $errorMessage,
	) {
```

lib/public/LanguageModel/ILanguageModelManager.php
come-nc, line 65: There is no `\OCP\LanguageModel\Events\TaskFinishedEvent`

lib/public/LanguageModel/ILanguageModelProvider.php
```diff
@@ -0,0 +1,49 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+namespace OCP\LanguageModel;
+
+use RuntimeException;
+
+/**
+ * This is the minimum interface is implemented by apps that
```
come-nc, line 32:
```suggestion
 * This is the minimum interface that is implemented by apps that
```
or "which"?

tests/lib/LanguageModel/LanguageModelManagerTest.php
```diff
@@ -0,0 +1,376 @@
...
+		$task2 = $this->languageModelManager->getTask($task->getId());
+		$this->assertEquals($task->getId(), $task2->getId());
+		$this->assertEquals('Hello', $task2->getInput());
+		$this->assertNull($task2->getOutput());
+		$this->assertEquals(ILanguageModelTask::STATUS_SCHEDULED, $task2->getStatus());
+
+		/** @var IEventDispatcher $this->eventDispatcher */
+		$this->eventDispatcher = \OC::$server->get(IEventDispatcher::class);
+		$successfulEventFired = false;
+		$this->eventDispatcher->addListener(TaskSuccessfulEvent::class, function (TaskSuccessfulEvent $event) use (&$successfulEventFired, $task) {
```
come-nc, line 220: It would be better to mock the event listener and expect the method call instead.

tests/lib/LanguageModel/LanguageModelManagerTest.php
```diff
@@ -0,0 +1,376 @@
...
+		// Try free prompt again
+		$this->assertEquals('Hello Free Prompt', $this->languageModelManager->runTask(new FreePromptTask('Hello', 'test', null)));
+
+		// Try headline task
+		$this->assertEquals('Hello Headline', $this->languageModelManager->runTask(new HeadlineTask('Hello', 'test', null)));
+
+		// Try summary task
+		$this->assertEquals('Hello Summarize', $this->languageModelManager->runTask(new SummaryTask('Hello', 'test', null)));
+
+		// Topics are not implemented by both the vanilla provider and the full provider
```
come-nc, line 275: Then why is it called TestFullLanguageModelProvider if it is not full :-P

come-nc (changes requested):

core/Controller/LanguageModelApiController.php
```diff
@@ -0,0 +1,101 @@
...
+		return new DataResponse([
+			'tasks' => $this->languageModelManager->getAvailableTaskTypes(),
+		]);
+	}
+
+	/**
+	 * @PublicPage
+	 * @UserRateThrottle(limit=20, period=120)
+	 * @AnonRateThrottle(limit=5, period=120)
+	 */
```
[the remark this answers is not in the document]
provokateurin: It's optional for now

core/Controller/LanguageModelApiController.php
```diff
@@ -0,0 +1,101 @@
...
+	public function schedule(string $text, string $type, ?string $appId, string $identifier = ''): DataResponse {
```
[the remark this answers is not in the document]
marcelklehr: Yes

core/Controller/LanguageModelApiController.php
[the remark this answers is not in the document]
marcelklehr: Done

lib/private/LanguageModel/LanguageModelManager.php
```diff
@@ -0,0 +1,185 @@
...
+	 * @inheritDoc
+	 */
+	public function runTask(ILanguageModelTask $task): string {
+		if (!$this->canHandleTask($task)) {
+			throw new PreConditionNotMetException('No LanguageModel provider is installed that can handle this task');
+		}
+		foreach ($this->getProviders() as $provider) {
+			if (!$task->canUseProvider($provider)) {
+				continue;
+			}
```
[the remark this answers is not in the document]
marcelklehr: A follow up PR will add admin settings for all AI APIs to set a preference

core/Controller/LanguageModelApiController.php
```diff
@@ -0,0 +1,128 @@
...
+		}
+	}
+
+	/**
+	 * This endpoint allows checking the status and results of a task.
+	 * Tasks are removed 1 week after receiving their last update.
+	 *
+	 * @PublicPage
+	 * @param int $id The id of the task
+	 * @return DataResponse<Htpp::STATUS_NOT_FOUND|Http::STATUS_INTERNAL_SERVER_ERROR, array{message:string}>|DataResponse<Http::STATUS_OK,array{task: array{id: int,type:string,status: ILanguageModelTask::STATUS_*,userId: string, appId: string,input: string,output: string, identifier: string}},array{}>
```
provokateurin, line 105:
```suggestion
	 * @return DataResponse<Htpp::STATUS_NOT_FOUND|Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}>|DataResponse<Http::STATUS_OK, array{task: array{id: int, type: string, status: ILanguageModelTask::STATUS_*, userId: string, appId: string, input: string, output: string, identifier: string}}, array{}>
```

Please use spaces then it will get more readable. Also applies to the other controller methods.

core/Controller/LanguageModelApiController.php
```diff
@@ -0,0 +1,128 @@
...
+	 * This endpoint allows scheduling a language model task
+	 *
+	 * @PublicPage
+	 * @UserRateThrottle(limit=20, period=120)
+	 * @AnonRateThrottle(limit=5, period=120)
+	 * @param string $input The input for the language model task
+	 * @param string $type The task type
+	 * @param string $appId The originating app ID
+	 * @param string $identifier An identifier to identify this task
+	 * @return DataResponse<Http::STATUS_OK,array{task: array{id: int,type:string,status: ILanguageModelTask::STATUS_*,userId: string, appId: string,input: string,output: string, identifier: string}},array{}>|DataResponse<Http::STATUS_PRECONDITION_FAILED,array{message: string}, array{}>
```
provokateurin, line 76: `ILanguageModelTask::STATUS_*` is not going to work with openapi-extractor, you will have to type it as int. Same for the controller method below.

provokateurin (changes requested):

marcelklehr:
Tests with --exclude-group DB,SLOWDB on master currently take 01:00.693 on my machine, while this branch takes 01:00.377 :shrug:

marcelklehr:
Should I add a capability for this OCS API? There's no docs on how this works, sadly.

core/Controller/LanguageModelApiController.php
```diff
@@ -0,0 +1,126 @@
...
+		try {
+			$task = $this->languageModelManager->getTask($id);
+
+			if ($this->userId !== $task->getUserId()) {
+				return new DataResponse(['message' => $this->l->t('Task not found')], Http::STATUS_NOT_FOUND);
+			}
+
+			return new DataResponse([
+				'task' => $task->jsonSerialize(),
+			]);
```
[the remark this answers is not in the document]
provokateurin: @marcelklehr I think you are just missing the return type annotation on the jsonSerialize method of the AbstractLanguageModelTask

core/Controller/LanguageModelApiController.php
```diff
@@ -0,0 +1,126 @@
...
+	 * This endpoint allows scheduling a language model task
+	 *
+	 * @PublicPage
+	 * @UserRateThrottle(limit=20, period=120)
+	 * @AnonRateThrottle(limit=5, period=120)
+	 * @param string $input The input for the language model task
+	 * @param string $type The task type
+	 * @param string $appId The originating app ID
+	 * @param string $identifier An identifier to identify this task
+	 * @return DataResponse<Http::STATUS_OK,array{task: array{id: int, type: string, status: int, userId: string, appId: string, input: string, output: string, identifier: string}},array{}> | DataResponse<Http::STATUS_PRECONDITION_FAILED | Http::STATUS_BAD_REQUEST, array{message: string}, array{}>
```
provokateurin, line 74:
```suggestion
	 * @return DataResponse<Http::STATUS_OK, array{task: array{id: int, type: string, status: int, userId: string, appId: string, input: string, output: string, identifier: string}}, array{}>| DataResponse<Http::STATUS_PRECONDITION_FAILED|Http::STATUS_BAD_REQUEST, array{message: string}, array{}>
```

For the union operator I wouldn't use spaces, just after every comma

provokateurin (changes requested):

core/Controller/LanguageModelApiController.php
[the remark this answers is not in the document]
marcelklehr: $task is a ILanguageModelTask here, though and that method comes from the Json serialize interface :/

core/Controller/LanguageModelApiController.php
```diff
@@ -86,8 +86,11 @@ public function schedule(string $input, string $type, string $appId, string $ide
 		try {
 			$this->languageModelManager->scheduleTask($task);
 
+			/** @var array{id: int|null, type: string, status: int, userId: string|null, appId: string, input: string, output: string|null, identifier: string} $json */
```
provokateurin, line 89: This annotation has to be on the method itself and not here where it is used

core/Controller/LanguageModelApiController.php
```diff
@@ -114,8 +117,11 @@ public function getTask(int $id): DataResponse {
 				return new DataResponse(['message' => $this->l->t('Task not found')], Http::STATUS_NOT_FOUND);
 			}
 
+			/** @var array{id: int|null, type: string, status: int, userId: string|null, appId: string, input: string, output: string|null, identifier: string} $json */
+			$json = $task->jsonSerialize();
```
provokateurin, line 121: Same again with the annotation

provokateurin (dismissed):

core/Controller/LanguageModelApiController.php
marcelklehr: The method is not from me, though

marcelklehr: Works fine for me

marcelklehr: This is where the method comes from: https://github.com/nextcloud/server/pull/38854/files#diff-4fcc874a025d38b83e66c558753978cb0f31ddc369330fc6fa03bd3721eace4bR32

provokateurin: You just need to set the annotation on the class that implements it (which should be https://github.com/nextcloud/server/pull/38854/files#diff-aa82def10581f49d1d2fc84749d32c130bab07f186a728ed92c8686875535d24R149).

marcelklehr: Setting a \@method annotation on AbstractLanguageModelTask will not work, because that's not the type I'm using. I'm using ILanguageModelTask. Setting an \@method annotation on that interface has no effects.

provokateurin: Psalm knows what implementations exist and it correctly infers the type from the implementation even if you always give it the interface. I know it works because I've already done it a few times in the server repo.

marcelklehr: > Psalm knows what implementations exist and it correctly infers the type from the implementation even if you always give it the interface.

That sounds like a dangerous feature.

I tried both 

```
@psalm-method array{id: int|null, type: string, status: int, userId: string|null, appId: string, input: string, output: string|null, identifier: string} jsonSerialize()
```

on the AbstractLanguageModel class and 

```
@return array{id: int|null, type: string, status: int, userId: string|null, appId: string, input: string, output: string|null, identifier: string}
``` 

on the method implementation. It infers mixed in both cases as the result of jsonSerialize().

provokateurin: Hm, feel free to just remove all the API annotations, I will fix this later

lib/private/TextProcessing/Manager.php
```diff
@@ -0,0 +1,182 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+namespace OC\TextProcessing;
+
+use OC\AppFramework\Bootstrap\Coordinator;
+use OC\TextProcessing\Db\Task as DbTask;
+use \OCP\TextProcessing\Task as OCPTask;
```
come-nc, line 30:
```suggestion
use OCP\TextProcessing\Task as OCPTask;
```

lib/public/TextProcessing/Events/TaskSuccessfulEvent.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+class TaskSuccessfulEvent extends AbstractTextProcessingEvent {
+	/**
+	 * @param Task $task
+	 * @since 27.1.0
+	 */
+	public function __construct(Task $task) {
+		parent::__construct($task);
+	}
+}
```
come-nc, line 18:
```suggestion
class TaskSuccessfulEvent extends AbstractTextProcessingEvent {
}
```

This is the same, no?

come-nc (approved):

lib/public/TextProcessing/Task.php
```diff
@@ -0,0 +1,238 @@
...
+	 * @param string $appId
+	 * @param string $identifier
+	 * @return Task
+	 * @throws \InvalidArgumentException
+	 * @since 27.1.0
+	 */
+	final public static function factory(string $type, string $input, ?string $userId, string $appId, string $identifier = ''): Task {
+		if (!in_array($type, self::TYPES)) {
+			throw new \InvalidArgumentException('Unknown task type');
+		}
```
julien-nc, line 235: This will prevent scheduling app-defined task types. Maybe this could rather check if the type is in `OC\TextProcessing\Manager::getAvailableTaskTypes()`.

lib/private/TextProcessing/Manager.php
```diff
@@ -0,0 +1,182 @@
...
+		if (!$this->canHandleTask($task)) {
+			throw new PreConditionNotMetException('No LanguageModel provider is installed that can handle this task');
+		}
+		$task->setStatus(OCPTask::STATUS_SCHEDULED);
+		$taskEntity = DbTask::fromPublicTask($task);
+		$this->taskMapper->insert($taskEntity);
+		$task->setId($taskEntity->getId());
+		$this->jobList->add(TaskBackgroundJob::class, [
+			'taskId' => $task->getId()
+		]);
```
julien-nc, line 161: Best case scenario, the job runs next time cron.php is executed, right?
If so, maybe we could think about allowing providers to say they will be fast and run the task immediately. All good for now, just thinking ahead.

julien-nc (changes requested):
Nice!

core/Controller/TextProcessingApiController.php
```diff
@@ -0,0 +1,137 @@
...
+	) {
+		parent::__construct($appName, $request);
+	}
+
+	/**
+	 * This endpoint returns all available LanguageModel task types
+	 *
+	 * @PublicPage
+	 */
+	public function taskTypes(): DataResponse {
```
julien-nc, line 62: As discussed, maybe it makes sense to have another endpoint to list the providers (maybe grouped by task type).

Also, maybe we could include the "selected provider" for each task type in the result of this `taskTypes` method.

WDYT?

marcelklehr: Do we really need the providers on the frontend though? I think we would usually want to hide them, no?

marcelklehr: > the "selected provider" for each task type

I tend to think this should be hidden from the user. Only the admin should see this.

> maybe it's useful for the settings part where admins can choose the "default" provider for each task type

I'll deal with that in the follow up PR :)

julien-nc (approved):
:tada:

marcelklehr:
Docs: https://github.com/nextcloud/documentation/pull/10714

AndyScherzinger:
@marcelklehr backport to 27 needed?

marcelklehr:
@AndyScherzinger Yes! If we backport now, this is gonna land in 27.0.2, though.

AndyScherzinger:
Ah, right 👍

Good point 🫣

lib/public/Common/Exception/NotFoundException.php
```diff
@@ -0,0 +1,41 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program. If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+namespace OCP\Common\Exception;
```
nickvergessen, line 25: This is quite an anti pattern and we really try hard to get rid of the `OC_Util` and `OCP\Util`, so let's not start the next thing we can never change because it is used by multiple mechanisms in different ways.

Can we deprecate this (or even revert since it was never released apart from beta) and instead add a dedicated TaskNotFound exception inside the text processing namespace?

lib/private/TextProcessing/Manager.php
```diff
@@ -0,0 +1,182 @@
...
+	 * @return OCPTask
+	 * @throws RuntimeException If the query failed
+	 * @throws NotFoundException If the task could not be found
+	 */
+	public function getTask(int $id): OCPTask {
+		try {
+			$taskEntity = $this->taskMapper->find($id);
+			return $taskEntity->toPublicTask();
+		} catch (DoesNotExistException $e) {
+			throw new NotFoundException('Could not find task with the provided id');
```
nickvergessen, line 175: As per other comment, either (re)throw `DoesNotExistException` which is already a generic exception with the same meaning, or a dedicated TaskNotFound

lib/public/Common/Exception/NotFoundException.php
marcelklehr: It was suggested to me by other reviewers :shrug: 
I'm on vacation until end of August. Feel free to change this in my absence.

nickvergessen: Just read up on the comments... alright then :(

And enjoy your hols! 😎

PR #38882, opened 2023-06-19
More comprehensive wording on devices' wipe action button

Jerome-Herbinet:
Replace "Wipe all devices" with "Disconnect all devices and delete local data".
IMO, of course, it's longer, but the "Wipe all devices" wording lacks clarity on the action incurred.
So this new wording might be a better compromise.

## Checklist


szaimen (approved):
LGTM

nimishavijay (changes requested):
Good idea to improve the wording! :) I am worried that "Disconnect all devices and delete local data" is quite long. How about "Delete data from all devices"? It provides more info than "Wipe all devices" (and probably translates better as well) and there is a confirmation modal to explain further anyway. What do you think? @nextcloud/designers

Jerome-Herbinet:
> Good idea to improve the wording! :) I am worried that "Disconnect all devices and delete local data" is quite long. How about "Delete data from all devices"? It provides more info than "Wipe all devices" (and probably translates better as well) and there is a confirmation modal to explain further anyway. What do you think? @nextcloud/designers

@nimishavijay thanks for your comment :-)
Your suggestion "Delete data from all devices" is longer than "Wipe all devices" but does not contain some information : 
- it does not talk about disconnection (IMO, this half of the feature is very important to talk about)
- it does not talk about the "local" or "synced" side of data
So the problem remains the same, even if better words are used, but with a longer wording ;-)
Instead of "Disconnect all devices and delete local data", a little shorter wording would be "Disconnect devices and wipe local data". 
After this, I suggest improving the modal's description because it's not talking about disconnection. This could be done by myself with your suggestions in this same PR. 
I would also (still in this PR) improve the wording for apps' passwords UI in the "Security" section ("Wipe device").
What do you think ?

nimishavijay:
Absolutely agreed about the modal! How about the item in the action menu is short and the modal has more info, like this: 

Action menu item:
- [ ] Change wording to "Delete data from all devices"
- [ ] Use the icon [phonelink-erase](https://fonts.google.com/icons?selected=Material%20Icons%3Aphonelink_erase%3A)
- [ ] Change the order of items in the action menu to better reflect the risk of the actions
  - [ ] Resend welcome email
  - [ ] Impersonate
  - [ ] Disable user
  - [ ] Delete data from all devices
  - [ ] Delete user

Modal:

```
Disconnect all devices and delete local data

In case of lost devices or exiting the organization, this will remotely wipe the Nextcloud data from all devices associated with [username] and disconnect the devices. Only works if the devices are connected to the internet.

[Cancel] [Wipe devices]

```

Things to note:
- [ ] Wording change in the heading of the modal
- [ ] Wording change in the description
- [ ] Wording change in the button (removed the username as the original text was `Wipe [username]'s devices`, and using an apostrophe ('s) may not translate well to other languages)

What do you think? Does this sufficiently communicate the result of the action? :)

nfebe (approved):
Another common way to say this for example in operating systems and hard disk manipulation software they often use "Erase device(s)" (Most times with a text that gives more details and of course a design that shows how dangerous the action is)

PR #38892, opened 2023-06-19
[stable26] use source cache when listing folder during recursive copy

juliusknorr:
backport of #38890


come-nc (approved):

AndyScherzinger (approved):

juliushaertl:
Can't tell why the node test doesn't report a status, but doesn't seem related at all.

juliushaertl:
Node test pending is unrelated as node-when-unrelated.yml is missing on stable26.

PR #38900, opened 2023-06-19
Add getDocBaseUrl on theming app

tanganellilore:
* Resolves: not opened for this type of PR

## Summary
With new theming introduced on nextcloud >25, we can customize DocBaseUrl only if we use "old fashion way", adding `defaults.php` under `/themes/<custom_theme>/ `.

With this PR I simply added a `getDocBaseUrl()` function, that will get theming `docBaseUrl` config and will use for doc base url.
It will fallback on the defaullt one if not found, in the same way as other function.

## TODO

N/A

## Checklist

### DIFF
--- a/apps/theming/lib/ThemingDefaults.php
+++ b/apps/theming/lib/ThemingDefaults.php
@@ -71,6 +71,7 @@ class ThemingDefaults extends \OC_Defaults {
 	private string $productName;
 	private string $url;
 	private string $color;
+	private string $docBaseUrl;
 
 	private string $iTunesAppId;
 	private string $iOSClientUrl;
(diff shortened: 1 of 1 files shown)


tanganellilore:
fixed psalm

tanganellilore:
@icewind1991 @ArtificialOwl @Fenn-CS  Any chance to check this?

nfebe (commented):
If  this config would live in `config.php`, then it should also be in the `config.sample.php`.

Otherwise, I don't know how much we need this so tagging @come-nc

apps/theming/lib/ThemingDefaults.php
```diff
@@ -163,6 +165,10 @@ public function getPrivacyUrl() {
 		return (string)$this->config->getAppValue('theming', 'privacyUrl', '');
 	}
 
+	public function getDocBaseUrl() {
```
come-nc, line 168: I do not understand where is this method used?

skjnldsv (approved):

apps/theming/lib/ThemingDefaults.php
skjnldsv: @come-nc it's just extending from `OC_Defaults`

come-nc (approved):

Issue #34637, opened 2022-10-17
[Bug]: Contact stop showing under contact menu when unchecking allow autocompletion when entering full name or email address 


AndyXheli:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Hello so i noticed once i deselected the highlighted below user stop showing under the Contact search. The while goal was to select Allow username autocompletion to users within the same group to only show users that that are in the same group for example if i have some external users. I don't want my external users to see internal users this part dose work but the contacts search menu dose display anything. If do select the below highlighted then they appear back

![image](https://user-images.githubusercontent.com/59488153/196203744-a3233040-76bc-42eb-afa7-a0692583b58c.png)

![image](https://user-images.githubusercontent.com/59488153/196203911-a261776b-c839-49d5-b75e-3b69d48f161c.png)

### Steps to reproduce

1. Settings
2. Admin
3.Sharing
4.Select Allow username autocompletion to users within the same group 
5. deselect Allow autocompletion when entering the full name or email address (ingoing missing phonebook match and being in the same group 
6. Check Contact search  
7. all gone 

### Expected behavior

Should show users under the same group. 

### Installation method

Community Manual installation with Archive

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### List of activated Apps

```shell
NA
```

szaimen:
Hi, please update to 24.0.9 or better 25.0.3 and report back if it fixes the issue. Thank you!

[...]

PR #38907, opened 2023-06-20
Fix: ContactManager search with fullmatch

JohannesGGE:
* Contributes to: https://github.com/nextcloud/mail/pull/8551

## Summary

If I understand the sharing options in the admin settings right, the following should lead to the suggestion of a SAB user in the same group, and don't show other SAB users even if there is a full match (name, email).

image

Currently, because of the 'fullmatch' option, I pass as false in mail, the if condition is triggered and no SAB user is suggested at all. To work correctly I think there should be the check for the access to the SAB as well.

Maybe I am wrong here, correct me if that's the case :)

## Checklist

### DIFF
--- a/lib/private/ContactsManager.php
+++ b/lib/private/ContactsManager.php
@@ -58,15 +58,18 @@ class ContactsManager implements IManager {
 			$strictSearch = array_key_exists('strict_search', $options) && $options['strict_search'] === true;
 
 			if ($addressBook->isSystemAddressBook()) {
+				$enumeration = !\array_key_exists('enumeration', $options) || $options['enumeration'] !== false;
 				$fullMatch = !\array_key_exists('fullmatch', $options) || $options['fullmatch'] !== false;
-				if (!$fullMatch) {
-					// Neither full match is allowed, so skip the system address book
+
+				if (!$enumeration && !$fullMatch) {
+					// No access to system address book AND no full match allowed
 					continue;
 				}
+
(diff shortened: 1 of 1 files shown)


ChristophWurst (commented):
That looks sane

ChristophWurst (approved):
Works as expected in combination with https://github.com/nextcloud/mail/pull/8551

lib/private/ContactsManager.php
```diff
@@ -58,15 +58,18 @@ public function search($pattern, $searchProperties = [], $options = []) {
...
 
 			if ($addressBook->isSystemAddressBook()) {
+				$enumeration = !\array_key_exists('enumeration', $options) || $options['enumeration'] !== false;
 				$fullMatch = !\array_key_exists('fullmatch', $options) || $options['fullmatch'] !== false;
-				if (!$fullMatch) {
-					// Neither full match is allowed, so skip the system address book
+
+				if (!$enumeration && !$fullMatch) {
+					// No access to system address book AND no full match allowed
 					continue;
```
ChristophWurst, line 66: I wonder if this PR could have any unexpected side effects

It would be interesting to set a breakpoint here and test if any other features that use contacts (file sharing, contacts menu) behave different with this change.

miaulalala (approved):

AndyXheli:
Looks like this is also fixed https://github.com/nextcloud/server/issues/34637

PR #38933, opened 2023-06-21
feat: remove orphaned entries from filecache_extended

kesselb:
* Resolves: # 

## Summary

Test scenario:

- Run sql: `select * from oc_filecache_extended left join oc_filecache on oc_filecache_extended.fileid = oc_filecache.fileid where oc_filecache.fileid is null;`
- Mount an external storage (e.g. local)
- Upload some files to the external storage
- Delete external storage
- Run sql: `select * from oc_filecache_extended left join oc_filecache on oc_filecache_extended.fileid = oc_filecache.fileid where oc_filecache.fileid is null;`
- See files without a matching record in filecache 

https://github.com/nextcloud/server/blob/e69c7c4d26731acd91cc683fbe118da1e423c460/lib/private/Files/Cache/Storage.php#L222

Best would be to clear filecache_extended when deleting the storage. 

## TODO

- [ ] OCI
- [x] CI

## Checklist

### DIFF
--- a/apps/files/lib/Command/DeleteOrphanedFiles.php
+++ b/apps/files/lib/Command/DeleteOrphanedFiles.php
@@ -24,9 +24,11 @@
  */
 namespace OCA\Files\Command;
 
+use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 /**
@@ -44,7 +46,8 @@ class DeleteOrphanedFiles extends Command {
 	protected function configure(): void {
 		$this
 			->setName('files:cleanup')
-			->setDescription('cleanup filecache');
+			->setDescription('cleanup filecache')
+			->addOption('filecache-extended', null, InputOption::VALUE_NONE, 'remove orphaned entries from filecache_extended');
 	}
 
 	public function execute(InputInterface $input, OutputInterface $output): int {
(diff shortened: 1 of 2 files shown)


apps/files/lib/Command/DeleteOrphanedFiles.php
```diff
@@ -48,7 +50,8 @@ public function __construct(IDBConnection $connection) {
 	protected function configure() {
 		$this
 			->setName('files:cleanup')
-			->setDescription('cleanup filecache');
+			->setDescription('cleanup filecache')
+			->addOption('filecache-extended', null, InputOption::VALUE_NONE, 'remove orphaned entries from filecache_extended');
```
juliushaertl, line 54: Any reason to not make this run by default?

juliusknorr (approved):
Code looks good, small question inside :)

apps/files/lib/Command/DeleteOrphanedFiles.php
kesselb: Technically, no. The execution for the left join seems quick, but I only tested with ~1000 files. 

My motivation, to hide it behind a flag, is my incomplete knowledge about filecache and especially external storages. However, I can't think of a case, why we would have a record in filecache_extended without a matching one in filecache. 

We can run it by default if you think it should ;)

kesselb:
Moving to 29

icewind1991 (approved):

apps/files/lib/Command/DeleteOrphanedFiles.php
icewind1991: Switched to running it by default and having the flag disable it in case we ever find cases where that is required

kesselb:
Howdy :wave: 

Thank you for taking care, I almost forgot about it :see_no_evil: 

I don't remember if I tested it with OCI back then.
The chunking (for the OCI in clauses limit) is there.

skjnldsv:
```
1) OCA\Files\Tests\Command\DeleteOrphanedFilesTest::testClearFiles
Expectation failed for method name is "writeln" when invoked 2 time(s)
Parameter 0 for invocation #1 Symfony\Component\Console\Output\OutputInterface::writeln('0 orphaned file cache extende...eleted', 0) does not match expected value.
--- Expected
+++ Actual
@@ @@
-'1 orphaned mount entries deleted'
+'0 orphaned file cache extended entries deleted'

```

kesselb:
Moving to 30

PR #38934, opened 2023-06-21
fix(s3): Pass SSE-C parameters for multipart upload

juliusknorr:
Follow up fix for #27034 to make it work properly with SSE-C encryption enabled.


artonge (approved):

tsdicloud:
This is a hot feature to combine with Verified Credentials ...

come-nc (approved):

ir0nhide:
@juliushaertl thank you for adding SSE-C support to the new S3 MultipartUpload feature, really appreciate your work here.
After review the changes, i think you also need to add SSE-C headers to the headObject call on line 99: https://github.com/nextcloud/server/blob/f824d87fc1f3aa99ddb141b7ba19b48641d2cb84/lib/private/Files/ObjectStore/S3.php#L99

Reference: https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

PR #38945, opened 2023-06-22
implement optimized getDirectoryContent for DAV

icewind1991:
- implement `getDirectoryContent` with a single propfind and rework `opendir` to reuse it
- implement `getMetaData` and rework other metadata methods to reuse it
- create shared logic to parse propfind response into file metadata

This both ensures that we only need to make a single propfind request to access a folder, and makes it fill the statcache so that any other metadata operations can reuse the cached info.
(Previously `opendir` didn't fill the cache because it didn't request the same properties)

### DIFF
--- a/lib/private/Files/Storage/DAV.php
+++ b/lib/private/Files/Storage/DAV.php
@@ -47,6 +47,7 @@ use OCP\Constants;
 use OCP\Diagnostics\IEventLogger;
 use OCP\Files\FileInfo;
 use OCP\Files\ForbiddenException;
+use OCP\Files\IMimeTypeDetector;
 use OCP\Files\StorageInvalidException;
 use OCP\Files\StorageNotAvailableException;
 use OCP\Http\Client\IClientService;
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):
Nice, also does the trick and is indeed a bit cleaner 👍 

Thanks @icewind1991

come-nc (approved):

Alexandero89:
@icewind1991 thanks for your effort, it seems to boost the performance!

I noticed we maybe should add
`$file = $this->cleanPath($file);`
in between 925 and 926. So it should look like this:
```
$file = urldecode($file);
$file = substr($file, strlen($this->root));
$file = $this->cleanPath($file);
$this->statCache->set($file, $response);
yield $this->getMetaFromPropfind($file, $response);
```

Otherwise directorypaths are saved incorrectly.

A directory named for example "Backups" is saved to the cache as "Backups/".
But when requesting for existing cache we ask for "Backups".
So directories are never found in the cache.

icewind1991:
> I noticed we maybe should add
> `$file = $this->cleanPath($file);`
> in between 925 and 926. So it should look like this:

https://github.com/nextcloud/server/pull/38945

solracsf:
@icewind1991 you linked to this same PR. 🤔

blizzz:
> @icewind1991 you linked to this same PR. thinking

#38965

PR #38946, opened 2023-06-22
Add a method to get all users with access to a file in OCP

come-nc:
* Resolves: # 

## Summary

In a lot of places we need the list of users having access to a file, this is especially often the case in activity providers in applications, to notify all users of an event.
Currently the implementation is done differently each time, this is an attempt to bring a method in OCP that allows to get the list of users having access to a file and under which path.
For comments for example, this fixes showing the activity event to users having access to the file through a groupfolder.

## TODO

- [ ] See if this is the right implementation
- [ ] Use it in all places that needs this feature
- [ ] Use it in applications where it makes sense, especially activity
- [ ] See how groupfolder ACLs could be integrated in there to avoid returning users which actually do not have access to the file

## Checklist

### DIFF
--- a/lib/private/Files/Config/UserMountCache.php
+++ b/lib/private/Files/Config/UserMountCache.php
@@ -395,12 +395,45 @@ class UserMountCache implements IUserMountCache {
 		$rootFolder = Server::get(IRootFolder::class);
 		$result = [];
 		foreach ($mounts as $mount) {
-			if (isset($result[$mount->getUser()->getUID()])) {
+			$uid = $mount->getUser()->getUID();
+			if (!isset($result[$uid])) {
+				$result[$uid] = [];
+			}
+
+			$userFolder = $rootFolder->getUserFolder($uid);
+			$result[$uid] = array_merge($result[$uid], $userFolder->getById($fileId));
+		}
+
+		return array_filter($result);
+	}
+
+	/**
+	 * Get all users having read access to a file, with a path they see it as.
+	 * They may see the same file under other paths,
+	 *  to get this information use the exhaustive getReadableNodesByUserForFileId
+	 *
+	 * @return array<string,string> Paths giving access to the given fileId, indexed by user ID
+	 * @since 28.0.0
+	 */
+	public function getReadablePathByUserForFileId(int $fileId): array {
+		$mounts = $this->getMountsForFileId($fileId);
+		$rootFolder = Server::get(IRootFolder::class);
+		$result = [];
+		foreach ($mounts as $mount) {
+			$uid = $mount->getUser()->getUID();
+			if (isset($result[$uid])) {
 				continue;
 			}
 
-			$userFolder = $rootFolder->getUserFolder($mount->getUser()->getUID());
-			$result[$mount->getUser()->getUID()] = $userFolder->getById($fileId);
+			$userFolder = $rootFolder->getUserFolder($uid);
+			$nodes = $userFolder->getById($fileId);
+			$node = reset($nodes);
+			if ($node) {
+				$path = $userFolder->getRelativePath($node->getPath());
+				if ($path !== null) {
+					$result[$uid] = $path;
+				}
+			}
 		}
 
 		return $result;
--- a/lib/private/Files/Config/UserMountCache.php
+++ b/lib/private/Files/Config/UserMountCache.php
@@ -14,10 +14,14 @@ use OCP\Diagnostics\IEventLogger;
 use OCP\Files\Config\ICachedMountFileInfo;
 use OCP\Files\Config\ICachedMountInfo;
 use OCP\Files\Config\IUserMountCache;
+use OCP\Files\IRootFolder;
+use OCP\Files\Mount\IMountPoint;
+use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\IDBConnection;
 use OCP\IUser;
 use OCP\IUserManager;
+use OCP\Server;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -378,6 +382,28 @@ class UserMountCache implements IUserMountCache {
 				$internalPath
 			);
 		}, $filteredMounts);
+	}
+
+	/**
+	 * Get all nodes that give access to a file
+	 *
+	 * @return array<string, Node[]> Nodes giving access to the given fileId, indexed by user ID
+	 * @since 28.0.0
+	 */
+	public function getReadableNodesByUserForFileId(int $fileId): array {
+		$mounts = $this->getMountsForFileId($fileId);
+		$rootFolder = Server::get(IRootFolder::class);
+		$result = [];
+		foreach ($mounts as $mount) {
+			if (isset($result[$mount->getUser()->getUID()])) {
+				continue;
+			}
+
+			$userFolder = $rootFolder->getUserFolder($mount->getUser()->getUID());
+			$result[$mount->getUser()->getUID()] = $userFolder->getById($fileId);
+		}
+
+		return $result;
 	}
 
 	/**
--- a/lib/public/Files/Config/IUserMountCache.php
+++ b/lib/public/Files/Config/IUserMountCache.php
@@ -8,6 +8,7 @@
 namespace OCP\Files\Config;
 
 use OCP\Files\Mount\IMountPoint;
+use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\IUser;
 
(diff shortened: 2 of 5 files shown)


lib/public/Files/Config/IUserMountCache.php
```diff
@@ -82,6 +83,14 @@ public function getMountsForRootId($rootFileId);
 	 */
 	public function getMountsForFileId($fileId, $user = null);
 
+	/**
+	 * Get all cached nodes that give access to a file
+	 *
+	 * @return array<string, Node[]>
+	 * @since 28.0.0
+	 */
+	public function getNodesByUserForFileId(int $fileId): array;
```
come-nc, line 92: - [ ] What would be the best name for the method?
- [ ] Is IUserMountCache the right place for it? (if we keep current implementation, I think it is)

lib/private/Files/Config/UserMountCache.php
```diff
@@ -383,6 +384,28 @@ public function getMountsForFileId($fileId, $user = null) {
...
 
+	/**
+	 * Get all nodes that give access to a file
+	 *
+	 * @return array<string, Node[]> Nodes giving access to the given fileId, indexed by user
+	 * @since 28.0.0
+	 */
+	public function getNodesByUserForFileId(int $fileId): array {
+		$mounts = $this->getMountsForFileId($id);
+		$rootFolder = Server::get(IRootFolder::class);
```
come-nc, line 395: - [ ] Can this be added to the DI or is there a problem with that?

lib/public/Files/Config/IUserMountCache.php
ChristophWurst: - [ ] Should we include the permission, e.g. to get the users with *read* permission but also the users with *write* permission?

lib/private/Files/Config/UserMountCache.php
```diff
@@ -383,6 +385,28 @@ public function getMountsForFileId($fileId, $user = null) {
 		}, $filteredMounts);
 	}
 
+	/**
+	 * Get all nodes that give access to a file
+	 *
+	 * @return array<string, Node[]> Nodes giving access to the given fileId, indexed by user
```
ChristophWurst, line 391:
```suggestion
	 * @return array<string, Node[]> Nodes giving access to the given fileId, indexed by user ID
```

just a suggestion for clarity

lib/public/Files/Config/IUserMountCache.php
come-nc: > * [ ]  Should we include the permission, e.g. to get the users with _read_ permission but also the users with _write_ permission?

Do you have an example of where this would be used? I did not see this usecase in our code yet

ChristophWurst: Fair enough. But then let's rename the method to `getReadableNodesByUserForFileId` to make adding other methods easier

come-nc:
Screenshot_20230627_114800
Works nicely

come-nc:
This already takes into account groupfolder ACLs, because `$userFolder->getById($fileId)` will return an empty array in this case.
What’s missing is a check for this to avoid returning an empty array for a user (in this case the user should not appear).

Also, the check `if (isset($result[$mount->getUser()->getUID()])) {` is problematic because it means for a user it will return only the first mount giving access to a file, but it should actually merge them.

That said, in the usecase of activity, which is the only one I found so far, we do not need to merge, we only want to get one path for each file. So I’m tempted to change the method to actually return only one path for each user who has access, as a string. This will make calling code a lot clearer.
The only usecase which needs more info is `info:file` command, to my knowledge.

come-nc:
So in the end I splitted between `getReadablePathByUserForFileId` and `getReadableNodesByUserForFileId`.
The first one stops at first path found for each user, and is useful for activity listeners.
The second one is used by info:file command only for now, but I still think it’s nice to have it in OCP.

I would like to use `getReadablePathByUserForFileId` as well in https://github.com/nextcloud/activity/blob/master/lib/FilesHooks.php , but there `IShareHelper::getPathsForAccessList` is used, which returns paths for local users as well as remote users it seems, and I failed to figure out how to replace only the part for local users.
I failed to find the code filling `$accessList['remotes']` so far.

But for comments and systemtags it works nicely with groupfolders now, even with ACLs. Let’s hope I did not break something else.

come-nc:
> I failed to figure out how to replace only the part for local users. I failed to find the code filling `$accessList['remotes']` so far.

It’s https://github.com/nextcloud/server/blob/b78f3a57d1d05fb5f81ec8361139b64a3e54d7b0/apps/federatedfilesharing/lib/FederatedShareProvider.php#L1090
I missed it because I was searching for `'remotes'`, but in ShareHelper::getPathsForAccessList `remote` is turned into `remotes` actually.

ChristophWurst (approved):
For the lib/public changes

come-nc:
Rebased on master and fixed psalm errors

lib/private/Files/Config/UserMountCache.php
```diff
@@ -383,6 +386,61 @@ public function getMountsForFileId($fileId, $user = null) {
...
+		$rootFolder = Server::get(IRootFolder::class);
+		$result = [];
+		foreach ($mounts as $mount) {
+			$uid = $mount->getUser()->getUID();
+			if (!isset($result[$uid])) {
+				$result[$uid] = [];
+			}
+
+			$userFolder = $rootFolder->getUserFolder($uid);
+			$result[$uid] = array_merge($result[$uid], $userFolder->getById($fileId));
```
icewind1991, line 406: I feel that there should be a more efficient way to do this here since we already know the mount.

Perhaps extracting some of the logic from `RootFolder::getByIdInPath` to create a `RootFolder::getByIdInMounts` or similar.

icewind1991 (commented):
Looks good overall but I would like to ensure this is somewhat performant

lib/private/Files/Config/UserMountCache.php
icewind1991: See https://github.com/nextcloud/server/pull/40482#discussion_r1332676488 for some alternate discussion around the same problem

marcelklehr:
It seems that IUserMountCache#getMountsForFileId() does not include shares because they are not listed in oc_mounts. For shares we would need something like `$shareManager->getAccessList` in addition to the mounts. And for groupfolders we would need an additional piece altogether.

Something like this might work, but hopefully there's a less hacky way:

```
// Assuming you have the file ID
$fileId = 123; // replace with your actual file ID

// Get the root folder
$rootFolder = \OC::$server->getRootFolder();

// Get the file by ID
$file = $rootFolder->getById($fileId);

if ($file) {
    $filePath = $file->getPath();

    // Now we need to find the group folder ID associated with this path
    // Assuming the group folders are stored in a specific path, e.g., "/__groupfolders/<groupfolder_id>/..."
    preg_match('/\/__groupfolders\/(\d+)\//', $filePath, $matches);

    if (isset($matches[1])) {
        $groupFolderId = $matches[1];
    } else {
        throw new Exception('Group folder ID not found');
    }
} else {
    throw new Exception('File not found');
}
```

icewind1991:
> It seems that IUserMountCache#getMountsForFileId() does not include shares because they are not listed in oc_mounts

Shares should be listen in `oc_mounts` (once the recipient of the share has setup their filesystem)

marcelklehr:
Aha! That explains a lot. I also used the wrong array concatenation method (:see_no_evil: ) so, groupfolders are afterall taken care of by this method, so: sorry for the noise, all good.

marcelklehr:
Sad times.

skjnldsv:
@marcelklehr if womeone wanna take over, or OP wanna work again on it, they can still re-open :)

come-nc:
Reopening because we need to figure this out at some point, it’s causing trouble for groupfolders. The question is whether we can do that without making the experience worse in other places.

come-nc:
reopen

come-nc:
Rebased

lib/private/Files/Config/UserMountCache.php
icewind1991: It might make more sense to move this method to the `RootFolder` instead

marcelklehr:
Since there is no other method than this currently, and we already use it in some apps, I vote to get this in, even if it's not super performant.

susnux:
Hello :wave: 
currently here have been no activity in the last years.
If this is still actual feel free to reopen the pull request and resolve the conflicts so we can incorporate it into Nextcloud.

PR #37727, opened 2023-04-14
[F2V] favorites view

skjnldsv:
- [x] Migrate favorites view
- [x] Add favorite action
- [ ] Add missing actions
- [x] Implement webdav prop API

Requires:
- [x] https://github.com/nextcloud/server/pull/38806
- [ ] https://github.com/nextcloud/server/pull/38939

Fix https://github.com/nextcloud/server/issues/7627

PR #38950, opened 2023-06-22
[f2v] Favorites view

### LINKED ISSUE #7627: Add multiple files to favorites
See proposed mockup:

bildschirmfoto 2017-12-26 um 18 35 06

- Nextcloud 13

### LINKED ISSUE #14530: Favorites bulk actions
Hi,

I'm on the last Nextcloud server version (15.0.5) and can't see a way to download all my favorites once I'm on the "favorites" page. It seems the bulk actions are not available as in a classic directory.

skjnldsv:
Supersede  #37727 
Fix #7627

- [x] Tests
- [x] Remove old favorite view when ready
- [ ] Backport https://github.com/nextcloud/server/pull/38950/commits/1fd22584f001ad190987131af29c405ea676b55b to stable27


apps/files/src/components/FileEntry.vue
```diff
@@ -33,38 +33,64 @@
...
+					:autofocus="true"
+					:minlength="3"
+					:required="true"
+					:value.sync="newName"
+					enterkeyhint="done"
+					@keyup="checkInputValidity"
+					@keyup.esc="stopRenaming" />
+			</form>
+
+			<a v-show="!isRenaming"
```
artonge, line 78: Why v-show instead of v-if?

apps/files/src/components/FileEntry.vue
```diff
@@ -237,8 +285,12 @@ export default Vue.extend({
...
 		},
 		displayName() {
-			return this.source.attributes.displayName
-				|| this.source.basename
+			const ext = (this.source.extension || '')
+			const name = (this.source.attributes.displayName
+				|| this.source.basename)
+
+			// Strip extension from name if defined
+			return !ext ? name : name.slice(0, 0 - ext.length)
```
artonge, line 293: Why strip the extension? I don't think this would be expected from a user POV

apps/files/src/components/FileEntry.vue
```diff
@@ -596,6 +656,133 @@ export default Vue.extend({
...
+			}
+
+			// Checking if already exists
+			if (this.checkIfNodeExists(newName)) {
+				showError(this.t('files', 'Another entry with the same name already exists'))
+				return
+			}
+
+			// Set loading state
+			this.loading = 'renaming'
```
artonge, line 742: I would use a different property to store the loading state of a rename, else it can conflict with an actual reload.

1. start reload -> loading = `true`
2. submit rename -> loading = `'rename'`
3. rename is done -> loading = `false`
4. reload is still ongoing, but loading is `false`

apps/files/src/components/FilesListVirtual.vue
```diff
@@ -121,6 +124,13 @@ export default Vue.extend({
 		summary() {
 			return translate('files', '{summaryFile} and {summaryFolder}', this)
 		},
+		isMtimeAvailable() {
+			// Hide mtime column on narrow screens
+			if (this.filesListWidth < 768) {
```
artonge, line 129: Might be useful to have a constant for 768.

apps/files/src/store/paths.ts
```diff
@@ -20,7 +20,7 @@
  *
  */
 /* eslint-disable */
```
artonge, line 22: What's the reason for that?

apps/files/src/store/paths.ts
```diff
@@ -20,7 +20,7 @@
...
-import type { PathOptions, ServicesState } from '../types.ts'
+import type { PathOptions, ServicesState } from '../types'
```
artonge, line 23: I thought our eslint config ask for extensions.

apps/files/src/store/renaming.ts
```diff
@@ -0,0 +1,48 @@
...
+	})
+
+	const renamingStore = store(...arguments)
+
+	// Make sure we only register the listeners once
+	if (!renamingStore._initialized) {
+		subscribe('files:node:rename', function(node: Node) {
+			renamingStore.renamingNode = node
+			renamingStore.newName = node.basename
+		})
```
artonge, line 43: Why is it needed?

apps/files/src/components/FileEntry.vue
skjnldsv: More reactive :)

apps/files/src/components/FileEntry.vue
skjnldsv: This is mostly for toasts messages and info, I figured it would me more readable ? :thinking:

apps/files/src/store/renaming.ts
skjnldsv: That's how a node is renamed, this is how the entire process happens.
You store the current renaming node as well as the new name to start renaming ?

apps/files/src/components/FileEntry.vue
artonge: Might be confusing if multiple file with different extensions. Or folder with the same name.
Extension should not be hidden IMO. But IIRC, Windows and maybe MacOS are hiding it in their file lists. So maybe ask the designers.

apps/files/src/components/FilesListVirtual.vue
skjnldsv: Yeah, for later :)

apps/files/src/store/renaming.ts
artonge: So `'files:node:rename'` is to ask the UI to show the form, and we have another event to inform that the rename was done?

skjnldsv: [`files:node:updated`](https://github.com/nextcloud/server/blob/fd9a4642d2a67817febd833115c3cb24f55418b1/apps/files/src/components/FileEntry.vue#L760) is fired yes :)
We could also fire `files:node:renamed`

artonge (approved):

apps/files/src/components/FileEntry.vue
skjnldsv: Reload?

`this.loading` isn't used for reload :), this is a fileEntry, this is used to target the part of the file row that is loading, it is usually an action id (here renaming is one too) 
https://github.com/nextcloud/server/blob/fd9a4642d2a67817febd833115c3cb24f55418b1/apps/files/src/components/FileEntry.vue#L571

artonge: Sorry for the incomprehension then.

The concern still hold no? If two actions are started at the same time, the one that finish the first will reset the loading state even though the second one is still pending.

skjnldsv: You cannot trigger two actions at the same time, they are disabled while one is running. Including batch actions :)

nfebe (approved):
Great, thank you!

PR #38952, opened 2023-06-23
Add cancel button for when user entry is being edited

### LINKED ISSUE #22608: Add a Cancel button when editing a user
### How to use GitHub

There is no obvious intuitive way to cancel editing a user. This is confusing for first time admins.

*Steps to reproduce*

1. Login to NC as an Admin
2. Create a new user, setting a username and display name
3. Edit the user
4. Change the user's display name
5. Realise that you did not actually want to change the user's display name

Now there is only the "OK" button with the check mark icon to submit your changes.
But there is no "Cancel" button to neglect your changes and go back to the original display name.

Add a second button "Cancel" next to the "OK" button. Clicking the "Cancel" button should ignore (and maybe revert) any changes you have made to the user's properties and return back to displaying the user's original properties.

The only way I have found to effectively cancel editing a user was by reloading the entire page (F5).
However, that is not immediately obvious to newcomers.

**Additional context**

JeaNugroho:
* Resolves: #22608

## Summary
Add a cancel button between checkmark button and three-dot button for when user entry is being edited.
### Before
Screen Shot 2023-06-22 at 8 02 07 PM

### After
Screen Shot 2023-06-21 at 10 40 21 PM

## Potential next steps
[This comment](https://github.com/nextcloud/server/issues/22608#issuecomment-751663701) may be the next fix to make user experience more intuitive and straightforward. I personally would think that the user editing process would be like this if I were a new user to the app. I can certainly implement a way to make this happen, either in the same PR or in the next PR after this one's completed.

## Checklist

- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included or are not applicable

### DIFF
--- a/apps/settings/src/components/Users/UserRow.vue
+++ b/apps/settings/src/components/Users/UserRow.vue
@@ -221,12 +221,10 @@
 		</div>
 
 		<div class="userActions">
-<<<<<<< HEAD
 			<UserRowActions v-if="!loading.all"
 				:actions="userActions"
 				:edit="true"
 				@update:edit="toggleEdit" />
-=======
 			<div v-if="!loading.all"
 				class="toggleUserActions">
 				<NcActions>
@@ -250,7 +248,6 @@
 				<div class="icon-checkmark" />
 				{{ feedbackMessage }}
 			</div>
->>>>>>> 7a2938b01f3 (Add cancel button for when user entry is being edited)
 		</div>
 	</div>
 </template>
--- a/apps/settings/src/components/Users/UserRow.vue
+++ b/apps/settings/src/components/Users/UserRow.vue
@@ -230,10 +230,6 @@
 			<div v-if="!loading.all"
 				class="toggleUserActions">
 				<NcActions :inline="2">
-					<NcActionButton icon="icon-checkmark"
-						:title="t('settings', 'Done')"
-						:aria-label="t('settings', 'Done')"
-						@click="editing = false" />
 					<NcActionButton icon="icon-close"
 						:title="t('settings', 'Cancel')"
 						:aria-label="t('settings', 'Cancel')"
--- a/apps/settings/src/components/Users/UserRow.vue
+++ b/apps/settings/src/components/Users/UserRow.vue
@@ -221,10 +221,40 @@
 		</div>
 
 		<div class="userActions">
+<<<<<<< HEAD
 			<UserRowActions v-if="!loading.all"
 				:actions="userActions"
 				:edit="true"
 				@update:edit="toggleEdit" />
+=======
+			<div v-if="!loading.all"
+				class="toggleUserActions">
+				<NcActions :inline="2">
+					<NcActionButton icon="icon-checkmark"
+						:title="t('settings', 'Done')"
+						:aria-label="t('settings', 'Done')"
+						@click="editing = false" />
+					<NcActionButton icon="icon-close"
+						:title="t('settings', 'Cancel')"
+						:aria-label="t('settings', 'Cancel')"
+						@click="editing = false" />
+				</NcActions>
+				<div v-click-outside="hideMenu" class="userPopoverMenuWrapper">
+					<button class="icon-more"
+						:aria-expanded="openedMenu"
+						:aria-label="t('settings', 'Toggle user actions menu')"
+						@click.prevent="toggleMenu" />
+					<div :class="{ 'open': openedMenu }" class="popovermenu">
+						<NcPopoverMenu :menu="userActions" />
+					</div>
+				</div>
+			</div>
+			<div :style="{opacity: feedbackMessage !== '' ? 1 : 0}"
+				class="feedback">
+				<div class="icon-checkmark" />
+				{{ feedbackMessage }}
+			</div>
+>>>>>>> 7a2938b01f3 (Add cancel button for when user entry is being edited)
 		</div>
 	</div>
 </template>
--- a/apps/settings/src/components/Users/UserRow.vue
+++ b/apps/settings/src/components/Users/UserRow.vue
@@ -229,7 +229,7 @@
 =======
 			<div v-if="!loading.all"
 				class="toggleUserActions">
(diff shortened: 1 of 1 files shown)


apps/settings/src/components/Users/UserRow.vue
```diff
@@ -239,11 +239,15 @@
 		<div class="userActions">
 			<div v-if="!loading.all"
 				class="toggleUserActions">
-				<NcActions>
+				<NcActions :inline="2">
 					<NcActionButton icon="icon-checkmark"
 						:title="t('settings', 'Done')"
 						:aria-label="t('settings', 'Done')"
 						@click="editing = false" />
```
Fenn-CS, line 246:
```suggestion
```

Perhaps? I mean the check and the cancel appears to do the same thing, so it's even more confusing or misleading because it could imply that the `check` is to validate the action.

We should probably remove it or make the check validate all "line edits"

nfebe (changes requested):
Thank you so far! Left an inline comment.

apps/settings/src/components/Users/UserRow.vue
JeaNugroho: @Fenn-CS I agree with you here that this is one of the fixes. 

However, I have a rough thought on how to implement the check button click as a validation to all line edits. This would mean submit any changes to the user fields by clicking on that check button and omit the "submit individual field changes by clicking on the arrow button in each field" functionality. I can add this as another commit to this PR. If it turns out that the fix is way harder than I thought, then I'll resort to deleting the check button.

Would you say the same?

JeaNugroho: @Fenn-CS I might think that it's best to follow your suggestion. I just realized there could be more complicated things happening while editing the user entries. For e.g., an authorization popup can appear while editing one of the fields. This fix you suggested gives a minimum change that wouldn't bother other features that I have yet to know.

I've also added an additional commit to omit the `:inline="2"` bind since we deleted the checkmark button. Please let me know if I can change any other things, otherwise I'm happy if this PR can be completed and changes can be merged. This'd my first PR for Nextcloud and open source contribution =)

artonge (commented):
Watch out as the component has been heavily modified recently, so your change might need to be adapted.

https://github.com/nextcloud/server/pull/38839

nimishavijay (changes requested):
Super nice! :) Only one tiny thing to nitpick: All the icons in a column should be center-aligned ideally. So the buttons should be moved slightly to the right, so that the close icon is nicely aligned with all the edit icons above it :)

apps/settings/src/components/Users/UserRow.vue
Fenn-CS: Thanks for you contribution @JeaNugroho !

This now makes better sense to me.

nfebe (commented):
This now looks ok to me however,

- There's a review about alignment which you should address
- Update the screenshot in the description to reflect the new changes

And we would be good to go! Thank you for your contribution :)

susnux:
This needs to be rebased as the @artonge said it was heavily modified.
The settings are not saved when pressing the check button "done" but when you save a field. Similar to all other settings on Nextcloud which are saved on change rather than that there is a "save" button.

JeaNugroho:
Have just rebased! I'll try to fix the alignment and update the original commit with the new screenshot of the layout. And then when any of the required CI checks fail, I hope someone can provide guidance for that 😊

@Fenn-CS @susnux @artonge @nimishavijay

JeaNugroho:
@Fenn-CS @artonge @susnux 
We now clearly know that we need to fix eslint and Node. I recall from a different PR I'm doing, we need to run `npm run build` to fix Node and run `npm run lint:fix` to fix eslint. So I don't mess this PR like the other one, in what sequence do I have to run these? And are there any other commands or things should I do before, after, or in between these two commands?

Fenn-CS:
@JeaNugroho  you should lint and then build ideally! 

However there are some merge conflicts in your PR, check the files tab closely. I would leave some inline comments to help you see them.

apps/settings/src/components/Users/UserRow.vue
```diff
@@ -221,10 +221,36 @@
...
+						<NcPopoverMenu :menu="userActions" />
+					</div>
+				</div>
+			</div>
+			<div :style="{opacity: feedbackMessage !== '' ? 1 : 0}"
+				class="feedback">
+				<div class="icon-checkmark" />
+				{{ feedbackMessage }}
+			</div>
+>>>>>>> 7a2938b01f3 (Add cancel button for when user entry is being edited)
```
Fenn-CS, line 253:
```suggestion
```

Merge conflict, and deleting this line is not the solution though you want to work with perhaps your code editor to help you resolve the conflicts.

apps/settings/src/components/Users/UserRow.vue
```diff
@@ -221,10 +221,36 @@
 		</div>
 
 		<div class="userActions">
+<<<<<<< HEAD
```
Fenn-CS, line 224:
```suggestion
```

Another conflict marker

apps/settings/src/components/Users/UserRow.vue
```diff
@@ -221,10 +221,36 @@
...
 			<UserRowActions v-if="!loading.all"
 				:actions="userActions"
 				:edit="true"
 				@update:edit="toggleEdit" />
+=======
```
Fenn-CS, line 229:
```suggestion
```

Conflict.

Pytal:
Thanks @JeaNugroho! We have some more modifications incoming, would it be fine with you if we revisit this after merging https://github.com/nextcloud/server/pull/39050 to prevent conflicts?

joshtrichards:
#39050 has been merged, @JeaNugroho. So feel free to proceed. :)

PR #38955, opened 2023-06-23
fix(eventdispatcher): Don't use all evaluating "or"

nickvergessen:
## Summary

`or` is always evaluating all expressions. There is no need to do this here, so in order to emphasize best practises it should be replaced with `||`

## Checklist


come-nc (approved):

come-nc:
> ## Summary
> `or` is always evaluating all expressions. There is no need to do this here, so in order to emphasize best practises it should be replaced with `||`

Source?

https://www.php.net/manual/en/language.operators.logical.php does not state that and https://3v4l.org/06rEj shows second part is not evaluated if first one is true.

PR #35560, opened 2022-12-02
Remove obsolete GD function overwrites

J0WI:
pick from #28279, those functions are part of PHP (7.2+)

PR #38959, opened 2023-06-23
[stable25] fix: BMP image without color info causes array access on `false`

ChristophWurst:
* Resolves: ``Trying to access array offset on value of type bool at lib/private/legacy/OC_Image.php#934``

## Summary

Invalid BMPs can cause PHP errors to be logged again and again.

26+ don't have this problem because the OC_Image method was replaced with a call to GD in https://github.com/nextcloud/server/pull/35560

## TODO

- [x] Fix array access
- [x] Manual tests to verify the error is gone

## Checklist


hamza221 (approved):
Tested ☑️

nfebe (approved):

PR #38987, opened 2023-06-25
Refactors lib/private/Mail.

fsamapoor:
Following [previous PRs taking advantage of PHP8's constructor property promotion](https://github.com/nextcloud/server/pulls?q=is%3Apr+author%3Afsamapoor+Uses+PHP8%27s+constructor+property+promotion) in `/core/` namespace, I have also made the required adjustments to the classes in `/lib/private/Mail` namespace.

The improvements in this PRs include but are not limited to:

- Adding return types
- Adding types to properties

### DIFF
--- a/lib/private/Mail/Message.php
+++ b/lib/private/Mail/Message.php
@@ -46,27 +46,19 @@ use Symfony\Component\Mime\Exception\RfcComplianceException;
  * @package OC\Mail
  */
 class Message implements IMessage {
-	private Email $symfonyEmail;
-	private bool $plainTextOnly;
+	private array $to = [];
+	private array $from = [];
+	private array $replyTo = [];
+	private array $cc = [];
+	private array $bcc = [];
 
-	private array $to;
-	private array $from;
-	private array $replyTo;
-	private array $cc;
-	private array $bcc;
-
-	public function __construct(Email $symfonyEmail, bool $plainTextOnly) {
-		$this->symfonyEmail = $symfonyEmail;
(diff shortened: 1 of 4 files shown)


skjnldsv (approved):

nfebe (approved):

lib/private/Mail/Message.php
```diff
@@ -46,27 +46,19 @@
...
+	private array $bcc = [];
+
+	public function __construct(
+		private Email $symfonyEmail,
+		private bool $plainTextOnly,
+	) {
 	}
 
 	/**
-	 * @return $this
```
come-nc, line 69: This was not redundant, it is adding information.

fsamapoor: The method has a return type of the parent interface and the body of the method has the `return $this` line.
Am I missing something here? Could you please elaborate on what information it is adding?

fsamapoor: Ah, didn't know about that. thank you.

How about the `self` or `static` return types? Would that trigger an error as well?
I changed the return type from `IMessage` to `self` in the minimal example that you linked and everything seems fine.
To be quite honest, I prefer avoiding annotations if there is a native workaround.

fsamapoor: Thank you for the provided details. So I'll just revert the removals of `@return $this`.
Are there any other changes that need to be made in this PR?

fsamapoor: Very well then. I have referenced the discussion in a new PR: #39020

PR #38996, opened 2023-06-25
Refactors lib/private/FullTextSearch

fsamapoor:
Following [previous PRs taking advantage of PHP8's constructor property promotion](https://github.com/nextcloud/server/pulls?q=is%3Apr+author%3Afsamapoor+Uses+PHP8%27s+constructor+property+promotion) in `/core/` namespace, I have also made the required adjustments to the classes in `/lib/private/FullTextSearch` namespace.

The improvements in this PRs include:

- Using early returns
- Adding return types
- Adding types to properties


lib/private/FullTextSearch/Model/IndexDocument.php
```diff
@@ -47,62 +47,41 @@
...
-	protected $id = '';
+	protected string $id = '';
 
-	/** @var string */
-	protected $providerId = '';
+	protected ?DocumentAccess $access;
 
-	/** @var DocumentAccess */
-	protected $access;
+	protected ?IIndex $index;
```
fsamapoor, line 54: Hello @ArtificialOwl. Would you please help me how to type the `$index` and `$access` properties?

The `getIndex(): IIndex` method returns `$this->index` and then there is the `hasIndex(): bool` method that checks if the index is defined (`$this->index !== null`).

Meaning it might be `null`, but the return type of the `getIndex(): IIndex` should remain `IIndex` because of the implemented interface.

So on the one hand it should be typed as nullable, on the other hand, it shouldn't.
Is there anything I am missing here? How should I type these parameters?

icewind1991 (approved):

lib/private/FullTextSearch/FullTextSearchManager.php
```diff
@@ -39,52 +39,46 @@
...
-			$this->searchService === null) {
+		if ($this->indexService === null) {
+			return false;
+		}
+
+		if ($this->providerService === null) {
+			return false;
+		}
+
+		if ($this->searchService === null) {
```
come-nc, line 81: Why?
Please revert this change

lib/private/FullTextSearch/Model/SearchRequestSimpleQuery.php
```diff
@@ -37,34 +37,24 @@
...
-	 *
 	 * @since 17.0.0
 	 */
-	public function __construct(string $field, int $type) {
-		$this->field = $field;
-		$this->type = $type;
+	public function __construct(
+		private string $field = '',
+		private int $type = 0,
+	) {
```
come-nc, line 51: Parameters were mandatory before, you change them to optional

come-nc (changes requested):

lib/private/FullTextSearch/FullTextSearchManager.php
fsamapoor: Will do. First of all, let me thank you again for your timely reviews and feedback. Really appreciate that.

These are the changes that I suggested a while back. Now that I'm more familiar with your taste, I wouldn't suggest these kinds of changes in my future PRs.

Personally, I find single condition checks a lot more readable and easier to follow, so I try to shape a happy path with  single guard clauses. Nothing wrong with the previous implementation though. Most of these kinds of decisions come down to personal preferences and as long as I shoot my shot to test the waters and be familiar with your preferences, I'm fine with reverting every single suggestion that I make.

lib/private/FullTextSearch/Model/SearchRequestSimpleQuery.php
fsamapoor: Thank you. Good catch! Should I remove the default values?

lib/private/FullTextSearch/Model/IndexDocument.php
fsamapoor: @come-nc Would you please help me here as well?

nfebe (approved):

lib/private/FullTextSearch/Model/SearchRequestSimpleQuery.php
come-nc: Yes, because they were unused anyway I think, they were always overwritten by the constructor

lib/private/FullTextSearch/Model/IndexDocument.php
come-nc: I do not have time to dig into it, sorry

fsamapoor: No problem. I'll do my best.

lib/private/FullTextSearch/Model/SearchRequestSimpleQuery.php
fsamapoor: Addressed and rebased.

lib/private/FullTextSearch/FullTextSearchManager.php
fsamapoor: Addressed and rebased.

fsamapoor:
@come-nc Anything I need to change in this PR?

I believe the integration workflow may have timed out.

lib/private/FullTextSearch/Model/IndexDocument.php
```diff
@@ -178,22 +146,15 @@ final public function getIndex(): IIndex {
 	 * return if Index is defined.
 	 *
 	 * @since 16.0.0
-	 *
-	 * @return bool
 	 */
 	final public function hasIndex(): bool {
-		return ($this->index !== null);
+		return isset($this->index);
 	}
```
come-nc, line 152: For me `index` should be typed `?IIndex`, and maybe `getIndex` will need to have a check and a throw.

fsamapoor: Thank you. Please take a look and let me know if I replicated the existing exception behavior in the app correctly.

come-nc (approved):

come-nc:
@fsamapoor autoloaders are not up to date, probably because of the new Exception class, please run build/autoloaderchecker.sh and commit the files with new classes.

fsamapoor:
> @fsamapoor autoloaders are not up to date, probably because of the new Exception class, please run build/autoloaderchecker.sh and commit the files with new classes.

Done.

lib/composer/composer/autoload_classmap.php
```diff
@@ -420,6 +420,7 @@
     'OCP\\Files\\UnseekableException' => $baseDir . '/lib/public/Files/UnseekableException.php',
     'OCP\\Files_FullTextSearch\\Model\\AFilesDocument' => $baseDir . '/lib/public/Files_FullTextSearch/Model/AFilesDocument.php',
     'OCP\\FullTextSearch\\Exceptions\\FullTextSearchAppNotAvailableException' => $baseDir . '/lib/public/FullTextSearch/Exceptions/FullTextSearchAppNotAvailableException.php',
+    'OCP\\FullTextSearch\\Exceptions\\FullTextSearchIndexNotAvailableException' => $baseDir . '/lib/public/FullTextSearch/Exceptions/FullTextSearchIndexNotAvailableException.php',
```
ChristophWurst, line 423: this should be committed :+1:

apps/workflowengine/composer/composer/installed.php
```diff
@@ -3,7 +3,7 @@
         'name' => '__root__',
         'pretty_version' => 'dev-master',
         'version' => 'dev-master',
-        'reference' => 'b1797842784b250fb01ed5e3bf130705eb94751b',
+        'reference' => 'c8456a1129f98e6ae16ce6adf4ad60044eb3b94a',
```
ChristophWurst, line 6: this should not

fsamapoor: @ChristophWurst Thank you for reviewing the changes. This is just an example, right? Basically, `lib/composer/composer` should be the only directory subject to change and no other changes in the `apps` namespace should be committed.

ChristophWurst: The apps/*/composer dirs are still to be committed if an app's autoloader class map changes.

It's specifically the `reference` that we don't commit. It constantly changes. If composer files changed, it's helpful to use git add with the patch option like `git add -p` and only select the good parts :)

lib/public/FullTextSearch/Exceptions/FullTextSearchIndexNotAvailableException.php
```diff
@@ -0,0 +1,35 @@
...
+ */
+namespace OCP\FullTextSearch\Exceptions;
+
+/**
+ * @since 28.0.0
+ *
+ * Class FullTextSearchIndexNotAvailableException
+ *
+ */
+class FullTextSearchIndexNotAvailableException extends \Exception {
```
ChristophWurst, line 34: is this used anywhere?

lib/private/FullTextSearch/Model/IndexDocument.php
```diff
@@ -166,34 +137,30 @@ final public function setIndex(IIndex $index): IIndexDocument {
...
 	 *
+	 * @throws FullTextSearchIndexNotAvailableException
 	 * @since 15.0.0
-	 *
-	 * @return IIndex
 	 */
 	final public function getIndex(): IIndex {
+		if ($this->index === null) {
+			throw new FullTextSearchIndexNotAvailableException('No IIndex generated');
+		}
```
fsamapoor, line 146: @ChristophWurst This is where we introduced the new exception.

apps/workflowengine/composer/composer/installed.php
fsamapoor: Yes. All of the changes in the `app` directory were reference changes, so I rebased and did not include them anymore. Thank you for the detailed explanation, Christoph!

ChristophWurst (approved):

PR #39008, opened 2023-06-26
switch to bootstrap font-stack

szaimen:
Switch to https://getbootstrap.com/docs/5.1/content/reboot/#native-font-stack

### DIFF
--- a/apps/theming/lib/Themes/DefaultTheme.php
+++ b/apps/theming/lib/Themes/DefaultTheme.php
@@ -171,7 +171,7 @@ class DefaultTheme implements ITheme {
 			'--color-border-dark' => $this->util->darken($colorMainBackground, 14),
 			'--color-border-maxcontrast' => $this->util->darken($colorMainBackground, 42),
 
-			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', 'Liberation Sans', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
+			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
 			'--default-font-size' => '15px',
 
 			// TODO: support "(prefers-reduced-motion)"
--- a/apps/theming/lib/Themes/DefaultTheme.php
+++ b/apps/theming/lib/Themes/DefaultTheme.php
@@ -171,7 +171,7 @@ class DefaultTheme implements ITheme {
 			'--color-border-dark' => $this->util->darken($colorMainBackground, 14),
 			'--color-border-maxcontrast' => $this->util->darken($colorMainBackground, 42),
 
-			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
+			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
 			'--default-font-size' => '15px',
 
 			// TODO: support "(prefers-reduced-motion)"
--- a/apps/theming/lib/Themes/DefaultTheme.php
+++ b/apps/theming/lib/Themes/DefaultTheme.php
@@ -171,7 +171,7 @@ class DefaultTheme implements ITheme {
 			'--color-border-dark' => $this->util->darken($colorMainBackground, 14),
 			'--color-border-maxcontrast' => $this->util->darken($colorMainBackground, 42),
 
-			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
+			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'",
 			'--default-font-size' => '15px',
 
 			// TODO: support "(prefers-reduced-motion)"
--- a/apps/theming/lib/Themes/DefaultTheme.php
+++ b/apps/theming/lib/Themes/DefaultTheme.php
@@ -171,7 +171,7 @@ class DefaultTheme implements ITheme {
 			'--color-border-dark' => $this->util->darken($colorMainBackground, 14),
 			'--color-border-maxcontrast' => $this->util->darken($colorMainBackground, 42),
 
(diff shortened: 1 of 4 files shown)


jancborchardt (commented):
Is this because you have one of the previous fonts in the list installed on the system? Cause otherwise the font stack is outdated and Segoe UI does not seem to be the default windows font anymore and we should update the font stack.

szaimen:
> Is this because you have one of the previous fonts in the list installed on the system? Cause otherwise the font stack is outdated and Segoe UI does not seem to be the default windows font anymore and we should update the font stack.

As far as I can see is Segoe UI still the default Font used on Windows. Unfortunately it brings the not-centered digits with it as shown above. So the suggested solution was to move it behind `sans-serif` which centers the digits correctly.

susnux:
`Segoe UI` seems to be required on Windows for Arabic (so should be at least at some position on the font stack): https://github.com/nextcloud/nextcloud-vue/issues/3780#issuecomment-1471724927

szaimen:
> `Segoe UI` seems to be required on Windows for Arabic (so should be at least at some position on the font stack): [nextcloud/nextcloud-vue#3780 (comment)](https://github.com/nextcloud/nextcloud-vue/issues/3780#issuecomment-1471724927)

Yes, I did not remove the font, just changed the order of it :)

solracsf:
What about the Bootstrap default font stack?
https://getbootstrap.com/docs/5.1/content/reboot/#native-font-stack

szaimen:
> What about the Bootstrap default font stack? https://getbootstrap.com/docs/5.1/content/reboot/#native-font-stack

This is unfortunately not going to fix the problem with bad digit rendering as Segoe UI apparently causes this.

jancborchardt:
> As far as I can see is Segoe UI still the default Font used on Windows.

But if sans-serif results in a different font (which one?), Segoe UI can't be the default, right? ;)

szaimen:
> But if sans-serif results in a different font (which one?)

It is sans-serif :) (there is a font that is called like this on windows afaik)

susnux:
`sans-serif` should be delivered as `Segoe UI` on Window, so I do not understand why this works here.
Moreover putting `Segoe` behind `sans-serif` will disable `Segoe` nearly always because every system has a font for `sans-serif`.

jancborchardt:
@szaimen The point is that if we move Segoe UI behind sans-serif and that fixes the issue by using a different font, it also means that:

- Segoe UI is not the default font on Windows (or you have a different one that overrides that default)
- We might as well remove Segoe UI from the font stack as it would never get picked. sans-serif is the catch-all.

I would prefer we just stick to the default Bootstrap font stack. It was probably updated since we adopted it.

susnux:
> I would prefer we just stick to the default Bootstrap font stack. It was probably updated since we adopted it.

It would look like this:
```scss
font-family:
  // Cross-platform generic font family (default user interface font)
  system-ui,
  // Safari for macOS and iOS (San Francisco)
  -apple-system,
  // Windows
  "Segoe UI",
  // Android
  Roboto,
  // older macOS and iOS
  "Helvetica Neue"
  // Linux
  "Noto Sans",
  "Liberation Sans",
  // Basic web fallback
  Arial,
  // Sans serif fallback
  sans-serif,
  // Emoji fonts
  "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji" !default;
```

szaimen:
> > I would prefer we just stick to the default Bootstrap font stack. It was probably updated since we adopted it.
> 
> [It would look like this](https://getbootstrap.com/docs/5.2/content/reboot/#native-font-stack):
> 
> ```css-scss
> font-family:
>   // Cross-platform generic font family (default user interface font)
>   system-ui,
>   // Safari for macOS and iOS (San Francisco)
>   -apple-system,
>   // Windows
>   "Segoe UI",
>   // Android
>   Roboto,
>   // older macOS and iOS
>   "Helvetica Neue"
>   // Linux
>   "Noto Sans",
>   "Liberation Sans",
>   // Basic web fallback
>   Arial,
>   // Sans serif fallback
>   sans-serif,
>   // Emoji fonts
>   "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji" !default;
> ```

So shall I keep it multilined? I thought having it on one line was preferred?

apps/theming/lib/Themes/DefaultTheme.php
```diff
@@ -171,7 +171,7 @@ public function getCSSVariables(): array {
 			'--color-border-dark' => $this->util->darken($colorMainBackground, 14),
 			'--color-border-maxcontrast' => $this->util->darken($colorMainBackground, 42),
 
-			'--font-face' => "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', Arial, sans-serif, 'Noto Color Emoji', 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'",
+			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', Arial, 'Noto Sans', 'Liberation Sans', sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
```
susnux, line 174:
```suggestion
			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
```

https://getbootstrap.com/docs/5.2/content/reboot/#native-font-stack

susnux:
> So shall I keep it multilined? I thought having it on one line was preferred?

No I wrote that comment just before you changed the PR^^
Single line is preferred :)

apps/theming/lib/Themes/DefaultTheme.php
szaimen: done

susnux (approved):

Pytal (approved):

jancborchardt (changes requested):
Only one adjustment – and we should include a comment like `Bootstrap font stack + Oxygen-Sans, Cantarell, Ubuntu`:

We need to add `Oxygen-Sans, Cantarell, Ubuntu` between Roboto and Helvetica Neue. This is so it looks native on KDE, GNOME and Ubuntu as well before it gets to the fallbacks.

szaimen:
> Only one adjustment – and we should include a comment like `Bootstrap font stack + Oxygen-Sans, Cantarell, Ubuntu`:
> 
> We need to add `Oxygen-Sans, Cantarell, Ubuntu` between Roboto and Helvetica Neue. This is so it looks native on KDE, GNOME and Ubuntu as well before it gets to the fallbacks.

But then it is not the native bootstrap font stack anymore? But would be fine for me...

jancborchardt:
> But then it is not the native bootstrap font stack anymore?

Yes, it is the one enhancement we need to make to their stack since they don’t really care about Linux except "Noto Sans" which is a bit generic. We use the actual native fonts for KDE, GNOME and Ubuntu, just like it is done for Windows and MacOS. This is why we should include a code comment there. :)

jancborchardt (approved):
All good! :)

apps/theming/lib/Themes/DefaultTheme.php
```diff
@@ -171,7 +171,7 @@ public function getCSSVariables(): array {
 			'--color-border-dark' => $this->util->darken($colorMainBackground, 14),
 			'--color-border-maxcontrast' => $this->util->darken($colorMainBackground, 42),
 
-			'--font-face' => "-apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', Arial, sans-serif, 'Noto Color Emoji', 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol'",
+			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';",
```
szaimen, line 174:
```suggestion
			'--font-face' => "system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji'",
```

szaimen:
Cypress failure unrelated afaics

PR #38915, opened 2023-06-20
adjust admin setup check to increase warning to configure https

szaimen:
Close https://github.com/nextcloud/server/issues/36582

Reason: Without https some important web functionality like "copy to clipboard" or "service workers" does not work!

PR #39009, opened 2023-06-26
fix setupcheck test

szaimen:
Follow-up to https://github.com/nextcloud/server/pull/38915


artonge (approved):

susnux:
1. If it is just "recommended" why it is an error?
2. HSTS is not required for "copy to clipboard" and "service workers", just HTTPS for connection (or wss) is enough, so this message might be misleading?

szaimen:
You are right. I adjusted the wrong place.

szaimen:
See https://github.com/nextcloud/server/pull/39014

PR #39013, opened 2023-06-26
[3/3] Refactors lib/private/Security

fsamapoor:
Following [previous PRs taking advantage of PHP8's constructor property promotion](https://github.com/nextcloud/server/pulls?q=is%3Apr+author%3Afsamapoor+Uses+PHP8%27s+constructor+property+promotion) in `/core/` namespace, I have also made the required adjustments to the classes in `/lib/private/Security` namespace.

I figured I should split the changes into three PRs to make reviewing the changes easier.

- Adding return types
- Adding types to properties

### DIFF
--- a/lib/private/Security/Certificate.php
+++ b/lib/private/Security/Certificate.php
@@ -32,9 +32,9 @@ use OCP\ICertificate;
 class Certificate implements ICertificate {
 	protected string $name;
 
-	protected mixed $commonName;
+	protected ?string $commonName;
 
-	protected mixed $organization;
+	protected ?string $organization;
 
 	protected $serial;
 
@@ -42,9 +42,9 @@ class Certificate implements ICertificate {
 
 	protected \DateTime $expireDate;
 
-	protected mixed $issuerName;
+	protected ?string $issuerName;
 
-	protected mixed $issuerOrganization;
+	protected ?string $issuerOrganization;
 
 	/**
 	 * @param string $data base64 encoded certificate
--- a/lib/private/Security/Certificate.php
+++ b/lib/private/Security/Certificate.php
@@ -36,7 +36,6 @@ class Certificate implements ICertificate {
 
 	protected ?string $organization;
 
-	protected $serial;
 
 	protected \DateTime $issueDate;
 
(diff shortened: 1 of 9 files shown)


lib/private/Security/Certificate.php
```diff
@@ -30,25 +30,24 @@
 use OCP\ICertificate;
 
 class Certificate implements ICertificate {
-	protected $name;
+	protected string $name;
 
-	protected $commonName;
+	protected mixed $commonName;
```
come-nc, line 35: Why the `mixed` types? It seems to be `?string` since getCommonName returns `?string`.

fsamapoor: It is being evaluated based on `$info['subject']['CN']` and I wasn't sure if its value would be of type `string`, but now that you mentioned it, I think it is safe to type it as it's getter's return type, which is `?string`.
I'll update other properties as well.

fsamapoor: @come-nc One other thing. I couldn't figure out the property type for the `$serial` since my IDE didn't find any usage of that property in the project. How could I know if it is needed or it is redundant?

lib/private/Security/Certificate.php
```diff
@@ -30,25 +30,24 @@
...
-	protected $name;
+	protected string $name;
 
-	protected $commonName;
+	protected ?string $commonName;
 
-	protected $organization;
+	protected ?string $organization;
 
 	protected $serial;
```
come-nc, line 39:
```suggestion
```

Indeed it is unused, always has been, please remove it.

come-nc (approved):
Good otherwise

Fenn-CS:
Hello, @fsamapoor please rebase.

nfebe (approved):

fsamapoor:
> Hello, @fsamapoor please rebase.

Messed it up a bit. I'll rebase it again to remove the merge commit and let you know afterward.

fsamapoor:
> Hello, @fsamapoor please rebase.

Rebased and removed the merge commit.

icewind1991 (approved):

PR #23932, opened 2020-11-06
Allow config to specify the bucket exists

rullzer:
In the 99% case the bucket is just always there. And if it is not the
read/write will fail hard anyways. Esp on big instances the Objectstore
is not always fast and this can save a few hundered ms of each request
that acess the objectstore.

In short it is adding

'verify_bucket_exists' => false

To the S3 config part

Signed-off-by: Roeland Jago Douma <roeland@famdouma.nl>

PR #39017, opened 2023-06-26
fix(s3): fix handling verify_bucket_exists parameter

tcitworld:
If `'verify_bucket_exists'` is set to false in the `config.php` s3 configuration, it's supposed to avoid verifying that the bucket exists. However `empty(falsy)` will always return true, so this condition would not work.

Follow-up to #23932

## Checklist

### DIFF
--- a/lib/private/Files/ObjectStore/S3ConnectionTrait.php
+++ b/lib/private/Files/ObjectStore/S3ConnectionTrait.php
@@ -92,7 +92,7 @@ trait S3ConnectionTrait {
 		if (!isset($params['port']) || $params['port'] === '') {
 			$params['port'] = (isset($params['use_ssl']) && $params['use_ssl'] === false) ? 80 : 443;
 		}
-		$params['verify_bucket_exists'] = empty($params['verify_bucket_exists']) ? true : $params['verify_bucket_exists'];
+		$params['verify_bucket_exists'] = $params['verify_bucket_exists'] ?? true;
 		$this->params = $params;
 	}
 


joshtrichards (approved):
Good catch.

Documentation covered few weeks back in nextcloud/documentation#10437 &  nextcloud/documentation#10436

joshtrichards:
There are also (some # of) installations out there with `autocreate` in their config to toggle this functionality with S3 (because until recently `autocreate` was in our docs - though it did nothing - oops!) 

Probably too late (risky) to make that parameter equivalent now without blowback. Also, probably not a big deal as long as it: (a) now works (b) the correct parameter is now documented.

juliusknorr (approved):

jdavid82:
I've been happily using NC with S3 integration for a while now.
But I just upgraded NC through the `stable` channel and then I started to get sync errors from the NC Desktop app. After further investigation I came across the following error message in the admin section in the web app: `Exception: Creation of bucket "..." failed. Error executing "CreateBucket" on "https://..."`. (which I think is related to this PR).

If it is in fact related to this PR, then would it be possible to get an ETA on this fix? In the meantime, I'll pause sync on Nexcloud desktop app and manually add/update the affected files to S3, thanks

EchedeyLR:
I was not aware of this underlaying issue, so even with documentation up-to-date there was a serious bug.

baldorch:
> I've been happily using NC with S3 integration for a while now.
> But I just upgraded NC through the `stable` channel and then I started to get sync errors from the NC Desktop app. After further investigation I came across the following error message in the admin section in the web app: `Exception: Creation of bucket "..." failed. Error executing "CreateBucket" on "https://..."`. (which I think is related to this PR).
> 
> If it is in fact related to this PR, then would it be possible to get an ETA on this fix? In the meantime, I'll pause sync on Nexcloud desktop app and manually add/update the affected files to S3, thanks

Does the bucket exist? Before it tries to create the bucket it will check if the bucket exists. I observed the same error, although my bucket existed. In my case the role used by NC lacked the "s3:ListBucket" Permission (which is needed to check if the bucket exists). Adding that removed the error.

jdavid82:
@baldorch yes like I said I've been using S3 integration for a while (220 days to be precise)

EchedeyLR:
@baldorch what you say only works with AWS S3 implementation for which ListBuckets is an existing permission.

For the rest, you either give all the permission or you have to use the option to prevent that check.

baldorch:
@EchedeyLR I am using Wasabi, they might have a complete AWS S3 Compliant thing going.

What I am wondering is - there are two condition checks in the code before it tries to create the bucket. The first is "Should I check at all?" (which is configured by the option affected by this bug) and the second is "Does the bucket exist?" (https://github.com/nextcloud/server/blob/bbb082741759f8a97bd7e5a97d93e069189fa348/lib/private/Files/ObjectStore/S3ConnectionTrait.php#L156). If `verify_bucket_exists` always evaluates to `true`, the outcome only depends on the existence check.

In my case the existence check failed because said permission was missing.

But I might be getting something wrong here, just saw the error mentioned in the comment and was like Leonardo DiCaprio: "I know that one!" :)

jdavid82:
Hi team, I just checked this pull request status and also went to the admin backend in my NC instance, and it turns out that AWS S3 bucket integration is now working fine again.

I don't know if this is due to some local cache being reset on my server or whether there was a temporary problem at the S3 side. But it seems to be working fine now... I just thought I'd mention it. Sorry for the red herring

tcitworld:
I've marked the above message off-topic as the bug was present from the very beginning this feature (and the parameter) was introduced.
https://github.com/nextcloud/server/pull/23932

susnux (approved):

Issue #32127, opened 2022-04-25
[Meta]: Port away from ILogger


CarlSchwan:
## Current pull requests
 - [x] https://github.com/nextcloud/server/pull/31609
 - [x] https://github.com/nextcloud/server/pull/31943
 - [x] https://github.com/nextcloud/server/pull/31948
 - [x] https://github.com/nextcloud/server/pull/40508
 - [ ] ...

## Current remaining modules to port in server
 - [x] lib/{private,public}/Files 

## Current remaining shipped app to port
 - [x] federation https://github.com/nextcloud/server/pull/39018
 - [x] encryption https://github.com/nextcloud/server/pull/39065
 - [x] federatedfilesharing #39160 
 - [x] files_sharing https://github.com/nextcloud/server/pull/39169
 - [ ] files_trashbin
 - [x] files_external https://github.com/nextcloud/server/pull/39480
 - [ ] dav
 - [x] weather_status
 - [ ] settings
 - [x] files
 - [ ] workflowengine
 - [x] user_status
 - [x] user_ldap https://github.com/nextcloud/server/pull/40504
 - [x] updatenotification
 - [x] twofactor_backupcodes
 - [x] theming
 - [x] sharebymail

## Current issues

- The background job interface exposes `ILogger` in its public API. We need to deprecate this
- In some places, like the updater. It's helpful to be able to write `ILogger:DEBUG` instead of `2`. We need to look if upstream psr would be interested to add enums for the log levels.

come-nc:
- [ ] https://github.com/nextcloud/server/pull/31775

come-nc:
- [ ] https://github.com/nextcloud/server/pull/31752

Altahrim:
- [ ] #39160

Altahrim:
- [ ] #39169

joshtrichards:
All that's left I think is to remove these deprecated references in the `\OCP\BackgroundJob` interfaces I think:

https://github.com/nextcloud/server/blob/1518ded8b193ca789b28f8f64114798ab8df1747/lib/public/BackgroundJob/Job.php#L42-L46](https://github.com/search?q=repo%3Anextcloud%2Fserver%20path%3A%2F%5Elib%5C%2Fpublic%5C%2FBackgroundJob%5C%2F%2F%20Ilogger&type=code)

come-nc:
See https://github.com/nextcloud/server/pull/48197 , it will be for Nextcloud 33 because of API stability rules.

PR #39018, opened 2023-06-26
Migrate federation application to LoggerInterface

come-nc:
See #32127

Migrate federation application from ILogger to LoggerInterface

## Checklist

### DIFF
--- a/apps/federation/lib/BackgroundJob/RequestSharedSecret.php
+++ b/apps/federation/lib/BackgroundJob/RequestSharedSecret.php
@@ -1,10 +1,14 @@
 <?php
+
+declare(strict_types=1);
+
 /**
  * @copyright Copyright (c) 2016, ownCloud, Inc.
  *
  * @author Bjoern Schiessle <bjoern@schiessle.org>
  * @author Björn Schießle <bjoern@schiessle.org>
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ * @author Côme Chilliet <come@chilliet.eu>
  * @author Joas Schilling <coding@schilljs.com>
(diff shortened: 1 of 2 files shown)


apps/federation/lib/BackgroundJob/RequestSharedSecret.php
```diff
@@ -49,74 +53,37 @@
...
 
 
 	/**
 	 * run the job, then remove it from the joblist
-	 *
-	 * @param IJobList $jobList
-	 * @param ILogger|null $logger
 	 */
-	public function execute(IJobList $jobList, ILogger $logger = null) {
+	public function start(IJobList $jobList): void {
```
Fenn-CS, line 82: This is a public method, just double checking to ensure you have checked for references in other parts of the code?

nfebe (approved):
Looks good otherwise.

apps/federation/lib/BackgroundJob/RequestSharedSecret.php
come-nc: Yes, the parent class `OCP\BackgroundJob\Job` is migrating from `execute` to `start` since 25, I think we should even deprecate `execute` since it’s always a problem for migrating away from `ILogger`.

Fenn-CS: Is there an issue for that? If not I could create one and deprecate it rn.

come-nc: Actually I just saw that `execute` is already deprecated in `IJob` interface.
Maybe it would still be a good idea to mark it as deprecated in `Job` class as well.

artonge (approved):

PR #39025, opened 2023-06-27
fix(files): Only render the menu if there are actions to show

juliusknorr:
## Summary

Filter action list that is used to determine if the actions menu should be visible properly so that:
- Default actions are not included
- Only the list of available actions for a file is used

Steps to reproduce:
- Share a folder as read only link and enable "hide download"

This change will make sure that we don't show the 3 dots menu for the files in the share as the only option available is "View" which is not in the menu as it is the default. Before that an empty menu was shown.
## Checklist

### DIFF
--- a/apps/files/js/fileactions.js
+++ b/apps/files/js/fileactions.js
@@ -597,8 +597,8 @@
 				Object.values = objectValues;
 			}
 
-			var menuActions = Object.values(this.actions.all).filter(function (action) {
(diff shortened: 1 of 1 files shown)


artonge (approved):

Pytal (approved):

susnux (approved):

juliushaertl:
Samba failure unrelated

PR #39029, opened 2023-06-27
chore: upgrade typescript, eslint and fix lint

skjnldsv:
- Upgrades typescript
- Upgrade eslint
- Fix eslint


.eslintrc.js
```diff
@@ -3,21 +3,18 @@ module.exports = {
...
+		moment: true,
+		oc_userconfig: true,
+		sinon: true,
 	},
 	plugins: [
 		'cypress',
 	],
 	extends: [
-		'@nextcloud',
+		'@nextcloud/eslint-config/typescript',
```
artonge, line 17: Should we change this in apps too?

artonge (approved):

.eslintrc.js
susnux: If they use Typescript I would recommend to do so.

susnux (approved):

PR #39040, opened 2023-06-28
fix(richdocuments): Mark some parameters of richdocuments WopiMapper as sensitive

nickvergessen:
Found in our own logs

## Checklist

### DIFF
--- a/lib/private/Log/ExceptionSerializer.php
+++ b/lib/private/Log/ExceptionSerializer.php
@@ -39,6 +39,7 @@ use OCA\Encryption\Crypto\Encryption;
 use OCA\Encryption\Hooks\UserHooks;
 use OCA\Encryption\KeyManager;
 use OCA\Encryption\Session;
+use OCA\Richdocuments\Db\WopiMapper;
 use OCP\HintException;
 
 class ExceptionSerializer {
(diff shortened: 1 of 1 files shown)


lib/private/Log/ExceptionSerializer.php
```diff
@@ -39,6 +39,7 @@
 use OCA\Encryption\Hooks\UserHooks;
 use OCA\Encryption\KeyManager;
 use OCA\Encryption\Session;
+use OCA\Richdocuments\Db\WopiMapper;
```
tcitworld, line 42: **question:** Won't this explode if richdocuments isn't actived?

juliushaertl: No just if you try to instantiate the class :)

juliusknorr (approved):

rullzer:
I do get the quick fix. However for future proofing this (and quicker updates). It would be great if there is a way for apps to register their own sensitive parameters. That way it would scale better and you don't have to wait for a server update each time.

blizzz:
> I do get the quick fix. However for future proofing this (and quicker updates). It would be great if there is a way for apps to register their own sensitive parameters. That way it would scale better and you don't have to wait for a server update each time.

We have that: https://github.com/nextcloud/server/blob/master/lib/public/AppFramework/Bootstrap/IRegistrationContext.php#L352

nickvergessen:
> I do get the quick fix. However for future proofing this (and quicker updates). It would be great if there is a way for apps to register their own sensitive parameters. That way it would scale better and you don't have to wait for a server update each time.

Well once PHP 8.2 is the minimum, the correct way is using the native PHP attribute (which I added in https://github.com/nextcloud/richdocuments/pull/3042 )

I was also not aware of the IRegisterContext function, so yeah will use that one instead.

PR #39044, opened 2023-06-28
Fix root mounts not being setup in some cases

icewind1991:
In some cases, the root setup is triggered before the `filesystem` apps are loaded. These are then loading during the filesystem wrapper setup, if the loading of an app attempts to interact with the filesystem, it will attempt to setup the filesystem recursively.

This recursion is prevented by the `SetupManager`, but that also means that the app loading code sees the filesystem with no mounts setup. Which will lead to errors.

By moving the wrapper setup to before the recursion setup we can prevent this issue

### DIFF
--- a/apps/files_sharing/lib/External/Manager.php
+++ b/apps/files_sharing/lib/External/Manager.php
@@ -606,6 +606,10 @@ class Manager {
 			$this->logger->error('Mount point to remove share not found', ['mountPoint' => $mountPoint]);
 			return false;
 		}
+		if (!$mountPointObj instanceof Mount) {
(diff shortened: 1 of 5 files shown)


nfebe (approved):

mr-field (approved):
I've not worked on Nextcloud before, but the changes look sensible to me. I've tested this on my setup and it successfully fixed #32194.

tgoeg:
This fixes active breakage of production systems and updates to those, the corresponding bug is already 1,5y open.
The review does not seem to be a big thing as the changes are minimal.
@ArtificialOwl, @sorbaugh: Can you try to fit this in in your schedule? Much appreciated.

sorbaugh (approved):

tgoeg:
Thanks a lot, much appreciated!

PR #39049, opened 2023-06-28
Update aria-expanded attribute for sidebar apps-slide-toggles

nfebe:
Some apps which use the machanism provided in `core/src/OC/apps.js` for toggle slide up/down in the side bar would equally rely on same for `aria-expanded` state change.

For example : https://github.com/nextcloud/activity/pull/1230


susnux (approved):
Lgtm!

skjnldsv (approved):

PR #39052, opened 2023-06-29
Allow null capabilities

provokateurin:
## Summary

From https://github.com/nextcloud/server/pull/36666

Allows apps to return null capabilities. This is needed to correctly type the capabilities and let apps return null for certain cases.

## Checklist

### DIFF
--- a/lib/private/CapabilitiesManager.php
+++ b/lib/private/CapabilitiesManager.php
@@ -71,7 +71,7 @@ class CapabilitiesManager {
 						// that we would otherwise inject to every page load
 						continue;
 					}
-					$capabilities = array_replace_recursive($capabilities, $c->getCapabilities());
+					$capabilities = array_replace_recursive($capabilities, $c->getCapabilities() ?? []);
 				}
 			} else {
 				throw new \InvalidArgumentException('The given Capability (' . get_class($c) . ') does not implement the ICapability interface');
--- a/lib/public/Capabilities/ICapability.php
+++ b/lib/public/Capabilities/ICapability.php
@@ -37,7 +37,7 @@ interface ICapability {
 	/**
 	 * Function an app uses to return the capabilities
 	 *
-	 * @return array Array containing the apps capabilities
+	 * @return ?array Array containing the apps capabilities
 	 * @since 8.2.0
 	 */
 	public function getCapabilities();


ChristophWurst:
> Allows apps to return null capabilities. This is needed to correctly type the capabilities and let apps return null for certain cases.

Can't apps return [] when they have no capabilities? Do null and [] have distinguished semantics, e.g. "unknown" vs "none".

provokateurin:
No, because of the array handling in psalm. If you return `[]` then you need to type it as `array<never, never>` which would turn into `[]` in json. Instead you could return `new \stdClass()`, but then the union type would look like `\stdClass|array{...}` which is also wrong, because in json it will just be the same with all the keys missing. It's a complicated matter that popped up many times during my OpenAPI work and the easiest way is to just return null.

provokateurin:
The app where I encountered this problem is Talk: https://github.com/nextcloud/spreed/blob/a77c3f54b190a0b211328c68849fdb65c19e9cd4/lib/Capabilities.php#L58

nickvergessen:
I don't get how your patch would solve the problem? Or would the next request be to ask Talk to not return an empty array but null instead (API break?)? (But then the code you added here `?? []` would make it an empty array again, so you didn't gain anything?

nickvergessen:
Talk itself also has keys which are arrays that can be empty: https://github.com/nextcloud/spreed/blob/a77c3f54b190a0b211328c68849fdb65c19e9cd4/lib/Capabilities.php#L137-L138

provokateurin:
> I don't get how your patch would solve the problem? Or would the next request be to ask Talk to not return an empty array but null instead (API break?)? (But then the code you added here ?? [] would make it an empty array again, so you didn't gain anything?

That's the point, in the end it doesn't change anything. This is only to make the typing work properly.

provokateurin:
I could do some type hackery in openapi-extractor, but that then depends on how you interpret the union type of `\stdClass|array{...}`. I'd rather not do it and do this instead.

provokateurin:
Sorry, I just realized this patch is not going to work as I intended :sweat_smile: . Type hackery it is then

ChristophWurst:
Check out https://psalm.dev/r/d635121f7c. I don't think hacks are necessary :)

provokateurin:
Yeah that works, but it doesn't let me know how the data structure inside looks :/ That's what I need to know

ChristophWurst:
Submit as https://github.com/nextcloud/server/pull/39057 nevertheless because even if it's not specific enough for your use case it makes it easier to implement the interface for app devs.

PR #39057, opened 2023-06-29
feat(ocp): More specific type for ICapabilities::getCapabilities

ChristophWurst:
## Summary

Be more specific about what we expect the method to return. We can become more specific in the future. For now this seems generic enough to allow the existing implementation to stay correct. This should not be a breaking change: https://psalm.dev/r/d635121f7c

Idea sparked in https://github.com/nextcloud/server/pull/39052#issuecomment-1612550369.

image

## TODO

- [x] Improve typing

## Checklist

### DIFF
--- a/lib/public/Capabilities/ICapability.php
+++ b/lib/public/Capabilities/ICapability.php
@@ -25,11 +25,8 @@ namespace OCP\Capabilities;
  * Minimal interface that has to be implemented for a class to be considered
  * a capability.
  *
- * In an application use:
- *   $this->getContainer()->registerCapability('OCA\MY_APP\Capabilities');
- * To register capabilities.
- *
- * The class 'OCA\MY_APP\Capabilities' must then implement ICapability
+ * In an application use \OCP\AppFramework\Bootstrap\IRegistrationContext::registerCapability
+ * to register capabilities.
  *
  * @since 8.2.0
  */
--- a/lib/public/Capabilities/ICapability.php
+++ b/lib/public/Capabilities/ICapability.php
@@ -37,7 +37,19 @@ interface ICapability {
 	/**
 	 * Function an app uses to return the capabilities
 	 *
-	 * @return array Array containing the apps capabilities
+	 * ```php
+	 * return [
+	 *     'myapp' => [
+	 *         'awesomefeature' => true,
+	 *         'featureversion' => 3,
+	 *     ],
+	 *     'morecomplex' => [
+	 *         'a' => [1, 2],
+	 *     ],
+	 * ];
+	 * ```
+	 *
+	 * @return array<string, array<string, mixed>> Indexed array containing the app's capabilities
 	 * @since 8.2.0
 	 */
 	public function getCapabilities();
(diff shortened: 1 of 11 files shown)


provokateurin (approved):

lib/public/Capabilities/ICapability.php
```diff
@@ -37,7 +37,7 @@ interface ICapability {
 	/**
 	 * Function an app uses to return the capabilities
 	 *
-	 * @return array Array containing the apps capabilities
+	 * @return array<string, array<string, mixed>> Indexed array containing the app's capabilities
```
come-nc, line 40: An example would be nice since this is a bit cryptic

come-nc (approved):

fsamapoor (approved):
It's always a treat to see more specificity.

lib/public/Capabilities/ICapability.php
nickvergessen: ERROR: [InvalidReturnStatement](https://psalm.dev/128) - 18:16 - The inferred type 'array{config: array{attachments: array{allowed: true}, call: array{'breakout-rooms': false, 'supported-reactions': list{'❤️', '🎉', '👏', '👍', '👎', '😂', '🤩', '🤔', '😲', '😥'}, enabled: false, recording: false}, chat: array{'max-length': 12, 'read-privacy': 1, 'typing-privacy': 12, translations: list{array{from: 'de', to: 'en'}}}, conversations: array{'can-create': true}, previews: array{'max-gif-size': 123}, signaling: array{'session-ping-limit': 200}}, features: list{'audio', 'video'}, version: 'abc'}' does not match the declared return type 'array<string, array<string, mixed>>' for CTalk::foo

ERROR: [InvalidReturnType](https://psalm.dev/011) - 17:28 - The declared return type 'array<string, array<string, mixed>>' for CTalk::foo is incorrect, got 'array{config: array{attachments: array{allowed: true}, call: array{'breakout-rooms': false, 'supported-reactions': list{'❤️', '🎉', '👏', '👍', '👎', '😂', '🤩', '🤔', '😲', '😥'}, enabled: false, recording: false}, chat: array{'max-length': 12, 'read-privacy': 1, 'typing-privacy': 12, translations: list{array{from: 'de', to: 'en'}}}, conversations: array{'can-create': true}, previews: array{'max-gif-size': 123}, signaling: array{'session-ping-limit': 200}}, features: list{'audio', 'video'}, version: 'abc'}'

https://psalm.dev/r/ee1f4419a5

come-nc: So, at least version has string value and not array, and features has int keys and not string.

`array<string, mixed>` then?

provokateurin: But on the top level it has to return `["spreed" => ["bla" => "bla"]]`. The code snippet is wrong @nickvergessen.

nickvergessen: Right, I missed that line at the end

nickvergessen: Psalm output (using commit [086b943](https://github.com/vimeo/psalm/commit/086b943)): 

No issues!

ChristophWurst:
Implementations overwriting the phpdoc have to be adjusted because `@return array` erases the specific array type and therefore causes a Psalm error

provokateurin:
@ChristophWurst ready to go? This PR would actually help me right now :)

ChristophWurst:
Rebased and squashed

come-nc (approved):

PR #39074, opened 2023-06-29
Fix a bug in the recursive rmdir

ajabep:
## Summary

The rmdirr tried to delete a non-empty directory, thus, it was failing.

______

A directory containing a directory containing files where never deleted.

Example:

Let's consider this configuration:

```
/tmp/root
└── intermediate
    ├── file1
    ├── file2
    └── file3
```

This configuration was able to occured during a plugin update.

If NC was running a `rmdirr("/tmp/root");` (which seems to be done during a cron, but, that's not important), it would never deleted anything.

Indeed, the `$files` variable would contians only the `intermediate` directory.

Thus, during the loop to delete the content of the `/tmp/root` content, when `$fileInfo` will contains the descriptor of `/tmp/root/intermediate`, the `rmdir` PHP function would fails because this directory is not empty.... And maked the final `rmdir` (in the `$deleteSelf` condition).

Tested on `PHP 8.1.20 (built: Jun 15 2023 01:23:25) (NTS)`, the PHP version of the docker container provided by NC-AIO.

## Checklist

* Code should be [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting) (linted using php -l and tested in prod on my server)
* [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
* Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are not included. I am currently testing it on my server, and it seems that does not break. I also have doubt about the presence of unit testing on this topic, else, the TypeError should never become in prod.
* No screenshots before/after for front-end changes: not a UI change.
* No Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) is required: not a new feature.
* No [Backports requested](https://github.com/nextcloud/backportbot/#usage), but feel free to request it.

### DIFF
--- a/lib/private/legacy/OC_Helper.php
+++ b/lib/private/legacy/OC_Helper.php
@@ -173,7 +173,7 @@ class OC_Helper {
 				if ($fileInfo->isLink()) {
 					unlink($fileInfo->getPathname());
 				} elseif ($fileInfo->isDir()) {
-					rmdir($fileInfo->getRealPath());
+					self::rmdirr($fileInfo->getRealPath());
 				} else {
 					unlink($fileInfo->getRealPath());
 				}


icewind1991:
The `RecursiveIteratorIterator::CHILD_FIRST` should cause any items in the folder to be deleted before that `rmdir` is hit

ajabep:
> The `RecursiveIteratorIterator::CHILD_FIRST` should cause any items in the folder to be deleted before that `rmdir` is hit

Actually, maybe it's a but in PHP (version used in your docker images), maybe it's a bad documentation, but, it was also what I thought initially ;-)

kesselb:
We did a lot of debugging about iterators in https://github.com/nextcloud/updater/pull/510 / https://github.com/php/php-src/issues/12574. 

It appears that in some constellations, it's not safe to modify the file system while iterating.
I assume that is also the case here, and therefore should be fixed by iterating first and modifying afterward. 

For example: https://github.com/nextcloud/updater/blob/55eca48fc6ed8b0f773dc3f14db84befa9ee1da9/lib/Updater.php#L776-L816 or by using `iterator_to_array`.

Issue #37754, opened 2023-04-15
DAV properties propertyvalue column has wrong column type


claywd:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

converting database from mysql to postgres fails for (more than one) reason but specifically this time because of encoding issues between the tables with error

An exception occurred while executing a query: SQLSTATE[22021]: Character not in repertoire: 7 ERROR:  invalid byte sequence for encoding "UTF8": 0x00
  CONTEXT:  unnamed portal parameter $5

### Steps to reproduce

1. Setup nextcloud with remote mysql as the db source
2. Setup a remote postgresql, create nextcloud database, nextcloud user, grant all privs on the db to the nextcloud user
3. run the command `./occ db:convert-type --password "redacted" pgsql nextcloud pg_db_name nextcloud` as the www-data user within the nextcloud directory



### Expected behavior

New database tables and schema are configured and populated in new postgresql db with transformed from previous mysql db. 

### Installation method

Community Docker image

### Nextcloud Server version

24

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Nginx

### Database engine version

PostgreSQL

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "memcache.local": "\\OC\\Memcache\\APCu",
        "apps_paths": [
            {
                "path": "\/var\/www\/html\/apps",
                "url": "\/apps",
                "writable": false
            },
            {
                "path": "\/var\/www\/html\/custom_apps",
                "url": "\/custom_apps",
                "writable": true
            }
        ],
[...]

Issue #9907, opened 2018-06-19
Error reporting for too long filepaths (>255 chars)


jzaers:
### Steps to reproduce
1. Use filenames with a resulting path longer than 255 characters
2. This seems to result in various strange side effects like move not executed correctly and 
the following error message:

```
[...]

PR #39084, opened 2023-06-30
Fix dav properties column types

tcitworld:
## Summary

- First commit changes column `propertyvalue` from `TEXT` type to `BLOB` as it may contain null 0x00 characters because of php serialization #37754
  - after removing all serialized data (that have very good chances to be invalid) 
  - by creating a new temporary column
  - moving data to it
  - dropping the old column
  - adding a new column with the right type
  - moving data to it
  - resetting not null on the new column
  - dropping the temporary column
- Second one increases simply `propertypath` column length #9907

If changing the type of the column can't be done, we'll need to create a new column and move data

## TODO

- [x] SQLite is accepting pretty much anything ✅
- [x] MySQL/MariaDB ✅ just needs the index to be on a prefix
- [x] Test on PostgreSQL ❌ (needs some cast magic)
- [x] Test on Oracle ❌ (is being Oracle)

## Checklist


tcitworld:
@skjnldsv This one is important and I need some help on the database migration.

tcitworld:
For PostgreSQL : https://stackoverflow.com/questions/19300113/unable-to-convert-postgresql-text-column-to-bytea
For Oracle : https://asktom.oracle.com/ords/f?p=100:11:::::P11_QUESTION_ID:510622111991

If we have less than 50k rows (value yet to determine), perform the migrations straight away.
Otherwise, skip the migration, set an appconfig and add a red flag in system check. Then include as an expensive repair step.

tcitworld:
Fixed through https://github.com/nextcloud/server/pull/49528

PR #39086, opened 2023-06-30
feat(admin_audit): Migrate to non-deprecated IEventDispatcher

nickvergessen:
One step closer for #38546 

- [x] Based on https://github.com/nextcloud/server/pull/39089

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #39089, opened 2023-06-30
Fix event names of 2FA related typed events

### LINKED ISSUE #39087: Events of TwoFactor provider and manager names are very confusing (at best)
The names of the typed events are in my opinion wrong and will lead to wrong assumptions:

## Case 1

* Dispatched when a user successfully passes a 2FA challange:
* https://github.com/nextcloud/server/blob/d9f85220038b2421870a36f9b6a860d6444a988a/lib/private/Authentication/TwoFactorAuth/Manager.php#L285-L292
* Legacy dispatcher: IProvider::EVENT_SUCCESS
* Typed: OCP\Authentication\TwoFactorAuth\TwoFactorProviderForUserEnabled
    * 😕  ❓  Was the provider disabled for a user?!
    * 💥 No, it auth was passed and a better name would be OCP\Authentication\TwoFactorAuth\TwoFactorProviderChallangePassed
    * The old event should be deprecated immediately and both events be dispatched for the time being

## Case 2

* Dispatched when a user does not pass a 2FA challange:
* https://github.com/nextcloud/server/blob/d9f85220038b2421870a36f9b6a860d6444a988a/lib/private/Authentication/TwoFactorAuth/Manager.php#L295-L301
* Legacy dispatcher: IProvider::EVENT_FAILED
* Typed: OCP\Authentication\TwoFactorAuth\TwoFactorProviderForUserDisabled
    * 😕  ❓ Was the provider disabled for a user?!
    * 💥 No, it auth was failed and a better name would be OCP\Authentication\TwoFactorAuth\TwoFactorProviderChallangeFailed
    * The old event should be deprecated immediately and both events be dispatched for the time being

## Case 3

* Dispatched when a user is being deleted inside a listener for UserDeletedEvent
* https://github.com/nextcloud/server/blob/f5c361cf44739058b79f322576a1bad2d8c142d9/lib/private/Authentication/Listeners/UserDeletedStoreCleanupListener.php#L44-L50
* https://github.com/nextcloud/server/blob/f5c361cf44739058b79f322576a1bad2d8c142d9/lib/private/Authentication/TwoFactorAuth/Registry.php#L68-L73
* Legacy dispatcher: none
* Typed: OCP\Authentication\TwoFactorAuth\TwoFactorProviderDisabled
    * 😕  ❓ Was the provider disabled for everyone?!
    * 💥 No, all entries of a single user get deleted OCP\Authentication\TwoFactorAuth\TwoFactorProviderForUserDisabled
        * But that name is from case 2, so need a different one Maybe "TwoFactorProviderUnregisteredForUser"
    * The old event should be deprecated immediately and both events be dispatched for the time being

nickvergessen:
* Resolves: #39087 

## Checklist


apps/admin_audit/composer/composer/installed.php
```diff
@@ -13,7 +13,7 @@
         '__root__' => array(
             'pretty_version' => 'dev-master',
             'version' => 'dev-master',
-            'reference' => 'dd3d689e04a5e1d558da937ca72980e0e2c7c404',
+            'reference' => 'b1797842784b250fb01ed5e3bf130705eb94751b',
```
marcelklehr, line 16: do we check in these changes?

lib/public/Authentication/TwoFactorAuth/TwoFactorProviderForUserRegistered.php
```diff
@@ -0,0 +1,34 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCP\Authentication\TwoFactorAuth;
+
+/**
+ * @since 28.0.0
+ */
+class TwoFactorProviderForUserRegistered extends RegistryEvent {
```
marcelklehr, line 33: RegistryEvent is marked as deprecated. If we really remove it in the future, this will no longer be able to inherit from it, no?

nickvergessen: I deprecate it in this PR because it should not be used directly anymore or as a typehint.
We can also duplicate the content and put it into the new events. Both is fine for me.

apps/admin_audit/composer/composer/installed.php
nickvergessen: If the autoloader changes, I commit it :P

marcelklehr: Will probably cause happy backporting issues :D

lib/public/Authentication/TwoFactorAuth/TwoFactorProviderForUserRegistered.php
marcelklehr: Not extending it in non-deprecated classes seems cleaner to me, yeah

apps/admin_audit/composer/composer/installed.php
nickvergessen: I'm not sure we backport this. Would allow earlier transition, yes. But deprecation period would be unchanged, so 🤷‍♂️

nickvergessen:
Come and Christoph approved the commit in https://github.com/nextcloud/server/pull/39086 already, so setting to merge

ChristophWurst (approved):

come-nc (approved):

marcelklehr (approved):

nickvergessen:
Forgot the since on the `public function __construct()` methods

nickvergessen:

PR #39108, opened 2023-07-02
Refactors lib/private/Lock

fsamapoor:
## Summary
Following [previous PRs taking advantage of PHP8's constructor property promotion](https://github.com/nextcloud/server/pulls?q=is%3Apr+author%3Afsamapoor+Uses+PHP8%27s+constructor+property+promotion) in `/core/` namespace, I have also made the required adjustments to the classes in `/lib/private/Lock` namespace.

The improvements in this PR include:

- Using match expression
- Adding types to properties

## Checklist

### DIFF
--- a/lib/private/Lock/DBLockingProvider.php
+++ b/lib/private/Lock/DBLockingProvider.php
@@ -39,21 +39,15 @@ use OCP\Lock\LockedException;
  * Locking provider that stores the locks in the database
  */
 class DBLockingProvider extends AbstractLockingProvider {
-	private IDBConnection $connection;
-	private ITimeFactory $timeFactory;
 	private array $sharedLocks = [];
-	private bool $cacheSharedLocks;
 
 	public function __construct(
-		IDBConnection $connection,
-		ITimeFactory $timeFactory,
+		private IDBConnection $connection,
+		private ITimeFactory $timeFactory,
 		int $ttl = 3600,
-		bool $cacheSharedLocks = true
+		private bool $cacheSharedLocks = true
 	) {
-		$this->connection = $connection;
-		$this->timeFactory = $timeFactory;
-		$this->ttl = $ttl;
-		$this->cacheSharedLocks = $cacheSharedLocks;
+		parent::__construct($ttl);
 	}
 
 	/**
--- a/lib/private/Lock/MemcacheLockingProvider.php
+++ b/lib/private/Lock/MemcacheLockingProvider.php
@@ -32,11 +32,11 @@ use OCP\IMemcacheTTL;
 use OCP\Lock\LockedException;
 
 class MemcacheLockingProvider extends AbstractLockingProvider {
-	private IMemcache $memcache;
-
-	public function __construct(IMemcache $memcache, int $ttl = 3600) {
-		$this->memcache = $memcache;
-		$this->ttl = $ttl;
+	public function __construct(
+		private IMemcache $memcache,
+		int $ttl = 3600,
+	) {
+		parent::__construct($ttl);
 	}
 
 	private function setTTL(string $path): void {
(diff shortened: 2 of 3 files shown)


nfebe (approved):
LGTM, thank you!

lib/private/Lock/DBLockingProvider.php
```diff
@@ -39,21 +39,14 @@
...
-	private bool $cacheSharedLocks;
 
 	public function __construct(
-		IDBConnection $connection,
-		ITimeFactory $timeFactory,
-		int $ttl = 3600,
-		bool $cacheSharedLocks = true
+		private IDBConnection $connection,
+		private ITimeFactory $timeFactory,
+		protected int $ttl = 3600,
```
come-nc, line 47: ttl property is from parent class, it’s ugly to redeclare it here.

come-nc:
```suggestion
		int $ttl = 3600,
```
Maybe add a constructor to parent class to be able to call parent::__construct($ttl) ?

lib/private/Lock/MemcacheLockingProvider.php
```diff
@@ -32,11 +32,10 @@
...
 
 class MemcacheLockingProvider extends AbstractLockingProvider {
-	private IMemcache $memcache;
-
-	public function __construct(IMemcache $memcache, int $ttl = 3600) {
-		$this->memcache = $memcache;
-		$this->ttl = $ttl;
+	public function __construct(
+		private IMemcache $memcache,
+		protected int $ttl = 3600,
```
come-nc, line 37: Same

come-nc (changes requested):
I dislike the match uses, I find them a lot less readable.

Also there is a problem with how the $ttl property is handled, see comments.

fsamapoor:
> I dislike the match uses, I find them a lot less readable.
> 
> Also there is a problem with how the $ttl property is handled, see comments.

Thank you, Come, for reviewing the PR. I really appreciate the timely feedback. I'm suggesting changes that I believe will improve code readability. However, I have no hard feelings if you prefer to revert or disregard them, as you are the maintainer.

Having said that, I've received multiple positive feedbacks in previous PRs regarding the use of `match` statements, and I personally try to minimize the use of `else` statements as much as I can.

I've moved the `ttl` property to the parent class. Please review the changes and let me know if I should revert the `match` statements.

come-nc:
> Thank you, Come, for reviewing the PR. I really appreciate the timely feedback. I'm suggesting changes that I believe will improve code readability. However, I have no hard feelings if you prefer to revert or disregard them, as you are the maintainer.
> 
> Having said that, I've received multiple positive feedbacks in previous PRs regarding the use of `match` statements, and I personally try to minimize the use of `else` statements as much as I can.
> 
> I've moved the `ttl` property to the parent class. Please review the changes and let me know if I should revert the `match` statements.

Yes please revert the `match` statements. There is no need for minimizing the `else` statements, `else` is a perfectly fine language construct.
I only find `match` interesting for simple cases where an input value directly implies an output value. When there is more logic than that it quickly looks like perl oneliners.
And I especially find `match(true)` really non-intuitive.

fsamapoor:
> > Thank you, Come, for reviewing the PR. I really appreciate the timely feedback. I'm suggesting changes that I believe will improve code readability. However, I have no hard feelings if you prefer to revert or disregard them, as you are the maintainer.
> > Having said that, I've received multiple positive feedbacks in previous PRs regarding the use of `match` statements, and I personally try to minimize the use of `else` statements as much as I can.
> > I've moved the `ttl` property to the parent class. Please review the changes and let me know if I should revert the `match` statements.
> 
> Yes please revert the `match` statements. There is no need for minimizing the `else` statements, `else` is a perfectly fine language construct. I only find `match` interesting for simple cases where an input value directly implies an output value. When there is more logic than that it quickly looks like perl oneliners. And I especially find `match(true)` really non-intuitive.

Done. Thank you for the explanation. Squashed the commits as well for a clean history.

come-nc (approved):

PR #39109, opened 2023-07-02
Refactors /Metadata, /Migration, and /Net namespaces in /lib/private/

fsamapoor:
## Summary

Following [previous PRs taking advantage of PHP8's constructor property promotion](https://github.com/nextcloud/server/pulls?q=is%3Apr+author%3Afsamapoor+Uses+PHP8%27s+constructor+property+promotion) in `/core/` namespace, I have also made the required adjustments to the classes in `/lib/private/Metadata`, `/lib/private/Migration`, and `/lib/private/Net` namespaces.

The improvements in this PR include:

- Adding return types
- Adding types to properties

## Checklist

### DIFF
--- a/lib/private/Migration/BackgroundRepair.php
+++ b/lib/private/Migration/BackgroundRepair.php
@@ -41,15 +41,13 @@ use Psr\Log\LoggerInterface;
  * @package OC\Migration
  */
 class BackgroundRepair extends TimedJob {
-	private IJobList $jobList;
-	private LoggerInterface $logger;
-	private IEventDispatcher $dispatcher;
-
-	public function __construct(IEventDispatcher $dispatcher, ITimeFactory $time, LoggerInterface $logger, IJobList $jobList) {
+	public function __construct(
+		private IEventDispatcher $dispatcher,
+		ITimeFactory $time,
+		private LoggerInterface $logger,
+		private IJobList $jobList,
+	) {
 		parent::__construct($time);
-		$this->dispatcher = $dispatcher;
-		$this->logger = $logger;
-		$this->jobList = $jobList;
 		$this->setInterval(15 * 60);
 	}
 
@@ -58,7 +56,7 @@ class BackgroundRepair extends TimedJob {
 	 * @throws \Exception
 	 * @throws \OC\NeedsUpdateException
 	 */
-	protected function run($argument) {
+	protected function run($argument): void {
 		if (!isset($argument['app']) || !isset($argument['step'])) {
 			// remove the job - we can never execute it
 			$this->jobList->remove($this, $this->argument);
@@ -101,7 +99,7 @@ class BackgroundRepair extends TimedJob {
 	 * @param $app
 	 * @throws NeedsUpdateException
 	 */
-	protected function loadApp($app) {
+	protected function loadApp($app): void {
 		OC_App::loadApp($app);
 	}
 }
--- a/lib/private/Migration/ConsoleOutput.php
+++ b/lib/private/Migration/ConsoleOutput.php
@@ -34,34 +34,31 @@ use Symfony\Component\Console\Output\OutputInterface;
  * @package OC\Migration
  */
 class ConsoleOutput implements IOutput {
-	/** @var OutputInterface */
-	private $output;
+	private ?ProgressBar $progressBar = null;
 
-	/** @var ProgressBar */
-	private $progressBar;
-
-	public function __construct(OutputInterface $output) {
-		$this->output = $output;
+	public function __construct(
+		private OutputInterface $output,
+	) {
 	}
 
 	/**
 	 * @param string $message
 	 */
-	public function info($message) {
+	public function info($message): void {
 		$this->output->writeln("<info>$message</info>");
 	}
 
 	/**
 	 * @param string $message
 	 */
-	public function warning($message) {
+	public function warning($message): void {
 		$this->output->writeln("<comment>$message</comment>");
 	}
 
 	/**
 	 * @param int $max
 	 */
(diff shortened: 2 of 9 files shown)


lib/private/Migration/ConsoleOutput.php
```diff
@@ -34,34 +34,31 @@
...
+	public function __construct(
+		private OutputInterface $output,
+	) {
 	}
 
 	/**
 	 * @param string $message
 	 */
-	public function info($message) {
+	public function info($message): void {
```
Fenn-CS, line 47:
```suggestion
	public function info(string $message): void {
```

lib/private/Migration/ConsoleOutput.php
```diff
@@ -73,7 +70,7 @@ public function startProgress($max = 0) {
 	 * @param int $step
 	 * @param string $description
 	 */
-	public function advance($step = 1, $description = '') {
+	public function advance($step = 1, $description = ''): void {
```
Fenn-CS, line 73:
```suggestion
	public function advance(int $step = 1, string $description = ''): void {
```

lib/private/Migration/SimpleOutput.php
```diff
@@ -33,48 +33,46 @@
...
+	public function warning($message): void {
 		$this->logger->warning($message, ['app' => $this->appName]);
 	}
 
 	/**
 	 * @param int $max
 	 * @since 9.1.0
 	 */
-	public function startProgress($max = 0) {
+	public function startProgress($max = 0): void {
```
Fenn-CS, line 62:
```suggestion
	public function startProgress(int $max = 0): void {
```

lib/private/Migration/SimpleOutput.php
```diff
@@ -33,48 +33,46 @@
...
 	}
 
 	/**
 	 * @param int $step
 	 * @param string $description
 	 * @since 9.1.0
 	 */
-	public function advance($step = 1, $description = '') {
+	public function advance($step = 1, $description = ''): void {
```
Fenn-CS, line 70:
```suggestion
	public function advance(int $step = 1, string $description = ''): void {
```

lib/private/Migration/ConsoleOutput.php
```diff
@@ -34,34 +34,31 @@
...
-	public function warning($message) {
+	public function warning($message): void {
 		$this->output->writeln("<comment>$message</comment>");
 	}
 
 	/**
 	 * @param int $max
 	 */
-	public function startProgress($max = 0) {
+	public function startProgress($max = 0): void {
```
Fenn-CS, line 61:
```suggestion
	public function startProgress(int $max = 0): void {
```

lib/private/Migration/BackgroundRepair.php
```diff
@@ -58,7 +56,7 @@ public function __construct(IEventDispatcher $dispatcher, ITimeFactory $time, Lo
 	 * @throws \Exception
 	 * @throws \OC\NeedsUpdateException
 	 */
-	protected function run($argument) {
+	protected function run($argument): void {
```
Fenn-CS, line 59:
```suggestion
	protected function run(array $argument): void {
```

nfebe (changes requested):
You added return types to some functions so you might want to add type hints as well.

Please do verify, that the suggestions I made tie with the expected function types, for example `$app` does not have a function annotation at the top but I am guessing it would be a string.

lib/private/Migration/ConsoleOutput.php
fsamapoor: I would love to, but this is not something that could be done right now. The interfaces do not have parameter type hints at the moment, and changing the interfaces would cause an absolute overhaul in the codebase.

Screenshot from 2023-07-14 11-31-09


I do my best to make improvements as much as I can, with the hope that eventually, the codebase would evolve into much more strict and precise definitions.

Now that we're talking about it, are there any plans in place to update the interfaces with type hints and return types?

lib/private/Migration/ConsoleOutput.php
fsamapoor: Addressed above.

lib/private/Migration/SimpleOutput.php
fsamapoor: Addressed above.

lib/private/Migration/SimpleOutput.php
fsamapoor: Addressed above.

lib/private/Migration/ConsoleOutput.php
fsamapoor: Addressed above.

lib/private/Migration/BackgroundRepair.php
fsamapoor: Same. Except for this time, it is an abstract class that has not parameter type hint for the `run` method and not an interface.

fsamapoor: Same. Except for this time, it is an abstract class that does not have the type hint for the `run` method's parameter an not the interface.

lib/private/Migration/ConsoleOutput.php
Fenn-CS: No worries for this then.

lib/private/Migration/BackgroundRepair.php
Fenn-CS: No worries.

nfebe (approved):

artonge (approved):

ChristophWurst (approved):

PR #39125, opened 2023-07-03
Rewrite OCS CSRF check to be readable

provokateurin:
## Summary

Previous code was unreadable and every time I come across it I had to read it multiple times to understand what it does. Should be much clearer now (at least it is to me).

## Checklist

### DIFF
--- a/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
@@ -206,7 +206,7 @@ class SecurityMiddleware extends Middleware {
 		}
 		// CSRF check - also registers the CSRF token since the session may be closed later
 		Util::callRegister();
-		if (!$this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class)) {
+		if ($this->isInvalidCSRFRequired($reflectionMethod)) {
 			/*
 			 * Only allow the CSRF check to fail on OCS Requests. This kind of
 			 * hacks around that we have no full token auth in place yet and we
@@ -215,12 +215,7 @@ class SecurityMiddleware extends Middleware {
 			 * Additionally we allow Bearer authenticated requests to pass on OCS routes.
 			 * This allows oauth apps (e.g. moodle) to use the OCS endpoints
 			 */
-			if (!$this->request->passesCSRFCheck() && !(
-				$controller instanceof OCSController && (
-					$this->request->getHeader('OCS-APIREQUEST') === 'true' ||
-					str_starts_with($this->request->getHeader('Authorization'), 'Bearer ')
-				)
-			)) {
+			if (!$controller instanceof OCSController || !$this->isValidOCSRequest()) {
 				throw new CrossSiteRequestForgeryException();
 			}
 		}
@@ -240,6 +235,19 @@ class SecurityMiddleware extends Middleware {
 		if ($appPath !== false && !$isPublicPage && !$this->appManager->isEnabledForUser($this->appName)) {
 			throw new AppNotEnabledException();
 		}
+	}
+
+	private function isInvalidCSRFRequired(ReflectionMethod $reflectionMethod): bool {
+		if ($this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class)) {
+			return false;
+		}
+
+		return !$this->request->passesCSRFCheck();
+	}
+
+	private function isValidOCSRequest(): bool {
+		return $this->request->getHeader('OCS-APIREQUEST') === 'true'
+			|| str_starts_with($this->request->getHeader('Authorization'), 'Bearer ');
 	}
 
 	/**


lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -215,12 +215,7 @@ public function beforeController($controller, $methodName) {
 			 * Additionally we allow Bearer authenticated requests to pass on OCS routes.
 			 * This allows oauth apps (e.g. moodle) to use the OCS endpoints
 			 */
-			if (!$this->request->passesCSRFCheck() && !(
-				$controller instanceof OCSController && (
-					$this->request->getHeader('OCS-APIREQUEST') === 'true' ||
-					str_starts_with($this->request->getHeader('Authorization'), 'Bearer ')
-				)
-			)) {
+			if (!$this->request->passesCSRFCheck() && $controller instanceof OCSController && ($this->request->getHeader('OCS-APIREQUEST') !== 'true' || !str_starts_with($this->request->getHeader('Authorization'), 'Bearer '))) {
```
provokateurin, line 218: Changing the order of arguments such that the CSRF check is the last check should lead to better performance on average. Yay or nay?

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -215,13 +215,10 @@ public function beforeController($controller, $methodName) {
...
 			 */
-			if (!$this->request->passesCSRFCheck() && !(
-				$controller instanceof OCSController && (
-					$this->request->getHeader('OCS-APIREQUEST') === 'true' ||
-					str_starts_with($this->request->getHeader('Authorization'), 'Bearer ')
-				)
-			)) {
-				throw new CrossSiteRequestForgeryException();
+			if (!$this->request->passesCSRFCheck()) {
+				if (!$controller instanceof OCSController || ($this->request->getHeader('OCS-APIREQUEST') !== 'true' || !str_starts_with($this->request->getHeader('Authorization'), 'Bearer '))) {
```
provokateurin, line 219: Still not great, but in the valid cases we just want to continue and can't just return

provokateurin:
This PR was hell, the previous logic was really not readable and understandable and comment was misleading...

blizzz (commented):
Personally, you would increase readability by breaking long lines down into many =) (but with leading operators).

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -206,21 +206,9 @@ public function beforeController($controller, $methodName) {
...
-			 * This allows oauth apps (e.g. moodle) to use the OCS endpoints
-			 */
-			if (!$this->request->passesCSRFCheck() && !(
-				$controller instanceof OCSController && (
-					$this->request->getHeader('OCS-APIREQUEST') === 'true' ||
-					str_starts_with($this->request->getHeader('Authorization'), 'Bearer ')
-				)
-			)) {
+		if (!$this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class) && !$this->request->passesCSRFCheck()) {
+			// If CSRF is required and the CSRF check does not pass, only allow OCS requests (OCS-APIREQUEST is true or Authorization is Bearer) to that go to an OCS controllers.
```
christianlupus, line 210:
```suggestion
			// CSRF is required and the CSRF check does not pass.
			// Thus, only allow OCS requests (OCS-APIREQUEST is true or Authorization is Bearer) that go to an OCS controllers.
```

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -206,21 +206,9 @@ public function beforeController($controller, $methodName) {
...
-			 * Additionally we allow Bearer authenticated requests to pass on OCS routes.
-			 * This allows oauth apps (e.g. moodle) to use the OCS endpoints
-			 */
-			if (!$this->request->passesCSRFCheck() && !(
-				$controller instanceof OCSController && (
-					$this->request->getHeader('OCS-APIREQUEST') === 'true' ||
-					str_starts_with($this->request->getHeader('Authorization'), 'Bearer ')
-				)
-			)) {
+		if (!$this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class) && !$this->request->passesCSRFCheck()) {
```
christianlupus, line 209:
```suggestion
		if (
			!$this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class)
			&& !$this->request->passesCSRFCheck()
		) {
```

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -206,21 +206,9 @@ public function beforeController($controller, $methodName) {
...
+			// If CSRF is required and the CSRF check does not pass, only allow OCS requests (OCS-APIREQUEST is true or Authorization is Bearer) to that go to an OCS controllers.
+			if (!$controller instanceof OCSController || ($this->request->getHeader('OCS-APIREQUEST') !== 'true' && !str_starts_with($this->request->getHeader('Authorization'), 'Bearer '))) {
```
christianlupus, line 211:
```suggestion
			if (
				!$controller instanceof OCSController
				|| (
					$this->request->getHeader('OCS-APIREQUEST') !== 'true'
					&& !str_starts_with($this->request->getHeader('Authorization'), 'Bearer ')
				)
			) {
```

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -206,21 +206,9 @@ public function beforeController($controller, $methodName) {
...
 		Util::callRegister();
-		if (!$this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class)) {
-			/*
-			 * Only allow the CSRF check to fail on OCS Requests. This kind of
-			 * hacks around that we have no full token auth in place yet and we
-			 * do want to offer CSRF checks for web requests.
-			 *
-			 * Additionally we allow Bearer authenticated requests to pass on OCS routes.
-			 * This allows oauth apps (e.g. moodle) to use the OCS endpoints
-			 */
```
christianlupus, line 217: There are some hints on why this was introduced here. Do we want to keep these in some way?

christianlupus (commented):
The changes I suggested are only cosmetic. As I used GitHub, I hope the indentations are correct and will satisfy any active cide style formatters.

One suggestion that might improve readability:
We could insert very small private methods that are just evaluating the `if` conditions. That would make it more readable, I guess. So, you would (for example) add a method
```php
private function isInvalidCSRFRequired($reflectionMethod) {
	if ($this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class) {
		return false;
	}

	return ! $this->request->passesCSRFCheck();
}
```
This is for sure more verbose but I guess more readable, especially as the main `beforeController` method will offload significant parts of the logic. Adding comments to various parts of the condition methods will allow more human understanding as well.
What do you think about this?

Apart from that I did not find any flaws with the logic. The behavior should be 100% the same.

In general, I think this increases the readability and that should be done in any case. (Maybe even more)

provokateurin:
@christianlupus I agree with you, it's a lot more readable when the logic is in a separate method. Then it's also easy to use early returns instead of all this mess.

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
provokateurin: They sound like it was only done for temporary reasons, but the code exists for a long time, so they probably don't apply anymore?

nickvergessen: Nothing works longer than a working intermediate solution.
Keeping the comment sounds helpful

christianlupus (approved):
This seems cleaner and I see no change in the implemented logic here.

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -242,6 +237,19 @@ public function beforeController($controller, $methodName) {
...
+		if ($this->hasAnnotationOrAttribute($reflectionMethod, 'NoCSRFRequired', NoCSRFRequired::class)) {
+			return false;
+		}
+
+		return !$this->request->passesCSRFCheck();
+	}
+
+	private function isValidOCSRequest(): bool {
+		return $this->request->getHeader('OCS-APIREQUEST') === 'true' ||
+			str_starts_with($this->request->getHeader('Authorization'), 'Bearer ');
```
julien-nc, line 250: Nitpicking, we mostly place boolean operators at line beginnings.
Otherwise LGTM.

julien-nc (approved):

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
provokateurin: wdym?

nickvergessen:
```suggestion
		return $this->request->getHeader('OCS-APIREQUEST') === 'true'
			|| str_starts_with($this->request->getHeader('Authorization'), 'Bearer ');
```

provokateurin:
How long do I have to rebase until this shit finally works :sweat_smile:

PR #39134, opened 2023-07-04
disable user unmount using 'allow_user_unmounting'

ArtificialOwl:
`./occ config:app:set files_external 'allow_user_unmounting' --value 'no'`
default value is 'yes'

- some migration to php8
- switching to `LoggerInterface`


ArtificialOwl:
(failing test does not seems related)

PR #39160, opened 2023-07-05
Migrate federated file sharing to PSR LoggerInterface

Altahrim:
See  #32127 

## Summary
Remove `OCP\ILogger` from `apps/federatedfilesharing` and use Psr\Log\LoggerInterface` instead.

### DIFF
--- a/apps/federatedfilesharing/lib/Controller/MountPublicLinkController.php
+++ b/apps/federatedfilesharing/lib/Controller/MountPublicLinkController.php
@@ -41,12 +41,12 @@ use OCP\Federation\ICloudIdManager;
 use OCP\HintException;
 use OCP\Http\Client\IClientService;
 use OCP\IL10N;
-use OCP\ILogger;
 use OCP\IRequest;
 use OCP\ISession;
 use OCP\IUserSession;
 use OCP\Share\IManager;
 use OCP\Share\IShare;
+use Psr\Log\LoggerInterface;
 
 /**
  * Class MountPublicLinkController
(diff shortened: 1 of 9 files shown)


apps/federatedfilesharing/lib/Controller/MountPublicLinkController.php
```diff
@@ -94,28 +69,22 @@ class MountPublicLinkController extends Controller {
 	 * @param IUserSession $userSession
 	 * @param IClientService $clientService
 	 * @param ICloudIdManager $cloudIdManager
+	 * @param LoggerInterface $logger
```
come-nc, line 72: You can remove the `@param` phpdoc tags when strong typing is used in the method prototype already.
Just keep `@param string $appName` if you do not strong type it, but I think you can strong type it as it’s the case in other controllers.

apps/federatedfilesharing/tests/Controller/MountPublicLinkControllerTest.php
```diff
@@ -48,48 +48,49 @@
 use OCP\IUserSession;
 use OCP\Share\IManager;
 use OCP\Share\IShare;
+use Psr\Log\LoggerInterface;
 
 class MountPublicLinkControllerTest extends \Test\TestCase {
-	/** @var IContactsManager|\PHPUnit\Framework\MockObject\MockObject */
+	/** @var IContactsManager|MockObject */
```
come-nc, line 54: You need to add a `use PHPUnit\Framework\MockObject\MockObject;`

apps/federatedfilesharing/tests/Controller/MountPublicLinkControllerTest.php
```diff
@@ -48,48 +48,49 @@
...
 	protected $contactsManager;
 
-	/** @var  MountPublicLinkController */
+	/** @var MountPublicLinkController */
 	private $controller;
 
-	/** @var  \OCP\IRequest | \PHPUnit\Framework\MockObject\MockObject */
+	/** @var \OCP\IRequest|MockObject */
```
come-nc, line 60:
```suggestion
	/** @var IRequest|MockObject */
```

apps/federatedfilesharing/lib/Controller/MountPublicLinkController.php
Altahrim: Done! 
I prefer it that way

come-nc (approved):

artonge (approved):

PR #39163, opened 2023-07-05
Refactor lib/private/Remote

shdehnavi:
## Summary
The required adjustments have been made to the classes in `/lib/private/Remote ` namespace.

The improvements:

- Adding return types
- Adding throws types
- Adding types to properties
- Using PHP8's `match` expression instead of `switch`

## Checklist

### DIFF
--- a/lib/private/Remote/Credentials.php
+++ b/lib/private/Remote/Credentials.php
@@ -38,14 +38,14 @@ class Credentials implements ICredentials {
 	/**
 	 * @return string
 	 */
-	public function getUsername() {
+	public function getUsername(): string {
 		return $this->user;
 	}
 
 	/**
 	 * @return string
 	 */
-	public function getPassword() {
+	public function getPassword(): string {
 		return $this->password;
 	}
 }
--- a/lib/private/Remote/Credentials.php
+++ b/lib/private/Remote/Credentials.php
@@ -25,18 +25,14 @@ namespace OC\Remote;
 use OCP\Remote\ICredentials;
 
 class Credentials implements ICredentials {
-	/** @var string */
-	private $user;
-	/** @var string */
-	private $password;
-
 	/**
 	 * @param string $user
 	 * @param string $password
 	 */
-	public function __construct($user, $password) {
-		$this->user = $user;
-		$this->password = $password;
+	public function __construct(
+		private string $user,
+		private string $password,
+	) {
 	}
 
 	/**
--- a/lib/private/Remote/Credentials.php
+++ b/lib/private/Remote/Credentials.php
@@ -31,9 +31,6 @@ class Credentials implements ICredentials {
 	) {
 	}
 
-	/**
-	 * @return string
-	 */
 	public function getUsername(): string {
 		return $this->user;
 	}
--- a/lib/private/Remote/Credentials.php
+++ b/lib/private/Remote/Credentials.php
@@ -25,10 +25,6 @@ namespace OC\Remote;
 use OCP\Remote\ICredentials;
 
 class Credentials implements ICredentials {
-	/**
-	 * @param string $user
-	 * @param string $password
-	 */
 	public function __construct(
 		private string $user,
 		private string $password,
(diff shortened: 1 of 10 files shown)


lib/private/Remote/Instance.php
```diff
@@ -32,55 +32,53 @@
...
 		return $this->getProtocol() . '://' . $this->getUrl();
 	}
 
 	/**
 	 * @return string The full version string in '13.1.2.3' format
+	 * @throws NotFoundException
 	 */
-	public function getVersion() {
+	public function getVersion(): string {
 		$status = $this->getStatus();
```
ChristophWurst, line 73: getStatus needs finer type annotation for Psalm to acknowledge that this method will always return a string

ChristophWurst: status might be `null`

shdehnavi: It's addressed in the next commit

lib/private/Remote/User.php
```diff
@@ -38,102 +38,100 @@ class User implements IUser {
...
-	public function getLanguage() {
+	public function getLanguage(): string {
 		return $this->data['language'];
 	}
 
 	/**
 	 * @return int
 	 */
-	public function getUsedSpace() {
+	public function getUsedSpace(): int {
```
nickvergessen, line 113: this and all other quota related fields can also be float

shdehnavi: It's addressed in the next commit

lib/private/Remote/Credentials.php
```diff
@@ -25,31 +25,27 @@
...
 class Credentials implements ICredentials {
-	/** @var string */
-	private $user;
-	/** @var string */
-	private $password;
-
 	/**
 	 * @param string $user
 	 * @param string $password
 	 */
```
nickvergessen, line 31:
```suggestion
```

lib/private/Remote/Credentials.php
```diff
@@ -25,31 +25,27 @@
...
-		$this->password = $password;
+	public function __construct(
+		private string $user,
+		private string $password,
+	) {
 	}
 
 	/**
 	 * @return string
 	 */
```
nickvergessen, line 40:
```suggestion
```

lib/private/Remote/Credentials.php
```diff
@@ -25,31 +25,27 @@
...
-	public function getUsername() {
+	public function getUsername(): string {
 		return $this->user;
 	}
 
 	/**
 	 * @return string
 	 */
```
nickvergessen, line 47:
```suggestion
```

lib/private/Remote/Instance.php
```diff
@@ -32,55 +32,53 @@
...
+	private ?array $status;
 
 	/**
 	 * @param string $url
 	 * @param ICache $cache
 	 * @param IClientService $clientService
 	 */
-	public function __construct($url, ICache $cache, IClientService $clientService) {
+	public function __construct(
+		$url,
```
nickvergessen, line 46:
```suggestion
		string $url,
```

lib/private/Remote/User.php
```diff
@@ -38,102 +38,100 @@ class User implements IUser {
...
-		$this->data = $data;
+	public function __construct(
+		private array $data,
+	) {
 	}
 
 
 	/**
 	 * @return string
 	 */
```
nickvergessen, line 49:
```suggestion
```

Also on all other getters.

nickvergessen (changes requested):

lib/private/Remote/User.php
```diff
@@ -38,102 +38,97 @@ class User implements IUser {
...
-	 * @return string
-	 */
-	public function getUserId() {
+	public function getUserId(): string {
 		return $this->data['id'];
 	}
 
 	/**
 	 * @return string
 	 */
```
nickvergessen, line 53:
```suggestion
```

Also the rest of the file

PR #39165, opened 2023-07-05
Enable all files_versions features when groupfolders is enabled

### LINKED ISSUE #39137: Make persistent versioning available even with groupfolders being activated
* Provide persistent versioning even with groupfolders being activated, yet just as is for non-groupfolders
* might be covered via groupfolders support [#39138](https://github.com/nextcloud/server/issues/39138)

artonge:
Fix https://github.com/nextcloud/server/issues/39137

-- |Before | After
-- | -- | --
In plain folder | Screenshot from 2023-07-05 15-54-16 | Screenshot from 2023-07-05 15-52-01
In group folder | Screenshot from 2023-07-05 15-54-22 | Screenshot from 2023-07-05 15-52-05


apps/files_versions/lib/Capabilities.php
```diff
@@ -44,13 +44,11 @@ public function __construct(
 	 * Return this classes capabilities
 	 */
 	public function getCapabilities() {
-		$groupFolderInstalled = $this->appManager->isInstalled('groupfolders');
-
 		return [
 			'files' => [
 				'versioning' => true,
-				'version_labeling' => !$groupFolderInstalled && $this->config->getSystemValueBool('enable_version_labeling', true),
-				'version_deletion' => !$groupFolderInstalled && $this->config->getSystemValueBool('enable_version_deletion', true),
```
artonge, line 53: @nextcloud/designers, should we drop those settings?

Pytal (approved):

apps/files_versions/lib/Capabilities.php
artonge: @szaimen or @nimishavijay ?

nimishavijay: Is there any particular reason to drop these settings? Version labelling seems like a pretty important feature, right? :)

nfebe (approved):

apps/files_versions/lib/Capabilities.php
artonge: Yes, but should an admin be able to disable labelling and deletion ?

tiredofit: I had thought that this PR was meant to _enable_ named versions when in a Group Folder. Is this just a PR to remove the functionality entirely even though it was touted as a launch day feature of NC 26?

PR #39168, opened 2023-07-05
remove no longer accessible shares

icewind1991:
When trying to access a share where the owner no longer has access to the source file, remove the share instead of only marking it as failed.

To test:

- Setup user `A`, `B` and `C`. Create a groupfolder `G` and give `A` and `B` access to it.
- User `B` uploads a file `F` to `G` and shares it with `C`
- User `A` moves `F` outside `G`
- `C` tries to download the file

Downside is that the share is now gone "forever" even if the owner re-gains access to the file. So I'm not sure if this is actually desired behavior

- [ ] Do we actually want this?

An alternative solution would be to mark the share as unavailable *somehow*, show the state to the user and make the user choose whether to remove the share or complain to the admin/owner of the file to give him his access back.

### DIFF
--- a/apps/files_sharing/lib/SharedStorage.php
+++ b/apps/files_sharing/lib/SharedStorage.php
@@ -146,9 +146,7 @@ class SharedStorage extends \OC\Files\Storage\Wrapper\Jail implements ISharedSto
 			/** @var Node|false $ownerNode */
 			$ownerNode = current($ownerNodes);
 			if (!$ownerNode) {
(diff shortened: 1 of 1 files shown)


icewind1991:
Also see https://github.com/nextcloud/server/pull/39170 for a command that admins can used for manual cleanup instead of automatic cleanup

come-nc:
Can you show screenshots for current situation on master?
It may be better to show the error, stating the owner of the share no longer have access to the file and let the user decide if they want to remove the share or not.

PR #39170, opened 2023-07-05
add command do delete orphan shares

icewind1991:
Compared to the existing background job, this command validates that the share owner still has access to the file instead of just checking that the file still exists.

I've decided against not adding this to the background command as the "owner lost access" case is recoverable by having the owner regain access. So automatically deleting them might be undesired.

Steps to create an orphaned share:

- Setup user `A`, `B` and `C`. Create a groupfolder `G` and give `A` and `B` access to it.
- User `B` uploads a file `F` to `G` and shares it with `C`
- User `A` moves `F` outside `G`
- `C` now has an orphaned share as `B` no longer has access to `F`


come-nc (approved):

artonge (commented):
- Does it really scale to fetch all share and then to set up the file system for each owner? Or does the `yield` statement makes it more manageable? Not familiar with it.
- Do we have a logic in place to prevent those from happening?

icewind1991:
> Does it really scale to fetch all share and then to set up the file system for each owner? Or does the yield statement makes it more manageable? Not familiar with it.

It will be slow yes, but there isn't really to much that can be done for that

AndyScherzinger:
If it only makes the command slow then I'd say that is fine or would it also impact "online" transactions?

AndyScherzinger (approved):

ChristophWurst:
The new occ command needs admin docs

Rello:
https://github.com/nextcloud/documentation/pull/11443

PR #39177, opened 2023-07-05
refactor federatedfilesharing app controllers

shdehnavi:
## Summary
The required adjustments have been made to the classes in `/lib/private/Remote ` namespace.

The improvements:

- Adding return types
- Adding throws types
- Adding types to properties
- Replacing qualifiers with imports
## Checklist

### DIFF
--- a/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
+++ b/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
@@ -290,7 +290,7 @@ class RequestHandlerController extends OCSController {
 		return new Http\DataResponse();
 	}
 
-	private function cleanupRemote($remote): string {
+	private function cleanupRemote(?string $remote = null): string {
 		$remote = substr($remote, strpos($remote, '://') + 3);
 
 		return rtrim($remote, '/');
@@ -374,7 +374,7 @@ class RequestHandlerController extends OCSController {
 	 * @param $ncPermissions
 	 * @return array
 	 */
-	protected function ncPermissions2ocmPermissions($ncPermissions): array {
+	protected function ncPermissions2ocmPermissions(?int $ncPermissions = null): array {
 		$ocmPermissions = [];
 
 		if ($ncPermissions & Constants::PERMISSION_SHARE) {
--- a/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
+++ b/apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
@@ -371,7 +371,7 @@ class RequestHandlerController extends OCSController {
 	/**
 	 * translate Nextcloud permissions to OCM Permissions
 	 *
-	 * @param $ncPermissions
+	 * @param int|null $ncPermissions
 	 * @return array
 	 */
 	protected function ncPermissions2ocmPermissions(?int $ncPermissions = null): array {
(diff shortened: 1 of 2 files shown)


apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -332,7 +290,7 @@
 		return new Http\DataResponse();
 	}
 
-	private function cleanupRemote($remote) {
+	private function cleanupRemote($remote): string {
```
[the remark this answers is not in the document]
shdehnavi: It's addressed in the next commit

apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -414,7 +374,7 @@
 	 * @param $ncPermissions
 	 * @return array
 	 */
-	protected function ncPermissions2ocmPermissions($ncPermissions) {
+	protected function ncPermissions2ocmPermissions($ncPermissions): array {
```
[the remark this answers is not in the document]
shdehnavi: It's addressed in the next commit

apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -332,7 +290,7 @@ public function unshare(int $id, ?string $token = null) {
 		return new Http\DataResponse();
 	}
 
-	private function cleanupRemote($remote) {
+	private function cleanupRemote(?string $remote = null): string {
```
nickvergessen, line 293: substr on null is invalid, so not sure why you added that type. It will explode in 8.2 anyway

shdehnavi: It's addressed in the next commit

apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -366,8 +324,10 @@ public function revoke(int $id, ?string $token = null) {
 	 *
 	 * @param bool $incoming
 	 * @return bool
+	 * @throws ContainerExceptionInterface
+	 * @throws NotFoundExceptionInterface
```
nickvergessen, line 328: This is not part of the API and only happens when the server code is broken.
So should not be documented and "exposed".

apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -411,10 +371,10 @@ public function updatePermissions(int $id, ?string $token = null, ?int $permissi
 	/**
 	 * translate Nextcloud permissions to OCM Permissions
 	 *
-	 * @param $ncPermissions
+	 * @param int|null $ncPermissions
 	 * @return array
 	 */
-	protected function ncPermissions2ocmPermissions($ncPermissions) {
+	protected function ncPermissions2ocmPermissions(?int $ncPermissions = null): array {
```
nickvergessen, line 377: null is not supported by the code.
```suggestion
	protected function ncPermissions2ocmPermissions(int $ncPermissions): array {
```

(Also adjust the doc block)

apps/federatedfilesharing/lib/Controller/RequestHandlerController.php
```diff
@@ -443,12 +403,14 @@ protected function ncPermissions2ocmPermissions($ncPermissions) {
 	 * @param string|null $token
 	 * @param string|null $remote
 	 * @param string|null $remote_id
-	 * @return Http\DataResponse
+	 * @return DataResponse
+	 * @throws ContainerExceptionInterface
+	 * @throws NotFoundExceptionInterface
```
nickvergessen, line 408: This is not part of the API and should therefore not be documented.

nickvergessen (changes requested):

PR #39183, opened 2023-07-06
Refactor core/Command/App

shdehnavi:
## Summary
The required adjustments have been made to the classes in `/core/Command/App ` namespace.

The improvements:

- Adding return types
- Adding throws types

## Checklist

### DIFF
--- a/core/Command/App/Install.php
+++ b/core/Command/App/Install.php
@@ -29,8 +29,6 @@ namespace OC\Core\Command\App;
 
 use OC\Installer;
 use OCP\App\IAppManager;
-use Psr\Container\ContainerExceptionInterface;
-use Psr\Container\NotFoundExceptionInterface;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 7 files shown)


ChristophWurst (approved):

core/Command/App/Install.php
```diff
@@ -66,6 +68,11 @@ protected function configure() {
 		;
 	}
 
+	/**
+	 * @throws ContainerExceptionInterface
+	 * @throws NotFoundExceptionInterface
+	 * @throws \Exception
```
nickvergessen, line 74: Since this is not a planned throw, but only happens when the server code is broken, I would not document them.

nickvergessen (changes requested):

core/Command/App/Install.php
shdehnavi: It's removed in the next commit

core/Command/App/Install.php
```diff
@@ -29,14 +29,16 @@
 
 use OC\Installer;
 use OCP\App\IAppManager;
+use Psr\Container\ContainerExceptionInterface;
+use Psr\Container\NotFoundExceptionInterface;
```
nickvergessen, line 33: unused import

fsamapoor (approved):

nickvergessen (approved):

PR #39184, opened 2023-07-06
Refactor core/Command/Background

shdehnavi:
## Summary
The required adjustments have been made to the classes in `/core/Command/Background ` namespace.

The improvements:

- Adding return types
- Adding throws types
- Updating doc blocks
- 
## Checklist

### DIFF
--- a/core/Command/Background/Job.php
+++ b/core/Command/Background/Job.php
@@ -58,9 +58,6 @@ class Job extends Command {
 		;
 	}
 
-	/**
-	 * @throws \ReflectionException
-	 */
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		$jobId = (int) $input->getArgument('job-id');
 
--- a/core/Command/Background/Job.php
+++ b/core/Command/Background/Job.php
@@ -105,9 +105,6 @@ class Job extends Command {
 		return 0;
 	}
 
-	/**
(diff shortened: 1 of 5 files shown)


core/Command/Background/Job.php
```diff
@@ -104,7 +107,10 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 		return 0;
 	}
 
-	protected function printJobInfo(int $jobId, IJob $job, OutputInterface$output): void {
+	/**
+	 * @throws \ReflectionException
```
ChristophWurst, line 111: I wouldn't document because this is not an exception to expect/catch when calling the method

shdehnavi: It's addressed in the next commit

skjnldsv (approved):

core/Command/Background/Job.php
```diff
@@ -60,6 +60,9 @@ protected function configure(): void {
 		;
 	}
 
+	/**
+	 * @throws \ReflectionException
```
ChristophWurst, line 64: same here. I would not document this.

shdehnavi: It's addressed in the next commit

ChristophWurst (approved):
Squash your commits into one please

:+1: thanks

PR #38546, opened 2023-05-31
fix(deps)!: Upgrade symfony/event-dispatcher to supported version 5.4.26

nickvergessen:
## Summary

- [x] https://github.com/nextcloud/server/pull/38545
- [x] https://github.com/nextcloud/server/pull/39086
- [x] https://github.com/nextcloud/server/pull/39190
- [x] https://github.com/nextcloud/server/pull/39304
- [x] https://github.com/nextcloud/server/pull/39305
- [x] https://github.com/nextcloud/server/pull/39485
- [x] https://github.com/nextcloud/server/pull/39487
- [x] https://github.com/nextcloud/server/pull/39571
- [x] https://github.com/nextcloud/server/pull/39595
- [x] https://github.com/nextcloud/server/pull/39605

## TODO

- [x] https://github.com/nextcloud/3rdparty/pull/1423

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #35388, opened 2022-11-24
Make CalDAV resource sync a service and expose as occ command

ChristophWurst:
## Summary

CalDAV resources and rooms are provided by apps. Currently the rooms and resources are read in a background job once an hour. Sometimes one wants to see the changes right away.

By moving the job logic into a reusable service we can also invoke it from an occ command.

This also drops the two unused events that aren't emitted anywhere
* ``\OCP\Calendar\Resource\ForceRefreshEvent``
* ``\OCP\Calendar\Room\ForceRefreshEvent``

## TODO

- [x] Move logic into a service
- [x] Add occ command
- [ ] Documentation

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [ ] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #39190, opened 2023-07-06
fix(dav): Move DAV app to non deprecated event dispatcher

nickvergessen:
One step closer for #38546 

## Checklist

### DIFF
--- a/apps/dav/lib/AppInfo/Application.php
+++ b/apps/dav/lib/AppInfo/Application.php
@@ -236,14 +236,14 @@ class Application extends App implements IBootstrap {
 		});
 
 
-		$dispatcher->addListener('\OCA\DAV\CalDAV\CalDavBackend::updateShares', function (GenericEvent $event) use ($container) {
+		$dispatcher->addListener(CalendarShareUpdatedEvent::class, function (CalendarShareUpdatedEvent $event) use ($container) {
 			/** @var Backend $backend */
 			$backend = $container->query(Backend::class);
 			$backend->onCalendarUpdateShares(
-				$event->getArgument('calendarData'),
-				$event->getArgument('shares'),
-				$event->getArgument('add'),
-				$event->getArgument('remove')
+				$event->getCalendarData(),
+				$event->getOldShares(),
+				$event->getAdded(),
+				$event->getRemoved()
 			);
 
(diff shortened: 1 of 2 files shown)


apps/dav/lib/AppInfo/Application.php
```diff
@@ -227,40 +224,26 @@ public function registerHooks(HookManager $hm,
...
-				$job = $container->query(UpdateCalendarResourcesRoomsBackgroundJob::class);
-				$job->run([]);
-				$serverContainer->getJobList()->setLastRun($job);
-			} catch (Exception $ex) {
-				$serverContainer->get(LoggerInterface::class)->error($ex->getMessage(), ['exception' => $ex]);
-			}
-		};
-
-		$dispatcher->addListener('\OCP\Calendar\Resource\ForceRefreshEvent', $eventHandler);
-		$dispatcher->addListener('\OCP\Calendar\Room\ForceRefreshEvent', $eventHandler);
```
nickvergessen, line 263: I could not find `ForceRefreshEvent` as a search term in our github org, apart from this occurance here. I assume there were plans that changed on the way, so I think this is dead code and I just remove it.

apps/dav/lib/AppInfo/Application.php
```diff
@@ -227,40 +224,26 @@ public function registerHooks(HookManager $hm,
 			}
 		});
 
-		$dispatcher->addListener('OC\AccountManager::userUpdated', function (GenericEvent $event) use ($container) {
+		$dispatcher->addListener('OC\AccountManager::userUpdated', function ($event) use ($container) {
```
ChristophWurst, line 227: what type is the event now? can it be hinted?

nickvergessen: Still GenericEvent, but then Psalm complains because the event listener does not consume IEvent. Removing silences psalm and makes it work. The actual fix again is creating and only listening to typed events, but time....

apps/dav/lib/AppInfo/Application.php
tcitworld: Also in https://github.com/nextcloud/server/pull/35388 :)

tcitworld (approved):

skjnldsv (approved):

PR #39196, opened 2023-07-06
feat(sharing): add sharing overview view

skjnldsv:
image

- [x] Overview
- [x] Shared with you
- [x] Shared by you
- [x] Link shares
- [x] Pending shares
- [x] Deleted shares
- [x] Tests
  - [x] View
  - [x] Actions
- Fixes
  - [x] Add mtime and size to backend
  - [x] Fix remote share backend path
  - [x] Fix own share permissions

### Followup
- Implement inline avatar/link icon action


apps/files_sharing/src/services/SharingService.ts
```diff
@@ -0,0 +1,181 @@
...
+		},
+		data: []
+	}
+}
+
+const headers = {
+	'Content-Type': 'application/json',
+}
+
+const ocsEntryToNode = function(ocsEntry: any): Folder | File | null {
```
artonge, line 49: Any reason to not write it like so:

```suggestion
function ocsEntryToNode(ocsEntry: any): Folder | File | null {
```

?

apps/files_sharing/src/actions/restoreShareAction.ts
```diff
@@ -0,0 +1,65 @@
...
+	},
+	async execBatch(nodes: Node[], view: Navigation, dir: string) {
+		return Promise.all(nodes.map(node => this.exec(node, view, dir)))
+	},
+
+	order: 1,
+	inline: () => true,
+})
+
+registerFileAction(action)
```
artonge, line 65: Just saw this pattern. Not a fan of side effects when importing a file.
Shouldn't we move the registration call outside of the declaration file?

apps/files_sharing/lib/Listener/LoadAdditionalListener.php
```diff
@@ -37,10 +38,13 @@ public function handle(Event $event): void {
...
 		}
 
-		// After files for the files list shared content
-		Util::addScript(Application::APP_ID, 'files_sharing', 'files');
 		// After files for the breadcrumb share indicator
 		Util::addScript(Application::APP_ID, 'additionalScripts', 'files');
 		Util::addStyle(Application::APP_ID, 'icons');
+
+		$shareManager = \OC::$server->get(IManager::class);
+		if ($shareManager->shareApiEnabled() && class_exists('\OCA\Files\App')) {
```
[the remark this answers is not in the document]
skjnldsv: No clue, this is from the previous code :shrug:

apps/files_sharing/src/services/SharingService.ts
skjnldsv: Personal preference and a few valid points:
- `const` prevents reassignment of the name while `function` does not
- same syntax as if it was an arrow function, while you cannot do that with `function`: `function () => {}`

In the end if looks as clean as any other variable, which in the end, it is.
And I like consistency ;)

apps/files_sharing/src/actions/restoreShareAction.ts
skjnldsv: I personally don't mind. We can simplify this in a followup.
But I'm not fond of duplicate code, so it means the main.ts file will look like this:

```ts
import { registerFileAction } from '../services/FileAction'

import { deleteAction } from './actions/deleteAction'
import { downloadAction } from './actions/downloadAction'
import { editLocallyAction } from './actions/editLocallyAction'
import { favoriteAction } from './actions/favoriteAction'
import { openFolderAction } from './actions/openFolderAction'
import { renameAction } from './actions/renameAction'
import { sidebarAction } from './actions/sidebarAction'
import { viewInFolderAction } from './actions/viewInFolderAction'

registerAction(deleteAction)
registerAction(downloadAction)
registerAction(editLocallyAction)
registerAction(favoriteAction)
registerAction(openFolderAction)
registerAction(renameAction)
registerAction(sidebarAction)

```

And it doesn't look as clean and simple as side-effect imports for actions :)

apps/files_sharing/src/services/SharingService.ts
artonge: :+1: Indeed, personal preference, I was curious for the rational :)

skjnldsv:
I forgot to add the open-in-files action for sharing
I have a few fixes to push tomorrow, feel free to review already what is here :shrug: :pray:

apps/files_sharing/src/actions/restoreShareAction.ts
skjnldsv: You're right, let's go for explicit before clean, it makes more sense! :+1:

skjnldsv: https://github.com/nextcloud/server/issues/39298

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -182,6 +182,11 @@ protected function formatShare(IShare $share, Node $recipientNode = null): array
 		$sharedBy = $this->userManager->get($share->getSharedBy());
 		$shareOwner = $this->userManager->get($share->getShareOwner());
 
+		$isOwnShare = false;
+		if ($shareOwner !== null) {
+			$isOwnShare = $shareOwner->getUID() === $this->currentUser;
+		}
+
```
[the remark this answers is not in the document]
skjnldsv: I used it in a previous commit and that got changed, it should not be an issue

artonge (approved):

susnux (approved):

PR #39201, opened 2023-07-06
Better untested app button's wording

Jerome-Herbinet:
Replace "Enable untested app" with "Authorize this untested app".

"Enable" creates confusion because when you click on this button, you only allow installing it ... and then it makes the install button appear ... and finally, the app is automatically enabled after install.

## Checklist


apps/settings/src/mixins/AppManagement.js
```diff
@@ -42,9 +42,9 @@ export default {
 		},
 		forceEnableButtonText() {
 			if (this.app.needsDownload) {
-				return t('settings', 'Enable untested app')
+				return t('settings', 'Authorize this untested app')
```
szaimen, line 45: What about `Activate this untested app` ? cc @nextcloud/designers

nimishavijay:
Definitely an improvement! I'm thinking it may be better to just get rid of the 2 steps, so authorize and automatically start the download action when the "Enable untrusted app" button is clicked. What do you think?

apps/settings/src/mixins/AppManagement.js
Jerome-Herbinet: For me, "Activate" is a synonym of "Enable". What about "Allow". Otherwise, why don't you like "Authorize". I think we should keep away from the "enable" (or activate) notion. What do you think ?

Jerome-Herbinet:
> Definitely an improvement! I'm thinking it may be better to just get rid of the 2 steps, so authorize and automatically start the download action when the "Enable untrusted app" button is clicked. What do you think?

"untrusted" is IMO too discriminant … and may let the user think that the whole app (whatever the version) is untrusted. Do you have another suggestion ?

jancborchardt (changes requested):
In an ideal world I would say @nimishavijay’s suggestion of a single-step process is great, but I assume this would take a lot more work.

So I would say @Jerome-Herbinet the best call is to adjust the wording to "Allow untested app". Short and still different from pressing "Enable …" 2 times.

Jerome-Herbinet:
New PR #39575

PR #39208, opened 2023-07-07
Use PHP8's match expression instead of switch

shdehnavi:
## Summary
The required adjustments have been made to the classes in the `/lib/public` namespace.

The improvements:

- Using PHP8's `match` expression instead of `switch`

## Checklist

### DIFF
--- a/lib/public/AppFramework/Db/QBMapper.php
+++ b/lib/public/AppFramework/Db/QBMapper.php
@@ -247,7 +247,6 @@ abstract class QBMapper {
 			'json' => IQueryBuilder::PARAM_JSON,
 			default => IQueryBuilder::PARAM_STR,
 		};
-
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


lib/public/AppFramework/Db/QBMapper.php
```diff
@@ -238,24 +238,15 @@ protected function getParameterTypeForProperty(Entity $entity, string $property)
...
-				return IQueryBuilder::PARAM_LOB;
-			case 'datetime':
-				return IQueryBuilder::PARAM_DATE;
-			case 'json':
-				return IQueryBuilder::PARAM_JSON;
-		}
-
-		return IQueryBuilder::PARAM_STR;
+		return match ($types[$property]) {
+			'int', 'integer' => IQueryBuilder::PARAM_INT,
```
nickvergessen, line 242: As mentioned in the talk PR I think this decreases readability and since it has no advantage we shouldn't do it, so that patches in the area can be backported with less chance for conflicts.

nickvergessen:
Closing as per comment

PR #39209, opened 2023-07-07
Use PHP8's match expression instead of switch

shdehnavi:
## Summary
Using PHP8's `match` expression instead of `switch` in the `/tests/lib` namespace classes.

## Checklist


come-nc:
I’m not sure what good that brings apart from complicating backport on tests.
For core code I can see the impulse for wanting to have the cleaner code as possible, but for tests it is less true, and here it is not clear what is the benefit.
Match is tiny bit shorter but it’s also less known by developers.

PR #39216, opened 2023-07-07
Refactor "substr" calls in lib/private to improve code readability

shdehnavi:
## Summary
Replacing `substr` with `str_starts_with` and `str_ends_with` in the `/lib/private` namespace to improve code readability

## Checklist

### DIFF
--- a/lib/private/DateTimeFormatter.php
+++ b/lib/private/DateTimeFormatter.php
@@ -125,7 +125,7 @@ class DateTimeFormatter implements \OCP\IDateTimeFormatter {
 	 * @return string Formatted relative date string
 	 */
 	public function formatDateRelativeDay($timestamp, $format = 'long', \DateTimeZone $timeZone = null, \OCP\IL10N $l = null) {
-		if (!str_ends_with($formatDate, '^') && !str_ends_with($formatDate, '*')) {
+		if (!str_ends_with($format, '^') && !str_ends_with($format, '*')) {
 			$format .= '^';
 		}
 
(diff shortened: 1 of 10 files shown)


lib/private/DateTimeFormatter.php
```diff
@@ -125,7 +125,7 @@ public function formatDate($timestamp, $format = 'long', \DateTimeZone $timeZone
 	 * @return string Formatted relative date string
 	 */
 	public function formatDateRelativeDay($timestamp, $format = 'long', \DateTimeZone $timeZone = null, \OCP\IL10N $l = null) {
-		if (substr($format, -1) !== '*' && substr($format, -1) !== '*') {
+		if (!str_ends_with($format, '*') && !str_ends_with($format, '*')) {
```
fsamapoor, line 128: Aren't these the same?

fsamapoor (approved):
Much much better. Thanks!
Other than the comment I left, looks good to me.

lib/private/DateTimeFormatter.php
shdehnavi: Thank you, it's addressed in the next commit

lib/private/DateTimeFormatter.php
```diff
@@ -125,7 +125,7 @@ public function formatDate($timestamp, $format = 'long', \DateTimeZone $timeZone
 	 * @return string Formatted relative date string
 	 */
 	public function formatDateRelativeDay($timestamp, $format = 'long', \DateTimeZone $timeZone = null, \OCP\IL10N $l = null) {
-		if (substr($format, -1) !== '*' && substr($format, -1) !== '*') {
+		if (!str_ends_with($format, '*')) {
```
artonge, line 128:
```suggestion
		if (!str_ends_with($formatDate, '^') && !str_ends_with($formatDate, '*')) {
```

I think the duplication was due to a bad copy-pasting. Having `*` or `^` at the end of the format seem on purpose and make sense if you look at the documentation of the library that we use. Same as on line 292.

https://punic.readthedocs.io/en/3.4.0/_static/api/class-Punic.Calendar.html
Screenshot from 2023-08-17 19-38-51

artonge (approved):

come-nc (changes requested):

come-nc (approved):

icewind1991 (approved):

PR #39221, opened 2023-07-07
Bugfix/bulk upload empty files

mgallien:
## Summary

Solve long running issue with bulk upload of empty files.
Require also fixes from desktop client see https://github.com/nextcloud/desktop/pull/5871

## TODO

- [ ] ...

## Checklist

should probably be backported to all maintained releases with bulk upload

### DIFF
--- a/apps/dav/lib/BulkUpload/MultipartRequestParser.php
+++ b/apps/dav/lib/BulkUpload/MultipartRequestParser.php
@@ -211,13 +211,17 @@ class MultipartRequestParser {
 			throw new BadRequest("Computed md5 hash is incorrect.");
 		}
 
-		$content = stream_get_line($this->stream, $length);
+		if ($length === 0) {
+			$content = '';
+		} else {
+			$content = stream_get_line($this->stream, $length);
+		}
 
 		if ($content === false) {
 			throw new Exception("Fail to read part's content.");
 		}
 
-		if (feof($this->stream)) {
+		if ($length !== 0 && feof($this->stream)) {
 			throw new Exception("Unexpected EOF while reading stream.");
 		}
 


kesselb:
Thanks for your patch :+1: 

The change for the desktop client is important. 

I could reproduce the issue with my development setup (nginx proxy -> php container), 
but the requests for `/remote.php/dav/bulk` did not trigger my breakpoint.

```
proxy_1      | nginx.1     | 2023/07/08 20:33:56 [info] 46#46: *34 client prematurely closed connection, client: 172.19.0.1, server: server.internal, request: "POST /remote.php/dav/bulk HTTP/1.1", host: "server.internal"
proxy_1      | nginx.1     | server.internal 172.19.0.1 - admin [08/Jul/2023:20:33:56 +0000] "POST /remote.php/dav/bulk HTTP/1.1" 400 0 "-" "Mozilla/5.0 (Linux) mirall/3.9.50 (build 16487) (Nextcloud, ubuntu-6.2.0-24-generic ClientArchitecture: x86_64 OsArchitecture: x86_64)" "-"
```

It seems to me that nginx did not forward the post request without a body to the php container. 
Using the build from your pr made it work. 
 
I did some testing with the patched desktop client and the server pr. 

1. Open a terminal
2. Change directory to your Nextcloud sync folder
3. Run `echo "123" > test61.txt; touch test62.txt; echo "456" > test63.txt`

Result:

```
cat test61.txt 
123
```

```
cat test62.txt 
--boundary_.oOo._EfhWH90em6640g8imIdDSkdgh4fStM3o
Content-Length: 4
Content-Type: application/octet-stream
X-File-MD5: d2d362cdc6579390f1c0617d74a7913d
X-File-Mtime: 1688851265
X-File-Path: /test63.txt

456

--boundary_.oOo._EfhWH90em6640g8imIdDSkdgh4fStM3o--
```

Notification: test63.txt could not be synced due to an error.
And a put request for test63.txt writing the proper content. 

I guess stream_get_line with $length = 0 is the problem here, 
because the default socket chunk size is used then. 

```diff
Index: apps/dav/lib/BulkUpload/MultipartRequestParser.php
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/apps/dav/lib/BulkUpload/MultipartRequestParser.php b/apps/dav/lib/BulkUpload/MultipartRequestParser.php
--- a/apps/dav/lib/BulkUpload/MultipartRequestParser.php	(revision 9bf69115593a9abaedd8cc8d886725253100f3c0)
+++ b/apps/dav/lib/BulkUpload/MultipartRequestParser.php	(date 1688852480282)
@@ -211,7 +211,11 @@
 			throw new BadRequest("Computed md5 hash is incorrect.");
 		}
 
-		$content = stream_get_line($this->stream, $length);
+		if ($length === 0) {
+			$content = '';
+		} else {
+			$content = stream_get_line($this->stream, $length);
+		}
 
 		if ($content === false) {
 			throw new Exception("Fail to read part's content.");
```

Made it work for me. 

Please keep my setup with the reverse proxy in mind. We should probably test the patch without a reverse proxy. 
And I don't know if the logic is backwards compatible.

apps/dav/lib/BulkUpload/MultipartRequestParser.php
```diff
@@ -217,10 +217,6 @@ private function readPartContent(int $length, string $md5): string {
 			throw new Exception("Fail to read part's content.");
 		}
 
-		if (feof($this->stream)) {
-			throw new Exception("Unexpected EOF while reading stream.");
-		}
-
```
artonge, line 223: Was this causing issues?

mgallien: yes
if the file has size 0 then upon receiving the part, `feof($this->stream)` will be true immediately after finding the boundaries of the part and this will throw instead of saving the empty file

artonge: Can we change to:


```suggestion
		if ($length !== 0 && feof($this->stream)) {
			throw new Exception("Unexpected EOF while reading stream.");
		}

```

mgallien: done

artonge (approved):

apps/dav/lib/BulkUpload/MultipartRequestParser.php
```diff
@@ -208,16 +208,20 @@ private function readPartContent(int $length, string $md5): string {
 		$computedMd5 = $this->computeMd5Hash($length);
 
 		if ($md5 !== $computedMd5) {
-			throw new BadRequest("Computed md5 hash is incorrect.");
+			throw new BadRequest("Computed md5 hash is incorrect. Expected: " . $md5 . " computed " . $computedMd5);
```
come-nc, line 211: Is that never a security issue?
I do not know the code paths for this but showing expected md5 to the caller looks suspicious.

mgallien: I agree
I could have kept this as a private change needed because I do not have php debugger setup

come-nc (approved):

kesselb:
Please don't forget to review the pr for 24 and check if we also need it: https://github.com/nextcloud/server/pull/39508

PR #39222, opened 2023-07-07
Migrate more apps to IEventDispatcher

nickvergessen:
One step closer for #38546 

- [x] files_trashbin: Emit existing event additionally as dispatchTyped
    - [x] documentation: https://github.com/nextcloud/documentation/pull/10737
- [x] workflowengine: Remove legacy event (deprecated since 17 (so allowed to drop since 26))
    - [x] documentation: https://github.com/nextcloud/documentation/pull/10737
- [X] lookup…: Move listener to IEventDispatcher

## Checklist

### DIFF
--- a/apps/lookup_server_connector/lib/AppInfo/Application.php
+++ b/apps/lookup_server_connector/lib/AppInfo/Application.php
@@ -34,9 +34,9 @@ use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
+use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IUser;
 use Psr\Container\ContainerInterface;
-use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\GenericEvent;
 
 class Application extends App implements IBootstrap {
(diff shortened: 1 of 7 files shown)


apps/workflowengine/lib/Manager.php
```diff
@@ -694,7 +687,6 @@ public function formatOperation(array $operation): array {
 	 */
 	public function getEntitiesList(): array {
 		$this->dispatcher->dispatchTyped(new RegisterEntitiesEvent($this));
-		$this->legacyEventDispatcher->dispatch(IManager::EVENT_NAME_REG_ENTITY, new GenericEvent($this));
```
come-nc, line 697: Code in applications which are registering a listener for this, will they get an error or just silently won’t be called anymore?

apps/lookup_server_connector/lib/AppInfo/Application.php
```diff
@@ -56,15 +56,17 @@ public function boot(IBootContext $context): void {
 	/**
 	 * @todo move the OCP events and then move the registration to `register`
 	 */
-	private function registerEventListeners(EventDispatcherInterface $dispatcher,
+	private function registerEventListeners(IEventDispatcher $dispatcher,
 											ContainerInterface $appContainer): void {
-		$dispatcher->addListener('OC\AccountManager::userUpdated', function (GenericEvent $event) use ($appContainer) {
-			/** @var IUser $user */
-			$user = $event->getSubject();
+		$dispatcher->addListener('OC\AccountManager::userUpdated', function ($event) use ($appContainer) {
```
come-nc, line 61: There is no typed event version for this one?
This change is not really updating to the new system, it’s still using the old one, no?

nickvergessen: > There is no typed event version for this one?

No unluckily not.

> This change is not really updating to the new system, it’s still using the old one, no?

For "listeners" there is no new/old system. It's the same. But yeah I'm not planing to migrate the things to typed events with the current work. I just want to free the way to be able to merge #38546 and that comes with a breaking API change, unless our wrapper is used when dispatching.
But for the ease of search and being sure, I'm move all code away from directly using the thirdparty.

apps/workflowengine/lib/Manager.php
nickvergessen: "Silently" won't be called anymore after 3+ years of deprecation. It is documented in the changelogs never the less:
https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_28.html#removed-events

come-nc (approved):

nickvergessen:
Rebased after conflicts

kesselb (approved):

PR #39244, opened 2023-07-08
Refactor dashboard app

shdehnavi:
## Summary
The required adjustments have been made to the `dashboard` app.

The improvements:

- Adding types to properties

## Checklist


juliusknorr (approved):

apps/dashboard/lib/Controller/DashboardController.php
```diff
@@ -44,32 +44,17 @@
...
-		IInitialState $initialState,
-		IEventDispatcher $eventDispatcher,
-		IManager $dashboardManager,
-		IConfig $config,
-		$userId
+		private IInitialState $initialState,
+		private IEventDispatcher $eventDispatcher,
+		private IManager $dashboardManager,
+		private IConfig $config,
+		private string $userId,
```
nickvergessen, line 54: While not documented, userId could be null when a guest invokes a URL and now the controller breaks and logs an error instead of returning a proper 401 response.
We can either make this nullable or we leave it vague with the type hint comment only, making psalm happy (not checking for null inside the controller methods where it can not be null), but at the same time allowing null on a code level

shdehnavi: It's addressed in the next commit

nickvergessen: Also in the other controller

shdehnavi: It's addressed in the next commit

nickvergessen (approved):

julien-nc (approved):

PR #39263, opened 2023-07-10
Add IgnoreOpenAPI attribute

provokateurin:
## Summary

Controller methods with this attribute (or annotation) will be ignored by https://github.com/nextcloud/openapi-extractor.

From https://github.com/nextcloud/server/pull/39223#discussion_r1257890403

## Checklist

### DIFF
--- a/lib/public/AppFramework/Http/Attribute/IgnoreOpenAPI.php
+++ b/lib/public/AppFramework/Http/Attribute/IgnoreOpenAPI.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Kate Döen <kate.doeen@nextcloud.com>
+ *
+ * @author Kate Döen <kate.doeen@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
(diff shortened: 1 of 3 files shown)


lib/public/AppFramework/Http/Attribute/IgnoreAPI.php
```diff
@@ -0,0 +1,37 @@
...
+namespace OCP\AppFramework\Http\Attribute;
+
+use Attribute;
+
+/**
+ * Attribute for controller methods that should be ignored when generating API documentation
+ *
+ * @since 28.0.0
+ */
+#[Attribute]
```
nickvergessen, line 35:
```suggestion
#[Attribute(Attribute::TARGET_METHOD)]
```

Or does it apply to full classes as well?

lib/public/AppFramework/Http/Attribute/IgnoreAPI.php
```diff
@@ -0,0 +1,37 @@
...
+class IgnoreAPI {
```
nickvergessen, line 36: Also should it contain the "documentation" keyword?

lib/public/AppFramework/Http/Attribute/IgnoreAPI.php
provokateurin: Right now no, but now that you mention it it would make a lot of sense. Will implement that, so the attribute will also work for controllers :+1:

lib/public/AppFramework/Http/Attribute/IgnoreAPI.php
provokateurin: Yeah that would make sense, but to be more specific `IgnoreOpenAPI` would be better imo

lib/public/AppFramework/Http/Attribute/IgnoreAPI.php
nickvergessen: Then:
```suggestion
#[Attribute(Attribute::TARGET_METHOD | Attribute::TARGET_CLASS)]
```

lib/public/AppFramework/Http/Attribute/IgnoreOpenAPI.php
nickvergessen, line 1: Missing `bash build/autoload...` execution

nickvergessen (approved):

bigcat88 (approved):

PR #39264, opened 2023-07-10
Add instance category while checking new updates

Altahrim:
## Summary
Send instance category when checking updates

### DIFF
--- a/apps/updatenotification/lib/Notification/BackgroundJob.php
+++ b/apps/updatenotification/lib/Notification/BackgroundJob.php
@@ -31,7 +31,6 @@ use OC\Updater\VersionCheck;
 use OCP\App\IAppManager;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\BackgroundJob\TimedJob;
-use OCP\Http\Client\IClientService;
 use OCP\IConfig;
 use OCP\IGroup;
 use OCP\IGroupManager;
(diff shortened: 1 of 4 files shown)


lib/private/Updater/VersionCheck.php
```diff
@@ -130,4 +126,25 @@ protected function getUrlContent($url) {
...
+		];
+
+		$nbUsers = $this->userManager->countSeenUsers();
+		foreach ($categoryBoundaries as $categoryId => $boundary) {
+			if ($nbUsers <= $boundary) {
+				return $categoryId;
+			}
+		}
+
+		return $categoryId + 1;
```
come-nc, line 148: I did not even know the foreach var was available to outside scope, it would be way clearer to me if it was not using it.
Maybe:
```suggestion
		return count($categoryBoundaries);
```

lib/private/Updater/VersionCheck.php
```diff
@@ -130,4 +126,25 @@ protected function getUrlContent($url) {
 		$response = $client->get($url);
 		return $response->getBody();
 	}
+
+	private function computeCategory() {
```
come-nc, line 130:
```suggestion
	private function computeCategory(): int {
```

lib/private/Updater/VersionCheck.php
Altahrim: I'll change it :)

apps/updatenotification/lib/Notification/BackgroundJob.php
```diff
@@ -264,7 +245,9 @@ protected function deleteOutdatedNotifications($app, $version) {
 	protected function createVersionCheck(): VersionCheck {
 		return new VersionCheck(
```
kesselb, line 246: Could we inject VersionCheck via DI as class dependency.
I don't see anything expensive in VersionCheck's constructor.

Altahrim: Indeed, I think there is a refactor that can be done here. I'll try something :)

come-nc (approved):

kesselb (approved):

PR #39268, opened 2023-07-10
fix(flow): Revert invalid rename of check class for RequestTime

juliusknorr:
Revert the wrongly adapted class name of the RequestTime check class which broke the flow UI for admins after #38878


apps/workflowengine/src/components/Checks/request.js
```diff
@@ -38,7 +38,7 @@ const RequestChecks = [
 		component: RequestURL,
 	},
 	{
-		class: 'OCA\\WorkflowEngine\\Check\\RequestTimeController',
+		class: 'OCA\\WorkflowEngine\\Check\\RequestTime',
```
provokateurin, line 41: It only needs to be fixed in the frontend?

juliushaertl: Yes, as this refers to the class from https://github.com/nextcloud/server/blob/9ba9f9a7dcd7654336fa714cbd178f6a2c1710a8/apps/workflowengine/lib/Check/RequestTime.php which is still named like that not the renamed controller

provokateurin: I see. I think my IDE just refactored the name and I looked at it and it seemed reasonable.

provokateurin (approved):

nickvergessen (approved):

PR #39275, opened 2023-07-10
Add optional to add backtrace to log items

icewind1991:
Disabled by default (for now?) due to the increased log size.

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -221,6 +221,12 @@ class Log implements ILogger, IDataLogger {
 			$this->eventDispatcher->dispatchTyped(new BeforeMessageLoggedEvent($app, $level, $entry));
 		}
 
+		$hasBacktrace = isset($entry['exception']);
+		$logBacktrace = $this->config->getValue('log.backtrace', false);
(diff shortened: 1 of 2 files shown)


blizzz (approved):

lib/private/Log.php
```diff
@@ -221,6 +221,12 @@ public function log(int $level, string $message, array $context = []) {
 			$this->eventDispatcher->dispatchTyped(new BeforeMessageLoggedEvent($app, $level, $entry));
 		}
 
+		$hasBacktrace = isset($entry['exception']);
+		$logBacktrace = $this->config->getValue('log.backtrace', false);
```
Fenn-CS, line 225: How does anyone know that this config option exists?

- Is it documented?
- Is it in the config template?

Otherwise, someone has to know this snippet of code exists to attempt using that which means it could potentially be dead-code without the above?

nfebe (changes requested):

blizzz:
Added an entry to config.sample.php (which is auto-synced to docs.nc.c).

Pytal (approved):

AndyScherzinger (approved):

PR #39276, opened 2023-07-10
full uppercase const

ArtificialOwl:
fix last character of `StorageConfig::MOUNT_TYPE_PERSONAl`

### DIFF
--- a/apps/files_external/lib/Lib/StorageConfig.php
+++ b/apps/files_external/lib/Lib/StorageConfig.php
@@ -40,7 +40,7 @@ use OCA\Files_External\ResponseDefinitions;
  */
 class StorageConfig implements \JsonSerializable {
 	public const MOUNT_TYPE_ADMIN = 1;
-	public const MOUNT_TYPE_PERSONAl = 2;
+	public const MOUNT_TYPE_PERSONAL = 2;
 
 	/**
 	 * Storage config id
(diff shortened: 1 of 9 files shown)


nickvergessen (approved):

apps/files_external/lib/Lib/StorageConfig.php
```diff
@@ -37,7 +37,7 @@
  */
 class StorageConfig implements \JsonSerializable {
 	public const MOUNT_TYPE_ADMIN = 1;
-	public const MOUNT_TYPE_PERSONAl = 2;
```
icewind1991, line 40: Have we checked if this is references in any other app?

Might be worth to leave the old const in place for compatibility

nickvergessen: Hmm I thought PHP was case forgiving and it just errors in Psalm. But seems that is only for Class names:
https://3v4l.org/bboIB

So yeah, should keep the old one in parallel.

apps/files_external/lib/Service/DBConfigService.php
```diff
@@ -37,7 +37,7 @@
  */
 class DBConfigService {
 	public const MOUNT_TYPE_ADMIN = 1;
-	public const MOUNT_TYPE_PERSONAl = 2;
```
nickvergessen, line 40: same here

ArtificialOwl: @icewind1991 should we merge or we keep both (`DBConfigService` and `StorageConfig`) definition ?

apps/files_external/lib/Lib/StorageConfig.php
ArtificialOwl: now marked as ~~deprecated~~

icewind1991 (approved):

joshtrichards (approved):

PR #30359, opened 2021-12-21
Allow calling cron jobs background job class with occ

juliusknorr:
This would allow important jobs to be handled in a separate cronjob execution, so they can get executed, even if other jobs have lined up before

e.g. have a separate cron job with `php occ background-job:worker OCA\\Files\\BackgroundJob\\ScanFiles`

PR #39290, opened 2023-07-10
feat: Move mimetype repair to background jobs

juliusknorr:
## Summary

Move the heavy and slow mime type repair steps into background jobs that are split by storage id.

Some considerations:
- The query is quite hard to optimize as we need to check for the file extension on all files
- Running an upgrade without this is usually fine as just old files might not immediately show up with the newly introduced mimetypes then
- Splitting by storage id seems to be the most reasonable thing to not have massive update transactions, but might not be enough yet

- [ ] Check if this mass of background jobs can become an issue on large instances, maybe those should be only during off-peak hours but this might have the risk of taking forever then
- [ ] Still have a way to manually run them, but we'd need to safeguard to not run the jobs then
  - [ ] Might be time to finish https://github.com/nextcloud/server/pull/30359
  - [ ] https://github.com/nextcloud/server/pull/37835 might also be helpful for this

## Checklist


joshtrichards:
Perhaps incorporate a check (for Admin->Overview) if it's still "in progress" that reports at info-only level. Maybe with some indication of how far along it is so the user can decide whether they need to do something to push it along faster.

juliushaertl:
Replaced by https://github.com/nextcloud/server/pull/45930

Issue #38503, opened 2023-05-29
[Bug]: the display does not change when clicking on a favorite in Files


maximelehericy:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

see title

### Steps to reproduce

click on a favorite.
display should show the content of your favorite folder.
nothing happens, but the URL changes.

### Expected behavior

 display should show the content of your favorite folder.

### Installation method

None

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### List of activated Apps

```shell
c.nc.c
```

szaimen:
cc @nextcloud/server-frontend

skjnldsv:
Clicking where, in the files list? In the navigation?

szaimen:
Apparently @Rello can reproduce this as well. 

Possibly caused by https://github.com/nextcloud/server/pull/38441?

arnowelzel:
JFTR: Happens here as well with a server I just upgraded from NC 26.0.3 to NC 27.0.0. It does not matter which browser I use and I have to explain nearly 40 users that the favorites are not usable any longer until this I fixed. Wasn't there a E2E test to make sure that works before releasing a new version?

arnowelzel:
> > JFTR: I have to explain nearly 40 users that the favorites are not usable any longer until this I fixed.
> 
> Favorites are still usable. Workaround:
> 
> * Click ON Favorites, (not on the expand [>]/[v] button):
> * the Favorites are listed in main window, where you can click and open them.

Thanks for the hint - I also found that out meanwhile ;-)

[...]

PR #39301, opened 2023-07-11
[stable27] fix(files): fix legacy router query change

skjnldsv:
Fix https://github.com/nextcloud/server/issues/38503


szaimen (approved):
Tested and seems to fix the problem. Thanks for fixing this @skjnldsv ! :)

artonge (approved):
Those? conditional? operators? are? really? useful !

PR #39303, opened 2023-07-11
feat: Use file cache query builder for searching for favorites

juliusknorr:
When trying to REPORT files that are favorites we still used the old code path to aggregate them, this moves this over to the filecache search query builder methods.

Request to trigger:

```
blackfire curl -X REPORT \
  --url http://admin:admin@nextcloud.local/remote.php/dav/files/admin/MyGroupfolder4 \
  --header 'Content-Type: application/xml' \
  --data '<?xml version="1.0"?>
<oc:filter-files  xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns">
         <oc:filter-rules>
                 <oc:favorite>1</oc:favorite>
         </oc:filter-rules>
        <d:prop>
            <d:resourcetype />
        </d:prop>
 </oc:filter-files> '
```

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
@@ -33,6 +33,7 @@ use OCP\Files\Folder;
 use OCP\Files\Node as INode;
 use OCP\IGroupManager;
 use OCP\ITagManager;
+use OCP\ITags;
 use OCP\IUserSession;
 use OCP\SystemTag\ISystemTagManager;
 use OCP\SystemTag\ISystemTagObjectMapper;
@@ -49,6 +50,10 @@ class FilesReportPlugin extends ServerPlugin {
 	// namespace
 	public const NS_OWNCLOUD = 'http://owncloud.org/ns';
 	public const NS_NEXTCLOUD = 'http://nextcloud.org/ns';
+
+	public const NS_OWNCLOUD_PREFIX = '{' . self::NS_OWNCLOUD. '}';
+	public const NS_NEXTCLOUD_PREFIX = '{' . self::NS_NEXTCLOUD. '}';
+
 	public const REPORT_NAME = '{http://owncloud.org/ns}filter-files';
 	public const SYSTEMTAG_PROPERTYNAME = '{http://owncloud.org/ns}systemtag';
 	public const CIRCLE_PROPERTYNAME = '{http://owncloud.org/ns}circle';
@@ -186,15 +191,13 @@ class FilesReportPlugin extends ServerPlugin {
 			return;
 		}
 
-		$ns = '{' . $this::NS_OWNCLOUD . '}';
-		$ncns = '{' . $this::NS_NEXTCLOUD . '}';
 		$requestedProps = [];
 		$filterRules = [];
 
 		// parse report properties and gather filter info
 		foreach ($report as $reportProps) {
 			$name = $reportProps['name'];
-			if ($name === $ns . 'filter-rules') {
+			if ($name === self::NS_OWNCLOUD_PREFIX . 'filter-rules') {
 				$filterRules = $reportProps['value'];
 			} elseif ($name === '{DAV:}prop') {
 				// propfind properties
@@ -205,7 +208,7 @@ class FilesReportPlugin extends ServerPlugin {
 				foreach ($reportProps['value'] as $propVal) {
 					if ($propVal['name'] === '{DAV:}nresults') {
 						$limit = (int)$propVal['value'];
-					} elseif ($propVal['name'] === $ncns . 'firstresult') {
+					} elseif ($propVal['name'] === self::NS_NEXTCLOUD_PREFIX . 'firstresult') {
 						$offset = (int)$propVal['value'];
 					}
 				}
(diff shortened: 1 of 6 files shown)


apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -363,6 +364,33 @@ protected function processFilterRulesForFileNodes(array $filterRules, ?int $limi
 			}
 		}
 
+		$ns = '{' . $this::NS_OWNCLOUD . '}';
```
blizzz, line 367: add as const like it is for systemtags and circles.

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -363,6 +364,33 @@ protected function processFilterRulesForFileNodes(array $filterRules, ?int $limi
...
+		$favoriteFilter = null;
+		foreach ($filterRules as $filterRule) {
+			if ($filterRule['name'] === $ns . 'favorite') {
+				$favoriteFilter = true;
```
blizzz, line 371: can break the loop afterwards (will not gain significant speed, of course, but still no reason to waste cycles as principle)

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -363,6 +364,33 @@ protected function processFilterRulesForFileNodes(array $filterRules, ?int $limi
...
+			}
+			if ($nodes === []) {
+				// there cannot be a common match when nodes are empty early.
+				return $nodes;
+			}
+
+			if ($limit !== null || $offset !== null) {
+				$nodes = array_slice($nodes, $offset, $limit);
+			}
+		}
```
blizzz, line 392: Duplicated code, move it to a separate func

lib/private/Files/Node/Folder.php
```diff
@@ -295,8 +295,8 @@ public function searchByMime($mimetype) {
 	 * @param string $userId owner of the tags
 	 * @return Node[]
 	 */
-	public function searchByTag($tag, $userId) {
-		$query = $this->queryFromOperator(new SearchComparison(ISearchComparison::COMPARE_EQUAL, 'tagname', $tag), $userId);
+	public function searchByTag($tag, $userId, int $limit = 0, int $offset = 0) {
```
blizzz, line 298: I suppose the interface needs to be extended as well? Though I don't see a psalm warning.
LazyFolder should look the same, alone for consistency.

blizzz (commented):
The favorites logic in `processFilterRulesForFileIDs()` should be removed, also. 

All the comments are no-brainers and I understand due to the fact this being an early draft. The approach is fine!

lib/private/Files/Node/Folder.php
juliushaertl: Adjusted, would need another adjustment in groupfolders again once merged.

For potential backports this is a bit more tricky but we may just use an ugly direct call on the private class implementation of a separate method then to avoid the interface adjustment.

juliushaertl:
Ready for review now

lib/private/Files/Node/Folder.php
blizzz: The foundational PRs were also backported and there the `Folder` changes were indeed made private. Since the signature is different, it's likely taking more to make at least tooling accept that state.

lib/private/Files/Node/NonExistingFolder.php
```diff
@@ -150,7 +150,7 @@ public function searchByMime($mimetype) {
 		throw new NotFoundException();
 	}
 
-	public function searchByTag($tag, $userId) {
+	public function searchByTag($tag, $userId, int $limit = 0, int $offset = 0) {
```
blizzz, line 153: That class is a good finding, i shall add searchBySystemTag here also – will be a separate PR.

blizzz (approved):

apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
```diff
@@ -345,27 +330,57 @@ protected function processFilterRulesForFileNodes(array $filterRules, ?int $limi
...
 			}
 
 			if (!$oneTagSearch && ($limit !== null || $offset !== null)) {
-				$nodes = array_slice($nodes, $offset, $limit);
+				$nodes = array_slice($nodes, $offset ?? 0, $limit);
+			}
+		}
+
+		if ($this->hasFilterFavorites($filterRules)) {
+			$tmpNodes = $this->userFolder->searchByTag(ITags::TAG_FAVORITE, $this->userSession->getUser()->getUID(), $limit ?? 0, $offset ?? 0);
```
marcelklehr, line 346: Isn't this this thing again where we slice multiple times and then intersect the results? Does this give proper results? I forgot.

marcelklehr:
```
limit = 3
offset = 2

a = [0,3,6,9,12,15,18,21,24]
b = [0,2,4,6,8,10,12,14,16,18,20,22,24]

slice(a) // 6,9,12
slice(b) // 4,6,8

intersect(slice(a), slice(b)) // 6

intersect(a,b) // 0,6,12,18,24
slice(intersect(a,b)) // 12,18,24
```

marcelklehr: So, it can give too little results. Sometimes even none when some would be expected.

juliushaertl: Good point, will see what could be done about that

icewind1991: Since both the `searchBySystemTag` and `searchByTag` end up doing a `search` with their specific search queries anyway. You can instead build a single search query that performs all the filtering in one go.

icewind1991: and you can probably also include the filtering for the fileids from `processFilterRulesForFileIDs` in the same query to remove the need for the `findNodesByFileIds` later down

marcelklehr: > You can instead build a single search query that performs all the filtering in one go.

Search stuff is not in OCP, though. Would be awesome if it was.

icewind1991: using it in the dav app is fine since it's releases are coupled with core

PR #39313, opened 2023-07-11
user_ldap: Add OpenAPI spec

provokateurin:
## Summary

From https://github.com/nextcloud/server/pull/36666. Adds the necessary annotations and descriptions.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Controller/ConfigAPIController.php
+++ b/apps/user_ldap/lib/Controller/ConfigAPIController.php
@@ -29,6 +29,7 @@ use OC\Security\IdentityProof\Manager;
 use OCA\User_LDAP\Configuration;
 use OCA\User_LDAP\ConnectionFactory;
 use OCA\User_LDAP\Helper;
+use OCP\AppFramework\Http;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCS\OCSBadRequestException;
 use OCP\AppFramework\OCS\OCSException;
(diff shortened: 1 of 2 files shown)


apps/user_ldap/lib/Controller/ConfigAPIController.php
```diff
@@ -220,83 +166,24 @@ public function modify($configID, $configData) {
...
-	 *     <hasMemberOfFilterSupport></hasMemberOfFilterSupport>
-	 *     <useMemberOfToDetectMembership>1</useMemberOfToDetectMembership>
-	 *     <ldapExpertUsernameAttr>uid</ldapExpertUsernameAttr>
-	 *     <ldapExpertUUIDUserAttr>uid</ldapExpertUUIDUserAttr>
-	 *     <ldapExpertUUIDGroupAttr></ldapExpertUUIDGroupAttr>
-	 *     <lastJpegPhotoLookup>0</lastJpegPhotoLookup>
-	 *     <ldapNestedGroups>0</ldapNestedGroups>
-	 *     <ldapPagingSize>500</ldapPagingSize>
-	 *     <turnOnPasswordChange>1</turnOnPasswordChange>
-	 *     <ldapDynamicGroupMemberURL></ldapDynamicGroupMemberURL>
```
provokateurin, line 283: I would like to preserve this information, but it's too much work right now because of the config key mapping which already discussed a while ago with @come-nc

bigcat88 (approved):

apps/user_ldap/lib/Controller/ConfigAPIController.php
provokateurin: @blizzz @come-nc What is your opinion on this?

come-nc: You can keep the comment, no? What is the question exactly? You would like to generate this from the mapping in the code?

apps/user_ldap/openapi.json
```diff
@@ -0,0 +1,384 @@
+{
+    "openapi": "3.0.3",
+    "info": {
+        "title": "user_ldap",
+        "version": "0.0.1",
```
come-nc, line 5: This does not need to match the application version?

provokateurin: No, this is the document and not the app version. I haven't yet figured out how to best utilize this field.

apps/user_ldap/lib/Controller/ConfigAPIController.php
provokateurin: We can keep the comment. I would prefer if it was possible to document this mapping in the spec itself and not just in a comment (which will be included in the spec as well).

provokateurin: I just had another look and this whole mapping thing is really bad. It's not possible to document this at all, so I will just bring back the comment and call it a day.

julien-nc (approved):

PR #39317, opened 2023-07-11
Change color for primary element on hover

### LINKED ISSUE #4193: [BITV] Buttons accessibility improvements
Update: I hid the previous conversation as we were going in the wrong direction. I'm updating the description with the specific accessibility problems that we have in the button component as per the call with the expert last week.  

To be addressed:

- [x] Focus state: currently there's not enough contrast between the background of the buttons and the dark outline we have. 
Solution: Add 2 outlines as focus state to buttons, both 2px thick. The inner should be `--color-main-text`, and the outer `--color-main-background`

Screenshot 2023-06-30 at 11 12 46

- [x] Primary button hover state doesn't meet text contrast requirements (needs solution)
- [x] Special buttons don't meet text contrast requirements (needs solution)

JuliaKirschenheuter:
* Fixes: https://github.com/nextcloud/nextcloud-vue/issues/4193

## Summary

Fixes color for primary element on hover state https://github.com/nextcloud/nextcloud-vue/issues/4193#issuecomment-1630476883

## Checklist

### DIFF
--- a/apps/theming/css/default.css
+++ b/apps/theming/css/default.css
@@ -72,7 +72,7 @@
   --color-primary-light-text: #002a41;
   --color-primary-light-hover: #dbe5ea;
   --color-primary-element: #006aa3;
-  --color-primary-element-hover: #3287b5;
+  --color-primary-element-hover: #1f7cae;
   --color-primary-element-text: #ffffff;
   --color-primary-element-light: #e5f0f5;
   --color-primary-element-light-hover: #dbe5ea;
--- a/apps/theming/lib/Themes/CommonThemeTrait.php
+++ b/apps/theming/lib/Themes/CommonThemeTrait.php
@@ -64,7 +64,7 @@ trait CommonThemeTrait {
 
 			// used for buttons, inputs...
 			'--color-primary-element' => $colorPrimaryElement,
-			'--color-primary-element-hover' => $this->util->mix($colorPrimaryElement, $colorMainBackground, 60),
+			'--color-primary-element-hover' => $this->util->mix($colorPrimaryElement, $colorMainBackground, 75),
 			'--color-primary-element-text' => $this->util->invertTextColor($colorPrimaryElement) ? '#000000' : '#ffffff',
 
 			// used for hover/focus states


artonge (approved):

skjnldsv (changes requested):
Please adjust generation script 

https://github.com/nextcloud/server/blob/0fca8238dd86d572f0f15eaa8e8a5ea836cc76bf/apps/theming/lib/Themes/CommonThemeTrait.php#L67

JuliaKirschenheuter:
@skjnldsv could you please help me how should this mix looks like?

skjnldsv:
> @skjnldsv could you please help me how should this mix looks like?

I can't, i have no idea how your hex colour got chosen.
Usage is as follow


> mix($hex, [$amount] ) : Allows you to mix another color to your color. Optionally you can decide to set the percent of second color or original color amount is ranged -100...0...100.

You have to pick the right amount to get as close as your `#2c7cae` as possible :shrug:

nimishavijay (approved):

skjnldsv:
> > I can't, i have no idea how your hex colour got chosen.
> 
> It is the slightly darker and bluer version of the previous hover color, but there was no logic per se. I can try and find another close color, how does the mix work?

Using `65` as a mix factor, we get `#2c84b3`. Which seems close to your `#2c7cae` @nimishavijay 

___
You can try this here: https://play.phpsandbox.io/mexitek/phpcolors
```php
<?php
use Mexitek\PHPColors\Color;

$color = new Color('#006AA3');
$background = new Color('#ffffff');

echo '#' . $color->mix($background, 65) . PHP_EOL;
```

nimishavijay:
Super nice! Thank you @skjnldsv :)

Unfortunately the color code you suggested does not have enough contrast against the main background color. However, using `75` as the mix factor, we get `#1f7cae` which is also basically the same and is accessible 🎉 so @JuliaKirschenheuter we can use  `#1f7cae` as the color here :) 

image

skjnldsv:
> Unfortunately the color code you suggested does not have enough contrast against the main background color.

This is not a variable used for text btw, so contrast rules are a bit more flexible for big elements iirc :)
But all good for me with `#1f7cae`  too

JuliaKirschenheuter:
@nimishavijay @skjnldsv 

After several tryings just saw you discussion. 
Thanks a lot for your cooperation, i've taken a mix factor `75` and `#1f7cae` as color ;))

nimishavijay (approved):
🚀 :)

Pytal (approved):

marcoambrosini (approved):

skjnldsv (approved):

PR #39320, opened 2023-07-11
files_trashbin: Add OpenAPI spec

provokateurin:
## Summary

From https://github.com/nextcloud/server/pull/36666. Adds the necessary annotations and descriptions.

## Checklist

### DIFF
--- a/apps/files_trashbin/openapi.json
+++ b/apps/files_trashbin/openapi.json
@@ -1,0 +1,134 @@
+{
+    "openapi": "3.0.3",
+    "info": {
+        "title": "files_trashbin",
+        "version": "0.0.1",
+        "description": "This application enables users to restore files that were deleted from the system.",
+        "license": {
+            "name": "agpl"
+        }
+    },
+    "components": {
+        "securitySchemes": {
+            "basic_auth": {
+                "type": "http",
+                "scheme": "basic"
+            },
+            "bearer_auth": {
+                "type": "http",
+                "scheme": "bearer"
+            }
+        },
+        "schemas": {
+            "Capabilities": {
+                "type": "object",
+                "required": [
+                    "files"
+                ],
+                "properties": {
+                    "files": {
+                        "type": "object",
+                        "required": [
+                            "undelete"
+                        ],
+                        "properties": {
+                            "undelete": {
+                                "type": "boolean"
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    },
+    "paths": {
+        "/index.php/apps/files_trashbin/preview": {
+            "get": {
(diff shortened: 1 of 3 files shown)


apps/files_trashbin/openapi.json
```diff
@@ -0,0 +1,134 @@
...
+                ],
+                "parameters": [
+                    {
+                        "name": "fileId",
+                        "in": "query",
+                        "description": "ID of the file",
+                        "schema": {
+                            "type": "integer",
+                            "format": "int64",
+                            "default": 1
```
bigcat88, line 68: maybe `-1`?

bigcat88 (changes requested):

apps/files_trashbin/openapi.json
provokateurin: Fixed it, there was a missing minus when parsing the default value.

bigcat88 (approved):

apps/files_trashbin/openapi.json
```diff
@@ -0,0 +1,134 @@
...
+                            "default": 32
+                        }
+                    },
+                    {
+                        "name": "a",
+                        "in": "query",
+                        "description": "Whether to crop the preview",
+                        "schema": {
+                            "type": "integer",
+                            "default": 0
```
Pytal, line 97:
```suggestion
                            "type": "boolean",
                            "default": false
```

apps/files_trashbin/openapi.json
```diff
@@ -0,0 +1,134 @@
...
+                        "schema": {
+                            "type": "integer",
+                            "format": "int64",
+                            "default": 32
+                        }
+                    },
+                    {
+                        "name": "a",
+                        "in": "query",
+                        "description": "Whether to crop the preview",
```
Pytal, line 94:
```suggestion
                        "description": "Whether to not crop the preview",
```

apps/files_trashbin/openapi.json
provokateurin: This is due to a quirk how booleans are handled by the server. Any value will result in `true`, except when sending a `0` which will turn into `false`. In theory [`allowEmptyValue`](https://swagger.io/docs/specification/describing-parameters/#nullable) exists and could be used to describe this case with the value being a boolean, but none of the generators that are widely used implement it (https://github.com/search?q=repo%3AOpenAPITools%2Fopenapi-generator%20allowEmptyValue&type=code). For the clients that are generated with them all boolean parameters would be broken, that's why the integer variant is used here.

skjnldsv (approved):

Pytal (approved):

PR #39333, opened 2023-07-12
[stable27] fix(files): sanitize name and ext display

skjnldsv:
Backport of https://github.com/nextcloud/server/commit/28c2d9e9063b05ad6eff9a227764a9571db4fcbe from https://github.com/nextcloud/server/pull/38950


apps/files/src/components/FileEntry.vue
```diff
@@ -59,7 +59,11 @@
 				</span>
 
 				<!-- File name -->
-				<span class="files-list__row-name-text">{{ displayName }}</span>
+				<span class="files-list__row-name-text">
+					<!-- Keep the displayName stuck to the extension to avoid whitespace rendering issues-->
+					<span class="files-list__row-name-name" v-text="displayName" />
```
susnux, line 64: We use the mustache notation everywhere else, sure we want to mix with the `v-text` notation?

susnux (approved):

szaimen (approved):
🐘

apps/files/src/components/FileEntry.vue
skjnldsv: Yes, because otherwise we have whitespace in the output too.
Vue does NOT trim whitespace. And for this specific example, we want to have none rendered. :)

PR #39353, opened 2023-07-13
[stable25] test(psalm): ignore known reports

blizzz:
this is code we also have in master, but raises flags here due to "race conditions" in merging and backporting competing with enabling psalm reacting on those types.


provokateurin (approved):

skjnldsv (approved):

blizzz:
merging as it is no production code change

PR #39366, opened 2023-07-13
prevent recursion with log events

icewind1991:
Prevent the event handling of a log event from creating more log events and causing infinite recursion.

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -210,6 +210,7 @@ class Log implements ILogger, IDataLogger {
 	 * @return void
 	 */
 	public function log(int $level, string $message, array $context = []) {
+		static $inLogEvent = false;
 		$minLevel = $this->getLogLevel($context);
 
 		array_walk($context, [$this->normalizer, 'format']);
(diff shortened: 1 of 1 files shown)


lib/private/Log.php
```diff
@@ -210,15 +210,21 @@ public function debug(string $message, array $context = []) {
...
 		$entry = $this->interpolateMessage($context, $message);
 
-		if ($this->eventDispatcher) {
-			$this->eventDispatcher->dispatchTyped(new BeforeMessageLoggedEvent($app, $level, $entry));
+		if ($this->eventDispatcher && !$inLogEvent) {
+			$inLogEvent = true;
+			try {
+				$this->eventDispatcher->dispatchTyped(new BeforeMessageLoggedEvent($app, $level, $entry));
+			} catch (Throwable $e) {
+				// ignore exceptions from inside the logger, as trying to log them would lead to recursion
```
come-nc, line 226: But that means a syntax error in BeforeMessageLoggedEvent class would go unnoticed, no?

icewind1991: I guess, but that seems better than breaking any logging from such an error.

(also static analysis would detect that specific error)

icewind1991: True, but an exception in the logging would still cause issues. We have similar logic in other logging places iirc

come-nc (approved):

skjnldsv:
Conflicts

PR #39367, opened 2023-07-13
handle more exceptions in AmazonS3::fopen

icewind1991:
returning `false` instead of throwing is the expected behavior.


icewind1991:
Other failing external storage tests

- vsftpd is a docker related issue
- kerberos sso is "complicated" and probably needs adjustments to the scripted setup

come-nc (approved):

Pytal (approved):

PR #39378, opened 2023-07-13
fix: Always apply fullscreen styling to sidebar

juliusknorr:
## Summary

Make sure that the full screen styling is always applied to the sidebar even if the loading hasn't finished yet.

Fixes a small UI glitch where a small space was left between the viewer modal and the sidebar.

This is a different approach I came up before stumbling over https://github.com/nextcloud/viewer/pull/1711 so I'd appreciate feedback especially from @ShGKme on this. It seems to also cover the drawback that the viewer PR has.

### Before

Screenshot 2023-07-13 at 22 15 23

### After

https://github.com/nextcloud/server/assets/3404133/69136fa3-3a6c-4006-ad8f-501f6c1a44fb

## TODO

- [ ] ...

## Checklist


ShGKme (commented):
I like this solution!

I didn't finish that PR in viewer because didn't find a good solution from the viewer side but didn't want to modify the server part :D
But this is definitely better.

---

Unfortunately, it still doesn't work in one case:
1. Open files
2. Open sidebar
3. Open viewer

image

It works in combination with my PR in viewer, but with a 0.2s transition...

ShGKme (approved):
Approved, regardless viewer, IMO it is better to have correct fullscreen state in all sidebar states in files.

Also, I'd consider to make fullscreen mode `NcAppSidebar` component feature, so we would be able to use it in any apps and have better sidebar integration in apps (for example, having shared fullscreen / resize events).

jancborchardt (commented):
@juliushaertl in your screen recording the viewer and sidebar open and close visibly at slightly different times when clicking the close "x" and the file name respectively. Is this something new based on the approach? (Maybe it’s also just because of the screen recording.)

jancborchardt (approved):
Looks good design-wise.

PR #39385, opened 2023-07-13
fix(a11y): Files table roles

Pytal:
## Summary

Before | After
--- | ---
image | image

## Checklist


apps/files/src/components/FilesListVirtual.vue
```diff
@@ -143,8 +143,10 @@ export default Vue.extend({
 	mounted() {
 		// Make the root recycle scroller a table for proper semantics
 		const slots = this.$el.querySelectorAll('.vue-recycle-scroller__slot')
-		slots[0].setAttribute('role', 'thead')
-		slots[1].setAttribute('role', 'tfoot')
+		slots[0].classList.add('files-list__header')
```
JuliaKirschenheuter, line 146: i don't have another solution right now but would it be probably possible to set this class direct to an element?

JuliaKirschenheuter (commented):
To which ticket this PR belongs to?

apps/files/src/components/FilesListVirtual.vue
Pytal: Would have done that if possible but this programmatic solution is the only available solution as the component from `vue-virtual-scroller` does not allow us to define a custom class another way :(

Pytal:
> To which ticket this PR belongs to?

Files2Vue a11y so would call this a followup to https://github.com/nextcloud/server/pull/36534

JuliaKirschenheuter (approved):

apps/files/src/views/FilesList.vue
```diff
@@ -23,7 +23,7 @@
 	<NcAppContent v-show="!currentView?.legacy"
 		:class="{'app-content--hidden': currentView?.legacy}"
 		data-cy-files-content>
-		<div class="files-list__header">
+		<div class="files-list__crumbs">
```
skjnldsv, line 26: Please no, more things than breadcrumbs are/will-be in here

apps/files/src/components/FilesListVirtual.vue
```diff
@@ -143,8 +143,10 @@ export default Vue.extend({
 	mounted() {
 		// Make the root recycle scroller a table for proper semantics
 		const slots = this.$el.querySelectorAll('.vue-recycle-scroller__slot')
-		slots[0].setAttribute('role', 'thead')
```
skjnldsv, line 146: Why is the thead not needed anymore suddenly?
The first table tr row is exactly a header, no?

https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Advanced#the_scope_attribute
image

skjnldsv (changes requested):

Pytal:
Superseded by https://github.com/nextcloud/server/pull/39808

PR #39397, opened 2023-07-14
[stable26] feat: Add public event for missing indices

juliusknorr:
Backport of https://github.com/nextcloud/server/pull/39389
Required for https://github.com/nextcloud/text/pull/4521

### DIFF
--- a/lib/public/DB/Events/AddMissingIndicesEvent.php
+++ b/lib/public/DB/Events/AddMissingIndicesEvent.php
@@ -1,0 +1,59 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2023 Julius Härtl <jus@bitgrid.net
+ *
+ * @author Julius Härtl <jus@bitgrid.net
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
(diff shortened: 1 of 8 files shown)


lib/public/DB/Events/AddMissingIndicesEvent.php
```diff
@@ -0,0 +1,59 @@
...
+	public function addMissingIndex(string $tableName, string $indexName, array $columns): void {
+		$this->missingIndices[] = [
+			'tableName' => $tableName,
+			'indexName' => $indexName,
+			'columns' => $columns
+		];
+	}
+
+	/**
+	 * @since 28.0.0
```
nickvergessen, line 53: Should we fix the version? Or keep it like this

nickvergessen (approved):

mejo- (approved):

lib/public/DB/Events/AddMissingIndicesEvent.php
juliushaertl: I'd keep it as then we are clear about that app developers shouldn't rely on this as it was only backported with a patch release

PR #39399, opened 2023-07-14
Allow using the profiler to selectively profile production instances

icewind1991:
- Fix app loading events not being ended correctly
- Fix legacy dav endpoints not being profiled
- Fix clearing profiles
- Add option to enable the profiler for a specific request using a shared secret in the request params
  - This is similar to the logic that exists to enable debug logging per request
  - This doesn't require the profiler app to be enabled and can thus be used to collect some profiles in production and then analyse them offline
  - More options for triggering the profiler should be added later (per user, per url, etc). But for now the shared secret will do as a POC

todo:

- [x] move the data collectors to core so profiles without the profiler app enabled have more info
- [x] move some of the `occ profiler:*` commands to core so profiles can be exported without requiring the profiler app.
- [ ] should we instead split up the profiler app into a front and backend? Where we can release/ship the backend without having to release the frontend


skjnldsv:
conflicts

PR #39400, opened 2023-07-14
fix(i18n) Changed case of "Files"

rakekniven:
Reported at Transifex.


skjnldsv (approved):

nickvergessen:
@rakekniven in case you rebased because of this:
grafik


You only need/should do it when the branch is quite old/conflicted. It's not necessary to do it every day or even multiple times a day. That just delays CI

skjnldsv:
Please also fix tests
```
FAIL apps/files_sharing/src/actions/openInFilesAction.spec.ts
  ● Open in files action conditions tests › Default values

    expect(received).toBe(expected) // Object.is equality

    Expected: "Open in files"
    Received: "Open in Files"

      51 | 		expect(action).toBeInstanceOf(FileAction)
      52 | 		expect(action.id).toBe('open-in-files')
    > 53 | 		expect(action.displayName([], validViews[0])).toBe('Open in files')

```

nickvergessen (approved):

PR #39446, opened 2023-07-18
Refactor user_ldap group membership cache and add check-group command

### LINKED ISSUE #29727: Force sync a ldap GROUP through command line or API
### How to use GitHub

Today, an internal workflow allows a user to create a share for a group (groupfolder).
The group creation is done in our LDAP, and once the group created, a script triggers on Nextcloud side the creation of the share.
The existence of the group can be chacked via a occ ldap :search --group myGroup.
The group then exists on Nextcloud side, but isn't populated: we then set permissions on an empty group, users seeing the group only once the group content had been updated by the Cron job.

A command `occ ldap: group-update myGroup` (or a group content feeding when a `occ ldap :search --group` command if the group wouldn't be referenced) would then make our process "immediate".

like `occ ldap:check-group --fetch $GROUPNAME` or `occ ldap:check-group --update $GROUPID`

**Additional content**
A similar demand exists to force the sync of a specific LDAP user : https://github.com/nextcloud/server/issues/27948

come-nc:
* Resolves: #29727

## Summary

Refactor user_ldap group membership cache, to be able to easily check if a membership is known or not.
Add a command check-group which works like check-user, with the same --update option to force update from LDAP.

## TODO

- [x] Figure out if indexes are needed on the new table
- [x] Improve command output (maybe use event listeners?)
- [x] Make sure command ignores cache for `--update` (currently not the case)

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required https://github.com/nextcloud/documentation/pull/11079


come-nc:
```json
{
"level":3,
"message":"Table \"oc_ldap_group_membership\" has no primary key and therefor will not behave sane in clustered setups. This will throw an exception and not be installable in a future version of Nextcloud.",
}
```

- [ ] Add primary key

come-nc:
- [ ] It only works if using full dn for non-mapped groups, would be better if using name works as well (for user it’s complicated but for groups it might be simpler to do that)
- [ ] It does not seem to correctly fire the group created event when passed a full dn of a new group

come-nc:
The `GroupCreatedEvent` firing is not completely working, but since GroupDeletedEvent is not fired I’m letting that for a follow-up PR.

apps/user_ldap/lib/Command/CheckGroup.php
```diff
@@ -0,0 +1,173 @@
...
+			$this->backend->getGroups($gid);
+			$exists = $this->backend->groupExistsOnLDAP($gid, true);
+			if ($exists === true) {
+				$output->writeln('The group is still available on LDAP.');
+				if ($input->getOption('update')) {
+					$this->backend->getLDAPAccess($gid)->connection->clearCache();
+					$this->updateGroup($gid, $output, $wasMapped);
+				}
+				return 0;
+			} elseif ($wasMapped) {
```
Altahrim, line 106: Since you use `return`, we can avoid the `else`:
```suggestion
			}
			if ($wasMapped) {
```
Same for lines `109-111`

apps/user_ldap/lib/Db/GroupMembershipMapper.php
```diff
@@ -0,0 +1,77 @@
...
+	/**
+	 * @return string[]
+	 */
+	public function getKnownGroups(): array {
+		$query = $this->db->getQueryBuilder();
+		$result = $query->selectDistinct('groupid')
+			->from($this->getTableName())
+			->executeQuery();
+
+		$groups = array_map(fn ($row) => $row['groupid'], $result->fetchAll());
```
Altahrim, line 50: No need to bind `$this`:
```suggestion
		$groups = array_column($result->fetchAll(), 'groupid');
```

Or maybe this one with DBAL: 
```suggestion
		$groups = $result->fetchFirstColumn();
```

apps/user_ldap/lib/Migration/Version1190Date20230706134108.php
```diff
@@ -0,0 +1,113 @@
...
+	public function preSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
+	}
+
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('ldap_group_membership')) {
+			$table = $schema->createTable('ldap_group_membership');
+			$table->addColumn('id', 'integer', [
```
Altahrim, line 51: If we only need `length = 4`, `SMALLINT` should be enough:
```suggestion
			$table->addColumn('id', Types:SMALLINT, [
```

apps/user_ldap/lib/Service/UpdateGroupsService.php
```diff
@@ -0,0 +1,188 @@
...
+	 * @throws Exception
+	 */
+	public function handleKnownGroups(array $groups): void {
+		$this->logger->debug('service "updateGroups" – Dealing with known Groups.');
+
+		foreach ($groups as $group) {
+			$this->logger->debug('service "updateGroups" – Dealing with {group}.', ['group' => $group]);
+			$groupMemberships = $this->groupMembershipMapper->findGroupMemberships($group);
+			$knownUsers = array_map(
+				fn (GroupMembership $groupMembership): string => $groupMembership->getUserid(),
```
Altahrim, line 93:
```suggestion
				static fn (GroupMembership $groupMembership): string => $groupMembership->getUserid(),
```
I think we can avoid binding `$this `

apps/user_ldap/lib/Db/GroupMembershipMapper.php
come-nc: `fetchFirstColumn` does not exists on `OCP\DB\IResult`.
I’ll switch to array_column.

apps/user_ldap/lib/Migration/Version1190Date20230706134108.php
come-nc: Hum this was from a copy/paste I think, from what I understood all ids should be BIGINT. I’ll switch to bigint, there might be millions of users in millions of groups.

apps/user_ldap/lib/Command/CheckGroup.php
come-nc: I find it clearer with the else.

come-nc:
Note: I looked into using LazyUser for events, but the test expects the event to only fire for existing users, which makes sense I suppose, so I abandoned the idea.

apps/user_ldap/lib/Migration/Version1190Date20230706134108.php
Altahrim: Why keep `length = 4` then?

Altahrim (approved):

AndyScherzinger (approved):

come-nc:
Rebased and squashed

PR #39472, opened 2023-07-19
Improve sharing flow

### LINKED ISSUE #26691: 🤝 Improvement to sharing design and flow
Currently we have several issues with sharing, across platforms:
- It looks a bit cluttered, especially the 3-dot dropdown menu with all the actions is too much
- It’s not clear to people when the share will be set & sent, leading to possible access control issues, and also 2 separate mails when you set a note
- Names are often cut off because of the "can edit" with checkbox
- It could be more consistent across platforms

Here’s a proposal to fix these issues which came out of a design discussion @FlexW and I had about improving the sharing flow for the Desktop client. We should of course make sure to keep this the same across all platforms. (It is to be seen as a mockup – of course the inputs and buttons should be the native elements of the platform.)

Share overview | Share details | Share details advanced
-|-|-
Share overview | Link details vertical | Link details vertical advanced

Mockup with slight interactivity at ~~https://design.penpot.app/#/view/b83aa940-a28e-11eb-bb49-81aaa3a367f9/b83aa941-a28e-11eb-bb49-81aaa3a367f9?token=9eUwuY6LzqqZ34pWFWSUnQ&index=0~~ https://design.penpot.app/#/view/b83aa940-a28e-11eb-bb49-81aaa3a367f9?page-id=b83aa941-a28e-11eb-bb49-81aaa3a367f9&index=0

## Share overview
- We add a subline to the entries where info on "View only" or "Can edit" or "File drop" is shown – this makes name and subline area actionable
   - For external shares of single files where there’s no editing, the "Can edit" function needs to be not shown (anything except txt, md, or office probably)
   - Even if there is only one option, we should still show the subline for feedback
- The avatar needs to stay a separate actionable area for the contact options
- Another direct action in the list is to "Copy link" because that needs to be quickly accessible.
- On the right there’s a 3-dot menu icon which brings up a menu with quick actions for
   - Share details
   - Resend email
   - Delete share / Unshare
- Going to "Share details" shows the detail page with further info
- The little primary-colored bar on the top is a little experiment and would mainly be something for the desktop client to make the dialog look less dead.
- **Alternatively on mobile:** Whole entry is tappable and opens the menu as a bottom sheet, in this case with
   - (Direct switch of View only / Edit / File drop if possible?)
   - Share details
   - Resend email
   - Delete share / Unshare

## Share detail
- Ideally, there is a quick slide left/right animation whenever one navigates in and out of this detail view so people have a sense of place.
- The header part needs to indicate both which file it is about, and if it is a share link or which person. In this mockup it is done by changing the header – this works for Desktop, for the other platforms we need to adjust.
- This opens with the relevant settings – when there is an expiration date set or enforced, the "Advanced settings" is already open
- Here all the share settings are listed, in this order:
   - View only / Edit / File drop
   - Note to recipient
   - Set password
   - Set expiration date
   - Share label
   - Delete link / Unshare
- Clicking the "Share & copy link" button automatically copies the link as expected. For shares to people, this could say "Share & send mail"

@nextcloud/designers and the leads @tobiasKaminsky @rullzer @marinofaggiana @mgallien @FlexW @LukasReschke @karlitschek

nfebe:
## Summary

Resolves: https://github.com/nextcloud/server/issues/26691


Fenn-CS:
# Share Details View

Share details | Share advanced (optional checks off) | Share details advanced (optional checks on)
-|-|-
share-details | share-details-advanced-checks-off | share-details-advanced

nimishavijay:
Looks super nice! 🚀 :) 
Some feedback from my side:
- [ ] The folder name under "Edit share link" can be removed, since it's already there in the top of the sidebar. Instead, if there is a share label set that can be shown in place of the folder name
- [ ] So the folder icon next to it can also be replaced by a [share icon](https://fonts.google.com/icons?selected=Material%20Icons%3Ashare%3A) 
- [ ] We could also add a few pixels of spacing between the icon and the text, and remove a bit of spacing between the "Edit share link" and its subline if it's shown. 
- [ ] The icon for the "Advanced settings" button can be the [expand more icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aexpand_more%3A)
- [ ] If the advanced settings are open, will clicking on "Advanced settings" button hide it again? Ideally it should, and in that case we can use the [expand less icon](https://fonts.google.com/icons?selected=Material%20Icons%3Aexpand_less%3A) instead
- [ ] The "Delete link" button can be styled to look like a tertiary icon-only button
- [ ] Generally we use sentence case for wording, so a captial letter only at the starting of the wording for each element:
  - [ ] Edit Share Link --> Edit share link
  - [ ] View only (Default) --> View only (default)
  - [ ] File Drop --> File drop
  - [ ] Advanced Settings --> Advanced settings

That's all from my side! It already looks great! :)

jancborchardt (commented):
> The folder name under "Edit share link" can be removed, since it's already there in the top of the sidebar. Instead, if there is a share label set that can be shown in place of the folder name

→ Reviews moved to comment below

Fenn-CS:
### Share flow updates

- Includes "missing detail" add here : https://github.com/nextcloud/server/issues/26691#issuecomment-1671537196
- [x] The folder name under "Edit share link" can be removed, since it's already there in the top of the sidebar. Instead, if there is a share label set that can be shown in place of the folder name
- [x] So the folder icon next to it can also be replaced by a [share icon](https://fonts.google.com/icons?selected=Material%20Icons%3Ashare%3A) 
- [x] The "Delete link" button can be styled to look like a tertiary icon-only button
- [x] Generally we use sentence case for wording, so a captial letter only at the starting of the wording for each element:
- [x] Some function wiring on the UI


Share view| Share details | Share advanced (most optional checks off) | Share details advanced (more optional checks on)
-|-|-|-
share-flow-1 | share-flow-2 | share-flow-3 | share-flow-4

jancborchardt (changes requested):
- [x] **"Share link" should use the same share details interface as internal shares**
  - [x] **Needs the 3-dot menu to have 3 entries, "Customize link", "Add another link" and "Unshare link" as per https://github.com/nextcloud/server/pull/39472#issuecomment-1682033889**
- [x] **File drop (Upload only functionality for shared links) needs to be fixed**
- [x] **When adding a new person or creating a new link, the detail view should always be opened. The share should only be created when you click "Share & copy link"**
   - [x] Type name in search --> click on person --> open details view 
- [x] **The "File drop" option should not show up for internal shares as we only have that for share links**

One thing which came up in a review but we decided against it is to expand the "Advanced settings" by default since it’s too complex. Instead:
- [ ] Advanced settings will expand automatically IF
   - [ ] for new shares if anything like password or expiration date is enforced
   - [x] for existing shares if anything in the advanced settings is set
- [x] "Custom permissions" is shown in the simple switcher on the top after all, also because only showing it conditionally caused jumping in the interface. This also makes advanced settings more easily reachable.



## Share view
- [ ] Currently links are sorted below emails once created, but sorting should be: Links, mails, users
- [x] **If custom permissions are set, the subline needs to show "Custom permissions" as indicator**
- [x] The arrow-down icon is too complex, it should only be a simple triangle icon as per mockup: https://pictogrammers.com/library/mdi/icon/triangle-small-down/
- [x] The "Can view/edit/share/file drop" is too wide, only needs to be as wide as the text is.
- [x] The text of the "Share link" and the person "Ambe Bandolo" does not left-align with each other
- [x] Also the circles on the left for Share link and avatar are much bigger than before, leading to them being bigger than the 2 below for "Others with access" and "Internal link". All of them should be same size
- [x] Subline text "Can view" should be called "View only"

## Share details & advanced
- [ ] **Navigating away from an active share flow needs to reset the view as per https://github.com/nextcloud/server/pull/39472#discussion_r1297020665**
- [x] **The subline of "Custom permissions" is intended to reflect what is actually set in the custom permissions. So in the 4th screenshot it should say "Read". If that is too complex, better remove the subline altogether.**
- [x] Move "Note to recipient" down, above "Custom permissions" and have it enable-able using a checkbox, not expanded by default (unless filled)
- [x] **Instead of duplicating the share icon left of "Share with [Person]", if possible the avatar should be shown there. If it is a link share, show a link icon on the left.**
- [x] **We need to make sure that the "Cancel" and "Share & copy link" are sticky on the bottom**
- [x] **"Delete link" should be renamed to "Delete share" (both for links and internal shares)**
- [x] Clicking "Delete share" should also close the detail view
- [x] "Delete share" button needs a `--color-main-background` background so that when the content scrolls, it doesn’t overlap with the text
- [x] The subline can be removed, as the filename already shows on the top
- [x] "Advanced Settings" --> "Advanced settings" (sentence case, only first word capitalized)
- [x] "Custom permisions" → "Custom permi**ss**ions"
- [x] The "Cancel" button being half the width looks a bit massive. Could you make both buttons only be as wide as their contents, Cancel flush left and "Share & copy link" flush right?

I bolded the ones I consider most important. :)

szaimen (dismissed):
The new sharing flow does not seem to work for link shares yet:
<details><summary>Screenshot</summary>
<p>

</p>
</details>

jancborchardt (dismissed):
Further things which came up in the Hub 6 design review:
→ All moved up to the other review at https://github.com/nextcloud/server/pull/39472#pullrequestreview-1571500093

Fenn-CS:
Share flow updates

Share view| Share details | Share advanced (most optional checks off) | Share details advanced (more optional checks on)
-|-|-|-
 Coming soon | share-2 | share-3 | share-4

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -330,6 +334,9 @@ export default {
 
 	data() {
 		return {
+			showDropdown: false,
+			selectedOption: 'Can edit',
+			options: ['View only', 'Can edit', 'File drop'],
```
artonge, line 339: Those need to be translated.

apps/settings/src/components/UserList/UserRow.vue
artonge, line 1: Copy-pasting from the user list? Is this needed here?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+					<CloseIcon :size="20" />
+				</template>
+				{{ t('files_sharing', 'Delete share') }}
+			</NcButton>
+			<div class="button-group">
+				<NcButton @click="$emit('close-sharing-details')">
+					{{ t('file_sharing', 'Cancel') }}
+				</NcButton>
+				<NcButton type="primary" @click="shareAndCopyLink">
+					{{ t('file_sharing', 'Share & copy link') }}
```
juliushaertl, line 110: Copy is not implemented yet, but this raises a question for me to @jancborchardt:

Should it be the same button when creating new shares for users/groups/circles? Should it then copy the internal link?

In addition, should the button have a different label when updating an existing share or just always stick with "Share" as the action, even though the share might already exist?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+			permissionsCreate: false,
+			permissionsDelete: false,
+			permissionsRead: false,
+			permissionsShare: false,
+		}
+	},
+
+	computed: {
+		title() {
+			let title = 'Share with '
```
juliushaertl, line 209: Should be translated, ideally also with the shareWithDisplayName parameter included, so that translators have context. This should work, but we may nee to think about a fallback for types that don't have a shareWithDisplayName

```suggestion
			let title = t('Share with {displayName}', { displayName: this.share.shareWithDisplayName })
```

apps/files_sharing/src/views/SharingTab.vue
```diff
@@ -355,6 +350,18 @@ export default {
...
+		toggleShareDetailsView(eventData) {
+			if (eventData) {
+				// If no share request value, set to null 
+				// Implies existing share
+				if (!eventData.shareRequestValue) {
+					eventData.shareRequestValue = null;
+				}
+				this.shareDetailsData = eventData;
+			}
+			this.showSharingDetailsView = !this.showSharingDetailsView
```
juliushaertl, line 363: @jancborchardt  The showSharingDetailsView should also be reset if switching to a different file in the sidebar i think, right? Question then is how to handle unsaved changes to the share data. Blocking the change might get tricky.

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+			type: Object,
+			default: false,
+		},
+	},
+	data() {
+		return {
+			setPassword: false,
+			password: '',
+			noteToRecipient: '',
+			sharingPermission: 'rw',
```
juliushaertl, line 192: Would be nice to have constants for different predefined share permissions that not use a bitmask

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+				<NcCheckboxRadioSwitch :checked.sync="hasExpirationDate">
+					{{ t('file_sharing', 'Set expiration date') }}
+				</NcCheckboxRadioSwitch>
+				<NcDatetimePicker v-if="hasExpirationDate" v-model="expirationDate" type="datetime"
+					:placeholder="t('file_sharing', 'Expiration date')" confirm />
+				<NcCheckboxRadioSwitch :checked.sync="hideDownload">
+					{{ t('file_sharing', 'Hide download') }}
+				</NcCheckboxRadioSwitch>
+				<NcInputField :value="shareLabel" :label-visible="true" :label="t('file_sharing', 'Share label')" />
+				<span style="display: none;"><label /></span>
```
juliushaertl, line 72: Probably not needed?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+
+				<NcCheckboxRadioSwitch :checked.sync="hasExpirationDate">
+					{{ t('file_sharing', 'Set expiration date') }}
+				</NcCheckboxRadioSwitch>
+				<NcDatetimePicker v-if="hasExpirationDate" v-model="expirationDate" type="datetime"
+					:placeholder="t('file_sharing', 'Expiration date')" confirm />
+				<NcCheckboxRadioSwitch :checked.sync="hideDownload">
+					{{ t('file_sharing', 'Hide download') }}
+				</NcCheckboxRadioSwitch>
+				<NcInputField :value="shareLabel" :label-visible="true" :label="t('file_sharing', 'Share label')" />
```
juliushaertl, line 71: Share labels are only possible for links/email shares as far as I see

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+					{{ t('file_sharing', 'Set password') }}
+				</NcCheckboxRadioSwitch>
+				<NcInputField v-if="setPassword" type="password" :value="password" :label="t('file_sharing', 'Password')" />
+
+				<NcCheckboxRadioSwitch :checked.sync="hasExpirationDate">
+					{{ t('file_sharing', 'Set expiration date') }}
+				</NcCheckboxRadioSwitch>
+				<NcDatetimePicker v-if="hasExpirationDate" v-model="expirationDate" type="datetime"
+					:placeholder="t('file_sharing', 'Expiration date')" confirm />
+				<NcCheckboxRadioSwitch :checked.sync="hideDownload">
```
juliushaertl, line 68: Only possible for link/email shares

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+				<template #icon>
+					<MenuDownIcon />
+				</template>
+			</NcButton>
+		</div>
+		<div v-if="advancedSectionAccordionExpanded" class="sharingTabDetailsView__advanced">
+			<section>
+				<label>{{ t('file_sharing', 'Note to recipient') }}</label>
+				<textarea :model="noteToRecipient" />
+				<NcCheckboxRadioSwitch :checked.sync="setPassword">
```
juliushaertl, line 58: Only possible for link/email shares

juliusknorr (commented):
A first round of code comments

juliushaertl:
Missing functional aspects from a quick test run:

- [ ] Creating shares does not work
- [ ] Editing user shares does not work -> leads to empty sidebar
- [ ] Updating shares not tested
- [ ] Quick permission change dropdown doesn't work on link/email shares
- [ ] QA: Needs side by side testing if all old options are possible or also properly limited with the new one

juliushaertl:
@jancborchardt What I couldn't find in the spec was how additional share links can be created. Currently this is in the three dots menu next to an existing share link, but from my understanding the icon should lead to the details view now. How would a user then add an additional link?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,525 @@
...
+			permissionsRead: false,
+			permissionsShare: false,
+		}
+	},
+
+	computed: {
+		title() {
+			let title = 'Share with '
+			if (this.share) {
+				title = title + this.share.shareWithDisplayName
```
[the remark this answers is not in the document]
Fenn-CS: Thanks, done.

apps/files_sharing/src/views/SharingDetailsTab.vue
jancborchardt: Yes, good points:
**For the MVP, we can just go for the wording: "Save share"** @Fenn-CS – then it works for both internal and link shares and copy doesn’t need to be implemented.

> Should it be the same button when creating new shares for users/groups/circles? Should it then copy the internal link?

That would be ideal, but yes probably something for next iteration.

apps/files_sharing/src/views/SharingDetailsTab.vue
artonge: https://github.com/nextcloud/server/blob/f3fa006484c6dc7f3dd2504ea005dc3689ed4d44/apps/files_sharing/src/lib/SharePermissionsToolBox.js#L32-L37

apps/files_sharing/src/views/SharingTab.vue
jancborchardt: > The showSharingDetailsView should also be reset if switching to a different file in the sidebar i think, right?

Yep, that needs to be reset @Fenn-CS.

- MVP could be to just drop the changes as if "Cancel" was pressed
- For the future we can think of showing a notice

jancborchardt:
> @jancborchardt What I couldn't find in the spec was how additional share links can be created. Currently this is in the three dots menu next to an existing share link, but from my understanding the icon should lead to the details view now. How would a user then add an additional link?

As discussed in the call, once the share link is ported to the new flow, the 3 dot menu should have 2 entries:
- Link settings
- Add another link

apps/files_sharing/src/views/SharingDetailsTab.vue
jancborchardt: This is done already by @Fenn-CS :)

apps/files_sharing/src/mixins/SharesMixin.js
```diff
@@ -121,11 +125,26 @@ export default {
 				monthFormat: 'MMM',
 			}
 		},
-
+		isExternalShare() {
```
juliushaertl, line 128: external share is used as internal wording for ferderated shares in the backend, maybe rather call this `isPublicShare`?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,691 @@
...
+		UploadIcon,
+		MenuDownIcon,
+		DotsHorizontalIcon,
+	},
+	mixins: [ShareTypes, ShareRequests, SharesMixin],
+	props: {
+		shareRequestValue: {
+			type: Object,
+			required: false,
+		},
```
juliushaertl, line 217: I was thinking about this approach a bit and would have a suggestion to simplify this. We could only have a `share` property which then always has the same structure independent of if the share already exists or not. The only difference would be that a non-existing share that is in creating would not have share id yet as it doesn't exist in the backend. This could simplify the overal logic quite a bit. What do you think about that?

Fenn-CS:
From @juliushaertl (Putting this as a comment to follow up later if can't fit in here)

> I was thinking about this approach a bit and would have a suggestion to simplify this. We could only have a share property which then always has the same structure independent of if the share already exists or not. The only difference would be that a non-existing share that is in creating would not have share id yet as it doesn't exist in the backend. This could simplify the overal logic quite a bit. What do you think about that?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,695 @@
...
+				this.fileInfo.sharePermissions = sharePermissionsSet
+			}
+			if (this.setCustomPermissions) {
+				this.updateAtomicPermissions()
+			}
+			if (!this.share) {
+				this.addShare(this.shareRequestValue, this.fileInfo, new Config())
+
+			}
+			this.queueUpdate('permissions')
```
juliushaertl, line 448: I think the attributes would be missing for the "allow download" permission on internal shares, see https://github.com/nextcloud/server/blob/stable27/apps/files_sharing/src/components/SharingEntry.vue#L446-L450 for reference

marcoambrosini (commented):
Sorry for the late review, for some reason this pr doesn't run properly in my development environment so @Fenn-CS kindly walked me through the feature in a call. I also thought there was no time for making design adjustments anymore, that's why I didn't comment on the mockups before. So anyway here are my thoughts:

The main issue I find here is the attempt to pack complex share settings in a sidebar tab. 
  
I think the use of a sidebar is warranted if 2 conditions are met: 
**1) I need to to perform actions related to the content** (the files list in this case) 
**2) There's a usability gain given by not loosing sight of the content while I use the functionality in the sidebar**
If these are not both met, we should put complex flows front and center, give proper spacing to everything and take the attention away from the rest of the UI.

In this case, while it's maybe nice to be able to have a list of shares in the sidebar, **once I want to edit a share and dive into its settings, the files list doesn't matter anymore, so condition 2) is not met.**
At that point I don't need to see the files list, so we're free to bring up (you guessed it) a nice spacious dialog where we can articulate all the complexity. 
I think that's the reason why all our competitors do it that way.

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
```diff
@@ -0,0 +1,182 @@
...
+		font-size: 12.5px;
+		gap: 2px;
+		color: var(--color-primary-element);
+	}
+
+	.share-select-dropdown-container {
+		position: absolute;
+		top: 100%;
+		left: 0;
+		background-color: #ffffff;
```
juliushaertl, line 150: Please use css variables for any hardcoded color values, otherwise dark mode is broken:

Screenshot 2023-08-22 at 09 28 19

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
```diff
@@ -0,0 +1,182 @@
+<template>
+	<div :class="{ 'active': showDropdown, 'share-select': true }">
+		<span class="trigger-text" @click="toggleDropdown">
+			{{ selectedOption }}
+			<DropdownIcon :size="15" />
+		</span>
+		<div v-if="showDropdown" class="share-select-dropdown-container">
```
juliushaertl, line 7: Follow up: This is not accessible, we should use proper elements to make this keyboard navigatable and usable by screen readers

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,938 @@
...
+					:value="new Date(share.expireDate).toLocaleDateString({ year: 'numeric', month: '2-digit', day: '2-digit' })"
+					:label="t('file_sharing', 'Expiration date')"
+					@update:value="onExpirationChange" />
+				<!-- REVIEW DISCUSSION OR PAIR FIX: @update:value="onExpirationChange" does not trigger any event for NcDateTimePicker -->
+				<NcDatetimePicker v-if="hasExpirationDate"
+					v-model="share.expireDate"
+					type="datetime"
+					:placeholder="t('file_sharing', 'Expiration date')"
+					confirm
+					@update:value="onExpirationChange" />
```
Fenn-CS, line 119: With `NcActionInputFiled` providing `new Date(share.expireDate)` would initialize the component correctly this does not work for `NcInputFiled` on the other hand `NcDateTimePicker` does not respond to the `update:value` event.

So I have both inputs so testers can see how each behave side by side and provide me with some context on what could be going on here.

cc: @artonge @juliushaertl

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		isPasswordEnforced() {
+			return this.isPublicShare && this.config.enforcePasswordForPublicLink
+		},
+		isExpiryDateEnforced() {
+			return this.config.isDefaultInternalExpireDateEnforced || this.config.isDefaultRemoteExpireDateEnabled || this.config.isDefaultExpireDateEnabled
+		},
+		existingExpiryDate() {
+			if (this.config.isDefaultInternalExpireDateEnabled && (this.isGroupShare || this.isUserShare)) {
+				return new Date(this.config.defaultInternalExpirationDate)
+			} else if (this.config.isDefaultRemoteExpireDateEnabled && isRemoteShare) {
```
juliushaertl, line 444:
```suggestion
			} else if (this.config.isDefaultRemoteExpireDateEnabled && this.isRemoteShare) {
```

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+			return this.isFolder || allowedMimetypes.includes(this.fileInfo.mimetype)
+		},
+		isPasswordEnforced() {
+			return this.isPublicShare && this.config.enforcePasswordForPublicLink
+		},
+		isExpiryDateEnforced() {
+			return this.config.isDefaultInternalExpireDateEnforced || this.config.isDefaultRemoteExpireDateEnabled || this.config.isDefaultExpireDateEnabled
+		},
+		existingExpiryDate() {
+			if (this.config.isDefaultInternalExpireDateEnabled && (this.isGroupShare || this.isUserShare)) {
```
juliushaertl, line 442: I'd prefer switching the condition order to make this easier to understand:

```suggestion
			if ((this.isGroupShare || this.isUserShare) && this.config.isDefaultInternalExpireDateEnabled) {
```

Same for the others

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+					id="share-date-picker"
+					v-model="existingExpiryDate"
+					:min="dateTomorrow"
+					:max="dateMaxEnforced"
+					:hide-label="true"
+					:disabled="isExpiryDateEnforced"
+					:placeholder="t('file_sharing', 'Expiration date')"
+					type="date"
+					@input="onExpirationChange" />
+				<NcCheckboxRadioSwitch :disabled="canChangeHideDownload"
```
juliushaertl, line 124: Should only be visible on link or mail shares

apps/files_sharing/src/mixins/SharesMixin.js
```diff
@@ -121,11 +125,24 @@ export default {
 				monthFormat: 'MMM',
 			}
 		},
-
+		isFolder() {
+			return this.fileInfo.type === 'dir'
+		},
+		isPublicShare() {
+			const shareType = this.share.shareType ?? this.share.type
```
juliushaertl, line 132: Why are there two different properties for that?

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
```diff
@@ -0,0 +1,184 @@
...
+			if (option === this.customPermissionsText) {
+				this.$emit('open-sharing-details')
+			} else {
+				this.share.permissions = this.dropDownPermissionValue
+				this.queueUpdate('permissions')
+			}
+			this.showDropdown = false
+		},
+		initializeComponent() {
+			this.selectedOption = this.preSelectedOption
```
artonge, line 127: Couldn't `selectedOption` be initialized in `data` ?

apps/files_sharing/src/components/SharingInput.vue
```diff
@@ -452,7 +453,6 @@ export default {
 			}
 
 			return {
-				id: `${result.value.shareType}-${result.value.shareWith}`,
```
artonge, line 455: Not needed anymore ?

artonge: Is it to discriminate whether the share exists or not in `ShareDetails.js` ?
If so, I would instead add a property as this `id` might be used somewhere.

apps/files_sharing/src/mixins/SharesMixin.js
artonge, line 132: Same, why are there two ways of accessing the type ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
+<template>
+	<div class="sharingTabDetailsView">
+		<div class="sharingTabDetailsView__header">
+			<span>
```
artonge, line 4: Useless `span` ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+					class="sharing-entry__avatar"
+					:is-no-user="share.shareType !== SHARE_TYPES.SHARE_TYPE_USER"
+					:user="share.shareWith"
+					:display-name="share.shareWithDisplayName"
+					:menu-position="'left'"
+					:url="share.shareWithAvatar" />
+				<component :is="getShareTypeIcon(share.type)" :size="32" />
+			</span>
+			<span>
+				<h1>{{ title }}</h1>
```
artonge, line 15: `h1` might be confusing here. I would use a `div`. @JuliaKirschenheuter do you have an opinion ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+				@click="advancedSectionAccordionExpanded = !advancedSectionAccordionExpanded">
+				{{ t('files_sharing', 'Advanced settings') }}
+				<template #icon>
+					<MenuDownIcon />
+				</template>
+			</NcButton>
+		</div>
+		<div v-if="advancedSectionAccordionExpanded" class="sharingTabDetailsView__advanced">
+			<section>
+				<NcInputField v-if="isPublicShare"
```
artonge, line 85: Use `NcTextField` https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcInputField/NcInputField.vue#L25-L29

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		ViewIcon,
+		MenuDownIcon,
+		DotsHorizontalIcon,
+	},
+	mixins: [ShareTypes, ShareRequests, SharesMixin],
+	props: {
+		shareRequestValue: {
+			type: Object,
+			required: false,
+		},
```
artonge, line 248: Don't `shareRequestValue` contains `fileInfo` and `share` ? Isn't it redundant ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		share: {
+			type: Object,
+			required: true,
+		},
+	},
+	data() {
+		return {
+			writeNoteToRecipientIsChecked: false,
+			sharingPermission: BUNDLED_PERMISSIONS.ALL.toString(),
+			setCustomPermissions: false,
```
artonge, line 262: I would rename to `showCustomPermissions`

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		},
+		share: {
+			type: Object,
+			required: true,
+		},
+	},
+	data() {
+		return {
+			writeNoteToRecipientIsChecked: false,
+			sharingPermission: BUNDLED_PERMISSIONS.ALL.toString(),
```
artonge, line 261: Couldn't this be inferred from the share's permissions ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+			setCustomPermissions: false,
+			passwordError: false,
+			advancedSectionAccordionExpanded: false,
```
artonge, line 264: Just to keep a consistent naming scheme: `showAdvancedSettings`

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+			if (isChecked) {
+				this.sharingPermission = 'custom'
+			} else {
+				this.sharingPermission = BUNDLED_PERMISSIONS.ALL.toString()
+			}
+		},
+		async isPasswordProtected(isChecked) {
+			if (isChecked) {
+				if (!this.isFirstComponentLoad) {
+					this.share.password = await GeneratePassword()
```
artonge, line 651: There might be a simpler way of doing this, like generating the password when the share object is created. But this works.

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+				this.updateAtomicPermissions({ isDownloadChecked: checked })
+			},
+		},
+		/**
+		 * Is this share readable
+		 * Needed for some federated shares that might have been added from file drop links
+		 */
+		hasRead: {
+			get() {
+				if (this.isNewShare && !this.allowsFileDrop) {
```
artonge, line 352: Why is this condition needed ? Or why does the share object does not contain the permissions ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		 *
+		 * @return {boolean}
+		 */
+		hasExpirationDate: {
+			get() {
+				return this.config.isDefaultInternalExpireDateEnforced || !!this.share.expireDate
+			},
+			set(enabled) {
+				const defaultExpirationDate = this.config.defaultInternalExpirationDate
+					|| new Date(new Date().setDate(new Date().getDate() + 1))
```
artonge, line 372: There might be an easier way of doing this with `@nextcloud/moment`

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		 * Is the current share a folder ?
+		 *
+		 * @return {boolean}
+		 */
+		isFolder() {
+			return this.fileInfo.type === 'dir'
+		},
+		dateMaxEnforced() {
+			if (!this.isRemote && this.config.isDefaultInternalExpireDateEnforced) {
+				return new Date(new Date().setDate(new Date().getDate() + 1 + this.config.defaultInternalExpireDate))
```
artonge, line 404: Same, moment could help here.

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		},
+		allowsFileDrop() {
+			if (this.isFolder) {
+				if (this.share.type === this.SHARE_TYPES.SHARE_TYPE_LINK || this.share.type === this.SHARE_TYPES.SHARE_TYPE_EMAIL) {
+					return true
+				}
+			}
+			return false
+		},
+		hasFileDropPermissions() {
```
artonge, line 471: `isFileDropPermissions`

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+		customPermissionsList() {
+			const perms = []
+			if (hasPermissions(this.share.permissions, ATOMIC_PERMISSIONS.READ)) {
+				perms.push('read')
+			}
+			if (hasPermissions(this.share.permissions, ATOMIC_PERMISSIONS.CREATE)) {
+				perms.push('create')
+			}
+			if (hasPermissions(this.share.permissions, ATOMIC_PERMISSIONS.UPDATE)) {
+				perms.push('Update')
```
artonge, line 623: Only one to be capitalized ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+			} else {
+				this.$delete(this.share, 'newPassword')
+			}
+		},
+	},
+	beforeMount() {
+		this.initializePermissions()
+		this.initializeAttributes()
+		console.debug('shareSentIn', this.share)
+		console.debug('config', this.config)
```
artonge, line 664: To remove

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+				| (isCreateChecked ? ATOMIC_PERMISSIONS.CREATE : 0)
+				| (isDeleteChecked ? ATOMIC_PERMISSIONS.DELETE : 0)
+				| (isEditChecked ? ATOMIC_PERMISSIONS.UPDATE : 0)
+				| (isReshareChecked ? ATOMIC_PERMISSIONS.SHARE : 0)
+			this.share.permissions = permissions
+			this.share.hasDownloadPermission = isDownloadChecked
+			this.share.download = isDownloadChecked
+			this.queueUpdate('permissions', 'attributes')
+
+		},
```
artonge, line 688: Why is this method needed ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+				hasAdvancedAttributes = true
+			}
+
+			if (hasAdvancedAttributes) {
+				this.advancedSectionAccordionExpanded = true
+			}
+
+		},
+		initializePermissions() {
+			if (this.share.share_type) {
```
artonge, line 725: Why is there `share.share_type` and `share.type` ?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -0,0 +1,1062 @@
...
+				this.share.type = this.share.share_type
+			}
+			// shareType 0 (USER_SHARE) would evaluate to zero
+			// Hence the use of hasOwnProperty
+			if ('shareType' in this.share) {
+				this.share.type = this.share.shareType
```
artonge, line 731: And `share.shareType` ?

apps/files_sharing/src/views/SharingDetailsTab.vue
juliushaertl: I'd say it should be the level below what the sidebar header is using

artonge (changes requested):
Critical:

- [x] When updating the share, if I do not touch the password, it is still updated to something else.
- [x] The advances settings are not saved if changed on a new share
- [x] When switching from "Custom permissions" to "View only", I end up with "Allow upload and editing" selected
- [x] When settings "View only" with the quick settings, and then reload, the quick settings show "Custom permissions"
- [x] Setting the expiration date on a new user share does not work
- [x] Setting the expiration date saves the changes, when it should not
- [x] Setting the custom permissions saves the changes, when it should not

Follow-ups:

- [x] The list of permissions below "Custom permissions" should reflect the current permissions
- [ ] Switching between permissions presets should not close the advanced settings view
- [ ] The content of the sidebar is scrollable when it should only be the content of the share tab
Screenshot from 2023-08-29 11-52-15
- [x] A deleted share link is still visible if I do not reload the page

UI comments:

- [ ] Button at the bottom of the details view should not move when scrolling to the bottom
- [ ] Related resources should not appear in the details view
- [ ] "Delete share" button text seems smaller than other buttons
- [ ] It is confusing to have the download permission under "Custom permissions" as it is not taken into account for bundles

artonge (changes requested):
UI comments

- [ ] Button at the bottom of the details view should not move when scrolling to the bottom
- [ ] Related resources should not appear in the details view
- [ ] "Delete share" button text seems smaller than other buttons

apps/files_sharing/src/views/SharingDetailsTab.vue
jancborchardt: To clarify: In the subline of "Custom permissions", only the first word should be capitalized.

Fenn-CS: It's the case, the function that handles that corrects this, but fixed for consistency.

apps/files_sharing/src/views/SharingTab.vue
```diff
@@ -74,6 +76,15 @@
 				:name="fileInfo.name" />
 		</div>
 
+		<!-- share details -->
+		<div v-else>
+			<SharingDetailsTab :file-info="shareDetailsData.fileInfo"
```
artonge, line 81: Any reason to use `shareDetailsData.fileInfo` instead of `fileInfo` ?

artonge (approved):

AndyScherzinger (approved):

artonge:
Backport to stable27: https://github.com/nextcloud/server/pull/40166

PR #39474, opened 2023-07-19
fix(i18n) Fixed grammar

rakekniven:
Reported at Transifex


Pytal (commented):
🙈 https://github.com/nextcloud/server/blob/3da48c2adb606ce26cbd27b7d4ddde1bc9565f77/apps/files/src/components/FilesListVirtual.vue#L49 as well please :)

Pytal (approved):

susnux (approved):

PR #39476, opened 2023-07-19
Remove opacity on federation control action buttons

### LINKED ISSUE #36975: [BITV] 9.1.4.11/6.3 - All buttons "Change scope level of ..." do not meet the contrast requirements for graphical user interface components - actual contrast is 2.8:1. (1)
- [ ] Fix
- [ ] Backport

97384da12ad39b1f7335e094bda19bb2

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-3dbfb7615a-v6-n3

</details>

JuliaKirschenheuter:
* Resolves: https://github.com/nextcloud/server/issues/36975

## Summary

Remove opacity on federation control action buttons.

Before:

97384da12ad39b1f7335e094bda19bb2

After:
Peek 2023-07-19 14-35

## Checklist


szaimen:
@JuliaKirschenheuter from UX-perspective, can we maybe also add the current verbosity to the icon so that it is more clear that 1. this is a button and 2. which level is currently seleccted as the icon-only is not that verbose?

JuliaKirschenheuter:
Thank you @simison!
@nimishavijay, @marcoambrosini what would you say?

szaimen:
> Thank you @simison!

That is not me 😅

marcoambrosini:
@JuliaKirschenheuter visually the actions component would do the job
Screenshot 2023-07-20 at 14 58 55

Although from an accessibility standpoint it should actually be a `<select />`

This seems like a common usage pattern. Is it worth having a component that looks exactly the same but uses select and option elements? @skjnldsv @raimund-schluessler @nextcloud/vuejs 
Or maybe a prop that makes it into a select?

JuliaKirschenheuter:
> Although from an accessibility standpoint it should actually be a `<select />`

As far i can remember @nimishavijay said we can't take them because otherwise it would be "too much" and place is not enough for select. 

Nimisha, what would you say about:

Screenshot 2023-07-20 at 14 58 55

marcoambrosini:
How about moving to 1 column and having the multiselect inline instead of above the field? Basically a "table" with a privacy "column" on the right of each field

JuliaKirschenheuter:
@marcoambrosini, @szaimen 

Thank you for your suggestions for improvements! I would do it if not a time pressure which we have on accessibility topic. As discussed with @AndyScherzinger we should fix accessibility issues in a first line.
@marcoambrosini / @szaimen could you please open an feature-request improvements for 1 column layout and NcSelect instead of action buttons?

I would say this PR fixes initial a11y problem.

AndyScherzinger (approved):

szaimen (approved):
Fine by me to do the change then in a follow-up

JuliaKirschenheuter:
**/drone/src/tests/acceptance/features/header.feature:47** is not relevant
**/drone/src/tests/acceptance/features/login.feature:38** is not relevant

nimishavijay:
I would suggest to not include the labels for this PR, just remove the opacity on the buttons. My reasoning:
- The labels don't make much sense for people who are new (for eg. "Federated" would be shown by default, which is not super easy to understand)
- A label may also take up a lot of space, esp on mobile and for translated labels
- Using a secondary styling will also draw a lot of attention for those elements, which are not the main focus at all. The result will be something like the screenshot attached (couldn't get the labels to work :/ ). Even if we are using tertiary button styling the button labels will be bold while the field labels will be regular weight, drawing more attention to the button.

image

Since there is a follow up issue anyway, I'd say for this a11y problem we can just change the opacity :) In the future when we change the layout of the entire page we can make sure that we use the correct components and styling :)

JuliaKirschenheuter:
**/drone/src/tests/acceptance/features/header.feature:47** not relevant
    **/drone/src/tests/acceptance/features/login.feature:38** not relevant

PR #39485, opened 2023-07-19
fix(dispatcher): Move remaining simple cases in apps/ folder to IEven…

nickvergessen:
…tDispatcher

* One step closer for https://github.com/nextcloud/server/pull/38546

## Checklist

### DIFF
--- a/apps/files_sharing/lib/AppInfo/Application.php
+++ b/apps/files_sharing/lib/AppInfo/Application.php
@@ -51,7 +51,6 @@ use OCA\Files_Sharing\Notification\Listener;
 use OCA\Files_Sharing\Notification\Notifier;
 use OCA\Files\Event\LoadAdditionalScriptsEvent;
 use OCA\Files\Event\LoadSidebar;
-use OCP\Files\Event\BeforeDirectGetEvent;
 use OCA\Files_Sharing\ShareBackend\File;
 use OCA\Files_Sharing\ShareBackend\Folder;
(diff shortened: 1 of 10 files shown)


come-nc (approved):

apps/files_sharing/lib/AppInfo/Application.php
```diff
@@ -148,12 +143,18 @@ public function registerEventsScripts(IEventDispatcher $dispatcher, EventDispatc
 		});
 
 		// notifications api to accept incoming user shares
-		$oldDispatcher->addListener('OCP\Share::postShare', function (OldGenericEvent $event) {
+		$dispatcher->addListener('OCP\Share::postShare', function ($event) {
+			if (!$event instanceof OldGenericEvent) {
+				return;
+			}
```
marcelklehr, line 149: Why is this guard needed?

nickvergessen: It helps to trick psalm, so it doesn't yield on `$listener->shareNotification($event)` Later on, which only eats GenericEvent.
But it only is an intermediate step anyway to move forward to 5.4.x of the dispatcher. When we have to update to 6.4.x at the end of the year, GenericEvent will no longer exist and we need to do more work I guess, so I took the soft/low effort approach and added this mini guard.

marcelklehr (approved):

susnux (approved):
:broom:

nickvergessen:

PR #39487, opened 2023-07-19
Typed events for db:add-missing-…

nickvergessen:
* One step closer to #38546 

## Summary

* Instead of the last commit [fix(db)!: Remove private legacy event because we can not keep it](https://github.com/nextcloud/server/commit/d83bb944915d26872163154b68b749391ebc56c0) We could also add a "dispatchLegacy" to `IEventDispatcher` that allows to dispatch symfony's `GenericEvent`. However that one is also deprecated and will be dropped sooner or later. From my POV introducing yet another legacy layer is just too much (especially since I couldn't find any listeners), and dropping is the better option.

## Checklist


kesselb:
Unrelated, but interesting: If you put your instance into maintenance mode to add missing indexes, apps cannot register their indexes because the app and event listener are not loaded, and the index is not added.[^1]

[^1]: https://github.com/nextcloud/server/pull/37178

core/Command/Db/AddMissingIndices.php
```diff
@@ -91,7 +88,22 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 					$table = $schema->getTable($missingIndex['tableName']);
 					if (!$table->hasIndex($missingIndex['indexName'])) {
 						$output->writeln('<info>Adding additional ' . $missingIndex['indexName'] . ' index to the ' . $table->getName() . ' table, this can take some time...</info>');
-						$table->addIndex($missingIndex['columns'], $missingIndex['indexName']);
+
+						if ($missingIndex['dropUnnamedIndex'])
+						foreach ($table->getIndexes() as $index) {
```
come-nc, line 93: Misleading indentation here, see php-cs failure.

core/Application.php
```diff
@@ -88,244 +84,237 @@ public function __construct() {
...
-					$table = $schema->getTable('comments');
-
-					if (!$table->hasColumn('reference_id')) {
-						$subject->addHintForMissingColumn($table->getName(), 'reference_id');
-					}
-				}
-			}
-		);
+		$eventDispatcher->addListener(AddMissingIndicesEvent::class, function (AddMissingIndicesEvent $event) {
+			$event->addMissingIndex(
```
juliushaertl, line 88: Maybe move this out to a dedicated listener class for core?

core/Command/Db/AddMissingIndices.php
nickvergessen: Fixed

core/Application.php
nickvergessen: Can we do that in a follow up? Just trying to get closer to #38546  ^_^
cleaning up the core application is something for another day

juliushaertl: Sure

ChristophWurst: > for another day

Today?

ChristophWurst (approved):

miaulalala (approved):

bigcat88 (approved):

nickvergessen:

PR #39507, opened 2023-07-20
[backport26] Replace hardcoded arm64 and php8.1 dev

tsdicloud:
Fix hardcoded architecture problem nextcloud#98461 and use php8.1 Docker for development of version 26.


tsdicloud:
Sorry, wrong repo. Intended for our fork.

PR #39525, opened 2023-07-21
fix(CardDAV): set owner-principal, displayname properties for SAB

miaulalala:
Found in https://github.com/nextcloud/server/pull/38397

## Summary
The SAB has no owner principal even though it is a "shared" address book - the system user shares it will all users automatically.

## Checklist

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Principal.php
+++ b/apps/dav/lib/Connector/Sabre/Principal.php
@@ -208,6 +208,11 @@ class Principal implements BackendInterface {
 					'{DAV:}displayname' => $group->getDisplayName(),
 				];
 			}
+		} elseif ($prefix === 'principals/system') {
+			return [
+				'uri' => 'principals/system/' . $name,
+				'{DAV:}displayname' => $this->languageFactory->get('dav')->t("Accounts"),
+			];
 		}
 		return null;
 	}
--- a/apps/dav/lib/CardDAV/AddressBook.php
+++ b/apps/dav/lib/CardDAV/AddressBook.php
@@ -223,10 +223,7 @@ class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMov
 	}
 
 	public function propPatch(PropPatch $propPatch) {
-		// shared address books will be handled by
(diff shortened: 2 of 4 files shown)


ChristophWurst (approved):

apps/dav/lib/Connector/Sabre/Principal.php
```diff
@@ -208,6 +208,11 @@ public function getPrincipalByPath($path) {
 					'{DAV:}displayname' => $group->getDisplayName(),
 				];
 			}
+		} elseif ($prefix === 'principals/system') {
+			return [
+				'uri' => 'principals/system/' . $name,
+				'{DAV:}displayname' => $name,
```
tcitworld, line 214: Shouldn't the display name be translated?

JohannesGGE (approved):

apps/dav/lib/Connector/Sabre/Principal.php
```diff
@@ -208,6 +208,11 @@ public function getPrincipalByPath($path) {
 					'{DAV:}displayname' => $group->getDisplayName(),
 				];
 			}
+		} elseif ($prefix === 'principals/system') {
+			return [
+				'uri' => 'principals/system/' . $name,
+				'{DAV:}displayname' => $this->languageFactory->get('dav')->t($name),
```
tcitworld, line 214: **nitpick:** This won't work for string extraction. Hardcode `"System Addressbook"` instead.

tcitworld (changes requested):

apps/dav/lib/Connector/Sabre/Principal.php
miaulalala: Needs to be "Accounts" afaik

tcitworld:
```
1) OCA\DAV\Tests\unit\CardDAV\AddressBookTest::testPropPatchNotShared
Expectation failed for method name is "updateAddressBook" when invoked at least 1 times.
Expected invocation at least 1 times but it occurred 0 time(s).
```

tcitworld (approved):

PR #39529, opened 2023-07-21
No session creation on API calls

tsdicloud:
Nextcloud still creates a new session on each API call.
This will flood MagentaCLOUD Redis setup, esp. due to frequent API calls from Magenta HomeOS to store
observation cam movie files.

There is no unambiguous identification found for API calls yet, so that Nextcloud#28311 is disabled
to date (stable27).


tsdicloud:
Sorry, wrong repo to place the PR (stupid Github)

PR #39568, opened 2023-07-25
Calendar optimizations

icewind1991:
- preload shares for calendars when listing calendars
- cache calendar objects when doing `calendarQuery`

These save N queries when listing calendars and when listing calendar objects respectively.


tcitworld (approved):
This looks really nice.

kesselb (approved):
Good change :+1:

PR #34160, opened 2022-09-20
Detect weird local ips

come-nc:
Use new dependency to detect weird syntaxes

- [x] https://github.com/nextcloud/3rdparty/pull/1173
- [x] rebase

PR #39582, opened 2023-07-26
Simplify IP address normalizer with IP masks

Altahrim:
## Summary

- Try to simplify IP address normalizer with IP masks
- Removes dead code (subnet sizes were never used)
- Should be a little more efficient

## Checklist

### DIFF
--- a/lib/private/Security/Normalizer/IpAddress.php
+++ b/lib/private/Security/Normalizer/IpAddress.php
@@ -38,7 +38,7 @@ namespace OC\Security\Normalizer;
  */
 class IpAddress {
 	/**
-	 * @param string $ip IP to normalized
+	 * @param string $ip IP to normalize
 	 */
 	public function __construct(
 		private string $ip,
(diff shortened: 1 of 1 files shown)


icewind1991 (approved):

lib/private/Security/Normalizer/IpAddress.php
```diff
@@ -93,50 +74,30 @@ private function getEmbeddedIpv4(string $ipv6): ?string {
...
 	 * Gets either the /32 (IPv4) or the /64 (IPv6) subnet of an IP address
 	 */
 	public function getSubnet(): string {
-		if (\preg_match('/^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$/', $this->ip)) {
-			return $this->getIPv4Subnet(
-				$this->ip,
-				32
-			);
+		if (filter_var($this->ip, FILTER_VALIDATE_IP, FILTER_FLAG_IPV4)) {
+			return $this->ip.'/32';
```
come-nc, line 92: Is that really enough? What about weirdnesses like https://github.com/nextcloud/server/pull/34160/files#diff-91847f791fa239b37abbe8e19fdf756d6c6a405d0430df1f8dd6f5eac7bffd5bR124-R142 , is that not a concern here?

Altahrim: At this point, `filter_var` is supposed to do the job, but maybe we could add some tests with it :)

Altahrim: I tested some, they are not recognized as IP address by `filter_var` so it should be fine.
The `getIPv4Subnet` wasn't doing really more…

come-nc (approved):

PR #39588, opened 2023-07-27
fix(files): url used to retrive storage stats

max-nextcloud:
## Checklist

- No tests - sorry. I assume this will be rewriten as part of files to vue rework anyway.
- Documentation is not required
- Backports are not required as this is not a **critical** bugfix. At least i have not found any impact of the requests returning an html response.

### DIFF
--- a/apps/files/js/files.js
+++ b/apps/files/js/files.js
@@ -25,7 +25,7 @@
 				state.call.abort();
 			}
 			state.dir = currentDir;
-			state.call = $.getJSON(OC.generateUrl('apps/files/ajax/getstoragestats?dir={dir}', {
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

juliusknorr (approved):

juliushaertl:
@max-nextcloud backports?

max-nextcloud:
> @max-nextcloud backports?

I'd say no as this is not a **critical** bugfix. At least i have not found any impact of the requests returning an html response other than funny content in the network tab.

max-nextcloud:
Well... I guess stable27 might still make sense as that will be around for a while and I also don't see any downside to backporting this.

PR #39613, opened 2023-07-28
feat(files): migrate recent view

skjnldsv:
- [x] Using DAV search so we're fully compatible with all required actions metadata
- [x] Remove old files


apps/files/src/services/Recent.ts
```diff
@@ -0,0 +1,149 @@
...
+		: new Folder(nodeData)
+}
+
+export const getContents = async (path = '/'): Promise<ContentsWithRoot> => {
+	const contentsResponse = await client.getDirectoryContents(path, {
+		details: true,
+		// Only filter favorites if we're at the root
+		data: searchPayload,
+		headers: {
+			// Patched in WebdavClient.ts
```
susnux, line 129: I do not think we patched `SEARCH`?

apps/files/src/services/Recent.ts
```diff
@@ -0,0 +1,149 @@
...
+
+	return node.type === 'file'
+		? new File(nodeData)
+		: new Folder(nodeData)
+}
+
+export const getContents = async (path = '/'): Promise<ContentsWithRoot> => {
+	const contentsResponse = await client.getDirectoryContents(path, {
+		details: true,
+		// Only filter favorites if we're at the root
```
susnux, line 126:
```suggestion
```

We are not filtering favorites here

susnux (approved):
just two comments

szaimen:
I tested this and found a few things:
1. Pressing on any file downloads the file instead of opening in the viewer (but should be fixed by migrating viewer to the new api)
2. Clicking on a folder in the recent view does not redirect nor open the folder
3. when a favorited file is displayed at sorted by name, it is shown always on top - probably needs some design decision if it should be shown and sorted on top
4. when pressing on a folder, navigating back does not open the recent files again (should be fixed by getting rid of legacy view)

Pytal (approved):

PR #39638, opened 2023-07-31
Raise default SMTP timeout from 10 to 30 sec

solracsf:
## Summary

With some hosted STMP services, like `smtp.office365.com` they have some security mesures like bruteforce protection, and SMTP response can take more than (default) 10s between SSL handshake, auth, EHLO, etc.. In the logs this can lead to things like:

```
Error: stream_socket_enable_crypto(): SSL: Handshake timed out at /3rdparty/swiftmailer/swiftmailer/lib/classes/Swift/Transport/StreamBuffer.php#94

Swift_TransportException: Failed to authenticate on SMTP server with username "no-reply@nextcloud" using 1 possible authenticators.
Authenticator LOGIN returned Expected response code 250 but got code "235", with message "235 2.7.0 Authentication successful ".
```

This PR raises the default from 10 to 30 seconds.

_Other changes here are just small code optimizations._

## Checklist


skjnldsv (approved):

lib/private/Mail/Mailer.php
```diff
@@ -269,19 +256,23 @@ protected function getInstance(): MailerInterface {
...
+			$mailSmtpSecure,
 			null,
 			$this->logger
 		);
+
 		/** @var SocketStream $stream */
 		$stream = $transport->getStream();
 		/** @psalm-suppress InternalMethod */
-		$stream->setTimeout($this->config->getSystemValueInt('mail_smtptimeout', 10));
+		$stream->setTimeout($this->config->getSystemValueInt('mail_smtptimeout', 30));
```
ChristophWurst, line 275: the downside is that server that are not reachable will make the PHP process run longer than 30 seconds and cause a timeout

I'm on the fence here. we could also improve our documentation about increasing the timeout if there are servers where this is necessary.

come-nc: Yeah I would also vote against it, timeout can already be set in the configuration and 10 seconds looks like a sane default.

PR #39641, opened 2023-07-31
Fix OpenAPI specs

provokateurin:
## Summary

Some stuff I forgot to commit and small fixes for union types. A future CI will catch the former problems.

## Checklist


bigcat88 (approved):
Does it a good idea to have such large description with an example of the reply?

provokateurin:
Sorry, I don't understand what you mean. Can you rephrase for me?

bigcat88:
I am about the part with the _description: Output can look like this: ..._

provokateurin:
Ah I see, that description isn't on me though. It's just extracted from the comment of the method.

julien-nc (approved):
Is it a fix that will have to be done on each future generated OpenAPI specs?

provokateurin:
If the openapi-extractor is updated and the output is changed, yes. I will add it as a dev dependency and also setup CI. Then the tool will be upgraded by bumping the dependency and thus needing to regenerate the specs. I expect the updates to get less and less in the future as the output gets better and better.

Issue #39181, opened 2023-07-06
📣 Set reminder: file/folder


AndyScherzinger:
### Tasks

- [x] https://github.com/nextcloud/server/pull/39651
- [x] https://github.com/nextcloud/server/pull/39685
- [x] https://github.com/nextcloud/server/pull/39798

---

Add ability to set reminders for a file/folder resulting in a notification pointing to the item at the chosen time, see example GMail:

https://github.com/nextcloud/server/issues/39181#issuecomment-1641880618
![image](https://github.com/nextcloud/server/assets/925062/261004db-d029-42da-89f7-09837ab586f9)

For a first iteration of the feature we only implement the fixed options, no option to pick an individual time.

For scheduling a notification please check Deck / @juliushaertl since Deck can already send notification for Deck cards when you hit their due date.

The notification should simply be "Reminder for {file-name or folder-name}" with the placeholder being bold + a action button "View file" or "View folder" that brings you directly to the file/folder in the files view (like the "View chat" does for notifications when being mentioned in a Talk room giving you an easy way to navigate there.

jancborchardt:
@nimishavijay this is very related to "snooze mails", do you want to specify both together? :)
https://github.com/nextcloud/mail/issues/2048

jancborchardt:
> Like where would the reminder option be located

We should simply put it in the 3-dot menu in files. Between "Move or copy" and "Edit locally" would be a good position.

> which options would be offer?

The options you mention are good:

Set a reminder
- In 1 hour
- Tomorrow
- Next week
- Custom (possibly as a follow-up feature)

Then at the specified time you would get a Nextcloud notification. If it is a shared file, still only the person who set a reminder should get a notification.

cc @nimishavijay to coordinate for the options we would have in Mail and other apps for this. She also mentioned a possible "Reminder note" feature, but let’s keep it simple first only with the reminder time.

AndyScherzinger:
@Pytal if you have any design questions please get in touch with @jancborchardt or @nimishavijay. @jancborchardt and @nimishavijay I took the liberty to summarize things in the issue description and also spec how the notification should look like and behave.

jancborchardt:
@Pytal @nickvergessen for consistency let’s call it "Set reminder" in both cases. We don’t use "me" anywhere in the interface, and "Set reminder" is a simple common expression and call to action. :)

Pytal:
Any icon changes needed @jancborchardt @nimishavijay ?

[...]

PR #34807, opened 2022-10-25
Compute notification parsed subject from rich subject when possible

come-nc:
Let’s avoid duplicated code and translation strings and ease application developpers lives.

PR #39651, opened 2023-08-01
feat: File reminders backend

Pytal:
- For https://github.com/nextcloud/server/issues/39181

## Summary

- [x] HTTP API
- [x] Notifier
- [x] Scheduled notifications
- [x] Clean up job

## Checklist

- Screenshots before/after for front-end changes


apps/files_reminders/lib/Command/ListCommand.php
```diff
@@ -0,0 +1,93 @@
...
+			}
+		}
+
+		$reminders = $this->reminderService->getAll($user ?? null);
+		if (empty($reminders)) {
+			$io->text('No reminders');
+			return 0;
+		}
+
+		$io->table(
```
come-nc, line 77: If possible also support json output as other commands do. When inheriting base output json option is always advertised in `--help` so it might as well work.

apps/files_reminders/lib/Migration/Version10000Date20230725162149.php
```diff
@@ -0,0 +1,78 @@
...
+
+class Version10000Date20230725162149 extends SimpleMigrationStep {
+	/**
+	 * @param Closure $schemaClosure The `\Closure` returns a `ISchemaWrapper`
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->createTable(ReminderMapper::TABLE_NAME);
```
come-nc, line 44: It should return null if the table already exists

apps/files_reminders/lib/Notification/Notifier.php
```diff
@@ -0,0 +1,130 @@
...
+								'path' => $path,
+								'link' => $link,
+							],
+						],
+					)
+					->setParsedSubject(str_replace(
+						['{name}'],
+						[$node->getName()],
+						$subject,
+					))
```
come-nc, line 102:
```suggestion
```

Should not be needed, see https://github.com/nextcloud/server/pull/34807

come-nc (approved):
I guess another listener for if the user is deleted is needed as well, good otherwise.

apps/files_reminders/lib/Listener/NodeDeletedListener.php
```diff
@@ -0,0 +1,53 @@
+<?php
```
susnux, line 1: From review: UserDeleted listener should be added

susnux (approved):
:+1:

PR #39652, opened 2023-08-01
Fixes in update-3rdparty command

come-nc:
## Summary

Should fix at least the reactions so it properly reports that it failed…


come-nc:
> /update-3rdparty

Forgot it’s not possible to test like that, master version will always be used

skjnldsv (approved):

kesselb (approved):
:+1:

PR #39655, opened 2023-08-01
AppEcosystem Authentification

bigcat88:
* Adds: **new way of authentication** for external applications.

## Summary

It's a draft for tomorrow's AppEcosystem call.
This is all that is needed (minimum changes) to authenticate applications that will be written for the new Appecosystem.

It is desirable that before the next releases this would already be included, starting from the 26th version of the cloud. (AppEcosystem is developed only for 26+ versions)

## TODO

- [x] Nothing, we had tested it for two month, here will be no changes, it is enough for this.

## Checklist


juliushaertl:
As talked in the call, i think the cleanest approach would be be to have a AuthenticationProvider interface that apps can implement an register their authentication services then, however since there are plans to move this to OCP anyways it might be ok for now to:
- Add the unknown class as a stub for psalm
- Wrap the `Server::get` with a try/catch as DI might still fail
- Leave a hint for furture code readers about the reasoning

AndyScherzinger (approved):

julien-nc (approved):
:+1:

juliusknorr (approved):

PR #39680, opened 2023-08-02
New OCS endpoint to list text processing tasks

julien-nc:
refs https://github.com/nextcloud/text/issues/4613

This adds a `/tasks/app/{appId}` endpoint to get a list of tasks for the current user, for a specific app and optionally with a specific identifier.

While playing with the API, I discovered that it's impossible to schedule a task while not being authenticated (mostly because of a NotNull Db constraint on the user_id column). But all the endpoints are public. Maybe we should sort this out first.

* Do we want to allow non-authenticated users to schedule a job?
    * If so, do we want to be able to list all tasks that were scheduled publicly?
    * If not, should we require auth for all those endpoint?
* Are we fine with a user only being able to list his/her own tasks?
    * This would mean publicly scheduled tasks can be listed by any other non-authenticated user. Are we fine with that? 

TODO
* [x] add proper index(es)
* [x] allow authenticated users to list all their tasks by app (and optionally by identifier)

### DIFF
--- a/core/Controller/TextProcessingApiController.php
+++ b/core/Controller/TextProcessingApiController.php
@@ -157,4 +157,36 @@ class TextProcessingApiController extends \OCP\AppFramework\OCSController {
 			return new DataResponse(['message' => $this->l->t('Internal error')], Http::STATUS_INTERNAL_SERVER_ERROR);
 		}
 	}
+
+	/**
+	 * This endpoint returns a list of tasks related with a specific appId and identifier
+	 *
+	 * @PublicPage
+	 * @UserRateThrottle(limit=20, period=120)
+	 *
+	 * @param string $appId
+	 * @param string|null $identifier
+	 * @return DataResponse<Http::STATUS_OK, array{tasks: array{CoreTextProcessingTask}}, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 *  200: Task list returned
+	 */
+	public function listTasksByApp(string $appId, ?string $identifier = null): DataResponse {
+		if ($this->userId === null) {
+			return new DataResponse([
+				'tasks' => [],
+			]);
+		}
+		try {
+			$tasks = $this->languageModelManager->getTasksByApp($this->userId, $appId, $identifier);
+			$json = array_map(static function (Task $task) {
+				return $task->jsonSerialize();
+			}, $tasks);
+
+			return new DataResponse([
+				'tasks' => $json,
+			]);
+		} catch (\RuntimeException $e) {
+			return new DataResponse(['message' => $this->l->t('Internal error')], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
 }
--- a/core/Controller/TextProcessingApiController.php
+++ b/core/Controller/TextProcessingApiController.php
@@ -157,6 +157,36 @@ class TextProcessingApiController extends \OCP\AppFramework\OCSController {
 	}
 
 	/**
+	 * This endpoint allows to delete a scheduled task for a user
+	 *
+	 * @param int $id The id of the task
+	 *
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTextProcessingTask}, array{}>|DataResponse<Http::STATUS_NOT_FOUND|Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 * 200: Task returned
+	 * 404: Task not found
+	 */
+	#[NoAdminRequired]
+	public function deleteTask(int $id): DataResponse {
+		try {
+			$task = $this->textProcessingManager->getUserTask($id, $this->userId);
+
+			$this->textProcessingManager->deleteTask($task);
+
+			$json = $task->jsonSerialize();
+
+			return new DataResponse([
+				'task' => $json,
+			]);
+		} catch (NotFoundException $e) {
+			return new DataResponse(['message' => $this->l->t('Task not found')], Http::STATUS_NOT_FOUND);
+		} catch (\RuntimeException $e) {
+			return new DataResponse(['message' => $this->l->t('Internal error')], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
+
+	/**
 	 * This endpoint returns a list of tasks of a user that are related
 	 * with a specific appId and optionally with an identifier
 	 *
--- a/core/Controller/TextProcessingApiController.php
+++ b/core/Controller/TextProcessingApiController.php
@@ -29,6 +29,10 @@ namespace OC\Core\Controller;
 use InvalidArgumentException;
 use OCA\Core\ResponseDefinitions;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\AnonRateLimit;
+use OCP\AppFramework\Http\Attribute\NoAdminRequired;
+use OCP\AppFramework\Http\Attribute\PublicPage;
+use OCP\AppFramework\Http\Attribute\UserRateLimit;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\Common\Exception\NotFoundException;
 use OCP\IL10N;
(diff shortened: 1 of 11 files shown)


lib/private/TextProcessing/Db/TaskMapper.php
```diff
@@ -59,6 +59,25 @@ public function find(int $id): Task {
...
+	 * @throws Exception
+	 */
+	public function findByApp(string $userId, string $appId, ?string $identifier = null): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select(Task::$columns)
+			->from($this->tableName)
+			->where($qb->expr()->eq('app_id', $qb->createPositionalParameter($appId)))
+			->andWhere($qb->expr()->eq('user_id', $qb->createPositionalParameter($userId)));
+		if ($identifier !== null) {
+			$qb->andWhere($qb->expr()->eq('identifier', $qb->createPositionalParameter($identifier)));
```
juliushaertl, line 76: iirc this column wasn’t indexed yes, so would be good to have a combined app, user, identifier index

julien-nc: Yes indeed, it's not indexed.
The question would be which combination of columns makes the most sense. What will be the most common requested info? I can imagine that the assistant will list all tasks for a user with a specific state (for the result history or example).
One index by common use case should do it.

julien-nc:
What's been done:
* Convert annotations to Php method attributes
* Rename languageModelManager to textProcessingManager in the API controller
* Add endpoint for authenticated users to list their tasks by appId (and optionally by identifier)
* Add manager and mapper methods to:
    * get a task by user id and task id
    * get a list of task by user id and app id (and optionally with a specific identifier)
* Allow null user_id in `textprocessing_tasks` so tasks can be scheduled anonymously
* Add `user_id, app_id, identifier` index to `textprocessing_tasks`

@juliushaertl So it's only possible to list tasks by app while being authenticated. This is to prevent anonymous users to be able to list all tasks that were created by other anonymous users. Any problem with that in the Text integration?

juliushaertl:
> So it's only possible to list tasks by app while being authenticated. This is to prevent anonymous users to be able to list all tasks that were created by other anonymous users. Any problem with that in the Text integration? 

Fine with me as the MVP. I'd then limit the text integration to logged in users only for now.

lib/private/TextProcessing/Db/TaskMapper.php
```diff
@@ -59,6 +59,46 @@ public function find(int $id): Task {
...
+	 * @param string|null $identifier
+	 * @return array
+	 * @throws Exception
+	 */
+	public function findUserTasksByApp(string $userId, string $appId, ?string $identifier = null): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select(Task::$columns)
+			->from($this->tableName)
+			->where($qb->expr()->eq('app_id', $qb->createPositionalParameter($appId)))
+			->andWhere($qb->expr()->eq('user_id', $qb->createPositionalParameter($userId)));
```
juliushaertl, line 95:
```suggestion
			->where($qb->expr()->eq('user_id', $qb->createPositionalParameter($userId)))
			->andWhere($qb->expr()->eq('app_id', $qb->createPositionalParameter($appId)));
```

Some databases are picky on the order, so maybe just switch them to match the index `['user_id', 'app_id', 'identifier']`

core/Controller/TextProcessingApiController.php
```diff
@@ -157,4 +157,36 @@
...
+
+	/**
+	 * This endpoint returns a list of tasks related with a specific appId and identifier
+	 *
+	 * @PublicPage
+	 * @UserRateThrottle(limit=20, period=120)
+	 *
+	 * @param string $appId
+	 * @param string|null $identifier
+	 * @return DataResponse<Http::STATUS_OK, array{tasks: array{CoreTextProcessingTask}}, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
```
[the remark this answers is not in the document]
marcelklehr: Is this fixable?

lib/public/TextProcessing/IManager.php
```diff
@@ -80,4 +80,23 @@ public function scheduleTask(Task $task) : void;
...
+
+	/**
+	 * @param int $id The id of the task
+	 * @param string|null $userId The user id that scheduled the task
+	 * @return Task
+	 * @throws RuntimeException If the query failed
+	 * @throws NotFoundException If the task could not be found
+	 * @since 27.1.0
+	 */
+	public function getUserTask(int $id, ?string $userId): Task;
```
marcelklehr, line 92:
```suggestion
	public function getUserTask(int|float $id, ?string $userId): Task;
```

For 32bit support

lib/private/TextProcessing/Manager.php
```diff
@@ -192,7 +194,51 @@ public function getTask(int $id): OCPTask {
...
+	 * Get a task from its user id and task id
+	 * If userId is null, this can only get a task that was scheduled anonymously
+	 *
+	 * @param int $id The id of the task
+	 * @param string|null $userId The user id that scheduled the task
+	 * @return OCPTask
+	 * @throws RuntimeException If the query failed
+	 * @throws NotFoundException If the task could not be found
+	 */
+	public function getUserTask(int $id, ?string $userId): OCPTask {
```
marcelklehr, line 211:
```suggestion
	public function getUserTask(int|float $id, ?string $userId): OCPTask {
```

lib/public/TextProcessing/IManager.php
julien-nc: Should we then change it also in the getTask method? And what about the controller's getTask method?

marcelklehr: See https://github.com/nextcloud/server/pull/39660

julien-nc: Ok i'll wait for #39660 to be merged, then rebase and apply your suggested changes.
Even if it's not covered by the tests, the controller also needs this change IMO.

core/Controller/TextProcessingApiController.php
```diff
@@ -156,6 +156,36 @@ public function getTask(int $id): DataResponse {
...
+	 *
+	 * @param int $id The id of the task
+	 *
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTextProcessingTask}, array{}>|DataResponse<Http::STATUS_NOT_FOUND|Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 * 200: Task returned
+	 * 404: Task not found
+	 */
+	#[NoAdminRequired]
+	public function deleteTask(int $id): DataResponse {
```
marcelklehr, line 170: Tasks are automatically cleaned up after a week

lib/public/TextProcessing/IManager.php
kesselb: Sorry, I sent you on the wrong track. 

We use `int|float` sometimes in server for the files handling on 32bit, but not for http requests / database ids. 
I think it's fine to use int for the database ids.

core/Controller/TextProcessingApiController.php
juliushaertl: We use the new listing endpoint in text to show tasks for a document, so this would be used to keep that list at a sane length by either dismissing them or cleaning up once they were used.

juliushaertl: See https://github.com/nextcloud/text/issues/4613#issuecomment-1665704873 for the most recent mockups

julien-nc:
Rebased on master, bumped server version to 28.0.0.2 to trigger the migration step. All good IMO.

juliusknorr (approved):
Seems good, tested and works for my use case. 👍 

Psalm is still not happy though.

marcelklehr (approved):
:sparkles: 

please make sure to fix psalm, though

juliushaertl:
Failing drone acceptance test is unrelated, also there on master https://drone.nextcloud.com/nextcloud/server/38240/57/4

juliushaertl:
@julien-nc @marcelklehr Is there already documentation for the text processing to extend on?

juliushaertl:
Not triggering the backport yet, but I have it on my list of requirements for 27.1

marcelklehr:
@juliushaertl

Yes, text processing docs have already been merged, should be easy to extend for this :)

I've also created https://github.com/nextcloud/server/issues/39738 :)

PR #39689, opened 2023-08-03
fix(notifications): Remove share notification when the node is deleted

### LINKED ISSUE #39135: [Bug]: Notifications with shares are not cleared if share has been revoked

### Bug description

1. After user (sender) clicks `Unshare` on a file, which is shared between other users, they still have a notification persisted, what a file is shared.

2. After user (recepient) clicks `Accept` button - notification item disappears from notification tab.  Previously shared file couldn't appear, as it's not available anymore.

3. After user (recepient) clicks `Reject` button - it display error "Failed to execute action." Also it couldn't disappear from notification tab

### Steps to reproduce

1. Login as user "A" and do some file internal share to user "B"
2. Now delete that file from user "A" which has shared to user "B"
3. Open user "B" account and navigate to 'Notification' from hamburger menu
4. Click on accept button from notification tab and observe it from dashboard as to see received share file
5. Share another file from user "A" to user "B" as a internal share and after share delete this file
6. Login as "A", navigate to notification tab and click on 'reject' button form file which has shared from user "B"
7. Observe notification tab after clicking 'reject' button from shared file

(Description as per customer ticket)

### Expected behavior

Notification should throw appropriate alert when clicking reject or accept button option for the notification file which has already deleted from source (`Failed to execute action` at the moment)
source code for accepting / rejecting share should be [there](https://github.com/nextcloud/server/blob/d2efd0e03cb0f30b748b86c324fc574e603334aa/apps/files_sharing/lib/Controller/ShareAPIController.php#L1448-L1469) and [there](https://github.com/nextcloud/server/blob/master/apps/files_sharing/lib/Controller/ShareAPIController.php#L528)

AND

Notified share file should automatically get deleted when source user deleted the file. As it works when user clicks 'Unshare' of that share file, in this case notified notification share file should disappear from notification tab.
`files_sharing` app is in charge: source code for unsharing [is there](https://github.com/nextcloud/server/blob/2c71a77f4bd35bf68674be8b0fcff6f9f3c42200/apps/files_sharing/lib/Controller/RemoteController.php#L177)

### Installation method

Official All-in-One appliance

### Nextcloud Server version

25

nickvergessen:
* Resolves: #39135 

## Steps

1. Optional: Set up a mobile app for better testing
2. As UserA go to index.php/settings/user/sharing and disable the checkbox
grafik

3. As UserB create a folder and share it to UserA
4. As UserA reload the page or refresh the notifications on the mobile app, so you see the share notification
5. As UserB delete the the folder from 3.
6. As UserA repeat 4.

## Expected
No notification anymore

## Actual

Notification was still visible and "Decline" yields an error

## Checklist

### DIFF
--- a/apps/files_sharing/lib/Notification/Notifier.php
+++ b/apps/files_sharing/lib/Notification/Notifier.php
@@ -29,6 +29,7 @@ declare(strict_types=1);
 namespace OCA\Files_Sharing\Notification;
 
 use OCP\Files\IRootFolder;
+use OCP\Files\NotFoundException;
 use OCP\IGroupManager;
 use OCP\IL10N;
 use OCP\IURLGenerator;
@@ -114,6 +115,13 @@ class Notifier implements INotifier {
 		try {
 			$share = $this->shareManager->getShareById($attemptId, $notification->getUser());
 		} catch (ShareNotFound $e) {
+			throw new AlreadyProcessedException();
+		}
+
+		try {
+			$share->getNode();
+		} catch (NotFoundException $e) {
+			// Node is already deleted, so discard the notification
 			throw new AlreadyProcessedException();
 		}
 


come-nc (approved):
This means that deleting a file does not delete shares related to it? (since here the share exists but not the file)

come-nc:
> This means that deleting a file does not delete shares related to it? (since here the share exists but not the file)

Corollary: What will happen to this share which was never accepted and for which notification was removed, will it live on forever in database?

Antreesy (approved):
Tested on web client, now behavior is aligned with unsharing - notification dismisses after file is deleted

nickvergessen:
> This means that deleting a file does not delete shares related to it? (since here the share exists but not the file)

There is a 15min background job: 
https://github.com/nextcloud/server/blob/a0f6a6545b59b1061e96e7d17d8ec6d6f32e615e/apps/files_sharing/lib/DeleteOrphanedSharesJob.php#L33

And a custom command:
https://github.com/nextcloud/server/blob/50501f8515361577e74c9e639853fc8e68178c2f/apps/files_sharing/lib/Command/DeleteOrphanShares.php#L35

But the 15min could be enough time to cause confusion.

> What will happen to this share which was never accepted and for which notification was removed, will it live on forever in database?

Yes, the "pending shares" view still allows accepting it

come-nc:
> > This means that deleting a file does not delete shares related to it? (since here the share exists but not the file)
> 
> There is a 15min background job

Why is that not done through a listener of the deletion event instead? (or on top of)

nickvergessen:
> Why is that not done through a listener of the deletion event instead? (or on top of)

We'd need to loop over all subfolders and files, potentially thousands to millions

PR #39694, opened 2023-08-03
expose gps data via webdav

tobiasKaminsky:
## Checklist

- [ ] https://github.com/nextcloud/documentation/pull/10957

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesPlugin.php
@@ -84,6 +84,7 @@ class FilesPlugin extends ServerPlugin {
 	public const SUBFOLDER_COUNT_PROPERTYNAME = '{http://nextcloud.org/ns}contained-folder-count';
 	public const SUBFILE_COUNT_PROPERTYNAME = '{http://nextcloud.org/ns}contained-file-count';
 	public const FILE_METADATA_SIZE = '{http://nextcloud.org/ns}file-metadata-size';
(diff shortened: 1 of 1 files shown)


apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -438,6 +439,27 @@ public function handleGetProperties(PropFind $propFind, \Sabre\DAV\INode $node)
 
 					return $sizeMetadata->getValue();
 				});
+
+				$propFind->handle(self::FILE_METADATA_GPS, function () use ($node) {
+					if (!str_starts_with($node->getFileInfo()->getMimetype(), 'image')) {
+						return json_encode((object)[], JSON_THROW_ON_ERROR);
```
Altahrim, line 445: Maybe just:
```suggestion
						return '{}';
```

Altahrim (approved):

susnux (approved):
Do not forget to add the prop to the documentation here:

https://github.com/nextcloud/documentation/blob/master/developer_manual/client_apis/WebDAV/basic.rst#supported-properties

apps/dav/lib/Connector/Sabre/FilesPlugin.php
AndyScherzinger: I guess it is a one-to-one replication of the code for `files-metadata-size`, so I would keep it as-is.

icewind1991:
Added the pre-loading and de-duplicated some logic

Altahrim (approved):

come-nc (approved):

PR #39695, opened 2023-08-03
Refactors files version app commands.

fsamapoor:
## Summary
I have made some adjustments to the apps/files_version/lib/Command classes to improve the code readability.

- Adding return types
- Replacing return code integers with more readable strings.
- Using early returns

## Checklist

### DIFF
--- a/apps/files_versions/lib/Command/CleanUp.php
+++ b/apps/files_versions/lib/Command/CleanUp.php
@@ -115,7 +115,7 @@ class CleanUp extends Command {
 	/**
 	 * delete versions for the given user
 	 */
-	protected function deleteVersions(string $user, string $path = null): void {
+	protected function deleteVersions(string $user, ?string $path = null): void {
 		\OC_Util::tearDownFS();
 		\OC_Util::setupFS($user);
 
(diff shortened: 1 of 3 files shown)


artonge (approved):

apps/files_versions/lib/Command/CleanUp.php
```diff
@@ -85,46 +75,45 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
 	}
 
 
 	/**
 	 * delete versions for the given user
-	 *
-	 * @param string      $user
-	 * @param string|null $path
 	 */
 	protected function deleteVersions(string $user, string $path = null): void {
```
come-nc, line 118:
```suggestion
	protected function deleteVersions(string $user, ?string $path = null): void {
```
Better to be explicit

apps/files_versions/lib/Command/Expire.php
```diff
@@ -62,7 +53,7 @@ public function handle() {
 			// In case of external storage and session credentials, the expiration
 			// fails because the command does not have those credentials
 
-			/** @var LoggerInterface */
+			/** @var LoggerInterface $logger */
```
come-nc, line 56:
```suggestion
```

come-nc (commented):
:+1:

apps/files_versions/lib/Command/CleanUp.php
fsamapoor: Sure.

come-nc (approved):

PR #26270, opened 2021-03-24
Catch invalid cache source storage path

skjnldsv:
`OC\Files\Cache\Cache::get` can also return false

Issue #19009, opened 2020-01-20
CacheQueryBuilder::whereFileId() must be of the type int, null given


acsfer:
- NC 19.0.4
- nGinx 1.19
- PHP 7.4.11
- APCU as Local cache
- Redis 5 as Distributed Memcache
- S3 Object Storage as primary
- MariaDB 10.4

Reproduce case here: https://github.com/nextcloud/server/issues/19009#issuecomment-621916168

```
[remote] Error: TypeError: Argument 1 passed to OC\Files\Cache\CacheQueryBuilder::whereFileId() must be of the type int, null given, called in /srv/lib/private/Files/Cache/Cache.php on line 655 at <<closure>>

 0. /srv/lib/private/Files/Cache/Cache.php line 655
    OC\Files\Cache\CacheQueryBuilder->whereFileId(null)
 1. /srv/lib/private/Files/Cache/Cache.php line 579
    OC\Files\Cache\Cache->moveFromCache(OC\Files\Cache\Cache {}, "files/TEST A/DSC_0073.jpg.part", "files/TEST A/DSC_0073.jpg")
 2. /srv/lib/private/Files/ObjectStore/ObjectStoreStorage.php line 481
    OC\Files\Cache\Cache->move("files/TEST A/DSC_0073.jpg.part", "files/TEST A/DSC_0073.jpg")
 3. /srv/lib/private/Files/Storage/Wrapper/Wrapper.php line 630
    OC\Files\ObjectStore\ObjectStoreStorage->writeStream("files/TEST A/DSC_0073.jpg", null, 10987936)
 4. /srv/lib/private/Files/Storage/Wrapper/Wrapper.php line 630
    OC\Files\Storage\Wrapper\Wrapper->writeStream("files/TEST A/DSC_0073.jpg", null, null)
 5. /srv/apps/dav/lib/Connector/Sabre/File.php line 192
    OC\Files\Storage\Wrapper\Wrapper->writeStream("files/TEST A/DSC_0073.jpg", null)
 6. /srv/apps/dav/lib/Connector/Sabre/Directory.php line 156
    OCA\DAV\Connector\Sabre\File->put(null)
 7. /srv/3rdparty/sabre/dav/lib/DAV/Tree.php line 316
    OCA\DAV\Connector\Sabre\Directory->createFile("DSC_0073.jpg", null)
 8. /srv/3rdparty/sabre/dav/lib/DAV/Tree.php line 130
[...]

PR #39698, opened 2023-08-03
Catch more invalid cache source storage paths

joshtrichards:
* Related: nextcloud/server#19009

## Summary

`OC\Files\Cache\Cache::get` can return string|false|null, not just string|false.

Since null is just as much of an indicator that something is wrong with the specified cache source storage path, handle it same way as false.

- nextcloud/server#26270 added handling of false, but null is needed too
- Well, or we change the default $resullt to false instead of null, but I'm not sure if that has other ramifications

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -668,7 +668,7 @@ class Cache implements ICache {
 			$targetPath = $this->normalize($targetPath);
 
 			$sourceData = $sourceCache->get($sourcePath);
-			if ($sourceData === false || $sourceData === null) {
+			if (!$sourceData) {
 				throw new \Exception('Invalid source storage path: ' . $sourcePath);
 			}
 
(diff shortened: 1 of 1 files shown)


lib/private/Files/Cache/Cache.php
```diff
@@ -668,7 +668,7 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
 			$targetPath = $this->normalize($targetPath);
 
 			$sourceData = $sourceCache->get($sourcePath);
-			if ($sourceData === false) {
+			if ($sourceData === false || $sourceData === null) {
```
solracsf, line 671: Simplification:
```suggestion
			if (!$sourceData) {
```

solracsf (approved):
Appart my comment above, makes sense.

skjnldsv (approved):

PR #39699, opened 2023-08-03
Hide shares by disabled users

### LINKED ISSUE #31444: Disable shares for deactivated users
### How to use GitHub

Implemented in #17077, share links are disabled for  deactivated users. In my opinion, this should also happen with normal shares to other users. At the moment, a shared folder of a disabled user is still accessable by others.

If a user is deactivated, shares from this user should no longer be visible / accesable for others.

I didn't found an easy way of deactivate shares of a user. My (ugly) solution is to backing up oc_share of this user in a table with same schema  `insert into my_share_backup select * from oc_share where uid_owner='userid';` and setting permissions in oc_share to 0 `update oc_share set permissions=0 where uid_owner='userid'`, restore data on re-activation of user.

**Additional context**
We need this for 2 reasons:
First, if deactivating someone because auf violation of TOS, noone should be able to change or access the users data. 
Second, if accounts expire, they are deactivated but not removed for a grace period. During this time also the data should not be accessable through other accounts.

(Using NC 22.2.5)

come-nc:
* Resolves: #31444

## Summary

1. Use `occ config:app:set files_sharing hide_disabled_user_shares --value yes` to enable
2. Once enabled shares by disabled user will appear as non-existing
3. Shares of files owned by disabled users will disappear as well, even if shared by someone else
4. Compatible with advanced option from user_ldap to disable users missing from LDAP

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required https://github.com/nextcloud/documentation/pull/11018

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -1559,7 +1559,7 @@ class Manager implements IManager {
 			$uids = array_unique([$share->getShareOwner(),$share->getSharedBy()]);
 			foreach ($uids as $uid) {
 				$user = $this->userManager->get($uid);
-				if (($user !== null) && !$user->isEnabled()) {
+				if ($user?->isEnabled() === false) {
 					throw new ShareNotFound($this->l->t('The requested share comes from a disabled user'));
 				}
 			}
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -1343,7 +1343,7 @@ class Manager implements IManager {
 			$added = 0;
 			foreach ($shares as $share) {
 				try {
-					$this->checkExpireDate($share);
+					$this->checkShare($share);
 				} catch (ShareNotFound $e) {
 					//Ignore since this basically means the share is deleted
 					continue;
@@ -1402,7 +1402,7 @@ class Manager implements IManager {
 		// remove all shares which are already expired
 		foreach ($shares as $key => $share) {
 			try {
-				$this->checkExpireDate($share);
+				$this->checkShare($share);
 			} catch (ShareNotFound $e) {
 				unset($shares[$key]);
 			}
@@ -1448,7 +1448,7 @@ class Manager implements IManager {
 
 		$share = $provider->getShareById($id, $recipient);
 
-		$this->checkExpireDate($share);
+		$this->checkShare($share);
 
 		return $share;
 	}
@@ -1532,7 +1532,7 @@ class Manager implements IManager {
 			throw new ShareNotFound($this->l->t('The requested share does not exist anymore'));
 		}
 
-		$this->checkExpireDate($share);
+		$this->checkShare($share);
 
 		/*
 		 * Reduce the permissions for link or email shares if public upload is not enabled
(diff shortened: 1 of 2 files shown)


lib/private/Share20/Manager.php
```diff
@@ -1554,11 +1554,25 @@ public function getShareByToken($token) {
...
 		if ($share->isExpired()) {
 			$this->deleteShare($share);
 			throw new ShareNotFound($this->l->t('The requested share does not exist anymore'));
 		}
+		if ($this->config->getAppValue('files_sharing', 'hide_disabled_user_shares', 'no') === 'yes') {
+			$uids = array_unique([$share->getShareOwner(),$share->getSharedBy()]);
+			foreach ($uids as $uid) {
+				$user = $this->userManager->get($uid);
+				if (($user !== null) && !$user->isEnabled()) {
+					throw new ShareNotFound($this->l->t('The requested share comes from a disabled user'));
```
Altahrim, line 1572: I wonder if we should add a kind of `ShareDenied` instead of `NotFound`. 
I think it will be closer to the real state :)

lib/private/Share20/Manager.php
```diff
@@ -1554,11 +1554,25 @@ public function getShareByToken($token) {
...
+	protected function checkShare(IShare $share): void {
 		if ($share->isExpired()) {
 			$this->deleteShare($share);
 			throw new ShareNotFound($this->l->t('The requested share does not exist anymore'));
 		}
+		if ($this->config->getAppValue('files_sharing', 'hide_disabled_user_shares', 'no') === 'yes') {
+			$uids = array_unique([$share->getShareOwner(),$share->getSharedBy()]);
+			foreach ($uids as $uid) {
+				$user = $this->userManager->get($uid);
+				if (($user !== null) && !$user->isEnabled()) {
```
Altahrim, line 1571: We can check all at once: 
```suggestion
				if ($user?->isEnabled() === false) {
```

lib/private/Share20/Manager.php
come-nc: @AndyScherzinger @icewind1991 What do you think?
Using a new Exception class will mean we have to adapt all places currently catching `ShareNotFound`, which seems to be a lot.
But it would be more appropriate, for expired shares `ShareNotFound` but the share is then deleted so it makes sense.
But it would open a whole lot of questions for each code path on wether we should actually treat those new Exception the same as `ShareNotFound` or not.

AndyScherzinger: Looping in @nickvergessen - from my pov it is fine to say "not found" to not disclose the info that the share exists. On the other hand one could argue that this is excactly what is the goal: To tell people that the share is denied (because the persons account got disabled and with it their shares).

Altahrim: We could also change the name of current `ShareNotFound` exception to show we don't know why the access is impossible (denied/not found/…).
`UnreachableShare`? I don't come up with a better name…

If we don't need to have a difference between "not found" and "denied", it will be easier to do.

Altahrim (approved):
My comment is not a blocker and will probably be addressed in another PR

lib/private/Share20/Manager.php
susnux: Or inherit from `ShareNotFound` so it is still catched?

nickvergessen: Since this is intentional, I'd say ShareNotFound is fine. If a user got disabled and there stuff should not be accessible anymore it should not be distinguishable

come-nc: Good to merge then?

tests/lib/Share20/ManagerTest.php
```diff
@@ -2714,10 +2714,12 @@ public function testGetSharesByExpiredLinkShares() {
 
 	public function testGetShareByToken() {
 		$this->config
-			->expects($this->once())
+			->expects($this->exactly(2))
```
icewind1991, line 2717: I prefer just removing these counts altogether as they don't actually contribute to the test.

AndyScherzinger (approved):

AndyScherzinger:
@come-nc I think having it in 27.1 would be nice to have some (more) upgrade motivation for the requesting party, no?

come-nc:
> What is the reasoning for not doing this by default?

Avoid a behavior change on update which will not please everyone.
Most people want to disable users to prevent them from logging and do not want to lose access to their shared data.

come-nc:
> @come-nc I think having it in 27.1 would be nice to have some (more) upgrade motivation for the requesting party, no?

Not my call, if you think it fits and 27.1 is still open we can backport.

AndyScherzinger:
> 27.1 is still open we can backport.

yes, still open and fine to backport 👍

PR #39702, opened 2023-08-03
Feat/26668/notifications for shared calendars 2

### LINKED ISSUE #26668: Send invitations for shared calendars

kesselb:
* Resolves: https://github.com/nextcloud/server/issues/26668

## Summary

Fun fact:  The default implementation for Sabre\CalDAV\Calendar.getOwner[^1] would always return the principaluri as owner. It's our overwrite in OCA\DAV\CalDAV\Calendar.getOwner[^2] causing the issues. 

[^1]: https://github.com/sabre-io/dav/blob/85b33f7c4b597bdb2a90e6886a48c5723e767062/lib/CalDAV/Calendar.php#L229-L240

[^2]: https://github.com/nextcloud/server/blob/780ece032fd42ec099d888a46a5349d8dea48dc2/apps/dav/lib/CalDAV/Calendar.php#L228-L232

## Issues

1. Calendar shared editable from Alice to Bob. Bob adds an event and invites Alice. An iTip message is generated for Alice and will update the event in the shared calendar.  If Alice adds an event and invites Bob, an additional event in Bob's personal calendar is created. The event in Bob's personal calendar is only created if Bob is already an attendee in the initial ics file, then the event in Bob's calendar is created before the event in Alice's calendar.

## TODO

- [ ] Sabre\CalDAV\Schedule\Plugin.beforeUnbind is related: https://github.com/nextcloud/3rdparty/blob/7293b4ffe86c0b48c3172bd8f180a12e0907fd60/sabre/dav/lib/CalDAV/Schedule/Plugin.php#L388-L393.
- [ ] Calendar Trash: Does not support shared calendars. Only the owner can restore objects from shared calendars.
- [ ] Tests

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/CalDavBackend.php
+++ b/apps/dav/lib/CalDAV/CalDavBackend.php
@@ -2244,21 +2244,58 @@ class CalDavBackend extends AbstractBackend implements SyncSupport, Subscription
 	 * @return string|null
 	 */
 	public function getCalendarObjectByUID($principalUri, $uid) {
+		// query for shared writable calendars
+		$principals = array_merge(
+			[$principalUri],
+			$this->principalBackend->getGroupMembership($principalUri, true),
+			$this->principalBackend->getCircleMembership($principalUri)
+		);
+
 		$query = $this->db->getQueryBuilder();
-		$query->selectAlias('c.uri', 'calendaruri')->selectAlias('co.uri', 'objecturi')
+		$query
+			->selectAlias('c.id', 'calendarid')
+			->selectAlias('c.principaluri', 'principaluri')
+			->selectAlias('c.uri', 'calendaruri')
+			->selectAlias('co.uri', 'objecturi')
+			->selectAlias('ds.access', 'access')
+			->selectAlias('ds.principaluri', 'shareprincipal')
 			->from('calendarobjects', 'co')
 			->leftJoin('co', 'calendars', 'c', $query->expr()->eq('co.calendarid', 'c.id'))
-			->where($query->expr()->eq('c.principaluri', $query->createNamedParameter($principalUri)))
-			->andWhere($query->expr()->eq('co.uid', $query->createNamedParameter($uid)))
-			->andWhere($query->expr()->isNull('co.deleted_at'));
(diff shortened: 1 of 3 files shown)


blizzz:

PR #39707, opened 2023-08-03
Allow ext storage Local to go unavailable

### LINKED ISSUE #39706: Detect external storage of type Local as unavailable when path is unmounted on system
### Steps

1. Create a local folder "/mnt/mountedstorage/nextcloud_external"
2. Give the web server permissions for that folder
3. Mount it as Local external storage under the local path "/nextcloud_external"
4. In the web UI, create a folder "/nextcloud_external/sub"
5. On the CLI, rename "nextcloud_external" to something else to simulate as if "/mnt/mountedstorage" was not mounted
6. PROPFIND "/nextcloud_external"

### Expected result
PROPFIND returns a 503 with StorageNotAvailable since the storage/path is temporarily not there.
A configured desktop client would not touch that folder locally, no deletion.

### Actual result
PROPFIND returns a 404 on the path.
A configured desktop client would proceed to delete the local "/nextcloud_external" folder until mounted again. (assuming this, I haven't tested)
The next time the mount point is back, the desktop client would redownload everything.

### Versions
Observed on master 9a7e2b15580578b4de3eb36808dc466a9fd6b976 and it was like that since forever :sweat_smile: 

### Notes
My actual use case is that I have an "archive" hard disk connected to my Rpi4 but I don't want to have it powered up every time.
My expectation is that whenever I power it up the mount point becomes available again to access the archive, but I don't want desktop clients to delete and redownload files again from that place.

From what I remember desktop clients are already able to deal with unavailable storages and will skip these seamlessly (503 StorageNotAvailable). @tobiasKaminsky 

Thoughts @icewind1991 @artonge @come-nc ?

I have a PR incoming...

PVince81:
* Resolves: #39706 

## Summary

Whenever an external storage of type Local points at a non-existing directory, process this as a StorageNotAvailable instead of returning 404.

This makes desktop clients ignore the folder instead of deleting it when it becomes unavailable.

The code change was limited to external storages to avoid issues during setup and with the default home storage.

## TODO

- [x] actually test with desktop client to confirm behavior: https://github.com/nextcloud/server/pull/39707#issuecomment-1664663872
- [ ] properly convert the exception so that it appears in Webdav response

## Checklist

- [x] Screenshots before/after for front-end changes -> N/A

### DIFF
--- a/apps/files_external/lib/Lib/Backend/Local.php
+++ b/apps/files_external/lib/Lib/Backend/Local.php
@@ -26,8 +26,10 @@ namespace OCA\Files_External\Lib\Backend;
 use OCA\Files_External\Lib\Auth\AuthMechanism;
 use OCA\Files_External\Lib\Auth\NullMechanism;
 use OCA\Files_External\Lib\DefinitionParameter;
+use OCA\Files_External\Lib\StorageConfig;
 use OCA\Files_External\Service\BackendService;
 use OCP\IL10N;
+use OCP\IUser;
 
 class Local extends Backend {
 	public function __construct(IL10N $l, NullMechanism $legacyAuth) {
@@ -44,5 +46,9 @@ class Local extends Backend {
 			->addAuthScheme(AuthMechanism::SCHEME_NULL)
 			->setLegacyAuthMechanism($legacyAuth)
 		;
+	}
+
+	public function manipulateStorageConfig(StorageConfig &$storage, IUser $user = null): void {
+		$storage->setBackendOption('isExternal', true);
 	}
 }
--- a/lib/private/Files/Storage/Local.php
+++ b/lib/private/Files/Storage/Local.php
@@ -51,6 +51,7 @@ use OCP\Files\ForbiddenException;
 use OCP\Files\GenericFileException;
 use OCP\Files\IMimeTypeDetector;
 use OCP\Files\Storage\IStorage;
+use OCP\Files\StorageNotAvailableException;
 use OCP\IConfig;
 use OCP\Util;
 use Psr\Log\LoggerInterface;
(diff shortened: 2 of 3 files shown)


PVince81:
I've tested with desktop client 3.9.0 and it did not delete the unavailable folder.
After restoring the folder, syncing continued.
Also, while the folder was unavailable, syncing of other folders worked fine.

**However** the desktop client complained repeatedly about the folder causing an error.
Either the response needs to be tweaked or the client needs adjustment.
The PROPFIND response when unavailable looks like this:
```xml
  <s:exception>Sabre\DAV\Exception\ServiceUnavailable</s:exception>
  <s:message>Storage is temporarily not available</s:message>
</d:error>
```

I'm guessing something is missing to have the exception name be properly propagated into the XML as StorageNotAvailableException.

- [ ] find out how to properly convert the exception

PVince81:
from what I see, other storages tend to throw that exception in regular methods, not in the constructor.
I purposefully chose to throw earlier as I didn't want to do the additional `stat('')` for any kind of exception happening in the methods.

Currently throwing the exception will trigger a `FailedStorage` in [constructStorage](https://github.com/nextcloud/server/blob/bugfix/39706/local-ext-storage-unavailable-mode/apps/files_external/lib/Config/ConfigAdapter.php#L128).
I'm guessing that the Webdav layer doesn't know how to deal with FailedStorage yet.

PVince81:
Found it, it seems that StorageNotAvailableException is converted to ServiceNotAvailableException... why not preserve it or convert to Webdav ? https://github.com/nextcloud/server/blob/bugfix/39706/local-ext-storage-unavailable-mode/apps/dav/lib/Connector/Sabre/ObjectTree.php#L161


I think I remembered it wrong and this distinction was never implemented on the client side, or a different approach was used ? Searching for StorageNotAvailableException in the desktop client code did not show results. @tobiasKaminsky 

Or is the distinction done whether the 503 is returned on the root or on a subdir ?

In any case it would be good to have errors be less flashy when a storage is not mounted, but still let the user know somehow, maybe only once the first time ?
image

max-nextcloud:
I'm currently investigating a scenario where metadata (tags, shares, etc) were lost after the main data directory which was on an NFS went offline.

Would such a check also make sense for the data directory itself?

icewind1991 (approved):

come-nc (approved):

szaimen:
@PVince81 is it fine when I merge this or do you want to address the missing point first? Would also be okay to do this in  follow-up imho :)

PR #39734, opened 2023-08-06
ci: move phpunit-nodb to actions

kesselb:
#### Learning 1 

`phpunit -c autotest-external` only runs tests for enable apps. 

This seems to make sense, but it is not obvious and does not make the CI configuration any easier.

The following apps are part of the server repository and are not enabled by default. 
Therefore, they must be enabled for CI runs.

```
admin_audit
encryption
files_external
user_ldap
```

#### Learning 2

`phpunit --list-tests` lists the available tests. 

Unfortunately, it does not list tests from dynamically enabled apps. 
I assume that's related to the way the tests are added. 

## TODO

- [x] CI
- [x] Decide: Run nodb tests for PHP 8.0, 8.1 and 8.2 or only PHP 8.0 (like on drone)
- [ ] Review
- [ ] Follow-up to remove nodb from drone
- [x] Follow-up to exclude PRIMARY-azure,PRIMARY-s3,PRIMARY-swift 
- [x] Follow-up to add php 8.3 (after https://github.com/nextcloud/server/pull/40630)
- [ ] Ask Côme if the additional apps should be enabled for phpunit-32bits too

## Migration

**Round 1:**

 - No changes to test suite

| | Actions | Drone |
|--------|--------|--------|
| Tests executed | 10554 | 10861 | 
| Warnings | 28 | 25 | 
| Skipped | 76 | 96 | 
| PHPUnit | 9.6.10 | 9.5.28 |

Warnings on Actions but not on Drone:

```
26) OCA\Theming\Tests\IconBuilderTest::testGetFaviconNotFound
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

27) OCA\Theming\Tests\IconBuilderTest::testGetTouchIconNotFound
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.

28) OCA\Theming\Tests\IconBuilderTest::testColorSvgNotFound
Expecting E_WARNING and E_USER_WARNING is deprecated and will no longer be possible in PHPUnit 10.
```
expectWarning is deprecated since PHPUnit 9.6

**Round 2:**

- Exclude PRIMARY-azure,PRIMARY-s3,PRIMARY-swift (they are skipped anyway because the configuration is missing)
- Added redis extension

| | Actions | Drone |
|--------|--------|--------|
| Tests executed | 10524 | 10861 | 
| Warnings | 28 | 25 | 
| Skipped | 46 | 96 | 
| PHPUnit | 9.6.10 | 9.5.28 |

**Round 3:**

- Added ldap extension
- Print list of tests to be executed

| | Actions | Drone |
|--------|--------|--------|
| Tests executed | 10524  | 10861 | 
| Warnings | 28 | 25 | 
| Skipped | 46 | 96 | 
| PHPUnit | 9.6.10 | 9.5.28 |

**Round 4:**

- Enable additional apps: `admin_audit encryption federatedfilesharing federation files_sharing files_trashbin files_versions provisioning_api user_ldap` (cf. https://github.com/nextcloud/server/blob/master/tests/enable_all.php)

| | Actions | Drone |
|--------|--------|--------|
| Tests executed | 10831 | 10861 | 
| Assertions | 23081 | 23020 | 
| Warnings | 28 | 25 | 
| Skipped | 46 | 96 | 
| PHPUnit | 9.6.10 | 9.5.28 |

Actions: https://github.com/nextcloud/server/actions/runs/5777400665/job/15657527803
Drone: https://drone.nextcloud.com/nextcloud/server/38222/9/4

**Round 5:**

- Include test suite PRIMARY-azure,PRIMARY-s3,PRIMARY-swift again because this cleanup should be done as follow up.

| | Actions | Drone |
|--------|--------|--------|
| Tests executed | 10861 | 10861 | 
| Assertions | 23081 | 23020 | 
| Warnings | 28 | 25 | 
| Skipped | 76 | 96 | 
| PHPUnit | 9.6.10 | 9.5.28 |

Actions: https://github.com/nextcloud/server/actions/runs/5777508182/job/15657761644
Drone: https://drone.nextcloud.com/nextcloud/server/38223/9/4

**Round 6:**

- Squashed the commits
- Removed the --list-tests step 

| | Actions | Drone |
|--------|--------|--------|
| Tests executed | 10861 | 10861 | 
| Assertions | 23081 | 23020 | 
| Warnings | 28 | 25 | 
| Skipped | 76 | 96 | 
| PHPUnit | 9.6.10 | 9.5.28 |

Actions: https://github.com/nextcloud/server/actions/runs/5777613755/job/15657986497
Drone: https://drone.nextcloud.com/nextcloud/server/38224/9/4

## Checklist

### DIFF
--- a/.github/workflows/nodb.yml
+++ b/.github/workflows/nodb.yml
@@ -63,7 +63,6 @@ jobs:
         with:
           php-version: ${{ matrix.php-versions }}
           extensions: apcu, ctype, curl, dom, fileinfo, gd, imagick, intl, json, ldap, mbstring, openssl, pcntl, pdo_sqlite, posix, redis, sqlite, xml, zip
-          tools: phpunit:9
           coverage: none
           ini-file: development
           ini-values:
@@ -75,6 +74,7 @@ jobs:
         env:
           DB_PORT: 4444
         run: |
+          composer install
           mkdir data
           ./occ maintenance:install --verbose --database=sqlite --database-name=nextcloud --database-host=127.0.0.1 --database-port=$DB_PORT --database-user=autotest --database-pass=rootpassword --admin-user admin --admin-pass admin
           php -f index.php
@@ -83,8 +83,7 @@ jobs:
         run: ./occ app:enable admin_audit encryption federatedfilesharing federation files_sharing files_trashbin files_versions provisioning_api user_ldap
 
       - name: PHPUnit
-        working-directory: tests
-        run: phpunit --configuration phpunit-autotest.xml --exclude-group DB,SLOWDB,PRIMARY-azure,PRIMARY-s3,PRIMARY-swift
+        run: composer run test -- --exclude-group DB,SLOWDB,PRIMARY-azure,PRIMARY-s3,PRIMARY-swift
 
   summary:
     runs-on: ubuntu-latest
--- a/.github/workflows/nodb.yml
+++ b/.github/workflows/nodb.yml
@@ -1,0 +1,79 @@
+name: PHPUnit nodb
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - master
+      - stable*
+
+permissions:
+  contents: read
+
+concurrency:
+  group: phpunit-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  phpunit-nodb:
+    runs-on: ubuntu-latest
+
+    if: ${{ github.repository_owner != 'nextcloud-gmbh' }}
+
+    strategy:
+      matrix:
+        php-versions: ['8.0', '8.1', '8.2', '8.3']
+
+    steps:
+      - name: Checkout server
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          submodules: true
+
+      - name: Install tools
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ffmpeg imagemagick libmagickcore-6.q16-3-extra
+
+      - name: Set up php ${{ matrix.php-versions }}
+        uses: shivammathur/setup-php@9c77701ae57b0c47f6732beebfbdec76e4e5c90a #debian bookworm fix
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: apcu, ctype, curl, dom, fileinfo, gd, imagick, intl, json, ldap, mbstring, openssl, pcntl, pdo_sqlite, posix, redis, sqlite, xml, zip
+          tools: phpunit:9
+          coverage: none
+          ini-file: development
+          ini-values:
+            apc.enabled=on, apc.enable_cli=on, disable_functions= # https://github.com/shivammathur/setup-php/discussions/573
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up Nextcloud
+        env:
+          DB_PORT: 4444
+        run: |
+          mkdir data
+          ./occ maintenance:install --verbose --database=sqlite --database-name=nextcloud --database-host=127.0.0.1 --database-port=$DB_PORT --database-user=autotest --database-pass=rootpassword --admin-user admin --admin-pass admin
+          php -f index.php
+
+      - name: Enable additional apps
+        run: ./occ app:enable admin_audit encryption federatedfilesharing federation files_sharing files_trashbin files_versions provisioning_api user_ldap
+
+      - name: PHPUnit
+        working-directory: tests
+        run: phpunit --configuration phpunit-autotest.xml --exclude-group DB,SLOWDB,PRIMARY-azure,PRIMARY-s3,PRIMARY-swift
+
+  summary-nodb:
+    permissions:
+      contents: none
+    runs-on: ubuntu-latest
+    needs: phpunit-nodb
+
+    if: always()
+
+    name: phpunit-nodb-summary
+
+    steps:
+      - name: Summary status
+        run: if ${{ needs.phpunit-nodb.result != 'success' }}; then exit 1; fi
--- a/.github/workflows/nodb.yml
+++ b/.github/workflows/nodb.yml
@@ -16,10 +16,32 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  phpunit-nodb:
+  changes:
     runs-on: ubuntu-latest
+    name: nodb-changes
 
-    if: ${{ github.repository_owner != 'nextcloud-gmbh' }}
+    # Required permissions
+    permissions:
+      pull-requests: read
+    # Set job outputs to values from filter step
+    outputs:
+      phpunit: ${{ steps.filter.outputs.phpunit }}
+    steps:
+      - name: Checkout server
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          submodules: true
+      - uses: dorny/paths-filter@4512585405083f25c027a35db413c2b3b9006d50 # v2.11.1
+        id: filter
+        with:
+          filters: .github/filters.yml
+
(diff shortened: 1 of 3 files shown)


skjnldsv:
I did some work in this one: https://github.com/skjnldsv/server/pull/13
If you want to have a look if anything is worth taking from

.github/workflows/nodb.yml
```diff
@@ -0,0 +1,92 @@
+name: PHPUnit nodb
+
+on:
+  pull_request:
+    paths:
```
kesselb, line 5: The paths filter hopefully works like https://github.com/nextcloud/server/blob/master/tests/drone-run-php-tests.sh :see_no_evil:

kesselb:
> If you want to have a look if anything is worth taking from

Thanks for letting me know, looks very nice :+1: 

I added the paths filter and `when-unrelated` workflow, hopefully it will work. 

Also tried adding the run script to composer.json, but that didn't work (in the workflow) because phpunit-autotest.xml must be called from the tests/ folder because the paths are relative.

.github/workflows/nodb.yml
```diff
@@ -0,0 +1,92 @@
...
+      - '**.json'
+      - '!package.json'
+      - '!package-lock.json'
```
nickvergessen, line 8: This is not the same as the chained grep + grep-v in the bash script?

kesselb:
```
 - '**.json'
 - '!package.json'
 - '!package-lock.json'
```

This should match every json file except package.json and package-lock.json[^1]

[^1]: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths

nickvergessen: > Using an exclamation mark (!) in front of a pattern negates it. When a file matches a pattern and also matches a negative pattern defined later in the file, the file will not be included.

Okay then

come-nc:
> > I’m not sure if you can use ini-file: development along with ini-values: but if yes it would make sure the tests are running without php deprecation warnings.
> 
> Thank you, good to know +1
> 
> I don't mind the deprecation warnings, just documented them to explain the difference in the test outcome. Do you want me to set ini-file to development?

Yes please, I tested in https://github.com/nextcloud/server/pull/39796 for 32 bits CI and it does work with ini-values.
The CI should run without deprecation warnings (so it should fail on those if there are any).

kesselb:
Make sense :+1: 

I first thought you were referring to the PHPUnit deprecation warnings, but you wanted to let me know that setup-php uses the php.ini-production by default and therefore `error_reporting = E_ALL & ~E_DEPRECATED & ~E_STRICT`.

kesselb:
```
1) Test\Command\CronBusTest::testClosureFromPreviousVersion
--- Expected
+++ Actual
@@ @@
-'opis'
+''
```

https://github.com/nextcloud/server/pull/40505

kesselb:
Good for review :tada:

.github/workflows/nodb-when-unrelated.yml
```diff
@@ -0,0 +1,34 @@
...
+      - '!**.json'
+      - 'package.json'
+      - 'package-lock.json'
+      - '!**.sh'
+      - '!**.yml'
+      - '!**.xml'
+      - '!**.php'
+      - '!**/tests/**'
+      - '!3rdparty'
+      - '!apps/theming/css'
```
nickvergessen, line 15: You have to use the quirk from https://github.com/nextcloud/.github/pull/230/files otherwise this executes way too often

kesselb: Good point.

I think the paths filter for the nodb-when-unrelated workflow is wrong anyway.
It only excludes paths, but does not include anything and will therefore never run :see_no_evil:

nickvergessen: > but does not include anything and will therefore never run 🙈

Yeah, I debugged this last week in https://github.com/nickvergessen/turbo-lamp/pulls and the pattern in https://github.com/nextcloud/.github/pull/230/files is what we need to use. So you need to paths-ignore the same thing in unrelated you require in the actual job and then add the work around with the second exclude list, to make sure the fake summary does not run when the real summary should be used.

kesselb: What do you think about https://github.com/nextcloud/server/pull/39734/commits/3d2665d8024a493c3e0791c9be88989627800ccd?

https://github.com/marketplace/actions/paths-changes-filter is similar to our drone-run-php-tests.sh script. 

The workflow is more complicated because we have a changes job, a conditional "main" job and the summary job with a check if the conditional job was required or not. This approach works without the when-unreated workflow. 

I think not having two workflows is a good trade-off for the higher complexity.

kesselb: Examples

- Vue/Js change without phpunit: https://github.com/nextcloud/server/actions/runs/6364565736/job/17280917479?pr=40729
- 3rdparty bump: https://github.com/nextcloud/server/actions/runs/6364580850/job/17280944950?pr=40730

.github/workflows/nodb.yml
```diff
@@ -0,0 +1,102 @@
...
+          ./occ maintenance:install --verbose --database=sqlite --database-name=nextcloud --database-host=127.0.0.1 --database-port=$DB_PORT --database-user=autotest --database-pass=rootpassword --admin-user admin --admin-pass admin
+          php -f index.php
+
+      - name: Enable additional apps
+        run: ./occ app:enable admin_audit encryption federatedfilesharing federation files_sharing files_trashbin files_versions provisioning_api user_ldap
+
+      - name: PHPUnit
+        working-directory: tests
+        # run: phpunit --configuration phpunit-autotest.xml --exclude-group DB,SLOWDB,PRIMARY-azure,PRIMARY-s3,PRIMARY-swift
+        run: phpunit --configuration phpunit-autotest.xml --exclude-group DB,SLOWDB
```
come-nc, line 88:
```suggestion
        run: phpunit --configuration phpunit-autotest.xml --exclude-group DB,SLOWDB,PRIMARY-azure,PRIMARY-s3,PRIMARY-swift
        # run: phpunit --configuration phpunit-autotest.xml --exclude-group DB,SLOWDB
```

See all the skipped tests in PHPUnit output: https://github.com/nextcloud/server/actions/runs/6364562541/job/17280916683?pr=39734
Either configure azure/s3/switft or exclude those tests.

ChristophWurst, line 88: :warning: soft-conflicts with the changes in https://github.com/nextcloud/server/pull/40678

kesselb: Done.

kesselb: Workflow adjusted.

kesselb:
> memcache and redis tests are skipped, not sure if we should ignore or configure those, or if they should be tested in their own job?

I think memcache and redis have another job in drone. 
The memcache and redis tests are also skipped on drone.

come-nc:
> > memcache and redis tests are skipped, not sure if we should ignore or configure those, or if they should be tested in their own job?
> 
> I think memcache and redis have another job in drone. The memcache and redis tests are also skipped on drone.

Yes but could they be added to a group that we add to the exclude-groups in  https://github.com/nextcloud/server/pull/39734/files#diff-2c10cc6204050ae8a99af53744687db4d3690d7157685a9a221fba3bb78e6946R86
That would clean up the output of the job, making it easier to see failures/warnings.

kesselb:
> Yes but could they be added to a group that we add to the exclude-groups in [#39734 (files)](https://github.com/nextcloud/server/pull/39734/files#diff-2c10cc6204050ae8a99af53744687db4d3690d7157685a9a221fba3bb78e6946R86)
> That would clean up the output of the job, making it easier to see failures/warnings.

Help yourself.

kesselb:
Replaced by https://github.com/nextcloud/server/pull/41003

PR #39740, opened 2023-08-07
[stable27] feat(HTTPClient): Provide wrapped access to Guzzle's asyncRequest()

nickvergessen:
Backport #38613

- [x] 🚧  To be merged after unfreeze


come-nc (approved):

lib/private/Http/Client/Client.php
```diff
@@ -403,4 +407,215 @@ public function options(string $uri, array $options = []): IResponse {
...
+	 *                   'strict'    => true,     // use "strict" RFC compliant redirects.
+	 *                   'referer'   => true,     // add a Referer header
+	 *                   'protocols' => ['https'] // only allow https URLs
+	 *              ],
+	 *              'sink' => '/path/to/file', // save to a file or a stream
+	 *              'verify' => true, // bool or string to CA file
+	 *              'debug' => true,
+	 *              'timeout' => 5,
+	 * @return IPromise
+	 */
```
Altahrim, line 445: Nit: 
PHPDoc already exists in interface, we could avoid to duplicate it in subclasses, right?

Altahrim (approved):

lib/private/Http/Client/Client.php
nickvergessen: yeah, but not going to change that in the backport :P

Altahrim: Of course! It seems I slightly lose the context during review :D

PR #39756, opened 2023-08-08
Add endpoint for getting disabled user list

come-nc:
* Resolves: # 

## Summary

Adds an endpoint to get the list of disabled users, with offset and limit.
This avoids fetching all the users when opening the disabled user list, fixing the disabled user list for LDAP backed installations which do not have a lot of disabled users. Also avoids spamming the LDAP server when opening disabled user list.

## TODO

- [x] Use the new endpoint to show the disabled user list efficiently https://github.com/nextcloud/server/pull/40169

## Checklist

### DIFF
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -238,6 +238,8 @@ class UsersController extends AUserData {
 	 * @param ?int $limit Limit the amount of users returned
 	 * @param int $offset Offset
 	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, ProvisioningApiUserDetails|array{id: string}>}, array{}>
+	 *
+	 * 200: Disabled users details returned
 	 */
 	public function getDisabledUsersDetails(?int $limit = null, int $offset = 0): DataResponse {
 		$currentUser = $this->userSession->getUser();
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -211,7 +211,6 @@ class UsersController extends AUserData {
 			$users = array_merge(...$users);
 		}
 
-		/** @var array<string, ProvisioningApiUserDetails|array{id: string}> $usersDetails */
 		$usersDetails = [];
 		foreach ($users as $userId) {
 			$userId = (string) $userId;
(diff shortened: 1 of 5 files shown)


apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -228,6 +228,71 @@ public function getUsersDetails(string $search = '', int $limit = null, int $off
...
+					)
+				);
+				if (($tempLimit !== null) && (count($users) >= $tempLimit)) {
+					break;
+				}
+			}
+			$users = array_slice($users, $offset);
+		}
+
+		/** @var array<string, ProvisioningApiUserDetails|array{id: string}> $usersDetails */
```
provokateurin, line 277: Doing this is dangerous. Is there no other way than overriding the type? I guess it has to be done because of the union type?

provokateurin:
There is a new CI workflow that checks that the OpenAPI specs are up-to-date. Please rebase onto master and regenerate the specs so that the workflow can do it's job.

apps/provisioning_api/lib/Controller/UsersController.php
provokateurin: I think that can be done with @psalm-assert or similar, although it shouldn't be necessary

come-nc: I removed it and psalm is still happy

provokateurin (approved):
API LGTM

nfebe (approved):
A default value for limit seems reasonable of if null, we should check for valid values different from zero... Such as negative numbers and 0

```php
    if ($limit !== null && $limit < 0) {
    // Whatever exception
        throw new InvalidArgumentException("Invalid limit value: $limit");
    }

    if ($offset < 0) {
        throw new InvalidArgumentException("Invalid offset value: $offset");
    }
```
Generally, this is a little slim or error handling.

come-nc:
> A default value for limit seems reasonable of if null, we should check for valid values different from zero... Such as negative numbers and 0
> 
> ```
>     if ($limit !== null && $limit < 0) {
>     // Whatever exception
>         throw new InvalidArgumentException("Invalid limit value: $limit");
>     }
> 
>     if ($offset < 0) {
>         throw new InvalidArgumentException("Invalid offset value: $offset");
>     }
> ```
> 
> Generally, this is a little slim or error handling.

Nice input, I’ve added that to the controller method.
For the internal method I consider the caller knows what he’s doing, but for the controller it makes sense to block invalid values.

Altahrim (approved):

PR #39770, opened 2023-08-09
fix: always use display name from correct backend

max-nextcloud:
Overwrite the display name after the account is initialized when using an instacne of IGetDisplayNameBackend.

Before when using a variation of user_oidc and registering a Backend.php implementing IGetDisplayNameBackend
the personal setting page shows 'uid'.

The UserManager/AccountManager seems not to use consistently the correct backend.

The correct backend is used in this sequence:

server/lib/private/TemplateLayout.php

          $userDisplayName = \OC_User::getDisplayName();
          $this->assign(user_displayname, $userDisplayName);

In the settings page, it definitely not calls the registered backend, but seems to fall back to default Backend and shows (usually) uid or a value from the standard account property table.

* Resolves: # 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Accounts/AccountManager.php
+++ b/lib/private/Accounts/AccountManager.php
@@ -62,6 +62,7 @@ use OCP\L10N\IFactory;
 use OCP\Mail\IMailer;
 use OCP\Security\ICrypto;
 use OCP\Security\VerificationToken\IVerificationToken;
+use OCP\User\Backend\IGetDisplayNameBackend;
 use OCP\Util;
 use Psr\Log\LoggerInterface;
 use function array_flip;
@@ -744,6 +745,10 @@ class AccountManager implements IAccountManager {
 			return $cached;
 		}
 		$account = $this->parseAccountData($user, $this->getUser($user));
+		if ($user->getBackend() instanceof IGetDisplayNameBackend) {
+			$property = $account->getProperty(self::PROPERTY_DISPLAYNAME);
+			$account->setProperty(self::PROPERTY_DISPLAYNAME, $user->getDisplayName(), $property->getScope(), $property->getVerified());
+		}
 		$this->internalCache->set($user->getUID(), $account);
 		return $account;
 	}


juliushaertl:
Looks sane, one thing I'll still want to double check is if this properly propagates to the system addressbook

lib/private/Accounts/AccountManager.php
```diff
@@ -744,6 +745,9 @@ public function getAccount(IUser $user): IAccount {
 			return $cached;
 		}
 		$account = $this->parseAccountData($user, $this->getUser($user));
+		if ($user->getBackend() instanceof IGetDisplayNameBackend) {
+			$account->setProperty(self::PROPERTY_DISPLAYNAME, $user->getDisplayName());
```
max-nextcloud, line 749: okay... so this would need a scope and verified parameter.

juliushaertl: We can just take the existing scope/verified parameter from the account object obtained earlier i think

max-nextcloud: I now reuse the ones that were set on the original property from the account data as that mimics what the previous patch did.
I'm not sure that's what we want though. Given that we are getting the display name from the backend it's probably verified even if the display name in the original account data was not.

juliushaertl: I think we do not make any use of the verified parameter anyways for the displayname, the scope is user configured so it should be fine with this approach.

juliusknorr (approved):
Looks good to me 👍

juliushaertl:
Asking @Pytal and @ChristophWurst for review as I think both worked with that accounts code in the past

ChristophWurst (approved):
Makes sense

tsdicloud:
Perfect. You helped me to save some nerves for the next release!

juliushaertl:
@max-nextcloud Can you take care of the manual backports?

PR #39779, opened 2023-08-09
Files : Left panel dropdown improvement for NC 25

Jerome-Herbinet:
Respawn of https://github.com/nextcloud/server/pull/38338

**/!\ Reviewers' test needed**

## Before : 
[before 2.webm](https://github.com/nextcloud/server/assets/33763786/823c8457-145a-423d-a9a9-39996205f607)

## After : 
[after 2.webm](https://github.com/nextcloud/server/assets/33763786/fe22628f-f9cf-4645-8e90-db76bbc458bc)

## Checklist


jancborchardt (changes requested):
Let me know if the video is outdated @Jerome-Herbinet, just same feedback as in the other review. :)

The triangle pointing to the right is confusing, it would be expected to have a triangle pointing down when in closed state, indicating more things open downwards. :)

When open, the arrow should still point down too. Then it would behave exactly like a standard HTML select element.

Jerome-Herbinet:
> Let me know if the video is outdated @Jerome-Herbinet, just same feedback as in the other review. :)
> 
> The triangle pointing to the right is confusing, it would be expected to have a triangle pointing down when in closed state, indicating more things open downwards. :)
> 
> When open, the arrow should still point down too. Then it would behave exactly like a standard HTML select element.

@jancborchardt here is an updated video : 
[Capture vidéo du 10-08-2023 11:11:35.webm](https://github.com/nextcloud/server/assets/33763786/36a5f22d-49a0-4915-94c7-839762e15509)

jancborchardt (approved):
Very nice! Would be good to also habe hover/focus feedback on the triangle button, but that can also be done in a follow-up.

Jerome-Herbinet:
> Very nice! Would be good to also habe hover/focus feedback on the triangle button, but that can also be done in a follow-up.

Thanks @jancborchardt ! Can you mention another appropriate reviewer for testing + approval ?

jancborchardt:
@skjnldsv @szaimen @susnux another review needed. :)

jancborchardt:
Also cc @AndyScherzinger would be nice to get this fix in for 27.1.0 as it makes the navigation subitems more obvious.

AndyScherzinger:
Checked with the devs, this PR is v25-only (!)
v26+ the left sidebar got migrated to vue in contrast to 25 being legacy JS. So this can't be used on anything but 25 and would need to be completely re-implemented for 26+ -> not happening for 27.1 then

jancborchardt:
@Jerome-Herbinet do you think this is still applicable then? I’d even say since it will soon be 3 versions old anyway, let’s look to the future instead?

AndyScherzinger:
> Ah got it, sorry for the noise then!

All good, still worth the question.

> @Jerome-Herbinet do you think this is still applicable then? I’d even say since it will soon be 3 versions old anyway, let’s look to the future instead?

I'd second that question. @Jerome-Herbinet would you be willing to work on the solution for the `master` branch? That is targeting v28 but we could then still discuss a back-port to `stable27` even though I believe it won't make it for the September release, so best maybe then to aim straight for v28.

skjnldsv:
I',m not sure I get it.
This is only about the triangle on the right instead of left, right?
Then Nextcloud 27 and 28 already have those :thinking: 

What's left to do on master then? :thinking:

Jerome-Herbinet:
> I'm not sure I get it. This is only about the triangle on the right instead of left, right?
> Then Nextcloud 27 and 28 already have those :thinking: 
> 
> What's left to do on master then? :thinking: 

ASFAIK recent major NC versions (since 26 ?), manage this in nextcloud-vue and it might be already done ; testing and confirmation might be necessary on 26 27 and 28 if possible.

skjnldsv:
Absolutely, since 26 this is nc/vue, so we're good.
I think we can close then :)

AndyScherzinger:
Well, then we could also just merge this one, no? Since it is already "resolved" on 26+ @skjnldsv @Jerome-Herbinet 
Or we drop it since it is resolved on 26+

skjnldsv:
I would tend to be on the careful side here and avoid touching `core/css/apps.scss` on a version that's going unsupported in a few weeks :)

Jerome-Herbinet:
> Well, then we could also just merge this one, no? Since it is already "resolved" on 26+ @skjnldsv @Jerome-Herbinet Or we drop it since it is resolved on 26+

@AndyScherzinger, I've just checked if my personal 26.0.5 instance ships this UI enhancement ... but it's not the case 🤔 ... that's strange, although this PR was merged https://github.com/nextcloud-libraries/nextcloud-vue/pull/4103
Cc @skjnldsv

skjnldsv:
Indeed, this is v8.0.0 and  v7.11.5+ only
So it will come in upcoming 26.0.6 https://github.com/nextcloud/server/pull/39812

PR #39780, opened 2023-08-09
log imaginary errors as info to not spam the server logs

szaimen:
Reason: usually if preview generation fails, the admin cannot do much about it. In order to not spam the error logs, move the logs to info mode. Additional logs are shown in the imaginary container anyway, so should still be pretty easy to debug.


AndyScherzinger (approved):

tcitworld (approved):

szaimen:
CI failure unrelated

Issue #39140, opened 2023-07-04
Compare 2 versions of a file


AndyScherzinger:
Provide an option to compare 2 versions of a file by marking them and opening both side-by-side in some kind of a "dual-pane", so the user can scroll both to visually compare them.

AndyScherzinger:
The idea is to just display 2 versions next to each other for the ones where we have a viewer available.
Like with viewing the version see https://github.com/nextcloud/server/issues/39139 any version should be read-only except for the current version

jancborchardt:
@nextcloud/designers it would be good to have a mockup for this, how to select the 2 versions and compare them. :)

marcoambrosini:
I'm taking this on :) 

The requirements are not very clear from the issue so can we formulate a problem statement for this feature? What is the goal? Is the comparison of the two files oriented at picking one of the two? 

Is it only for text files or also for other files like spreadsheets, pictures, videos and so on?

jancborchardt:
@marcoambrosini it would be a very simple side-by side comparison of 2 documents.

To keep it simple:
- It is only possible to pick one version to compare to the current one. This would be done by an action menu entry "Compare version"
- It is a side-by-side view, no diff or anything like that. The current version is on the right.
- Changes can be made only to the current version, the other document is read-only.
- We will just not make this possible on mobile, the option to "Compare version" will just open the version, as if you clicked on the name.
- The individual views side-by-side could have individual x-close buttons, which would then lead to the other side taking back the whole width.

@artonge do you need mockups beyond this?

juliushaertl:
No further input needed for now from the design side I think. 

@skjnldsv This would require us to have some sort of API to trigger a side-by-side rendering in viewer. My suggestion also considering the short time available would be to simply add an optional parameter to pass a second file info object:

```
OCA.Viewer.open({
	fileInfo: currentFileInfo,
	fileInfoCompare: versionFileInfo,
})
```

This would align with the work already done in https://github.com/nextcloud/server/pull/39171/files#diff-8c16136a0fe1baf1357890af01439aebe1da2380f96f80afb1eab039157311b8R191-R198

Any concerns or objections on that?

skjnldsv:
I'm actually not sure about this.
I think this is not the job of the Viewer, no?

Any other app would benefit from this feature otherwise?
Aside from Text?

juliushaertl:
> Any other app would benefit from this feature otherwise?
> Aside from Text?
[...]

PR #39786, opened 2023-08-10
fix: Also cleanup version metadata if expiring

juliusknorr:
Found while writing e2e tests for #39140 

Steps to reproduce:

- Upload some file multiple times with close timestamps
- Run cron which expires versions (ref https://github.com/nextcloud/server/blob/26c7cda5899d88470cf6e73fb7003c0ee0df05ac/apps/files_versions/lib/Storage.php#L87-L100)

This leads to the version files being deleted while the oc_files_versions table still contains references, meaning that they show up as broken files in the versions sidebar.

Ideally expiration should also be handled by the version manager, but this requires some larger efforts and is not backportable.

Quick script to reproduce:
```bash
#!/bin/bash
NC_URL=https://admin:admin@nextcloud.local/remote.php/webdav/

filename=$(date | md5sum | cut -d " " -f 1)
echo $filename

echo a1 > 1/$filename.md
sleep 2
echo a2 > 2/$filename.md
sleep 2
echo a3 > 3/$filename.md
sleep 2
echo a4 > 4/$filename.md
sleep 2
echo a5 > 5/$filename.md

curl -T "1/$filename.md" "$NC_URL/$filename.md"
curl -T "2/$filename.md" "$NC_URL/$filename.md"
curl -T "3/$filename.md" "$NC_URL/$filename.md"
curl -T "4/$filename.md" "$NC_URL/$filename.md"
curl -T "5/$filename.md" "$NC_URL/$filename.md"
```

Afterwards you can compare the following sql queries before and after cron for validation:

```sql
# with the echoed filename
select * from oc_filecache where name like '3882259fffd1ef36e300511488e561be%';
# with the file id
select * from oc_files_versions where file_id = 8740;
```

## Checklist

### DIFF
--- a/apps/files_versions/lib/Storage.php
+++ b/apps/files_versions/lib/Storage.php
@@ -913,6 +913,21 @@ class Storage {
 			}
 
 			foreach ($toDelete as $key => $path) {
+				// Make sure to cleanup version table relations as expire does not pass deleteVersion
+				try {
+					/** @var VersionsMapper $versionsMapper */
+					$versionsMapper = \OC::$server->get(VersionsMapper::class);
+					$file = \OC::$server->get(IRootFolder::class)->getUserFolder($uid)->get($filename);
+					$pathparts = pathinfo($path);
+					$timestamp = (int)substr($pathparts['extension'] ?? '', 1);
+					$versionEntity = $versionsMapper->findVersionForFileId($file->getId(), $timestamp);
+					if ($versionEntity->getLabel() !== '') {
+						continue;
+					}
+					$versionsMapper->delete($versionEntity);
+				} catch (DoesNotExistException $e) {
+				}
+
 				\OC_Hook::emit('\OCP\Versions', 'preDelete', ['path' => $path, 'trigger' => self::DELETE_TRIGGER_QUOTA_EXCEEDED]);
 				self::deleteVersion($versionsFileview, $path);
 				\OC_Hook::emit('\OCP\Versions', 'delete', ['path' => $path, 'trigger' => self::DELETE_TRIGGER_QUOTA_EXCEEDED]);


apps/files_versions/lib/Storage.php
```diff
@@ -914,7 +914,21 @@ public static function expire($filename, $uid) {
...
 				\OC_Hook::emit('\OCP\Versions', 'preDelete', ['path' => $path, 'trigger' => self::DELETE_TRIGGER_QUOTA_EXCEEDED]);
+
+				// Make sure to cleanup version table relations as expire does not pass deleteVersion
+				try {
+					/** @var VersionsMapper $versionsMapper */
+					$versionsMapper = \OC::$server->get(VersionsMapper::class);
+					$file = \OC::$server->get(IRootFolder::class)->getUserFolder($uid)->get($filename);
+					$pathparts = pathinfo($path);
+					$timestamp = (int)substr($pathparts['extension'] ?? '', 1);
+					$version = $versionsMapper->findVersionForFileId($file->getId(), $timestamp);
```
juliushaertl, line 925: We should probably also skip the expire if the version has a label

artonge: Yes, that would be expected, as in:

https://github.com/nextcloud/server/blob/c3c58b66f2f60a9191d673cee3c2544a23acdfdf/apps/files_versions/lib/Storage.php#L595-L607

juliushaertl: Pushed.

artonge (approved):

icewind1991 (approved):

solracsf:
@juliushaertl no 25 ?

juliushaertl:
Only 26+ is affected where the new table for metadata was introduced

Issue #39185, opened 2023-07-06
File picker: use same sort order as in the main Files view


AndyScherzinger:

AndyScherzinger:
@jancborchardt @nimishavijay regarding the sorting. Given the headline, the sorting shall always be matching the main Files view? This is fine for starters I guess. In a future iteration we could also discuss if we have sorting options in the picker and they are persisted and whenever a user changed the sorting in the picker then from that moment on that value will be used indefinitely.

jancborchardt:
Yep @AndyScherzinger, exactly how I would say we should cut it into 2 steps, with the automatic sorting takeover being simplest.

Issue #38926, opened 2023-06-21
📍 File picker dialog improvements


jancborchardt:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
The current file picker is very basic at the moment – after talking with @jospoortvliet about it we identified some improvements:

**Describe the solution you'd like**
- [ ] #39185
- [x] #39186
- [x] #39187
- [x] #39188

What do you think @nextcloud/designers – any additional points, and who is interested in doing mockups for it? :)

AndyScherzinger:
@jancborchardt since we also had this issue in the main view a while ago:

> Ability to search in the picker

* 🔎_Search_ or 🌪️_Filter_ (files currently does both, first filters current folder listing, yet searches everywhere in parallel)
* Where would you search? In the folder you are currently in? "Everywhere" files related? (bullet point (1) might implicitly answer it dep4ending on your answer to it)

jancborchardt:
@AndyScherzinger we could start with Filter in current view rather than search. But also, out of the things listed, that is the least priority issue (also cause probably needs a lot of investment).

nimishavijay:
Will pick this up for mockups! 

Additionally, it would be great to allow sharing via link from the file picker itself wherever appropriate. Related issue: https://github.com/nextcloud/mail/issues/6080 what do you think?

jancborchardt:
> Additionally, it would be great to allow sharing via link from the file picker itself wherever appropriate. Related issue: https://github.com/nextcloud/mail/issues/6080 what do you think?

@nimishavijay sounds good as an additional optional mockup, yep! :)

AndyScherzinger:
please align any specific @nimishavijay and @susnux - I guess sorting and remembering the last location are clear - however I would challenge the sorting part, but will do in the respective ticket.

nimishavijay:
@susnux The new file picker can look like this:

![image](https://github.com/nextcloud/server/assets/52440189/9a8ee72b-b023-46a0-a0f8-7719fbe0f191)
![image](https://github.com/nextcloud/server/assets/52440189/c32495a1-8a72-4578-8711-91683d69b604)


[...]

PR #39792, opened 2023-08-10
feat: Add new Vue FilePicker from `@nextcloud/dialogs` and use it by default

### LINKED ISSUE #39186: File picker: remember last location/folder you were in, not start in root folder every time

### LINKED ISSUE #39187: File picker: ability to filter

### LINKED ISSUE #39188: File picker: left sidebar with the favorite folders for quick jumping

susnux:
* Part of #38926
  * Resolves: #39186
  * Resolves: #39187
  * Resolves: #39188

## Summary
*(Still providing the legacy one until the Vue FilePicker is out of beta.)*

Basically just use the new component from the dialogs package, see information in the PR https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/878

### Screenshots

image

https://github.com/nextcloud-libraries/nextcloud-dialogs/assets/1855448/3f5ec8fe-ec11-4308-824f-5dca90c72752

And on smaller screens:
image

## TODO

- [ ] See list on the dialogs PR: https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/878
- Especially https://github.com/nextcloud/server/issues/39185 will come after the first beta.

## Checklist


AndyScherzinger (approved):

core/src/OC/dialogs.js
```diff
@@ -246,7 +250,7 @@ const Dialogs = {
 	 * @param {Object} [options] additonal options that need to be set
 	 * @param {Function} [options.filter] filter function for advanced filtering
 	 */
-	filepicker: function(title, callback, multiselect, mimetypeFilter, modal, type, path, options) {
+	legacyFilepicker: function(title, callback, multiselect, mimetypeFilter, modal, type, path, options) {
```
skjnldsv, line 253: deprecate message ? :)

AndyScherzinger:
> (Still providing the legacy one until the Vue FilePicker is out of beta.)

but files will use the new one right-away?

core/src/OC/dialogs.js
susnux: sure :)

susnux:
> but files will use the new one right-away?

Correct by default the new one is used everywhere.

szaimen (dismissed):
A bit feedback from design perspective:
1. Why is there an empty space between home button and plus button?
2. Why is there a checkbox to select folders?
image
3. On mobile, the seearch and multiselect looks distorted. They should probably be on different lines?
image

jancborchardt (dismissed):
Nice! :) Only 2 additional things besides @szaimen’s points:
- [ ] The loading skeleton screen elements are too dark on the background. We had a similar issue in Mail or elsewhere, I thought it was fixed in the components?
- [ ] The "Modified" column should be left-aligned like it is now in files. Only "Size" is right-aligned as it has numbers which can extend to the left, but modified is numbers + text.

susnux:
Thank you for your feedback :)

@szaimen 
> Why is there an empty space between home button and plus button?
I think this is an issue with the nextcloud vue library as this is from the NcBreadcrumbs component.

> Why is there a checkbox to select folders?
Currently there is always checkboxes shown (if multiselect is enabled then there is also a select all one in the table header).
Which brings up the question if this should be changed.

---

@jancborchardt 
>  The loading skeleton screen elements are too dark on the background.

You are right, on darkmode I did not noticed it that much but I agree we should make them lighter.

> The "Modified" column should be left-aligned like it is now in files.
Thank you, will change this :)

szaimen:
Thanks @susnux! :)

> > Why is there an empty space between home button and plus button?
> 
> I think this is an issue with the nextcloud vue library as this is from the NcBreadcrumbs component.

I see, I can create an issue for this if you want :)

> > Why is there a checkbox to select folders?
> 
> Currently there is always checkboxes shown (if multiselect is enabled then there is also a select all one in the table header).
> Which brings up the question if this should be changed.

Yes, I think so :)

---

Please dont forget my third point :)

> On mobile, the seearch and multiselect looks distorted. They should probably be on different lines?

susnux:
> Please dont forget my third point :)
Yes :sweat_smile: 

I thought I had already fixed this but seems like the styling was not updated.

nimishavijay:
Super cool! :) Some small points from my side as well:
- [ ] The heading for the entire dialog is somewhat off center, would it be possible to center it? :) Similar to Talk/Mail/Files settings dialogs 
image

- [ ] The icons for each file/folder also seem a bit big, we could change it to a 20px [MDI folder icon](https://fonts.google.com/icons?selected=Material%20Icons%20Round%3Afolder%3A). This is different from the way it currently looks in Files, but we should change it in Files also after F2V.
- [ ] Small detail, in the header "Name" can be left aligned with the text in the rows below (currently it is left aligned with the icon)



> Currently there is always checkboxes shown (if multiselect is enabled then there is also a select all one in the table header).
Which brings up the question if this should be changed.

If only one file/folder can be selected (eg. while moving/copying a file) then it doesn't make much sense to show the checkboxes as they wouldn't do anything. However if multiple files/folders can be selected (eg. when adding attachments to an email) then the checkboxes could be shown to indicate selection, what do you think?

skjnldsv:
Maybe we can summarize all design requests in an issue in https://github.com/nextcloud-libraries/nextcloud-dialogs ?

Pytal (approved):

susnux:
Comments of @jancborchardt are resolved. The third one of @szaimen also (the first one is probably a nextcloud-vue bug, the second one will be in a future beta release).

---
@nimishavijay Thank you very much :)

> Super cool! :) Some small points from my side as well:
> 
>     * [ ]  The heading for the entire dialog is somewhat off center, would it be possible to center it? :) Similar to Talk/Mail/Files settings dialogs

Well to be honest I implemented it according to the mockup in https://github.com/nextcloud/server/issues/38926#issuecomment-1632560834
We should (at a different location) discuss if we should adjust the settings dialog instead by using the new general purpose dialog component as the base.

>     * [ ]  The icons for each file/folder also seem a bit big, we could change it to a 20px [MDI folder icon](https://fonts.google.com/icons?selected=Material%20Icons%20Round%3Afolder%3A). This is different from the way it currently looks in Files, but we should change it in Files also after F2V.

Currently the same as the files app (even as the new f2v field app), so with Nextcloud 28 we can adjust this, what do you think?
Especially when previews instead of generic icons are shown.

But generally lets open a new issue on the dialogs library and do a design review session for the next iteration, would that work for you? :)

nimishavijay:
> Well to be honest I implemented it according to the mockup 

Ah! I see the confusion, in the mockup it is actually centered, but it just happens to align perfectly with the file list section :)
Would it be possible to just center it for the upcoming release, and address the other points for 28? :)

skjnldsv (approved):

skjnldsv:
Rebasing...

skjnldsv:
Needs https://github.com/nextcloud/server/pull/39955

skjnldsv:
If the cypress user test fails with `Request failed with status code 500`
It means the path given and returner by the picker is invalid :)

skjnldsv:
Added a theming fix.
Now we need a new dialogs release and we should be good :rocket:

susnux:
should be ready to go?

skjnldsv:
Congrats!! :tada: :tada: :tada:

PR #39806, opened 2023-08-11
fix(updatenotification): Skip update check

Pytal:
## Summary

Skip update check if the config option https://github.com/nextcloud/server/blob/f0db3d0de1785c17097f6e8ffd73e0f9221528d9/config/config.sample.php#L818-L823 is `false`

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #39829, opened 2023-08-11
[stable24] fix(updatenotification): Skip update check

Pytal:
Backport https://github.com/nextcloud/server/pull/39806


Pytal:
EOL

PR #39860, opened 2023-08-14
don't preload metadata for the sub-childen

icewind1991:
Fixes to broad preloading logic introduced with https://github.com/nextcloud/server/pull/39694

- Only preload when `Depth = 1`, this stops preloading for the children of sub-folders
- Don't trigger preload logic when no metadata is requested.

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesPlugin.php
@@ -465,7 +465,7 @@ class FilesPlugin extends ServerPlugin {
 
 			$requestProperties = $propFind->getRequestedProperties();
 
-			if ($this->config->getSystemValueBool('enable_file_metadata', true)) {
+			if ($this->config->getSystemValueBool('enable_file_metadata', true) && $propFind->getDepth() === 1) {
 				$requestedMetaData = [];
 				foreach ($requestProperties as $requestProperty) {
 					if (isset(self::ALL_METADATA_PROPS[$requestProperty])) {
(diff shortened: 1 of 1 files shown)


artonge (approved):

apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -465,13 +465,17 @@ public function handleGetProperties(PropFind $propFind, \Sabre\DAV\INode $node)
...
+			$requestedMetaData = [];
+			foreach ($requestProperties as $requestProperty) {
+				if (isset(self::ALL_METADATA_PROPS[$requestProperty])) {
+					$requestedMetaData[] = self::ALL_METADATA_PROPS[$requestProperty];
 				}
+			}
+			if (
+				$this->config->getSystemValueBool('enable_file_metadata', true) &&
+				$propFind->getDepth() === 1 &&
+				$requestedMetaData
```
ArtificialOwl, line 477: any reason for this 3rd condition ?

ArtificialOwl (approved):

Altahrim (approved):

apps/dav/lib/Connector/Sabre/FilesPlugin.php
icewind1991: it allows skipping the logic if no metadata fields are requested

ArtificialOwl: but it is stays `true` even if array is empty, right ?

ArtificialOwl: nevermind, it seems that an empty array will returns `false`

pulsejet:
This patch cut down my SQL requests to half 👀
Should've been labelled Performance 🚀😉

PR #39864, opened 2023-08-14
reuse l10n and request in dav folder listing

icewind1991:
instead of having to query those once for every node

saved 15-20% when loading large folders during local testing: https://blackfire.io/profiles/compare/d33a9993-e3a4-4ea4-b9f2-f1b38f576937/graph

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Directory.php
+++ b/apps/dav/lib/Connector/Sabre/Directory.php
@@ -35,16 +35,19 @@ namespace OCA\DAV\Connector\Sabre;
 use OC\Files\Mount\MoveableMount;
 use OC\Files\View;
 use OC\Metadata\FileMetadata;
+use OCA\DAV\AppInfo\Application;
 use OCA\DAV\Connector\Sabre\Exception\FileLocked;
 use OCA\DAV\Connector\Sabre\Exception\Forbidden;
 use OCA\DAV\Connector\Sabre\Exception\InvalidPath;
-use OCA\DAV\Upload\FutureFile;
 use OCP\Files\FileInfo;
 use OCP\Files\Folder;
 use OCP\Files\ForbiddenException;
 use OCP\Files\InvalidPathException;
 use OCP\Files\NotPermittedException;
 use OCP\Files\StorageNotAvailableException;
+use OCP\IL10N;
+use OCP\IRequest;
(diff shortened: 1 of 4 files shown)


apps/dav/lib/Connector/Sabre/Directory.php
```diff
@@ -203,7 +208,7 @@ public function createDirectory($name) {
 	 * @throws \Sabre\DAV\Exception\NotFound
 	 * @throws \Sabre\DAV\Exception\ServiceUnavailable
 	 */
-	public function getChild($name, $info = null) {
+	public function getChild($name, $info = null, Request $request = null, IL10N $l10n = null) {
```
nickvergessen, line 211:
```suggestion
	public function getChild($name, $info = null, IRequest $request = null, IL10N $l10n = null) {
```

Any reason the interface is not sufficient?

apps/dav/tests/unit/Connector/Sabre/RequestTest/Sapi.php
```diff
@@ -23,6 +23,7 @@
  */
 namespace OCA\DAV\Tests\unit\Connector\Sabre\RequestTest;
 
+use OCP\IRequest;
```
nickvergessen, line 26:
```suggestion
```

skjnldsv (approved):

apps/dav/lib/Connector/Sabre/Directory.php
```diff
@@ -265,8 +268,10 @@ public function getChildren() {
 		}
 
 		$nodes = [];
+		$request = \OC::$server->get(IRequest::class);
+		$l10nFactory = \OC::$server->get(IFactory::class);
 		foreach ($folderContent as $info) {
-			$node = $this->getChild($info->getName(), $info);
+			$node = $this->getChild($info->getName(), $info, $request, $l10nFactory->get(Application::APP_ID));
```
nickvergessen, line 274: Let's also drop the `$l10nFactory->get` from the loop?
```
		$l = \OC::$server->get(IFactory::class)->get(Application::APP_ID);
		foreach ($folderContent as $info) {
			$node = $this->getChild($info->getName(), $info, $request, $l);
```

nickvergessen (approved):

PR #39867, opened 2023-08-14
Add a separate event for login page rendering

icewind1991:
Only very few apps actually need to add their scripts/styles to the login page, but it's currently easy for apps to accidentally do that anyway.

By having a separate event, having your scripts on the login page becomes opt-in rather than opt-out.

- Add a separate event for login page rendering
- Move some more event loading logic to the correct events.
- [x] test that nothing explodes?

### DIFF
--- a/apps/files_sharing/lib/AppInfo/Application.php
+++ b/apps/files_sharing/lib/AppInfo/Application.php
@@ -116,11 +116,6 @@ class Application extends App implements IBootstrap {
 
 		Share::registerBackend('file', File::class);
 		Share::registerBackend('folder', Folder::class, 'file');
-
-		/**
-		 * Always add main sharing script
-		 */
-		Util::addScript(self::APP_ID, 'main');
 	}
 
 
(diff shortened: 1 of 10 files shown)


icewind1991:
https://github.com/nextcloud/unsplash/blob/4c04d275e9e2469c5b6f1b3d2b6f24cfe9db9887/lib/EventListener/BeforeTemplateRenderedEventListener.php#L18 is the only non-core app I could find in a quick search that would need to be adjusted

nickvergessen (approved):

icewind1991:
- 2fa apps have their own templates to do script/style adding with
- register app uses the "alternative login" feature instead of events
- user_saml doesn't seem to rely on script injection

Seems safe to me to apply this change.

AndyScherzinger (approved):

ChristophWurst:
image

breaks the installer styling and scripts.

ChristophWurst (commented):
The new event is not emitted

icewind1991:
The downside of having your dev setup perform the setup automatically :see_no_evil:

icewind1991:
https://github.com/nextcloud/server/pull/40494

PR #39883, opened 2023-08-15
Refactor files trashbin app

danialRahimy:
## Summary

I did these items in a single file `apps/files_trashbin/lib/Trashbin.php` for two reasons:
1. Pull request hasn't been a lot of changes.
2. I was not sure about some changes.

- Return early
- Change the const name for more readability
- Add functions' return type
- Update doc types

## Checklist

### DIFF
--- a/apps/files_trashbin/lib/Trashbin.php
+++ b/apps/files_trashbin/lib/Trashbin.php
@@ -198,7 +198,7 @@ class Trashbin {
 	 * @throws CopyRecursiveException
 	 * @throws InvalidPathException
 	 */
-	private static function copyFilesToUser($sourcePath, $owner, $targetPath, $user, $timestamp): void {
+	private static function copyFilesToUser(string $sourcePath, string $owner, string $targetPath, string $user, int $timestamp): void {
 		self::setUpTrash($owner);
 
 		$targetFilename = basename($targetPath);
--- a/apps/files_trashbin/lib/Trashbin.php
+++ b/apps/files_trashbin/lib/Trashbin.php
@@ -399,22 +399,24 @@ class Trashbin {
 	 * @param int $timestamp when the file was deleted
 	 */
 	private static function retainVersions($filename, $owner, $ownerPath, $timestamp) {
-		if (\OCP\Server::get(IAppManager::class)->isEnabledForUser('files_versions') && !empty($ownerPath)) {
-			$user = OC_User::getUser();
-			$rootView = new View('/');
+		if (!\OCP\Server::get(IAppManager::class)->isEnabledForUser('files_versions') || empty($ownerPath)) {
+			return null;
+		}
 
-			if ($rootView->is_dir($owner . '/files_versions/' . $ownerPath)) {
+		$user = OC_User::getUser();
+		$rootView = new View('/');
+
+		if ($rootView->is_dir($owner . '/files_versions/' . $ownerPath)) {
+			if ($owner !== $user) {
+				self::copy_recursive($owner . '/files_versions/' . $ownerPath, $owner . '/files_trashbin/versions/' . static::getTrashFilename(basename($ownerPath), $timestamp), $rootView);
+			}
+			self::move($rootView, $owner . '/files_versions/' . $ownerPath, $user . '/files_trashbin/versions/' . static::getTrashFilename($filename, $timestamp));
+		} elseif ($versions = \OCA\Files_Versions\Storage::getVersions($owner, $ownerPath)) {
+			foreach ($versions as $v) {
 				if ($owner !== $user) {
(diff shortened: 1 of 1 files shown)


apps/files_trashbin/lib/Trashbin.php
```diff
@@ -227,16 +246,21 @@ private static function copyFilesToUser($sourcePath, $owner, $targetPath, $user,
...
 	 * @param bool $ownerOnly delete for owner only (if file gets moved out of a shared folder)
-	 *
 	 * @return bool
+	 * @throws Exception
+	 * @throws CopyRecursiveException
+	 * @throws InvalidPathException
+	 * @throws LockedException
+	 * @throws NoUserException
+	 * @throws QueryException
+	 * @throws ContainerExceptionInterface
```
nickvergessen, line 261: I'm very much against these.
They are not expected and not part of the API. If they happen it's not a user / API consumer fault, but basically the whole nextcloud is offline/broken.

danialRahimy: You're right, and I appreciate your input.
I'll revert the changes to their previous state.

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -66,15 +68,15 @@
...
+	public const DEFAULT_MAX_SIZE = 50;
 
 	/**
 	 * Ensure we don't need to scan the file during the move to trash
 	 * by triggering the scan in the pre-hook
 	 *
 	 * @param array $params
 	 */
-	public static function ensureFileScannedHook($params) {
+	public static function ensureFileScannedHook($params): void {
```
Fenn-CS, line 79:
```suggestion
	public static function ensureFileScannedHook(array $params): void {
```

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -88,9 +90,8 @@ public static function ensureFileScannedHook($params) {
 	 *
 	 * @param string $filename
 	 * @return array
-	 * @throws \OC\User\NoUserException
 	 */
-	public static function getUidAndFilename($filename) {
+	public static function getUidAndFilename($filename): array {
```
Fenn-CS, line 94:
```suggestion
	public static function getUidAndFilename(string $filename): array {
```

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -122,7 +123,7 @@ public static function getUidAndFilename($filename) {
 	 * @param string $user
 	 * @return array (filename => array (timestamp => original location))
 	 */
-	public static function getLocations($user) {
+	public static function getLocations($user): array {
```
Fenn-CS, line 126:
```suggestion
	public static function getLocations(string $user): array {
```

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -146,9 +147,9 @@ public static function getLocations($user) {
 	 * @param string $user
 	 * @param string $filename
 	 * @param string $timestamp
-	 * @return string original location
+	 * @return bool|string original location
 	 */
-	public static function getLocation($user, $filename, $timestamp) {
+	public static function getLocation($user, $filename, $timestamp): bool|string {
```
Fenn-CS, line 152:
```suggestion
	public static function getLocation(string $user, string $filename, string $timestamp): bool|string {
```

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -167,7 +168,10 @@ public static function getLocation($user, $filename, $timestamp) {
 		}
 	}
 
-	private static function setUpTrash($user) {
+	/**
+	 * @param string $user
+	 */
+	private static function setUpTrash($user): void {
```
Fenn-CS, line 174:
```suggestion
	private static function setUpTrash(string $user): void {
```

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -183,17 +187,18 @@ private static function setUpTrash($user) {
...
 	 * @param string $owner
 	 * @param string $targetPath
-	 * @param $user
+	 * @param string $user
 	 * @param int $timestamp
+	 * @throws CopyRecursiveException
+	 * @throws InvalidPathException
 	 */
-	private static function copyFilesToUser($sourcePath, $owner, $targetPath, $user, $timestamp) {
+	private static function copyFilesToUser($sourcePath, $owner, $targetPath, $user, $timestamp): void {
```
Fenn-CS, line 201:
```suggestion
	private static function copyFilesToUser(string $sourcePath, string $owner, string $targetPath, string $user, int $timestamp): void {
```

nfebe (changes requested):
Since this is just a refactor and no main functional change you can just use build in type hinting for PHP... I left a few examples but not all, most of the functions can use built in type hinting and that would allow you to delete the comment annotations at the top (leaving behind just the exceptions)

PR #39888, opened 2023-08-15
Reduce the number of container queries

icewind1991:
Tweaks DI in various places to prevent calling `Server::get` in a loop.

While the container is fairly fast, doing a lot of queries adds up.

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Directory.php
+++ b/apps/dav/lib/Connector/Sabre/Directory.php
@@ -315,20 +315,19 @@ class Directory extends \OCA\DAV\Connector\Sabre\Node implements \Sabre\DAV\ICol
 		}
 	}
 
+	private function getLogger(): LoggerInterface {
+		return \OC::$server->get(LoggerInterface::class);
+	}
+
 	/**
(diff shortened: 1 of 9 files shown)


AndyScherzinger (approved):

Altahrim (approved):

icewind1991:
acceptance failures might be related, will look into it later

apps/dav/lib/Connector/Sabre/Directory.php
```diff
@@ -315,20 +315,19 @@ public function delete() {
...
+
 	/**
 	 * Returns available diskspace information
 	 *
 	 * @return array
 	 */
 	public function getQuotaInfo() {
-		/** @var LoggerInterface $logger */
-		$logger = \OC::$server->get(LoggerInterface::class);
-		if ($this->quotaInfo) {
```
nickvergessen, line 326: is this intentional?

icewind1991: 🙈

nickvergessen (approved):

PR #39897, opened 2023-08-16
sharebymail: remove ILogger and upgrade to PHP 8 syntax

Altahrim:
Ref: #32127 

## Summary
 - Remove ILogger from `sharebymail` application
 - Upgrade to PHP 8 syntax (types, promoted properties)

## Checklist


apps/sharebymail/tests/ShareByMailProviderTest.php
```diff
@@ -247,9 +244,7 @@ public function testCreateSendPasswordByMailWithoutEnforcedPasswordProtection()
 
 		$this->mailer->expects($this->never())->method('send');
 
-		$this->assertSame('shareObject',
-			$instance->create($share)
-		);
+		$this->assertInstanceOf(IShare::class, $instance->create($share));
```
come-nc, line 247: This feels a bit weak as an assertion, create is typed to return an IShare, no?

come-nc (commented):
I would feel better if tests are hardened, otherwise good.

apps/sharebymail/tests/ShareByMailProviderTest.php
Altahrim: Yes, it basically test if above `expects` behave as expected and it is able to build the object.
The refactor don't really change the test.

Do you have another idea about something to test?

ArtificialOwl (approved):

apps/sharebymail/tests/ShareByMailProviderTest.php
come-nc: That the object returned is the one returned by createShareObject, or that its id is 42?

Altahrim: Done!

come-nc (approved):
Apply the same change to the other modified tests please, then all good.

Issue #36980, opened 2023-03-01
[BITV] 9.2.1.1/3.2 - The "Details" area - The combobox for selecting "Tags" is not fully usable with the keyboard. (2)


AndyScherzinger:
It is not possible, among other things, to select or edit a tag from the list. Additionally, the keyboard focus in the list is not noticeable. 

<img width="427" alt="5e91a02de42ea9052604ec3630e2fc8e" src="https://user-images.githubusercontent.com/1315170/222244985-031bed68-a5b4-4c2b-b98c-bfbeac5b848a.png">

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-6bd94ebc49-v3-n2

</details>

- [ ] 9.2.1.1/10.2 - https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-6bd94ebc49-v10-n2

ShGKme:
Seems to be already fixed by moving to `NcSelect`?
![image](https://github.com/nextcloud/server/assets/25978914/abc6c920-ee81-4725-b019-40cdaeb9b106)

AndyScherzinger:
@ShGKme if keyboard navigation works fine for select/edit/remove/add then the keyboard navigation part seems fine. However looking at your screenshot I expect "the keyboard focus in the list is not noticeable." to be still a valid issue that needs fixing (I would expect the contrast to be too low from an a11y perspective would need to be 1:3 or even 1:4,5). Since @JuliaKirschenheuter is on vacation maybe @jancborchardt or @nimishavijay know the right contrast or which to pick (or add a border?)

JuliaKirschenheuter:
If we would use same border as on parent NcSelect element, like 

![Screenshot from 2023-08-14 14-23-28](https://github.com/nextcloud/server/assets/6078378/30d57576-1650-41dc-93be-0f2a5bafc7a4), 
`border: 2px solid var(--color-border-maxcontrast);` then it is ok for a11y.
@nimishavijay @marcoambrosini would you approve such solution?

JuliaKirschenheuter:
>  if keyboard navigation works fine for select/edit/remove/add

select/edit/remove/add seems to work well besides focus (after selecting a tag NcSelect have to be selected itself)

jancborchardt:
@JuliaKirschenheuter yeah sounds good to me! 2 points:
[...]

PR #39909, opened 2023-08-16
fix(settings): focus lost on selects in users settings

ShGKme:
* Resolves: Similar to https://github.com/nextcloud/server/issues/36980, but in User Settigns.

## Summary

After adding/removing groups the focus is lost from `NcSelect`. You need to catch the select again with <kbd>tab</kbd>. This happens because during loading the `NcSelect` become disabled and cannot handle focus. To fix - remove disabling on loading state.

Also fixed some very minor issues in this component.

## 🖼️ Screenshots

Before | After
---|---
before | after
image | image
image | image

## Todo

- [x] Don't disable `NcSelect`s on loading, only keep its loading state 
- [x] Minor fixes
  - [x] Fix loading indicator of user manager update (lost `await`)
  - [x] Fix label set user manager label

## Drawbacks

Now `NcSelect` is still available during the loading, so the user can add a new group or delete a group. But it should work fine and it is the same as what we have with collaborative tags in the Files sidebar.

## Checklist


JuliaKirschenheuter (approved):

JuliaKirschenheuter (commented):
Thank you! 

I guess a text here doesn't have enough contrast ;(
image
Is it an issue of NcSelect?

ShGKme:
> should we remove

It is still used for loading indicators, I only remove its usage from :disable

ShGKme:
> Is it an issue of NcSelect?

Yes...

ChristophWurst (approved):

susnux (approved):

Pytal (approved):

PR #39910, opened 2023-08-16
fix(memcache): Fix comparison of Memcache configs to classes

nickvergessen:
## Summary

In some cases we compared the cache class with a string starting `\` and in some others with the `::class` which does not have the leading `\`, so depending on the config value certain bits of functionality would kick in or not.
Now I'm aligning the class names to be without leading `\` so we can always compare to `::class`

## Checklist


icewind1991 (approved):

ChristophWurst (approved):
:see_no_evil:

PR #39924, opened 2023-08-17
fix: Set X-Requested-With header on all requests to avoid browser auth dialogs

juliusknorr:
This used to work for most cases in the far past as jquery adds this on requests (https://github.com/jquery/jquery/blob/48cc402a917d6011c7d3e75f779f11ef91b474fb/src/ajax/xhr.js#L43-L50)

https://github.com/nextcloud-libraries/nextcloud-axios/pull/637 adds this to our axios wrapper but this PR handles cases where apps or code parts are not using that.

Server has logic in place to return a dummyauth response header in case this one is set:
https://github.com/nextcloud/server/commit/dfc3536d2b95fea1b54b3e85651e3a66c2d0088e

This helps to handle requests that happen in the web UI after authentication was revoked (e.g. by disabling a user) more gracefully, otherwise a request would lead to the browser showing a basic auth dialog.

Steps to reproduce:
- Login as a user
- Disable the user through occ
- Wait until the notifications app sends another request

## Checklist


artonge (approved):

artonge:
In which cases do we not use axios?

juliushaertl:
Old code or apps that might not use our libraries, rare cases but might still happen

skjnldsv (approved):

Pytal (approved):

core/src/utils/xhr-request.js
```diff
@@ -0,0 +1,55 @@
...
+			}
+			if (!init.headers) {
+				init.headers = new Headers()
+			}
+
+			if (init.headers instanceof Headers && !init.headers.has('X-Requested-With')) {
+				init.headers.append('X-Requested-With', 'XMLHttpRequest')
+			} else if (init.headers instanceof Object && !init.headers['X-Requested-With']) {
+				init.headers['X-Requested-With'] = 'XMLHttpRequest'
+			}
```
skjnldsv, line 50: @juliushaertl a small find as of today:
image

If `headers` is an instance of `Headers` and already have `X-Requested-With`, then the second `else if` will still be true.
Luckily, `['X-Requested-With']` on a `Headers` object is ignored by fetch :see_no_evil:

PR #39925, opened 2023-08-17
fix(CardDAV): allow disabling of the system address book 

### LINKED ISSUE #38880: Provide a way to hide / remove the system address book globally
### Describe the solution you'd like

The latest version v27 introduces the system address book or more specifically exposes it to every user as a separate address book named "accounts".
This addressbook is also synced to client devices, like iOS, via CardDAV, which is something I don't want.

I'm missing a global switch to not expose this system address book for all users.

Ref https://github.com/nextcloud/server/issues/19575

miaulalala:
via config option and OCC command

* Resolves: https://github.com/nextcloud/server/issues/38880

## Summary
Allow admins to disable the System Addressbook systemwide.

This might cause complaints from connected clients if the SAB is the default addressbook in their env.

## TODO

- [x] Documentation

## Checklist

### DIFF
--- a/apps/dav/lib/CardDAV/UserAddressBooks.php
+++ b/apps/dav/lib/CardDAV/UserAddressBooks.php
@@ -88,7 +88,8 @@ class UserAddressBooks extends \Sabre\CardDAV\AddressBookHome {
 		$addressBooks = $this->carddavBackend->getAddressBooksForUser($this->principalUri);
 		// add the system address book
 		$systemAddressBook = null;
-		if (is_string($principal) && $principal !== 'principals/system/system' && $this->carddavBackend instanceof CardDavBackend) {
+		$systemAddressBookExposed = $this->config->getAppValue('dav', 'system_addressbook_exposed', 'yes') === 'yes';
+		if ($systemAddressBookExposed && is_string($principal) && $principal !== 'principals/system/system' && $this->carddavBackend instanceof CardDavBackend) {
 			$systemAddressBook = $this->carddavBackend->getAddressBooksByUri('principals/system/system', 'system');
 			if ($systemAddressBook !== null) {
 				$systemAddressBook['uri'] = SystemAddressbook::URI_SHARED;


ChristophWurst:
config:app:set/delete should be sufficient. we don't need dedicated commands right now: https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/occ_command.html#config-commands

tcitworld (changes requested):
I feel like just adding an `if` to the above block where the system address book is added in the first place would be much more efficient, why not do that instead?

(in any case, you need to get the `getAppValue` call outside of the `array_map` so that it doesn't get called for each iteration)

apps/dav/lib/CardDAV/UserAddressBooks.php
```diff
@@ -101,7 +102,7 @@ public function getChildren() {
 		$objects = [];
 		if (!empty($addressBooks)) {
 			/** @var IAddressBook[] $objects */
-			$objects = array_map(function (array $addressBook) {
+			$objects = array_filter(array_map(function (array $addressBook) {
```
tcitworld, line 105: **question:** Still needed?

miaulalala: No, thanks!

tcitworld (approved):

apps/dav/lib/CardDAV/UserAddressBooks.php
```diff
@@ -88,7 +88,8 @@ public function getChildren() {
 		$addressBooks = $this->carddavBackend->getAddressBooksForUser($this->principalUri);
 		// add the system address book
 		$systemAddressBook = null;
-		if (is_string($principal) && $principal !== 'principals/system/system' && $this->carddavBackend instanceof CardDavBackend) {
+		$systemAddressBookDisabled = $this->config->getAppValue('dav', 'system_addressbook_disabled', 'no') === 'yes';
```
tcitworld, line 91: **suggestion:** Now that I think about it, is `disabled` the correct term, since it continues to exist and is synced by `SyncService` for federation? Maybe `system_addressbook_exposed` instead (with default to `'yes'`) ?

tcitworld (commented):
Code is fine but I didn't check if that has other consequences elsewhere.

apps/dav/lib/CardDAV/UserAddressBooks.php
miaulalala: hm that's true but that is no different than the situation before we added the SAB. Federation sync has always been happening

tcitworld: Well no, the system address book was exposed only on the `principals/system/system` principal, not on all principals.

andrey18106 (approved):

miaulalala:
Failing cypress test is unrelated methinks:


```
  Login with a new user and open the files app
    1) "after all" hook for "See the default file welcome.txt in the files list"


  0 passing (7s)
  1 failing

  1) Login with a new user and open the files app
       "after all" hook for "See the default file welcome.txt in the files list":
     CypressError: The following error originated from your application code, not from Cypress. It was caused by an unhandled promise rejection.
```

mbiebl:
Applied this patch and everything is working nicely again after disabling the SAB. Thanks a lot!

sudoer:
I was able to comment out this block of code, and I saw that the "Accounts" address book was no longer exported, as we had hoped.  So I am looking forward to the configuration-based fix soon.

I did have a question.  When this pull request is merged, how would someone set the configuration option?  Is it in a config file?  Or is there an "occ" command to set the option?  Is it eventually stored in the database, or in the code?  I imagine this question is a bit outside of the scope of this PR, but it sure would help to have a pointer to how to enable/disable it.  Thanks!

tcitworld:
```sh
php occ config:app:set dav system_addressbook_exposed --value=no
```

MicKress:
Hi, I have a question. Will the deactivation of system address book only be possible via occ-Command or will there be an point in the administration panel? If you have a managed nextcloud, you cannot often work with occ-commands. Best Regards

exuded:
Hello, when using the command [given in the documentation](https://docs.nextcloud.com/server/latest/admin_manual/groupware/contacts.html#system-address-book), I get the following error:
CleanShot 2024-05-13 at 11 56 35

 I am a bit confused because in the command is no "-a" part

PR #39928, opened 2023-08-17
Refactor userـldap app commands

fsamapoor:
## Summary
I have made some adjustments to the `apps/user_ldap/lib/Command` classes to improve the code readability.

- Adding return types
- Replacing return code integers with more readable strings.
- Using early returns

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Command/CreateEmptyConfig.php
+++ b/apps/user_ldap/lib/Command/CreateEmptyConfig.php
@@ -32,18 +32,13 @@ use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 
 class CreateEmptyConfig extends Command {
-	/** @var \OCA\User_LDAP\Helper */
-	protected $helper;
-
-	/**
-	 * @param Helper $helper
-	 */
-	public function __construct(Helper $helper) {
-		$this->helper = $helper;
+	public function __construct(
+		protected Helper $helper,
+	) {
 		parent::__construct();
 	}
 
-	protected function configure() {
+	protected function configure(): void {
 		$this
 			->setName('ldap:create-empty-config')
 			->setDescription('creates an empty LDAP configuration')
@@ -67,6 +62,6 @@ class CreateEmptyConfig extends Command {
 			$prose = 'Created new configuration with configID ';
 		}
 		$output->writeln($prose . "{$configPrefix}");
-		return 0;
+		return self::SUCCESS;
 	}
 }
--- a/apps/user_ldap/lib/Command/ResetUser.php
+++ b/apps/user_ldap/lib/Command/ResetUser.php
@@ -36,25 +36,15 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Question\Question;
 
 class ResetUser extends Command {
-	/** @var DeletedUsersIndex */
-	protected $dui;
-	/** @var IUserManager */
-	private $userManager;
-	/** @var UserPluginManager */
-	private $pluginManager;
-
 	public function __construct(
-		DeletedUsersIndex $dui,
-		IUserManager $userManager,
-		UserPluginManager $pluginManager
+		protected DeletedUsersIndex $dui,
+		private IUserManager $userManager,
+		private UserPluginManager $pluginManager,
 	) {
-		$this->dui = $dui;
-		$this->userManager = $userManager;
-		$this->pluginManager = $pluginManager;
 		parent::__construct();
 	}
 
-	protected function configure() {
+	protected function configure(): void {
 		$this
 			->setName('ldap:reset-user')
 			->setDescription('deletes an LDAP user independent of the user state')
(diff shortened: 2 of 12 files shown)


apps/user_ldap/lib/Command/CreateEmptyConfig.php
```diff
@@ -67,6 +62,6 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			$prose = 'Created new configuration with configID ';
 		}
 		$output->writeln($prose . "{$configPrefix}");
-		return 0;
+		return self::SUCCESS;
```
Fenn-CS, line 65: Why is a method annotated with void returning something?

apps/user_ldap/lib/Command/ResetUser.php
```diff
@@ -36,25 +36,15 @@
...
+		private UserPluginManager $pluginManager,
 	) {
-		$this->dui = $dui;
-		$this->userManager = $userManager;
-		$this->pluginManager = $pluginManager;
 		parent::__construct();
 	}
 
-	protected function configure() {
+	protected function configure(): void {
```
Fenn-CS, line 47: Same comment as above, this method returns values. Can't be void.

nfebe (commented):
Looks like some void-annotated methods do actually return values?

fsamapoor:
> Looks like some void-annotated methods do actually return values?

Hello @Fenn-CS, Thank you for reviewing the PR.

No method with the return type of `void` returns any values. The git diff seems like the return belongs to the method with a `void` return type, but it actually does not.

These two lines do not belong to the same method:

image

Here is the expanded diff:

image

apps/user_ldap/lib/Command/CreateEmptyConfig.php
fsamapoor: The git diff seems confusing. No value is returned form a method with the `void` return type. I explained it in a separate comment.

apps/user_ldap/lib/Command/ResetUser.php
fsamapoor: The git diff seems confusing. No value is returned from a method with the `void` return type. I explained it in a separate comment.

nfebe (approved):
Thanks Faraz!

blizzz (approved):

apps/user_ldap/lib/Command/Search.php
```diff
@@ -87,22 +81,20 @@ protected function configure() {
...
+		$message = match (true) {
+			($limit < 0) => 'limit must be 0 or greater',
+			($offset < 0) => 'offset must be 0 or greater',
+			($limit === 0 && $offset !== 0) => 'offset must be 0 if limit is also set to 0',
+			($offset > 0 && ($offset % $limit !== 0)) => 'offset must be a multiple of limit',
+			default => '',
+		};
+
+		if (!empty($message)) {
+			throw new \InvalidArgumentException($message);
```
come-nc, line 97: Could you revert this part?
I really see no point apart from using the shiny new match keyword. It’s harder to read.

apps/user_ldap/lib/Command/UpdateUUID.php
```diff
@@ -49,45 +49,29 @@ class UuidUpdateReport {
...
-		$this->dn = $dn;
-		$this->isUser = $isUser;
-		$this->state = $state;
-		$this->oldUuid = $oldUuid;
-		$this->newUuid = $newUuid;
+	public function __construct(
+		public string $id = '',
+		public string $dn = '',
+		public bool $isUser = true,
+		public int $state = self::UNCHANGED,
```
come-nc, line 56:
```suggestion
		public string $id,
		public string $dn,
		public bool $isUser,
		public int $state,
```
These should stay mandatory parameters

come-nc (commented):
I’m not sure why there is so much hate towards `else` statements, I find things clearer with them that without. Especially the `continue` in the loop instead of a simple `else` looks convoluted.
I like the strong typing, constructor promotion and constant usage in returns though.

Please revert at least the `match (true)`.

fsamapoor:
> I’m not sure why there is so much hate towards `else` statements, I find things clearer with them that without. Especially the `continue` in the loop instead of a simple `else` looks convoluted. I like the strong typing, constructor promotion and constant usage in returns though.
> 
> Please revert at least the `match (true)`.

Thank you for taking the time to review the changes Come. This was one of my older PRs and I appreciate your suggestions.

apps/user_ldap/lib/Command/Search.php
```diff
@@ -87,11 +81,10 @@ protected function configure() {
 
 	/**
 	 * Tests whether the offset and limit options are valid
-	 * @param int $offset
-	 * @param int $limit
+	 *
 	 * @throws \InvalidArgumentException
 	 */
-	protected function validateOffsetAndLimit($offset, $limit) {
+	protected function validateOffsetAndLimit($offset, $limit): void {
```
come-nc, line 87:
```suggestion
	protected function validateOffsetAndLimit(int $offset, int $limit): void {
```

come-nc (approved):

PR #39944, opened 2023-08-17
pass the share to the cache instead of having to ask the storage

icewind1991:
Going through the storage includes storage wrappers so it has a bit more overhead than you might expect.

A [modest improvement](https://blackfire.io/profiles/compare/6741c0d9-f280-482b-a03b-139bda1c25c7/graph) for a small change


AndyScherzinger (approved):

kesselb (approved):
Nice :+1:

PR #39950, opened 2023-08-18
fix: Avoid throwing on folder previews as there are none

juliusknorr:
Otherwise there is an uncaught exception thrown which blocks cypress tests in text.


apps/files/src/components/FileEntry.vue
```diff
@@ -395,6 +395,10 @@ export default Vue.extend({
 			return this.userConfig.crop_image_previews
 		},
 		previewUrl() {
+			if (this.source.type === 'folder') {
```
skjnldsv, line 398:
```suggestion
			import { FileType } from '@nextcloud/files'
			if (this.source.type === FileType.Folder) {
```

skjnldsv (approved):

max-nextcloud (approved):

PR #39952, opened 2023-08-18
[stable27] allow to upgrade from 26.0.x to 27.1.x directly

szaimen:
Do this to prevent upgrade issues if custom update solutions are in place that are not written to handle the case needing to upgrade from 26.0.x to 27.0.x and then to 27.1.x

As discussed with @skjnldsv and @AndyScherzinger

### DIFF
--- a/version.php
+++ b/version.php
@@ -37,6 +37,7 @@ $OC_VersionString = '27.1.0 beta 2';
 
 $OC_VersionCanBeUpgradedFrom = [
 	'nextcloud' => [
+		'26.0' => true,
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):
Not an expert, so +1 for the threshold-only

skjnldsv (dismissed):
Let's wait for @blizzz

nickvergessen (approved):
This should definitely be in.
The updater will always return the latest 27.* to the newest 26
We don't want to make another loop, that is one of the reasons we did a minor.

PR #39955, opened 2023-08-18
chore: use Navigation and Actions from `@nextcloud/files`

skjnldsv:
Needs 
- [x] https://github.com/nextcloud-libraries/nextcloud-files/pull/734
- [x] https://github.com/nextcloud-libraries/nextcloud-files/pull/740


susnux:
@skjnldsv sounds like an import issue? Like if a commonjs projects exports like
```js
module.exports = function() {}
```
and you import like
```js
import fo from 'bar'
```

As there is no default import fo will probably be undefined, the correct import would be:
``js
import * as fo from 'bar'
```

**Review by susnux — APPROVED** [review]

File: apps/files/src/components/FilesListFooter.vue [hunk 1]
```diff
@@ -57,8 +57,7 @@
 </template>
 
 <script lang="ts">
-import { formatFileSize } from '@nextcloud/files'
-import { translate } from '@nextcloud/l10n'
+import { formatFileSize, translate } from '@nextcloud/l10n'
```
nickvergessen, line 60: I'm pretty sure this is not what you meant

skjnldsv: Hum, why didn't it throw....

nickvergessen (approved):
Looks good

PR #39985, opened 2023-08-21
fix: Make sure that handleLogin is wrapped with a session lock

juliusknorr:
There might be cases where multiple requests try to renew a token from cookie data. In this case we should make sure that only one of those runs at the same time, so that after the first generates a new token, the second would just fail and need another retry.

Best to be reviewed as https://github.com/nextcloud/server/pull/39985/files?w=1


nickvergessen (approved):
Worth a try

juliushaertl:
Didn't seem to have an effect :/ Will close then

PR #39987, opened 2023-08-21
Fix predefined status buttons

### LINKED ISSUE #39926: [BITV] Increase contrast on focus state of "Status Message"
Unfortunately, I've noticed another issue. The elements in the "Status Message" list receive a poorly noticeable keyboard focus. This needs to be changed.
CleanShot 2023-08-16 at 20 04 46@2x

_Originally posted by @michaelnissenbaum in https://github.com/nextcloud/server/issues/37011#issuecomment-1681062093_

JuliaKirschenheuter:
* Resolves: https://github.com/nextcloud/server/issues/39926

## Summary

- Add logic for predefined status to be visible.
- Add styles for radio buttons to be visible in checked, active and focus-visible state.
- Styles are same as for online status radio buttons https://github.com/nextcloud/server/pull/39332

| Peek 2023-08-21 14-24 | Peek 2023-08-22 13-16

## Checklist


jancborchardt:
@marcoambrosini what do you think of the style?

apps/user_status/src/components/PredefinedStatusesList.vue
```diff
@@ -30,7 +30,7 @@
 			:icon="status.icon"
 			:message="status.message"
 			:clear-at="status.clearAt"
-			:selected="!isCustomStatus && lastSelected === status.id"
```
ShGKme, line 33: Is was the only usage of `isCustomStatus`. I think, we can now remove this prop at all, and remove passing the value from the parent

ShGKme (approved):

marcoambrosini (commented):
I think that the border is fine, but I think we should put the radio button symbol too @JuliaKirschenheuter 
I'm afraid that we're confusing people a lot with these radio buttons that don't look like radios.

JuliaKirschenheuter:
> I'm afraid that we're confusing people a lot with these radio buttons that don't look like radios.
yes, i'm agree that this buttons are not really radio buttons. But let them be radio buttons for now. We had only an issue  that they are not really recognizable.

> I think that the border is fine, but I think we should put the radio button symbol too
I think then it will be lots of "flashing elements". A11y-wise it is not required.

If border is ok - could we leave it like this for now? All improvements we can do afterwards

marcoambrosini (approved):
@JuliaKirschenheuter let's add a normal border radius and get this in, going forward I think we should add the radio button (which is now `.hidden-visually`) to all radio buttons

Screenshot 2023-08-22 at 19 47 52

JuliaKirschenheuter:
> @JuliaKirschenheuter let's add a normal border radius and get this in

do you mean the same border radius like in "online status"?

marcoambrosini:
> do you mean the same border radius like in "online status"?

Yes @JuliaKirschenheuter `var(--border-radius-large)`

PR #39990, opened 2023-08-21
add some recrusive detection/prevention

icewind1991:
- Set a maximum depth for the number of nested shared storage setups
- Once setup, check if the share source comes from the current share

While we try to prevent recursive shares from being created, they are known to sometimes pass through the checks, attempting to use a recursive share leads to hard to debug out of memory issues


skjnldsv (approved):

skjnldsv:
psalm is valid

nfebe (approved):

PR #39991, opened 2023-08-22
[stable26] fix: always use display name from correct backend

max-nextcloud:
Backport of https://github.com/nextcloud/server/pull/39770


blizzz (approved):

ChristophWurst (approved):

max-nextcloud:
cypress seems to fail like this fairly often on stable26.

PR #39996, opened 2023-08-22
fix(middleware): Fix header injection for bruteforce middleware

nickvergessen:
Calling `setHeaders(getHeaders())` breaks the CSP nonce for unknown reasons So shifting back to old standard practise for now

## Summary

* Open a talk conversation you are not a member off so you get bruteforce throttled
* Load the index.php/apps/spreed
* See console glowing red with weird unrelated errors (CSP is the route cause)

## Checklist

### DIFF
--- a/lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
@@ -130,11 +130,7 @@ class BruteForceMiddleware extends Middleware {
 		}
 
 		if ($this->delaySlept) {
-			$headers = $response->getHeaders();
-			if (!isset($headers['X-Nextcloud-Bruteforce-Throttled'])) {
-				$headers['X-Nextcloud-Bruteforce-Throttled'] = $this->delaySlept . 'ms';
-				$response->setHeaders($headers);
-			}
+			$response->addHeader('X-Nextcloud-Bruteforce-Throttled', $this->delaySlept . 'ms');
 		}
 
 		return parent::afterController($controller, $methodName, $response);


lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
```diff
@@ -130,11 +130,7 @@ public function afterController($controller, $methodName, Response $response) {
 		}
 
 		if ($this->delaySlept) {
-			$headers = $response->getHeaders();
-			if (!isset($headers['X-Nextcloud-Bruteforce-Throttled'])) {
-				$headers['X-Nextcloud-Bruteforce-Throttled'] = $this->delaySlept . 'ms';
-				$response->setHeaders($headers);
-			}
+			$response->addHeader('X-Nextcloud-Bruteforce-Throttled', $this->delaySlept . 'ms');
```
ShGKme, line 133: Check for `if (!isset($headers['X-Nextcloud-Bruteforce-Throttled']))` is not required here?

nickvergessen: No, just setting it again will replace the array entry and we will not save a write either, so not needed anymore

Antreesy (approved):
:seal:

bigcat88 (approved):

lib/private/AppFramework/Middleware/Security/BruteForceMiddleware.php
ShGKme: To better understand PHP/server. Was this `if` required in the previous solution?

nickvergessen: Not required, but when it was already set I didn't want to set all headers again with the same values. Now that we are only writing 1 header anyway it's not worth the checking anymore.

PR #40014, opened 2023-08-23
Create right table view

### LINKED ISSUE #37899: [BITV] 9.1.3.1e/8.1 - The apps table represents a table element and should be implemented as such.
Same issue as https://github.com/nextcloud/server/issues/36921

At the moment, the data is implemented in an unstructured manner using "div" elements. More information on grid implementation can be found at the following URLs: https://sarahmhigley.com/writing/grids-part1/, https://sarahmhigley.com/writing/grids-part2/ or https://www.w3.org/WAI/ARIA/apg/patterns/grid/. 

- [ ] Fix

Screenshot from 2023-04-24 14-34-50

<details>
<summary>Details</summary>

Same as: 
https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-3f0ff0bc0e-v8-n1

</details>

JuliaKirschenheuter:
* Resolves: https://github.com/nextcloud/server/issues/37899

## Summary

Create right table view for list of the apps

| Screenshot from 2023-08-23 18-14-39 | Screenshot from 2023-08-23 18-38-16 |-------------|---------------------------------------------------------------|
| Screenshot from 2023-08-23 18-11-30 | Screenshot from 2023-08-23 18-37-53

## Checklist


apps/settings/src/components/AppList.vue
```diff
@@ -38,7 +38,24 @@
...
+						<th class="app-version">
+							<span class="hidden-visually">{{ t('settings', 'App version') }}</span>
+						</th>
+						<th class="app-level">
+							<span class="hidden-visually">{{ t('settings', 'App level') }}</span>
+						</th>
+						<th class="actions">
+							<span class="hidden-visually">{{ t('settings', 'Actions') }}</span>
+						</th>
+					</tr>
```
artonge, line 58: Is it required to have headers for accessibility? Even if it is hidden?

apps/settings/src/components/AppList.vue
```diff
@@ -48,19 +65,24 @@
...
-						<div class="app-version" />
-						<div class="app-level" />
-						<div class="app-groups" />
-						<div class="actions">
-							&nbsp;
-						</div>
-					</div>
+					<tr :key="bundle.id" class="apps-header">
+						<th class="app-image">
+							<span class="hidden-visually">{{ t('settings', 'App icon') }}</span>
```
artonge, line 73: Same, what is the reason to add it and to keep it hidden?

apps/settings/src/components/AppList.vue
```diff
@@ -48,19 +65,24 @@
...
+						</th>
+						<th><h2>{{ bundle.name }} <input type="button" :value="bundleToggleText(bundle.id)" @click="toggleBundle(bundle.id)"></h2></th>
```
artonge, line 75: Wouldn't something like that be better?

```suggestion
						<th colspan="4"><h2>{{ bundle.name }} <input type="button" :value="bundleToggleText(bundle.id)" @click="toggleBundle(bundle.id)"></h2></th>
```

apps/settings/src/components/AppList.vue
JuliaKirschenheuter: yes, because there is no other possibility for users with assessing technologies to recognize which columns are available

apps/settings/src/components/AppList.vue
JuliaKirschenheuter: i've tried it out and that makes no difference. Actually it is only a name of the bundle available and nothing else.

artonge: Thinking about it further, accessibility wise, won't this be confusing to have the bundle name in the app name column ?

Maybe we should have a table per bundle ? Like we did in timelines for Photos, one list per month.

JuliaKirschenheuter: i get your point. I've checked it before and (actually) multiple headers should be fine https://www.w3.org/WAI/tutorials/tables/multi-level/

JuliaKirschenheuter: i mean it is nicer solution with different tables but multiple should be fine too

JuliaKirschenheuter: @Pytal @susnux what do you mean?

apps/settings/src/components/AppList.vue
ShGKme: They are hidden visually, but visible to screenreader while reading the table can understand what each cell means.

w3: https://www.w3.org/TR/WCAG20-TECHS/H51
> Check that table markup with at least the elements table, tr, th, and td is used.

Example:

https://github.com/nextcloud/server/assets/25978914/92f8876c-0759-4ebd-a84e-7f7037ce1ea5

ShGKme: Maybe simplify titles to?

Icon | Name | Version | ? level ? | Actions
---|---|---|---|---
. | . | . | . | .

apps/settings/src/components/AppList.vue
ShGKme: > Thinking about it further, accessibility wise, won't this be confusing to have the bundle name in the app name column ?
> 
> Maybe we should have a table per bundle ? Like we did in timelines for Photos, one list per month.

I agree, `bundle.name` should not be nixed with column titles.

It should be either another heading level or captions of separated tables. And with bundles I think separated tables looks good, simpler and easier to use by voice (you can go one table to next one).

JuliaKirschenheuter: > Thinking about it further, accessibility wise, won't this be confusing to have the bundle name in the app name column ?

Good point! But there are even more problems `input type="button" ` https://github.com/nextcloud/server/blob/f44a0a6dcc38238bc044072088804d179422b7a8/apps/settings/src/components/AppList.vue#L75 🙈

JuliaKirschenheuter: i will make an implementation like here: example2, Summaries for more complex tables, https://www.w3.org/WAI/tutorials/tables/caption-summary/#:~:text=A%20caption%20functions%20like%20a,primary%20mechanism%20to%20identify%20tables.

JuliaKirschenheuter: done

JuliaKirschenheuter:
Hi @michaelnissenbaum 

what would you say is it an acceptable solution in your eyes? 

Screenshot from 2023-08-25 14-11-50

ShGKme:
There is a problem with multi-level table in the bundles view.

Sub headers like "Hub bundle" all works as headers via screen readers.

So when a user reads a "Calendar" app cell in "Name" column in "Hub bundle" section it reads:
```
Name
Enterprise bundleDownload and enable all
Hub bundleDownload and enable all
Groupware bundleDownload and enable all
Social sharing bundleDownload and enable all
Education EditionDownload and enable all  column 2  clickable  Calendar
```

In theory it must be fixed by `headers` attribute that allows to set what exactly headers has each cell. Unfortunately it is not supported by most of screen readers as far as I know 😭

It could be a good alternative to separate it to different tables having 1 caption per table. But then they will have independent column widths... 🥲. (We also cannot make columns fixed).

The only option I can see here is to not use `<th>` for bundle headers. To make it just a table cell

michaelnissenbaum:
@JuliaKirschenheuter The code snippet you sent me, from my perspective, is not accessible. If we intend to structure the content as a data table, we need meaningful column headers like "Name," "Version," etc., which we don't have in this case. As for the text "Enterprise bundle," it should be treated as a caption - refer to https://www.w3.org/WAI/tutorials/tables/caption-summary/, or a heading. Additionally, please avoid using regular headings ("h") within column headers ("th").

ShGKme:
Hello @michaelnissenbaum!

The complex part is that "Enterprise bundle" is not the table caption, but sub-header, the header of a part of the table. This is why we cannot use `<caption>` here.

image

We tried to make a table according to the w3 tutorial: [Tables with Multi-Level Headers
 / Example 2: Table with three headers related to each data cell](https://www.w3.org/WAI/tutorials/tables/multi-level/)

I tested the table with NVDA screen reader and it works awful even on w3 example. It doesn't support the `headers` attribute, so it always read all the sub-header for each cell.

For example, a cell with an app "Calendar" in a Name column is read as
```
Name
Enterprise bundleDownload and enable all
Hub bundleDownload and enable all
Groupware bundleDownload and enable all
Social sharing bundleDownload and enable all
Education EditionDownload and enable all  column 2  clickable  Calendar
```

I asked one accessibility community and they asked that most of screen readers doesn't support `headers` attribute and multi-header tables in general.

Could we ask you help here? 

What software should we use for testing screen reading? Or is it fine to just follow the standard even if it is not well supported by the screen readers?

Maybe in this specific case, we can make sub-headers in the table not headers but just cells?

JuliaKirschenheuter:
Thank you @michaelnissenbaum!

>  If we intend to structure the content as a data table, we need meaningful column headers like "Name," "Version," etc., which we don't have in this case. 

That part was already implemented, it was just hidden as visually-hidden `<tr>`: `apps-header visually-hidden` This part is not a big deal.

> As for the text "Enterprise bundle," it should be treated as a caption - refer to https://www.w3.org/WAI/tutorials/tables/caption-summary/, or a heading. 

Please seen Grigorii's answer.

> Additionally, please avoid using regular headings ("h") within column headers ("th").

Yes, i will do it (just forgotten because of other difficulties).

JuliaKirschenheuter:
@michaelnissenbaum i'm wondering how it could be possible that examples from https://www.w3.org/WAI/tutorials/tables/multi-level/ are not accessible (massive screen reader problem)

michaelnissenbaum:
@JuliaKirschenheuter Screen readers are software products and therefore not flawless (bug free). It's also important to use the right combination of screen reader and browser and to test with different browser/screen reader combinations. In my experience, the WAI examples are generally fine and work in most cases.

michaelnissenbaum:
@ShGKme I might be misunderstanding, but from my point of view, it could be possible to divide the contents of the page into separate tables with captions or headings like "Enterprise Bundle" and "Hub Bundle." In this case, we no longer have any issues.

JuliaKirschenheuter:
Thank yopu @michaelnissenbaum again!

>  it could be possible to divide the contents of the page into separate tables with captions or headings like "Enterprise Bundle" and "Hub Bundle."

The issue is: 
> "It could be a good alternative to separate it to different tables having 1 caption per table. But then they will have independent column widths... 🥲. (We also cannot make columns fixed)."

Could we implement a table like "Table with three headers related to each data cell" from https://www.w3.org/WAI/tutorials/tables/multi-level/?

michaelnissenbaum:
@JuliaKirschenheuter You can do that. Just make sure the headings are correctly linked. Refer to the description of the test step here: https://ergebnis.bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-1-3-1f-zuordnung-von-tabellenzellen.

JuliaKirschenheuter:
In example2, "Table with three headers related to each data cell", https://www.w3.org/WAI/tutorials/tables/multi-level/ there is wrong implementation of ` <th scope="colgroup">`

right implementation is under: https://html.spec.whatwg.org/multipage/tables.html#attr-th-scope. should be `<th scope=rowgroup>`

Pytal (approved):

ShGKme (approved):

Pytal (approved):

ShGKme:
Fixed ESLint errors with formatting, no noticeable changes.

ShGKme:
Acceptance tests selectors updated according to layout update

Pytal (approved):

PR #40023, opened 2023-08-23
Improve function readability

solracsf:
From switch to match. Less code, more readable.

## Checklist

### DIFF
--- a/lib/private/Log/ErrorHandler.php
+++ b/lib/private/Log/ErrorHandler.php
@@ -94,20 +94,12 @@ class ErrorHandler {
 	}
 
 	private static function errnoToLogLevel(int $errno): int {
-		switch ($errno) {
-			case E_USER_WARNING:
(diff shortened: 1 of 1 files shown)


susnux (approved):

Altahrim (approved):
A lot of them are missing, mainly `E_WARNING` and `E_NOTICE`. 
Does someone knows why?
Here is [the list](https://www.php.net/manual/fr/errorfunc.constants.php)

nickvergessen:
> Does someone knows why?

Because we log them as errors.

> Conflicting files
lib/private/Log/ErrorHandler.php 

https://github.com/nextcloud/server/blob/master/lib/private/Log/ErrorHandler.php#L96-L101

seems like 2 people had this idea.

Altahrim:
> Because we log them as errors.

Yes, logging a warning or strict as error seems strange to me ^^

PR #40026, opened 2023-08-24
feat: Add auth token list and delete commands

### LINKED ISSUE #39690: Add `occ user:app-passwords <uid>` command
### How to use GitHub

As an admin of a Nextcloud instance I want to be able to list the app passwords of a user so that I can see if they have old data that could be cleared.

Add an occ command `user:app-passwords <uid>` that exports all app tokens with their id, name and creation time. Ideally there should be an option for json or other structured output so the data can be used in a script.

N/a

**Additional context**

There is an `occ user:add-app-password` command

### LINKED ISSUE #39691: Add `occ user:delete-app-password` command
### How to use GitHub

As an admin of a Nextcloud instance I want to be able to delete app passwords and web sessions of my users.

Add an occ command `user:delete-app-password`.
The command could take the database ID as identification, paired with https://github.com/nextcloud/server/issues/39690, or allow clearing app tokens older than a given timestamp or duration.

N/a

**Additional context**

Best enjoyed with https://github.com/nextcloud/server/issues/39690

lhsazevedo:
:wave: Hi, this is my first contribution :)

* Resolves #39690
* Resolves #39691

## Summary
Add two new commands requested in #39690 and #39691.

Since both commands can handle app passwords as well as session tokens, I've named them `occ user:auth-tokens` and `occ user:delete-auth-token` instead to better align with the nouns used in the codebase. Let me know if the requested names should be kept.

## Sample output:
```
$ ./occ user:auth-tokens admin
  - 0:
    - id: 8
    - name: Mozilla/5.0 (X11; Linux x86_64; rv:107.0) Gecko/20100101 Firefox/107.0
    - lastActivity: 1692835829
    - type: 0
    - scope:
      - filesystem: true
  - 1:
    - id: 9
    - name: my-app-password
    - lastActivity: 1692835843
    - type: 1
    - scope:
      - filesystem: true

$ ./occ user:delete-auth-token 9
$ ./occ user:auth-tokens --output=json_pretty admin
[
    {
        "id": 8,
        "name": "Mozilla\/5.0 (X11; Linux x86_64; rv:107.0) Gecko\/20100101 Firefox\/107.0",
        "lastActivity": 1692845431,
        "type": 0,
        "scope": {
            "filesystem": true
        }
    }
]
```

## Checklist


lhsazevedo:
I'll open the documentation PR as soon as we agree on this one!

Also, while implementing the commands I noticed some possible minor follow ups:

1. #39691 also mentions the possibility of deleting tokens older than a given timestamp or age. I think this could be done is by adding `--before` and `--age` options, either in the introduced `delete-auth-token` command (by making `<id>` optional) or in a new one to avoid confusion. 
2. `OC\Core\Command\Base::configure()`:
   1. A few commands are manually setting up the `--output` option instead of relying on the parent `OC\Core\Command\Base::configure()` method for this.
      - core/Command/Group/Info.php
      - core/Command/Group/ListCommand.php
      - core/Command/User/Info.php
      - core/Command/User/ListCommand.php
   2. Despite not allowing control over the output format, some commands have the `--output` option because they are calling `parent::configure()`:
      - core/Command/App/GetPath.php
      - core/Command/TwoFactorAuth/Enable.php
      - core/Command/TwoFactorAuth/Disable.php
      - core/Command/TwoFactorAuth/Cleanup.php
      - core/Command/TwoFactorAuth/State.php
      - core/Command/Config/App/DeleteConfig.php
      - core/Command/Config/App/SetConfig.php
      - core/Command/Config/System/DeleteConfig.php
      - core/Command/Config/System/SetConfig.php
   
   I guess that both of the above happened because it isn't clear that `OC\Core\Command\Base::configure()` only sets up the `output` option. What do you think of renaming it to `setupOutputFormatOption()`?.
3. There now three auth token related commands in `core/Commands/User`. Maybe we could group them inside a nested `AuthToken` namespace?

Is there anything in the list that you think should be done?

nickvergessen:
> 1. https://github.com/nextcloud/server/issues/39691 also mentions the possibility of deleting tokens older than a given timestamp or age. I think this could be done is by adding --before and --age options, either in the introduced delete-auth-token command (by making <id> optional) or in a new one to avoid confusion.

I would make id optional and go with a `--before` parameter as you suggested

2. `configure()` is a "magic" predefined function of the Symfony console, so would like to keep that (also it would break apps that use it). Calling `parent::configure()` is the prefered way and if the used options are the same, those not using it should adjusted in the future.

3. Makes sense to me

core/Command/User/AuthTokens.php
```diff
@@ -0,0 +1,70 @@
...
+		protected IProvider $tokenProvider,
+	) {
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		parent::configure();
+
+		$this
+			->setName('user:auth-tokens')
```
nickvergessen, line 45: How about namespacing them as well?
```suggestion
			->setName('user:auth-tokens:list')
```

and rename the others to
```
			->setName('user:auth-tokens:delete')
```
```
			->setName('user:auth-tokens:add')
			->setAliases(['user:add-app-password'])
```

core/Command/User/AuthTokens.php
```diff
@@ -0,0 +1,70 @@
...
+		$user = $this->userManager->get($input->getArgument('user'));
+
+		if (is_null($user)) {
+			$output->writeln('<error>user not found</error>');
+			return 1;
+		}
+
+		$tokens = $this->tokenProvider->getTokenByUser($user->getUID());
+
+		$data = array_map(fn (IToken $token): mixed => $token->jsonSerialize(), $tokens);
```
nickvergessen, line 64: I think we should filter out some columns here:
- password
- password_hash
- token
- public_key
- private_key

core/register_command.php
```diff
@@ -193,6 +193,8 @@
 	$application->add(new OC\Core\Command\User\ListCommand(\OC::$server->getUserManager(), \OC::$server->getGroupManager()));
 	$application->add(new OC\Core\Command\User\Info(\OC::$server->getUserManager(), \OC::$server->getGroupManager()));
 	$application->add(new OC\Core\Command\User\AddAppPassword(\OC::$server->get(\OCP\IUserManager::class), \OC::$server->get(\OC\Authentication\Token\IProvider::class), \OC::$server->get(\OCP\Security\ISecureRandom::class), \OC::$server->get(\OCP\EventDispatcher\IEventDispatcher::class)));
+	$application->add(new OC\Core\Command\User\AuthTokens(\OC::$server->get(\OCP\IUserManager::class), \OC::$server->get(\OC\Authentication\Token\IProvider::class)));
+	$application->add(new OC\Core\Command\User\DeleteAuthToken(\OC::$server->get(\OC\Authentication\Token\IProvider::class)));
```
nickvergessen, line 197: Since all parameters are DI-able we can also just use:
```suggestion
	$application->add(\OC::$server->get(\OC\Core\Command\User\AuthTokens::class));
	$application->add(\OC::$server->get(\OC\Core\Command\User\DeleteAuthToken::class));
```

nickvergessen:
> I would make id optional and go with a --before parameter as you suggested

That being said, with `--before` you definetly want to limit it to a user. So maybe the delete should take the userId as first argument as some insurance if you delete by id and to allow `--before` being scoped to a user?

nickvergessen:
Otherwise that's a nice and handy PR :)

core/Command/User/AuthTokens.php
nickvergessen, line 1: To fix CI:
> Please run: bash build/autoloaderchecker.sh
> And commit the result

core/Command/User/AuthTokens.php
```diff
@@ -0,0 +1,70 @@
...
+
+		$this->writeArrayInOutputFormat($input, $output, $data);
```
tcitworld, line 66: **suggestion:** Maybe `writeTableInOutputFormat` is more suited?

core/Command/User/AuthTokens.php
lhsazevedo: Done. For the record, `PublicKeyToken::jsonSerialize()` already explicitly lists allowed fields, ao we are adding a second guard here to be on the safe side.

core/register_command.php
lhsazevedo: Oh I see, nice

core/Command/User/AuthTokens.php
lhsazevedo: Sounds great, the alias was a good idea as well. Done.

core/Command/User/AuthTokens.php
lhsazevedo: Thanks

nickvergessen:
Let's use `--last-used-before`, then it should be clear and we can add the other one later if needed?

core/Command/User/AuthTokens.php
lhsazevedo: The only block I found here is that the scope field is an array (decoded JSON), and `writeTableInOutputFormat` expects fields to be scalars (or objects implementing `__toString()`). I was only able to find references to the `filesystem` scope, but there could be more in the future and I'm not sure if it would fit nicely in a table cell. We might as well omit the scope field when using the plain output format (table), or proceed with the simpler `writeArrayInOutputFormat`...

What do you think?

lhsazevedo:
Done, thanks for reviewing! I could squash all the commits but that might make reviewing harder, so feel free to squash when merging...

core/Command/User/AuthTokens.php
tcitworld: As long as the array stays kinda the same we're safe.

A possibility would be to `array_filter` the array for `true` values, get only the keys (`array_keys`) from what's left and `implode` that to a string.

In any case, adding a test on the output should make sure format changes don't break anything.

core/Command/User/AuthTokens/Delete.php
```diff
@@ -0,0 +1,120 @@
...
+				'last-used-before',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'Delete tokens last used before a given date.'
+			);
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$uid = $input->getArgument('uid');
+		$id = $input->getArgument('id');
```
nickvergessen, line 66:
```suggestion
		$id = (int) $input->getArgument('id');
```

core/Command/User/AuthTokens/Delete.php
```diff
@@ -0,0 +1,120 @@
...
+		return $this->deleteById($uid, $id);
+	}
+
+	protected function deleteById(string $uid, string $id) {
+		$this->tokenProvider->invalidateTokenById($uid, $id);
+
+		return Command::SUCCESS;
+	}
+
+	protected function deleteLastUsedBefore(string $uid, string $before) {
```
nickvergessen, line 89:
```suggestion
	protected function deleteLastUsedBefore(string $uid, string $before): int {
```

lib/private/Authentication/Token/PublicKeyTokenMapper.php
```diff
@@ -69,6 +69,15 @@ public function invalidateOld(int $olderThan, int $remember = IToken::DO_NOT_REM
 			->execute();
 	}
 
+	public function invalidateLastUsedBefore(string $uid, int $before): int {
+		$qb = $this->db->getQueryBuilder();
+		return $qb->delete($this->tableName)
+			->where($qb->expr()->eq('uid', $qb->createNamedParameter($uid)))
+			->andWhere($qb->expr()->lt('last_activity', $qb->createNamedParameter($before, IQueryBuilder::PARAM_INT)))
+			->andWhere($qb->expr()->eq('version', $qb->createNamedParameter(PublicKeyToken::VERSION, IQueryBuilder::PARAM_INT)))
+			->executeStatement();
```
nickvergessen, line 78:
```suggestion
		$qb->delete($this->tableName)
			->where($qb->expr()->eq('uid', $qb->createNamedParameter($uid)))
			->andWhere($qb->expr()->lt('last_activity', $qb->createNamedParameter($before, IQueryBuilder::PARAM_INT)))
			->andWhere($qb->expr()->eq('version', $qb->createNamedParameter(PublicKeyToken::VERSION, IQueryBuilder::PARAM_INT)));
		return $query->executeStatement();
```

Let's split type changing calls to a new expression

nickvergessen (commented):
Almost there, let's make Psalm a bit more happy

nickvergessen:
> I could squash all the commits but that might make reviewing a little harder, feel free to do so when merging.

Squashing after reviews are done is what we mostly do 👍🏼

core/Command/User/AuthTokens.php
lhsazevedo: Here's the result

image
~chef's kiss

This is done only when the output format is `OUTPUT_FORMAT_PLAIN`. Otherwise, we keep the array to be consumed by scripts.

tcitworld: @lhsazevedo Could you format `lastActivity` to an ISO8601 date string and map [the `type` constants](https://github.com/nextcloud/server/blob/b570d76355c9c463c7ded49ac384a2aaa5ba36a4/lib/private/Authentication/Token/IToken.php#L32-L34C20) to strings?

lhsazevedo: Sure

lhsazevedo: @tcitworld Done, I took the opportunity to extract the plain output code to a method.

```
$ occ user:auth-tokens:list --no-ansi admin
+----+------------------------------------------------------------------------+---------------------------+-----------+------------+
| id | name                                                                   | lastActivity              | type      | scope      |
+----+------------------------------------------------------------------------+---------------------------+-----------+------------+
| 3  | cli                                                                    | 2023-08-28T01:32:34+00:00 | permanent | filesystem |
| 6  | Mozilla/5.0 (X11; Linux x86_64; rv:107.0) Gecko/20100101 Firefox/107.0 | 2023-08-28T01:47:08+00:00 | temporary | filesystem |
| 7  | my-test-app                                                            | 2023-08-28T01:47:40+00:00 | permanent | filesystem |
+----+------------------------------------------------------------------------+---------------------------+-----------+------------+
```

nickvergessen (approved):

tcitworld (approved):

lhsazevedo:
Conflict resolved

nickvergessen:
Awesome story @lhsazevedo I invited you to the org

lhsazevedo:
Thanks for inviting me @nickvergessen, just joined it. I'm looking forward to contribute more with the project.

PR #40042, opened 2023-08-25
(enh) Change `occ background-job:list` limit default 10->500

joshtrichards:
## Summary

I'd been wondering for awhile why not all background jobs would show up under `occ background-job:list` even though all jobs were clearly being executed. Spent an embarrassing amount of time reviewing the code on this... only to realize there was a default limit of `10` in the command. :rofl: 

I understand why there is a limit (large environments) - and it can be easily overridden - but I feel 10 is  bit _too_ conservative for a default. Being this low means the job list is nearly always incomplete in the default execution mode, which makes the overriding of the default nearly a 100% certainty in day to day use.

Nearly all environments >10 and many (most?) <100 jobs. This changes the default to 100 so fewer have to feel as silly as me. :smiley:  And a lot more people get all their jobs shown by default. At the same time, this higher default still remains conservative enough to avoid problems in really large environments.

## TODO

- [ ] ...

## Checklist

- ~~[ ] Screenshots before/after for front-end changes~~
- ~~[ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)~~

### DIFF
--- a/core/Command/Background/ListCommand.php
+++ b/core/Command/Background/ListCommand.php
@@ -66,11 +66,12 @@ class ListCommand extends Base {
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output): int {
-		$jobs = $this->jobList->getJobsIterator($input->getOption('class'), (int)$input->getOption('limit'), (int)$input->getOption('offset'));
-		if ($input->getOption('output') === self::OUTPUT_FORMAT_PLAIN) {
-			$output->writeln("Output is currently limited to first " .  $input->getOption('limit') . " jobs. Specify `-l, --limit[=LIMIT]` to override.");
+		$limit = (int)$input->getOption('limit');
+		$jobsInfo = $this->formatJobs($this->jobList->getJobsIterator($input->getOption('class'), $limit, (int)$input->getOption('offset')));
+		$this->writeTableInOutputFormat($input, $output, $jobsInfo);
+		if ($input->getOption('output') === self::OUTPUT_FORMAT_PLAIN && count($jobsInfo) >= $limit) {
+			$output->writeln("\n<comment>Output is currently limited to first " .  $limit . " jobs. Specify `-l, --limit[=LIMIT]` to override.</comment>");
 		}
-		$this->writeTableInOutputFormat($input, $output, $this->formatJobs($jobs));
 		return 0;
 	}
 
--- a/core/Command/Background/ListCommand.php
+++ b/core/Command/Background/ListCommand.php
@@ -53,7 +53,7 @@ class ListCommand extends Base {
 				'l',
 				InputOption::VALUE_OPTIONAL,
 				'Number of jobs to retrieve',
-				'100'
+				'500'
 			)->addOption(
 				'offset',
 				'o',
--- a/core/Command/Background/ListCommand.php
+++ b/core/Command/Background/ListCommand.php
@@ -67,7 +67,9 @@ class ListCommand extends Base {
 
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		$jobs = $this->jobList->getJobsIterator($input->getOption('class'), (int)$input->getOption('limit'), (int)$input->getOption('offset'));
-		$output->writeln("Output is currently limited to first " .  $input->getOption('limit') . " jobs. Specify `-l, --limit[=LIMIT]` to override.");
+		if ($input->getOption('output') === self::OUTPUT_FORMAT_PLAIN) {
+			$output->writeln("Output is currently limited to first " .  $input->getOption('limit') . " jobs. Specify `-l, --limit[=LIMIT]` to override.");
+		}
 		$this->writeTableInOutputFormat($input, $output, $this->formatJobs($jobs));
 		return 0;
 	}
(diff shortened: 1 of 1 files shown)


solracsf:
Another possibility would be to have an `-l` option like `user:list` has.
```
-l, --limit[=LIMIT]    Number of background jobs to retrieve [default: "10"]
```

joshtrichards:
> Why just dont `occ background-job:list -l 100` ? 🤔
> 
> ```
> -l, --limit[=LIMIT]    Number of jobs to retrieve [default: "10"]
> ```

Yes, the parameter works. I just think `10` is an overly low default. It'll almost never show all the jobs, even in tiny installations. Effectively the operator has to know to *always* specify a limit if they actually want to see an accurate job list. This nullifies the `10` as a useful default.

It's may seem a nitpick, but it also creates more mental load for what seems no net benefit from what I can tell.

SystemKeeper:
Maybe it makes sense to add a note to the output like „Showing the first X jobs“, or „Output is limited to X jobs“ to indicate that there’s a limit active.

come-nc (approved):
Maybe an indication in the output if the list is complete or not would indeed be also good.

joshtrichards:
Done. Though, technically, it breaks json output format...

core/Command/Background/ListCommand.php
```diff
@@ -67,6 +67,7 @@ protected function configure(): void {
 
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		$jobs = $this->jobList->getJobsIterator($input->getOption('class'), (int)$input->getOption('limit'), (int)$input->getOption('offset'));
+		$output->writeln("Output is currently limited to first " .  $input->getOption('limit') . " jobs. Specify `-l, --limit[=LIMIT]` to override.");
```
come-nc, line 70: Can you show what it looks like when json output is used then?
I know some commands have a conditional on output format to hide this stuff is needed.

Also, it would be nicer if the line does not show when there are less jobs than the limit.

mejo- (approved):

core/Command/Background/ListCommand.php
kesselb: The result is an invald json: 

image

kesselb: I assume if you pick json or json_pretty that you would like to work with the result and therefore suggest not applying a limit if possible and printing all jobs.

kesselb, line 70:
```suggestion
		if ($input->getOption('output') === self::OUTPUT_FORMAT_PLAIN) {
			$output->writeln("Output is currently limited to first " .  $input->getOption('limit') . " jobs. Specify `-l, --limit[=LIMIT]` to override.");
		}
```

To output the warning only for plain.

kesselb (commented):
Good idea :+1:

core/Command/Background/ListCommand.php
mejo-: Looks like a good solution for now. Let's get this in, no?

come-nc (approved):

kesselb (approved):

core/Command/Background/ListCommand.php
```diff
@@ -67,7 +67,11 @@ protected function configure(): void {
 
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		$jobs = $this->jobList->getJobsIterator($input->getOption('class'), (int)$input->getOption('limit'), (int)$input->getOption('offset'));
-		$this->writeTableInOutputFormat($input, $output, $this->formatJobs($jobs));
+		$jobsInfo = $this->formatJobs($jobs);
+		$this->writeTableInOutputFormat($input, $output, $jobsInfo);
+		if ($input->getOption('output') === self::OUTPUT_FORMAT_PLAIN && count($jobsInfo) >= (int)$input->getOption('limit')) {
+			$output->writeln("\n<comment>Output is currently limited to first " .  $input->getOption('limit') . " jobs. Specify `-l, --limit[=LIMIT]` to override.</comment>");
```
nickvergessen, line 73: Depending on the offset it's not "first" 😅


```suggestion
			$output->writeln("\n<comment>Output is currently limited to " .  $input->getOption('limit') . " jobs. Specify `-l, --limit[=LIMIT]` to override.</comment>");
```

core/Command/Background/ListCommand.php
```diff
@@ -66,8 +66,12 @@ protected function configure(): void {
 	}
 
 	protected function execute(InputInterface $input, OutputInterface $output): int {
-		$jobs = $this->jobList->getJobsIterator($input->getOption('class'), (int)$input->getOption('limit'), (int)$input->getOption('offset'));
-		$this->writeTableInOutputFormat($input, $output, $this->formatJobs($jobs));
+		$limit = (int)$input->getOption('limit');
+		$jobsInfo = $this->formatJobs($this->jobList->getJobsIterator($input->getOption('class'), $limit, (int)$input->getOption('offset')));
+		$this->writeTableInOutputFormat($input, $output, $jobsInfo);
+		if ($input->getOption('output') === self::OUTPUT_FORMAT_PLAIN && count($jobsInfo) >= $limit) {
+			$output->writeln("\n<comment>Output is currently limited to first " .  $limit . " jobs. Specify `-l, --limit[=LIMIT]` to override.</comment>");
```
come-nc, line 73:
```suggestion
			$output->writeln("\n<comment>Output is currently limited to " .  $limit . " jobs. Specify `-l, --limit[=LIMIT]` to override.</comment>");
```
Missed it :see_no_evil:

nickvergessen (approved):

Issue #39914, opened 2023-08-16
[f2v] Final overview


skjnldsv:
## Main modules
- [x] Files: #39808 
- [x] Systemtags: https://github.com/nextcloud/server/pull/40475

### Submodules
- [x] File upload + New files/folder menu: https://github.com/nextcloud/server/pull/39945
- [x] Breadcrumbs sharing icon https://github.com/nextcloud/server/pull/40192
- [x] Drag and drop + Move/Copy action #39998 
- [x] Sort favorites first option #37946
- [x] Grid view  https://github.com/nextcloud/server/pull/40917
- [ ] Search
  - closes https://github.com/nextcloud/server/issues/1274
  - closes https://github.com/nextcloud/server/issues/38312
- [x] Folder icon overlay https://github.com/nextcloud/server/pull/40209
- [x] Dropzone https://github.com/nextcloud/server/pull/40674

### Actions
- [ ] Support nested actions
  - closes https://github.com/nextcloud/server/issues/39718
- [x] Inline tag display https://github.com/nextcloud/server/pull/40284
- [x] Inline comments action https://github.com/nextcloud/server/pull/40409
- [ ] Inline locked state https://github.com/nextcloud/files_lock/issues/154
- [x] Inline share action https://github.com/nextcloud/server/pull/40517
- [x] Template picker https://github.com/nextcloud/server/pull/40854
- [ ] Disconnect/leave share delete node: https://github.com/nextcloud/server/issues/40940 https://github.com/nextcloud/server/blob/dd3cd29224cd93c6168cd42e93e68cdd6ac5e921/apps/files/js/fileactions.js#L787-L791

#### Undecided / later
- [ ] Files public (the public files app is behaving very differently and needs dedicated separate work) 
- [ ] My files view
  - closes #9528

## Regressions

- #40514 

## Testing
- [ ] Acceptance cypress

## Follow-ups
- [x] Dive into a potential perf boost on the virtual scrolling with some pseudo-code key mapping https://github.com/nextcloud/server/pull/40917
- [ ] Implement keyboard navigation and shortcut page: https://github.com/nextcloud/server/issues/30484
- [x] Deprecate files_rightclick
- [ ] Remove `/apps/files/js`
- [ ] Photos previews get outdated after an edit from the viewer https://github.com/nextcloud/server/issues/40230

## Apps customizations
- [x] Disabled style for end-to-end encrypted folders

artonge:
[...]

PR #40065, opened 2023-08-27
fix(files): do not load legacy files app js

max-nextcloud:
`OCP.Files.Navigation` is undefined.

* Resolves: failures like https://cloud.cypress.io/projects/hx9gqy/runs/11854/overview?roarHideRunsWithDiffGroupsAndTags=1

cypress-screenshot

## Checklist

- [x] Tests are fixed by this
- [x] Screenshot of test failurs
- [x] Documentation is not required
- [x] Backports are not required


skjnldsv:
Where is legacy files still loaded?
Those scripts shuold not even be here anymore :)

max-nextcloud:
uh... interesting - i will investigate.

max-nextcloud:
Loaded directly from the html inside the cypress tests. I will see if it also loads without cypress...

grafik

max-nextcloud:
It's served with the initial template when loading `apps/files`:
grafik

skjnldsv:
Ah, found
This ain't supposed to still be there
https://github.com/nextcloud/server/blob/3846ec2899d476c03765859487bce4a8ccd0e799/apps/files/js/app.js#L403

skjnldsv:
It's a multi step, but requires a bit more:
 - Disable [this line](https://github.com/nextcloud/server/blob/be41dc4fa79310328647da48c3426a8eaed75f29/apps/files/lib/Controller/ViewController.php#L221)
 - Remove [systemtagsfilelist](https://github.com/nextcloud/server/blob/0fca8238dd86d572f0f15eaa8e8a5ea836cc76bf/apps/systemtags/src/systemtags.js#L25) 
 - Fix Sidebar to not rely on `OCA.Files.App.fileList.filesClient`
  https://github.com/nextcloud/server/blob/be41dc4fa79310328647da48c3426a8eaed75f29/apps/files/src/services/FileInfo.js#L36-L39

Then it should be good, public pages seems to still work, I gues they load their own minimal stuff

max-nextcloud:
> This ain't supposed to still be there
> 
> https://github.com/nextcloud/server/blob/3846ec2899d476c03765859487bce4a8ccd0e799/apps/files/js/app.js#L403

So can i also drop `initialize()` entirely and `navigation` and `fileList` as they are only written to in `initialize()`?

skjnldsv:
if we don't include the file, dropping the initialize will not matter.
I think we can keep those files here until  https://github.com/nextcloud/server/issues/39914 is complete and we do a cleanup pass

max-nextcloud:
I think I got it... or at least some intermediate state.

Circles is missing the filesList though:
grafik

skjnldsv:
> Circles is missing the filesList though:

Circles is not upgraded for 28 though :thinking: 
I think that's a different topic anyway

skjnldsv (approved):

apps/files/lib/Controller/ViewController.php
```diff
@@ -218,7 +218,7 @@ public function index($dir = '', $view = '', $fileid = null, $fileNotFound = fal
 
 		// Load the files we need
 		\OCP\Util::addStyle('files', 'merged');
-		\OCP\Util::addScript('files', 'merged-index', 'files');
+		\OCP\Util::addScript('files', 'fileactions');
```
max-nextcloud, line 221: without the fileactions import files won't open anymore in the viewer. So I guess we still need it.

@skjnldsv - what do you think?

skjnldsv: Tried latest Viewer?
We don't use the old `fileactions` ?

max-nextcloud: Yes... latest viewer, disabled contacts app which was causing errors and early return from circles app so it does not interfere. Still `.md` `.pdf` and `.jpg`s have no view action and clicking files triggers a download. Did not try other file types.

skjnldsv:
Is it still relevant?

max-nextcloud:
> Is it still relevant?

I think so. Cleaned it up and the main thing now is not to `addScript('files', 'merged-index', 'files');`.

skjnldsv (approved):

susnux (approved):

PR #40109, opened 2023-08-29
clear sftp stat cache when opening a write stream

icewind1991:
the writable stream does some lower level stuff that bypasses the normal stat cache clear of the library


icewind1991:
This code is merged already

Issue #37090, opened 2023-03-07
[BITV] 9.3.3.2/8.2 - The "Add group" input field does not have a visual label, it only has a placeholder element. (1)


AndyScherzinger:
see also https://github.com/nextcloud/server/issues/37091

- [x] https://github.com/nextcloud-libraries/nextcloud-vue/pull/4494
- [x] Release
- [x] https://github.com/nextcloud/server/pull/40692
- [x] https://github.com/nextcloud/server/pull/40110

<img width="291" alt="61c68544e79ec5b016e35a21f21e15a7" src="https://user-images.githubusercontent.com/1315170/223564162-159c6335-14e1-4647-8448-84ea8c396c09.png">

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-9f92015c98-v8-n2

</details>

Pytal:
Would need a redesign of NcAppNavigationNewItem

![image](https://github.com/nextcloud/server/assets/24800714/0d58bff7-3a1e-4496-8037-91fc6393dbe0)

Could be NcActions with NcActionInput inside and a visible label @nimishavijay?

JuliaKirschenheuter:
@marcoambrosini?

marcoambrosini:
Agree with @Pytal, we could use `NcAppnavigationCaption` which comes with an `NcActionMenu` on the right. 
This is exactly what we do for contacts groups.


<img width="391" alt="Screenshot 2023-08-29 at 10 49 50" src="https://github.com/nextcloud/server/assets/26852655/b2206e1e-55f8-418a-98eb-5c27b06dc99b">


And we should probably deprecate the NcAppNavigationNew component. cc @skjnldsv what are your thoughts on that?

skjnldsv:
> cc @skjnldsv what are your thoughts on that?

None :)
As long as the component code and logic make sense, it doesn't trubles me. 
Nonetheless, using an `NcAppnavigationCaption` seems counter intuitive. SO maybe there can just be a bit of change in the `NcAppNavigationNew` to have it use the `NcAppnavigationCaption` :shrug:

Pytal:
> None :) As long as the component code and logic make sense, it doesn't trubles me. Nonetheless, using an `NcAppnavigationCaption` seems counter intuitive. SO maybe there can just be a bit of change in the `NcAppNavigationNew` to have it use the `NcAppnavigationCaption` 🤷

[...]

PR #40110, opened 2023-08-30
enh(settings): Add and remove groups accessibly

Pytal:
- Base branch (merge first) https://github.com/nextcloud/server/pull/40858
- For https://github.com/nextcloud/server/issues/37090

## Summary

Before | After
--- | ---
image | image
image | image

## TODO

- [x] Tests

## Checklist

- Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)


marcoambrosini (commented):
@Pytal I would still prefer to have a heading in the menu that states what's the current operation that the user is undergoing, similar to contacts. 

Something like 
heading of the tooltip "Create group"
Label of the input: "Group name"

Pytal:
> @Pytal I would still prefer to have a heading in the menu that states what's the current operation that the user is undergoing, similar to contacts.
> 
> Something like heading of the tooltip "Create group" Label of the input: "Group name"

Done!

Also moved the actions next to the "Groups" navigation caption

apps/settings/src/views/Users.vue
```diff
@@ -284,42 +299,31 @@ export default {
...
 		 * Create a new group
 		 *
-		 * @param {string} gid The group id
+		 * @param {SubmitEvent} event The event
 		 */
-		async createGroup(gid) {
-			// group is not valid
-			if (gid.trim() === '') {
+		async createGroup(event) {
+			this.$refs.navigationCaption.$children[0].opened = false
```
ShGKme, line 305: IMO, this is very not a Vue way and explicit to manipulate `NcAction` by `$refs.navigationCaption.$children[0]`. It's very easy to break by change `NcAppNavigationCaption`.

Also, `$children` is removed in Vue and will break during Vue 3 migration. See: https://v3-migration.vuejs.org/breaking-changes/children.html

ShGKme: As an alternative way here I can propose to add new data property like `isAddGroupOpen` and use `open` prop of `NcAction`. We can pass it directly on `NcAppNavigationCaption`, it is proxied. Then instead of

```js
this.$refs.navigationCaption.$children[0].opened = false
```

we can pass `false` to this property:

```js
this.isAddGroupOpen = false
```

apps/settings/src/views/Users.vue
```diff
@@ -358,11 +362,6 @@ export default {
 	max-height: 100%;
 }
 
-// force hiding the editing action for the add group entry
-.app-navigation__list #addgroup::v-deep .app-navigation-entry__utils {
-	display: none;
-}
-
```
ShGKme, line 365: Thank for cleaning up the component and removing unused code!

ShGKme (changes requested):
Works and looks good. But we should avoid using `$children` when possible.

apps/settings/src/views/Users.vue
Pytal: Will also need to proxy listeners with https://github.com/nextcloud-libraries/nextcloud-vue/pull/4494 too

apps/settings/src/views/Users.vue
```diff
@@ -92,7 +81,32 @@
 						</template>
 					</NcAppNavigationItem>
 
-					<NcAppNavigationCaption v-if="groupList.length > 0" :name="t('settings', 'Groups')" />
+					<!-- TODO remove class -->
+					<NcAppNavigationCaption class="caption"
+						:name="t('settings', 'Groups')"
+						:disabled="loadingAddGroup"
+						:aria-label="loadingAddGroup ? t('settings', 'Creating group…') : t('settings', 'Create group')"
```
ShGKme, line 88: I'm not sure, but ше seems that we usually add a space before `…`, don't we?

```suggestion
						:aria-label="loadingAddGroup ? t('settings', 'Creating group …') : t('settings', 'Create group')"
```

apps/settings/src/views/Users.vue
```diff
@@ -92,7 +81,32 @@
...
+								<template #icon>
+									<AccountGroup :size="20" />
+								</template>
+								{{ t('settings', 'Create group') }}
+							</NcActionText>
+							<NcActionInput :label="t('settings', 'Group name')"
+								:label-outside="false"
+								:disabled="loadingAddGroup"
+								:value.sync="newGroupName"
+								@submit="createGroup" />
```
ShGKme, line 106: `trailing-button-label` by default has `aria-label="Clear text"`, but maybe we should fix it on the `NcActionInput` side

apps/settings/src/views/Users.vue
```diff
@@ -287,43 +308,29 @@ export default {
...
-		 * @param {string} gid The group id
 		 */
-		async createGroup(gid) {
-			// group is not valid
-			if (gid.trim() === '') {
+		async createGroup() {
+			this.isAddGroupOpen = false
+			const groupId = this.newGroupName.trim()
+			if (groupId === '') {
+				showError(t('settings', 'Please enter a valid group name'))
```
ShGKme, line 316: Can we show it in a `hint-text` of the input?

ShGKme (approved):

apps/settings/src/views/Users.vue
Pytal: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4632

marcoambrosini (approved):

JuliaKirschenheuter (approved):

ShGKme (approved):

PR #40121, opened 2023-08-30
Refactor `OC\Server::getDatabaseConnection`

summersab:
This PR refactors the deprecated method `OC\Server::getDatabaseConnection` and replaces it with `OC\Server::get(\OCP\IDBConnection::class)` throughout the entire NC codebase (excluding `./apps` and `./3rdparty`).

Additionally, where necessary, the `\OCP\IDBConnection` class is imported via the `use` directive.


core/register_command.php
```diff
@@ -107,7 +109,7 @@
 	$application->add(\OC::$server->get(OC\Core\Command\Info\Space::class));
 
 	$application->add(new OC\Core\Command\Db\ConvertType(\OC::$server->getConfig(), new \OC\DB\ConnectionFactory(\OC::$server->getSystemConfig())));
-	$application->add(new OC\Core\Command\Db\ConvertMysqlToMB4(\OC::$server->getConfig(), \OC::$server->getDatabaseConnection(), \OC::$server->getURLGenerator(), \OC::$server->get(LoggerInterface::class)));
+	$application->add(new OC\Core\Command\Db\ConvertMysqlToMB4(\OC::$server->getConfig(), \OC::$server->get(IDBConnection::class), \OC::$server->getURLGenerator(), \OC::$server->get(LoggerInterface::class)));
```
nickvergessen, line 112:
```suggestion
	$application->add(\OCP\Server::get(\OC\Core\Command\Db\ConvertMysqlToMB4::class));
```

lib/private/Repair.php
```diff
@@ -189,16 +190,16 @@ public static function getRepairSteps(): array {
 				\OC::$server->getConfig()
 			),
 			new MigrateOauthTables(\OC::$server->get(Connection::class)),
-			new FixMountStorages(\OC::$server->getDatabaseConnection()),
-			new UpdateLanguageCodes(\OC::$server->getDatabaseConnection(), \OC::$server->getConfig()),
+			new FixMountStorages(\OC::$server->get(IDBConnection::class)),
```
nickvergessen, line 193:
```suggestion
			\OCP\Server::get(FixMountStorages::class),
```

core/register_command.php
summersab: I would say that this is out of scope, but I have no complaints - makes sense to me!

lib/private/Repair.php
summersab: I would say that this is out of scope, but I have no complaints - makes sense to me!

core/register_command.php
nickvergessen: Well it will make 3 other PRs to touch the same line obsolete.
You could even send a PR and fix this `core/register_command.php` file completely, (or you push it into this PR here to avoid one more PR to be rebased afterwards)

lib/private/Files/Cache/Cache.php
```diff
@@ -123,7 +123,7 @@ public function __construct(IStorage $storage) {
 
 		$this->storageCache = new Storage($storage);
 		$this->mimetypeLoader = \OC::$server->getMimeTypeLoader();
-		$this->connection = \OC::$server->getDatabaseConnection();
+		$this->connection = \OC::$server->get(IDBConnection::class);
```
ChristophWurst, line 126:
```suggestion
		$this->connection = \OCP\Server::get(IDBConnection::class);
```

let's use the public server container API. same applies to most changes below

ChristophWurst (approved):

lib/private/Files/Cache/Cache.php
summersab: I forgot to make this change before pushing, and I'm currently running tests. I don't know if you can cancel the tests that are currently running to free up resources, but feel free.

Also, why the change to the public container? In general, when should this be used?

summersab:
I'm getting the following failure:
```
1) OCA\DAV\Tests\unit\Connector\Sabre\FilesReportPluginTest::testPrepareResponses
Failed asserting that null matches expected 200.

./server/apps/dav/tests/unit/Connector/Sabre/FilesReportPluginTest.php:445
```

Since this has to do with the `dav` app and not the core, how should I address this? I DO plan to refactor apps once my PRs for the core are approved. This seems a bit like a chicken and egg situation.

lib/private/Files/Cache/Cache.php
nickvergessen: > In general, when should this be used?

In new code: always
In old code: when you touch it anyways

skjnldsv:
Too much conflicts, closing until this is done fresh again

PR #40164, opened 2023-08-30
Translate all share exceptions

solracsf:
## Summary

Fixes ambiguous messages like :
image

## Checklist

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -164,7 +164,7 @@ class Manager implements IManager {
 		} elseif ($share->getShareType() === IShare::TYPE_SCIENCEMESH) {
 		} else {
 			// We cannot handle other types yet
-			throw new \InvalidArgumentException($this->l->t('unknown share type'));
+			throw new \InvalidArgumentException($this->l->t('Unknown share type'));
 		}
 
 		// Verify the initiator of the share is set
@@ -1095,7 +1095,7 @@ class Manager implements IManager {
 	public function getSharesInFolder($userId, Folder $node, $reshares = false, $shallow = true) {
 		$providers = $this->factory->getAllProviders();
 		if (!$shallow) {
-			throw new \Exception($this->l->t('non-shallow getSharesInFolder is no longer supported'));
+			throw new \Exception('non-shallow getSharesInFolder is no longer supported');
 		}
 
 		return array_reduce($providers, function ($shares, IShareProvider $provider) use ($userId, $node, $reshares) {
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -103,7 +103,7 @@ class Manager implements IManager {
 		if ($password === null) {
 			// No password is set, check if this is allowed.
 			if ($this->shareApiLinkEnforcePassword()) {
-				throw new \InvalidArgumentException('Passwords are enforced for link and mail shares');
+				throw new \InvalidArgumentException($this->l->t('Passwords are enforced for link and mail shares'));
 			}
 
 			return;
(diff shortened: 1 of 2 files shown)


come-nc (dismissed):
Why all the $message_t intermediate var, just passed the result of the call to the exception constructor directly?
Also why in some places you rename an existing var?

solracsf:
The `$message_t` is already present in the codebase, so i've just applied the same logic.
Where the exitent `$var` has been renamed? Maybe a mistake?

artonge (dismissed):
Indeed, I would rather see the result used directly instead of having intermediate variables.

@come-nc where have you seen the variable rename ?

lib/private/Share20/Manager.php
```diff
@@ -427,15 +444,16 @@ protected function validateExpirationDateInternal(IShare $share) {
...
 			}
 
 			$date = new \DateTime();
 			$date->setTime(0, 0, 0);
 			$date->add(new \DateInterval('P' . $defaultExpireDays . 'D'));
 			if ($date < $expirationDate) {
-				$message = $this->l->n('Cannot set expiration date more than %n day in the future', 'Cannot set expiration date more than %n days in the future', $defaultExpireDays);
-				throw new GenericShareException($message, $message, 404);
+				$message_t = $this->l->n('Cannot set expiration date more than %n day in the future', 'Cannot set expiration date more than %n days in the future', $defaultExpireDays);
+				throw new GenericShareException($message_t, $message_t, 404);
```
come-nc, line 456: @artonge Here (and a few other places below)

skjnldsv:
> So, _in fine_, what's the status here?

put the translation message straight into the Exception :)

nfebe (commented):
According to previous reviews, there seems to be consensus to avoid using intermediate `$message` or `$message_t` variables, can this be updated?

solracsf:
Will do as soon as I find some time, no problem. 👍

Valdnet (approved):

lib/private/Share20/Manager.php
```diff
@@ -130,63 +130,63 @@ protected function generalCreateChecks(IShare $share, bool $isUpdate = false) {
...
-				throw new \InvalidArgumentException('SharedWith is not a valid circle');
+				throw new \InvalidArgumentException($this->l->t('SharedWith is not a valid circle'));
 			}
 		} elseif ($share->getShareType() === IShare::TYPE_ROOM) {
 		} elseif ($share->getShareType() === IShare::TYPE_DECK) {
 		} elseif ($share->getShareType() === IShare::TYPE_SCIENCEMESH) {
 		} else {
 			// We cannot handle other types yet
-			throw new \InvalidArgumentException('unknown share type');
+			throw new \InvalidArgumentException($this->l->t('unknown share type'));
```
Fenn-CS, line 167:
```suggestion
			throw new \InvalidArgumentException($this->l->t('Unknown share type'));
```

Fenn-CS, line 167:
```suggestion
			throw new \InvalidArgumentException($this->l->t('Unknown share type'));
```

lib/private/Share20/Manager.php
```diff
@@ -1102,7 +1095,7 @@ public function moveShare(IShare $share, $recipientId) {
 	public function getSharesInFolder($userId, Folder $node, $reshares = false, $shallow = true) {
 		$providers = $this->factory->getAllProviders();
 		if (!$shallow) {
-			throw new \Exception('non-shallow getSharesInFolder is no longer supported');
+			throw new \Exception($this->l->t('non-shallow getSharesInFolder is no longer supported'));
```
Fenn-CS, line 1098:
```suggestion
			throw new \Exception($this->l->t('Non-shallow getSharesInFolder is no longer supported'));
```
Or maybe `'"non-shallow" getSharesInFolder....`??

nfebe (approved):
Looks good, some capitalization suggestions. Thank you.

lib/private/Share20/Manager.php
susnux: To be honest, I think this message makes no sense to translate as this is an error message that does not help users. It is a developers only error, so maybe just not translate it?

```suggestion
			throw new \Exception('non-shallow getSharesInFolder is no longer supported');
```

lib/private/Share20/Manager.php
susnux: This is also an exception that should not be user facing, so I do not see any reason to translate it.
(I think translated developer facing exception make debugging log files much harder as you need to figure out what that error means (e.g. if a user reports a bug and you do not speak the users configured language)).

But yes otherwise I fully agree with the suggestion :)

skjnldsv (approved):

Valdnet (approved):

skjnldsv:
Sharding test error unrelated

solracsf:
Worth a backport to 30 at least ?

susnux:
If you manually backport, then please include: https://github.com/nextcloud/server/pull/40164

PR #40169, opened 2023-08-31
enh(settings): Load from disabled users endpoint

### LINKED ISSUE #41342: Disabled users tab not working unless searching
### Steps

1. Go to user management
2. Switch to "Disabled users (1)"

### Expected

See disabled users

### Actually

If none of the first 25 users of the instance are disabled the list will be empty:
[023-11-08 14-53-41.webm](https://github.com/nextcloud/server/assets/213943/1bca0798-2ce7-453d-b0bf-ee60eda21b2d)

### Problem

It queries `…/ocs/v2.php/cloud/users/details?offset=0&limit=25&search=` without a filter for disabled users and it does not "loop"

* Ref mostlikely https://github.com/nextcloud/server/pull/39050

cc @Pytal

Pytal:
- Use endpoint https://github.com/nextcloud/server/pull/39756

## Summary

- Load users directly from disabled users endpoint

## Checklist

- Screenshots before/after for front-end changes


come-nc (changes requested):
User count seems broken, also avatar of freshly disabled users do not load.

apps/settings/src/components/UserList.vue
```diff
@@ -314,6 +314,10 @@ export default {
 		},
 
 		async handleScrollEnd() {
+			// Do not load more users when all have already been loaded
+			if (this.users.length % 25 !== 0) {
+				return
+			}
```
come-nc, line 320: Hum, what if I load the list, enable a user from the list, and then scroll to the end? Length will be 24, no?

Pytal: Nope, only `filteredUsers` would change while `users` which contains all users loaded would stay the same

come-nc:
It works for disabled user list but for active user list it does not load the next ones when scrolling to the end it seems @Pytal

artonge (approved):

come-nc:
This is still not working correctly for me for active user list.
The problem is that with LDAP backend it returns more users than the limit, because the limit is applied to each backend indivdually. This may be seen as a bug but has been this way for a long time. (and fixing that brings perf questions)

So the test that user count is a multiple of 25 does not work, in may case the first request with limit 25 and offset 0 returns 37 users (25 from ldap, the rest from DB).
What you need to test to know if there are no more users is if the number of returned users is less than the limit.

Note that for disabled users endpoint it behaves better and will slice to return only 25 users when limit is 25, but maybe it’s still safer to consider 25 users **or more** valid sizes for non-complete answer.

come-nc (approved):

susnux (approved):

PR #40192, opened 2023-09-01
feat(files): add sharing icon in header 

skjnldsv:
|Not shared|
|:---|
||
|Shared by link|
||
|Shared|
||


szaimen:
> Shared
> image

I am not sure about this style. Shouldn't we better use a different icon, e.g. this one?
image

skjnldsv:
@szaimen Feel free to open an issue, but for now this is feature parity.
Please approve so we can move forward :)

tcitworld:
The indicator when "not shared" had an opacity of 0.2 to indicate that it's not currently shared, it's just a button to reach the sharing sidebar.

There's still a differentiation with the new "shared" status, but the "not shared" status matches the previous "shared" status, so people might be confused.

nimishavijay:
There doesn't seem to be a clear distinction between a button that does an action and an indicator especially because it's styled as a secondary button so it's quite prominent, and @szaimen's icon suggestion makes sense (although it should be the filled variant and not outlined).

How about:

Not shared
image

Shared
image

Shared via link
image

skjnldsv:
Icons before the name is not supported.
You can pick whatever design you want on the sharing icon, but not before (feel free to open an issue in the vue components, but this is not a blocker for this PR)

skjnldsv:
> The indicator when "not shared" had an opacity of 0.2 to indicate that it's not currently shared, it's just a button to reach the sharing sidebar.

Here you go, we now have exactly what we had on the old files app.
For any change, please open a dedicated issue, let's not discuss this here, thank you :)

szaimen:
I created https://github.com/nextcloud/server/issues/40202

szaimen:
I tested this and the icon does not seem to directly update after I shared the folder for example. I guess this should be the case though?

skjnldsv:
> I tested this and the icon does not seem to directly update after I shared the folder for example. I guess this should be the case though? Apart from that seems to work well so good work!

Needs a separate fix. We'll need to do a pass on global events. But right now the Sidebar doesn't update the files store, you are correct :)

szaimen:
> > I tested this and the icon does not seem to directly update after I shared the folder for example. I guess this should be the case though? Apart from that seems to work well so good work!
> 
> Needs a separate fix. We'll need to do a pass on global events. But right now the Sidebar doesn't update the files store, you are correct :)

Shall I create a follow-up issue then?

lhsazevedo (approved):
Working well for me as well

szaimen:
> Shall I create a follow-up issue then?

I've created https://github.com/nextcloud/server/issues/40263

szaimen (approved):
Okay for me then for now

PR #40199, opened 2023-09-01
remove old fileUploadSpec jsunit test and fix flaky cypress files test 

skjnldsv:
Jsunit was broken on master it seems and cypress files is actually going too fast and the preview preload throw because it logout before being done fetching it :see_no_evil:


szaimen (approved):
Seems to fix it

kesselb (approved):

PR #40208, opened 2023-09-01
Fix sharing and around three errors related to Undefined array key "CLOUD"

klondi:
## Summary
We try to access $contact['CLOUD'] when it might not be set. As a consequence of that 3 PHP errors are generated the last one being an uncaught exception that makes it impossible to share files locally.

This patch adds a simple check to ensure the value exists before using it.

Please backport this to stable 26 too as it also has the same issue.

## Checklist

### DIFF
--- a/lib/private/Collaboration/Collaborators/MailPlugin.php
+++ b/lib/private/Collaboration/Collaborators/MailPlugin.php
@@ -186,7 +186,7 @@ class MailPlugin implements ISearchPlugin {
 							return false;
 						}
 
-						if ($this->shareeEnumeration) {
+						if ($this->shareeEnumeration && isset($contact['CLOUD'])) {
 							try {
 								$cloud = $this->cloudIdManager->resolveCloudId($contact['CLOUD'][0]);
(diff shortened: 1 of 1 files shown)


klondi:
For more details this seems to be triggered by the  LDAP Contacts Backend as the contacts returned do not (for obvious reasons) have a CLOUD parameter.

joshtrichards:
Is this associated with an open Issue?

lib/private/Collaboration/Collaborators/MailPlugin.php
```diff
@@ -186,7 +186,7 @@ public function search($search, $limit, $offset, ISearchResult $searchResult) {
 							return false;
 						}
 
-						if ($this->shareeEnumeration) {
+						if ($this->shareeEnumeration && isset($contact['CLOUD'])) {
```
tcitworld, line 189: **nitpick:** Please add the check on the condition line 164 as well.

susnux:
This should be fixed now with the latest refactoring of the mail plugin.
But thank you for your contribution!

PR #40209, opened 2023-09-01
feat(files): add folder icon overlay

skjnldsv:
Ref https://github.com/nextcloud/server/issues/39914

I decided to hardcode it for now as I couldn't find a good API way to tackle that.
We can always talk and improve from there :)

image

@nextcloud/designers if you want to adjust those icons, suggest new ones from the material design set as a review comment. I currently used the ones we already use or some that got asked by Jan.


skjnldsv:
I don't have an e2ee setup available, can you try it yourself @artonge ? :pray:

szaimen (approved):
LGTM but didnt test

jancborchardt (approved):
Looks good design-wise!

It would be nice to switch the share-triangle to this less technical "add person" icon:
https://pictogrammers.com/library/mdi/icon/account-plus/
I think I mentioned to you before @skjnldsv? But probably separate issue as it should be done across the board – your pick.

skjnldsv:
> I think I mentioned to you before @skjnldsv?

We did that for the Navigation yes
Do I understand you want me to replace them every time I encounter the old ones?

artonge:
Both are shared by the current user.
`test_normal` is not e2e encrypted.
`test_4` is e2e encrypted.
Screenshot from 2023-09-05 10-54-36
Is it what you're expecting ?

skjnldsv:
> Is it what you're expecting ?

What is the old behaviour? We show the encrypted feedback before the sharing feedback?
Does the key show on encrypted but non-shared folder?

artonge:
Old behavior:
Screenshot from 2023-09-05 11-02-11

For history, it was recently fixed by https://github.com/nextcloud/server/pull/36748

szaimen:
Yes a key looks good to me 👍

cc @jancborchardt @nimishavijay

skjnldsv:
Yep, icons are changing more and more along the releases.
As we're moving them all to MaterialDesign

tobiasKaminsky:
Once we start, we will ask/check for a complete list, so that all clients have same as web.

allexzander:
@skjnldsv Where can I get the exact same icons as in your screenshot?

Issue #40107, opened 2023-08-29
Prevent continuous file-reading when stopping/restarting streaming triggered by in-video searches


AndyScherzinger:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
* Share a big video by link, in my test I use a 6GB mp4 video
* Open the link in the browser, I use FireFox
* Open the network tab of the browser to watch networking activity
* Start top on the server to watch apache processes and server load
* Start playing the video, then slowly pull the seek handle to the right
* This produces lots of range requests and the same amount of apache processes serving one request each
* If the serving processes don't monitor connection status then all of them read the video from the start point requested in the range request to the end of the file, producing a lot of load on the server.
* The problem is worse if the client is on a relatively slow DSL connection

**Describe the solution you'd like**
Opening such a link, playing the video directly in the browser and then seeking in the video produces (lots of) range requests for the video file. This is okay, as the browser closes all the connections for range requests it is no longer interested in. But PHP does not care and the way the video was served resulted in all the requests reading the video to the end, producing a lot of load on the storage system and on the server itself for no good reason. Prevent this problem by checking connection status regularly.

**Describe alternatives you've considered**
none

**Additional context**
none

PR #40233, opened 2023-09-04
[stable25] Detect aborted connection in OC\Files\View and stop writing data to the output buffer

come-nc:
* Resolves: #40107

## Summary

When reading a video through a public link, and moving the time picker to different times in the video, each request keeps running after being aborted by the web browser and that overloads the server.
This is not the same as https://github.com/sabre-io/http/pull/207 because in this case the download does not go through dav.

## Checklist

### DIFF
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -426,11 +426,11 @@ class View {
 		}
 		$handle = $this->fopen($path, 'rb');
 		if ($handle) {
-			$chunkSize = 524288; // 512 kB chunks
+			$chunkSize = 524288; // 512 kiB chunks
 			while (!feof($handle)) {
 				echo fread($handle, $chunkSize);
-				$this->checkConnectionStatus();
 				flush();
+				$this->checkConnectionStatus();
 			}
 			fclose($handle);
 			return $this->filesize($path);
@@ -482,8 +482,8 @@ class View {
 						$len = $chunkSize;
 					}
 					echo fread($handle, $len);
-					$this->checkConnectionStatus();
 					flush();
+					$this->checkConnectionStatus();
 				}
 				return ftell($handle) - $from;
 			}
@@ -496,7 +496,7 @@ class View {
 
 	private function checkConnectionStatus(): void {
 		$connectionStatus = \connection_status();
-		if ($connectionStatus !== 0) {
+		if ($connectionStatus !== CONNECTION_NORMAL) {
 			throw new ConnectionLostException("Connection lost. Status: $connectionStatus");
 		}
 	}
--- a/lib/private/Files/View.php
+++ b/lib/private/Files/View.php
@@ -53,6 +53,7 @@ use OC\User\LazyUser;
 use OCA\Files_Sharing\SharedMount;
 use OCP\Constants;
 use OCP\Files\Cache\ICacheEntry;
+use OCP\Files\ConnectionLostException;
 use OCP\Files\EmptyFileNameException;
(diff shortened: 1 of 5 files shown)


lib/private/Files/View.php
```diff
@@ -490,6 +493,14 @@ public function readfilePart($path, $from, $to) {
 		return false;
 	}
 
+
+	private function checkConnectionStatus(): void {
+		$connectionStatus = \connection_status();
+		if ($connectionStatus !== 0) {
```
Altahrim, line 499: I think it can be more explicit that way:
```suggestion
		if (\connection_aborted()) {
```

lib/private/Files/View.php
```diff
@@ -428,6 +429,7 @@ public function readfile($path) {
 			$chunkSize = 524288; // 512 kB chunks
 			while (!feof($handle)) {
 				echo fread($handle, $chunkSize);
+				$this->checkConnectionStatus();
```
Altahrim, line 432: I think it would be better to put this line after the `flush()`.
According to [php.net ignore-user-abort](https://www.php.net/manual/en/function.ignore-user-abort.php): 
> PHP will not detect that the user has aborted the connection until an attempt is made to send information to the client. Simply using an echo statement does not guarantee that information is sent, see [flush()](https://www.php.net/manual/en/function.flush.php).

I guess it also applies to `connection_status` & cie

AndyScherzinger (approved):

lib/private/Files/View.php
come-nc: Yeah but it gives less information, in the debug log the connectionStatus is stored, which may help debug weird edge cases related to this later.

lib/private/Files/View.php
come-nc: Hum given it’s in a loop it does check between each flush anyway, no?

lib/private/Files/View.php
Altahrim: We can keep the `connection_status()` for the exception.

At first I was bothered by the "arbitrary" value of zero.
So, maybe we can write it like this instead:
```suggestion
		if ($connectionStatus !== CONNECTION_NORMAL) {
```

lib/private/Files/View.php
```diff
@@ -428,6 +429,7 @@ public function readfile($path) {
 			$chunkSize = 524288; // 512 kB chunks
```
Altahrim, line 429: Nit: 
```suggestion
			$chunkSize = 524288; // 512 kiB chunks
```

lib/private/Files/View.php
```diff
@@ -428,6 +429,7 @@ public function readfile($path) {
...
 			while (!feof($handle)) {
 				echo fread($handle, $chunkSize);
+				$this->checkConnectionStatus();
 				flush();
 			}
```
come-nc, line 434:
```suggestion
			while (!feof($handle)) {
				echo fread($handle, $chunkSize);
				flush();
				$this->checkConnectionStatus();
			}
```

lib/private/Files/View.php
```diff
@@ -480,6 +482,7 @@ public function readfilePart($path, $from, $to) {
 						$len = $chunkSize;
 					}
 					echo fread($handle, $len);
+					$this->checkConnectionStatus();
 					flush();
```
Altahrim, line 486: It should also be changed here:
```suggestion
					flush();
					$this->checkConnectionStatus();
```

Altahrim (approved):

blizzz (approved):

PR #40266, opened 2023-09-05
40172 Polish new sharing flow :  accesibility, expand bahavior, click outside behaviour

### LINKED ISSUE #40172: ⛑️ Polishing for new sharing design and flow 🤝
Follow-up to #26691 listing the remaining open issues to be fixed in `master` as well as `stable27`

---

https://github.com/nextcloud/server/pull/39472#issuecomment-1680970478
## Share list
- [x] BUG: Quick access menu should close when click outside

### Bugs specific to share settings
- [x] BUG: Setting "Allow public upload" disabled should also disable options to set file drop, according permissions

Screenshot 2023-08-17 at 12 23 33

- [x] BUG: Lots of optional share settings for the ones that should affect front-end:
  - [x] BUG: Public upload can be blocked
- [x] BUG: https://github.com/nextcloud/server/issues/40289
---

- [x] BUG: https://github.com/nextcloud/server/issues/40228 - Edit (louis): seem unrelated
- [x] BUG: https://github.com/nextcloud/server/issues/40299

---

- [x] BUG: Create a new share to a user, change to custom permission, click save share, go back to the share details -> now update and share check-boxes are missing (the custom permissions are the same as view only, but nevertheless the check-boxes should of course still be available)
- [x] BUG: Creating shares should probably not reload the full file list. Edit (louis): only an issue with 27 as the logic to handle share changes is not fine-grained enough to only refresh the share status of a file. Also, it was not introduced by the new sharing flow.
- [x] BUG: Related resources should not appear in the share details view - https://github.com/nextcloud/server/pull/40365

## Non-blocking
- [x] Accessibility fixes for quick menu
- [x] "Delete share" button does not need to be sticky, can just be a tertiary button (no background by default) below the advanced settings toggle (not inside advanced settings)
- [ ] The "Cancel" and primary button on the bottom need a backdrop to separate them a bit from the background, best a fade from color-main-background on the bottom to transparent on the top.
- [ ] Currently links are sorted below emails once created, but sorting should be: Links, mails, users
- [x] Too many share saved/updated notifications when updating/creating shares
- [ ] QA: Needs side by side testing if all old options are possible or also properly limited with the new one
- [ ] #40436
- [ ] If allow public upload admin setting is unchecked we should rename "Allow upload and editing" to "allow editing" and file drop / create permission should be disabled state

---

https://github.com/nextcloud/server/pull/39472#pullrequestreview-1599979725

### Follow-ups:
- [x] Switching between permissions presets should not close the advanced settings view
- [ ] The content of the sidebar is scroll-able when it should only be the content of the share tab

Screenshot from 2023-08-29 11-52-15

### UI comments:
- [x] Button at the bottom of the details view should not move when scrolling to the bottom
- [x] "Delete share" button text seems smaller than other buttons
- [ ] It is confusing to have the download permission under "Custom permissions" as it is not taken into account for bundles

nfebe:
Addresses the following issues
----

https://github.com/nextcloud/server/pull/39472#issuecomment-1680970478
## Share list
- [x] BUG: Quick access menu should close when click outside

- [x] BUG: Create a new share to a user, change to custom permission, click save share, go back to the share details -> now update and share check-boxes are missing (the custom permissions are the same as view only, but nevertheless the check-boxes should of course still be available)

## Non-blocking
- [x] Accessibility fixes for quick menu

### Follow-ups:
- [x] Switching between permissions presets should not close the advanced settings view

Resolves : https://github.com/nextcloud/server/issues/40172


apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
```diff
@@ -1,16 +1,32 @@
...
 			{{ selectedOption }}
 			<DropdownIcon :size="15" />
 		</span>
-		<div v-if="showDropdown" class="share-select-dropdown-container">
-			<div v-for="option in options"
+		<div v-if="showDropdown"
+			ref="quickShareDropdown"
+			class="share-select-dropdown"
+			:aria-labelledby="dropdownId"
+			tabindex="0"
```
artonge, line 20: Why does the container need to be focusable ?

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
```diff
@@ -1,16 +1,32 @@
...
+			@keydown.esc="closeDropdown">
+			<button v-for="option in options"
 				:key="option"
 				:class="{ 'dropdown-item': true, 'selected': option === selectedOption }"
+				tabindex="0"
```
artonge, line 25: Should not be needed if the element is a button.

artonge (commented):
Could be useful to have @Pytal's review for the focus trap as I am not familiar with it

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
Fenn-CS: Fixed

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
Fenn-CS: So that it can respond to the keyboard escape event.

juliusknorr (approved):

AndyScherzinger (approved):

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
artonge: Can it be done from the children instead ?

apps/files_sharing/src/components/SharingEntryQuickShareSelect.vue
```diff
@@ -128,6 +166,47 @@ export default {
...
+
+			if (dropdownContainer && !dropdownContainer.contains(event.target)) {
+				this.showDropdown = false
+			}
+		},
+		useFocusTrap() {
+			const dropdownElement = this.$refs.quickShareDropdown
+			this.focusTrap = createFocusTrap(dropdownElement, {
+				allowOutsideClick: true,
+			})
```
Pytal, line 180: Let's add `trapStack` to ensure no fighting traps
https://github.com/nextcloud/server/blob/5cb2713ea4499eaa5bd108cec74effe3cd13c7be/core/src/jquery/ocdialog.js#L258-L264

Pytal (approved):
Other than the remark below

PR #40267, opened 2023-09-05
[stabl26] LDAP: Increase profile picture limit to 512

marcelklehr:
## Checklist


kesselb (approved):
Looks like a reasonable change :+1:

blizzz:
first master, then backports?

marcelklehr:
If you insist. The customer we needed a patch for was on 26, so this is how it went :shrug:

blizzz:
> If you insist. The customer we needed a patch for was on 26, so this is how it went 🤷

You know, it ends up too often that stuff is then merged in older versions, but does not make its way up. A stitch in time saves nine! ;)

PR #40284, opened 2023-09-06
[f2v] Render inline system tags using new DAV properties API

lhsazevedo:
Ref: #39914 #37938

## Summary
Display inline system tags using the new DAV properties API.

image

image

I kept the action in the files app since the previous plugin was in there as well, and I didn't found any action registration outside the files* apps. But I can move it to the `systemtags` app if that is preferred!

## TODO

- [x] Decide if we should move the action to the `systemtags` app

## Checklist

- ~Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required~
- ~[Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)~


lhsazevedo:
Possible follow ups:

- Maybe we should hide the system tags (or all inline action?) on mobile to avoid truncating the node name?
  image
- Looks like the search by tag feature needs to be migrated as well.

skjnldsv:
> * Maybe we should hide the system tags (or all inline action?) on mobile to avoid truncating the node name?

Yes please!

> * Looks like the search by tag feature needs to be migrated as well.

Absolutely, this is ongoing already, see [overview](https://github.com/nextcloud/server/issues/39914)

skjnldsv (commented):
Nice!!
Looks good so far. A few suggestions:
- [x] Move those scripts to the `systemtags` app
- [x] Fix eslint

lhsazevedo:
Done! Added some tests as well. For now I'm hiding the on mobile using CSS media queries, as I'm not sure if registering an size observer inside `renderInline` would be the proper way to make an action react to a screen resize or if we should modify the FileAction API for this instead.

Unfortunately the tags aren't being rendered on the initial DAV request after moving to the `systemtags` app. I checked to ensure the action is being called, and it is indeed, but looks like now the dav property isn't being registered in time for the first request. Subsequent requests have the property registered:

https://github.com/nextcloud/server/assets/7695608/10f4424a-60bf-4697-8c08-0dbd68e9c2ac

I believe it isn't possible to register the property earlier while still in the `systemtags` app, is it?

skjnldsv:
> Done! Added some tests as well. For now I'm hiding the tags on mobile using CSS media queries, as I'm not sure if registering an size observer inside `renderInline` would be the proper way to make an action react to a screen resize or if we should modify the FileAction API for this instead.

Media queries is how it should be done.
If you wanna prevent the fetching for ressources optimisation, you can also use `document.documentElement.clientWidth < 512` and not register the dav props? :thinking: 

> Unfortunately the tags aren't being rendered on the initial DAV request after moving to the `systemtags` app. I checked to ensure the action is being called, and it is indeed, but looks like now the dav property isn't being registered in time for the first request. Subsequent requests have the property registered:

We actually discovered that with @juliushaertl yesterday and worked on a fix, you can ignore it :)

skjnldsv (approved):

lhsazevedo:
> If you wanna prevent the fetching for ressources optimisation, you can also use document.documentElement.clientWidth < 512 and not register the dav props? 🤔

Not sure if the complexity trade off is worth it... We'll probably consume them in other places of the app soon, so I guess it's Ok for them to be available on mobile as well for now?

Altahrim (approved):

PR #40296, opened 2023-09-06
Enable new versions feature for groupfolders

artonge:
## New
- Some logic that should have been part of the main version backend were in the shared logic. So I created a `INeedSyncVersionBackend` interface to discriminate backends that need to update their DB on some node events. The main backend and the groupfolders backend both need it.
- Unlock the frontend to allow access to the new versions features for groupfolders
- Fix naming the current version, it was lost when working on opening versions in the viewer.

## Todo
- [x] Check tests

## Note
~~This PR is kept before F2V on purpose to make it easier to test.~~


AndyScherzinger (approved):

come-nc (approved):

artonge:
Failing test is unrelated

apps/files_versions/src/views/VersionTab.vue
```diff
@@ -249,7 +253,13 @@ export default {
 
 			// Versions previews are too small for our use case, so we override hasPreview and previewUrl
 			// which makes the viewer render the original file.
-			const versions = this.versions.map(version => ({ ...version, hasPreview: false, previewUrl: undefined }))
+			// We also point to the original filename if the version is the current one.
+			const versions = this.versions.map(version => ({
+				...version,
+				filename: version.mtime === this.fileInfo.mtime ? path.join('files', getCurrentUser()?.uid ?? '', fileInfo.path, fileInfo.name) : version.filename,
```
juliushaertl, line 259: Caused a regression on the version preview as fileInfo is not defined here (should be `this.fileInfo`). Fix is at https://github.com/nextcloud/server/pull/40348

PR #40312, opened 2023-09-06
fix: Pass parent to NonExistingFile instances

juliusknorr:
## Summary

Fixes calls to getParent after #38150 when calling on NonExistingFile instances, e.g. within a BeforeNodeRenamedEvent.

Found in failing tests for moving files in text which has special handling. CI is passing with this branch again: https://github.com/nextcloud/text/pull/4779

## Checklist

### DIFF
--- a/lib/private/Files/Node/Node.php
+++ b/lib/private/Files/Node/Node.php
@@ -297,10 +297,19 @@ class Node implements INode {
 				return $this->root;
 			}
 
+			// Manually fetch the parent if the current node doesn't have a file info yet
+			try {
+				$fileInfo = $this->getFileInfo();
+			} catch (NotFoundException) {
+				$this->parent = $this->root->get($newPath);
+				/** @var \OCP\Files\Folder $this->parent */
+				return $this->parent;
+			}
+
 			// gather the metadata we already know about our parent
 			$parentData = [
 				'path' => $newPath,
-				'fileid' => $this->getFileInfo()->getParentId(),
+				'fileid' => $fileInfo->getParentId(),
 			];
 
 			// and create lazy folder with it instead of always querying
--- a/tests/lib/Files/Node/NodeTest.php
+++ b/tests/lib/Files/Node/NodeTest.php
@@ -481,8 +481,7 @@ abstract class NodeTest extends \Test\TestCase {
 		$parentNode = new \OC\Files\Node\Folder($this->root, $this->view, '/bar');
 		$newNode = $this->createTestNode($this->root, $this->view, '/bar/asd');
 
-		$this->root->expects($this->exactly(2))
-			->method('get')
+		$this->root->method('get')
 			->willReturnMap([
 				['/bar/asd', $newNode],
 				['/bar', $parentNode]


skjnldsv (approved):

lib/private/Files/Node/File.php
```diff
@@ -40,7 +40,7 @@ class File extends Node implements \OCP\Files\File {
 	 * @return NonExistingFile non-existing node
 	 */
 	protected function createNonExistingNode($path) {
-		return new NonExistingFile($this->root, $this->view, $path);
+		return new NonExistingFile($this->root, $this->view, $path, $this->root->get(dirname($path)));
```
come-nc, line 43: Can you go into the details of what breaks without this and why is the lazy parent not enough?

juliushaertl: getParent used to just directly get it in https://github.com/nextcloud/server/pull/38150/files#diff-edc136e5646ea4171607db53b38833e8f6caaacc64874a0a850aa4444f0a56edL303-R309 but now always makes use of the file info which is not available for a nonexisting file.

come-nc: Then wouldn’t the right fix be to check for fileinfo presence before calling getParentId on it and fallback on previous behavior if fileInfo is not there? I’m afraid there might be other side effects we missed.

juliushaertl: Yeah might be indeed a good idea, will check that

juliushaertl: Pushed a more generic approach but still passing the parent to the NonExistingFile to avoid an extra queries in that case.

lib/private/Files/Node/Node.php
```diff
@@ -297,10 +297,18 @@ public function getParent(): INode|IRootFolder {
 				return $this->root;
 			}
 
+			// Manually the parent if the current node doesn't have a file info yet
```
come-nc, line 300:
```suggestion
			// Manually fetch the parent if the current node doesn't have a file info yet
```
Or get? But a word is missing there I think.

come-nc (approved):

lib/private/Files/Node/Node.php
juliushaertl: Yep, fixed

juliushaertl:
I cannot see the same psalm error locally :/

tests/lib/Files/Node/NodeTest.php
```diff
@@ -481,7 +481,7 @@ public function testCopySameStorage() {
 		$parentNode = new \OC\Files\Node\Folder($this->root, $this->view, '/bar');
 		$newNode = $this->createTestNode($this->root, $this->view, '/bar/asd');
 
-		$this->root->expects($this->exactly(2))
+		$this->root->expects($this->any())
```
icewind1991, line 484: You can just remove the `expects` altogether

lib/private/Files/Node/File.php
icewind1991: > to avoid an extra queries in that case.

doesn't it just move the query to the constructor call here?

As is it always does the "get parent" query even when `getParentId` isn't called, without this it only does the query when `getParentId` is called

icewind1991 (dismissed):

icewind1991 (approved):

PR #40323, opened 2023-09-07
feat: Add dedicated method to load init scripts

juliusknorr:
During development on text I noticed that the new files2vue file list is loading too fast and fetching the file list before text has even had a chance to register additional dav properties on the initial page load.

This adds a method to allow apps registering init scripts which are loaded first (even before any core javascript). They need to be handled with care, but I indicated that also on the phpdoc comment.

Change itself is from @skjnldsv, I just shaped it to a pull request.

Example usage: https://github.com/nextcloud/text/pull/4776/commits/0eb23bbf8a3afc534a62e692957d00f4222a56ba

## Checklist

### DIFF
--- a/lib/public/Util.php
+++ b/lib/public/Util.php
@@ -49,6 +49,7 @@ namespace OCP;
 use OC\AppScriptDependency;
 use OC\AppScriptSort;
 use bantu\IniGetWrapper\IniGetWrapper;
+use OCP\Share\IManager;
 use Psr\Container\ContainerExceptionInterface;
 
 /**
@@ -57,17 +58,11 @@ use Psr\Container\ContainerExceptionInterface;
  * @since 4.0.0
  */
 class Util {
-	/** @var \OCP\Share\IManager */
-	private static $shareManager;
+	private static ?IManager $shareManager = null;
 
-	/** @var array */
-	private static $scripts = [];
-
-	/** @var array */
-	private static $scriptDeps = [];
-
-	/** @var array */
-	private static $sortedScriptDeps = [];
+	private static array $scriptsInit = [];
+	private static array $scripts = [];
+	private static array $scriptDeps = [];
 
 	/**
 	 * get the current installed version of Nextcloud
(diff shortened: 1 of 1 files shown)


lib/public/Util.php
```diff
@@ -163,6 +158,26 @@ public static function addStyle($application, $file = null) {
 		\OC_Util::addStyle($application, $file);
 	}
 
+	/**
+	 * Add a standalone init js file that is loaded for initialization
+	 *
+	 * Be careful loading scripts using this method as they are loaded early
+	 * and block the initial page rendering.
+	 * They should not have dependencies on any of the core javascript logic
+	 * as it is also not loaded at that stage.
```
skjnldsv, line 167: This is not really true, like you can require OC properties at this stage
It is purposely being pushed after the `core-common` file.

juliushaertl: Ah then my memory was wrong in that, will adjust

skjnldsv: image

skjnldsv (approved):

lib/public/Util.php
juliushaertl: Adjusted

skjnldsv (approved):

come-nc (approved):

lib/public/Util.php
```diff
@@ -214,6 +228,7 @@ public static function getScripts(): array {
 		$sortedScripts = $scriptSort->sort(self::$scripts, self::$scriptDeps);
 
 		// Flatten array and remove duplicates
+		$sortedScripts = array_merge([self::$scriptsInit], $sortedScripts);
 		$sortedScripts = $sortedScripts ? array_merge(...array_values(($sortedScripts))) : [];
```
come-nc, line 232:
```suggestion
		$sortedScripts = array_merge(...array_values($sortedScripts));
```

ChristophWurst:
this new feature needs docs

PR #40326, opened 2023-09-07
Implement TextToImage OCP API

marcelklehr:
## Summary
Implements an OCP API for registering and consuming text to image providers like stable diffusion.

## Todo
- [x] cleanup images in appdata

## Checklist


core/Controller/TextToImageApiController.php
```diff
@@ -0,0 +1,220 @@
...
+	 * Tasks are removed 1 week after receiving their last update.
+	 *
+	 * @param int $id The id of the task
+	 *
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTextToImageTask}, array{}>|DataResponse<Http::STATUS_NOT_FOUND|Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 * 200: Task returned
+	 * 404: Task not found
+	 */
+	#[PublicPage]
```
come-nc, line 114: It seems id are from DB autoincrement, so easy to guess. With all these being public pages, anyone can spy on the result of the tasks, no?

marcelklehr: Only tasks created by not logged-in users can be seen by all not-logged-in users, tasks created by a specific user are only shown to that user (see Manager#getUserTask)

lib/public/TextToImage/IManager.php
```diff
@@ -0,0 +1,98 @@
...
+	 * @throws TaskNotFoundException If the task could not be found
+	 * @since 28.0.0
+	 */
+	public function getUserTask(int $id, ?string $userId): Task;
+
+	/**
+	 * @param string $userId
+	 * @param string $appId
+	 * @param string|null $identifier
+	 * @return array
```
juliushaertl, line 94:
```suggestion
	 * @return Task[]
```

lib/private/TextToImage/Manager.php
```diff
@@ -0,0 +1,244 @@
...
+				} else {
+					$this->taskMapper->update(DbTask::fromPublicTask($task));
+				}
+				try {
+					$folder = $this->appData->getFolder('text2image');
+				} catch(\OCP\Files\NotFoundException $e) {
+					$folder = $this->appData->newFolder('text2image');
+				}
+				$file = $folder->newFile((string) $task->getId());
+				$provider->generate($task->getInput(), $file->write());
```
juliushaertl, line 132: Maybe close the file resource afterwards to not require every provider to do it?

lib/private/TextToImage/Manager.php
```diff
@@ -0,0 +1,244 @@
...
+			} catch (\RuntimeException $e) {
+				$this->logger->info('Text2Image generation using provider ' . $provider->getName() . ' failed', ['exception' => $e]);
+				$task->setStatus(Task::STATUS_FAILED);
+				$this->taskMapper->update(DbTask::fromPublicTask($task));
+				throw $e;
+			} catch (\Throwable $e) {
+				$this->logger->info('Text2Image generation using provider ' . $provider->getName() . ' failed', ['exception' => $e]);
+				$task->setStatus(Task::STATUS_FAILED);
+				$this->taskMapper->update(DbTask::fromPublicTask($task));
+				throw new RuntimeException('Text2Image generation using provider ' . $provider->getName() . ' failed: ' . $e->getMessage(), 0, $e);
```
juliushaertl, line 145: Shall we cleanup the file if an error happened in generate?

lib/public/TextToImage/Exception/Exception.php
```diff
@@ -0,0 +1,28 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\TextToImage\Exception;
+
+class Exception extends \Exception {
```
juliushaertl, line 27: Can we name this a bit more specific?

lib/public/TextToImage/Events/TaskSuccessfulEvent.php
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\TextToImage\Events;
+
+/**
+ * @since 27.1.0
```
juliushaertl, line 30:
```suggestion
 * @since 28.0.0
```

lib/private/TextToImage/TaskBackgroundJob.php
```diff
@@ -0,0 +1,63 @@
...
+use OCP\EventDispatcher\IEventDispatcher;
+use OCP\TextToImage\Events\TaskFailedEvent;
+use OCP\TextToImage\Events\TaskSuccessfulEvent;
+use OCP\TextToImage\IManager;
+
+class TaskBackgroundJob extends QueuedJob {
+	public function __construct(
+		ITimeFactory             $timeFactory,
+		private IManager         $text2imageManager,
+		private IEventDispatcher $eventDispatcher,
```
juliushaertl, line 40:
```suggestion
		ITimeFactory $timeFactory,
		private IManager $text2imageManager,
		private IEventDispatcher $eventDispatcher,
```

lib/private/TextToImage/Manager.php
```diff
@@ -0,0 +1,244 @@
...
+			$className = json_decode($json, true);
+			$provider = current(array_filter($providers, fn ($provider) => $provider::class === $className));
+			if ($provider !== false) {
+				$providers = [$provider];
+			}
+		}
+
+		foreach ($providers as $provider) {
+			try {
+				$task->setStatus(Task::STATUS_RUNNING);
```
juliushaertl, line 119: Would it make sense to log the processing a bit more in detail at info or at least at debug level?

core/Migrations/Version28000Date20230906104802.php
```diff
@@ -0,0 +1,100 @@
...
+				'notnull' => true,
+				'length' => 32,
+				'default' => '',
+			]);
+			$table->addColumn('identifier', Types::STRING, [
+				'notnull' => true,
+				'length' => 255,
+				'default' => '',
+			]);
+			$table->addColumn('last_updated', Types::INTEGER, [
```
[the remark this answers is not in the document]
marcelklehr: Ah, I don't have too much experience using the datetime type, will this require changing https://github.com/nextcloud/server/pull/40326/files/1899a05326828603bad6c17f52543c7b0f32d6a8#diff-1c2d622cfc3d84c8702f2c18a7752f3deacc8c87daf12e768154cb78c54e2562R110 as well?

core/Controller/TextToImageApiController.php
```diff
@@ -0,0 +1,220 @@
...
+	public function getImage(int $id): DataResponse|FileDisplayResponse {
+		try {
+			$task = $this->textToImageManager->getUserTask($id, $this->userId);
+			try {
+				$folder = $this->appData->getFolder('text2image');
+			} catch(\OCP\Files\NotFoundException) {
+				$folder = $this->appData->newFolder('text2image');
```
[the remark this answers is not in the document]
marcelklehr: Added the cleanup mechanism now

lib/private/TextToImage/Db/TaskMapper.php
```diff
@@ -0,0 +1,128 @@
...
+	 * @return Task[] the deleted tasks
+	 * @throws Exception
+	 */
+	public function deleteOlderThan(int $timeout): array {
+		$datetime = new DateTime();
+		$datetime->sub(new \DateInterval('PT'.$timeout.'S'));
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->tableName)
+			->where($qb->expr()->lt('last_updated', $qb->createPositionalParameter($datetime, IQueryBuilder::PARAM_DATE)));
```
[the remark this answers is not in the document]
marcelklehr: Mh, am I using the wrong constant here?

lib/private/TextToImage/Db/Task.php
```diff
@@ -0,0 +1,120 @@
...
+ * @method setResultPath(string $resultPath)
+ * @method string getResultPath()
+ * @method setStatus(int $type)
+ * @method int getStatus()
+ * @method setUserId(?string $userId)
+ * @method string|null getUserId()
+ * @method setAppId(string $type)
+ * @method string getAppId()
+ * @method setIdentifier(string $identifier)
+ * @method string getIdentifier()
```
nickvergessen, line 50:
```suggestion
 * @method setIdentifier(string $identifier)
 * @method string|null getIdentifier()
```

For oracle

lib/private/TextToImage/Db/TaskMapper.php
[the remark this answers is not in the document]
nickvergessen: Fix in https://github.com/nextcloud/server/pull/40959

You are just the first one using `createPositionalParameter` with it, `createNamedParameter` does not have a typehint to int and therefore allowed anything.

lib/private/TextToImage/Manager.php
```diff
@@ -0,0 +1,280 @@
...
+		$json = $this->config->getAppValue('core', 'ai.text2image_provider', '');
+		if ($json !== '') {
+			try {
+				$className = json_decode($json, true, 512, JSON_THROW_ON_ERROR);
+				$provider = current(array_filter($providers, fn ($provider) => $provider::class === $className));
+				if ($provider !== false) {
+					$providers = [$provider];
+				}
+			} catch (\JsonException $e) {
+				$this->logger->warning('Failed to decode Text2Image setting `ai.text2image_provider`', ['exception' => $e]);
```
nickvergessen, line 120: return? Or can we proceed without valid settings?

lib/public/TextToImage/Exception/TaskNotFoundException.php
```diff
@@ -0,0 +1,28 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\TextToImage\Exception;
+
+class TaskNotFoundException extends TextToImageException {
```
nickvergessen, line 27:
```suggestion
/**
 * @since 28.0.0
 */
class TaskNotFoundException extends TextToImageException {
```

lib/public/TextToImage/Exception/TextToImageException.php
```diff
@@ -0,0 +1,28 @@
...
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\TextToImage\Exception;
+
+class TextToImageException extends \Exception {
```
nickvergessen, line 27:
```suggestion
/**
 * @since 28.0.0
 */
class TextToImageException extends \Exception {
```

core/Migrations/Version28000Date20230906104802.php
```diff
@@ -0,0 +1,92 @@
...
+				'notnull' => false,
+				'length' => 64,
+			]);
+			$table->addColumn('app_id', Types::STRING, [
+				'notnull' => true,
+				'length' => 32,
+				'default' => '',
+			]);
+			$table->addColumn('identifier', Types::STRING, [
+				'notnull' => true,
```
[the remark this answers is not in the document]
marcelklehr: gotta love oracle

lib/private/TextToImage/Manager.php
marcelklehr: It will use the first registered provider if the setting is not set or can't be read. We can change it to error out if the setting isn't valid, but I don't see a good reason.

lib/private/TextToImage/Db/TaskMapper.php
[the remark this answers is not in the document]
nickvergessen: Merged

core/Controller/TextToImageApiController.php
```diff
@@ -0,0 +1,219 @@
...
+		try {
+			$task = $this->textToImageManager->getUserTask($id, $this->userId);
+
+			$json = $task->jsonSerialize();
+
+			return new DataResponse([
+				'task' => $json,
+			]);
+		} catch (TaskNotFoundException) {
+			return new DataResponse(['message' => $this->l->t('Task not found')], Http::STATUS_NOT_FOUND);
```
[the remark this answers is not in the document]
nickvergessen: You added rate limit now, which is cool :)

However it's not the same as bruteforce protection which should happen on negative attempts, blocking the user from doing any more requests for 12h when they tried to find things they can not find.

But it's as easy to add, just an attribute on the method and a function call on the response before returning it:
https://docs.nextcloud.com/server/latest/developer_manual/basics/controllers.html#brute-force-protection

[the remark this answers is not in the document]
marcelklehr: Oooh, right. I always confuse the two...

core/Controller/TextToImageApiController.php
```diff
@@ -0,0 +1,241 @@
...
+	 * @param int $id The id of the task
+	 *
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTextToImageTask}, array{}>|DataResponse<Http::STATUS_NOT_FOUND|Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 * 200: Task returned
+	 * 404: Task not found
+	 */
+	#[PublicPage]
+	#[BruteForceProtection(action: 'not-found')]
+	#[BruteForceProtection(action: 'error')]
```
nickvergessen, line 117: Errors should not trigger BFP, only the not-found

Also the action should be specific to the endpoint, so e.g. `text2image` would be a good one here compared with `not-found`

marcelklehr: Oh. The docs are not clear on the scope of those action identifiers

lib/private/TextToImage/Db/Task.php
```diff
@@ -0,0 +1,120 @@
...
+		return array_combine(self::$columns, array_map(function ($field) {
+			return $this->{'get'.ucfirst($field)}();
+		}, self::$fields));
+	}
+
+	public static function fromPublicTask(OCPTask $task): Task {
+		/** @var Task $dbTask */
+		$dbTask = Task::fromParams([
+			'id' => $task->getId(),
+			'lastUpdated' => time(),
```
MB-Finski, line 93: Should this use new DateTime('now')?

marcelklehr:
I've made some small-ish additions to what @come-nc ,  @juliushaertl and @nickvergessen  reviewed:

- Multiple images can now be generated within one task
- Each provider must now expose its expected runtime for one task, this is also exposed by the OCS API per task
- I've added a Fake provider to the testing app which always returns the nextcloud logo

lib/public/TextToImage/Task.php
```diff
@@ -0,0 +1,212 @@
...
+		protected ?string $identifier = '',
+	) {
+	}
+
+	/**
+	 * @return IImage[]|null
+	 * @since 28.0.0
+	 */
+	final public function getOutputImages(): ?array {
+		$appData = \OC::$server->get(IAppDataFactory::class)->get('core');
```
marcelklehr, line 93: Not sure how I can avoid this to appease psalm.

marcelklehr:
Added the relevant AI settings:

image

lib/private/TextToImage/Db/Task.php
```diff
@@ -0,0 +1,116 @@
...
+		return array_combine(self::$columns, array_map(function ($field) {
+			return $this->{'get'.ucfirst($field)}();
+		}, self::$fields));
+	}
+
+	public static function fromPublicTask(OCPTask $task): Task {
+		/** @var Task $dbTask */
+		$dbTask = Task::fromParams([
+			'id' => $task->getId(),
+			'lastUpdated' => new DateTime('now'),
```
nickvergessen, line 97: `ITimeFactory::getDateTime()`?
Since you can't inject it via the constructor, you would have to inject it on this method. 😝

lib/private/TextToImage/Db/TaskMapper.php
```diff
@@ -0,0 +1,128 @@
...
+		return $this->findEntities($qb);
+	}
+
+	/**
+	 * @param int $timeout
+	 * @return Task[] the deleted tasks
+	 * @throws Exception
+	 */
+	public function deleteOlderThan(int $timeout): array {
+		$datetime = new DateTime();
```
nickvergessen, line 110:
```suggestion
		$datetime = $this->timeFactory->getDateTime();
```

lib/private/TextToImage/Db/TaskMapper.php
```diff
@@ -0,0 +1,128 @@
...
+		$deletedTasks = $this->findEntities($qb);
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->tableName)
+			->where($qb->expr()->lt('last_updated', $qb->createPositionalParameter($datetime, IQueryBuilder::PARAM_DATE)));
+		$qb->executeStatement();
+		return $deletedTasks;
+	}
+
+	public function update(Entity $entity): Entity {
+		$entity->setLastUpdated(DateTime::createFromImmutable($this->timeFactory->now()));
```
nickvergessen, line 125:
```suggestion
		$entity->setLastUpdated($this->timeFactory->getDateTime());
```

lib/private/TextToImage/Manager.php
```diff
@@ -0,0 +1,329 @@
...
+	public function __construct(
+		private IServerContainer $serverContainer,
+		private Coordinator $coordinator,
+		private LoggerInterface $logger,
+		private IJobList $jobList,
+		private TaskMapper $taskMapper,
+		private IConfig $config,
+		private IAppDataFactory $appDataFactory,
+	) {
+		$this->appData = $this->appDataFactory->get('core');
```
nickvergessen, line 64: You don't need to "save it" when you only need it in the constructor:

```suggestion
		IAppDataFactory $appDataFactory,
	) {
		$this->appData = $appDataFactory->get('core');
```

lib/private/TextToImage/RemoveOldTasksBackgroundJob.php
```diff
@@ -0,0 +1,78 @@
...
+	private IAppData $appData;
+
+	public function __construct(
+		ITimeFactory $timeFactory,
+		private TaskMapper $taskMapper,
+		private LoggerInterface $logger,
+		private IAppDataFactory $appDataFactory,
+	) {
+		parent::__construct($timeFactory);
+		$this->appData = $this->appDataFactory->get('core');
```
nickvergessen, line 51:
```suggestion
		IAppDataFactory $appDataFactory,
	) {
		parent::__construct($timeFactory);
		$this->appData = $appDataFactory->get('core');
```

lib/public/TextToImage/Task.php
nickvergessen:
```suggestion
		$appData = \OCP\Server::get(IAppDataFactory::class)->get('core');
```

lib/public/TextToImage/Task.php
```diff
@@ -0,0 +1,212 @@
...
+	/**
+	 * @return string|null
+	 * @since 28.0.0
+	 */
+	final public function getUserId(): ?string {
+		return $this->userId;
+	}
+
+	/**
+	 * @psalm-return array{id: ?int, status: 0|1|2|3|4, userId: ?string, appId: string, input: string, identifier: ?string, numberOfImages: int, completionExpectedAt: ?int}
```
nickvergessen, line 197: Btw for psalm you can do:
```suggestion
	 * @psalm-return array{id: ?int, status: self::STATUS_*, userId: ?string, appId: string, input: string, identifier: ?string, numberOfImages: int, completionExpectedAt: ?int}
```

Only the openapi extractor needs the plain numbers atm (so in `core/ResponseDefinitions.php`)

lib/private/TextToImage/Manager.php
```diff
@@ -0,0 +1,331 @@
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function scheduleTask(Task $task): void {
+		if (!$this->hasProviders()) {
+			throw new PreConditionNotMetException('No text to image provider is installed that can handle this task');
+		}
+		$this->logger->warning('Scheduling Text2Image Task');
```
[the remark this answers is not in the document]
marcelklehr: hah, copypasta..

lib/public/AppFramework/Bootstrap/IRegistrationContext.php
```diff
@@ -230,6 +231,16 @@ public function registerSpeechToTextProvider(string $providerClass): void;
 	 */
 	public function registerTextProcessingProvider(string $providerClass): void;
 
+	/**
+	 * Register a custom text2image provider class that provides the possibility to generate images
+	 * through the OCP\TextToImage APIs
+	 *
+	 * @param string $providerClass
+	 * @psalm-param class-string<ITextToImageProvider> $providerClass
+	 * @since 27.1.0
```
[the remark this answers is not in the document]
marcelklehr: copypasta

core/Controller/TextToImageApiController.php
```diff
@@ -0,0 +1,245 @@
...
+	#[PublicPage]
+	#[UserRateLimit(limit: 20, period: 120)]
+	#[AnonRateLimit(limit: 5, period: 120)]
+	public function schedule(string $input, string $appId, string $identifier = '', int $numberOfImages = 8): DataResponse {
+		$task = new Task($input, $appId, $numberOfImages, $this->userId, $identifier);
+		try {
+			try {
+				$this->textToImageManager->runOrScheduleTask($task);
+			} catch (\RuntimeException) {
+				// noop
```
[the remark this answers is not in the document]
marcelklehr: `runOrSchedule` dynamically decides if it should schedule the task or run it directly. This endpoint only represents the schedule part, though, so even if there is an error we return 200 for "Scheduling successful". If there is an error, it will be thrown, but will also be persisted in the status of the task object, which is returned. So, downstream code can still see from the status of the returned task that it failed.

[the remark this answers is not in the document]
come-nc: But PHP itself can throw RuntimeException in some cases and this would silence the problem, no?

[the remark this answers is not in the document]
marcelklehr: Yes, that's right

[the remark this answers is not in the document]
marcelklehr: Do you have a recommendation?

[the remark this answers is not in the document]
come-nc: Use your own exception class to mean «problem encountered and task status was already updated», and throw/catch this one.

lib/private/TextToImage/Manager.php
```diff
@@ -0,0 +1,328 @@
...
+				try {
+					$this->taskMapper->update(DbTask::fromPublicTask($task));
+				} catch (Exception $e) {
+					$this->logger->warning('Failed to update database after Text2Image error', ['exception' => $e]);
+				}
+				throw new TaskFailureException('Text2Image generation using provider "' . $provider->getName() . '" failed: ' . $e->getMessage(), 0, $e);
+			}
+		}
+
+		throw new TaskFailureException('Could not run task');
```
come-nc, line 202: In this case the task status is not set to failure, is it?

marcelklehr: Good catch!

marcelklehr: Fixed

lib/public/TextToImage/IProvider.php
```diff
@@ -0,0 +1,64 @@
...
+use RuntimeException;
+
+/**
+ * This is the interface that is implemented by apps that
+ * implement a text to image provider
+ * @since 28.0.0
+ */
+interface IProvider {
+	/**
+	 * The ID of this provider
```
come-nc, line 37:
```suggestion
	 * An arbitrary unique text string identifying this provider
```

come-nc: (maybe even give an example to set a pattern dev will use, or suggest to return self::class

come-nc (approved):

AndyScherzinger (approved):

marcelklehr:
:tada:

marcelklehr:
Failures seem unrelated, please merge :pray:

ChristophWurst:
the new public API needs docs

PR #40329, opened 2023-09-07
theming: Revert broken SVG optimization in default-source.svg

provokateurin:
## Summary

https://github.com/nextcloud/server/commit/f2afccfd569b7c6de0234514a321409dc238dc2a did some optimizations which inlined the included jpgs into the svg. This will make the image no longer update when the jpg files are updated and also bloats the svg itself. I'm pretty sure this was not intended.

## Checklist


bigcat88 (approved):

miaulalala (approved):

nickvergessen:
The next automatic run will bring this back. So the script at https://github.com/nextcloud/server/blob/master/build/image-optimization.sh needs adjusting I guess (could start with ignoring the file for now, future step find a parameter that skips "embedding")

provokateurin:
Looking at the other "optinizations" it doesn't seem usefule to have that script. Changing the order of xml element attributes does nothing at all.

Issue #40289, opened 2023-09-06
Custom permissions should show permissions in subline


AndyScherzinger:
<img width="366" alt="2023-09-06 12_10_41-Shared with others - Nextcloud – Mozilla Firefox" src="https://github.com/nextcloud/server/assets/1315170/0983de6b-087d-41ce-a154-6ac51b3ab99e">

* customer permissions should be in a separate line 
* icons are mission (optional - only to be done if quick fix, else just the subline)

juliushaertl:
Seems only an issue on stable27, maybe caused by the different version of `@nextcloud/vue` being used.

PR #40343, opened 2023-09-07
[stable27] 40289 fix appearance

nfebe:
* Resolves: https://github.com/nextcloud/server/issues/40289

| Before | Currently |  Without Icons |
| --------- | ------------ | -------------- |
|2023-09-06 12_10_41-Shared with others - Nextcloud – Mozilla Firefox |Screenshot from 2023-09-08 00-23-51 | Screenshot from 2023-09-08 00-18-11 |


AndyScherzinger:
@juliushaertl @Fenn-CS nice 👍 Can we left-align the content so we are closer to the mockups and v28 as described in https://github.com/nextcloud/server/issues/26691 ?

Fenn-CS:
> @juliushaertl @Fenn-CS nice 👍 Can we left-align the content so we are closer to the mockups and v28 as described in https://github.com/nextcloud/server/issues/26691 ?

That was the default, option and I even got a screenshot for it earlier, I just forgot to add it here. The reason, I centered it with the icons, is that, I realized it was centered without them on 27... But yeah done.

Screenshot from 2023-09-08 00-12-15

cc: @AndyScherzinger

AndyScherzinger (approved):

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -951,11 +946,25 @@ export default {
...
+							flex: 10%;
+						}
+
+						&:last-of-type {
+							flex: 90%
+						}
 					}
+
+					small {
+                       flex-basis: 100% !important;
```
skjnldsv, line 965:
```suggestion
						flex-basis: 100% !important;
```

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -951,11 +946,25 @@ export default {
...
 						display: flex;
-						flex-direction: column;
+
+						&:first-of-type {
+							flex: 10%;
+						}
+
+						&:last-of-type {
+							flex: 90%
+						}
```
skjnldsv, line 961: Ideally this should be done in the vue components with a proper slot like you did initially :)

artonge (approved):

jancborchardt (commented):
Just design feedback, not blocking:
- [x] The subline alignment of "Create, update, delete, etc." is too far to the left, it should ideally be left-aligned with the "Custom permissions" text as per the mockup. If the subline is too long it should ellipsize after 1 line

apps/files_sharing/src/views/SharingDetailsTab.vue
Fenn-CS: Yeah @skjnldsv, this changes where already made in the component, but where not back-ported (not sure why), so went this route.

Fenn-CS:
> Just design feedback, not blocking:
> 
>     The subline alignment of "Create, update, delete, etc." is too far to the left, it should ideally be left-aligned with the "Custom >permissions" text as per the mockup. If the subline is too long it should ellipsize after 1 line

Screenshot from 2023-09-11 13-25-59



cc: @jancborchardt

jancborchardt (approved):
Nice! :)

sorbaugh (approved):

PR #40340, opened 2023-09-07
Fix access to fileInfo in VersionTab.vue

artonge:
Missed after https://github.com/nextcloud/server/pull/40296 and eslint didn't catch it

PR #40348, opened 2023-09-08
fix: Use proper fileInfo when trying to preview versions

juliusknorr:
Fix the regression from https://github.com/nextcloud/server/pull/40296/files#r1319816091 which broke opening version previews with viewer.

Catched by text cypress tests


artonge:
Or https://github.com/nextcloud/server/pull/40340
Sorry for the trouble

juliushaertl:
No worries, thanks for the fix, I actually forgot to check if there was a pr already 🙈

PR #40363, opened 2023-09-11
improve logic for getting the watcher for a shared external storage

icewind1991:
with the new-ish lazy mount loading the underlying mount might not be loaded yet, so ask the mount cache instead

To test

- Setup an instance with user A and user B
- mount external storage M for user A
- have user A share M with B
- modify the data in M directly (bypassing NC)
- as B check if the file listing is updated


come-nc (approved):
(did not test)

skjnldsv (approved):

PR #40368, opened 2023-09-11
feat: update sign up link in share pages

luka-nextcloud:
## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/TemplateLayout.php
+++ b/lib/private/TemplateLayout.php
@@ -47,11 +47,13 @@ use OC\Search\SearchQuery;
 use OC\Template\CSSResourceLocator;
 use OC\Template\JSConfigHelper;
 use OC\Template\JSResourceLocator;
+use OCP\App\IAppManager;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\Defaults;
 use OCP\IConfig;
 use OCP\IInitialStateService;
 use OCP\INavigationManager;
+use OCP\IURLGenerator;
 use OCP\IUserSession;
 use OCP\Support\Subscription\IRegistry;
 use OCP\Util;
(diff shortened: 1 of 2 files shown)


lib/private/TemplateLayout.php
```diff
@@ -196,6 +198,14 @@ public function __construct($renderAs, $appId = '') {
 				$showSimpleSignup = false;
 			}
 			$this->assign('showSimpleSignUpLink', $showSimpleSignup);
+
+			$urlGenerator = \OCP\Server::get(IURLGenerator::class);
+			$signUpLink = $this->config->getSystemValueString('registration_link', $urlGenerator->getAbsoluteURL('/index.php/signup/'));
```
juliushaertl, line 203:
```suggestion
			$signUpLink = $this->config->getSystemValueString('registration_link', 'https://nextcloud.com/signup/');
```

lib/private/TemplateLayout.php
```diff
@@ -196,6 +198,14 @@ public function __construct($renderAs, $appId = '') {
...
+			$appManager = \OCP\Server::get(IAppManager::class);
```
juliushaertl, line 204: We should also make sure that `showSimpleSignUpLink` assigned above is still true if a custom URL is configured.

juliusknorr (changes requested):
Two small additions, otherwise this seems fine 👍

juliusknorr (approved):

lib/private/TemplateLayout.php
```diff
@@ -195,7 +197,21 @@ public function __construct($renderAs, $appId = '') {
 			if ($showSimpleSignup && $subscription->delegateHasValidSubscription()) {
 				$showSimpleSignup = false;
 			}
+
+			$urlGenerator = \OCP\Server::get(IURLGenerator::class);
+			$defaultSignUpLink = 'https://nextcloud.com/signup/';
+			$signUpLink = $this->config->getSystemValueString('registration_link', $defaultSignUpLink);
+			if ($signUpLink !== $defaultSignUpLink) {
+				$showSimpleSignup = true;
```
icewind1991, line 205: do we want to set this back to true if is was set to false on line 198?

luka-nextcloud: Yes, it should be. As juliushaertl has suggested [here](https://github.com/nextcloud/server/pull/40368/files/e92d396200e5536b2daa3d78c10e594b6ed5538e#r1322043083).

mejo- (approved):
Looks sensible :ok_hand:

Altahrim (approved):

PR #40376, opened 2023-09-12
test: add tests for dns pin middleware

kesselb:
## TODO

- [x] CI

## Checklist

### DIFF
--- a/lib/private/Http/Client/DnsPinMiddleware.php
+++ b/lib/private/Http/Client/DnsPinMiddleware.php
@@ -55,7 +55,7 @@ class DnsPinMiddleware {
 		$second = array_pop($labels);
 
 		$hostname = $second . '.' . $top;
-		$responses = dns_get_record($hostname, DNS_SOA);
+		$responses = $this->dnsGetRecord($hostname, DNS_SOA);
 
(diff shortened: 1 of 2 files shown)


miaulalala (approved):

lib/private/Http/Client/DnsPinMiddleware.php
```diff
@@ -104,6 +104,13 @@ private function dnsResolve(string $target, int $recursionCount) : array {
 		return $targetIps;
 	}
 
+	/**
+	 * Wrapper for dns_get_record
+	 */
+	protected function dnsGetRecord(string $hostname, int $type): array|false {
```
ChristophWurst, line 110: works but IMO it feels cleaner if this is a simple class that is injected

ChristophWurst (approved):

lib/private/Http/Client/DnsPinMiddleware.php
kesselb: That was my first approach. Discarded it because it felt like overengineering.

ChristophWurst: But I like overengineering :smiling_face_with_tear:

PR #40393, opened 2023-09-13
Display user status by the side in sharing flow

nfebe:
The quick share drop down now takes the place where the user status would show in the past, so we have the option of removing it (as it's available in the profile dropdown) or putting it at the side as such.

It's put on the side without the icon, as the status icon appears on the avatar when set.

| Before | Now |
| --------  | ------ |
| Screenshot from 2023-09-13 11-32-46 | Screenshot from 2023-09-13 12-00-20 |


apps/files_sharing/src/components/SharingEntry.vue
```diff
@@ -35,12 +35,9 @@
 				:aria-label="tooltip"
 				:href="share.shareWithLink"
 				class="sharing-entry__desc">
-				<span>{{ title }}<span v-if="!isUnique" class="sharing-entry__desc-unique"> ({{
-					share.shareWithDisplayNameUnique }})</span></span>
-				<p v-if="hasStatus">
-					<span>{{ share.status.icon || '' }}</span>
-					<span>{{ share.status.message || '' }}</span>
-				</p>
+				<span>{{ title }} <small v-if="hasStatus">({{ share.status.message || '' }})</small>
```
juliushaertl, line 38: I think we should keep the text size at the same level so just use a span tag, but maybe @jancborchardt can confirm. Also the status should probably be after the unique display name part below.

AndyScherzinger (approved):

apps/files_sharing/src/components/SharingEntry.vue
jancborchardt: Yes, text size should be same size for accessibility. :)
@Fenn-CS also make sure that long lines will be correctly ellipsized.

Fenn-CS: > Yes, text size should be same size for accessibility. :)

I do not agree, while the the font size of elements greatly affects accessibility, here the font-size is not customized it is just the `<small>` element which is not considered inherently inaccessible.

- `<small>` is recommended for content that is genuinely less important or supplementary to the main content, such as fine print or legal disclaimers. Which is 100% true for this case.

- It's also recommended to avoid using <small> for text that is critical for understanding or interacting with the page. Important information should be presented in a clear and legible manner. Which is 100% true for this case. 

- The default font size appears to 15 and `small` seems to be calculated at `12.5` which is bigger than the recommended meaning of `12`

As a matter of fact, `<span>` is semantically wrong in this context as it would give the side note same important as the name of the person to whom the share pertains : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/small and can in turn affect accessibility negatively if a screen reader for example reads that as part of the share-name.

Fenn-CS:
Updates that take into consideration disambiguation text as well as wrap. (Also blurred the text, just like the disambiguation text)

| Without wrap | With wrap |
| --------  | ------ |
| Screenshot from 2023-09-13 15-07-07 | Screenshot from 2023-09-13 15-07-48 |

juliusknorr (approved):

AndyScherzinger:
Looks good, just some minor issues as mentioned partially already

* is Ellipsizing implemented?
* text size of the parenthesis text is not yet unified, see status vs. mail address for example
* is the grey tone accessible? Given the font size the contrast ratio needs to be 1:4,5 (relevant more for Master than stable27)

Fenn-CS:
> is Ellipsizing implemented?

So no wrap? things wrap in a decent way when text is too long as shown in the last screen shots.

> text size of the parenthesis text is not yet unified, see status vs. mail address for example
> is the grey tone accessible? Given the font size the contrast ratio needs to be 1:4,5 (relevant more for Master than stable27)

`color: var(--color-text-maxcontrast);` is used.

cc: @AndyScherzinger

jancborchardt:
Sorry but with wrapping it looks way too wonky, we shouldn't do that. It needs to be ellipsized to one line. :)

Also, the gap between the first line and the permission line is too big for users. For "Share link" it looks good, but for the users below it's larger.

skjnldsv:
I have no idea why, but this broke scrolling on Files.
image

PR #40394, opened 2023-09-13
fix mimetype not being updated when changing file extention on objectstore

icewind1991:
With object store the source entry no longer exists once we hit the updater, so the old code path for changing the mimetype wasn't triggered.

### DIFF
--- a/lib/private/Files/Cache/Updater.php
+++ b/lib/private/Files/Cache/Updater.php
@@ -205,21 +205,25 @@ class Updater implements IUpdater {
 
 		$sourceInfo = $sourceCache->get($source);
 
+		$sourceExtension = pathinfo($source, PATHINFO_EXTENSION);
+		$targetExtension = pathinfo($target, PATHINFO_EXTENSION);
+		$targetIsTrash = preg_match("/^d\d+$/", $targetExtension);
+
 		if ($sourceInfo !== false) {
 			if (!$this->storage->instanceOfStorage(ObjectStoreStorage::class)) {
 				$operation($sourceCache, $sourceInfo);
 			}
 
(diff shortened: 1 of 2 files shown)


lib/private/Files/Cache/Updater.php
```diff
@@ -194,6 +194,10 @@ public function renameFromStorage(IStorage $sourceStorage, $source, $target) {
 
 		$sourceInfo = $sourceCache->get($source);
 
+		$sourceExtension = pathinfo($source, PATHINFO_EXTENSION);
+		$targetExtension = pathinfo($target, PATHINFO_EXTENSION);
+		$targetIsTrash = preg_match("/d\d+/", $targetExtension);
```
come-nc, line 199: Seems fragile, shouldn’t this test if target is in files_trashbin instead?
Or at least have a regex with `^` or `$`, not something that can match anything in the middle.
https://www.file-extension.info/format/d3d This format for instance would match the regex while not being trash.

lib/private/Files/Cache/Updater.php
```diff
@@ -209,12 +213,16 @@ public function renameFromStorage(IStorage $sourceStorage, $source, $target) {
 			$targetExtension = pathinfo($target, PATHINFO_EXTENSION);
 			$targetIsTrash = preg_match("/d\d+/", $targetExtension);
```
come-nc, line 214: This is now done twice?

come-nc (approved):

provokateurin (approved):

solracsf:
This is being backported to old versions, shouldn't 28/29/30 be backported too? 🤔

provokateurin:
True, the original PR was just quite old

PR #40409, opened 2023-09-14
refactor(f2v): Migrate unread comments action to the new FileAction API

lhsazevedo:
* Ref: #39914 

## Summary
Migrate inline unread comments action to the new FileAction API.

To display the inline action as a icon only button, I had to remove the `force-name` prop from the `<NcActions />` in the `FileEntry.vue`, otherwise the `NcButton` is rendered with text as well. But I imagine the action buttons were meant to be icon only, so I guess this was unintentionally added? 

## Screenshots
image
image

## TODO

- [ ] Hide action after opening the sidebar. Should I use `Vue.set()` to overwrite the `comments-unread` attribute and rely on reactivity to disable the action?

## Checklist

- ~Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required~
- ~[Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)~


skjnldsv (approved):

skjnldsv:
> To display the inline action as a icon only button, I had to remove the force-name prop from the <NcActions /> in the FileEntry.vue, otherwise the NcButton is rendered with text as well. But I imagine the action buttons were meant to be icon only, so I guess this was unintentionally added?

Maybe, I don't recall. We'll see if that causes issues later on! Let get this merged :)

AndyScherzinger (approved):

icewind1991 (approved):

PR #40413, opened 2023-09-14
Add in:users as a search filter to limit searches to users

### LINKED ISSUE #40287: Add `in:users` as a search filter to limit searches to users
### How to use GitHub

Currently when you have large installation in terms of users and you need to search for a user, you can only do that via unified search. This works and filters down the list of users in case you are at `*/settings/users` on your instance. The downside is that this then also trigger searches in all available search providers.

Provide a app-filter `in:users` that can be added to the search query like it can be done for other apps like files and within the actual user search just remove any occurrence of `in:users` and just search with the rest of the search string (trimmed). Due to the in-filter all search providers will be skipped which is what is the desired behavior.

Implement an in-app search in users like present in talk but that would be way more work.

---

This shall be backported down to v25

sorbaugh:
* Resolves: #40287

## Summary
Add in:users filter to search bar via SearchProvider to have a mechanism to limit the search load in settings/users.

## TODO

- [x] Add new UserSearch Provider
- [x] Trim search query before executing actual search

## Checklist


apps/settings/lib/Search/UserSearch.php
```diff
@@ -0,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2020 Joas Schilling <coding@schilljs.com>
+ *
+ * @author Joas Schilling <coding@schilljs.com>
+ * @author John Molakvoæ <skjnldsv@protonmail.com>
+ * @author Morris Jobke <hey@morrisjobke.de>
```
artonge, line 10: Needs to be adapted with your name

apps/settings/lib/Search/UserSearch.php
```diff
@@ -0,0 +1,106 @@
...
+								IUserManager $userManager,
+								IAccountManager $accountManager,
+								IL10N $l) {
+		$this->settingsManager = $settingsManager;
+		$this->groupManager = $groupManager;
+		$this->urlGenerator = $urlGenerator;
+		$this->userManager = $userManager;
+		$this->accountManager = $accountManager;
+		$this->l = $l;
+	}
```
artonge, line 73: We can now simplify that by declaring the class properties in the constructor arguments like so: 
https://github.com/nextcloud/activity/pull/1285/files#diff-523ecbb0068bfbac78757e8bea247f4216c49b6508c44e0039be8ddda8d631dcR35-R37

artonge: As said, be careful if this needs to be backported to 25, the new syntax might not be compatible.

sorbaugh: Suuuper late reply but yes, since this must be backported to 25 I think we should leave like this.

artonge (commented):
Tested and the code looks good.
Still need to address the copyright comment.

Enjoy the rebase !

apps/settings/lib/Search/UserSearch.php
```diff
@@ -0,0 +1,105 @@
...
+								IGroupManager $groupManager,
+								IURLGenerator $urlGenerator,
+								IUserManager $userManager,
+								IAccountManager $accountManager,
+								IL10N $l) {
+		$this->settingsManager = $settingsManager;
+		$this->groupManager = $groupManager;
+		$this->urlGenerator = $urlGenerator;
+		$this->userManager = $userManager;
+		$this->accountManager = $accountManager;
```
blizzz, line 70: Maybe I am overseeing something, but they are all not being used?

apps/settings/lib/Search/UserSearch.php
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	/**
+	 * @inheritDoc
+	 */
+	public function search(IUser $user, ISearchQuery $query): SearchResult {
+
+		return SearchResult::complete(
+			$this->l->t('Users'),
+			[]
```
blizzz, line 102: Do I understand how it works correctly, that this search plugin intentionally returns an empty result set? For users page it make sense as results are shown on the content and not in the pull down. Maybe I am misinterpreting how that search works though.

The naming would block potential valid future searches for users, although there is the contacts menu for this. It's acceptable, but want to point it out.

sorbaugh: Yes I agree this is somewhat strange. ATM the main goal is for a customer to more efficiently search for contacts without triggering everything. You correctly pointed out that the goal is to not have a list of users shown in the pulldown. See commit where how this used to be solved (and how we want to revisit in the future)

https://github.com/nextcloud/server/pull/40413/commits/54a21d431bf097feda9eedf552be19410f0a9df1

apps/settings/lib/Search/UserSearch.php
sorbaugh: Cleaned this one up, thanks for pointing it out 👍

blizzz (approved):
Just the copyright info ;)

AndyScherzinger (approved):

nfebe (approved):

artonge (approved):

nickvergessen:
/backport ab81cd1e1e0685952d2c1818160954cbbaa9fa6b,65c70d7cc6425af0bf202d70c67eabd579d913a6,c38bafba9e13e4e735ed2d6698b6c4f5d9efc7d5,a3a599855e20b60627ac4e65fed8fdfa6b2c04f8,f66e4ee07246b285f79173c300de943b3a857bad,bdf0fe6a03beccc58a119f05d9cdb7ab643ebcfd to stable27

blizzz:
Backportbot seems to have day off :-/ fastest to do it manually it seems

PR #40422, opened 2023-09-14
fix: skip mtime hack for tag search

kesselb:
## TODO

- [x] CI

## Checklist


icewind1991:
I'm still not convinced that this will lead to better performance in most cases, having mysql use the index for the mtime `ORDER` (seems to) stop it from using any index for the `WHERE` clauses.

And while there are certainly cases where this is fine, I believe that in more cases the query gains more performance from being able to use the the index for filtering the storages to search in.

PR #40423, opened 2023-09-14
enh(db): provide database providers via IDB API

### LINKED ISSUE #30877: Wrap instance reporting for database schema in an API
## Problem
Currently, we're checking for the Database Type of an instance something like this:

```php
$schema->getDatabasePlatform() instanceof PostgreSQL94Platform
 ```

If we ever change the underlying structure, these checks might break.

## Proposed solution

Add a new API method for `ISchemaWrapper` that will return the database type as a string and a few constants that map to the [supported databases](https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/linux_database_configuration.html#:~:text=Nextcloud%20requires%20a%20database%20in,PostgreSQL) to compare the Database string too:

```php

public CONST PostgreSQL94Platform = 'PostgreSQL94Platform';
....

public function getDatabaseType(): string;
```

[Context](https://github.com/nextcloud/mail/pull/5935/files#r792481561)

miaulalala:
## Summary

To avoid leaking internals (OC), wrap the getDatabasePlatform and provide the associated constants

* Resolves: https://github.com/nextcloud/server/issues/30877

## Checklist

- [ ] ~~[Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)~~

### DIFF
--- a/lib/private/DB/ConnectionAdapter.php
+++ b/lib/private/DB/ConnectionAdapter.php
@@ -27,6 +27,10 @@ namespace OC\DB;
 
 use Doctrine\DBAL\Exception;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Platforms\OraclePlatform;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
+use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Schema\Schema;
 use OC\DB\Exceptions\DbalException;
 use OCP\DB\IPreparedStatement;
@@ -241,5 +245,20 @@ class ConnectionAdapter implements IDBConnection {
 
 	public function getInner(): Connection {
 		return $this->inner;
+	}
+
+	public function getDatabaseProvider(): string {
+		$platform = $this->inner->getDatabasePlatform();
+		if ($platform instanceof MySQLPlatform) {
+			return IDBConnection::PLATFORM_MYSQL;
+		} elseif ($platform instanceof OraclePlatform) {
+			return IDBConnection::PLATFORM_ORACLE;
+		} elseif ($platform instanceof PostgreSQLPlatform) {
+			return IDBConnection::PLATFORM_POSTGRES;
+		} elseif ($platform instanceof SqlitePlatform) {
+			return IDBConnection::PLATFORM_SQLITE;
+		} else {
+			throw new \Exception('Database ' . $platform::class . ' not supported');
+		}
 	}
 }
--- a/lib/public/IDBConnection.php
+++ b/lib/public/IDBConnection.php
@@ -45,6 +45,18 @@ use OCP\DB\QueryBuilder\IQueryBuilder;
  * @since 6.0.0
  */
 interface IDBConnection {
+	/* @since 28.0.0 */
+	public const PLATFORM_MYSQL = 'mysql';
+
+	/* @since 28.0.0 */
+	public const PLATFORM_ORACLE = 'oracle';
+
+	/* @since 28.0.0 */
+	public const PLATFORM_POSTGRES = 'postgres';
+
+	/* @since 28.0.0 */
+	public const PLATFORM_SQLITE = 'sqlite';
+
 	/**
 	 * Gets the QueryBuilder for the connection.
 	 *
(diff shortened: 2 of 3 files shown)


lib/public/IDBConnection.php
```diff
@@ -45,6 +45,11 @@
  * @since 6.0.0
  */
 interface IDBConnection {
+	public const MYSQL = 'mysql';
+	public const POSTGRES = 'postgres';
+	public const ORACLE = 'oracle';
```
kesselb, line 50:
```suggestion
	public const ORACLE = 'oracle';
	public const POSTGRES = 'postgres';
```

lib/public/IDBConnection.php
```diff
@@ -45,6 +45,11 @@
  * @since 6.0.0
  */
 interface IDBConnection {
+	public const MYSQL = 'mysql';
```
ChristophWurst, line 48:
```suggestion
	public const PLATFORM_MYSQL = 'mysql';
```

so the constants can be used in typing

all constants need an `@since` tag

lib/public/IDBConnection.php
```diff
@@ -339,4 +344,13 @@ public function createSchema(): Schema;
 	 * @since 13.0.0
 	 */
 	public function migrateToSchema(Schema $toSchema): void;
+
+	/**
+	 * Returns the database provider name
+	 *
+	 * @throws Exception
```
ChristophWurst, line 351: Avoid throwing if possible

lib/public/IDBConnection.php
```diff
@@ -339,4 +344,13 @@ public function createSchema(): Schema;
...
+	 * @link https://github.com/nextcloud/server/issues/30877
```
ChristophWurst, line 352: document return type as `IDBConnection::PLATFORM_*`

lib/private/DB/ConnectionAdapter.php
```diff
@@ -242,4 +246,19 @@ public function migrateToSchema(Schema $toSchema): void {
...
+		if ($platform instanceof MySQLPlatform) {
+			return IDBConnection::MYSQL;
+		} elseif ($platform instanceof PostgreSQLPlatform) {
+			return IDBConnection::POSTGRES;
+		} elseif ($platform instanceof OraclePlatform) {
+			return IDBConnection::ORACLE;
+		} elseif ($platform instanceof SqlitePlatform) {
+			return IDBConnection::SQLITE;
+		} else {
+			return $platform::class;
```
kesselb, line 261: I'm not sure about it.

lib/public/IDBConnection.php
miaulalala: well if the db isn't there then the exception can't be handled any way so I won't catch it and let it bubble up.

lib/private/DB/ConnectionAdapter.php
ChristophWurst: this should be impossible to reach, no? you could drop the last elseif and return sqlite in else

ChristophWurst:
This adds a new public API so it does need documentation

lib/private/DB/ConnectionAdapter.php
miaulalala: well maybe? There might be people out there that use NC with DB2 or MsSQL, so it would be weird to return sqlite IMHO.

icewind1991: just throw an error for those, they are not supported

icewind1991 (approved):

kesselb (approved):

lib/private/DB/ConnectionAdapter.php
```diff
@@ -242,4 +246,19 @@ public function migrateToSchema(Schema $toSchema): void {
...
+		if ($platform instanceof MySQLPlatform) {
+			return IDBConnection::PLATFORM_MYSQL;
+		} elseif ($platform instanceof OraclePlatform) {
+			return IDBConnection::PLATFORM_ORACLE;
+		} elseif ($platform instanceof PostgreSQLPlatform) {
+			return IDBConnection::PLATFORM_POSTGRES;
+		} elseif ($platform instanceof SqlitePlatform) {
+			return IDBConnection::PLATFORM_SQLITE;
+		} else {
+			throw new \Exception('Database provider not supported');
```
ChristophWurst, line 261: nit: include platform in string so the error message

ChristophWurst (approved):

ChristophWurst (approved):

PR #40428, opened 2023-09-14
Multiple bug fix in users list

artonge:
- Do not display userCount when it is equal to 0
- Reset loading state on UserRow mount
- Do not preload potential manager list on UserRow mount
- Do not manipulate userCount when it is equal to 0


Pytal:
Pushed a ux fix to show initial loading state of managers

Pytal (approved):

susnux (approved):

come-nc:
@Pytal Could you backport this to stable27? Not all commits will be needed I think but some of them applies the same I think.

PR #40439, opened 2023-09-15
Make logo in header clickable in the shared file view

Koc:
## Summary
Now we use same `logo_url` settings for shared file view that are used for normal user view.

image

## Open questions
- How to deal if `logo_url` is not configured? Use home link (like now) or make logo non-clickable?

## How to test
1. update `config/config.php` with
    ```php
      'logo_url' => 'https://google.com',
    ```
1. share file by creating public link
1. open shared link
1. click on logo

## Checklist

### DIFF
--- a/core/templates/layout.public.php
+++ b/core/templates/layout.public.php
@@ -38,14 +38,22 @@ p($theme->getTitle());
 
 	<header id="header">
 		<div class="header-left">
-			<div class="logo logo-icon svg"></div>
-			<span id="nextcloud" class="header-appname">
+			<div id="nextcloud" class="header-appname">
+				<?php if ($_['logoUrl']): ?>
+					<a href="<?php print_unescaped($_['logoUrl']); ?>"
+					   aria-label="<?php p($l->t('Go to %s', [$_['logoUrl']])); ?>">
+						<div class="logo logo-icon"></div>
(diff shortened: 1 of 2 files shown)


artonge (approved):

nfebe (approved):

core/templates/layout.public.php
```diff
@@ -38,14 +38,18 @@
 
 	<header id="header">
 		<div class="header-left">
-			<div class="logo logo-icon svg"></div>
-			<span id="nextcloud" class="header-appname">
+			<div id="nextcloud" class="header-appname">
+				<a href="<?php print_unescaped($_['logoUrl'] ?: link_to('', 'index.php')); ?>"
```
skjnldsv, line 42: Falling back to index.php seems counter intuitive.
You're in a public page, redirecting will ask the current user to log in. WHich make no sense in terms of context.

I would only set the link if logoUrl is set, otherwise have no link at all.

skjnldsv (changes requested):

core/templates/layout.public.php
Koc: done

susnux (approved):

skjnldsv (approved):

PR #40443, opened 2023-09-15
fix(ldap): store last known user groups

blizzz:
* Resolves: Share recipients loose groupfolder-based shares when the sharer is an LDAP user and was removed from the directory

## Summary

The LDAP groups of an LDAP user are fetched live, unless they are cached in Redis or similar. When an LDAP user is being deleted on the LDAP directory, the resulting groups are empty, thus the user may use access to certain files.

When LDAP users are removed on the directy, we still perform normal operations, e.g. shared files are still accessible to the recipients. This works with regular local shares, but the above scenario requires to keep track of the groups.

This approach here is a minimal fix, and not ideal in some aspects, but rather simple and easy to backport. 

It is a bit cumbersome to do it connection/configuration-based, but the Group_Proxy can be kept clean and lean this way. Still tainting the Group backend with User bits cannot be prevented.

If I would create the whole logic again, from scratch you could say, we still have the group management abstraction method entry point in the group proxy. There, I would for regular operations directly return the known groups (apart of initially), and not query LDAP. he goal would be to provide fast results to the user. Then, the user sync background job would be extended to also pull and store group information. The information would not be very up to date, the refresh happens twice a day. It should be accompanied with a way to signal Nextcloud to update a user (we have this need anyway). The Provisioning API could get an endpoint and emit an event RequestUpdateExternalUserDataEvent or something like this. Basically analogue to occ ldap:check-user --update (which might need to be extended on the group memberships as well).
But being a behavorial change, a solution like this is only possible for a major release. It could be an opportunity to adjust any other user- and group method to work like that.

@come-nc what do you think? Is this approach acceptable for you, also to backporting? 

## TODO

- [ ] discuss

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Group_LDAP.php
+++ b/apps/user_ldap/lib/Group_LDAP.php
@@ -53,6 +53,7 @@ use OCP\Group\Backend\ABackend;
 use OCP\Group\Backend\IDeleteGroupBackend;
 use OCP\Group\Backend\IGetDisplayNameBackend;
 use OCP\IConfig;
+use OCP\IUserManager;
 use OCP\Server;
 use Psr\Log\LoggerInterface;
 use function json_decode;
@@ -75,8 +76,14 @@ class Group_LDAP extends ABackend implements GroupInterface, IGroupLDAP, IGetDis
 	 */
 	protected string $ldapGroupMemberAssocAttr;
 	private IConfig $config;
+	private IUserManager $ncUserManager;
 
-	public function __construct(Access $access, GroupPluginManager $groupPluginManager, IConfig $config) {
+	public function __construct(
+		Access $access,
+		GroupPluginManager $groupPluginManager,
+		IConfig $config,
+		IUserManager $ncUserManager
+	) {
 		$this->access = $access;
 		$filter = $this->access->connection->ldapGroupFilter;
 		$gAssoc = $this->access->connection->ldapGroupMemberAssocAttr;
@@ -91,6 +98,7 @@ class Group_LDAP extends ABackend implements GroupInterface, IGroupLDAP, IGetDis
 		$this->logger = Server::get(LoggerInterface::class);
 		$this->ldapGroupMemberAssocAttr = strtolower((string)$gAssoc);
 		$this->config = $config;
+		$this->ncUserManager = $ncUserManager;
 	}
 
 	/**
@@ -445,6 +453,7 @@ class Group_LDAP extends ABackend implements GroupInterface, IGroupLDAP, IGetDis
 	public function getUserGidNumber(string $dn) {
 		$gidNumber = false;
 		if ($this->access->connection->hasGidNumber) {
+			// FIXME: when $dn does not exist on LDAP anymore, this will be set wrongly to false :/
 			$gidNumber = $this->getEntryGidNumber($dn, $this->access->connection->ldapGidNumber);
 			if ($gidNumber === false) {
 				$this->access->connection->hasGidNumber = false;
@@ -659,6 +668,25 @@ class Group_LDAP extends ABackend implements GroupInterface, IGroupLDAP, IGetDis
 		return false;
 	}
 
+	private function isUserOnLDAP(string $uid): bool {
+		// forces a user exists check - but does not help if a positive result is cached, while group info is not
+		$ncUser = $this->ncUserManager->get($uid);
+		if ($ncUser === null) {
+			return false;
+		}
+		$backend = $ncUser->getBackend();
+		if ($backend instanceof User_Proxy) {
+			// ignoring cache as safeguard (and we are behind the group cache check anyway)
+			return $backend->userExistsOnLDAP($uid, true);
+		}
+		return false;
+	}
+
+	protected function getCachedGroupsForUserId(string $uid): array {
+		$groupStr = $this->config->getUserValue($uid, 'user_ldap', 'cached-group-memberships-' . $this->access->connection->getConfigPrefix(), '[]');
+		return json_decode($groupStr) ?? [];
+	}
+
 	/**
 	 * This function fetches all groups a user belongs to. It does not check
 	 * if the user exists at all.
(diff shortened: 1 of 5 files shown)


come-nc:
I do not get what this is about. Can you write a step by step scenario of what happens with current code which would happen differently with your idea?

We now have a store of known group<>user association in oc_ldap_group_membership since https://github.com/nextcloud/server/pull/39446 , and have a `check-group --update` similar to `check-user --update`.

There is still the discrepancy of using live membership from LDAP while announced memberships (through events) are done by the background job.
I try to improve on that in https://github.com/nextcloud/server/pull/40367 which checks at login if all groups of the user have been announced (which does not add LDAP requests because group membership is fetched at login anyway).

come-nc:
> > Can you write a step by step scenario of what happens with current code which would happen differently with your idea?
> 
> 1. Have a group folder with content.
> 2. Have no memcache enabled (to not have to wait for TTL running out)
> 3. Have LDAP users Alice with access to the group folder, and Bob without access to the group folder
> 4. As Alice, share a folder from the group folder with Bob
> 5. As Bob, confirm you have received that share and can access it
> 6. Delete Alice from LDAP
> 7. Run `occ ldap:check-user?  against Alice
> 8. As Bob, try to access the share now
> 
> Expected: you still can access the share
> 
> Actual: The folder is visible, cannot be entered (no content shown, error instead).

And which part is misbehaving here, is it a listener on the user delete which is suppose to update the share somehow?

If I re-read your first message, the problem is that once the user is deleted we do not know in which groups they were before deletion? If so, we can fetch this information from `OCA\User_LDAP\Db\GroupMembershipMapper` now I think.

blizzz:
> > > Can you write a step by step scenario of what happens with current code which would happen differently with your idea?
> > 
> > 
> > 
> > 1. Have a group folder with content.
> > 2. Have no memcache enabled (to not have to wait for TTL running out)
> > 3. Have LDAP users Alice with access to the group folder, and Bob without access to the group folder
> > 4. As Alice, share a folder from the group folder with Bob
> > 5. As Bob, confirm you have received that share and can access it
> > 6. Delete Alice from LDAP
> > 7. Run `occ ldap:check-user?  against Alice
> > 8. As Bob, try to access the share now
> > 
> > Expected: you still can access the share
> > Actual: The folder is visible, cannot be entered (no content shown, error instead).
> 
> And which part is misbehaving here, is it a listener on the user delete which is suppose to update the share somehow?

`getUserGroups()` returns an empty array.

> If I re-read your first message, the problem is that once the user is deleted we do not know in which groups they were before deletion? 

Correct

> If so, we can fetch this information from `OCA\User_LDAP\Db\GroupMembershipMapper` now I think.

Affected instance is on 23 :roll_eyes:

apps/user_ldap/lib/Group_LDAP.php
```diff
@@ -664,10 +666,20 @@ public function getUserPrimaryGroup(string $dn) {
...
 		}
+		$ncUid = $uid;
+
+		$user = $this->access->userManager->get($uid);
+		if ($user instanceof OfflineUser) {
+			/** @var IConfig $config */
+			$config = \OCP\Server::get(IConfig::class);
+			$groupStr = $config->getUserValue($uid, 'user_ldap', 'group-memberships-' . $this->access->connection->getConfigPrefix(), '[]');
+			return \json_decode($groupStr) ?? [];
+		}
```
come-nc, line 681:
```suggestion
		if ($user instanceof OfflineUser) {
			/* For remnants, we load known group memberships from configuration because LDAP server does not contain them anymore */
			/** @var IConfig $config */
			$config = \OCP\Server::get(IConfig::class);
			$groupStr = $config->getUserValue($uid, 'user_ldap', 'group-memberships-' . $this->access->connection->getConfigPrefix(), '[]');
			return \json_decode($groupStr) ?? [];
		}
```

apps/user_ldap/lib/Group_LDAP.php
```diff
@@ -786,6 +798,10 @@ public function getUserGroups($uid) {
 
 		$groups = array_unique($groups, SORT_LOCALE_STRING);
 		$this->access->connection->writeToCache($cacheKey, $groups);
+		/** @var IConfig $config */
+		$config = \OCP\Server::get(IConfig::class);
+		$groupStr = \json_encode($groups);
+		$config->setUserValue($ncUid, 'user_ldap', 'group-memberships-' . $this->access->connection->getConfigPrefix(), $groupStr);
```
come-nc, line 804:
```suggestion
		$config->setUserValue($ncUid, 'user_ldap', 'cache-group-memberships-' . $this->access->connection->getConfigPrefix(), $groupStr);
```
I’d prefer something prefixed with cache or backup to avoid misusages of this configuration key when people will stumble upon it.
Also, for modern releases if we could avoid this config key and use the group membership mapper I’d feel better, but we need to check first if the group membership mapping is kept in these situations.

apps/user_ldap/lib/Group_LDAP.php
```diff
@@ -664,10 +666,20 @@ public function getUserPrimaryGroup(string $dn) {
...
+		if ($user instanceof OfflineUser) {
+			/** @var IConfig $config */
+			$config = \OCP\Server::get(IConfig::class);
+			$groupStr = $config->getUserValue($uid, 'user_ldap', 'group-memberships-' . $this->access->connection->getConfigPrefix(), '[]');
+			return \json_decode($groupStr) ?? [];
+		}
+
 		$cacheKey = 'getUserGroups' . $uid;
 		$userGroups = $this->access->connection->getFromCache($cacheKey);
 		if (!is_null($userGroups)) {
```
come-nc, line 685: Cache could be checked first, no?
This avoids searching for the user object when we already have the info in cache.

blizzz:
Addressed the remarks. Evaluating utilizing the Mapper would be good for a followup? So we can have this working and backported.

come-nc (approved):

blizzz:
Rebased and added a test.

Alas, I think I found a race condition while smoke testing :-/

blizzz:
> Alas, I think I found a race condition while smoke testing :-/

Should be a bit obvious. If the user is not detected yet as "offline", it still goes through as normal user and there is no userExists check. 

> Rebased and added a test.
> 
> Alas, I think I found a race condition while smoke testing :-/

Have a potential fix for this, relies on https://github.com/nextcloud/server/pull/40839, but would contact LDAP each time once it passed the first cache check, when the user object is not `OfflineUser`. Not so cool. Maybe I get another idea.

blizzz:
Completed now and the race condition resolved

nfebe (approved):

come-nc (approved):

PR #40463, opened 2023-09-17
Revert "Non sticky delete button"

skjnldsv:
This reverts commit 2b2622ddd151d3092bc9ec8b099cb3c2c9a04852.

image

For some reason, this seems to break files scrolling on master. I bisected like this:
One good commit: e0a7003f08b07d2c9c614120e46201a683f74a7c
One bad commit: 79d1870aa7754c3ed15eec0e17cf7c50d45adc2c

```
2b2622ddd151d3092bc9ec8b099cb3c2c9a04852 is the first bad commit
commit 2b2622ddd151d3092bc9ec8b099cb3c2c9a04852
Author: fenn-cs <fenn25.fn@gmail.com>
Date:   Wed Sep 13 15:15:26 2023 +0100

    Non sticky delete button
    
    Removes button from footer and removes button background.
    
    Signed-off-by: fenn-cs <fenn25.fn@gmail.com>

 ```


susnux:
I reverted and recompiled to trace this down, #40103 was the last merged PR that worked, #40284 broke it.

The reason reverting that commit works for you seems to be that #40284 changed source but did not compile (at least the git history says so). No idea how that could pass the CI.

Meaning
* `git checkout 4f0a151eb4661070c97a726b68511f0c4456c66b && npm ci && npm run dev` works ✅
* `git checkout 165c54d62079e3ec3459dddc1b9f7e23f2140720 && npm ci && npm run dev` does not ❌

Furthermore I tried to track down which commit caused the issue on that PR and it seems to be 146e9eeb9f3c7d4462ea575788be9327a51413ea

Issue #8283, opened 2018-02-09
Missing "Enable recovery key" field in the admin settings


CamZie:
### Steps to reproduce
1. Install Nextcloud 13
2. Enable server-side encryption in the admin settings
3. Enable the app "Default encryption module" 

### Expected behaviour
The fields to enable recovery key should appear in the admin settings under "Encryption".

### Actual behaviour
The fields to enable recovery key does not appear, hence we can not create the recovery key.

### Server configuration

**Operating system**: Debian 9.3 (stretch)

**Web server:** NGINX 1.10

**Database:** Maria-DB 10.1.26

**PHP version:** PHP 7

**Nextcloud version:** 13

**Updated from an older Nextcloud/ownCloud or fresh install:** Fresh install

**Signing status:**
<details>
<summary>Signing status</summary>

```
No errors have been found.
```
</details>

**List of activated apps:**
<details>
<summary>App list</summary>

```
Enabled:
  - activity: 2.6.1
  - comments: 1.3.0
  - dav: 1.4.6
  - encryption: 2.0.0
  - federatedfilesharing: 1.3.1
  - federation: 1.3.0
  - files: 1.8.0
  - files_pdfviewer: 1.2.0
  - files_sharing: 1.5.0
  - files_texteditor: 2.5.1
  - files_trashbin: 1.3.0
  - files_versions: 1.6.0
  - files_videoplayer: 1.2.0
  - firstrunwizard: 2.2.1
  - gallery: 18.0.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.1.0
  - nextcloud_announcements: 1.2.0
  - notifications: 2.1.2
  - oauth2: 1.1.0
  - password_policy: 1.3.0
  - provisioning_api: 1.3.0
  - serverinfo: 1.3.0
  - sharebymail: 1.3.0
  - survey_client: 1.1.0
  - systemtags: 1.3.0
  - theming: 1.4.1
  - twofactor_backupcodes: 1.2.3
  - updatenotification: 1.3.0
  - workflowengine: 1.3.0
Disabled:
  - admin_audit
  - files_external
  - user_external
  - user_ldap
```
</details>

**Nextcloud configuration:**
<details>
<summary>Config report</summary>

```
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
[...]

PR #40468, opened 2023-09-17
[Proposal] Inform user about master key generation on encryption enabling

func0der:
* Resolves: kinda discussions in https://github.com/nextcloud/server/issues/8283 but also a lot of weirdness and questionmarks around this topic.

## Summary

The user/admin should be informed about the fact that the encryption is generating a master key for the whole instance.
This default behavior was introduced in Nextcloud 13 and was not properly communicated (as it seems). There was no trace to be found in the docs, nor was there anything in the interface warning the non-RTFM users about this.

This change to a less secure, but more user friendly setup, should be clearly communicated to the user since Nextcloud is used in more and more setups which require them to be compliant with very high privacy standards.
If the user is not pro-actively informed about this setup, they might violate those standards by accident.

## TODO

- [ ] Add language file entries for all languages (done when this change finds acceptance)

## Checklist

Before | After
-- | --
before | after

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required -> https://github.com/nextcloud/documentation/pull/11122


func0der:
Anything that clears this matter up is a welcome change :D

apps/settings/src/components/Encryption.vue
```diff
@@ -18,6 +18,7 @@
 			<p>{{ t('settings', 'Please read carefully before activating server-side encryption:') }}</p>
 			<ul>
 				<li>{{ t('settings', 'Once encryption is enabled, all files uploaded to the server from that point forward will be encrypted at rest on the server. It will only be possible to disable encryption at a later date if the active encryption module supports that function, and all pre-conditions (e.g. setting a recover key) are met.') }}</li>
+				<li>{{ t('settings', 'By default a master key for the whole instance will be generated. Please check if that level of access is compliant with your needs.') }}</li>
```
susnux, line 21:
```suggestion
				<li>{{ t('settings', 'By default a master key for the whole instance will be generated. Please check if that level of access is compliant with your needs. Otherwise refer to the documentation on how to enable per-user keys.') }}</li>
```

szaimen:
Hm... wouldnt this conflict with https://github.com/notifications?query=is%3Adone?

susnux:
@szaimen yes but that information is still there and we should add that part

apps/settings/src/components/Encryption.vue
szaimen: Sounds good to add this

susnux:
@func0der took the liberty to rebase the PR

susnux (approved):

szaimen (approved):
LGTM but didnt test

AndyScherzinger (approved):

susnux:
@AndyScherzinger compile does not work on forks. Manually compiled and pushed.
Cypress is disabled on forks too but there should be no issue as this is only adding text (and we do not even have tests for it).

PR #40471, opened 2023-09-18
Only add x-requested-with header in requests to Nextcloud

julien-nc:
refs #39924

Adding this header systematically is breaking any usage of the MapLibre map library for example because the lib reaches endpoints with a strict CORS policy which does not accept extra headers.


artonge (approved):

juliusknorr (approved):

juliushaertl:
Conflicts on dist bundles ;)

julien-nc:
Cypress failures are unrelated (theming and user settings).

juliushaertl:
Small follow up fix is in https://github.com/nextcloud/server/pull/40944

PR #40472, opened 2023-09-18
fix(twofactor): avoid DB error on Twofactor (en/dis)abled event

Altahrim:
* Required for: nextcloud/twofactor_admin#35

## Summary
With PostgreSQL, a transaction is aborted each time an error happen with the following message:
```
ERROR:  current transaction is aborted, commands ignored until end of transaction block
```
This PR avoid the error so the transaction can continue as expected.

## Checklist

### DIFF
--- a/lib/private/Authentication/TwoFactorAuth/Db/ProviderUserAssignmentDao.php
+++ b/lib/private/Authentication/TwoFactorAuth/Db/ProviderUserAssignmentDao.php
@@ -25,8 +25,6 @@ declare(strict_types=1);
  */
 namespace OC\Authentication\TwoFactorAuth\Db;
 
-use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
-use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
 use function array_map;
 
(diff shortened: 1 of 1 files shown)


lib/private/Authentication/TwoFactorAuth/Db/ProviderUserAssignmentDao.php
```diff
@@ -70,25 +68,24 @@ public function getState(string $uid): array {
...
-			// There is already an entry -> update it
-			$updateQuery = $qb->update(self::TABLE_NAME)
-				->set('enabled', $qb->createNamedParameter($enabled))
-				->where($qb->expr()->eq('provider_id', $qb->createNamedParameter($providerId)))
-				->andWhere($qb->expr()->eq('uid', $qb->createNamedParameter($uid)));
-			$updateQuery->execute();
+		$conn = $this->conn;
+
+		// Insert a new entry
+		if ($conn->insertIgnoreConflict(self::TABLE_NAME, [
```
nickvergessen, line 74: It does exactly the same?
https://github.com/nextcloud/server/blob/5c78adb20dc7d9b1db3cf07e3dc42d982afb0ce3/lib/private/DB/Adapter.php#L139-L153

Altahrim: As I understand it, the method is overridden in PGAdapter for PG >= 9.5:
https://github.com/nextcloud/server/blob/5c78adb20dc7d9b1db3cf07e3dc42d982afb0ce3/lib/private/DB/AdapterPgSql.php#L46-L60

nickvergessen: alright then!

nickvergessen (approved):

ChristophWurst (approved):
I'm very new to this approach but it seems reasonable

Altahrim:
CI failure seems unrelated

nickvergessen:
We are in freeze anyway, so you have to wait until friday

PR #40474, opened 2023-09-18
fix(dav): report status

skjnldsv:
Fixed favorites and (incoming) tag view

### Context
If a status is defined on the response, it will be used and forced-rendered outside of `propstat`. Which is invalid according to the RFC and broke in the last webdav update (which is now compliant to the rfc too)

https://github.com/nextcloud/3rdparty/blob/1269091e96a2a550912db37c0d2b29ed5c0aa60f/sabre/dav/lib/DAV/Xml/Element/Response.php#L115-L117

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesReportPlugin.php
@@ -406,7 +406,6 @@ class FilesReportPlugin extends ServerPlugin {
 			$responses[] = new Response(
 				rtrim($this->server->getBaseUri(), '/') . $filesUri . $node->getPath(),
 				$result,
-				200
 			);
 		}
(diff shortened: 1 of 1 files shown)


icewind1991 (approved):

artonge (approved):

susnux:
For reference: This one is a workaround for https://github.com/sabre-io/dav/pull/1488

icewind1991:
Tests meed updating?

```
There was 1 failure:
--
Failed asserting that null matches expected 200. 
--
/drone/src/apps/dav/tests/unit/Connector/Sabre/FilesReportPluginTest.php:445
```

skjnldsv:
@icewind1991 https://github.com/nextcloud/server/pull/40485

PR #40482, opened 2023-09-18
fix(isLegitimatedForUserId): Setup mountpoints to check file access

### LINKED ISSUE #71: No notification from groupfolders
I have a 'flow' set to notify when a file with a specific name is edited.
It works well for a file in a shared folder, but no notification is pushed for a file in a group folder.

mejo-:
This fixes workflows on groupfolders, as it will consider access to files in groupfolders.

It also fixes false positives where access to files was limited by other means not taken into account before, e.g. access control.

For postDelete events, check for permissions of the parent folder instead, as the file itself no longer exists.

Fixes: nextcloud/flow_notifications#71

Idea taken from https://github.com/nextcloud/server/pull/38946

## Checklist

### DIFF
--- a/apps/workflowengine/lib/Entity/File.php
+++ b/apps/workflowengine/lib/Entity/File.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
  *
  * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
+ * @author Jonas Meurer <jonas@freesources.org>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -31,6 +32,7 @@ use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\GenericEvent;
 use OCP\Files\InvalidPathException;
 use OCP\Files\IRootFolder;
+use OCP\Files\Mount\IMountManager;
 use OCP\Files\Node;
 use OCP\Files\NotFoundException;
 use OCP\IL10N;
@@ -74,6 +76,8 @@ class File implements IEntity, IDisplayText, IUrl, IIcon, IContextPortation {
 	private $userManager;
 	/** @var UserMountCache */
 	private $userMountCache;
+	/** @var IMountManager */
+	private $mountManager;
 
 	public function __construct(
 		IL10N $l10n,
@@ -82,7 +86,8 @@ class File implements IEntity, IDisplayText, IUrl, IIcon, IContextPortation {
 		IUserSession $userSession,
 		ISystemTagManager $tagManager,
 		IUserManager $userManager,
-		UserMountCache $userMountCache
+		UserMountCache $userMountCache,
+		IMountManager $mountManager
 	) {
 		$this->l10n = $l10n;
 		$this->urlGenerator = $urlGenerator;
@@ -91,6 +96,7 @@ class File implements IEntity, IDisplayText, IUrl, IIcon, IContextPortation {
 		$this->tagManager = $tagManager;
 		$this->userManager = $userManager;
 		$this->userMountCache = $userMountCache;
+		$this->mountManager = $mountManager;
 	}
 
 	public function getName(): string {
@@ -143,10 +149,10 @@ class File implements IEntity, IDisplayText, IUrl, IIcon, IContextPortation {
 				$fileId = $node->getId();
 			}
 
-			$mounts = $this->userMountCache->getMountsForFileId($fileId, $uid);
-			foreach ($mounts as $mount) {
-				$userFolder = $this->root->getUserFolder($uid);
-				if (!empty($userFolder->getById($fileId))) {
+			$mountInfos = $this->userMountCache->getMountsForFileId($fileId, $uid);
+			foreach ($mountInfos as $mountInfo) {
+				$mount = $this->mountManager->getMountFromMountInfo($mountInfo);
+				if ($mount && $mount->getStorage() && !empty($mount->getStorage()->getCache()->get($fileId))) {
 					return true;
 				}
 			}
--- a/apps/workflowengine/lib/Entity/File.php
+++ b/apps/workflowengine/lib/Entity/File.php
@@ -26,6 +26,7 @@ declare(strict_types=1);
  */
 namespace OCA\WorkflowEngine\Entity;
 
+use OC\Files\Config\UserMountCache;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\GenericEvent;
 use OCP\Files\InvalidPathException;
@@ -37,7 +38,6 @@ use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
 use OCP\IUserSession;
-use OCP\Share\IManager as ShareManager;
 use OCP\SystemTag\ISystemTag;
 use OCP\SystemTag\ISystemTagManager;
 use OCP\SystemTag\MapperEvent;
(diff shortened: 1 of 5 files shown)


apps/workflowengine/lib/Entity/File.php
```diff
@@ -136,12 +141,19 @@ public function prepareRuleMatcher(IRuleMatcher $ruleMatcher, string $eventName,
 
 	public function isLegitimatedForUserId(string $uid): bool {
 		try {
-			$node = $this->getNode();
-			if ($node->getOwner()->getUID() === $uid) {
-				return true;
+			$fileId = $this->getNode()->getId();
+			$mounts = $this->userMountCache->getMountsForFileId($fileId);
```
icewind1991, line 145: you can add the user filter here since we're only checking for one user

apps/workflowengine/lib/Entity/File.php
```diff
@@ -136,12 +136,15 @@ public function prepareRuleMatcher(IRuleMatcher $ruleMatcher, string $eventName,
 
 	public function isLegitimatedForUserId(string $uid): bool {
 		try {
-			$node = $this->getNode();
-			if ($node->getOwner()->getUID() === $uid) {
-				return true;
```
icewind1991, line 141: Probably best to leave this in as a fast path

icewind1991:
> @icewind1991 do you think this can be a performance killer?

It isn't worse than the previous code (`getAccessList` is probably just as bad), but if this gets called in a loop (like for every file in a folder that is being accessed) than performance will be dead yes.
I do not know enough about workflow to know how this code is called

apps/workflowengine/lib/Entity/File.php
mejo-: Good point, fixed now :+1:

apps/workflowengine/lib/Entity/File.php
mejo-: Agreed, also fixed now.

mejo-:
Cypress test failures seem unrelated. PR should be ready for review :blush:

mejo-:
By the way, I manually tested the following scenarios (all with a `flow_notification` flow defined by admin for files with a system tag):

* User `jane` creates/changes a file in a tagged folder shared with her by `admin` -> `isLegitimatedForUserId` returns `true`
* User `jane` deletes a file in a tagged folder shared with her by `admin` -> `isLegitimatedForUserId` returns `true`
* User `jane` creates/changes a file in a tagged groupfolder where `admin` has access  -> `isLegitimatedForUserId` returns `true`
* User `jane` deletes a file in a tagged groupfolder where `admin` has access  ->  -> `isLegitimatedForUserId` returns `true`
* User `jane` creates/changes a file in a tagged groupfolder where `admin` doesn't have access  -> `isLegitimatedForUserId` returns `false`
* User `jane` deletes a file in a tagged groupfolder where `admin` doesn't have access  ->  -> `isLegitimatedForUserId` returns `false`

apps/workflowengine/lib/Entity/File.php
```diff
@@ -140,8 +140,19 @@ public function isLegitimatedForUserId(string $uid): bool {
...
-			$acl = $this->shareManager->getAccessList($node, true, true);
-			return isset($acl['users']) && array_key_exists($uid, $acl['users']);
+			$fileId = $node->getId();
+			$mounts = $this->userMountCache->getMountsForFileId($fileId, $uid);
+			foreach ($mounts as $mount) {
+				$userFolder = $this->root->getUserFolder($uid);
+				if ($this->eventName === self::EVENT_NAMESPACE . 'postDelete') {
+					// At postDelete, the file no longer exists. Check for parent folder instead.
+					$fileId = $node->getParentId();
+				}
```
icewind1991, line 150: this can be moved outside the loop

apps/workflowengine/lib/Entity/File.php
```diff
@@ -140,8 +140,19 @@ public function isLegitimatedForUserId(string $uid): bool {
...
+				if (!empty($userFolder->getById($fileId))) {
```
icewind1991, line 151: I'm not a fan of doing `getById` here, it not only duplicates some of the logic we just did (getting the mount from the UserMountCache), it also does extra work to setup the nodes that we then just throw away here.

An alternative approach would be to create a `getMountFromMountInfo` (or similarly named) method in the `MountManager` to get the `MountPoint` instance matching a `ICachedMountInfo`. This would setup the mounts for the user and find the mount that matches.

When we have a `MountPoint` we can do `$mount->getStorage()->getCache()->get($id)` to check if the user actually has access.

apps/workflowengine/lib/Entity/File.php
mejo-: Agreed, changed now.

apps/workflowengine/lib/Entity/File.php
mejo-: See commit 585897014c5e077bdbd99bafd8670a1f076737cc for my current implementation attempt.

icewind1991: That approach looks good.

> Given that the fix needs to be backported until stable26, I guess it's not an option to add getMountFromMountInfo() to the public IMountManager interface, right? Sure we can add it there as well, but at least in the backports we would have to use the private Mount/Manager class, or am I missing something?

Yes, for the backports we can just leave out the interface changes

> What exactly would you compare to match the mount from ICachedMountFileInfo? The rootId?

Mountpoint

mejo-: Thanks for the comments @icewind1991. Unfortunately comparing `$mount->getMountPoint()` and `$info->getMountPoint()` doesn't work in my tests with groupfolders. While `$info->getMountPoint()` is e.g. `/admin/files/testgroupfolder` (for workflows configured by `admin`), `$mount->getMountPoint()` is `/jane/files/testgroupfolder` (for changes done by `jane`). Comparing the `rootId`/`storageRoot` on the other hand would work. What do you think @icewind1991?

lib/private/Files/Mount/Manager.php
```diff
@@ -226,4 +227,21 @@ public function getMountsByMountProvider(string $path, array $mountProviders) {
...
+
+	/**
+	 * Return the mount matching a cached mount info (or mount file info)
+	 *
+	 * @param ICachedMountInfo $info
+	 *
+	 * @return IMountPoint
+	 * @throws NotFoundException
+	 */
+	public function getMountFromMountInfo(ICachedMountInfo $info): IMountPoint {
```
icewind1991, line 239: Probably better to return `null` instead of throwing an exception when the mount is not found imo

apps/workflowengine/lib/Entity/File.php
icewind1991: `rootId` is not unique by itself (the root of an ext storage/gf being shared to a user that already has access will lead to 2 mounts with the same `rootId`) which could cause issues as the duplicate mounts might have different permissions.

Might just need to setup the filesystem for `$uid` here

mejo-: How about comparing `$info->getRootInternalPath()` to `$mount->getSourcePath()`? Or does this have the same problem as `rootId`, because it compares the source instead of the mountpoint?

icewind1991: It still has the same `rootid` yes, `mountpoint` is the only field in the mountcache that is unique.

for the specific use case it doesn't really mattert (since we only care that the user has at least one mountpoint that can access the file), but for a method in `IMountManager` it needs to deal with "duplicate" mounts

mejo-: I implemented it like @icewind1991 suggested now: Setup the filesystem for the mountpoint in `getMountFromMountInfo()`.

icewind1991 (approved):

juliusknorr (approved):
One psalm complaint, but otherwise seems reasonable :+1:

mejo-:
Thanks @juliushaertl. I addressed the psalm complaint now.

mejo-:
Will need manual backports without the public interface changes to stable27 and stable26.

PR #40500, opened 2023-09-19
Consider share type for enforced share expiry dates

nfebe:
We have three kind of share expiry date enforcements.

- `isDefaultExpireDateEnforced`
- `isDefaultInternalExpireDateEnforced`
- `isDefaultRemoteExpireDateEnforced`

Before these commit, `isExpiryDateEnforced` that is used to disable/enable the set expiry date checkbox does not take those into consideration which is problematic as those have different applications.

In addition, this commit now uses `isExpiryDateEnforced` to disable/enable the input for expiry date that shows up before the creation of link shares.


AndyScherzinger (approved):

apps/files_sharing/src/mixins/SharesMixin.js
```diff
@@ -135,6 +135,15 @@ export default {
...
 		},
+		isExpiryDateEnforced() {
+			if (this.isPublicShare) {
+				return this.config.isDefaultExpireDateEnforced
+			}
+			if (this.isRemoteShare) {
+			    return this.config.isDefaultRemoteExpireDateEnforced
+			}
+			return this.config.isDefaultInternalExpireDateEnforced
+		},
```
artonge, line 146: Does `isDefaultExpireDateEnforced` only apply to public shares? If not, then this should be better: 
```suggestion
		isExpiryDateEnforced() {
			if (this.isPublicShare) {
				return this.config.isDefaultExpireDateEnforced
			}
			if (this.isRemoteShare) {
			    return this.config.isDefaultRemoteExpireDateEnforced || this.config.isDefaultExpireDateEnforced
			}
			return this.config.isDefaultInternalExpireDateEnforced || this.config.isDefaultExpireDateEnforced
		},
```

artonge (approved):

PR #40517, opened 2023-09-19
feat(files): add files_sharing indicator

### LINKED ISSUE #38156: Sharing icon not displayed for files shared to the user
Most files on Nextcloud have a sharing icon on the right, next to the file's menu.

However, when another user shares a file with me, the sharing icon is replaced by the avatar of said user. This change can be quite confusing for some users, as they won't know where to click if they want to open the sharing panel on the right.

Artboard Copy 10

I conducted a series of user tests, in which practically all users, even the advanced ones, hesitated when asked to share a folder that had been shared to them. Some of them also bypassed the issue by going into the details, inside the file's menu. 

The lack of sharing icon is confusing for users. I suggest to make it permanent to make things more cohesive and easier to understand. The icon also doesn't take a lot of space.

Artboard Copy 6

@jancborchardt @nimishavijay what do you think ? :)

skjnldsv:
image
Ref: https://github.com/nextcloud/server/issues/39914
Closes https://github.com/nextcloud/server/issues/38156

### Additional infos
It might be good to extend the action API to allow title on inline Actions :grin:


apps/files/src/components/FileEntry.vue
```diff
@@ -109,16 +109,19 @@
...
+				:source="source"
+				class="files-list__row-action--inline" />
 
 			<!-- Menu actions -->
 			<NcActions v-if="visible"
 				ref="actionsMenu"
 				:boundaries-element="getBoundariesElement()"
 				:container="getBoundariesElement()"
 				:disabled="source._loading"
+				:force-name="true"
```
skjnldsv, line 124: @lhsazevedo this was needed because otherwise we cannot get a mix between inline actions with text and actions without, see screenshot :)

skjnldsv:
For the icon reverse, I think @susnux worked on something related? :thinking: 
This can always be improved later anyway :)

artonge (approved):

szaimen (approved):
LGTM :)

apps/files/src/components/FileEntry.vue
lhsazevedo: Oh I see. Extending the API to allow title display control sounds good then!

lhsazevedo: Oops, looks like I missed #40551 Nice!

PR #40522, opened 2023-09-20
delete filecache entries when the object doesn't exist

icewind1991:
If the cache and object store get out of sync for any reason we can have items in the cache for an object store with no matching objects. This changes it so the cache item is deleted automatically when such a file is being read.

It logs the full cache item to provide some breadcrumbs if these files ever need to be recovered.
I would like to have a more complete "trashbin" for these items to help with debugging/resolving the cache desync issues but I can't think of anything atm.


sorbaugh:
Aligned with @icewind1991 . We should revive this but it's being debated if there can be an improvement made on how to track which files are deleted to give admins the chance to review these.

PR #40530, opened 2023-09-20
[stable27] Consider share type for enforced share expiry dates 

nfebe:
Manual backport #40500


AndyScherzinger (approved):

artonge (approved):

blizzz:
@Fenn-CS I am a irritated. Do we need another compile, or is it actually good? It looks to me like that, with CI being green.

Fenn-CS:
@blizzz I don't think we need another compile :)

blizzz:
failing test is unrelated → merging

Issue #34898, opened 2022-10-31
CORS issue in login flow


salonikumawat28:
*Context*
I have a website(not the NextCloud app) from which I want to login into NextCloud so that I can get credentials which I can use to download and upload files.

First step I did was using the hardcoded username:password and passed it as `Authorization` header to download and upload URLs which failed on CORS. I was able to resolve the CORS issues once I installed `WebAppPassword` and whitelisted my website.

As a second step, I tried the [login v2 flow](https://docs.nextcloud.com/server/latest/developer_manual/client_apis/LoginFlow/index.html#login-flow-v2) and it failed on CORS. This was surprising as my website origin was already whitelisted in`WebAppPassword` and was working for download and upload URLs.

As a temporary hack, I added `@CORS` annotation in the login v2 controllers and it worked on all URLs except `login/v2/poll` because it fails on `preflight` request as it's a `POST` call with json content type. To solve this, another temporary hack I did is to add preflight URL for login v2 flow.

*Problem*
As these hacks were direct code changes in the NextCloud server code, I want to replace these hacks with a proper solutions. Things I tried:
1. I created an app within the NextCloud and registered a Middleware to resolve the CORS issues. The problem is that the Middleware works only the URLs pointed to the app I wrote and it doesn't work on the login URLs. Even if it works somehow, it will not resolve the preflight issue.

Can you please help me on how to resolve the CORS and preflight issues?

salonikumawat28:
I looked into similar issues like https://github.com/nextcloud/server/issues/3131 and all the solution attempts and temporary hacks mentioned in the issue were based on the suggested mentioned in the issue, NextCloud [login v2 flow](https://docs.nextcloud.com/server/stable/developer_manual/client_apis/LoginFlow/index.html#login-flow-v2) doc and NextCloud [OAuth flow](https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/oauth2.html) doc but I was not able to find a solution which doesn't involve changes in NextCloud server.

salonikumawat28:
Gentle Reminde : Can someone help me into this ?

salonikumawat28:
Can please someone help me on this issue?

julien-nc:
@juliushaertl Not really, no. The only topic where it is currently possible to canonically deal with CORS headers is WebDAV endpoints (by adding a Sabre plugin like the WebAppPassword app does).

[...]

PR #40537, opened 2023-09-20
Allow to configure "allowed domains" for CORS on DAV

### LINKED ISSUE #3131: Support for cross-domain WebDAV access (CORS)
It would be great if web applications hosted on a different domain could access Nextcloud files via WebDAV. This would enable web app developers to offer users to store personal application data in their Nextcloud without the need to provide a dedicated Nextcloud app.

Currently this is not possible as the necessary CORS headers are not set.

@LukasReschke

### LINKED ISSUE #30964: CORS setting at user level
Hello

Some user need to allow CORS (Cross-origin resource sharing) to be able sharing data with external in-browser application. Many software make CORS settings available to administrator level.

**But it will be fairest to let the user choose, isn't it ?**

Regards & cheers :-)

### LINKED ISSUE #37716: Cors origin allowed list check

susnux:
* Resolves: #3131 
* Resolves: #30964
* <del>check if #34898</del> (not yet)
* Resolves: #37716 (by checking a list of allowed domains, so the sharing routes can make use of it)

## Summary
This adds CORS support to the DAV routes, the default is non breaking -> so no other domain is allowed.
But the admin can configure a list of allowed domains which will be allowed to access the DAV routes using CORS.

Moreover the admin can enable user defined lists, meaning that users can add their own allowed domains for WebDAV related requests, see #30964.

### Why do we need CORS on DAV?
See #3131

Basically to access it within the browser from a different page, like e.g. the Dropbox file picker with allows you to pick files from your cloud onto another page (e.g. upload something from cloud to form etc).

### Screenshot

Screenshot_20230920_150622

## Todo
* Follow-up: User config UI
* Follow-up: allowed header UI
* If this is decided: Write documentation

## Checklist


apps/dav/lib/Connector/Sabre/CorsPlugin.php
```diff
@@ -0,0 +1,176 @@
...
+use Sabre\HTTP\ResponseInterface;
+
+/**
+ * Class CorsPlugin is a plugin which adds CORS headers to the responses
+ */
+class CorsPlugin extends ServerPlugin {
+	/**
+	 * Reference to main server object
+	 */
+	private \Sabre\DAV\Server $server;
```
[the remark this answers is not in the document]
susnux: I think this is not possible and can only be done in `initialize`

skjnldsv:
I'm too irrelevant with this PR, removing myself as reviewer :)

apps/dav/lib/Server.php
```diff
@@ -130,6 +131,8 @@
 		$this->server->addPlugin(new ProfilerPlugin($this->request));
 		$this->server->addPlugin(new BlockLegacyClientPlugin(\OC::$server->getConfig()));
 		$this->server->addPlugin(new AnonymousOptionsPlugin());
+		$this->server->addPlugin(new CorsPlugin(\OC::$server->getUserSession(), \OC::$server->getConfig()));
```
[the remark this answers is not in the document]
Altahrim: I think you can change to:
```suggestion
		$this->server->addPlugin(new CorsPlugin(\OC::$server->get(IUserSession), \OC::$server->get(IConfig::class)));
```
With:
```php
use \OCP\IConfig;
use \OCP\IUserSession
```

lib/public/Util.php
```diff
@@ -624,4 +624,60 @@ public static function isFunctionEnabled(string $functionName): bool {
...
+			throw new \InvalidArgumentException('Invalid url "' . $url . '"');
+		}
+		if (!isset($parts['scheme']) || !isset($parts['host'])) {
+			throw new \InvalidArgumentException('Invalid url "' . $url . '"');
+		}
+		$protocol = \strtolower($parts['scheme']);
+		$host = \strtolower($parts['host']);
+
+		$port = $parts['port'] ?? null;
+		if ($port === null || $port === 0) {
```
Altahrim, line 653:
```suggestion
		$port = (int) $parts['port'] ?? 0;
		if ($port === 0) {
```

settings/Controller/CorsController.php
```diff
@@ -0,0 +1,156 @@
...
+								IURLGenerator $urlGenerator,
+								IConfig $config) {
+		parent::__construct($AppName, $request);
+
+		$this->AppName = $AppName;
+		$this->config = $config;
+		$this->userId = $userSession->getUser()->getUID();
+		$this->logger = $logger;
+		$this->urlGenerator = $urlGenerator;
+	}
```
Altahrim, line 70: Can use properties promotion:
```suggestion
	public function __construct(
		$AppName, IRequest $request,
		IUserSession $userSession,
		private LoggerInterface $logger,
		private IURLGenerator $urlGenerator,
		private IConfig $config
	) {
		parent::__construct($AppName, $request);
		$this->userId = $userSession->getUser()->getUID();
	}
```
Also replace deprecated ILogger

settings/Controller/CorsController.php
```diff
@@ -0,0 +1,156 @@
...
+	 * @return JSONResponse All the White-listed domains
+	 */
+	public function getDomains() {
+		$userId = $this->userId;
+
+		if (empty($this->config->getUserValue($userId, 'core', 'domains'))) {
+			$domains = [];
+		} else {
+			$domains = json_decode($this->config->getUserValue($userId, 'core', 'domains'));
+		}
```
Altahrim, line 97:
```suggestion
		$domains =  empty($this->config->getUserValue($userId, 'core', 'domains')
			? []
			: json_decode($this->config->getUserValue($userId, 'core', 'domains'), flags:  JSON_THROW_ON_ERROR);
```

settings/Controller/CorsController.php
```diff
@@ -0,0 +1,156 @@
...
+		return new JSONResponse($domains);
+	}
+
+	/**
+	 * WhiteLists a domain for CORS
+	 *
+	 * @param string $domain The domain to whitelist
+	 * @return RedirectResponse Redirection to the settings page.
+	 */
+	public function addDomain($domain) {
```
Altahrim, line 108: Can we enforce it here?
```suggestion
	public function addDomain(string $domain) {
```

Altahrim (commented):
First quick read ^^

aleixq:
I am coming from https://github.com/nextcloud/server/issues/37716 for which I was proposing https://github.com/nextcloud/server/pull/37896 . 
If the reason to allow dav cors is to make it easier to offer a filepicker, I think that filesharing api and filepreview api may be covered to be allowed in a similar way too. If not it will be a raw basic one IMHO.
By now the way to offer a functional filepicker is to use the webapppassword app as it also includes the filesharing api extended and preview api extended,  for which I made the accepted PR some time ago when I was trying to offer Nextcloud as a DAM for a CMS (drupal).

PR #40583, opened 2023-09-22
feat(files): properly format buttons, align mtime to the left and apply opacity based on file last modification

skjnldsv:
- Mtime is alligned left
- Mtime needs colour fade from now to 31 days old
- Buttons should not be messed with the padding as `padding-block` is used and not simple padding

image


apps/files/src/components/FileEntry.vue
```diff
@@ -389,6 +390,24 @@ export default Vue.extend({
 			}
 			return this.t('files_trashbin', 'A long time ago')
 		},
+		mtimeOpacity() {
+			// Whatever theme is active, the contrast will pass WCAG AA
```
artonge, line 394: So I assume that this accessible.
@JuliaKirschenheuter wasn't this remove from the old files view?

skjnldsv: Jan talked to me about it recently and said it was missing :)
but yes, it does pass WCAG AA (not AAA, which is ok for secondary text)
image

artonge (approved):
Let's maybe wait for @JuliaKirschenheuter input

szaimen (approved):

apps/files/src/components/FileEntry.vue
skjnldsv: cc @jancborchardt

PR #40585, opened 2023-09-22
fix(files): disallow illegal characters

### LINKED ISSUE #38614: Disallow illegal characters / ? < > \ : * | " and so on in file and folder names
Reopening [issue that was closed by a bot](https://github.com/nextcloud/server/issues/27891).

Nextcloud allows characters in file and folder names that are illegal on some operating systems (windows). 

Users in one of my installs use double quotes and colons. Nextcloud cannot open files or folder with symbols. Windows clients refuse to sync if there is a colon in the folder name.

My suggestion is to optionally not allow these special characters when creating files and folders in the web interface (or implement a setting where the administrator will specify an array of forbidden characters). Files and folders with these symbols in the sync folder must be renamed and the specified symbol removed before it can be distributed to other clients on different OSes.

skjnldsv:
fix #38614


susnux:
I disagree here, just because one system (Window) can not handle those symbols we should not hard forbid it here.
E.g. I often use some of those.

See also the git blame of `OCP\Constants::FILENAME_INVALID_CHARS`.

skjnldsv:
> See also the git blame of `OCP\Constants::FILENAME_INVALID_CHARS`.
> 
> Especially we should not have multiple different character lists that we forbid, I would simply use `OCP\Constants::FILENAME_INVALID_CHARS` and maybe make if configurable to forbid more (system config value?)

Nice, I search for this, and I could only find a variable in files_external, which would ot have been proper to use. 

> I disagree here, just because one system (Window) can not handle those symbols we should not hard forbid it here.

I don't think that's a big of an issue to be honest, but  I agree with you.
What shall we do then? Close the other issue?

nickvergessen:
> Do we actually support windows officially?

Not as server plattform.

But it's of course still bad UX when people create such files on ubuntu/mac and then share with windows users.

The list you have matches the old value of the constant:
https://github.com/nextcloud/server/commit/6bd0d51b2a9742108f64b406fba55d5af6264fb2

So I would say the only disallowed chars are `/` and `\` so you can use the const.

Some external storages have more restrict limits (SMB?), but I'd say it's fine to not block that in the UI in the beginning and just show an error afterwards.

skjnldsv:
> So I would say the only disallowed chars are `/` and `\` so you can use the const.

The const is a weird format though. It should be an array of string, not a string...
Especially the var name is confusing, being plural :see_no_evil:

nickvergessen:
> It should be an array of string, not a string...

We have been growing a long way. It simply was not allowed in PHP until 5.6
grafik

skjnldsv:
Done! Please review :)

apps/files/lib/Controller/ViewController.php
```diff
@@ -246,6 +247,10 @@ public function index($dir = '', $view = '', $fileid = null, $fileNotFound = fal
 		$filesSortingConfig = json_decode($this->config->getUserValue($userId, 'files', 'files_sorting_configs', '{}'), true);
 		$this->initialState->provideInitialState('filesSortingConfig', $filesSortingConfig);
 
+		// Forbidden file characters
+		$forbiddenCharacters = $this->config->getSystemValue('forbidden_chars', []);
+		$this->initialState->provideInitialState('forbiddenCharacters', Constants::FILENAME_INVALID_CHARS . join('', $forbiddenCharacters));
```
nickvergessen, line 252:
```suggestion
		$this->initialState->provideInitialState('forbiddenCharacters', Constants::FILENAME_INVALID_CHARS . implode('', $forbiddenCharacters));
```

join is just an alias

nickvergessen (approved):

config/config.sample.php
```diff
@@ -1933,6 +1933,16 @@
  */
 'blacklisted_files' => ['.htaccess'],
 
+/**
+ * Blacklist characters from being used in filenames. This is useful if you
+ * have a filesystem or OS which does not support certain characters like windows.
+ *
+ * Example for windows systems: ``array('?', '<', '>', ':', '*', '|', '"', chr(0), "\n", "\r")``
```
Altahrim, line 1940: Maybe add a link to help people find forbidden characters for their own filesystem: 
https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits

max-nextcloud (approved):

Altahrim (approved):

susnux:
Very nice solution :+1:

Issue #40245, opened 2023-09-04
↔️ Create a mockup for the settings-block under "Appearance and accessibility"


AndyScherzinger:
mockup/design for #40241

szaimen:
Why was this ticket created? To better track the different sub-tasks?

szaimen:
Hi, here is the MVP mockup for this feature:

![image](https://github.com/nextcloud/server/assets/42591237/c9a4ead1-b3f9-4cc6-945a-16e0ead7c3db)

It should be put into the `Appearance and Accessibility` settings section at the very bottom (below the keyboard shortcuts section).

Let me know what you think! :)

---

For reference here also a mockup how the feature could look like in the very end if there should be enough time:

<details>
<summary>click here to expand</summary>

![image](https://github.com/nextcloud/server/assets/42591237/f8ce776b-f3c3-49fe-8a4f-b66e9ec196a2)

This would then have an additional admin section in theming where we would show this switch.

</details>

sorbaugh:
Hi @skjnldsv ,

question came up yesterday if we need new Vue components to implement this or do we already have them? 🤔

skjnldsv:
There might be a few we can already use, but that doesn't matter tbh, the whole UI needs to be built from scratch

susnux:
We have a similar use case at the forms app where we implemented something similar for question order, so probably we can reuse that :)

come-nc:
> We have a similar use case at the forms app where we implemented something similar for question order, so probably we can reuse that :)

There is the same thing in table view creation UI as well: https://github.com/nextcloud/tables/blob/main/src/modules/main/partials/editViewPartials/SelectedViewColumns.vue

Issue #40241, opened 2023-09-04
↔️ User-configurable app order


sorbaugh:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Users should have the possibility to modify the order in which apps are listed in the top bar and also to define a "default" app that should always be loaded when a user "starts" Nextcloud. 

**Mockup and comments by @szaimen**
Here -> https://github.com/nextcloud/server/issues/40245#issuecomment-1709999731

Reference to old issue asking for this feature: #39162
Reference to comment defining the requirements:  https://github.com/nextcloud/server/issues/4917#issuecomment-1699025806

* This is a user-setting only, no admin setting
* First sorted app will be default app (if set, it will override the default app config setting)
* This has to be a settings block in "Appearance and accessibility", similar to the sorting in "Artificial Intelligence" or the column sorting in Tables so it is both accessible via mouse and keyboard. No sorting by drag & drop directly in the navigation

<img width="929" alt="Bildschirmfoto 2023-09-04 um 14 45 22" src="https://github.com/sorbaugh/example1/assets/62374139/699f528c-f081-4cfb-a443-7608898b44ca">

Task breakdown:
- [ ] #40245
- [x] #40254
- [ ] #40255
- [x] #40256

sorbaugh:
See Mockup:  https://github.com/nextcloud/server/issues/40245#issuecomment-1709999731

ichier:
this is great news! but i've got one question:
[#4917](https://github.com/nextcloud/server/issues/4917) states that:

> Admins should be able to
>     set an default order for apps
>     set, whether users should be allowed to change the order of the apps

is this the case now?

because this is the only reason i didn't upgrade after 24.
imho it is less important for users to be able to customize, than for admins to set up the environment first to make sense according to the use-case.

most important of course is, that there's no dashboard-fuzz but plain files after login. by default. not per user customization.

the ability for users to customize sure is a nice feature, but 99% won't. its the task of admins to provide a clutter-free work-space with no customization necessary.

susnux:
[...]

PR #40600, opened 2023-09-23
[frontend] Add possibility to set default app

### LINKED ISSUE #29474: Ability turnoff dashboard (or choose the default app) on a per user basis
Hey all, first: I love NC. :D To my point: My Org recently installed the NC dashboard app and maybe other ppl find it useful, but I certainly don't. However I cannot turn the dashboard off for me. I also cannot chose to always land in the calendar page after login (not dashboard or files). Could a setting like this maybe be added somewhere or did I miss something? Thanks for all your work

susnux:
* Resolves: https://github.com/nextcloud/server/issues/29474

## Summary
The adds support for setting the default app, that is the app opened after login or when clicking the logo in the navigation.

First I had to extend the `IAppManager` class to support setting the default app(s) (for current user or global as a admin setting).

Secondly I added settings to the theming app to allow configure global default apps (as the admin) and to enable user defined default apps. Also a personal setting for the user defined default app is added.

### Example recording
*(Sorry for German UI)*

https://github.com/nextcloud/server/assets/1855448/d467e178-9595-4491-9b3e-a6b7fb797edb

## Checklist


szaimen:
Hi, I wonder if this might conflict with https://github.com/nextcloud/server/issues/40245?

susnux:
> Hi, I wonder if this might conflict with #40245?

I created this initially during the hack week, just polished it now. What do you think? (I attached a screen recording).

jancborchardt:
So during the discussion for https://github.com/nextcloud/server/issues/40241 we said, as also written in the issue:
> This is a user-setting only, no admin setting

And I would stick to that, as it introduces a lot of complexity into something that is not as important, right? So let’s only go forward with the user setting.
If the existing default_app variable is set in the config.php by the admin that of course will still get honored, but the user-specific setting will override it.

cc @sorbaugh @AndyScherzinger

susnux:
Close in favor of https://github.com/nextcloud/server/pull/40844

PR #40515, opened 2023-09-19
fix: redirect to proper directory if file not found

skjnldsv:
Fix https://github.com/nextcloud/server/issues/32177
Fix https://github.com/nextcloud/server/issues/40104

1. If you ask for a fileid in a dir
2. And the fileid doesn't exists in that directory
3. BUT the file only exists in one occurrence in another directory
4. Then redirect to that file

PR #40604, opened 2023-09-24
fix(files): open folder fileid while navigating without opening the sidebar

skjnldsv:
Related: https://github.com/nextcloud/server/pull/40515

Make sure you can copy/paste URLs and still end up in the same folder even if the path doesn't match :shrug:


skjnldsv:
Wrong  mentioned team, sorry ! :see_no_evil:

susnux (approved):

nfebe (approved):

PR #40606, opened 2023-09-24
[stable26] Do not throw while restoring unloaded files list entries

susnux:
## Summary
When restoring files from trashbin while not all files are loaded into the file list (e.g. overflowing -> scrolling), then the restore all action throws because the `$el` of the unloaded elements in undefined.
In this case simply do not call remove.

For 27 this is fixed with files2vue.

## Checklist


susnux:
/backport 95acf8a0e9b005ee1f27698cbd94427a290ad8e9 to stable25

artonge (approved):

szaimen (approved):
Makes sense but didnt test

PR #40564, opened 2023-09-21
fix(userstatus): Track message timestamp too

ChristophWurst:
* Resolves: https://github.com/nextcloud/server/issues/40558

## Summary

Add new database column to track the message change date.

The fix requires a database schema change and data migration. I would vote against backporting.

## How to test

1) Set up user A
2) Set up user B
3) Log in as A
4) Log in as B in a private window
5) Set your status to DND and the status message to "test" as user A
6) Wait 2 minutes
7) Set your status to online as user A (**do not** use the *Save status message* button. Online status change doesn't need a confirmation)
7) Open the dashboard and recent statuses as user B

master: User A shows "test, seconds ago"
this branch: User A shows "test, 2 minutes ago"

## Screenshots

| Before | After |
|--------|--------|
| Bildschirmfoto vom 2023-09-26 14-31-52 |  Bildschirmfoto vom 2023-09-26 14-31-22 | 

## TODO

- [x] Adjust schema
- [x] Adjust db layer
- [x] Adjust dashboard widget
- [x] Test
- [x] Update branch after https://github.com/nextcloud/server/pull/40614

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #28403, opened 2021-08-12
Add an API to set and rollback the user status

CarlSchwan:
Signed-off-by: Carl Schwan <carl@carlschwan.eu>

PR #40614, opened 2023-09-25
fix(userstatus): Sync migration version number with app version

ChristophWurst:
## Summary

The migration version number was synced with server in https://github.com/nextcloud/server/pull/28403. The app was at 1.3.0. Now we are at 1.8.0 and it just feels wrong to continue the out-of-sync naming in https://github.com/nextcloud/server/pull/40564.

Renaming causes the migration to run again if it hasn't run already. It's adjusted to not cause an issue in that case.

I'm also open for other ideas that bring the versioning state into consistency.

## TODO

- [x] Adjust version number
- [x] Test an upgrade/migration

## Checklist


nickvergessen (approved):
🙈

apps/user_status/lib/Migration/Version1030Date20210809144824.php
```diff
@@ -33,7 +33,7 @@
 /**
  * @package OCA\UserStatus\Migration
  */
-class Version2301Date20210809144824 extends SimpleMigrationStep {
+class Version1030Date20210809144824 extends SimpleMigrationStep {
```
nickvergessen, line 36: If the app was `1.3.0` the expected name would be:
```suggestion
class Version1003Date20210809144824 extends SimpleMigrationStep {
```

nickvergessen (changes requested):

nickvergessen (approved):

st3iny (approved):
LGTM!

kesselb (approved):

ChristophWurst:
mysql job timed out after 1h -> ignore

Issue #39162, opened 2023-07-05
🔎 Advanced search


jancborchardt:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
The current unified search dialog in the top right is quite small and compact. 

**Describe the solution you'd like**
For the powerful things it supports, it would make more sense to have a bit more ways of filtering. E.g. having a left sidebar with filtering which app to focus on, or some detailed selections on owner of file, or date range, or size.

Edit by @sorbaugh (Last check: 04-09-2023): Since the requirements have shifted somewhat over the course of the discussion (and will likely continue to do so), here is a breakdown of the current state in tasks to minimize confusion. I will make an effort to keep this up to date:

Mockup (original comment -> https://github.com/nextcloud/server/issues/39162#issuecomment-1651037351) 
<img width="736" alt="Bildschirmfoto 2023-09-04 um 13 50 04" src="https://github.com/sorbaugh/example1/assets/62374139/8f61efb8-0a57-4a33-b305-1e3c08b8c574">

Frontend Task breakdown:

- [ ] #40238
- [ ] #40239
- [ ] #40240
- [ ] #40237

Backend Task breakdown:
- [ ] #40246
- [ ] #40249
- [ ] #40250

---

**Date**
For things like date range, we have to make sure it works across different apps like Files, Calendar, Contacts, Talk, Mail etc with different types of dates, so we need to define which date it is for the relevant type:
- Files: modification date (probably more relevant than creation date)
- Calendar: event date (probably more relevant than creation date) https://github.com/nextcloud/server/issues/41345
- Deck card: due date (probably more relevant than creation date, if it exists)
- Talk: message date
- Mail: message date https://github.com/nextcloud/mail/issues/8864
- Contacts: ? birthday
- General fallback would be "Modified date"
- etc. …

**People**
The person dropdown could pick from what the "People menu" in the top right also lists, which is both Nextcloud accounts and Contacts. This would then e.g. for Mail also cover the "Recently contacted" address book which is nice.

- Files: owners and sharees
- Calendar: participants https://github.com/nextcloud/server/issues/41345
- Deck card: assignee
- Talk: message poster and/or person replied to
[...]

PR #40618, opened 2023-09-25
Advanced search: backend allows multiples terms to search

### LINKED ISSUE #3608: 🔎 Allow contact and manager search in advanced search
### Is your feature request related to a problem? Please describe.

Extend advanced search in server with Contacts specifics, like searching in cards and managers.

### Describe the solution you'd like

Integrate with the new seach in Hub 7 (details follow)

### Describe alternatives you've considered

_No response_

### Additional context

https://github.com/nextcloud/server/issues/39162

### LINKED ISSUE #41345: 🔎 Allow event date and attendee search in advanced search
### Is your feature request related to a problem? Please describe.

As a user with lots of data I want to be able to find events based on event date and attendees

### Describe the solution you'd like

Integrate with the new seach in Hub 7

### Describe alternatives you've considered

_No response_

### Additional context

https://github.com/nextcloud/server/issues/39162

Altahrim:
Ref #39162 

Partly fixes https://github.com/nextcloud/server/issues/41345
Fixes https://github.com/nextcloud/contacts/issues/3608

## Summary
Allows API to return results of an advanced search

## Checklist


lib/private/Search/SearchQuery.php
```diff
@@ -32,51 +32,39 @@
...
-	/** @var int */
-	private $limit;
-
-	/** @var int|string|null */
-	private $cursor;
-
-	/** @var string */
-	private $route;
+	/**
+	 * @param mixed[] $filters
```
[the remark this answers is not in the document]
Altahrim: I updated the code to reduce the number of `mixed` types.
It seems difficult to get `SearchComparison` automatically because pattern can change. Sometimes we want `%value%`, sometimes `value/%`…

lib/public/Search/ISearchQuery.php
```diff
@@ -46,11 +46,32 @@ interface ISearchQuery {
 	/**
 	 * Get the user-entered search term to find matches for
 	 *
+	 * Alias for getFilter('term', '')
+	 *
 	 * @return string the search term
 	 * @since 20.0.0
+	 * @deprecated
```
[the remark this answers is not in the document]
Altahrim: Done

apps/files/lib/Search/FilesSearchProvider.php
```diff
@@ -141,6 +132,35 @@ public function search(IUser $user, ISearchQuery $query): SearchResult {
...
+			$comparison = match ($filter->name) {
+				'term', 's_term' => new SearchComparison(ISearchComparison::COMPARE_LIKE, 'name', '%' . $filter->value . '%'),
+				't_since' => new SearchComparison(ISearchComparison::COMPARE_GREATER_THAN_EQUAL, 'mtime', $filter->value),
+				't_until' => new SearchComparison(ISearchComparison::COMPARE_LESS_THAN_EQUAL, 'mtime', $filter->value),
+				'i_min-size' => new SearchComparison(ISearchComparison::COMPARE_GREATER_THAN_EQUAL, 'size', $filter->value),
+				'i_max-size' => new SearchComparison(ISearchComparison::COMPARE_LESS_THAN_EQUAL, 'size', $filter->value),
+				's_mime' => new SearchComparison(ISearchComparison::COMPARE_EQUAL, 'mimetype', $filter->value),
+				's_type' => new SearchComparison(ISearchComparison::COMPARE_LIKE, 'mimetype', $filter->value . '/%'),
+				default => false,
+			};
```
come-nc, line 147:
```suggestion
			$comparison = $filter->getSearchComparison();
```

lib/private/Search/Filter.php
```diff
@@ -0,0 +1,48 @@
...
+		try {
+			$this->value = $this->parse($value);
+		} catch (InvalidFilter $e) {
+			throw $e;
+		} catch (Throwable $e) {
+			throw new InvalidFilter($name, $value, $e);
+		}
+	}
+
+	abstract protected function parse(string $value): mixed;
```
come-nc, line 47:
```suggestion
	abstract protected function parse(string $value): mixed;
	
	abstract public function getSearchComparison(): ISearchComparison;
```

lib/private/Search/Filter/TypeDateTime.php
```diff
@@ -0,0 +1,44 @@
...
+		if (filter_var($value, FILTER_VALIDATE_INT)) {
+			$value = '@'.$value;
+		}
+
+		return new DateTimeImmutable($value);
+	}
+
+	public function get(): DateTimeImmutable {
+		return $this->value;
+	}
```
come-nc, line 43:
```suggestion
	}
	
	public function getSearchComparison(): ISearchComparison {
		return new SearchComparison(ISearchComparison::COMPARE_GREATER_THAN_EQUAL, $this->name, $this->value);
	}
```

And either you have two classes for >= and <=, or you add a property to store the operator?

lib/private/Search/Filter/TypeFloat.php
```diff
@@ -0,0 +1,45 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\Search\Filter;
+
+use OC\Search\Filter;
+use OC\Search\InvalidFilter;
+
+class TypeFloat extends Filter {
```
come-nc, line 32: Why are the name prefixed with Type?
Should be FloatFilter or FilterFloat imo.

lib/private/Search/Filter.php
```diff
@@ -0,0 +1,48 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\Search;
+
+use Throwable;
+
+abstract class Filter {
+	public readonly mixed $value;
```
come-nc, line 32: For readonly you need to wait for 8.0 support to be dropped.
See #40636

come-nc: But value should be private, and I think it’s better if it’s only defined in subclasses so that they can type it properly.

lib/private/Search/Filter/TypeFloat.php
Altahrim: Updated

lib/private/Search/Filter.php
Altahrim: Done :)

lib/public/Search/IFilterCollection.php
```diff
@@ -0,0 +1,64 @@
...
+ * @since 28.0.0
+ * @implements IteratorAggregate<string,IFilter>
+ */
+interface IFilterCollection extends IteratorAggregate {
+	/**
+	 * Constructor
+	 *
+	 * @since 28.0.0
+	 */
+	public function __construct(IFilter ...$filters);
```
ChristophWurst, line 42: apps don't know about the implementation of the interface, so they also must not care about the construction. remove the constructor from the interface

lib/public/Search/IProviderV2.php
```diff
@@ -0,0 +1,73 @@
...
+ * Interface for search providers
+ *
+ * These providers will be implemented in apps, so they can participate in the
+ * global search results of Nextcloud. If an app provides more than one type of
+ * resource, e.g. contacts and address books in Nextcloud Contacts, it should
+ * register one provider per group.
+ *
+ * @since 28.0.0
+ */
+interface IProviderV2 extends IProvider {
```
ChristophWurst, line 38: if IProvider doesn't really go away but is *extended*, can we spare the deprecation and keep search providers that do not offer filtering? that's less code than implementing provider v2 and returning empty arrays.

this extending interface could be called `IFilteringProvider`.

lib/public/Search/IProviderV2.php
```diff
@@ -0,0 +1,73 @@
...
+	public function getAlternateIds(): array;
+
+	/**
+	 * Allows application to declare custom filters
+	 *
+	 * @see SearchComposer::FILTERS for an example
+	 * @since 28.0.0
+	 * @return array<string, string> Array with filter name => type
+	 */
+	public function registerCustomFilters(): array;
```
ChristophWurst, line 72:
```suggestion
	public function getCustomFilters(): array;
```

*register* sounds like an action but this is only a *getter* I assume

lib/public/Search/Filter/UserFilter.php
```diff
@@ -0,0 +1,49 @@
...
+namespace OCP\Search\Filter;
+
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\Search\IFilter;
+
+class UserFilter implements IFilter {
+	private IUser $user;
+
+	public function __construct(
```
ChristophWurst, line 36: constructors in public API are tricky because we can't just change the signature in the future. could the instantiation of filters be moved to a factory?

lib/public/Search/Filter/BooleanFilter.php
```diff
@@ -0,0 +1,46 @@
...
+
+namespace OCP\Search\Filter;
+
+use InvalidArgumentException;
+use OCP\Search\IFilter;
+
+class BooleanFilter implements IFilter {
+	private bool $value;
+
+	public function __construct(string $value) {
```
ChristophWurst, line 35:
```suggestion
	public function __construct(bool $value) {
```

:hand_over_mouth:

ChristophWurst (commented):
Early feedback on the public parts

lib/public/Search/IProviderV2.php
Altahrim: I plan to remove it later: move the code into `V2` and drop the original one. 
The `extends` allows to detect both interfaces in one call

lib/public/Search/Filter/BooleanFilter.php
Altahrim: In current implementation, all filters receive `string` or `string[]` values (direct values from query string)

lib/public/Search/Filter/UserFilter.php
Altahrim: The factory is already there: https://github.com/nextcloud/server/blob/feat/39162/advanced_search/lib/private/Search/FilterFactory.php
Actually, all these filters could go in private. The interface is here for that.

lib/public/Search/IProviderV2.php
ChristophWurst: Let's not do that please unless the v1 methods have to change. Removing the old interface completely without a real need means that app developers have to adjust their app for basic unified search integration even if nothing changed.

Altahrim: Yes, you are right.
I was thinking of removing it because it adds a bit a complexity in our UnifiedSearch. I think I can have good compatibility with both interfaces anyway.
I'll keep it and rename it as you suggested :)

core/Controller/UnifiedSearchController.php
```diff
@@ -95,22 +95,23 @@ public function getProviders(string $from = ''): DataResponse {
 	 * 400: Searching is not possible
 	 */
 	public function search(string $providerId,
-						   string $term = '',
```
provokateurin, line 98: Why is the parameter removed? This old API should still continue to work

Altahrim: It's still working, but the compatibility is done later in `SearchComposer`

provokateurin: But this way it is also remove from the OpenAPI specification and that is definitely wrong. I haven't looked at any of the code here, what does the SearchComposer do?

Altahrim: > But this way it is also remove from the OpenAPI specification and that is definitely wrong

You are right about it, but I also need to advertise new parameters. `term` is the default, but now `since`, `until`, `person` and a lot more are also supported. It is possible to do a research without `term`.
I don't know how to do it yet.


> what does the SearchComposer do?

It loads all search providers and give them the filters to work with.

provokateurin: > You are right about it, but I also need to advertise new parameters. term is the default, but now since, until, person and a lot more are also supported. It is possible to do a research without term.

Ah now I understand

>  I don't know how to do it yet.

Just make the parameter nullable then it is optional (or keep the empty string default).

ChristophWurst (commented):
Public API looks good

lib/public/Search/FilterDefinition.php
```diff
@@ -0,0 +1,101 @@
...
+class FilterDefinition {
+	public const TYPE_BOOL = 'bool';
+	public const TYPE_INT = 'int';
+	public const TYPE_FLOAT = 'float';
+	public const TYPE_STRING = 'string';
+	public const TYPE_STRINGS = 'strings';
+	public const TYPE_DATETIME = 'datetime';
+	public const TYPE_PERSON = 'person';
+	public const TYPE_USER = 'user';
+	public const TYPE_GROUP = 'group';
```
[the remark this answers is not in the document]
Altahrim: When modifying it, I noticed we use theses names in [`IShare`](https://github.com/nextcloud/server/blob/master/lib/public/Share/IShare.php#L46-L54). So now, I am in doubt :D

core/openapi.json
```diff
@@ -4013,15 +4013,6 @@
...
                 "parameters": [
-                    {
-                        "name": "term",
-                        "in": "query",
-                        "description": "Term to search",
-                        "schema": {
-                            "type": "string",
-                            "default": ""
-                        }
-                    },
```
provokateurin, line 4024: Has to be reverted

provokateurin (changes requested):

lib/public/Search/FilterDefinition.php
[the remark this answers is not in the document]
ChristophWurst: I'm also fine with *user group* :)

core/Controller/UnifiedSearchController.php
```diff
@@ -82,35 +83,45 @@ public function getProviders(string $from = ''): DataResponse {
...
-			return new DataResponse(null, Http::STATUS_BAD_REQUEST);
-		}
+	public function search(
+		string $providerId,
+		?int $sortOrder = null,
+		?int $limit = null,
+		$cursor = null,
+		string $from = '',
+		// Unused parameter for OpenAPI spec generator
+		string $term = null,
```
provokateurin, line 110: Please keep the term parameter in the beginning to avoid the diff in the OpenAPI spec

core/Controller/UnifiedSearchController.php
```diff
@@ -80,37 +81,46 @@ public function getProviders(string $from = ''): DataResponse {
 	 * @NoAdminRequired
 	 * @NoCSRFRequired
 	 *
-	 * Search
+	 * Launch a search for a specific search provider.
+	 *
+	 * Additionnal filters are available for each provider.
```
provokateurin, line 86:
```suggestion
	 * Additional filters are available for each provider.
```

core/Controller/UnifiedSearchController.php
```diff
@@ -80,37 +81,46 @@ public function getProviders(string $from = ''): DataResponse {
...
+	 * Send a request to /providers endpoint to list providers with their available filters.
 	 *
 	 * @param string $providerId ID of the provider
 	 * @param string $term Term to search
 	 * @param int|null $sortOrder Order of entries
 	 * @param int|null $limit Maximum amount of entries
 	 * @param int|string|null $cursor Offset for searching
 	 * @param string $from The current user URL
+	 * @param string $term Term to search
```
provokateurin, line 95: Duplicate param

provokateurin (changes requested):

provokateurin (approved):
OpenAPI side looks good

lib/public/Search/FilterDefinition.php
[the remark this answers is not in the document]
Altahrim: Updated with `nc_user`/`nc_group`

ChristophWurst (approved):
For lib/public

ChristophWurst:
image

Visual feedback for no search results is not idea. The descriptive text only shows on hover.

lib/private/Search/SearchQuery.php
```diff
@@ -27,89 +27,57 @@
...
 
-	/**
-	 * @inheritDoc
-	 */
 	public function getTerm(): string {
-		return $this->term;
+		return $this->getFilter('term')?->get() ?? '';
+	}
+
+	public function getFilter(string $name): ?IFilter {
```
ChristophWurst, line 55: Idea for the future:

for more ergonomic consumption of the new API I would love to see something like a second parameter for a filter class. Then I can call `getFilter('person', UserFilter::class)`. It only returns if there is a filter and the filter is of type UserFilter.
Right now I first have to check for null, then the filter type before I can access anything.

## Before

```
if ($query->getFilter('person') !== null) {
  $personFilter = $query->getFilter('person');
  if ($personFilter instanceof UserFilter) {
    $email = $personFilter->get()->getEMailAddress();
  }
}
```

## After
```
$email = $query->getFilter('person', UserFilter::class)?->get()?->getEMailAddress();
```

This whole thing can be typed with Psalm to get static guarantees.

Altahrim: The filter `person` can't be anything else than a `UserFilter`. The declaration [is here](https://github.com/nextcloud/server/pull/40618/files/70e8c78f3acb9c14d85c80d001975bd21f45085f#diff-b892980bbda0284c7230f57fd9c5a658b81d5dbfc8cd51fba24be61d4250799bR87)
It applies for each filter (either declared globally or declared by search provider). 

But you are right, the code doesn't reflect that… I'll try to think about it.

ChristophWurst: Filter definitions are only useful when processed by a human. The typing system (used by PHP itself, IDEs and Psalm) is totally unaware of any relations between a filter name and its type.

We can add a second argument to getFilter in the future without backwards compatibility issues. We can make the change later.

Altahrim: Yes, I am agree with you. I am just wondering if there is a way to make typing system understand it, so we don't have to repeat it.

The other thing that bother me is `*Filter` classes are currently private. Maybe there is also something to change there.

ChristophWurst: > The other thing that bother me is `*Filter` classes are currently private. Maybe there is also something to change there.

I hadn't noticed but that is indeed an issue.

So right now one has to use `get()` and then check the PHP type, not the type/class of the filter.

ChristophWurst:
CI is done

PR #40620, opened 2023-09-25
feat(phonenumber): Add OCP wrapper for PhoneNumber utils library

nickvergessen:
* Required for https://github.com/nextcloud/spreed/issues/10410

## Summary

- [x] Extract current code that interacts with 3rdparty `libphonenumber` into an OCP wrapper, so it can be used savely by apps, e.g. Talk for the upcoming SIP dial-out.

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required https://github.com/nextcloud/documentation/pull/11147


ChristophWurst (approved):
80cgu4

nickvergessen:
Docs in https://github.com/nextcloud/documentation/pull/11147

lib/private/PhoneNumberUtil.php
```diff
@@ -0,0 +1,61 @@
...
+ * @since 28.0.0
+ */
+class PhoneNumberUtil implements IPhoneNumberUtil {
+	/**
+	 * {@inheritDoc}
+	 */
+	public function getCountryCodeForRegion(string $regionCode): ?int {
+		$phoneUtil = \libphonenumber\PhoneNumberUtil::getInstance();
+		$countryCode = $phoneUtil->getCountryCodeForRegion($regionCode);
+		return $countryCode === 0 ? null : $countryCode;
```
fsamapoor, line 43: Maybe I'm missing something here, but could it ever return `0` at all? I thought `null` is the return value that indicates an invalid/unsupported number.

nickvergessen: Upstream library returns int only and then 0 means "invalid, nothing matching found". I found this too hidden and opted for converting it to null in this case instead.

fsamapoor: Right. Thank you for the explanation Joas!

fsamapoor (approved):

Issue #38874, opened 2023-06-18
wrong user quota in the user administration


ghost:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Entering 10GB will allocate 10.7GB as the storage size.  1GB becomes 1.1GB, 500 becomes 575.5GB.

### Steps to reproduce

1. Go to user management
2.Specify an Arbitrary size for the user disk space and see how it becomes a different size.

### Expected behavior

-

### Installation method

Community Docker image

### Nextcloud Server version

27

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

PostgreSQL

### Is this bug present after an update or on a fresh install?

Upgraded to a MAJOR version (ex. 22 to 23)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

jalil-salame:
I can replicate this using the NixOS configuration for nextcloud27

LuXc-NL:
Encountered the same issue..

dgedgedge:
### More or less same issue :
I have a user with 850GB of files.
If I try to put 1000GB then it goes to unlimited and / or 1GB...
I saw both.


Server configuration detail

Operating system: Linux 6.1.0-9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1 (2023-05-08) x86_64

Webserver: Apache (fpm-fcgi)

Database: mysql 10.11.3

PHP version: 8.2.7

[...]

PR #40621, opened 2023-09-25
fix: Use correct unit for file sizes and fix user quota

susnux:
## Summary
We always used binary file sizes (base 2, meaning 2048 bytes = 2KiB), but we added the wrong unit (we used the decimal (base 10) unit e.g. 2KB).

This fixes the user management showing a wrong quota (https://github.com/nextcloud/server/issues/38874 for master)

Also the OC_Helper now supports base 10 and base 2 in `computerFileSize`, for backwards compatibility it defaults to force even units like `KB` to be binary.

### Screenshots
Screenshot | comment
---|---
Screenshot_20230925_163450 | As discussed during the Nextcloud conf, we always expect the user to mean the binary size (mostly they are confused by windows using the wrong unit)
Screenshot 2023-09-25 at 16-33-31 Aktive Benutzer - Benutzer - Benutzer - Nextcloud|Quota is now displayed correctly
Screenshot 2023-09-25 at 16-32-58 Dateien - Nextcloud|The file list automatically adapted to the correct unit

## Checklist


artonge (approved):

skjnldsv (approved):

susnux:
cc @sorbaugh and @AndyScherzinger 

The problem:
- On Nextcloud up to (including) 26 we used decimal units like `10 KB` or `20 GB` but we always meant `10 KB = 10 * 1024 bytes` and not `10 KB = 10 * 1000 bytes`.
- On Nextcloud 27 there was a regression (`@nextcloud/files`) that switched to even decimal values, meaning `10KB = 10 * 1000 bytes` but our backend was and is always using binary values (1024).
- On Nextcloud 27.1 we patched this and reverted back to binary values (10* 1024) but with decimal units (`KB` instead of the proper `KiB`).

But for current master the frontend is still broken, we have to options:
1. Everything stays like it was on Nextcloud 26 and below, this requires the `@nextcloud/files` library to be adjusted again.
2. Move to decimal values (we should not do this the only people doing this are hard drive manufacturers and Apple)
3. Show the proper units (`KiB` instead of `KB`, `GiB` instead of `GB`...) this is what this PR is about.

From my point of view only 1 and 3 are really considerable solutions.
Pro 1:
- Everything stays the same UI wise
- Windows is using the wrong units for decades now, so the majority of users are used to them
Con 1:
- See the confusion in the linked issue, having one unit used for multiple different things confuses

Pro 3 (this PR):
- It is clear what unit and file sizes to expect, so less confusion
Con 3:
- Non technical users might choke on the `i` in the unit (but probably it will even not attract attention)
- Changing UI

sorbaugh:
While it's true that lots of users are conditioned to read KB, GB, etc, I think switching to the correct units would be a "soft enough" change UX wise so I would vote for option 3 (but would be fine with option 1).

jancborchardt:
I’d say it’s best in this case to go with what Windows does since @susnux as you say that is what the majority of people are used to. Yes it is not correct, but it is sort of convention on a topic where many people are confused anyway.

sorbaugh:
I don't have a strong opinion on this, so option 1 is fine as well if we can foresee that option 3 would be confusing to people.

susnux:
It was decided to keep current style.

artonge:
@nextcloud/designers can one of you tell us why we should stick to the current style?
The current situation is creating confusion for user and admins as the size does not match what they see in other softwares.

jancborchardt:
My last comment on it is 3 years old https://github.com/nextcloud/server/pull/40621#issuecomment-1766098807

So I am fine to switch if that is also an established way of showing things e.g. in MS and Google products. For details like this it’s best to have least surprise.

artonge:
> So I am fine to switch if that is also an established way of showing things e.g. in MS and Google products. For details like this it’s best to have least surprise.

Neither of them are consistent, they are both displaying GiB values next to the GB unit. So I would rather move away from what they are doing.

provokateurin:
> For details like this it’s best to have least surprise.

I think the current way is the biggest surprise, because it is inconsistent and regularly leads to problems if you actually have to deal with these values (e.g. quotas).

Issue #37492, opened 2023-03-30
[Bug]: Tried to log in "user" but could not verify token


smart7324:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

As soon as I open Nextcloud in a new tab, I get redirected to login page and have to login again. Then always the first login fails/nothing happens, so I have to login twice. I am seeing lots of "Tried to log in "user" but could not verify token" errors in log.

It is only happening on Safari (macOS, iOS, iPadOS), tried several versions, also did a clean install of Nextcloud 26 and still the same. Also tried with another user account on a different Mac.

At first I thought it could be related to #33919, but it doesn't seem to be the case. I really spent many hours in trying to get this fixed, but I have no clue, why it is not working.

### Steps to reproduce

1. Login to Nextcloud in Safari
2. Open another tab and open Nextcloud (alternatively close browser and open it again)
3. You will be redirected to login page and the message "Tried to log in "user" but could not verify token" is in log file.

### Expected behavior

The user should still be logged in and not be redirected to login page.

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

26

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MySQL

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "***REMOVED SENSITIVE VALUE***"
        ],
[...]

PR #40628, opened 2023-09-26
fix(session): Replace remember-me tokens in transaction

ChristophWurst:
* Contributes to https://github.com/nextcloud/server/issues/37492

## Summary

Running `delete from oc_preferences where …` in two READ COMMITTED transactions makes the database serialize the operation and allow detection of a second DELETE that happened concurrently.

This is not a real fix but it can help us get a better insight into the problem.

## TODO

- [x] Improve token rotation
- [ ] Adjust unit tests

## Checklist

### DIFF
--- a/lib/private/AllConfig.php
+++ b/lib/private/AllConfig.php
@@ -345,19 +345,12 @@ class AllConfig implements IConfig {
 		}
 	}
 
-	/**
-	 * Delete a user value
-	 *
-	 * @param string $userId the userId of the user that we want to store the value under
-	 * @param string $appName the appName that we stored the value under
-	 * @param string $key the key under which the value is being stored
-	 */
-	public function deleteUserValue($userId, $appName, $key) {
+	public function deleteUserValue($userId, $appName, $key): bool {
 		// TODO - FIXME
 		$this->fixDIInit();
 
 		$qb = $this->connection->getQueryBuilder();
-		$qb->delete('preferences')
+		$rowsAffected = $qb->delete('preferences')
 			->where($qb->expr()->eq('userid', $qb->createNamedParameter($userId, IQueryBuilder::PARAM_STR)))
 			->andWhere($qb->expr()->eq('appid', $qb->createNamedParameter($appName, IQueryBuilder::PARAM_STR)))
 			->andWhere($qb->expr()->eq('configkey', $qb->createNamedParameter($key, IQueryBuilder::PARAM_STR)))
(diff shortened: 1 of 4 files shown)


lib/private/User/Session.php
```diff
@@ -871,32 +878,63 @@ public function loginWithCookie($uid, $currentToken, $oldSessionId) {
...
+		 *
+		 * The READ COMMITTED isolation level causes the database to serialize
+		 * the DELETE query, making it possible to detect if two concurrent
+		 * processes try to replace the same login token.
+		 * Replacing more than once doesn't work because the app token behind
+		 * the session can only be replaced once.
+		 */
+		$newToken = $this->atomic(function() use ($uid, $currentToken): ?string {
+			// get stored tokens
+			$tokens = $this->config->getUserKeys($uid, 'login_token');
```
ChristophWurst, line 892: food for thought: we might be able to skip the SELECT and go for the DELETE directly

lib/private/User/Session.php
```diff
@@ -871,32 +878,63 @@ public function loginWithCookie($uid, $currentToken, $oldSessionId) {
...
+				]);
+				return null;
+			}
+			$newToken = $this->random->generate(32);
+			$this->config->setUserValue($uid, 'login_token', $newToken, (string)$this->timeFactory->getTime());
+			return $newToken;
+		}, $this->connection);
+
+		// Token verification or replacement failed. Session can't be revived.
+		if ($newToken === null) {
```
rullzer, line 919: so if the transaction aborts it just defaults to  null? Or is there an exception that needs handling somewhere?

lib/private/AllConfig.php
```diff
@@ -339,19 +339,12 @@ public function getUserKeys($userId, $appName) {
...
 
-	/**
-	 * Delete a user value
-	 *
-	 * @param string $userId the userId of the user that we want to store the value under
-	 * @param string $appName the appName that we stored the value under
-	 * @param string $key the key under which the value is being stored
-	 */
-	public function deleteUserValue($userId, $appName, $key) {
+	public function deleteUserValue($userId, $appName, $key): bool {
```
rullzer, line 342: I like this in general :+1:

lib/private/User/Session.php
ChristophWurst: the DELETE and INSERT should not trigger a conflict, even in concurrent situations, because the unique contraint is on appid and configkey. configkey is the new, random token.

only in theory there might be two processes with the same random token ;-)

PR #40633, opened 2023-09-26
Use OCP\Util::getL10N instead of OC::$server->getL10N

come-nc:
## Summary

`OCP` should not refer to private `OC` namespace, ever.
`$server->getL10N` is deprecated.
As there already was a helper in `OCP\Util`, I am using that instead of having to call tedious `\OCP\Server::get(OCP\L10N\IFactory::class)->get('appname')` each time.

## TODO

- [x] Do the same change in `core/` and `lib/private`

## Checklist

### DIFF
--- a/lib/private/Server.php
+++ b/lib/private/Server.php
@@ -1724,7 +1724,7 @@ class Server extends ServerContainer implements IServerContainer {
 	 * @param string $app appid
 	 * @param string $lang
 	 * @return IL10N
-	 * @deprecated 20.0.0
+	 * @deprecated 20.0.0 use \OCP\Util::getL10N instead
 	 */
 	public function getL10N($app, $lang = null) {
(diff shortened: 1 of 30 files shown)


lib/private/Server.php
```diff
@@ -1724,7 +1724,7 @@ public function getL10NFactory() {
 	 * @param string $app appid
 	 * @param string $lang
 	 * @return IL10N
-	 * @deprecated 20.0.0
+	 * @deprecated 20.0.0 use DI of IL10N instead, or \OCP\Util::getL10N as a last resort
```
nickvergessen, line 1727:
```suggestion
	 * @deprecated 20.0.0 use DI of {@see IL10N} or {@see IFactory} instead, or {@see \OCP\Util::getL10N()} as a last resort
```

nickvergessen (approved):

Altahrim (approved):

Issue #37082, opened 2023-03-07
[BITV] 9.3.3.2/0.1 - Note: "Placeholder" attributes are not a substitute for visual labels - they disappear as soon as users start entering data. (1)


AndyScherzinger:
<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-9f92015c98-v0-n1

</details>

* #37083
* #37084
* #37085
* nextcloud/nextcloud-vue#3871
* #37086
* #37087
*  #37088
* #37089
* nextcloud/photos#1685
* #37090

AndyScherzinger:
@JuliaKirschenheuter @artonge @Pytal you sure this is "XS" ?

JuliaKirschenheuter:
> @JuliaKirschenheuter @artonge @Pytal you sure this is "XS" ?

I will change this ticket to "S" for now.

not sure but actually can be covered by https://github.com/nextcloud-libraries/nextcloud-vue/pull/4394. All places have to be  adjusted to new component.

JuliaKirschenheuter:
Places to be replaced:

![Screenshot from 2023-08-28 19-00-07](https://github.com/nextcloud/server/assets/6078378/ddc37a5a-7ac0-4db1-a280-c64adfcd21c1)
![Screenshot from 2023-08-28 18-59-51](https://github.com/nextcloud/server/assets/6078378/e94405f3-86fc-421a-88c6-4cbdca186ae3)
![Screenshot from 2023-08-28 18-59-38](https://github.com/nextcloud/server/assets/6078378/541cd3ac-2f1b-498b-9a2d-b88ee847179e)
![Screenshot from 2023-08-28 18-53-09](https://github.com/nextcloud/server/assets/6078378/48466355-d6ab-4a98-8956-0d051f77e68a)
![Screenshot from 2023-08-28 18-52-53](https://github.com/nextcloud/server/assets/6078378/2782ba27-3c17-4761-85f3-49ff93b37037)
![Screenshot from 2023-08-28 18-52-15](https://github.com/nextcloud/server/assets/6078378/0568065c-2e7e-4779-a13b-3dfbf2002674)
![Screenshot from 2023-08-28 18-46-25](https://github.com/nextcloud/server/assets/6078378/f6ed5c3b-13e6-4131-9b31-60ecc1d90c08)
[...]

PR #40643, opened 2023-09-26
Set visible label for input field

JuliaKirschenheuter:
* Part of https://github.com/nextcloud/server/issues/37082

## Summary

Replace password dialog with dialog from `nextcloud-password-confirmation` library. No visual changes.

## Checklist


ShGKme:
Looks a bit weird... Would it be valid here to use "This action requires you to confirm your password" as a label?

artonge (approved):

JuliaKirschenheuter:
> Looks a bit weird... Would it be valid here to use "This action requires you to confirm your password" as a label?

I think "This action requires you to confirm your password" is too long for a label. That would change width of modal window too.
Would you have some other suggestion? But actually we can do all improvements after a11y certification.

JuliaKirschenheuter:
@nimishavijay would you approve it?

nimishavijay:
Would using a password field with an internal label work? 


https://github.com/nextcloud/server/assets/52440189/8c4f237a-f1ea-4f4f-aa4a-a8dd8454a3db


If not, then I propose the following changes:

image

- [ ] Change heading to "Confirm your password"
- [ ] Change subline to "This action needs authentication"
- [ ] Use an external label, but on the same row as the input field

core/src/OC/password-confirmation.js
ShGKme, line 1: After migrating got `@nextcloud/password-confirmation` all the private methods and properties are not used anymore.

So we need to remove as well:
- [ ] properties:
  - [ ] `callback`
  - [ ] 'pageLoadTime'
- [ ] `init` method and its usage in `src/core/init.js`
  - [ ] including its usage
- [ ] `_confirmPassword`

---

`requiresPasswordConfirmation` is still needed, but its implementation must be replaced with `isPasswordConfirmationRequired` from the same package

---

Then, I think we should also mark these methods of OC as deprecated in favor of the package

ShGKme (changes requested):

core/src/OC/password-confirmation.js
JuliaKirschenheuter: > * pageLoadTime

If i would remove `pageLoadTime` then i have to remove ore replace `serverTimeDiff` and `timeSinceLogin` as well as `requiresPasswordConfirmation` method. Should i remove all this logic?

ShGKme (approved):

ShGKme:
@artonge Do you want to re-review after a large change?

susnux (approved):

core/src/OC/password-confirmation.js
```diff
@@ -22,110 +22,23 @@
  *
  */
 
-import _ from 'underscore'
-import $ from 'jquery'
-import moment from 'moment'
-import { generateUrl } from '@nextcloud/router'
-
-import OC from './index.js'
+import { confirmPassword, isPasswordConfirmationRequired } from '@nextcloud/password-confirmation'
```
Pytal, line 25: `isPasswordConfirmationRequired` is not exported https://github.com/nextcloud-libraries/nextcloud-password-confirmation/blob/607a4f425adcff19f6ea56df1d4ba13762129ee0/src/main.ts#L10 ?

ShGKme: nice catch

JuliaKirschenheuter: needs new release https://github.com/nextcloud-libraries/nextcloud-password-confirmation/pull/614

JuliaKirschenheuter: https://github.com/nextcloud-libraries/nextcloud-password-confirmation/pull/609

ShGKme, line 25: We also need to import styles here.

It works because they were already imported from another place, but we cannot count on that:
- May break if another place changes
- May break on chunks separating

```suggestion
import { confirmPassword, isPasswordConfirmationRequired } from '@nextcloud/password-confirmation'
import '@nextcloud/password-confirmation/style.css'
```

Pytal (approved):
Still needs https://github.com/nextcloud/server/pull/40874

Otherwise good

ShGKme:
Cypress `users_modify` seems to be a flaky test, works fine locally:
image

Another test is a known problem

JuliaKirschenheuter:
failing tests are unrelated

PR #40644, opened 2023-09-26
fix: encode uri for delete, restore and favorite

kesselb:
## Summary

To encode # and other characters properly

Similar to https://github.com/nextcloud/server/pull/39842 but different ;)

`#` is a valid character for our filenames but also for uris in general and therefore not encoded by default. 
To use `#` in filenames we need to encode it.

## TODO

- [ ] CI
- [ ] Review

## Checklist


kesselb:
Version for stable27: https://github.com/nextcloud/server/pull/40642

ChristophWurst (approved):
Makes sense

Pytal (approved):

ChristophWurst:
CI failure is unrelated to the code changes

skjnldsv:
UI is broken and show lots of encoded files now :)
Reverting this in followup. The data model should NOT include encoded string, it should not be request-aware. A Node object contains raw data and should never be transformed/oriented towards a specific usage (here a request context for example)

apps/files_trashbin/src/services/trashbin.ts
```diff
@@ -46,7 +46,7 @@ const resultToNode = function(node: FileStat): File | Folder {
 
 	const nodeData = {
 		id: node.props?.fileid as number || 0,
-		source: generateRemoteUrl('dav' + rootPath + node.filename),
+		source: generateRemoteUrl(encodePath('dav' + rootPath + node.filename)),
```
skjnldsv, line 49: `encodePath` is no even defined here...

PR #40656, opened 2023-09-27
feat(backgroundjob): Schedule job after <timestamp>

ChristophWurst:
## Summary

For a planned execution in the future. Avoids a job that picked up again and again until it's time has come.

Credits go to https://github.com/nextcloud/server/blob/f8f437072ac13a4556dea18219d55f11466497e5/apps/dav/lib/BackgroundJob/UserStatusAutomation.php#L174-L177 and @nickvergessen.

## TODO

- [x] Adjust API

## Checklist

  - https://github.com/nextcloud/documentation/pull/11157

### DIFF
--- a/lib/private/BackgroundJob/JobList.php
+++ b/lib/private/BackgroundJob/JobList.php
@@ -41,6 +41,10 @@ use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IConfig;
 use OCP\IDBConnection;
 use Psr\Log\LoggerInterface;
+use function get_class;
+use function json_encode;
+use function md5;
+use function strlen;
 
 class JobList implements IJobList {
 	protected IDBConnection $connection;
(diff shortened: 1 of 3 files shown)


tests/lib/BackgroundJob/DummyJobList.php
```diff
@@ -46,6 +46,10 @@ public function add($job, $argument = null): void {
 		}
 	}
 
+	public function scheduleAfter(string $job, int $runAfter, $argument = null): void {
+		$this->add($job, $argument);
```
marcelklehr, line 50:
```suggestion
		$this->add($job, $argument, $runAfter);
```

Would have expected $runAfter as third argument here, no?

lib/private/BackgroundJob/JobList.php
```diff
@@ -55,11 +59,10 @@ public function __construct(IDBConnection $connection, IConfig $config, ITimeFac
 		$this->logger = $logger;
 	}
 
-	/**
-	 * @param IJob|class-string<IJob> $job
-	 * @param mixed $argument
-	 */
-	public function add($job, $argument = null): void {
+	public function add($job, $argument = null, int $firstCheck = null): void {
```
nickvergessen, line 62:
```suggestion
	public function add($job, $argument = null, ?int $firstCheck = null): void {
```

Or is it okay like that?

ChristophWurst: :potato: :potato: apparently

https://psalm.dev/r/5a7fbe6ae1
https://3v4l.org/K9mhc

nickvergessen: Roger 🙈

come-nc: Both are equivalent but ?int is a lot clearer

come-nc (approved):

nickvergessen (approved):
Then time to squash and merge?

marcelklehr (approved):

Issue #37080, opened 2023-03-07
[BITV] 9.3.3.1/8.1 - Modale dialog "Authentication required" - If the "Password" input field is left empty, an error message appears. This error message "Failed to authenticate, please try again" is unspecific and does not give any information about which input caused the error. (1)


AndyScherzinger:
Active users - Users - Nextcloud
https://try.nextcloud.com/ltd/a11y/index.php/settings/users

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-c3b56e67ab-v8-n1

</details>

JuliaKirschenheuter:
@michaelnissenbaum could you please check it if this issue is already resolved?

In my opinion it is but i want to be sure. Thanks a lot!

![Peek 2023-08-24 09-44](https://github.com/nextcloud/server/assets/6078378/ece7e4e8-2d54-431f-aeac-e21b95a6614f)

JuliaKirschenheuter:
https://github.com/nextcloud/server/pull/38839

michaelnissenbaum:
Hi @JuliaKirschenheuter. No, from my perspective, the issue still remains. This concerns the modal dialog with the error message "Failed to authenticate, please try again." The error message is unspecific and unclear in this case.
![CleanShot 2023-08-24 at 20 41 33@2x](https://github.com/nextcloud/server/assets/48948959/67f47d0a-de12-4aff-ad87-6d4c120e516d)

Additionally I found another issue. The button with the eye icon doesn't have alternative text - an empty "aria-label" attribute is used.

![CleanShot 2023-08-24 at 20 45 46@2x](https://github.com/nextcloud/server/assets/48948959/20641c3c-5681-4b16-94ae-3bfad6ed147f)

And the button with the arrow has an incorrect alternative text - aria-label="Clear text".
![CleanShot 2023-08-24 at 20 46 41@2x](https://github.com/nextcloud/server/assets/48948959/9e67db51-710d-4174-a110-0fa9c248d702)

JuliaKirschenheuter:
Thank you @michaelnissenbaum!

JuliaKirschenheuter:
@artonge  am i right that we would need much more backend differentiation for showing a detailed information? 
Or could i probably just change this message: 

[...]

PR #40662, opened 2023-09-27
Correct `aria-label` on action input and text field

JuliaKirschenheuter:
* Part of https://github.com/nextcloud/server/issues/37080#issuecomment-1692235248
## Summary

Before:
And the button with the arrow has an incorrect alternative text - aria-label="Clear text".
CleanShot 2023-08-24 at 20 46 41@2x

After:
Screenshot from 2023-09-27 17-28-51

## Checklist


apps/settings/src/components/Users/UserRow.vue
```diff
@@ -68,6 +69,7 @@
 			<template v-if="idState.editing && settings.canChangePassword && user.backendCapabilities.setPassword">
 				<NcTextField class="user-row-text-field"
 					:class="{'icon-loading-small': idState.loading.password}"
+					:trailing-button-label="t('settings', 'Add new password')"
```
artonge, line 72: Shouldn't it be "Set new password" ? Same thing for the label.

apps/settings/src/components/GroupListItem.vue
```diff
@@ -61,10 +62,12 @@ import NcAppNavigationItem from '@nextcloud/vue/dist/Components/NcAppNavigationI
 import NcCounterBubble from '@nextcloud/vue/dist/Components/NcCounterBubble.js'
 
 import AccountGroup from 'vue-material-design-icons/AccountGroup.vue'
+import NcTextField from '@nextcloud/vue/dist/Components/NcTextField'
```
artonge, line 65: Is it needed ?

Pytal (approved):

apps/settings/src/components/GroupListItem.vue
```diff
@@ -41,6 +41,7 @@
 			<NcActionInput v-if="id !== 'admin' && id !== 'disabled' && settings.isAdmin"
 				ref="displayNameInput"
 				icon="icon-edit"
+				:trailing-button-label="t('settings', 'Rename group')"
```
ShGKme, line 44: It seems, `NcActionInput` has no `trailing-button-label` prop. We need to change `NcActionInput` to at least pass "Submit" as a text (because it always has submit trailing button) or add an optional prop.

apps/settings/src/components/Users/UserRow.vue
```diff
@@ -68,11 +69,12 @@
 			<template v-if="idState.editing && settings.canChangePassword && user.backendCapabilities.setPassword">
 				<NcTextField class="user-row-text-field"
 					:class="{'icon-loading-small': idState.loading.password}"
+					:trailing-button-label="t('settings', 'Set new password')"
 					:show-trailing-button="true"
 					:disabled="idState.loading.password || isLoadingField"
 					:minlength="minPasswordLength"
 					maxlength="469"
-					:label="t('settings', 'Add new password')"
+					:label="t('settings', 'Set new password')"
```
ShGKme, line 77: Isn't it a problem that 2 interactive elements have the same label?

ShGKme (changes requested):

artonge (approved):

apps/settings/src/components/GroupListItem.vue
JuliaKirschenheuter: Ok, should i leave it like it was before (without any change) if it is anyway "submit"?

artonge: "Submit new ..." ?

ShGKme: > Ok, should i leave it like it was before (without any change) if it is anyway "submit"?

Currently, it is not `Submit` but `Clear text`, which is not correct. It should be fixed in `@nextcloud/vue/NcActionInput`

ShGKme: Sorry, I miss-looked, `NcActionInput` proxies attrs to `NcTextField`, it works fine

ShGKme (approved):

apps/settings/src/components/GroupListItem.vue
JuliaKirschenheuter: It works, i've checked. Thank you!

susnux (approved):

PR #40674, opened 2023-09-28
feat: allow external drop and add dropzone

### LINKED ISSUE #4483: Introduce a notice for instant upload / drag&drop in below file list
As discussed in #4273 we want to break down @Espina2  ideas into smaller improvements. I propose to introduce the notice for drag & drop in the whitespace below the files. This is in particular helpful in public sharing / to non-nextcloud users.

Here is a quick sketch:
files-drag drop

If the file list takes the whole height, the info could move directly above the first file as in @Espina2 proposal.

@nextcloud/designers

skjnldsv:
<table>
<tr>
 <td colspan="2"> Drag and drop in current folder (with tall header because of recommended files) </td>
</tr>
<tr>
 <td colspan="2">

 </td>
</tr>
<tr>
 <td> Drop in sub-folder </td>
 <td> Narrow height header look </td>
</tr>
<tr>
 <td> </td>
 <td> </td>
</tr>
</table>

- [x] Allow dropping external files from your OS to Nextcloud
- [x] Add a drop zone with a way to hide it (see #4483)

Fix #4483

@nextcloud/designers for visual input


skjnldsv:
> How does it handle rich workspaces?

Above
image

apps/files/src/components/FileEntry.vue
```diff
@@ -781,15 +782,15 @@ export default Vue.extend({
 		isFileNameValid(name) {
 			const trimmedName = name.trim()
 			if (trimmedName === '.' || trimmedName === '..') {
-				throw new Error(this.t('files', '"{name}" is an invalid file name.', { name }))
+				throw new Error(t('files', '"{name}" is an invalid file name.', { name }))
 			} else if (trimmedName.length === 0) {
-				throw new Error(this.t('files', 'File name cannot be empty.'))
+				throw new Error(t('files', 'NcFile name cannot be empty.'))
```
artonge, line 787: On purpose or find and replace ?
I think it makes sense to keep "File" here as it is an error message.

artonge (approved):
Code looks good, but not a fan of how it looks :/
Have we looked at how others are doing it ?

skjnldsv:
> but not a fan of how it looks :/

Me neither ahah!
Quickly mocked up from the linked issue and following Jan's recommendations:

> If the file list takes the whole height, the info could move directly above the first file as in Espina2 proposal.
>
> I do agree though that it should be on the top of the file list, maximum 2 rows height, and with an x-Close button. Then it’s always in the same place and doesn’t depend on the size of the view.

I'm not merging until Design team reviewed it :)

nimishavijay:
Hm, that is using up a lot of vertical space. Maybe we can make the following changes:

- [ ] The height lessened by around 20px 
- [ ] The icon can be made smaller, and also I would recommend using a different icon, maybe this [upload icon](https://fonts.google.com/icons?selected=Material%20Icons%3Afile_upload%3A), because this one looks pretty similar to a download icon actually
- [ ] The close button can be closer to the text, maybe even next to it. 

- [ ] Question: Is this shown for all folders or just the home folder and is it remembered if it is closed?

Checking Google Drive and OneDrive, neither of these show a drag and drop notice (Gdrive shows it in an empty folder), so I'm wondering if drag and drop is just expected behaviour at this point. We could definitely show it in the empty content in an empty folder though. but ccing @jancborchardt for this as he was involved in the original issue :)

skjnldsv:
> * Question: Is this shown for all folders or just the home folder and is it remembered if it is closed?

all

> Checking Google Drive and OneDrive, neither of these show a drag and drop notice (Gdrive shows it in an empty folder), so I'm wondering if drag and drop is just expected behaviour at this point. We could definitely show it in the empty content in an empty folder though. but ccing @jancborchardt for this as he was involved in the original issue :)

No opinion for me. But note that I do agree that it's hard to know where to drop when you want to drop on the current folder. 
e.g your current folder have a lot of folders, where do you drop your files?
image

Pytal (approved):

artonge:
Bold opinion: maybe we should not support dropping to a subfolder et simply display a dropzone covering the whole file list.

jancborchardt:
> Bold opinion: maybe we should not support dropping to a subfolder et simply display a dropzone covering the whole file list.

I sort of agree with that, yeah.

If it should be possible to drop into a subfolder, we can display the dropzone in the area of controls bar & table title.

I would say important is that the height of the page doesn’t change by having a dropzone notice inserted.

skjnldsv:
@jancborchardt, so, what do I do, because this is not clear.
Do I remove the drop to subfolder feature and only display an overlay on top of the entire page when dragging?
If so, please provide visuals

apps/files/src/components/FileEntry.vue
```diff
@@ -487,7 +488,7 @@ export default Vue.extend({
 
 			try {
 				const previewUrl = this.source.attributes.previewUrl
-					|| generateUrl('/core/preview?fileid={fileid}', {
+					|| generateUrl('/core/preview?fileId={fileid}', {
```
skjnldsv, line 491: @JuliaKirschenheuter here is the fix! :see_no_evil:

AndyScherzinger: Thanks a lot @skjnldsv 💙

skjnldsv:
@nimishavijay please advise updated design in the first comment!

skjnldsv:
> @nimishavijay @jancborchardt please advise updated design in the first comment!

Design tuning can/will be a followup though

nimishavijay:
Nice! This looks good already, nice touch with the folder opening icon :) 

Follow up points:

- [ ] While dragging in the header, if your cursor is hovering over the header, there should be some hover feedback, similar to when your cursor is over a subfolder. So the background of the header section can change to `--color-primary-light` and the font color to `--color-primary` for example. 
- [ ] Since this is shown only while dragging, the message could be even more attention grabbing, so we could make the font bold
  

(ref. Gdrive's drop indication which is _really_ attention grabbing)
image

skjnldsv:
> * While dragging in the header, if your cursor is hovering over the header, there should be some hover feedback, similar to when your cursor is over a subfolder. So the background of the header section can change to `--color-primary-light` and the font color to `--color-primary` for example.

Those two colours doesn't really go together? Is this a mistake? :)

Issue #40636, opened 2023-09-26
Prepare for PHP 8.0 end of life


come-nc:
PHP version 8.0 official security support will end on November 26: https://www.php.net/supported-versions.php

Also, we may not easily keep supporting it as our dependencies might drop 8.0 as well when adding support for 8.3. (I think this is not currently the case, if you have information please comment)

Preparing for this drop, we need to first update the following CI:

- [x] github:
  - sftp.yml
  - static-code-analysis.yml
  - phpunit-32bits.yml
  - update-psalm-baseline.yml
  - s3-primary-integration.yml
  - s3-primary.yml
  - ftp.yml
  - performance.yml
  - [x] vendor-bin/psalm
  - [x] vendor-bin/cs-fixer https://github.com/nextcloud/server/pull/39271
- [x] drone:
  - [x] ghcr.io/nextcloud/continuous-integration-php8.0:latest
  - [x] ghcr.io/nextcloud/continuous-integration-litmus-php8.0:latest
  - [x] ghcr.io/nextcloud/continuous-integration-samba-native-php8.0:latest
  - [x] ghcr.io/nextcloud/continuous-integration-samba-non-native-php8.0:latest ?
  - [x] ghcr.io/nextcloud/continuous-integration-acceptance-php8.0:latest
  - [x] ghcr.io/nextcloud/continuous-integration-integration-php8.0:latest
  - [x] ghcr.io/nextcloud/continuous-integration-mysql-8.0:mysql-8.0
  - [x] ghcr.io/nextcloud/continuous-integration-php8.0-memcached:latest
  - [x] ghcr.io/nextcloud/continuous-integration-webdav-apache-php8.0:latest

isdnfan:
hi team another heads up - for some reason NC30 still supports PHP 8.0 which is EOL since 26.11.**2023**!! 
please keep in mind PHP 8.1 "*Active support*" ends in less than 2 months on on 1 Jan 2025

come-nc:
> hi team another heads up - for some reason NC30 still supports PHP 8.0 which is EOL since 26.11.**2023**!! please keep in mind PHP 8.1 "_Active support_" ends in less than 2 months on on 1 Jan 2025

No, Nextcloud 30 dropped support for PHP 8.0, as you can see in both the documentation: https://docs.nextcloud.com/server/stable/admin_manual/installation/system_requirements.html#server and source code: https://github.com/nextcloud/server/blob/v30.0.0/lib/versioncheck.php

Where did you get your information?

PR #40678, opened 2023-09-28
chore: Install phpunit via vendor bin

ChristophWurst:
## Summary

This means we can all run the same version. And when switching to stableX it can pull the corresponding version too. Also local dev envs and CI run the same phpunit.

~~Set to draft until I know if this actually works.~~ It works.

## TODO

- [x] Add PHPUnit 9 in a new composer bin
- [x] Add composer scripts to run main test configuration and files_external test configuration
- [x] Migrate github workflow scripts to use pinned phpunit instead of a global one

## Follow-ups

- Migrate autotest.sh

## Checklist

### DIFF
--- a/vendor-bin/phpunit/composer.json
+++ b/vendor-bin/phpunit/composer.json
@@ -1,0 +1,11 @@
+{
+    "config": {
+        "sort-packages": true,
+        "platform": {
+            "php": "8.0"
+        }
+    },
+    "require": {
+        "phpunit/phpunit": "9.6"
+    }
+}
--- a/.github/workflows/ftp.yml
+++ b/.github/workflows/ftp.yml
@@ -49,13 +49,13 @@ jobs:
         uses: shivammathur/setup-php@c5fc0d8281aba02c7fda07d3a70cc5371548067d #v2.25.2
         with:
           php-version: ${{ matrix.php-versions }}
-          tools: phpunit:9
           extensions: mbstring, fileinfo, intl, sqlite, pdo_sqlite, zip, gd
         env:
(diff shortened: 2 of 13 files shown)


blizzz (approved):

nickvergessen (approved):

vendor-bin/phpunit/composer.json
```diff
@@ -0,0 +1,11 @@
+{
+    "config": {
+        "sort-packages": true,
+        "platform": {
+            "php": "8.0"
```
come-nc, line 5: 8.1 possible?
See #40636

ChristophWurst: Sure. Once master doesn't work with 8.0 anymore we can bump the platform here :+1:

come-nc: Well the point of #40636 is to update the CI before dropping 8.0.

Because I fear dropping 8.0 will be decided late in the release cycle.

We have a php lint step that checks all supported versions, it’s not needed for phpunit to run on the oldest one.

ChristophWurst: If I drop the platform requirement here, composer will pull sub dependencies of phpunit that are incompatible with PHP8.0. So CI won't pass. It's really just about one incompatible package. There is an older version that works with 8.0 as well, which is installed now.

IMO dropping PHP8.0 support and removing it from CI is an atomic step. Either master doesn't support 8.0 and we don't test it, or it still supports it and we have tests. Dropping 8.0 CI now is a risk if 8.0 support is not removed in the this release cycle.

PR #40700, opened 2023-09-29
Remove space

Valdnet:


rakekniven (approved):

szaimen (approved):

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -278,7 +278,7 @@ export default {
 
 	computed: {
 		title() {
-			let title = t('files_sharing', 'Share with ')
+			let title = t('files_sharing', 'Share with')
 			if (this.share.type === this.SHARE_TYPES.SHARE_TYPE_USER) {
 				title = title + this.share.shareWithDisplayName
```
blizzz, line 283: correct me if i am wrong, but the space is missing then here where it is concatenated with the display name.  At least here, didn't check further.
```suggestion
			let title = t('files_sharing', 'Share with') + " "
			if (this.share.type === this.SHARE_TYPES.SHARE_TYPE_USER) {
				title = title + this.share.shareWithDisplayName
```

blizzz (changes requested):

szaimen:
@Valdnet this needs a manual rebase and then you can run the compile command again

blizzz:
@nickvergessen was my suggestion actually ok, or are there smarter/better ways? perhaps also with RTL in mind?

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -278,7 +278,7 @@ export default {
 
 	computed: {
 		title() {
-			let title = t('files_sharing', 'Share with ')
+			let title = t('files_sharing', 'Share with') + " "
 			if (this.share.type === this.SHARE_TYPES.SHARE_TYPE_USER) {
 				title = title + this.share.shareWithDisplayName
```
nickvergessen, line 283: As per https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#improving-your-translations this is not good.
It would be better to have a full t() in each of the else-if branches?

AndyScherzinger:
@Valdnet do you want to take care of the manual rebase and fix the issues from the comments or close and started from scratch?

Issue #38911, opened 2023-06-20
[Bug]: PDF previews don't work after upgrade to 27 - Error 400


Chartman123:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

On one of my instances PDF previews do no longer work after upgrading to NC 27. On both my other instances it works as expected.

The failing one is using the official Docker images. The other two instances are classic installations at shared web hosters. Other differences: DB (pgsql <-> mariadb, redis <-> non redis)

PDF previews work with Imaginary disabled, so I think it's related to Imaginary.

Also, preview generation works for `x >= 257` and `y >= 257`

### Steps to reproduce

1. Browse to folder with PDF files
2. No previews, just the PDF logo

### Expected behavior

See previews of files

### Installation method

Community Docker image

### Nextcloud Server version

27

### Operating system

Other

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database engine version

PostgreSQL

### Is this bug present after an update or on a fresh install?

Upgraded to a MAJOR version (ex. 22 to 23)

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [X] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "htaccess.RewriteBase": "\/",
        "memcache.local": "\\OC\\Memcache\\APCu",
        "apps_paths": [
            {
                "path": "\/var\/www\/html\/apps",
                "url": "\/apps",
                "writable": false
            },
            {
                "path": "\/var\/www\/html\/custom_apps",
                "url": "\/custom_apps",
                "writable": true
            }
        ],
        "instanceid": "***REMOVED SENSITIVE VALUE***",
[...]

PR #40731, opened 2023-09-30
Generate all preview sizes for Imaginary

enoch85:
Based on discussions here: https://github.com/nextcloud/server/issues/38911#issuecomment-1741819188

This fixes the case were not all previews are generated, for example in the activity view: https://github.com/nextcloud/server/issues/38911#issuecomment-1738886106

* Resolves: # 
https://github.com/nextcloud/server/issues/38911

## TODO

- [ ] ...

## Checklist

- [X] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)


Chartman123:
Is `getSmallImagePreview` still needed then? I couldn't find any other call for this method...

solracsf:
This superseedes https://github.com/nextcloud/server/pull/40670 as `getSmallImagePreview()` function should also be removed

enoch85:
> Is `getSmallImagePreview` still needed then? I couldn't find any other call for this method...

I guess the devs have to choose which one to merge.

solracsf:
Nevertheless, for this PR to be complete, function has to be removed.

enoch85:
FWIW, just tested this with a client that had 1000s of PDFs, everything is super fast and it generates everything. 

I think we should close the other PR, let's have a vote. :-1:  for no, and :+1: for keeping it. gogogo

Chartman123:
Just go on and close it... The other fix would be lost by merging this here ;)

szaimen (approved):
Tested by the community and confirmed to make things work but cannot judge the code changes

enoch85:
@juliushaertl Would you mind having a look at this? It removes a function which makes everything work again.

ChristophWurst:
(partly) reverts the changes from https://github.com/nextcloud/server/pull/31864 so :eyes: @skjnldsv

nfebe (commented):
I am skeptical about deleting the performance related previews all together, the real question is *why does the deleted snippet or `getSmallImagePreview` not generate the previews? 

Theoretically we should be able to generate smaller previews no?

enoch85:
> I am skeptical about deleting the performance related previews all together, the real question is what does the deleted snippet or `getSmallImagePreview` to generate the previews?
> 
> Theoretically we should be able to generate smaller previews no?

The function is buggy and doesn't make a huge impact when used. 

Ok, if you're on a Raspberry Pi or some legacy hardware - sure, but not for the "normal" user of Nextcloud and certainly not Enterprise users.

nfebe (approved):

enoch85:
@Fenn-CS Can you please help me with the backport?

Fenn-CS:
Sure @enoch85 would look into it.

PR #40745, opened 2023-10-02
fix ocm-provider rewrite rules

ArtificialOwl:
Fix the rewrite rules about `ocm-provider` in 2 locations:

- the default .htaccess, 
- the generated part of htaccess from `Setup.php`

I were only able to reproduce the issue in _some_ conditions, each time it seems that this patch fixed them.


sorbaugh (approved):

Altahrim (approved):
Not tested, but LGTM

Issue #40676, opened 2023-09-28
Making the Metadata API public


sorbaugh:
# Requirements

### Existing

- [x] Store width and height for pictures
- [x] Store GPS coordinate for pictures
- [x] Store user readable location for pictures

### Planned

- [ ] nextcloud/photos#87
- [ ] nextcloud/server#40242
- [ ] nextcloud/server#39714
- [ ] Generate blurhash for pictures for files, photos, and social

### Potential

- [ ] Tags on files
- [ ] Favorite state

### Already implemented

- [x] Store arbitrary metadata as a string and link it to a fileid and a data name.
- [x] Allow the metadata to be manually and granularly exposed to WebDAV requests.
- [x] Trigger providers to generated metadata during files:scan and files' updates (NodeWrittenEvent).
- [ ] Delete metadata upon files deletion (NodeDeletedEvent). But needs adjustments: nextcloud/server#34424

### Already missing

- [ ] Support dependency among providers to allow providers to depend on the work of other providers. Example: MapMediaToPlaceJob in the Photos app.
- [ ] Execute the provider in a background process to not slow down the request. Example: MapMediaToPlaceJob in the Photos app.
- [ ] Support listing all different metadata value for a given user and metadata name. Example: listing all places linked to pictures in the Photos app.
- [ ] Support listing all files with a given metadata value and name for a given user. Example: listing all pictures linked a place in the Photos app.

### Needed for the new use cases

- [ ] Ability to use a metadata in an orderby directive in a WebDAV SEARCH request (sorting picture by taken date)
- [ ] Automatically handle the exposition of the metadata in WebDAV requests. (show EXIF data in sidebar)
- [ ] Allow clients to set the content of a metadata. (live photos)

artonge:
# Technical requirements

## Database

Indexing all the metadata is unnecessary, so we have one table to store the all the metadata, and another table to store indexed metadata.
The second table contains a copy of some metadata contained in the first table.
The second table can be rebuilt from the first one. This opens the way to be smarter about which data we keep in the indexed table. In the future, we could drop some of its data, or populate it on demand.

The `last_update` column is used to know how old is a given set of metadata. The `unique` column is used to avoid race condition when updating metadata

The indexed table contains two columns for the value. One is a `varchar`, the other a `bigint`. This allows to optimize the usage of both type of data.

**oc_metadata**
Name | Type
-- | --
fileid | `varchar`
metadata | `text`
unique | `varchar`
last_update | `datetime`

[...]

PR #40761, opened 2023-10-04
IFilesMetadata

ArtificialOwl:
https://github.com/nextcloud/server/issues/40676

### Introduction

This is a feature that allow any app to store files metadata.

When a file is created or modified, an event is broadcasted so that a listener can:

- request for an identical event, but on the background job process
- set metadata, key and value
- define metadata index
- help building queries to search on an indexed metadata
- help building queries that returns metadata linked to `fileid`
 
### tasks

- [x] model Metadata
  * [x] store type of metadata (string, int, bool, array, float)
  * [x] serialize/deserialize and import from database
- [x] store metadata,
  * [x] update database only if new data (or truly edited) is available
  * [x] avoid parallel update in database
  * [x] delete metadata based on key
  * [x] delete metadata based on prefix
- [x] index metadata (file id, metadata key, metadata value)
  * [x] generate entries in metadata_index based on value type
- [x] queries
  * [x] helper to left join metadata based on file_id
  * [x] helper to search on metadata key
  * [x] helper to search on metadata value
  * [x] helper to get metadata key and metadata value table field
  * [x] helper to convert row's data to metadata object
- [x] events
  * [x] listener on new file
  * [x] listener on edited file
  * [x] on request, dispatch event on background job process
  * [x] occ metadata:get 

### How to add metadata from your own app:

_lib/AppInfo/Application.php_
```

		$context->registerEventListener(MetadataLiveEvent::class, UpdateFilesMetadata::class);
		$context->registerEventListener(MetadataBackgroundEvent::class, UpdateFilesMetadata::class);
```

_UpdateFilesMetadata.php_
```
use OCP\EventDispatcher\Event;
use OCP\EventDispatcher\IEventListener;
use OCP\FilesMetadata\Event\MetadataBackgroundEvent;
use OCP\FilesMetadata\Event\MetadataLiveEvent;

class UpdateFilesMetadata implements IEventListener {
	public function __construct() {}

	public function handle(Event $event): void {
		if (!($event instanceof MetadataLiveEvent) &&
			!($event instanceof MetadataBackgroundEvent)) {
			return;
		}

		$node = $event->getNode();
		// only few files are interesting.
		if ($node->getMimetype() !== 'image/jpeg') {
			return;
		}

		// we only work on process initiated as background job, so ask for a re-run later
		if ($event instanceof MetadataLiveEvent) {
			$event->requestBackgroundJob();
			return;
		}

		$fd = $node->fopen('rb');
		if (false === $fd) {
			return;
		}

		try {
			$oldBufferSize = stream_set_chunk_size($fd, 1);
			$data = @exif_read_data($fd, 'ANY_TAG', true);
			stream_set_chunk_size($fd, $oldBufferSize);

                        // we have exif data in $data, let's index the photo taken date 
			$metadata = $event->getMetadata();
			$metadata->setInt('photo-taken', strtotime($data['EXIF']['DateTimeOriginal']), true);
		} catch (\Exception $e) {
		}
	}
}
```

### How to use the _(current)_ QueryHelper
```
		/** @var ConnectionAdapter $conn */
		$conn = \OCP\Server::get(ConnectionAdapter::class);
		$qb = $conn->getQueryBuilder();

		$qb->select('file.fileid', 'file.path')->from('filecache', 'file');

                // linking metadataQuery to existing query builder
		$metadataQuery = $this->filesMetadataManager->getMetadataQuery($qb, 'file', 'fileid');
                // add metadata to the request's results
		$metadataQuery->retrieveMetadata();
                // only returns result havin
		$metadataQuery->joinIndex('another');

		$expr = $qb->expr();
		$qb->where(
			$expr->lte(
				$metadataQuery->getMetadataValueIntField(),
				$qb->createNamedParameter(1, IQueryBuilder::PARAM_INT)
			)
		);
		$qb->orderBy($metadataQuery->getMetadataValueIntField(), 'desc');

```


lib/private/FilesMetadata/Model/MetadataQuery.php
```diff
@@ -0,0 +1,182 @@
...
+	 * @see self::retrieveMetadata()
+	 * @since 28.0.0
+	 */
+	public function extractMetadata(array $row): IFilesMetadata {
+		$fileId = (array_key_exists($this->fileIdField, $row)) ? $row[$this->fileIdField] : 0;
+		$metadata = new FilesMetadata($fileId);
+		try {
+			$metadata->importFromDatabase($row, $this->alias . '_');
+		} catch (FilesMetadataNotFoundException $e) {
+		}
```
come-nc, line 76: Log or comment

lib/private/FilesMetadata/Model/MetadataQuery.php
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	/**
+	 * @param string $metadataKey metadata key
+	 *
+	 * @inheritDoc
+	 * @since 28.0.0
+	 */
+	public function enforceMetadataKey(string $metadataKey): void {
```
come-nc, line 116: Why is it the same as enforceMetadataValue?

ArtificialOwl, line 116: still needed since joinIndex do the same !?

lib/private/FilesMetadata/Model/MetadataQuery.php
```diff
@@ -0,0 +1,182 @@
...
+	/**
+	 * @param string $value metadata value
+	 * @inheritDoc
+	 * @since 28.0.0
+	 */
+	public function enforceMetadataValue(string $value): void {
+		$expr = $this->queryBuilder->expr();
+		$this->queryBuilder->andWhere(
+			$expr->eq(
+				$this->getMetadataKeyField(),
```
ArtificialOwl, line 135: `->getMetadataValueField()`

lib/private/FilesMetadata/Service/IndexRequestService.php
```diff
@@ -0,0 +1,177 @@
...
+			match ($filesMetadata->getType($key)) {
+				IMetadataValueWrapper::TYPE_STRING
+				=> $this->insertIndexString($fileId, $key, $filesMetadata->get($key)),
+				IMetadataValueWrapper::TYPE_INT
+				=> $this->insertIndexInt($fileId, $key, $filesMetadata->getInt($key)),
+				IMetadataValueWrapper::TYPE_STRING_LIST
+				=> $this->insertIndexStringList($fileId, $key, $filesMetadata->getStringList($key)),
+				IMetadataValueWrapper::TYPE_INT_LIST
+				=> $this->insertIndexIntList($fileId, $key, $filesMetadata->getIntList($key))
+			};
```
come-nc, line 83: Should be in a transaction for drop -> insert, maybe?

lib/private/FilesMetadata/Service/MetadataRequestService.php
```diff
@@ -0,0 +1,160 @@
...
+	private function generateSyncToken(): string {
+		$chars = 'qwertyuiopasdfghjklzxcvbnmQWERTYUIOPASDFGHJKLZXCVBNM1234567890';
+
+		$str = '';
+		$max = strlen($chars);
+		for ($i = 0; $i < 7; $i++) {
+			try {
+				$str .= $chars[random_int(0, $max - 2)];
+			} catch (Exception $e) {
+			}
```
come-nc, line 155: ?

lib/public/FilesMetadata/IFilesMetadataManager.php
```diff
@@ -0,0 +1,116 @@
...
+use OCP\FilesMetadata\Model\IMetadataQuery;
+
+/**
+ * Manager for FilesMetadata; manage files' metadata.
+ *
+ * @since 28.0.0
+ */
+interface IFilesMetadataManager {
+	public const PROCESS_LIVE = 1;
+	public const PROCESS_BACKGROUND = 2;
```
come-nc, line 42: `@since`

lib/public/FilesMetadata/Model/IMetadataValueWrapper.php
```diff
@@ -0,0 +1,297 @@
...
+ * @since 28.0.0
+ */
+interface IMetadataValueWrapper extends JsonSerializable {
+	public const TYPE_STRING = 'string';
+	public const TYPE_INT = 'int';
+	public const TYPE_FLOAT = 'float';
+	public const TYPE_BOOL = 'bool';
+	public const TYPE_ARRAY = 'array';
+	public const TYPE_STRING_LIST = 'string[]';
+	public const TYPE_INT_LIST = 'int[]';
```
come-nc, line 46: `@since`

core/Migrations/Version28000Date20231004103301.php
```diff
@@ -0,0 +1,80 @@
...
+			$table = $schema->createTable('files_metadata_index');
+			$table->addColumn('id', Types::BIGINT, [
+				'autoincrement' => true,
+				'notnull' => true,
+				'length' => 15,
+				'unsigned' => true,
+			]);
+			$table->addColumn('file_id', Types::BIGINT, ['notnull' => false, 'length' => 15]);
+			$table->addColumn('meta_key', Types::STRING, ['notnull' => false, 'length' => 31]);
+			$table->addColumn('meta_value', Types::STRING, ['notnull' => false, 'length' => 63]);
```
artonge, line 68: Half of the API default to string, and the other require specifying a type. Shall we make the type explicit everywhere ?

```suggestion
			$table->addColumn('meta_value_string', Types::STRING, ['notnull' => false, 'length' => 63]);
```

core/Command/FilesMetadata/Get.php
```diff
@@ -0,0 +1,119 @@
...
+			$metadata = $this->filesMetadataManager->refreshMetadata(
+				$node[0],
+				IFilesMetadataManager::PROCESS_LIVE | IFilesMetadataManager::PROCESS_BACKGROUND
+			);
+		} else {
+			$metadata = $this->filesMetadataManager->getMetadata($fileId);
+		}
+
+		if ($input->getOption('as-array')) {
+			$output->writeln(json_encode($metadata->asArray(), JSON_PRETTY_PRINT));
```
artonge, line 112: `JSON_THROW_ON_ERROR` could be added for all JSON function calls.

icewind1991 (commented):
Do I understand correctly that the metadata is only loaded on search requests atm?

I don't see any logic to trigger the loading for propfinds

lib/public/Files/Search/ISearchComparison.php
```diff
@@ -55,6 +56,23 @@ public function getType();
...
+	public function getExtra(): string;
+
+	/**
+	 * returns if data are 'extra' or not
+	 *
+	 * @return bool
+	 * @since 28.0.0
+	 */
+	public function isExtra(): bool;
+
```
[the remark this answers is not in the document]
ArtificialOwl: it might not be necessary

ArtificialOwl:
> Do I understand correctly that the metadata is only loaded on search requests atm?

GET request also wants to get their metadata, 
and metadata will also be updated from iOS client (another PR is on its way)

> I don't see any logic to trigger the loading for propfinds

If we don't load known indexed metadata, exceptions are thrown by https://github.com/nextcloud/3rdparty/blob/e6c45c6c0d4f92ffec621446fcc3b34584772b13/icewind/searchdav/src/DAV/SearchHandler.php#L102-L128

We could edit the code there, but not sure this is something we want ?

artonge (approved):

come-nc (approved):
3000 lines /o\

nickvergessen:
This PR has no version bump and therefore breaks all master setups because while the table is used, the migration to create it did not trigger.

nickvergessen:
Follow up fix in https://github.com/nextcloud/server/pull/41351

PR #40763, opened 2023-10-04
fix(OC_Template.php): ensure we throw all errors

Psycojoker:
## Summary

In one branch of the try/catch combination, in one specific path no error was thrown at all, thus the user ended up with a 500 error page online but absolutely no error anyway in the logs.

## Checklist

- ~~[ ] Screenshots before/after for front-end changes~~
- ~~[ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)~~

### DIFF
--- a/lib/private/legacy/OC_Template.php
+++ b/lib/private/legacy/OC_Template.php
@@ -311,6 +311,9 @@ class OC_Template extends \OC\Template\Base {
 			print("The server encountered an internal error and was unable to complete your request.\n");
 			print("Please contact the server administrator if this error reappears multiple times, please include the technical details below in your report.\n");
 			print("More details can be found in the server log.\n");
+
+			// and then throw it again to log it at least to the web server error log
+			throw $e;
 		}
 		die();
 	}


kesselb:
If you reach line 316, that means no error occurred while logging the exception. 
Are you sure there was nothing in your nextcloud logs?

Psycojoker:
> Are you sure there was nothing in your nextcloud logs?

Yes, I've looked at every possible log imaginable:
- nextcloud.log ofc
- everything possible in /var/log/**/*.log
- all fpm process, activating logs etc...
- every possible related systemd process

Adding those lines was the only possible way to get the logging of the exact error somewhere (the error was that nextcloud couldn't connect to the database because it was stopped)

kesselb:
I assume we fixed that already for master/28 with https://github.com/nextcloud/server/pull/37458.

Psycojoker:
Maybe it would be a good idea to still have a fallback in the case the logging system broke for whatever reason again?

susnux:
We no longer use OC_Template and the remaining functions are deprecated as of 32.

PR #40766, opened 2023-10-04
Make OAuth2 authorization code expire

julien-nc:
Oauth authorization codes now expire 10 minutes after being created.
To know if a row in `oauth2_access_tokens` is an authorization code or an access token, we keep track of the number of delivered access tokens. If no token was delivered, it means the row still contains an authorization code (in the `encrypted_token` column).

* Refuse to deliver an access token from an expired authorization code (and immediately delete the related DB entry)
* Expired authorization codes are cleaned up every month
* Fix: A refresh token could be used as an authorization code (with the `authorization_code` grant type). This is no longer possible
* A few tests have been implemented

### DIFF
--- a/apps/oauth2/lib/BackgroundJob/CleanupExpiredAuthorizationCode.php
+++ b/apps/oauth2/lib/BackgroundJob/CleanupExpiredAuthorizationCode.php
@@ -1,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Julien Veyssier <julien-nc@posteo.net>
+ *
+ * @author Julien Veyssier <julien-nc@posteo.net>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+namespace OCA\OAuth2\BackgroundJob;
+
+use OCA\OAuth2\Db\AccessTokenMapper;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\IJob;
+use OCP\BackgroundJob\TimedJob;
+use OCP\DB\Exception;
+use Psr\Log\LoggerInterface;
+
+class CleanupExpiredAuthorizationCode extends TimedJob {
+
+	public function __construct(
+		ITimeFactory $timeFactory,
+		private AccessTokenMapper $accessTokenMapper,
+		private LoggerInterface $logger,
+
+	) {
+		parent::__construct($timeFactory);
+		// 30 days
+		$this->setInterval(60 * 60 * 24 * 30);
+		$this->setTimeSensitivity(IJob::TIME_INSENSITIVE);
+	}
+
+	/**
+	 * @param mixed $argument
+	 * @inheritDoc
+	 */
+	protected function run($argument) {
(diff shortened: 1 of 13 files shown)


AndyScherzinger (approved):

apps/oauth2/lib/Migration/Version011603Date20230620111039.php
```diff
@@ -0,0 +1,84 @@
...
+			}
+			if ($dbChanged) {
+				return $schema;
+			}
+		}
+
+		return null;
+	}
+
+	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options) {
```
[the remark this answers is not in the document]
nickvergessen:
```suggestion
	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
```

apps/oauth2/lib/Migration/Version011603Date20230620111039.php
```diff
@@ -0,0 +1,84 @@
...
+use OCP\Migration\SimpleMigrationStep;
+
+class Version011603Date20230620111039 extends SimpleMigrationStep {
+
+	public function __construct(
+		private IDBConnection $connection,
+	) {
+	}
+
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options) {
```
nickvergessen, line 43:
```suggestion
	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
```

apps/oauth2/lib/Migration/Version011603Date20230620111039.php
```diff
@@ -0,0 +1,84 @@
...
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('oauth2_access_tokens')) {
+			$table = $schema->getTable('oauth2_access_tokens');
+			$dbChanged = false;
+			if (!$table->hasColumn('code_created_at')) {
+				$table->addColumn('code_created_at', Types::BIGINT, [
+					'notnull' => true,
+					'default' => 0,
```
nickvergessen, line 53:
```suggestion
					'unsigned' => true,
```

Default 0 is automatically, unsigned will allow more numbers

apps/oauth2/lib/Migration/Version011603Date20230620111039.php
```diff
@@ -0,0 +1,84 @@
...
+				]);
+				$dbChanged = true;
+			}
+			if (!$table->hasColumn('token_count')) {
+				$table->addColumn('token_count', Types::BIGINT, [
+					'notnull' => true,
+					'default' => 0,
```
nickvergessen, line 60:
```suggestion
					'unsigned' => true,
```

Default 0 is automatically, unsigned will allow more numbers

apps/oauth2/lib/BackgroundJob/CleanupExpiredAuthorizationCode.php
nickvergessen, line 1: File is missing from autoloader

apps/oauth2/lib/Migration/Version011603Date20230620111039.php
julien-nc: Why not explicitly setting the default to 0? I didn't know about it, maybe others don't know either.

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -70,16 +74,20 @@ public function __construct(
...
 	 *
 	 * @param string $grant_type Token type that should be granted
-	 * @param string $code Code of the flow
-	 * @param string $refresh_token Refresh token
-	 * @param string $client_id Client ID
-	 * @param string $client_secret Client secret
+	 * @param string|null $code Code of the flow
+	 * @param string|null $refresh_token Refresh token
+	 * @param string|null $client_id Client ID
+	 * @param string|null $client_secret Client secret
```
provokateurin, line 80:
```suggestion
	 * @param ?string $code Code of the flow
	 * @param ?string $refresh_token Refresh token
	 * @param ?string $client_id Client ID
	 * @param ?string $client_secret Client secret
```

provokateurin (approved):
API changes LGTM

apps/oauth2/lib/Migration/Version011603Date20230620111039.php
nickvergessen: Also fine 👍🏼

PR #40770, opened 2023-10-04
fix(cypress): empty group, tag and ci-build-id when not using cypress cloud

max-nextcloud:
Cypress is currently failing in master and stable branches due to #40736 
grafik
https://github.com/nextcloud/server/actions/runs/6408685873/job/17398449900

This fixes it by leaving the settings that only work with `record` empty.

Tested here:
https://github.com/nextcloud/server/actions/runs/6409023706/job/17399426108

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -103,14 +103,14 @@ jobs:
       - name: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }} cypress tests
         uses: cypress-io/github-action@59810ebfa5a5ac6fcfdcfdf036d1cd4d083a88f2 # v6.5.0
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cypress.yml
```diff
@@ -120,6 +120,7 @@ jobs:
           TESTING: true
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+          DEBUG: '@cypress/github-action'
```
artonge, line 123: To remove ?

artonge (approved):

mejo- (approved):

PR #40782, opened 2023-10-05
fix(files_external): basic auth user storage trigger

### LINKED ISSUE #40776: [Bug]: F2V: Not possible to enter credentials for external storage
### Bug description

When the authentication method for an external storage is set to "User entered, store in database", the file view should display a form to allow the user to enter the credentials.

This was previously done by: https://github.com/nextcloud/server/blob/stable26/apps/files_external/js/statusmanager.js#L389

And I do not see any modal opening in the current implementation: https://github.com/nextcloud/server/blob/7e5e8873070b7fd3071a5eaab5ed467686baa774/apps/files_external/src/actions/enterCredentialsAction.ts#L76-L77

Might be looking at the wrong place. But I was not able to trigger any modal during my testing.

### Steps to reproduce

1. Create an external storage
2. Set the authentication method to "User entered, store in database"
3. Go to the F2V file list
4. Click on the external storage
5. Nothing happens

### Expected behavior

A modal should be displayed to prompt the user for credentials.

### Nextcloud Server version

master

@skjnldsv

skjnldsv:
Fix https://github.com/nextcloud/server/issues/40776


apps/files_external/src/actions/enterCredentialsAction.ts
```diff
@@ -75,11 +75,13 @@ export const action = new FileAction({
 
 	async exec(node: Node) {
 		// always resolve auth request, we'll process the data afterwards
-		const response = await axios.get(generateOcsUrl('/apps/files_external/api/v1/auth'), {
-			validateStatus: () => true,
+		// Using fetch as axios have integrated auth handling and X-Requested-With header
+		const response = await fetch(generateOcsUrl('/apps/files_external/api/v1/auth'), {
```
artonge, line 79: You mean @nc/axios? Maybe you can use the bare axios. But it is fine like that.

apps/files_external/lib/Controller/ApiController.php
```diff
@@ -126,6 +126,7 @@ public function getUserMounts(): DataResponse {
 
 	/**
 	 * @NoAdminRequired
+	 * @NoCSRFRequired
```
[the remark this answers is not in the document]
skjnldsv: Yes, this method is just a dummy 401/auth request.
it doesn't matter, it does not trigger anything related to Nextcloud.

apps/files_external/src/actions/enterCredentialsAction.ts
skjnldsv: Nah, it was also interfering weirdly :thinking:

artonge (approved):

susnux (approved):

ArtificialOwl (approved):

AndyScherzinger (approved):

PR #40785, opened 2023-10-05
fix: Log critical session renewal and logout paths

ChristophWurst:
* Contributes to: https://github.com/nextcloud/server/issues/37492

## Summary

Extracted and extended the generic logging pieces of https://github.com/nextcloud/server/pull/40628.

## TODO

- [x] Add logging

## Checklist

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -783,6 +783,11 @@ class Session implements IUserSession, Emitter {
 		try {
 			$dbToken = $this->tokenProvider->getToken($token);
 		} catch (InvalidTokenException $ex) {
+			$this->logger->warning('Session token is invalid because it does not exist', [
+				'app' => 'core',
+				'user' => $user,
+				'exception' => $ex,
+			]);
 			return false;
 		}
 
@@ -802,6 +807,10 @@ class Session implements IUserSession, Emitter {
 		}
 
 		if (!$this->checkTokenCredentials($dbToken, $token)) {
+			$this->logger->warning('Session token credentials are invalid', [
+				'app' => 'core',
+				'user' => $user,
+			]);
 			return false;
 		}
 
@@ -877,9 +886,9 @@ class Session implements IUserSession, Emitter {
 		$tokens = $this->config->getUserKeys($uid, 'login_token');
 		// test cookies token against stored tokens
 		if (!in_array($currentToken, $tokens, true)) {
-			$this->logger->info('Tried to log in {uid} but could not verify token', [
+			$this->logger->info('Tried to log in but could not verify token', [
 				'app' => 'core',
-				'uid' => $uid,
+				'user' => $uid,
 			]);
 			return false;
 		}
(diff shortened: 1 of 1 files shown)


lib/private/User/Session.php
```diff
@@ -878,25 +888,38 @@ public function loginWithCookie($uid, $currentToken, $oldSessionId) {
 			$this->logger->info('Tried to log in {uid} but could not verify token', [
 				'app' => 'core',
 				'uid' => $uid,
+				'user' => $uid,
```
Altahrim, line 891: We log `$uid` twice. Should it be `$user`?
```suggestion
				'user' => $user,
```
Later in code, we have other cases like `'user' => $uid` or `'uid' => $user`

nickvergessen: `$user` is an `IUser` which can not be written to the log.

The thing here is that `user` is a magic array key so he wants to set that, while not touching `uid` which is used in the log message itself.

Altahrim: Thanks for the explanation!

lib/private/User/Session.php
```diff
@@ -935,10 +958,17 @@ public function logout() {
...
+				$token = $this->session->getId();
+				$this->tokenProvider->invalidateToken($token);
+				$this->logger->debug('Session token invalidated before logout', [
+					'user' => $user->getUID(),
+				]);
 			} catch (SessionNotAvailableException $ex) {
 			}
 		}
+		$this->logger->info('Logging out', [
+			'user' => $user === null ? null : $user->getUID(),
```
Altahrim, line 970:
```suggestion
			'user' => $user?->getUID(),
```

Altahrim (approved):

lib/private/User/Session.php
```diff
@@ -878,25 +888,38 @@ public function loginWithCookie($uid, $currentToken, $oldSessionId) {
 			$this->logger->info('Tried to log in {uid} but could not verify token', [
 				'app' => 'core',
 				'uid' => $uid,
+				'user' => $uid,
 			]);
```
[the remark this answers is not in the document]
ChristophWurst: I can actually remove the username from the message and allow tools to aggregate the same error for multiple users :)

lib/private/User/Session.php
ChristophWurst: I was tempted but also afraid this would bite me in backports :grimacing:

lib/private/User/Session.php
```diff
@@ -935,10 +957,17 @@ public function logout() {
...
-				$this->tokenProvider->invalidateToken($this->session->getId());
+				$token = $this->session->getId();
+				$this->tokenProvider->invalidateToken($token);
+				$this->logger->debug('Session token invalidated before logout', [
+					'user' => $user->getUID(),
+				]);
 			} catch (SessionNotAvailableException $ex) {
 			}
 		}
+		$this->logger->info('Logging out', [
```
come-nc, line 968: Log all logouts as info feels a bit too noisy, no?

nickvergessen: debug maybe?

come-nc (approved):

PR #40796, opened 2023-10-06
Another system config: allow_user_to_change_email_address

### LINKED ISSUE #16823: 'allow_user_to_change_display_name' only affecting 'display name' not 'email address'
As suggested in the closed https://github.com/nextcloud/server/issues/642

We need an option to disable change of display name without disabling the possibility to change the email address.

(Why do we need this?
for example:
Users (school) change their display names - make 'funny' things with shared content - rechange their name again - in activity log is a wrong name, and a wrong user is blamed...
Not good - but the users need to be able to change their email address by themselves)

'allow_user_to_change_display_name' = false should only affect display name.

Another setting
'allow_user_to_change_email_address'
could do the same for the email address.

Thanks for reevaluation.

theronakpatel:
* Resolves: https://github.com/nextcloud/server/issues/16823


joshtrichards:
This has too many conflicts to be tested as-is at the moment.

susnux:
Replaced with https://github.com/nextcloud/server/pull/51745

PR #40822, opened 2023-10-08
fix(files): correct typo that broke thumbnails

starypatyk:
## Summary

Loading thumbnails in files seems to be broken after #39998 - apparently due to a typo in https://github.com/nextcloud/server/commit/f9d2e3af0cfb087ed57c8c4a445618e0d24dde8f.

If I understand correctly the query parameter names are case sensitive. Since the `fileId` parameter was accidentally changed to `fileid`, the parameter value was not passed to the `PreviewController::getPreviewByFileId` method.

This change seems to correct the issue.


skjnldsv (approved):

skjnldsv:
Yep, I fixed it in https://github.com/nextcloud/server/pull/40674 too :see_no_evil: 
Thanks for checking it out! :+1:

skjnldsv:
Maybe let's go with https://github.com/nextcloud/server/pull/40674 as this is ready to get in already :)
But thanks again for the fix @starypatyk

starypatyk:
Fair enough @skjnldsv. Good that the fix is merged now. :+1:

PR #40850, opened 2023-10-10
Read configuration from files during installation

Kixunil:
## Summary

The installation script required users to either enter sensitive information in command line arguments, which was unsecure, or enter it interactively in the input, which prevented automated installation. This commit makes it possible to pre-configure the server in config.php so that automated installation can be performed safely.

I intentionally didn't add tests for now as I seek guidance on how to test CLI/installation and whether this approach is good.

## TODO

- [ ] add tests

## Checklist

- [ ] ~~Screenshots before/after for front-end changes~~ - doesn't apply

### DIFF
--- a/core/Command/Maintenance/Install.php
+++ b/core/Command/Maintenance/Install.php
@@ -30,6 +30,9 @@
  */
 namespace OC\Core\Command\Maintenance;
 
+require_once __DIR__ . '/../../../lib/base.php';
+
+use OC;
 use bantu\IniGetWrapper\IniGetWrapper;
 use InvalidArgumentException;
 use OC\Installer;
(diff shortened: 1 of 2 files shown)


kesselb:
https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/automatic_configuration.html does not work for you?

It looks wrong to include base.php here.

core/Command/Maintenance/Install.php
```diff
@@ -111,40 +114,60 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+			'database-name' => 'dbname',
+			'database-host' => 'dbhost',
+			'admin-user' => 'adminlogin',
+			'admin-pass' => 'adminpass',
+			'admin-email' => 'adminemail',
+			'data-dir' => 'datadirectory',
+		];
+		$option = $input->getOption($name);
+		if (is_null($option)) {
+			return OC::$config->getValue($option_mapping[$name]);
```
kesselb, line 131: I assume you can already access the config mapper via $this->config (it's injected via di).

Kixunil:
Wow, I totally missed autoconfig.php! Thanks! I will test it and close this PR if it works fine for me or let you know in case of problems to find a better solution.

Kixunil:
Looks like it works, however:
* `occ maintenance:install` doesn't work, loading `index.php` does - is this intended?
* It requires creating `CAN_INSTALL` file which is not documented and it's not clear what's its purpose. Is it related to security? Are there any dangers around creating it?
* It loses admin password by hashing it - I guess this is intentional but better document it.

So I think this PR is not needed but it'd be nice to update documentation or improve the installation process (or both).

Issue #40559, opened 2023-09-21
Show status instead of email in contacts/people menu


ChristophWurst:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

## Is your feature request related to a problem? Please describe.

As a Nextcloud user I want to see what other people are doing. If I click the *people menu* (formerly known as *contacts menu*) in the top right I see the first x contacts sorted by their name. For each contact I see the email address.

## Describe the solution you'd like

Remove the email address subline and show the user status instead. Sort the contacts by their last status update.

- [x] Blocked by https://github.com/nextcloud/server/issues/40558

## Additional context

N/a

ChristophWurst:
@jancborchardt @marcoambrosini @nimishavijay could you create a mockup for this? Should the menu also show the user online status (green dot etc)?

jancborchardt:
- We can pretty much have it look the same as the "Participants" tab in Talk, except with the "(moderator)" indicator removed. That is also very similar to the "Recent statuses" dashboard widget (which should ideally also be adjusted to include the online status).
- Click events and menu items would still stay the same for now, this step is only about the subline, online status indicator and sorting.
- Currently the list is very short at 4 entries (and also not clear one can scroll). We can make it larger to 6½ entries so that there are more entries and the half cut off entry suggests scrollability more.
![image](https://github.com/nextcloud/server/assets/925062/4aa0238e-a6fb-4268-9e10-6b5023b1d500)

ChristophWurst:
Medium-size technical barrier for this feature: https://github.com/nextcloud/server/blob/master/core/src/views/ContactsMenu.vue smells like a Vue components but the internals are still jquery and backbone https://github.com/nextcloud/server/blob/5ddffcb7b41326b5d3d2528e2a0e9cb6f8f8d76a/core/src/views/ContactsMenu.vue#L57-L59. 

To not increase technical debt further I should probably rewrite the menu in Vue first.

ChristophWurst:
This will revert the changes of https://github.com/nextcloud/server/pull/28822. cc @kesselb

ChristophWurst:
[...]

PR #40852, opened 2023-10-10
feat(contactsmenu): Show user status

ChristophWurst:
* Resolves part 1 of https://github.com/nextcloud/server/issues/40559

## Summary

Show the user status for system contacts if they have a status set. Show the email address otherwise.

| Before | After |
|--------|--------|
| Bildschirmfoto vom 2023-10-17 13-18-45 | Bildschirmfoto vom 2023-10-18 20-33-12 | 

## TODO

- [x] ...

## Checklist

  - https://github.com/nextcloud/documentation/pull/11226


ChristophWurst:
requesting early design review based on the before/after screenshots

I think the feature is done. I only have to adjust/write tests

nimishavijay (commented):
Looks good! small cosmetic changes:

- [x] "Show all contacts" remove ellipsis and text in bold so it looks like a button 
- [x] size of avatar could be 44px (ignore if already the case, it looks smaller than 44px to me)
- [x] increase space between avatar and text by 2px 
- [x] decrease space between main text and subline by 2px
Basically similar to the styling of a list item in, for eg. the contacts app.
image
- [ ] Also, if clicking on the item opens the profile anyway, I'd say we don't need a separate profile button, it could just be a 3 dot menu

nimishavijay:
> Do you want me to remove the profile action or just move it into the actions menu? 

Ah okay, I understood, so for emails that would be a mail action? 

How about we make Talk the priority action and keep profile in the 3 dot menu. That way, when you open the contact search you see users whom you can Talk to and email addresses that you can send mails to, and if you click on the user you would go to their profile anyway. 

My idea is that you are probably looking for a contact to do something specific like chat or email more than view the profile. What do you think?

jancborchardt (approved):
Looks nice! :) I’d agree with @nimishavijay that having "Talk" as the primary action makes more sense as it’s a more direct action.
However then clicking the name/status part of the person should do the same to not have 4 different click areas (avatar, name, action icon, action menu), and "Profile" has to be put into the action menu – so your call if this should be in here or in a separate pull request @ChristophWurst.

ChristophWurst:
Let's do these changes separately :+1:

lib/private/Contacts/ContactsMenu/ActionProviderStore.php
```diff
@@ -47,7 +48,6 @@ public function __construct(
 	}
 
 	/**
-	 * @return IProvider[]
```
nickvergessen, line 50:
```
	 * @return IProvider[]|IBulkProvider[]
```

lib/private/Contacts/ContactsMenu/Entry.php
```diff
@@ -142,7 +159,7 @@ public function getProperty(string $key): mixed {
 	}
 
 	/**
-	 * @return array{id: int|string|null, fullName: string, avatar: string|null, topAction: mixed, actions: array, lastMessage: '', emailAddresses: string[], profileTitle: string|null, profileUrl: string|null}
+	 * @return array{id: int|string|null, fullName: string, avatar: string|null, topAction: mixed, actions: array, lastMessage: '', emailAddresses: string[], profileTitle: string|null, profileUrl: string|null, status: string|null, statusMessage: null|string, statusIcon: null|string, isUser: bool, uid: mixed}
```
nickvergessen, line 162: I guess you need to recreate openapi docs?

nickvergessen (approved):
Only checked PHP

lib/private/Contacts/ContactsMenu/Entry.php
ChristophWurst: > OpenAPI specifications up to date. Carry on

lib/private/Contacts/ContactsMenu/ActionProviderStore.php
ChristophWurst: https://psalm.dev/r/705c8b2f87 vs https://psalm.dev/r/eef0455894 

:smiling_face_with_tear:

nickvergessen: Oh come on psalm :(

nickvergessen: Hmm could call the method bulkProcess in the bulk provider, then a provider could be both :D
But well the else if is already taking care. So yolo from here

marcoambrosini (approved):

ChristophWurst:
Red CI unrelated. Cypress is random and postgres timed out.

ChristophWurst:
I hovered the button to stop automerge and when I clicked the page jumped and my click went to the upate button. Great. Just four more hours.

PR #40854, opened 2023-10-10
feat(files): migrate template picker

skjnldsv:
Ref: https://github.com/nextcloud/server/issues/39914

image


juliushaertl:
- Should also trigger the inline name editing after the file was created?
- I feel like we should still open the file directly after creating (as it used to be before) but not sure how easy that is to trigger after entering the name

skjnldsv:
> * Should also trigger the inline name editing after the file was created?

Absolutely!

> I feel like we should still open the file directly after creating (as it used to be before) but not sure how easy that is to trigger after entering the name

Good point too! Can I do a followup? :)

skjnldsv:
Ping everyone!

juliusknorr (approved):

apps/files/src/views/TemplatePicker.vue
```diff
@@ -199,12 +209,11 @@ export default {
 
 		async onSubmit() {
 			this.loading = true
-			const currentDirectory = getCurrentDirectory()
-			const fileList = OCA?.Files?.App?.currentFileList
+			const currentDirectory = new URL(window.location.href).searchParams.get('dir') || '/'
```
artonge, line 212: Why can't we use something like ` this.$route.params.dir` ?

apps/files/src/views/TemplatePicker.vue
```diff
@@ -216,35 +225,43 @@ export default {
...
-						fileList,
-						fileActions: fileList?.fileActions,
-						fileInfoModel: model,
-					})
-				}
+				// Update files list
+				emit('files:node:created', node)
 
+				// Open the new file
+				window.OCP.Files.Router.goToRoute(
```
artonge, line 248: Wasn't there another way to navigate without using `window.OCP` ?

artonge (approved):

apps/files/src/views/TemplatePicker.vue
skjnldsv: The templatepicker is standalone and not aware of the Files app router.
We do not expose the route raw to keep things clean and limit what devs can do with it

apps/files/src/views/TemplatePicker.vue
skjnldsv: Same as above: The templatepicker is standalone` and works without the Files app

Issue #40800, opened 2023-10-06
[Bug]: Caching in background jobs (almost) worthless when instance uses APCu for local caching


ChristophWurst:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

APCu cache memory is volatile. For web servers it lives as long as the webserver process. For CLI it lives as long as the cache.

We recommend APCu for local caching because it's faster than Redis and the TCP roundtrip.

Nextcloud uses a lot of background (cron) processes. When set up for performance, background jobs are triggered by a system cron. Within this cron, the APCu cache does work. Between two cron executions all cached values are lost. This severely limits the usefulness of caching.

### Steps to reproduce

1. Set up an instance with APCu as local cache
2. Cache data in a background job
3. Build on the cached value later on

### Expected behavior

Cached value is there

### Installation method

None

### Nextcloud Server version

27

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

ChristophWurst:
How is this not a bug?

PR #40860, opened 2023-10-11
feat: Give more control over caching

ChristophWurst:
## Summary

* Make it clear when caching is actually available and populating makes sense
* Add factory method for in-memory caches

This is quick PR putting some ideas around https://github.com/nextcloud/server/issues/40800 into code.

## Checklist


ChristophWurst:
There are method to check availability before acquiring an instance. This isn't strictly required.

PR #40864, opened 2023-10-11
fix(federation): Use `sharing.federation.allowSelfSignedCertificates` config for all OCM requests

nickvergessen:
## Summary

The config already existed in [many places](https://github.com/nextcloud/server/blob/8b9e7e235dd7711bba31441e9a8b09dc33f85d38/lib/private/Federation/CloudFederationProviderManager.php#L127), but was missed in these two, breaking it, because federation requests actually didn't work for self-signed certificates:

## Checklist

### DIFF
--- a/apps/files_sharing/lib/Controller/ExternalSharesController.php
+++ b/apps/files_sharing/lib/Controller/ExternalSharesController.php
@@ -29,6 +29,7 @@ use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\Http\Client\IClientService;
+use OCP\IConfig;
 use OCP\IRequest;
 
 /**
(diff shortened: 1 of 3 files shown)


come-nc (approved):

apps/files_sharing/lib/Controller/ExternalSharesController.php
```diff
@@ -37,25 +38,14 @@
...
-	 * @param \OCA\Files_Sharing\External\Manager $externalManager
-	 * @param IClientService $clientService
-	 */
-	public function __construct($appName,
-								IRequest $request,
-								\OCA\Files_Sharing\External\Manager $externalManager,
-								IClientService $clientService) {
+	public function __construct(
+		string $appName,
+		IRequest $request,
```
mejo-, line 43: What's the reason that `$appName` and `IRequest` are not defined as `private`? :thinking:

mejo- (approved):
Looks good apart from my comment.

apps/files_sharing/lib/Controller/ExternalSharesController.php
nickvergessen: It would conflict with the extended class which already defines them.

Issue #36965, opened 2023-03-01
[BITV] 9.1.4.10/6.2 - At a browser window width of 320 CSS pixels (or at a browser window width of 1280 CSS pixels and 400% zoom magnification), the page contents do not wrap - some information and functions are not available without users having to scroll horizontally. (1)


AndyScherzinger:
The issue affects both Firefox and Chromium-based browsers, with the whole page being affected in the case of Chromium-based browsers.

Page — Personal info
Personal info - Personal settings - Nextcloud
https://try.nextcloud.com/ltd/a11y/index.php/settings/user

<details>
<summary>Details</summary>

https://report.bitvtest.de/default-en/d63601ac-cb34-4645-8256-66bec78964a0.html#checkpoint-81d891d46b-v6-n2

</details>

JuliaKirschenheuter:
https://ergebnis.bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-1-4-10-inhalte-brechen-um

ShGKme:
Note:

> In the alternative test with 1280px viewport width and 400% zoom, it should not be rated negatively if permanently positioned content (such as drop-out menus) is not fully accessible due to the very low available height of the browser window. The only check is whether content successfully breaks into a view that does not need to be scrolled horizontally. To check the loss of information and functionality, the test at 320px viewport width is crucial.

So only the width is checked, not the height.

JuliaKirschenheuter:
closing ticket. As follow up (not exactly related to this ticket with horizontal scrolling):

- [BITV] Fix left sidebar from a11y perspective, https://github.com/nextcloud/server/issues/40954
- [BITV]: Sidebar is broken, https://github.com/nextcloud/activity/issues/1360

ShGKme:
> So only the width is checked, not the height.

UPD: this is not correct.

According to the requirement:

> 2.2 Alternative test with 1280px viewport width
> 1. In a browser that supports zoom enlargement by 400% (e.g. B. Chrome) set the browser window to a width of 1280 CSS pixels and a height of 1024 CSS pixels. For this, the Resize-function of Web Developer Toolbar use.

Height `1024px + 400%` or `256px` also should work.

ShGKme:
Files | Sidebar | Settings dialog | Apps sidebar
---|---|---|---
[...]

PR #40871, opened 2023-10-11
Fix my-apps view on mobile

szaimen:
Address part of https://github.com/nextcloud/server/issues/36965

### before
image

### After

https://github.com/nextcloud/server/assets/42591237/0d4d2615-ec4c-4f69-b674-42304dd270a9


ShGKme:
> Address part of #36965

Note: for tables it's also allowed to use horizontal scrolling

> Exceptions apply to content that requires a two-dimensional layout, e.g. B. Images, maps, diagrams, videos, games, presentations, data tables and application interfaces in which the processing of content requires the permanent availability of toolbars.

https://ergebnis.bitvtest.de/pruefschritt/bitv-20-web/bitv-20-web-9-1-4-10-inhalte-brechen-um

ShGKme:
I'm not sure this is a good approach to removing the data on a small screen. From the issue:

> some information and functions are not available without users having to scroll horizontally

In this solution, we don't have the scroll, but some information is not available at all.

Maybe we can have a horizontal scrolling here, or switch to a list layout (using isMobile util from nextcloud/vue)?

skjnldsv (approved):

JuliaKirschenheuter (approved):

JuliaKirschenheuter:
> In this solution, we don't have the scroll, but some information is not available at all.

Which functionality is missing? 

> Maybe we can have a horizontal scrolling here, or switch to the grid layout (using isMobile util from nextcloud/vue)?

The issue is that sticky header had not full width and action buttons were not fully available, like

Screenshot from 2023-10-12 10-50-04

szaimen:
> I'm not sure this is a good approach to remove the data on a small screen. From the issue:
> 
> > some information and functions are not available without users having to scroll horizontally
> 
> In this solution, we don't have the scroll, but some information is not available at all.

Actually I did not change the behaviour. Data was already removed before. I only adjusted the css so that it looks good on mobile :)

JuliaKirschenheuter (changes requested):
sorry, tested again and it doesn't seems to be fixed ;(

Screenshot from 2023-10-12 10-52-39

ShGKme:
> Actually I did not change the behaviour. Data was already removed before. I only adjusted the css so that it looks good on mobile :)

Oh, sorry, I missed the "Before" section =D

But still, hiding the information seems to be the a11y issue here

szaimen:
> But still, hiding the information seems to be the a11y issue here

Are you sure that was the problem here? I thought only the sticky header?

JuliaKirschenheuter (approved):

ShGKme:
> Are you sure that was the problem here? I thought only the sticky header which is not going to the right?

I'm not...

szaimen:
CI failure unrelated

PR #40885, opened 2023-10-12
Fix OCM provider public API and handling to allow apps to register

nickvergessen:
Follow up to https://github.com/nextcloud/server/pull/39574

## TODO

- [x] Make the public API only rely on OCP
- [x] Make the OCM provider stateful so apps can add resources
- [x] Allow apps to register OCM resources

## Checklist

### DIFF
--- a/lib/private/OCM/Model/OCMProvider.php
+++ b/lib/private/OCM/Model/OCMProvider.php
@@ -26,27 +26,27 @@ declare(strict_types=1);
 
 namespace OC\OCM\Model;
 
-use JsonSerializable;
 use OCP\OCM\Exceptions\OCMArgumentException;
 use OCP\OCM\Exceptions\OCMProviderException;
 use OCP\OCM\IOCMProvider;
+use OCP\OCM\IOCMResource;
 
 /**
  * @since 28.0.0
  */
-class OCMProvider implements IOCMProvider, JsonSerializable {
+class OCMProvider implements IOCMProvider {
 	private bool $enabled = false;
 	private string $apiVersion = '';
 	private string $endPoint = '';
-	/** @var OCMResource[] */
+	/** @var IOCMResource[] */
 	private array $resourceTypes = [];
 
 	/**
 	 * @param bool $enabled
 	 *
-	 * @return OCMProvider
+	 * @return $this
 	 */
-	public function setEnabled(bool $enabled): self {
+	public function setEnabled(bool $enabled): static {
 		$this->enabled = $enabled;
 
 		return $this;
@@ -62,9 +62,9 @@ class OCMProvider implements IOCMProvider, JsonSerializable {
 	/**
 	 * @param string $apiVersion
 	 *
-	 * @return OCMProvider
+	 * @return $this
 	 */
-	public function setApiVersion(string $apiVersion): self {
+	public function setApiVersion(string $apiVersion): static {
 		$this->apiVersion = $apiVersion;
 
 		return $this;
@@ -80,9 +80,9 @@ class OCMProvider implements IOCMProvider, JsonSerializable {
 	/**
 	 * @param string $endPoint
 	 *
-	 * @return OCMProvider
+	 * @return $this
 	 */
-	public function setEndPoint(string $endPoint): self {
+	public function setEndPoint(string $endPoint): static {
 		$this->endPoint = $endPoint;
 
 		return $this;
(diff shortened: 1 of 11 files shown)


ChristophWurst (approved):

lib/private/OCM/Model/OCMProvider.php
```diff
@@ -96,20 +105,28 @@ public function getEndPoint(): string {
...
+	 */
+	public function createNewResource(): IOCMResource {
+		return new OCMResource();
+	}
+
+	/**
+	 * @param IOCMResource $resource
 	 *
-	 * @return $this
+	 * @return self
```
come-nc, line 118: `@return $this` is more precise than `@return self`

lib/public/OCM/Events/BeforeResourceTypesGetEvent.php
```diff
@@ -0,0 +1,34 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\OCM\Events;
+
+use OCP\EventDispatcher\Event;
+
+/**
```
come-nc, line 30: A description of what it’s for would be nice

come-nc (approved):
Fine apart from nitpick

lib/private/OCM/Model/OCMProvider.php
nickvergessen: should then do that also on the interfaces I guess?

nickvergessen: Switched to static and $this as discussed.

lib/public/OCM/Events/BeforeResourceTypesGetEvent.php
```diff
@@ -0,0 +1,37 @@
...
+ * Use this event to register additional OCM resources before the API returns
+ * them in the OCM provider list and capability
+ *
+ * @since 28.0.0
+ */
+class BeforeResourceTypesGetEvent extends Event {
```
ChristophWurst, line 36:
```suggestion
class BeforeResourceTypesReturnedEvent extends Event {
```

or something else that makes the purpose of the event clearer?

nickvergessen: ResourceTypeRegisterEvent?

ChristophWurst: and could we improve the DX by passing the IOCMProvider implementation into the event and exposing a getter?

nickvergessen: Well the main job is not getting but adding,

See https://github.com/nextcloud/spreed/pull/10692/files#diff-3b678f93168d9f4120e050fae0237b8f70b8a03c8a3a3d9c1f81e63f05ffa676R56 for a sample listener

nickvergessen: Ah you mean a getter to get the provider? Can also do that.

ArtificialOwl (approved):

lib/public/OCM/Events/BeforeResourceTypesGetEvent.php
nickvergessen: > why do I create a new resource but then add a resource type? should it be createResourceType or addResource?

That's a good question but the inconsitancy is already part of the public api 🙈

We could add a `registerResourceType()` directly to the event, so consumers don't even need to bother with the `IProvider` and `IOCMResource` themselves? 🤔 sounds good?

ChristophWurst: sounds good!

nickvergessen: > ooooor we register the resource (types) via the application class?

That sounds a bit bloating for the fact we have 2 federation things only (files+ talk soon)

ChristophWurst (approved):

nickvergessen:
/backport 4dbe0677ad25d9ccbc4efcdc67c0e5b3d736d70f to stable27

nickvergessen:
/backport 4dbe0677ad25d9ccbc4efcdc67c0e5b3d736d70f to stable26

solracsf:
Needs manual backport it seems.

nickvergessen:
Since stable branches are currently frozen (in RC stage), will have to wait until after Thursday

PR #40888, opened 2023-10-12
fix: Move contacts menu JSON format to public interface

ChristophWurst:
## Summary

Doesn't really fix anything but the spec should be on the interface IMO.

PHP allows this: https://3v4l.org/JbouL

## Checklist

### DIFF
--- a/lib/private/Contacts/ContactsMenu/Entry.php
+++ b/lib/private/Contacts/ContactsMenu/Entry.php
@@ -141,9 +141,6 @@ class Entry implements IEntry {
 		return $this->properties[$key];
 	}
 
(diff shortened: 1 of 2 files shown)


provokateurin (approved):
I know this can cause problems, but if psalm is green then it's good

ChristophWurst:
which problems?

provokateurin:
Some places only work with the implementation so psalm doesn't understand that it needs to inherit the annotation from the interface. But as I said, if psalm is green the change is good.

ChristophWurst:
Conflicts and has to be redone.

PR #40898, opened 2023-10-13
Add custom apps translation scripts and image path for consistency

### LINKED ISSUE #40897: Custom apps translation and image asset handling is not consistent with core apps
### How to use GitHub

- Custom app translation scripts are not delivered at all
  - For files which is a core app, I see a `GET apps/files/l10n/fr.js`
  - But for a custom app, this is not there
- I can replace core app translations via theme but not for custom apps
- I can replace core app images via theme but not for custom apps

- Consistent behaviour for both core and custom apps

- N/A

**Additional context**
- N/A

akhil1508:
* Resolves: https://github.com/nextcloud/server/issues/40897

## Summary
- Add checks for custom app translations and images

## Checklist

### DIFF
--- a/lib/private/Template/JSResourceLocator.php
+++ b/lib/private/Template/JSResourceLocator.php
@@ -60,8 +60,13 @@ class JSResourceLocator extends ResourceLocator {
 			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.'core/'.$script);
 			$found += $this->appendScriptIfExist($this->serverroot, $script);
 			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.$script);
-			$found += $this->appendScriptIfExist($this->serverroot, 'apps/'.$script);
-			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.'apps/'.$script);
+
+			foreach (\OC::$APPSROOTS as $appRoot) {
+				$dirName = basename($appRoot['path']);
+				$rootPath = dirname($appRoot['path']);
+				$found += $this->appendScriptIfExist($rootPath, $dirName.'/'.$script);
+				$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.$dirName.'/'.$script);
+			}
 
 			if ($found) {
 				return;


lib/private/Template/JSResourceLocator.php
```diff
@@ -62,6 +62,8 @@ public function doFind($script) {
 			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.$script);
 			$found += $this->appendScriptIfExist($this->serverroot, 'apps/'.$script);
 			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.'apps/'.$script);
+			$found += $this->appendScriptIfExist($this->serverroot, 'custom_apps/'.$script);
```
kesselb, line 65: It's possible to pick a different name than custom_apps and therefore this approach is wrong.

kesselb (changes requested):
I'm unsure about this change in general.
I have plenty of apps in apps-extra and see the requests for the translation files.

akhil1508:
> I'm unsure about this change in general.
> I have plenty of apps in apps-extra and see the requests for the translation files.

For me, `occ app:install calendar`  installed it in `custom_apps` folder. I don't see the requests for `custom_apps/calendar/l10n/$lang.js` for any language. Nor if I modify via theme..

lib/private/Template/JSResourceLocator.php
akhil1508: Yes, my bad. Different approach now :)

lib/private/Template/JSResourceLocator.php
```diff
@@ -60,8 +60,13 @@ public function doFind($script) {
 			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.'core/'.$script);
 			$found += $this->appendScriptIfExist($this->serverroot, $script);
 			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.$script);
-			$found += $this->appendScriptIfExist($this->serverroot, 'apps/'.$script);
-			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.'apps/'.$script);
+
+			foreach (\OC::$APPSROOTS AS $appRoot) {
```
kesselb, line 64:
```suggestion
			foreach (\OC::$APPSROOTS as $appRoot) {
```

kesselb:
@juliushaertl @skjnldsv I'm totally unsure if the patch makes sense, we need it, etc .. :see_no_evil: 

However, I dislike the ability to modify files from apps via a theme.

juliushaertl:
I'm all in for dropping the theme generally once we have proper ways to achieve its mainly used functionality  otherwise https://github.com/nextcloud/server/issues/27433

I think for now the change makes sense, to ensure that the legacy theme mechanism works as expected with custom app paths.

juliusknorr (approved):
Would be good to squash the commits into one, otherwise this seems good.

akhil1508:
> Would be good to squash the commits into one

@juliushaertl Squashed

susnux (approved):

juliushaertl:
Failures seem unrelated.

PR #40899, opened 2023-10-13
refactor: Unify running a dev server (for automated testing)

ChristophWurst:
## Summary

Makes it quite easy to spin up a dev web server:

```bash
composer serve
```

It's possible to influence hostname, port and number of worker processes:

```bash
NEXTCLOUD_HOST=127.0.0.1 NEXTCLOUD_PORT=8081 NEXTCLOUD_WORKERS=10 composer serve
```

Ref https://github.com/php-pds/composer-script-names_research for the naming convention.

## TODO

- [x] Add new composer script
- [x] Set PHP_CLI_SERVER_WORKERS and make it overridable to have a bit of concurrency

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,6 +60,10 @@
 		"psalm": "psalm --no-cache --threads=$(nproc)",
 		"psalm:ci": "psalm --no-cache --threads=1",
 		"psalm:update-baseline": "psalm --no-cache --threads=$(nproc) --update-baseline",
+		"serve": [
+			"Composer\\Config::disableProcessTimeout",
+			"PHP_CLI_SERVER_WORKERS=${NEXTCLOUD_WORKERS:=4} php -S ${NEXTCLOUD_HOST:=localhost}:${NEXTCLOUD_PORT:=8080} -t ./"
+		],
 		"test": "phpunit --configuration tests/phpunit-autotest.xml",
 		"test:db": "@composer run test -- --group DB,SLOWDB",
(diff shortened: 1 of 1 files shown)


ChristophWurst:
Needs a config to prevent composer killing the command with "exceeded the timeout of 300 seconds.": https://github.com/spatie/phpunit-watcher/issues/72

skjnldsv (approved):

provokateurin (approved):

composer.json
```diff
@@ -6,6 +6,7 @@
 		"platform": {
 			"php": "8.0"
 		},
+		"process-timeout": 0,
```
Altahrim, line 9: Nit: 
Reading this again, I am a bit annoyed by this part.
It applies to all jobs and I imagine it can be an useful security in some cases.

The only other option I think of is using a `Makefile` instead.

ChristophWurst: Which jobs benefit from a timeout?

CIs do kill jobs if they don't have output for x seconds

provokateurin: I somewhat agree. but I don't think we will really have problems with it

ChristophWurst: Let me try https://stackoverflow.com/questions/18917768/why-composer-install-timeouts-after-300-seconds#comment102528678_49304952

Altahrim (approved):

composer.json
Altahrim: I forgot CI can also abort a long running job :) Your solution is perfect :)

PR #40902, opened 2023-10-13
Append to body all selects inside of user table

JuliaKirschenheuter:
### ☑️ Resolves

* Part of https://github.com/nextcloud/server/issues/36965
* Fixes: NcSelect is broken so far table will be moved to left or right
* After state is a bit better, but don't know how to fix this issue completely. Probably it is ok to leave it like it is now.

### 🖼️ Screenshots

🏚️ Before

Peek 2023-10-11 10-18

 🏡 After

Peek 2023-10-16 11-42

## Checklist


szaimen (changes requested):
> Based on your video, the select dropdown still overlaps?

Yes, but not so much as before.

@Pytal do you have some other ideas?

Pytal:
> > Based on your video, the select dropdown still overlaps?
> 
> Yes, but not so much as before.

Reply in the same comment !?🤪

Pytal:
> @Pytal do you have some other ideas?

Pushed a fix :)

JuliaKirschenheuter:
> Pushed a fix :)

Thank you!

szaimen (commented):
Still looks a bit weird if there are not enough user rows but otherwise it works fine now :)

image

susnux (approved):

apps/settings/src/components/Users/shared/styles.scss
```diff
@@ -38,6 +38,7 @@
 		min-width: var(--cell-width);
 		width: var(--cell-width);
 		color: var(--color-main-text);
+		border-bottom: 1px solid var(--color-border);
```
ShGKme, line 41: This adds an unwanted bottom border to the users count in the end:

image

ShGKme (changes requested):
Works with select great now, but there is an extra border in the table footer now now

Pytal (approved):

ShGKme (approved):

PR #40905, opened 2023-10-13
Reduce left margin of files list header

JuliaKirschenheuter:
* Part of: https://github.com/nextcloud/server/issues/36965
* Resolves: files overview part

## Summary

🏚️ Before | 🏡 After
---|---
 image | Screenshot from 2023-10-13 15-57-49

## Checklist


apps/files/src/views/FilesList.vue
```diff
@@ -441,7 +441,7 @@ export default Vue.extend({
 }
 
 $margin: 4px;
-$navigationToggleSize: 50px;
+$navigationToggleSize: 45px;
```
szaimen, line 444: isn't our baseline that usually 44px? maybe we should use the normal variable here? WDYT @skjnldsv

skjnldsv (changes requested):
Okay, I had to dig into this so we can move forward
- [x] 1: Fix the min-width on the nav in the vue components (like mentioned [here](https://github.com/nextcloud-libraries/nextcloud-vue/pull/4640#pullrequestreview-1674090462))
- [x] 2: Add a `	max-width: 100%;` on the `.files-list__header`

Then it will be fixed.
Peek 13-10-2023 16-47

skjnldsv (approved):

skjnldsv:
The mixin is broken, fixed in https://github.com/nextcloud/server/commit/11728d69d9701e783fa2799d15de7ec0577dc36d

Pytal (approved):

AndyScherzinger (approved):

PR #40924, opened 2023-10-15
Shortcut Ctrl+F for different keyboard layouts

AIlkiv:
## Summary

For example, on the Ukrainian keyboard layout, when you press the 'f' button, event.key is equal to "а", and event.code is equal to "KeyF".
Documentation: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code

If you activate the Caps Lock and then press CTRL+f on an English keyboard layout, event.key becomes "F" instead of 'f', causing the search window not to open.

This pull request solves both of these problems.

image

## Checklist


susnux:
> For example, on the Ukrainian keyboard layout, when you press the 'f' button, event.key is equal to "а", and event.code is equal to "KeyF".

Should it not be the opposite? `event.key` should always be the letter that is "printed" and `event.code` should be the physical code. Meaning if you press `f` you get `event.key = f` but `event.code` depends on your keyboard layout.
E.g. if I press `z` I get `event.key = z` but `event.code = keyY` because of QWERTZ vs QWERTY.

So my problem is: If for Ukrainian keyboards search shortcut is "ctrl + a" (and a is on the place were on qwerty is the f) then this will work, but I am not sure this is valid for all languages.

susnux (approved):
Generally ok with this, but see my comment.

AIlkiv:
> Should it not be the opposite? `event.key` should always be the letter that is "printed" and `event.code` should be the physical code. Meaning if you press `f` you get `event.key = f` but `event.code` depends on your keyboard layout.
> E.g. if I press `z` I get `event.key = z` but `event.code = keyY` because of QWERTZ vs QWERTY.

@susnux Thanks for the answer. Many languages lack the letter "F," but there is an overlay of QWERTY + the native language. We use the physical placement of keys (tested: Chrome, LibreOffice, Office365).

Here are examples of several keyboard layouts:

Keyboards with Ukrainian layout

image

Keyboards with Greek layout

image

I've corrected the use of keyCode; it is more accurate for different layouts. It works correctly even on the Dvorak keyboard layout.

susnux:
Thank you for the explanation! Makes sense now.
But please revert to use `.code` as `.keyCode` is deprecated

AIlkiv:
> Thank you for the explanation! Makes sense now. But please revert to use `.code` as `.keyCode` is deprecated

Reverted

Pytal (approved):

PR #40927, opened 2023-10-15
Allow share expiry dates lower than enforced limits

nfebe:
Previously, users could change the share expiry date up to the enforced maximum.

The new share flow imposed the enforced share expiry date maximum literally and did not allow even dates lower than the maximum enforced.

That does not make much sense, if the enforced expiry date is 30 days from creation date, then it's logical to allow users set the date to anything less than 30 days from the creation date.

Resolves : _No github issue_


susnux (approved):

sorbaugh (approved):

come-nc:
I tested and I confirm https://github.com/nextcloud/server/pull/40928#pullrequestreview-1679980436 , this change breaks default date handling for remote and public shares, the default date is use there as a maximum even if enforced checkbox is not checked for those types of shares.
For remote shares it also seem to not even be used as default value.

Fenn-CS:
@come-nc nice catch : https://github.com/nextcloud/server/pull/40933

PR #40932, opened 2023-10-16
fix profile visibility settings being too wide

szaimen:
Address part of https://github.com/nextcloud/server/issues/36965

| Before | After mobile | After Desktop |
|---|---|---|
| image | image image | image |

<details>

```
docker run -it \
--name nextcloud-easy-test \
-p 8443:443 \
-e TRUSTED_DOMAIN=192.168.24.128 \
-e SERVER_BRANCH=enh/36965/fix-account-settings \
-e COMPILE_SERVER=1 \
-e VIEWER_BRANCH=master \
```

</details>


AndyScherzinger (approved):

susnux (approved):

apps/settings/css/settings.scss
```diff
@@ -120,6 +120,12 @@ select {
 	}
 }
 
+@media only screen and (max-width: math.div(variables.$breakpoint-mobile, 2)) {
```
ShGKme, line 123: I'm not sure it is fine to have a calculation over breakpoint-mobile here.

There is no semantics in `breakpoint-mobile / 2` value (it is not another breakpoint) and it is not exactly 0.5.

apps/settings/css/settings.scss
```diff
@@ -120,6 +120,12 @@ select {
...
+	#personal-settings {
+		display: block;
+	}
+}
```
ShGKme, line 127: And I don't see what actually `display: block` does here. It has a grid with 1 column anyway. What does it change?

apps/settings/css/settings.scss
szaimen: We do this already in multiple places so I thought it would be okay to do it here as well

ShGKme: I haven't found any on the server, except some recent updates this issue related. But here it also seems to change nothing, or I'm looking somewhere wrong.

ShGKme: > did you look for math.div?

No, for `breakpoint-mobile`

szaimen: ah, yes we already have at least one example in nextcloud/vue

apps/settings/css/settings.scss
szaimen: | With display: block | With display: grid |
|---|---|
| image | imageb|

JuliaKirschenheuter (approved):
Looks good, thank you!

apps/settings/css/settings.scss
ShGKme: Oh, I see know.

It is caused by min grid column size: 
`grid-template-columns: repeat(auto-fill, minmax(300px, 1fr));`

What about adding min with `100%` to not overlap when it is less than 300px?

`repeat(auto-fill, minmax(min(100%, 300px), 1fr))`

ShGKme (approved):

szaimen:
CI failure unrelated

PR #40935, opened 2023-10-16
add some support for rename on case insensitive local filesystems

icewind1991:
Add *some* logic for handling case insensitive filesystems when `localstorage.case_insensitive` is set.

This adds a bunch of overhead as `file_exists` now needs to read the directory content to perform the case sensitive compare.

With my local testing (case insensitive zfs), the `rename('foo', 'Foo')` call returns true but doesn't actually change the on-disk name, so an additional check is added that the name has actually changed. So the rename fails but at least doesn't cause a filecache desync.

### DIFF
--- a/lib/private/Files/Storage/Local.php
+++ b/lib/private/Files/Storage/Local.php
@@ -74,6 +74,8 @@ class Local extends \OC\Files\Storage\Common {
 
 	protected bool $unlinkOnTruncate;
 
+	protected bool $caseInsensitive = false;
+
 	public function __construct($arguments) {
 		if (!isset($arguments['datadir']) || !is_string($arguments['datadir'])) {
 			throw new \InvalidArgumentException('No data directory set for local storage');
@@ -93,6 +95,7 @@ class Local extends \OC\Files\Storage\Common {
 		$this->config = \OC::$server->get(IConfig::class);
 		$this->mimeTypeDetector = \OC::$server->get(IMimeTypeDetector::class);
 		$this->defUMask = $this->config->getSystemValue('localstorage.umask', 0022);
+		$this->caseInsensitive = $this->config->getSystemValueBool('localstorage.case_insensitive', false);
 
 		// support Write-Once-Read-Many file systems
 		$this->unlinkOnTruncate = $this->config->getSystemValueBool('localstorage.unlink_on_truncate', false);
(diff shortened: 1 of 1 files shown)


icewind1991:
Note that having proper support for case insensitive file systems is a significant amount of work

Altahrim:
Tested on `stable27`  :+1:

lib/private/Files/Storage/Local.php
```diff
@@ -372,6 +387,11 @@ public function rename($source, $target): bool {
 		}
 
 		if (@rename($this->getSourcePath($source), $this->getSourcePath($target))) {
+			if ($this->caseInsensitive) {
+				if (strtolower($target) === strtolower($source) && !$this->file_exists($target)) {
```
Altahrim, line 391: I think we should care to multibyte strings:
```suggestion
				if (mb_strtolower($target) === mb_strtolower($source) && !$this->file_exists($target)) {
```
I renamed a file `É` to `é` and I had the issue.

lib/private/Files/Storage/Local.php
```diff
@@ -74,6 +74,8 @@ class Local extends \OC\Files\Storage\Common {
 
 	protected bool $unlinkOnTruncate;
 
+	private $caseInsensitive = false;
```
juliushaertl, line 77:
```suggestion
	private bool $caseInsensitive = false;
```

lib/private/Files/Storage/Local.php
```diff
@@ -93,6 +95,7 @@ public function __construct($arguments) {
 		$this->config = \OC::$server->get(IConfig::class);
 		$this->mimeTypeDetector = \OC::$server->get(IMimeTypeDetector::class);
 		$this->defUMask = $this->config->getSystemValue('localstorage.umask', 0022);
+		$this->caseInsensitive = $this->config->getSystemValue('localstorage.caseInsensitive', false);
```
kesselb, line 98: We are already using

localstorage.umask
localstorage.unlink_on_truncate
localstorage.allowsymlinks

and therefore I would prefer localstorage.case_insensitive to keep it a bit consistent ;)

lib/private/Files/Storage/Local.php
```diff
@@ -74,6 +74,8 @@ class Local extends \OC\Files\Storage\Common {
 
 	protected bool $unlinkOnTruncate;
 
+	private bool $caseInsensitive = false;
```
Altahrim, line 77: May I suggest to add it as `protected` in `Common` so we'll be able to use it for other storages too

Altahrim (approved):

juliusknorr (approved):

PR #37065, opened 2023-03-07
Add accessible system tags select

Pytal:
- For https://github.com/nextcloud/server/issues/36718

## Summary

Before | After
--- | ---
image | image
image | image

<details>
<summary>Notes</summary>

FIXMEs are out-of-scope

The `parseIdFromLocation()` util was adapted from https://github.com/nextcloud/server/blob/73000175d097a5d04cfa9abcfc964fd641ba2021/core/src/OC/backbone-webdav.js#L86-L102

</details>

## TODO

- [x] Fetch last used tags and sort first
- [x] Sort selected and created tags to front
- [x] Use [`webdav`](https://github.com/perry-mitchell/webdav-client) lib for DAV requests
- [x] Extract methods from ApiMixin and export from separate non-mixin file
- [x] Fix eslint
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3968
- [x] Remove legacy systemtags code
- [x] https://github.com/nextcloud/server/pull/37809
- [x] https://github.com/nextcloud/nextcloud-vue/pull/3981
- [x] https://github.com/nextcloud/server/pull/37804
- [x] Compile

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #40939, opened 2023-10-17
sidebar: allow turning off tags view by default

pulsejet:
With #37065, there is no way to hide the tags from the sidebar by default when they are not relevant or redundant (e.g. the tab may already show the file's tags). This can be annyoing especially when the file has many tags. This patch adds an option to hide the tags from the sidebar by default (the user can still open the tags tab manually).

This also reduces one request when opening the sidebar when the tags are turned off, since all tags don't need to be fetched anymore.

Screenshot: tags are redundant and take up a lot of space.

image


artonge (approved):
This won't be persisted, right ?

pulsejet:
> This won't be persisted, right ?

Yeah, a reload will switch it back to showing the tags again (current behavior). But I expect the app to call this method to not show the tags on load.

Pytal:
> With #37065, there is no way to hide the tags from the sidebar by default when they are not relevant or redundant (e.g. the tab may already show the file's tags).

Worked this way actually since legacy times :)

Pytal (approved):

pulsejet:
> Worked this way actually since legacy times :)

Well there was `OCA.SystemTags.View.toggle()`. Technically you could combine that with a `MutationObserver` ...

Pytal:
> Well there was `OCA.SystemTags.View.toggle()`. Technically you could combine that with a `MutationObserver` ...

🤯, I suppose default was the same then

pulsejet:
Any reason this was ignored for 28?

artonge:
Beside red CI, I see no reason :)

artonge:
> is it okay if I enable auto-merge on my PRs?

Sure, but some CI steps are not required to auto-merge, so the merge won't get blocked if they don't pass. If you're confident that all tests will pass then go for it :).

> EDIT2: another issue is the quick merge conflict that arises

That's hard to do without. But if you have the needed approvals, and if CI failures are not related, ping one of us stating the situation, and we'll force merge for you.

artonge:
You can also try to restart some failing CI steps. Note that for cypress, you need to restart all the steps and not only the failing ones.

solracsf:
/backport https://github.com/nextcloud/server/commit/424eb124ef2c7b8360d1ed39f006e82f18d9d0f5 to stable28

solracsf:
/backport 424eb124ef2c7b8360d1ed39f006e82f18d9d0f5 to stable28

PR #40955, opened 2023-10-17
perf: switch file list to vue-virtual-scroller

pulsejet:
This patch significantly improves the performance of the file list by using `vue-virtual-scroller` over the current simplistic implementation. `vue-virtual-scroller` has excellent performance since it uses CSS transforms to place the virtual elements and has very good component reuse.

The implementation isn't complete yet but I wanted to get some early feedback / thoughts.

It's somewhat hard to capture the improvement in a video, so here's a before and after with a 4x simulated slowdown:

Before:
https://github.com/nextcloud/server/assets/10709794/0d769798-3d85-44ea-a2b8-cd3fba0aeb53

After:
https://github.com/nextcloud/server/assets/10709794/1d2a2a83-0180-4a23-833b-12d6f69099fa

### DIFF
--- a/apps/files/src/components/FileEntry.vue
+++ b/apps/files/src/components/FileEntry.vue
@@ -50,9 +50,9 @@
 			<!-- Icon or preview -->
 			<span class="files-list__row-icon" @click="execDefaultAction">
 				<template v-if="source.type === 'folder'">
-					<FolderOpenIcon v-if="dragover" />
+					<FolderOpenIcon v-once v-if="dragover" />
 					<template v-else>
-						<FolderIcon />
+						<FolderIcon v-once />
 						<OverlayIcon :is="folderOverlay"
 							v-if="folderOverlay"
 							class="files-list__row-icon-overlay" />
@@ -69,13 +69,13 @@
 					@error="backgroundFailed = true"
 					@load="backgroundFailed = false">
 
-				<FileIcon v-else />
+				<FileIcon v-once v-else />
 
 				<!-- Favorite icon -->
 				<span v-if="isFavorite"
 					class="files-list__row-icon-favorite"
 					:aria-label="t('files', 'Favorite')">
-					<FavoriteIcon />
+					<FavoriteIcon v-once />
 				</span>
 			</span>
 
--- a/apps/files/src/components/FilesListTableFooter.vue
+++ b/apps/files/src/components/FilesListTableFooter.vue
@@ -159,7 +159,6 @@ export default Vue.extend({
 <style scoped lang="scss">
 // Scoped row
 tr {
-	padding-bottom: 300px;
 	border-top: 1px solid var(--color-border);
 	// Prevent hover effect on the whole row
 	background-color: transparent !important;
--- a/apps/files/src/components/FilesListVirtual.vue
+++ b/apps/files/src/components/FilesListVirtual.vue
@@ -278,20 +278,9 @@ export default Vue.extend({
 	--clickable-area: 44px;
 	--icon-preview-size: 32px;
 
-	display: block;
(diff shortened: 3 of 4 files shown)


skjnldsv:
We actually moved away from it purposely, sorry.

apps/files/src/components/FileEntry.vue
```diff
@@ -50,9 +50,9 @@
 			<!-- Icon or preview -->
 			<span class="files-list__row-icon" @click="execDefaultAction">
 				<template v-if="source.type === 'folder'">
-					<FolderOpenIcon v-if="dragover" />
+					<FolderOpenIcon v-once v-if="dragover" />
```
skjnldsv, line 53: `v-once` is still a good idea! :+1:

pulsejet:
Ah I see, thanks for the explanation.

I agree `vue-virtual-scroller` has an accessibility and maintenance problem, but I do belive using transforms gives better performance (I did a side-by-side comparison of using of the various ways to do virtual scrolling a while back and this approach has fewer reflows than anything else). Anyway, I agree with the reasoning of staying with an in-house component 👍🏻, but I think there are some lessons to be learned from `vue-virtual-scroller`.

> You're missing the latest master, which includes the recycle part of the vue-virtual-scroller which gives all the performance boost you're seeing

Indeed it does seem better than before, but still not as fluid. Will try to profile this a bit to understand what's happening.

BTW I might throw out a couple of PRs for some smaller optimizations; feel free to knock these down if they make a lot of sense.

skjnldsv:
> but I think there are some lessons to be learned from `vue-virtual-scroller`.

:100: 

> BTW I might throw out a couple of PRs for some smaller optimizations

:hugs: :bow:

pulsejet:
> I do belive using transforms gives better performance

So I did a deeper dive into this and thought I'd document my findings here.

The reason using absolute positioning is faster is subtle. The current virtual list code does maintain the list of keys to be the same. Whenever the user scrolls down one element, the array `renderedItems` is reordered, so that the first key (which is now unused) moves to the last. As a result the components in the virtual DOM are reused.

Here's the problem: since the array was reordered, Vue needs to actually remove the first physical node and append it at the end and set up all hooks again. So even when the virtual DOM is untouched, the actual DOM is manipulated as the user scrolls down, which is quite slow.

On the other hand in the case of using absolute positioning, even the actual DOM is untouched and only the CSS changes. Even better when using CSS transforms, since it all happens on the GPU.

As a POC, I tried [switching the list to use transforms](https://github.com/nextcloud/server/commit/a27d9e241fc3da54c54f3748abeb0a96b4f59bd4) and I can confirm that it gives the 60fps butter smooth scrolling effect. But the problem is, still, there's no easy way I've found to make this accessible. It _might_ be possible to make the tab order work correctly using some JS, not sure how this'll fare with screenreaders though.

Let me know if any thoughts @skjnldsv

pulsejet:
To be clear: I'm not suggesting we do this / it's a good idea. Just satisfying my academic curiosity.

skjnldsv:
> As a POC, I tried [switching the list to use transforms](https://github.com/nextcloud/server/commit/a27d9e241fc3da54c54f3748abeb0a96b4f59bd4) and I can confirm that it gives the 60fps butter smooth scrolling effect. But the problem is, still, there's no easy way I've found to make this accessible. It _might_ be possible to make the tab order work correctly using some JS, not sure how this'll fare with screenreaders though.

Considering the points above and the insane amount of accessibility issues we had to fix over the last months, I'm not taking any risks, especially for a subtle perf change :)

But thanks for the research, this makes sense and is indeed quite insightful! :wink:

PR #40958, opened 2023-10-18
files-list: performance optimizations

pulsejet:
Some optimizations in the virtual scroller @skjnldsv 

There are comments on the individual changes. Again, it's hard to measure performance here.


apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -70,6 +70,8 @@ import NcActions from '@nextcloud/vue/dist/Components/NcActions.js'
 import NcIconSvgWrapper from '@nextcloud/vue/dist/Components/NcIconSvgWrapper.js'
 import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 
+import CustomElementRender from '../CustomElementRender.vue'
```
pulsejet, line 73: missing import

apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -182,18 +185,18 @@ export default Vue.extend({
...
 
-	methods: {
 		/**
 		 * Making this a function in case the files-list
 		 * reference changes in the future. That way we're
 		 * sure there is one at the time we call it.
 		 */
-		getBoundariesElement() {
+		 getBoundariesElement() {
 			return document.querySelector('.app-content > table.files-list')
```
pulsejet, line 195: I think the return value for this can be cached since it doesn't change in the lifetime of this component? Correct me if I'm wrong

apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -41,13 +41,13 @@
...
 
-		<FileIcon v-else />
+		<FileIcon v-once v-else />
 
 		<!-- Favorite icon -->
 		<span v-if="isFavorite"
 			class="files-list__row-icon-favorite"
 			:aria-label="t('files', 'Favorite')">
-			<FavoriteIcon />
+			<FavoriteIcon v-once />
```
pulsejet, line 50: Icons don't need re-rendering

apps/files/src/components/FilesListVirtual.vue
```diff
@@ -312,6 +312,8 @@ export default Vue.extend({
 	&::v-deep {
 		// Table head, body and footer
 		tbody {
+			will-change: scroll-position, padding;
+			contain: layout paint style;
```
pulsejet, line 316: Some CSS tricks to improve rendering performance

apps/files/src/components/VirtualList.vue
```diff
@@ -18,7 +18,7 @@
 			<component :is="dataComponent"
 				v-for="({key, item}, i) in renderedItems"
 				:key="key"
-				:visible="(i >= bufferItems - 1 || index <= bufferItems) && (i <= shownItems - bufferItems)"
+				:visible="true"
```
pulsejet, line 21: Toggling `visible` actually hurts performance (a bit) since it causes reflows and prevents proper component reuse.

apps/files/src/components/VirtualList.vue
```diff
@@ -211,7 +211,7 @@ export default Vue.extend({
 		}
 
 		// Adding scroll listener AFTER the initial scroll to index
-		this.$el.addEventListener('scroll', this.onScroll)
+		this.$el.addEventListener('scroll', this.onScroll, { passive: true })
```
pulsejet, line 214: Passive listeners are faster

apps/files/src/components/VirtualList.vue
```diff
@@ -232,11 +232,14 @@ export default Vue.extend({
 		},
 
 		onScroll() {
-			const topScroll = this.$el.scrollTop - this.beforeHeight
-			const index = Math.floor(topScroll / this.itemHeight) * this.columnCount
-			// Max 0 to prevent negative index
-			this.index = Math.max(0, index)
-			this.$emit('scroll')
+			this._onScrollHandle ??= requestAnimationFrame(() => {
```
pulsejet, line 235: Throttling this call to once per frame. Notably, `this.$el.scrollTop` is a getter and causes layout thrashing, so this is the most important change here.

apps/files/src/components/FileEntry/FileEntryActions.vue
skjnldsv: It totally can!
The table isn't supposed to be unmounted anyway :)

apps/files/src/components/VirtualList.vue
skjnldsv: Alright, I guess we don't show that many extra rows.
Maybe we can entirely remove the visible feature then?

apps/files/src/components/VirtualList.vue
skjnldsv: That's something I wanted to experiment with! Nice take at this issue! It make sense :)

apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -182,18 +185,18 @@ export default Vue.extend({
...
-	},
 
-	methods: {
 		/**
 		 * Making this a function in case the files-list
 		 * reference changes in the future. That way we're
 		 * sure there is one at the time we call it.
 		 */
-		getBoundariesElement() {
+		 getBoundariesElement() {
```
skjnldsv, line 194:
```suggestion
		getBoundariesElement() {
```

apps/files/src/components/FilesListVirtual.vue
```diff
@@ -312,6 +312,8 @@ export default Vue.extend({
 	&::v-deep {
 		// Table head, body and footer
 		tbody {
+			will-change: scroll-position, padding;
```
skjnldsv, line 315: Not sure about padding. Is it even a valid value for `will-change` ? :thinking:

apps/files/src/components/VirtualList.vue
skjnldsv: Ah, and this will cause issues with the drag to scroll feature:
https://github.com/nextcloud/server/blob/8c2473783109653b9ae778ad5634b3cee554a8d0/apps/files/src/components/FilesListVirtual.vue#L263-L275

skjnldsv:
Pushed some fixes, performances are definitely positively impacted!! :rocket:

apps/files/src/components/VirtualList.vue
skjnldsv: Done

skjnldsv (approved):

apps/files/src/components/FilesListVirtual.vue
pulsejet: Well `will-change` takes [<custom-ident>](https://developer.mozilla.org/en-US/docs/Web/CSS/custom-ident) as a valid value so it's all up to the browser. I don't think there are any real optimizations for padding changes right now though 🤷🏻

pulsejet:
Thanks for fixing this up @skjnldsv 😄

skjnldsv:
Drone failure unrelated: https://drone.nextcloud.com/nextcloud/server/42176

Altahrim (approved):

juliushaertl:
This seems to have caused a regression where file list headers are broken now and not rendering anymore. Namely recommendations and text app.

PR #40963, opened 2023-10-18
Reduce width of sharing entry

JuliaKirschenheuter:
* Part of https://github.com/nextcloud/server/issues/36965
* Fixes "Files right sidebar"

## Summary

| image | Screenshot from 2023-10-18 16-30-24

## Checklist


Pytal (approved):

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -756,7 +756,7 @@ export default {
 		display: flex;
 		justify-content: space-between;
 		width: 80%;
-		min-width: 80%;
+		min-width: 70%;
```
ShGKme, line 759: `min-width: 70%` fixes the issue on a small screen.

But `80%` is always a bit not the same as other elements.

What about replacing the fixed 80/70% width with flex-grow? Then it will always have just all the space between buttons on all screens.

(with `min-width: 0;` to with overlapping issue)

Large screen | Small screen
---|---
**Current master** `width: 80%; min-width: 80%` | .
image | image
**This PR** `width: 80%; min-width: 70%` | .
image | image
**Proposal** `flex: 1 0; min-width: 0;` | .
image | image

JuliaKirschenheuter: Thanks a lot! Done

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -755,8 +755,9 @@ export default {
 		padding-left: 10px;
 		display: flex;
 		justify-content: space-between;
+		flex: 1 0;
 		width: 80%;
-		min-width: 80%;
+		min-width: 0;
```
ShGKme, line 760: With `flex: 1 0` there should not be fixed `width: 80%;`. Otherwise, the base size is still 80% and it is broken on a small screen.

ShGKme (changes requested):

apps/files_sharing/src/components/SharingEntryLink.vue
ShGKme: `width: 80%;` is the main problem, it makes a shift on both small and large screens, because `80% != 100% - buttons and avatars`

ShGKme (commented):
There is also a second component that has exactly the same styles with the same issue in this place: `SharingEntry.vue`

image

Let's fix it as well.

ShGKme (approved):

JuliaKirschenheuter:
failed continuous-integration/drone/pr tests are not related

PR #29357, opened 2021-10-21
Prevent duplicate auth token activity updates

ChristophWurst:
The auth token activity logic works as follows
* Read auth token
* Compare last activity time stamp to current time
* Update auth token activity if it's older than x seconds

This works fine in isolation but with concurrency that means that
occasionally the same token is read simultaneously by two processes and
both of these processes will trigger an update of the same row.
Affectively the second update doesn't add much value. It might set the
time stamp to the exact same time stamp or one a few seconds later. But
the last activity is no precise science, we don't need this accuracy.

This patch changes the UPDATE query to include the expected value in a
comparison with the current data. This results in an affected row when
the data in the DB still has an old time stamp, but won't affect a row
if the time stamp is (nearly) up to date.

This is a micro optimization and will possibly not show any significant
performance improvement. Yet in setups with a DB cluster it means that
the write node has to send fewer changes to the read nodes due to the
lower number of actual changes.

## How to test

* Assign the result of ``$update->executeStatement();`` to a var. It contains the number of affected rows.
* Set a break point at the same line
* Set the `last_activity` to 0 for the auth token that you want to test, e.g. the browser session token
* Trigger some (ajax) requests and wait until more than one processes hit the break point
* Take one step in each debugger session

You should see that the number of affected rows is 1 for the first process but 0 for the others.

PR #40971, opened 2023-10-19
fix: Make authtoken activity updater simpler to avoid deadlocks

ChristophWurst:
* Resolves: # 

## Summary

https://github.com/nextcloud/server/pull/29357 made the query more complex to avoid too many UPDATE queries, but this can lead to deadlock situations. This patch reverts the optimization. This means there might be some lost updates in concurrent situations and there will be a lot more actual UPDATEs.

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Authentication/Token/PublicKeyTokenMapper.php
+++ b/lib/private/Authentication/Token/PublicKeyTokenMapper.php
@@ -203,28 +203,6 @@ class PublicKeyTokenMapper extends QBMapper {
 	/**
 	 * Update the last activity timestamp
 	 *
-	 * In highly concurrent setups it can happen that two parallel processes
-	 * trigger the update at (nearly) the same time. In that special case it's
-	 * not necessary to hit the database with two actual updates. Therefore the
-	 * target last activity is included in the WHERE clause with a few seconds
-	 * of tolerance.
-	 *
-	 * Example:
-	 * - process 1 (P1) reads the token at timestamp 1500
-	 * - process 1 (P2) reads the token at timestamp 1501
-	 * - activity update interval is 100
-	 *
-	 * This means
-	 *
-	 * - P1 will see a last_activity smaller than the current time and update
-	 *   the token row
-	 * - If P2 reads after P1 had written, it will see 1600 as last activity
-	 *   and the comparison on last_activity won't be truthy. This means no rows
-	 *   need to be updated a second time
-	 * - If P2 reads before P1 had written, it will see 1501 as last activity,
-	 *   but the comparison on last_activity will still not be truthy and the
-	 *   token row is not updated a second time
-	 *
 	 * @param IToken $token
 	 * @param int $now
 	 */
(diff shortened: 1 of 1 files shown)


lib/private/Authentication/Token/PublicKeyTokenMapper.php
```diff
@@ -234,7 +212,6 @@ public function updateActivity(IToken $token, int $now): void {
 			->set('last_activity', $qb->createNamedParameter($now, IQueryBuilder::PARAM_INT))
 			->where(
 				$qb->expr()->eq('id', $qb->createNamedParameter($token->getId(), IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT),
-				$qb->expr()->lt('last_activity', $qb->createNamedParameter($now - 15, IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT)
```
Altahrim, line 237: Do we have a way to know current `last_activity` of the token?
If so, we could write this:
```php
$qb->expr()->eq('last_activity', $qb->createNamedParameter($token->lastActivity(), IQueryBuilder::PARAM_INT), IQueryBuilder::PARAM_INT)
```
I think it can help to reduce locking

nickvergessen: https://github.com/nextcloud/server/commit/7dd7256cfe8bdb2e11def60167e1b418458e68ea

> This is a micro optimization and will possibly not show any significant
performance improvement. Yet in setups with a DB cluster it means that
the write node has to send fewer changes to the read nodes due to the
lower number of actual changes.

This is exactly what I'm wondering about. Not sure if DB clusters are optimized enough to not resync changes when the query is "empty", but the idea was to avoid contacting the master.

ChristophWurst: @Altahrim I thought about that too. We definitely have access to the previous last_activity value. But then I wonder if the additional clause in WHERE makes the query less or more prone to locking issues. The row is already looked up by the primary key. I can't imagine there being much difference between a less than and an equals check. They should be comparably fast/slow.

Altahrim: > Not sure if DB clusters are optimized enough to not resync changes when the query is "empty"

If I recall correctly, Galera enforce row replication, so no change should lead to no replication sent.

> The row is already looked up by the primary key.

I am agree. I would like to be sure it doesn't lock anything else because of the `less than`.
I generally saw equal instead of less than to check if row as been updated.

----
Maybe we can simply avoid this request PHP side if we detect a close enough timestamp?

Altahrim:
Did you manage to reproduce the issue somewhere?
Unless working on several rows at the same time, I wasn't able to.

ChristophWurst:
No, I was not. The deadlocks are only showing on a production instance backed by a Galera cluster.

ChristophWurst:
The change doesn't show any improvements on a affected production instance.

PR #40978, opened 2023-10-19
HTML Validation: replace non standard `autocorrect` with `spellcheck`

ShGKme:
* Resolves: a part of https://github.com/nextcloud/andy/issues/5

## Summary

`autocorrect` attribute is not a standard and works only in Safari.

Replaced with HTML 5 `spellcheck`. Works in Safari since Safari supports HTML 5.

No visual changes.

See:
- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input#autocorrect
- https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck

## Checklist


szaimen (approved):
LGTM but didnt test

Pytal (approved):

susnux (approved):

PR #40986, opened 2023-10-19
fix(tests): Only run tests on CI that are needed

susnux:
## Summary

Do not run PHP unit tests when only JS code is changed, should save us some CI time.

### TODO

- [ ] Make `node-test-summary` required instead  of `jsunit`, `handlebars` and `test` (node test)

## Checklist


.github/workflows/lint-php-cs-when-unrelated.yml
```diff
@@ -0,0 +1,35 @@
+# This workflow is provided via the organization template repository
```
nickvergessen, line 1: This does not work as I found out the hard way. Please see github.com/nextcloud/.github for updated logic

nickvergessen (changes requested):

.github/workflows/lint-php-cs-when-unrelated.yml
susnux: Good to know! Move to the new logic :)

pulsejet:
Don't know enough about this part

skjnldsv (approved):

.github/workflows/lint-eslint.yml
```diff
@@ -31,14 +19,42 @@ concurrency:
...
+
+    steps:
+      - uses: dorny/paths-filter@4512585405083f25c027a35db413c2b3b9006d50 # v2.11.1
+        id: changes
+        continue-on-error: true
+        with:
+          filters: |
+            src:
+              - '.github/workflows/**'
+              - 'src/**'
```
nickvergessen, line 36: same question for this, does it match code `core/src/` and `apps/**/src/` or only root src?

.github/workflows/lint-eslint.yml
```diff
@@ -31,14 +19,42 @@ concurrency:
...
+              - 'appinfo/info.xml'
```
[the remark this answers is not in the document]
susnux: I wonder if we can remove that line, why would be need to lint the JS code if the appinfo is changed?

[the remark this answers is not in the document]
nickvergessen: Because that is on releases/branch off which is a sane snapshot. But yeah less useful in server than in apps

.github/workflows/lint-eslint.yml
susnux: Fixed!

susnux:
I think this is good now :)
Drone is unrelated :wink:

nickvergessen (approved):

PR #40994, opened 2023-10-19
fix(user_menu): pass only actual prop to user menu entries

ShGKme:
* Resolves: a part of https://github.com/nextcloud/andy/issues/5

## Summary

- **Error: Attribute `order` not allowed on element [`li`](https://html.spec.whatwg.org/multipage/#the-li-element) at this point.**
- **Error: Attribute `classes` not allowed on element [`li`](https://html.spec.whatwg.org/multipage/#the-li-element) at this point.**
- **Error: Attribute `unread` not allowed on element [`li`](https://html.spec.whatwg.org/multipage/#the-li-element) at this point.**
- **Error: The `type` attribute on the `li` element is obsolete.**

All these attributes were set because Navigation Entries was set using `v-bind="entry"`.

User Menu has only a part of the entry properties as props, e.g. `name` or `href`. But others are not props.

## TODO

- [x] Do not bind a whole object with `v-bind="entry"` but instead bind only necessary props one by one

### Before

image

### After

image

## Checklist


Pytal (approved):

susnux (approved):

JuliaKirschenheuter (approved):

ShGKme:
Drone error unrelated

PR #40995, opened 2023-10-19
optimize cache jail creation

icewind1991:
- remove duplicate constructor assignments
- allow the cache jail to re-use the inner cache

A [modest improvement](https://blackfire.io/profiles/compare/da5ade2a-0f2a-4b00-8992-0f21e0658ad6/graph) for little effort (tested with listing 2k group folders)

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Jail.php
+++ b/lib/private/Files/Storage/Wrapper/Jail.php
@@ -396,10 +396,7 @@ class Jail extends Wrapper {
 	 * @return \OC\Files\Cache\Cache
 	 */
 	public function getCache($path = '', $storage = null) {
-		if (!$storage) {
-			$storage = $this->getWrapperStorage();
-		}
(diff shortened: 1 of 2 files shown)


lib/private/Files/Storage/Wrapper/Jail.php
```diff
@@ -396,10 +396,7 @@ public function hasUpdated($path, $time) {
 	 * @return \OC\Files\Cache\Cache
 	 */
 	public function getCache($path = '', $storage = null) {
```
Altahrim, line 398: Now useless?
```suggestion
	public function getCache($path = '') {
```

icewind1991: The parameter is part of the parent class' method

Altahrim (approved):

blizzz (approved):

PR #40998, opened 2023-10-19
chore(tests): Move OCS integration tests from drone to cypress

susnux:
## Summary

Drone takes ~3 minutes (CPU time) for preparing the tests and then ~20 seconds for compiling and running.
Cypress only takes ~10 seconds for compiling and running the tests.

Also added similar tests for v2 and extended them a bit.

BTW: The new cypress command probably can be reused afterwards for the other commands or even be moved to the nextcloud-cypress library(?).

## Checklist


nickvergessen:
Talk is running integration tests just fine with behat on GitHub actions: https://github.com/nextcloud/spreed/blob/master/.github/workflows/integration-pgsql.yml
I would heavily prefer to keep the php/API tests on php stack so the devs that work on the apis don't need to learn another stack for their CI. You can add cypress for frontend/end2end tests, that's fine and cool, but actual API testing should stay in php 🤔

susnux:
Sounds reasonable @nickvergessen I just thought nobody wants to keep that test files as you have to maintain the "tests" and than in some other location the logic that gets called by the tests.

But as I am not mainly a backend dev, I am happy with every solution (should be a backender decision).

PR #40999, opened 2023-10-19
files: fix virtual list will-change value (60fps superpowers)

pulsejet:
Follow up of #40958

I got the `will-change` wrong previously; it's the table that's scrolling, not the `tbody`. The performance with this patch is absolutely insane (we might want to increase the buffer a bit from 3 though). It's so good, actually, that I'm considering making this scroller into a standalone library (unless there's already a plan for that).

Here's how the updates during scrolling look like:

| wc_before | wc_after |


pulsejet:
Note: I noticed this is somewhat browser dependent. Firefox seems to already have been detecting and applying this optimization. So the main difference here is for Chromium (and maybe WebKit)

susnux (approved):

susnux:
Nice! :rocket: 

But please fix the missing signoff :wink:

pulsejet:
> But please fix the missing signoff 😉

Argh not again 😩 .. done!

> we might want to increase the buffer a bit from 3 though

need opinion on this btw, if anyone can test scrolling fast in a large directory. I think 5 or 6 might be good

skjnldsv (approved):

skjnldsv:
> need opinion on this btw, if anyone can test scrolling fast in a large directory. I think 5 or 6 might be good (EDIT: or maybe it's not that important ... can't see a lot of difference)

All my tests and development over the last months have been done on a 1k images folder :)

skjnldsv:
@pulsejet your commit names doesn't follow [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
`files: fix virtual list will-change value` should be `fix(files): virtual list will-change value` (adjusted)

PR #41010, opened 2023-10-20
feat(files): support nested actions

### LINKED ISSUE #39718: Add API for FileAction submenu actions
An API should be added to https://github.com/nextcloud/server/blob/8606f162c5b5b5d6a355e7e2f27e05585b3d0360/apps/files/src/services/FileAction.ts#L80 to allow for an action to open a submenu of actions

skjnldsv:
- Fix https://github.com/nextcloud/server/issues/39718
- [x] Needs https://github.com/nextcloud-libraries/nextcloud-files/pull/814
- [x] Needs https://github.com/nextcloud-libraries/nextcloud-files/pull/823
- [x] Needs https://github.com/nextcloud-libraries/nextcloud-vue/pull/4754

Peek 03-11-2023 18-52


apps/files/src/init.ts
```diff
@@ -62,3 +63,39 @@ registerRecentView()
...
+	exec: async () => alert('Hello 3'),
+	parent: 'menu',
+}))
+registerFileAction(new FileAction({
+	id: 'submenu4',
+	displayName: () => 'Submenu 4',
+	iconSvgInline: () => MenuIcon,
+	exec: async () => alert('Hello 4'),
+	parent: 'menu',
+}))
```
skjnldsv, line 101: Obviously testing purpose

skjnldsv:
Setting as draft, please review anyway, then I'll release a new files lib update and we can rebase/compile/merge :rocket:

skjnldsv:
> I am not completely sure if possible but from design perspective maybe a menu should rather be shown like this for example:
> 
> image
> 
> (quick and dirty mockup)
> 
> WDYT @nextcloud/designers

That's exactly what is written in the first post :grimacing:

szaimen:
> That's exactly what is written in the first post 😬

sorry, skipped this section 😅

Pytal (commented):
For rich functionality other action types, i.e. NcAction* components, should be supported in addition to the simple button action

For the reminder feature at least NcActionInput would be needed

skjnldsv:
> For rich functionality other action types, i.e. NcAction* components, should be supported in addition to the simple button action

That was not part of the plan :see_no_evil: 
That would mean we're scoping our API to be locked with the vue component, which I think we should not. IIRC from our discussion with the design team, there was only one "rich" element, which was the date picker, and that can be mitigated or worked around I think. :thinking: 

Quick thinking on our options, but for the custom date picker, we could:
- Mount a datepicker somewhere and trigger the browser date picker
- Use a popup

marcoambrosini (commented):
I think that for this we should use floating-vue's submenus
https://floating-vue.starpad.dev/guide/component#sub-menu

https://github.com/nextcloud/server/assets/26852655/5b5ded99-e949-4b75-a04d-f2c887e11bf4

skjnldsv:
@marcoambrosini irrelevant here, we're not impelmenting a new design.
Any menu like yours needs to be discussed globally and handled in the vue components.

here we're implementing the same layout as other apps as it's been decided a while ago
Please raise your ticket in https://github.com/nextcloud-libraries/nextcloud-vue

nimishavijay:
> I think that for this we should use floating-vue's submenus

Hm, I agree that I see this cascading submenu more in desktop applications, but it also has more edge cases where it could look not so ideal, for example in mobile view, at the right edge of the screen, etc. I am okay with the paginated submenu we have atm. 

> Still showing the provided icon and putting a chevron icon at the end would be nice, but it's not supported by the vue components afaik

Agreed, best way forward here, but if not possible we could just stick to the provided icon rather than changing it to a chevron.

In the sublevel, I would also say instead of "Back" we show the parent level item text to help with understanding which level you are at.

It would also be nice to include an animation for the transition to make it look smoother, sort of like this:

https://github.com/nextcloud/server/assets/52440189/bf75d801-b499-40f6-b55f-03f83107df35

skjnldsv:
> In the sublevel, I would also say instead of "Back" we show the parent level item text to help with understanding which level you are at.

Will be done here

For the following points:
>> Still showing the provided icon and putting a chevron icon at the end would be nice, but it's not supported by the vue components afaik
>
> Agreed, best way forward here, but if not possible we could just stick to the provided icon rather than changing it to a chevron.
>
> It would also be nice to include an animation for the transition to make it look smoother, sort of like this:

This is for the vue components

Pytal:
> * Use a popup: [Use a dialog for custom date form instead of including it in the action menu spreed#10202](https://github.com/nextcloud/spreed/issues/10202)

+1 for popup

Since this would be a new component we'd need final approval from designers and some way to have the native datetime picker expanded in the dialog without having to click the input first

jancborchardt (commented):
I’d say it does work and is nice – but agree that an animation for the Vue component is needed to communicate better what happens. :)

jancborchardt:
And cc @marcoambrosini since he brought up https://github.com/nextcloud/spreed/issues/10202

skjnldsv:
@Pytal please review the logic!
Especially the condition on where a reminder is shown or not, as I wasn't part of the original discussions! I used your services, so I think it should be good.
But here are a few questions to be safe
1. Is it normal that depending on the time of day, only some options are visible?
2. Is this feature enabled for all nodes, no matter what they are and what permissions they have? (folder/files, etc etc)

apps/files_reminders/src/actions/setReminderSuggestionActions.scss
```diff
@@ -0,0 +1,38 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+ // TODO: remove when/if the actions API supports a separator
+ // This the last preset action, so we need to add a separator
+.files-list__row-action-set-reminder-3 {
```
Pytal, line 24: Target `.files-list__row-action-set-reminder-custom` instead and use margin-top, etc. ? As the "Next week" action is not always rendered

apps/files_reminders/src/actions/setReminderSuggestionActions.ts
```diff
@@ -0,0 +1,106 @@
...
+
+// Generate the default preset actions
+export const actions = [laterToday, tomorrow, thisWeekend, nextWeek].map((option): FileAction|null => {
+	const dateTime = getDateTime(option.dateTimePreset)
+	if (!dateTime) {
+		return null
+	}
+
+	return new FileAction({
+		id: `set-reminder-${option.dateTimePreset}`,
```
Pytal, line 74: Let's change DateTimePreset to a string enum then if it's used this way

apps/files_reminders/src/actions/setReminderSuggestionActions.ts
```diff
@@ -0,0 +1,106 @@
...
+		displayName: () => `${option.label} - ${getDateString(dateTime)}`,
+		title: () => `${option.ariaLabel} – ${getVerboseDateString(dateTime)}`,
```
Pytal, line 76:
```suggestion
		displayName: () => `${option.label} – ${getDateString(dateTime)}`,
		title: () => `${option.ariaLabel} – ${getVerboseDateString(dateTime)}`,
```
En dash

apps/files_reminders/src/actions/setReminderSuggestionActions.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+const nextWeek: ReminderOption = {
+	dateTimePreset: DateTimePreset.NextWeek,
+	label: t('files_reminders', 'Next week'),
+	ariaLabel: t('files_reminders', 'Set reminder for next week'),
+}
+
+// Generate the default preset actions
+export const actions = [laterToday, tomorrow, thisWeekend, nextWeek].map((option): FileAction|null => {
```
Pytal, line 67: Let's extract this anonymous function to a `const computeAction` and write this as `.map(computeAction).filter(Boolean)` for readability

apps/files_reminders/src/components/SetCustomReminderModal.vue
```diff
@@ -0,0 +1,199 @@
...
+			setTimeout(() => {
+				const input = document.getElementById('set-custom-reminder') as HTMLInputElement
+				input.focus()
+				input.showPicker()
+			}, 300)
+		},
+
+		async setCustom(): Promise<void> {
+			// Handle input cleared
+			if (this.customDueDate === '') {
```
Pytal, line 152:
```suggestion
			if (this.customDueDate === null) {
```
Since https://github.com/nextcloud-libraries/nextcloud-vue/pull/4540/commits/05d58f63c7f34558226fa248377693d8c3f09cbf

apps/files_reminders/src/services/customPicker.ts
```diff
@@ -0,0 +1,48 @@
...
+document.body.appendChild(mount)
+
+// Create a new Vue instance and mount it to our modal container
+const CustomReminderModal = new View({
+	name: 'SetCustomReminderModal',
+	el: mount,
+})
+
+export const pickCustomDate = async (node: Node): Promise<void> => {
+	console.debug('CustomReminderModal', mount, CustomReminderModal)
```
Pytal, line 40:
```suggestion
```

Pytal:
> 1. Is it normal that depending on the time of day, only some options are visible?

Yes this is by design

> 2. Is this feature enabled for all nodes, no matter what they are and what permissions they have? (folder/files, etc etc)

Yes, if you can see (read) the file then you should be able to set a reminder on it

apps/files_reminders/src/actions/setReminderSuggestionActions.ts
skjnldsv: Done

susnux (changes requested):
The focus handling does not work properly:
If you click (keyboard) on "set reminder" the submenu opens but the focus is not set on the submenu's first element.

Also if you click the back button the first submenu is reopend, but the focus is not on the "set reminder" entry.

In both cases the focus is on some random non visible element.

apps/files_reminders/src/actions/setReminderSuggestionActions.ts
```diff
@@ -0,0 +1,116 @@
...
+			// Silent success as we display our own notification
+			return null
+		},
+
+		order: 21,
+	})
+}
+
+// Generate the default preset actions
+export const actions = [laterToday, tomorrow, thisWeekend, nextWeek]
```
Pytal, line 114: Since the presence of the action should be evaluated ideally on click instead of on script execution I think the FileAction enabled callback needs to be set to something like `() => Boolean(getDateTime(option.dateTimePreset))` to be more reactive

skjnldsv: Can be improved later

susnux (approved):
:elephant: a11y in follow up?

Pytal (approved):

PR #41013, opened 2023-10-20
Explicitly close the directory stream

solracsf:
## Summary

The directory handle `$dh` is explicitly closed using `closedir($dh)` to release resources after iterating through the directory.
This ensures proper resource management.

## Checklist

### DIFF
--- a/apps/encryption/lib/Command/FixKeyLocation.php
+++ b/apps/encryption/lib/Command/FixKeyLocation.php
@@ -366,6 +366,7 @@ class FixKeyLocation extends Command {
 					}
 				}
 			}
+			closedir($dh);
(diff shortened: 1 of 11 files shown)


Altahrim (approved):

nickvergessen (approved):

artonge (approved):

icewind1991 (changes requested):
When the function returns the handles get dropped and the resources cleaned up automatically. Manually closing the handles only makes a difference if there is a significant amount of work being done without dropping the handle.

solracsf:
You're right, closing then as there is no point to do it. 👍

PR #41017, opened 2023-10-20
Move IToken and IProvider::getToken to OCP

come-nc:
## Summary

Move `IToken` to `OCP` and add `getToken` method in the IProvider already in OCP, so that applications can rely upon it.
I made it so that applications using the interfaces from OC do not have to change their code until they want to.

More method could/should be added to `OCP\Authentication\Token\IProvider` once the idea is approved. If you know which methods applications are using you can list them in the comments.

## Checklist


lib/private/Authentication/Exceptions/ExpiredTokenException.php
```diff
@@ -27,17 +27,19 @@
...
-class ExpiredTokenException extends InvalidTokenException {
-	/** @var IToken */
-	private $token;
-
-	public function __construct(IToken $token) {
-		parent::__construct();
-
-		$this->token = $token;
+/**
+ * @deprecated 28.0.0 use OCP version instead
```
nickvergessen, line 31:
```suggestion
 * @deprecated 28.0.0 use {@see \OCP\Authentication\Exceptions\ExpiredTokenException} instead
```

That is clickable from phpstorm at least :)

lib/private/Authentication/Exceptions/ExpiredTokenException.php
```diff
@@ -27,17 +27,19 @@
...
+	public function __construct(
+		IToken $token,
+	) {
+		parent::__construct($token);
 	}
 
 	public function getToken(): IToken {
-		return $this->token;
+		$token = parent::getToken();
+		/** @var IToken $token We know that we passed OC interface from constructor */
```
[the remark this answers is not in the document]
nickvergessen: or would you need to make the `$token` protected then instead (which I would favor then)

lib/private/Authentication/Token/IToken.php
```diff
@@ -26,109 +26,7 @@
...
-	 */
-	public function setPassword(string $password);
-
-	/**
-	 * Set the expiration time of the token
-	 *
-	 * @param int|null $expires
-	 */
-	public function setExpires($expires);
+interface IToken extends OCPIToken {
```
nickvergessen, line 31: Add deprecation?

lib/private/Authentication/Token/PublicKeyToken.php
```diff
@@ -191,7 +187,7 @@ public function getScopeAsArray(): array {
 		return $scope;
 	}
 
-	public function setScope($scope) {
+	public function setScope(mixed $scope): void {
```
nickvergessen, line 190: Could be more specific on the supported types:
```suggestion
	public function setScope(string|array $scope): void {
```

lib/public/Authentication/Exceptions/ExpiredTokenException.php
```diff
@@ -0,0 +1,49 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCP\Authentication\Exceptions;
+
+use OCP\Authentication\Token\IToken;
+
+/**
+ * @since 28.0
```
nickvergessen, line 31: Usually we add the full version
```suggestion
 * @since 28.0.0
```

lib/public/Authentication/Token/IToken.php
```diff
@@ -0,0 +1,124 @@
...
+ */
+namespace OCP\Authentication\Token;
+
+use JsonSerializable;
+
+/**
+ * @since 28.0.0
+ */
+interface IToken extends JsonSerializable {
+	public const TEMPORARY_TOKEN = 0;
```
nickvergessen, line 35: Also need since annotations

lib/private/Authentication/Exceptions/ExpiredTokenException.php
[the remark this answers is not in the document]
come-nc: No, this function is there so that when you catch the `OC` exception and use `getToken` you know that you get the `OC` `IToken`. This allows current application code to keep running without psalm errors.

lib/public/Authentication/Token/IToken.php
come-nc: Ok, the CI is unaware it seems.

lib/private/Authentication/Exceptions/ExpiredTokenException.php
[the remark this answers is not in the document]
come-nc: Hum how would you handle that with a protected $token? I found it quite clean that it’s encapsulated in super-class and this one is only ensuring the type through constructor.

I tried with a protected property on psalm.dev it’s not happy if I declare it again with a subtype on the child class.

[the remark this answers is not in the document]
nickvergessen: > No, this function is there so that when you catch the OC exception and use getToken you know that you get the OC IToken. This allows current application code to keep running without psalm errors.

Ah, I missed that both things refer to different `IToken` implementations

lib/private/Authentication/Token/PublicKeyToken.php
come-nc: It can also be null it seems

nickvergessen (approved):

lib/public/Authentication/Token/IToken.php
```diff
@@ -0,0 +1,139 @@
...
+	 * Get the timestamp of the last password check
+	 * @since 28.0.0
+	 */
+	public function getLastCheck(): int;
+
+	/**
+	 * Set the timestamp of the last password check
+	 * @since 28.0.0
+	 */
+	public function setLastCheck(int $time): void;
```
ChristophWurst, line 90: nit: do we need to expose setters as well our would read access be sufficient for the public API?

ChristophWurst (approved):

lib/public/Authentication/Token/IToken.php
come-nc: I think it’s better to be able to deprecate the OC interface and avoid having to juggle with the two interfaces forever.
It would be pretty to have the setters in the OC but I fear it’s not worth the hassle.

PR #41018, opened 2023-10-20
fix(files): disable size containment on headers

pulsejet:
Fixes a regression caused by the header being variable height: see https://github.com/nextcloud/server/pull/40958#issuecomment-1772366976

cc @juliushaertl


skjnldsv (approved):
AH, yes, makes more sense indeed :)

susnux (approved):

juliusknorr (approved):

juliushaertl:
Thanks for the quick fix 💙

susnux (approved):

juliushaertl:
Sorry to bug again, unfortunately this still seems to cause issues with the popover menu that text might show:

Screenshot 2023-10-24 at 15 34 03

Any chance you could have another look at that as well?

PR #41021, opened 2023-10-20
fix(tests): Move leftover acceptance tests for users from drone to Cypress

susnux:
## Summary

Move the tiny rest of acceptance tests for user setting to Cypress and add missing test cases:
* Change quota
* Set subadmin
* Add manager

## Checklist


susnux:
BTW using random users should make it also less flaky, as now every test has a defined state.

skjnldsv (approved):

skjnldsv:
> BTW using random users should make it also less flaky, as now every test has a defined state.

If random users makes it flaky, the tests or our codeis probably the issue then.
I would prefer not hardcoding users, they do create a new universal "fresh Nextcloud" base and we should rely on it :)

susnux:
>  I would prefer not hardcoding users

Exactly, previously hardcoded "jane doe" was used, but that might lead to inconsistent state.
Now random users are created, so at least the user does not exists and we start with a fresh profile.

It would need some more work (I do not have the time currently) to rewrite the tests, as the best way (and the best practice from Cypress) would be to ensure we have a clean state **before** the tests and not clean up after the tests.

So e.g. removing the datadir and resetting the database before each test.

Thinking about this hardcoding users might also be good if we always use the same and have something like:
```
before each: delete user "test" if exists. Recreate a fresh user "test".
```

But especially as our Cypress tests are also changing admin settings (we have a reset but only for theming), the golden standard would be to reset all settings by running every testsuite with the initial database.
So not sure if possible but creating a snapshot of the datadir right before cypress starts (as that includes the sqlite database) and then having a command that resets the datadir to the snapshot in the `before` hook.

apps/settings/src/components/Users/UserRow.vue
```diff
@@ -36,11 +36,11 @@
 				:user="user.id" />
 		</td>
 
-		<td class="row__cell row__cell--displayname" data-test-id="cell-displayname">
+		<td class="row__cell row__cell--displayname" data-testid="cell-displayname">
 			<template v-if="editing && user.backendCapabilities.setDisplayName">
 				<NcTextField ref="displayNameField"
 					class="user-row-text-field"
-					data-test-id="input-displayName"
```
skjnldsv, line 43: Ah, we talked abut this before with various other devs. I advocated for the usage of `data-cy-xxxx` and seemed that other devs agreed too after a some discussion. This allow to scope the elements.

I know what the [best-practice](https://docs.cypress.io/guides/references/best-practice) says, but I think they do not have our overall ecosystem, so things differs a bit :)

The data attribute value should be used when it have a real changing value. Like `data-cy-userslist-user="admin"`. Otherwise it makes really weird to target a specific element. 
I also prefer to have more unique data attributes. The fact that you can grep all `[data-cy="files-list-row"]` makes no sense imho.
But `[data-cy-files-list-row]` and `[data-cy-files-list-row-fileid="12345"]` does :)

Any thoughts or counter arguments?
cc @artonge @max-nextcloud @Pytal  as I recall you were in that discussion before :wink:

skjnldsv:
> Exactly, previously hardcoded "jane doe" was used, but that might lead to inconsistent state.
> Now random users are created, so at least the user does not exists and we start with a fresh profile.

Ah, I read it as the opposite :see_no_evil: 
I thought you meant we should go for hardcoded users with Cypress :wink:

apps/settings/src/components/Users/UserRow.vue
susnux: For me it makes no difference if I use `data-testid` or `data-cy` both a pretty common and standard, so I am fine with both.

> The fact that you can grep all [data-cy="files-list-row"] makes no sense imho.
> But [data-cy-files-list-row] and [data-cy-files-list-row-fileid="12345"] does :)

Well for me not that much :sweat_smile:
`data-cy="files-list-row"` for me this is like the `id` (without its singleton limitation), so it says "Select all elements that are file list rows". While `data-cy-files-list-row-fileid="12345"` says "get the file list row that for file ID 12345".

So my problem with `data-cy-files-list-row` vs  `data-cy="files-list-row"` is that there is no consistent naming for identifying elements, you always need to use a specific attribute rather that a specific attribute ID.
This feels wrong in a HTML context.

But this is not about feels, so I have no problem to stick with one notation for all of our code and if it was decided to go with `data-cy-some-identifier` as boolean attribute then I am fine with it. As there is no technical downside :smiley: 

Need some time to adjust this PR then for `data-cy` and we should probably add this to our development guides :)

susnux: TL;DR: Will take some time this week to adjust the test selectors :wink:

skjnldsv: > `data-cy="files-list-row"` for me this is like the `id` (without its singleton limitation), so it says "Select all elements that are file list rows". While `data-cy-files-list-row-fileid="12345"` says "get the file list row that for file ID 12345".

If you want all the lines, then you can also do `[data-cy-files-list-row-fileid]`, without the id.
That's the benefit of scoped selectors. And with `[data-cy="files-list-row"]`, you cannot select by Ids with that syntax.
Or with something like `[data-cy="files-list-row-123456"]`, but I find it incredibly messy :shrug:

artonge (approved):

Pytal (approved):

susnux:
Cypress error caused by theming, so will do a follow up fixing theming tests.

PR #41034, opened 2023-10-21
fix(settings): users page html validation

ShGKme:
* Resolves: A part of https://github.com/nextcloud/andy/issues/5

## Summary

- **Error: Attribute `button-class` not allowed on element [`div`](https://html.spec.whatwg.org/multipage/#the-div-element) at this point.**
  - [x] Remove non-existing `button-class` from `NcAppNavigationNew`
- **Error: The value of the `for` attribute of the `label` element must be the ID of a non-hidden form control.**
  - [x] Typo: use `input-id` instead of `id` on `NcSelect` to be labeled
- **Error: Bad value `quota-progressAdmin 25x1kcgi6srs` for attribute `for` on element [`label`](https://html.spec.whatwg.org/multipage/#the-label-element): An ID must not contain whitespace.**
  **Error: Bad value `quota-progressAdmin 25x1kcgi6srs` for attribute `id` on element [`progress`](https://html.spec.whatwg.org/multipage/#the-progress-element): An ID must not contain whitespace.**
  - [x] Encode `uniqueId` with user's ID, because `user.id` may contain space

Also change `<label>` for `<progress>` in user quota with `aria-labelledby`, because `<label>` doesn't work here in some screen readers (NDVA) and there is no need in an actual clickable `<label>` link.

No visual changes.

## Checklist


susnux (approved):

JuliaKirschenheuter (approved):

ShGKme:
Cypress fail **is related**

susnux:
> Or not... Works fine locally

Not related, the tests are quite flaky. Pushed a fix here:
https://github.com/nextcloud/server/pull/41021

Issue #40690, opened 2023-09-28
[Bug]: EXTERNAL STORAGE: renaming object by just changing case on SMB storage leads to strange results


alexanderharm:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Renaming an object (here folder/directory) on external storage connected via SMB leads to strange results depending on the client.

WebGUI:
- renaming `99_test` to `99_TEST` produces an error message saying that the name is already in use and that I should use a different name
- HOWEVER: in the GUI one can see now both folders `99_test` AND `99_TEST`
- if I create an object in `99_test` it also shows in `99_TEST` and vice versa
- if I delete either of the folders then both are either deleted or one is deleted and the other dysfunctional (cannot create any files or folders inside it)
- on the connected Nextcloud client on Windows the user can just see the original `99_test`

Nextcloud client (Windows) with virtual files:
- renaming `99_test` to `99_TEST` produces an error message saying that it failed caused by a protocol error
- client sees folder as `99_test`, sync status remains active
- creating any objects in `99_test` fails in client and WebGUI, folder is dysfunctional

### Steps to reproduce

1. create folder on external storage connected via SMB
2. rename the folder by just changing the case of a letter

### Expected behavior

renaming by just changing case is detected by WebGUI and Nextcloud client as illegal on SMB storage and no file operations are executed.

### Installation method

Other Community project

### Nextcloud Server version

27

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database engine version

PostgreSQL

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

[...]

PR #41053, opened 2023-10-23
fix(files_external): on case insensitive system, block case change

Altahrim:
* Ref: #40690 

## Summary

Add an option to declare a SMB share case-insensitive.

When a file/directory is renamed to the same name with only case change, the rename fail. We block this kind of rename.

The user will have to rename to another name first.

### DIFF
--- a/apps/files_external/lib/Lib/Storage/SMB.php
+++ b/apps/files_external/lib/Lib/Storage/SMB.php
@@ -93,6 +93,8 @@ class SMB extends Common implements INotifyStorage {
 	/** @var bool */
 	protected $showHidden;
 
+	private bool $caseSensitive;
+
 	/** @var bool */
 	protected $checkAcl;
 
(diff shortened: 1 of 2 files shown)


apps/files_external/lib/Lib/Storage/SMB.php
```diff
@@ -674,6 +683,16 @@ public function mkdir($path) {
...
+			if ($this->caseSensitive === false) {
+				$filename = basename($path);
+				$siblings = $this->getDirectoryContent(dirname($this->buildPath($path)));
+				foreach ($siblings as $sibling) {
+					if ($sibling['name'] === $filename) {
+						return true;
+					}
+				}
+				return false;
+			}
```
come-nc, line 695: This is to make the file_exists case sensitive? Or the other way around?

From what I understand this is to avoid `file_exists('FOO')` to return `true` for a file named `foo`, but I’m not sure why?

Altahrim: Yes, now `file_exists` returns `true` only if file exists with the same case on file system. 
The same kind of fix have been written for Local storage: https://github.com/nextcloud/server/pull/40935
It allows to keep the same kind of behaviour with case-sensitive/insensitive filesystems.

come-nc (approved):

juliusknorr (approved):

Altahrim:
CI failure seems unrelated

PR #41055, opened 2023-10-23
feat(profile): Add public interface for profile manager so apps can check config

nickvergessen:
## Summary

- Required for apps that register profile fields to limit the API side of things based on the users selection, e.g. https://github.com/nextcloud/spreed/pull/10542

## Checklist

### DIFF
--- a/core/Db/ProfileConfig.php
+++ b/core/Db/ProfileConfig.php
@@ -26,7 +26,6 @@ declare(strict_types=1);
 
 namespace OC\Core\Db;
 
-use OCP\Profile\IProfileManager;
 use function Safe\json_decode;
 use function Safe\json_encode;
 use \JsonSerializable;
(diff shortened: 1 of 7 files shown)


come-nc (approved):

lib/private/Profile/ProfileManager.php
```diff
@@ -26,8 +26,9 @@
 
 namespace OC\Profile;
 
-use function Safe\array_flip;
-use function Safe\usort;
+use OCP\Profile\IProfileManager;
+use function array_flip;
+use function usort;
```
fsamapoor, line 31: Would you please educate me on the reason for this change?

core/Db/ProfileConfig.php
```diff
@@ -29,7 +29,6 @@
 use function Safe\json_decode;
 use function Safe\json_encode;
```
fsamapoor, line 30: I want to know whether there is a standard in place for using safe functions in the codebase.

fsamapoor (approved):

lib/private/Profile/ProfileManager.php
nickvergessen: It was never intended to use functions of an indirect dependency of server 3rdparty.
The problem is if you import the function using PHPStorm, it will always import them from the library, ignoring the fact it's an indirect dependency of the project, so any update of the library using it would drop it and break this usage by accident.

core/Db/ProfileConfig.php
nickvergessen: Will send a follow up adjusting all of them

PR #41056, opened 2023-10-23
optimize isShared and isMounted

icewind1991:
Checking the mountpoint is faster than `instanceOfStorage` because is doesn't have to deal with wrappers.


icewind1991:
Merging this into #41119 due to code overlap

PR #41057, opened 2023-10-23
optimize UserMountCache::registerStorage

icewind1991:
Simplify the logic a bit and use less intermediate arrays

### DIFF
--- a/lib/public/Files/Config/ICachedMountInfo.php
+++ b/lib/public/Files/Config/ICachedMountInfo.php
@@ -84,4 +84,12 @@ interface ICachedMountInfo {
 	 * @since 24.0.0
 	 */
 	public function getMountProvider(): string;
+
+	/**
+	 * Get a key that uniquely identifies the mount
+	 *
+	 * @return string
+	 * @since 28.0.0
+	 */
+	public function getKey(): string;
 }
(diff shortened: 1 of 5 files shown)


lib/public/Files/Config/ICachedMountInfo.php
```diff
@@ -84,4 +84,11 @@ public function getRootInternalPath(): string;
 	 * @since 24.0.0
 	 */
 	public function getMountProvider(): string;
+
+	/**
+	 * Get a key that uniquely identifies the mount
+	 *
+	 * @return string
+	 */
+	public function getKey(): string;
```
blizzz, line 93: missing @since annotation

blizzz (changes requested):
Missing required annotation, otherwise :+1:

lib/public/Files/Config/ICachedMountInfo.php
Altahrim: Just added it :)

blizzz (approved):

juliusknorr (approved):

juliusknorr (changes requested):
Ah wait CI shows some issue:

```
PHP Fatal error:  Nesting level too deep - recursive dependency? in /home/runner/work/server/server/lib/private/Files/Config/UserMountCache.php on line 130
```

.htaccess
```diff
@@ -96,3 +96,7 @@
 
 AddDefaultCharset utf-8
 Options -Indexes
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
juliushaertl, line 99: Unrelated :)

Altahrim: Arf yes, I don't know why it's modified… Thanks!

juliusknorr (approved):

PR #21734, opened 2020-07-07
disable touch to update mtime for external s3 storage

icewind1991:
since updating the mtime requires doing a copy this can have a significant overhead and it messes with versioning

Signed-off-by: Robin Appelman <robin@icewind.nl>

PR #41062, opened 2023-10-23
Return correct status if touch failed

artonge:
When uploading a file to an external s3 bucket, the original mtime is not kept. It boils down to the [touch operation being deactivated for existing file for s3 buckets](https://github.com/nextcloud/server/blob/834c9a209ebef7b8d5795b53a018577fe4e534ca/apps/files_external/lib/Lib/Storage/AmazonS3.php#L553). Limitation introduced here: https://github.com/nextcloud/server/pull/21734

This PR changes the return value of the `touch` method. Now this method returns `false` when the file already exists to trigger the following condition: https://github.com/nextcloud/server/blob/e66e8bad1a711ec3073e901e824edc3f92dbd107/lib/private/Files/View.php#L560-L571

### DIFF
--- a/apps/files_external/lib/Lib/Storage/AmazonS3.php
+++ b/apps/files_external/lib/Lib/Storage/AmazonS3.php
@@ -550,18 +550,20 @@ class AmazonS3 extends \OC\Files\Storage\Common {
 		];
 
 		try {
-			if (!$this->file_exists($path)) {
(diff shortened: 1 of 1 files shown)


icewind1991 (approved):

nfebe (approved):

artonge:
CI failure is unrelated

PR #41081, opened 2023-10-24
Add a command to run the setup checks from CLI

come-nc:
## Summary

Add a command to run the setup checks from the cli.
This will help setup checks development by having an easy way to test setup checks and see their names&descriptions.
This will also be useful for admins who want to check their server problems from cli.

Screenshot:
Screenshot_20231026_120306

Note that names, categories and descriptions are not finals. The point of this command is also to be able to see all of them easily and adapt what looks wrong.

## TODO

- [x] Only failure by default and success with --verbose?
- [x] Improve warning/info differences
- [x] Fix descriptions and categories (not in this PR) https://github.com/nextcloud/server/pull/41083
- [x] Decide on naming: is setupchecks the right command name? Should this replace check? Some other name?

## Checklist

### DIFF
--- a/core/Command/SetupChecks.php
+++ b/core/Command/SetupChecks.php
@@ -64,6 +64,7 @@ class SetupChecks extends Base {
 						$emoji = match ($check->getSeverity()) {
 							'success' => '✓',
 							'error' => '❌',
+							'warning' => '⚠',
 							default => 'ℹ',
 						};
 						$verbosity = ($check->getSeverity() === 'error' ? OutputInterface::VERBOSITY_QUIET : OutputInterface::VERBOSITY_NORMAL);
--- a/core/Command/SetupChecks.php
+++ b/core/Command/SetupChecks.php
@@ -66,12 +66,15 @@ class SetupChecks extends Base {
 							'error' => '❌',
 							default => 'ℹ',
 						};
+						$verbosity = ($check->getSeverity() === 'error' ? OutputInterface::VERBOSITY_QUIET : OutputInterface::VERBOSITY_NORMAL);
+						$description = $check->getDescription();
 						$output->writeln(
 							"\t\t<{$styleTag}>".
 							"{$emoji} ".
 							$title.
-							($check->getDescription() !== null ? ': '.$check->getDescription() : '').
-							"</{$styleTag}>"
+							($description !== null ? ': '.$description : '').
+							"</{$styleTag}>",
+							$verbosity
 						);
 					}
 				}
@@ -79,10 +82,10 @@ class SetupChecks extends Base {
 		foreach ($results as $category => $checks) {
 			foreach ($checks as $title => $check) {
 				if ($check->getSeverity() !== 'success') {
-					return 1;
+					return self::FAILURE;
 				}
 			}
 		}
-		return 0;
+		return self::SUCCESS;
 	}
 }
--- a/core/Command/SetupChecks.php
+++ b/core/Command/SetupChecks.php
@@ -59,7 +59,8 @@ class SetupChecks extends Base {
 						$styleTag = match ($check->getSeverity()) {
 							'success' => 'info',
 							'error' => 'error',
-							default => 'comment',
+							'warning' => 'comment',
+							default => null,
 						};
 						$emoji = match ($check->getSeverity()) {
 							'success' => '✓',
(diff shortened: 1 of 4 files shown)


core/Command/SetupChecks.php
```diff
@@ -0,0 +1,88 @@
...
+							'error' => '❌',
+							default => 'ℹ',
+						};
+						$output->writeln(
+							"\t\t<{$styleTag}>".
+							"{$emoji} ".
+							$title.
+							($check->getDescription() !== null ? ': '.$check->getDescription() : '').
+							"</{$styleTag}>"
+						);
```
Altahrim, line 75: I think we can use different `VERBOSE` level here, so with `-q` it will only display errors:
```suggestion
						$verbosity = $check->getSeverity() === 'error' ? OutputInterface::VERBOSITY_QUIET : OutputInterface::VERBOSITY_NORMAL;
						$output->writeln(
							"\t\t<{$styleTag}>".
							"{$emoji} ".
							$title.
							($check->getDescription() !== null ? ': '.$check->getDescription() : '').
							"</{$styleTag}>",
							$verbosity,
						);
```

core/Command/SetupChecks.php
```diff
@@ -0,0 +1,88 @@
...
+					}
+				}
+		}
+		foreach ($results as $category => $checks) {
+			foreach ($checks as $title => $check) {
+				if ($check->getSeverity() !== 'success') {
+					return 1;
```
Altahrim, line 82:
```suggestion
					return Base::SUCCESS;
```

core/Command/SetupChecks.php
```diff
@@ -0,0 +1,88 @@
...
+				}
+			}
+		}
+		return 0;
```
Altahrim, line 86:
```suggestion
		return Base::FAILURE;
```

core/Command/SetupChecks.php
```diff
@@ -0,0 +1,92 @@
...
+						$styleTag = match ($check->getSeverity()) {
+							'success' => 'info',
+							'error' => 'error',
+							default => 'comment',
+						};
+						$emoji = match ($check->getSeverity()) {
+							'success' => '✓',
+							'error' => '❌',
+							'warning' => '⚠',
+							default => 'ℹ',
```
Altahrim, line 68: Nit:
To keep the same style: 
```suggestion
							'success' => '✓',
							'error' => '✗',
							'warning' => '⚠',
							default => 'ℹ',
```
or
```suggestion
							'success' => '✅',
							'error' => '❌',
							'warning' => '⚠️',
							default => 'ℹ️',
```

core/Command/SetupChecks.php
```diff
@@ -0,0 +1,92 @@
...
+				$this->writeArrayInOutputFormat($input, $output, $results);
+				break;
+			default:
+				foreach ($results as $category => $checks) {
+					$output->writeln("\t{$category}:");
+					foreach ($checks as $title => $check) {
+						$styleTag = match ($check->getSeverity()) {
+							'success' => 'info',
+							'error' => 'error',
+							default => 'comment',
```
Altahrim, line 62: No warning here?

come-nc: Did not find it in symfony docs.
I tried and `warning` does not work. I did not want to go into manual styling so :shrug:

core/Command/SetupChecks.php
come-nc: Screenshot_20231026_153458
Screenshot_20231026_153410
I think I prefer the "text" version (first one), what do you think?

core/Command/SetupChecks.php
nickvergessen: `comment` is the `warning` style... it's so confusing and throwing me off everytime I use it.

core/Command/SetupChecks.php
nickvergessen: I agree, it's better to keep those informative and OK thing "passive"

core/Command/SetupChecks.php
come-nc: Yeah, but setupchecks have `success`/`info`/`warning`/`error`, and in symfony styles I have only 3 available I think, so both `info` and `warning` end up with `comment` style, only the icon differs.

nickvergessen: https://github.com/nextcloud/3rdparty/blob/fcc80609824be396681d5a68ba495f9b85c9b0bb/symfony/console/Formatter/OutputFormatter.php#L74-L77

there are 4, question would fit for info, but its black font on cyan background so might be too drastic for info. Maybe just use default text color for info, instead of yellow comment??

Altahrim (approved):

ChristophWurst (approved):

come-nc:
Added to documentation by https://github.com/nextcloud/documentation/pull/12171

PR #41082, opened 2023-10-24
[stable27] fix(isLegitimatedForUserId): Setup mountpoints to check file access

mejo-:
Manual backport of #40482

This fixes workflows on groupfolders, as it will consider access to files in groupfolders.

It also fixes false positives where access to files was limited by other means not taken into account before, e.g. access control.

For postDelete events, check for permissions of the parent folder instead, as the file itself no longer exists.

Fixes: https://github.com/nextcloud/flow_notifications/issues/71

Idea taken from https://github.com/nextcloud/server/pull/38946


mejo-:
Manually tested on stable27.

icewind1991 (approved):

juliusknorr (approved):

PR #41088, opened 2023-10-24
enh(TextProcessing): Allow providers and task types to declare a dynamic ID instead of using className

marcelklehr:
this allows AppAPI to register anonymous classes as TextProcessing providers and task types

- [ ] Change settings implementation

## Checklist


marcelklehr:
(I'm sort of sad about the loss of beauty with this, but alas)

marcelklehr:
It's not as straightforward as with the STT API, because we need to be able to instantiate task types from their id. When we are using the class name as ID it works fine, but if we're using a getId() method we need to register task types as well. I'm not sure how we can do this in a backward-compatible way.

marcelklehr:
Decided with @julien-nc not to have task types implementable by ExApps. Then we can keep the class names for TaskTypes and this whole thing gets a lot easier.

marcelklehr:
Please review

lib/private/TextProcessing/Manager.php
```diff
@@ -282,7 +283,12 @@ public function getPreferredProviders(OCPTask $task): array {
 			$preferences = json_decode($json, true);
 			if (isset($preferences[$task->getType()])) {
 				// If a preference for this task type is set, move the preferred provider to the start
-				$provider = current(array_values(array_filter($providers, fn ($provider) => $provider::class === $preferences[$task->getType()])));
+				$provider = current(array_values(array_filter($providers, function ($provider) use ($preferences, $task) {
+          if ($provider instanceof IProviderWithId) {
+						return $provider->getId() === $preferences[$task->getType()];
+					}
+          $provider::class === $preferences[$task->getType()];
+        })));
```
juliushaertl, line 291: Indentation is off

juliusknorr (approved):
Small indent issue, otherwise looks good 👍

andrey18106 (approved):

bigcat88 (approved):

PR #41093, opened 2023-10-24
fix(files): let popper overflow in header

pulsejet:
Fix for https://github.com/nextcloud/server/pull/41018#issuecomment-1777217306

Tbh, I don't really like the popper being inside the header element because it violates the visual scope of the header. My first attempt at fixing this was to disable containment when the popper exists (easy: `has`, with fallback to class on body for Firefox). But then I realised other apps might add headers doing weirder stuff and this'll eventually cause problems :(

Anyway, with #40999 we no longer repaint the header unless necessary (it's never destroyed unlike the list elements), so this is no big deal.

BTW `has` is coming to stable FF very soon!


juliusknorr (approved):

skjnldsv (approved):

skjnldsv:
Please rebase and compile :)

PR #41105, opened 2023-10-25
fix(docs): Use PHP 8.1 for building the PHP API docs

juliusknorr:
We still need the additional `run`


skjnldsv (approved):
Seems to work without though :see_no_evil:

juliushaertl:
> Seems to work without though 🙈

I guess I need another coffee as well. I checked the wrong netlify build. Lets close that then.

PR #41110, opened 2023-10-25
fix(cypress): Fix selector for the user menu button

susnux:
## Summary

Adjusted the selector for the user menu toggle (upper right menu containing the logout button).

## Checklist

### DIFF
--- a/cypress/e2e/login/login.cy.ts
+++ b/cypress/e2e/login/login.cy.ts
@@ -137,7 +137,7 @@ describe('Login', () => {
 		cy.url().should('match', /apps\/dashboard(\/|$)/)
 
 		// When click logout
-		cy.get('#user-menu button').should('exist').click()
+		cy.get('#user-menu > button').should('exist').click()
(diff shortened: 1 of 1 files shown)


artonge (approved):

szaimen (approved):
Wouldn't the old selector have worked as well? Anyway, fine with the change :)

susnux:
> Wouldn't the old selector have worked as well? Anyway, fine with the change :)

No because it selected more than one button (when writing those tests it worked but now somehow a second button is there??).
So this simply selects only the direct child-button which is the menu toggle.

susnux:
Cypress unrelated:
```
  │ ✖  theming/admin-settings.cy.ts             01:58       42       40        2        -        - │
  ```

szaimen:
> No because it selected more than one button (when writing those tests it worked but now somehow a second button is there??).
> So this simply selects only the direct child-button which is the menu toggle.

Ah I see. That makes sense

PR #41116, opened 2023-10-25
remove some unneeded normalizePath calls

icewind1991:
While normalizing isn't that expensive, it does add up if we do it multiple times for every item in a folder

### DIFF
--- a/lib/private/legacy/OC_Helper.php
+++ b/lib/private/legacy/OC_Helper.php
@@ -491,7 +491,7 @@ class OC_Helper {
 		if (!$view) {
 			throw new \OCP\Files\NotFoundException();
 		}
-		$fullPath = Filesystem::normalizePath($view->getAbsolutePath($path));
+		$fullPath = rtrim($view->getAbsolutePath($path), '/');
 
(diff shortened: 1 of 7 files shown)


Altahrim (approved):

lib/private/legacy/OC_Helper.php
```diff
@@ -491,7 +491,7 @@ public static function getStorageInfo($path, $rootInfo = null, $includeMountPoin
 		if (!$view) {
 			throw new \OCP\Files\NotFoundException();
 		}
-		$fullPath = Filesystem::normalizePath($view->getAbsolutePath($path));
+		$fullPath = rtrim($view->getAbsolutePath($path), '/');
```
juliushaertl, line 494: I'm not entirely sure if we should skip the normalization here as `$path` might get passed from public API and is used as a cache key which could trigger the uncached calculation then more often

juliusknorr (commented):
I'm uncomfortable getting this merged in a draft state as I cannot tell about any potentially leftovers and the change itself feels a bit risky to do as I fear there might be side effects whenever a non-normalized path is used in other code parts.

PR #41135, opened 2023-10-26
fix(cypress): Also clone 3rdparty submodule to keep in sync for cypress tests

susnux:
## Summary

If `lib` uses updated `3rdparty` libraries but the shallow server is outdated (like now the server is 28 days old -> dependencies have been changed) it might lead to error 500 as the Nextcloud server can not find that changed dependencies.

So we need to also pull submodules on Cypress init.

## Checklist


Pytal (approved):

skjnldsv (approved):

nickvergessen:
Failing node test seems unrelated

PR #41003, opened 2023-10-19
chore(tests): Run all unit tests on GitHub instead of Drone

susnux:
* Part of https://github.com/nextcloud/server/issues/41187

## Summary

Lets move from drone to GitHub this way we can use our runners more flexible and restart flaky tests separatly instead of "drone unrelated".

## TODO
- [ ] Do we want to backport this for all branches?
- [ ] Make all the phpunit-summary tests required (someone with super powers)

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #41139, opened 2023-10-26
Move integration tests to github actions

nickvergessen:
* Part of https://github.com/nextcloud/server/issues/41187

## Summary

Run all integration tests on GitHub instead of Drone.

## TODO

- [ ] Make `integration-sqlite-summary` required

## Checklist


.github/workflows/integration-sqlite.yml
```diff
@@ -0,0 +1,169 @@
...
+              - '.github/workflows/**'
+              - 'appinfo/**'
+              - 'lib/**'
+              - 'templates/**'
+              - 'tests/integration/**'
+              - 'vendor/**'
+              - 'vendor-bin/**'
+              - '.php-cs-fixer.dist.php'
+              - 'composer.json'
+              - 'composer.lock'
```
susnux, line 46: Should be adjusted to match also apps/ subdirectories
```suggestion
             - '.github/workflows/**'
             - '3rdparty/**'
             - '**/*.php'
             - '**/lib/**'
             - '**/tests/**'
             - '**/vendor-bin/**'
             - '.php-cs-fixer.dist.php'
             - 'composer.json'
            - 'composer.lock'
```

.github/workflows/integration-sqlite.yml
```diff
@@ -0,0 +1,169 @@
...
+          ./occ config:system:set hashing_default_password --value=true --type=boolean
+
+      - name: Configure caching
+        if: ${{ matrix.test-suite == 'ldap_features' }}
+        run: |
+          ./occ config:system:set redis host --value=localhost
+          ./occ config:system:set redis port --value=6379 --type=integer
+          ./occ config:system:set redis timeout --value=0 --type=integer
+          ./occ config:system:set memcache.local --value='\OC\Memcache\Redis'
+          ./occ config:system:set memcache.distributed --value='\OC\Memcache\Redis'
```
susnux, line 146: Why not use redis everywhere might make tests faster?
We could then simplify to:
```suggestion
          cp tests/redis.config.php config/
          ./occ maintenance:install --verbose --database=sqlite --database-name=nextcloud --database-host=127.0.0.1 --database-port=$DB_PORT --database-user=root --database-pass=rootpassword --admin-user admin --admin-pass admin
          ./occ config:system:set hashing_default_password --value=true --type=boolean
```

susnux: (oh this only works with the changes from the unit test PR as there I changed the redis config to work with GH actions)

.github/workflows/integration-sqlite.yml
```diff
@@ -0,0 +1,172 @@
...
+        test-suite:
+          - 'capabilities_features'
+          - 'collaboration_features'
+          - 'comments_features'
+          - 'dav_features'
+          - 'features'
+          - 'federation_features'
+          - '--tags ~@large files_features'
+          - 'filesdrop_features'
+          - 'openldap_features'
```
nickvergessen, line 62: Why the openldap split. I merged folders as hard as possible to avoid 2.5-4mins of set up time for less then 5min of actual test execution 🙈

susnux: I know but the reason for the failing LDAP test were:
1. wrong host (`openldap` is not a working host as for github actions only localhost work)
2. missing port forwarding for openldap service
3. wrong data dir (the tests assert using `/dev/shm`)
4. **The `openldap_...` tests do not work together, they pollute the ldap server and break following tests**

Drone works because every tests is run separately and I do not have enough knowledge of Behat to fix the tests to leave in usable state or reset the state afterwards.

susnux:
@nickvergessen Tests are now working. But maybe you can have a look and replace the fixups with better solutions?
Seems like tests rely on different assumptions, e.g. LDAP rely on data directory being `/dev/shm` while some other tests require real access to that files (`federated.features).

Maybe adjust on of those locations so the tests all use the same data directory?

.github/workflows/integration-sqlite.yml
nickvergessen: Ah, the joy of dirty states

nickvergessen:
> Seems like tests rely on different assumptions, e.g. LDAP rely on data directory being /dev/shm while some other tests require real access to that files (`federated.features).

Yeah, let's keep it for now. The path is hardcoded for LDAP (maybe even into the LDAP config of the test docker, so changing would cause issues on stable branches, etc). Feel free to raise a follow up issue

.github/workflows/integration-sqlite.yml
nickvergessen: > Why not use redis everywhere might make tests faster?

Because I'm not sure we should have this. CI is not only about speed. But e.g. when redis is there, behaviour changes (e.g. BFP has a redis and a DB backend, etc), so each test should decide if they need it.

.github/workflows/integration-sqlite.yml
nickvergessen: yeah copy pastes from talk for now.

also needs `build/integration` added

susnux: Out of curiosity but why are the integration tests located in `build/` rather then `tests/`?

nickvergessen: Because the coworker doing it back at owncloud times put it there.
I would be fine with moving it, but it "breaks" local setups/scripts and open PRs (but there shouldn't be too many), so maybe we use this "refactoring" as the right point in time to do it?

.github/workflows/integration-s3-primary.yml
```diff
@@ -66,15 +72,15 @@ jobs:
...
+          echo '<?php $CONFIG=["${{ matrix.key }}" => ["class" => "OC\Files\ObjectStore\S3", "arguments" => ["bucket" => "nextcloud", "autocreate" => true, "key" => "nextcloud", "secret" => "bWluaW8tc2VjcmV0LWtleS1uZXh0Y2xvdWQ=", "hostname" => "localhost", "port" => 9000, "use_ssl" => false, "use_path_style" => true, "uploadPartSize" => 52428800]]];' > config/config.php
           echo '<?php $CONFIG=["redis" => ["host" => "localhost", "port" => 6379], "memcache.local" => "\OC\Memcache\Redis", "memcache.distributed" => "\OC\Memcache\Redis"];' > config/redis.config.php
           ./occ maintenance:install --verbose --database=sqlite --database-name=nextcloud --database-host=127.0.0.1 --database-user=root --database-pass=rootpassword --admin-user admin --admin-pass admin
           php -f index.php
 
       - name: Integration
         run: |
           cd build/integration
-          bash run.sh --tags "~@failure-s3" features/webdav-related.feature
+          bash run.sh --tags "~@failure-s3" rename to build/integration/dav_features/webdav-related.feature
```
nickvergessen, line 83: you failed here?

susnux: When you press save while git is in the commit message editor view... it does not get commited together :facepalm: 
Sorry, fixed that.

nickvergessen:
Rebased on master, ready to merge for less load on Drone

susnux:
> Rebased on master, ready to merge for less load on Drone

Lets then also do https://github.com/nextcloud/server/pull/41003

Especially nodb would be useful

build/integration/features/ratelimiting.feature
```diff
@@ -1,9 +1,9 @@
 Feature: ratelimiting
-  
+
   Background:
     Given user "user0" exists
     Given As an "admin"
-    Given app "testing" is enabled
+    Given invoking occ with "app:enable --force testing"
```
mejo-, line 6: Not sure about this change. It changes the logic from only checking if the app is enabled to force-enabling it, right? What changed that it was already enabled before and has to be enabled now?

mejo- (approved):
Overall looks good. Thanks so much for working on this :space_invader: 

Only minor comments, nothing blocking.

Bit by bit
stone by stone
the day will come
we'll sunset drone

build/integration/composer.json
```diff
@@ -1,11 +1,16 @@
...
-    "dms/phpunit-arraysubset-asserts": "^0.5",
-    "behat/behat": "~3.13.0",
-    "guzzlehttp/guzzle": "7.8.1",
-    "jarnaiz/behat-junit-formatter": "^1.3",
-    "sabre/dav": "4.6.0",
-    "symfony/event-dispatcher": "~5.3"
-  }
+	"config": {
+		"platform": {
+			"php": "8.0.2"
```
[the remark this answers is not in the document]
nickvergessen: > Is it important to define a PHP version here? I fear this will increase maintenance work when the minimum PHP version is raised for no good reason.

Otherwise some dependencies can not be updated, there was a sever bug which made symfony container require it iirc. Same as https://github.com/nextcloud/3rdparty/blob/master/composer.json#L11 basically.

And well, I can help on the bump next time around :)

build/integration/features/ratelimiting.feature
nickvergessen: yeah, it was previously "force enabled" by the drone script:
https://github.com/nextcloud/server/pull/41139/files#diff-b54b39f1afced2465e1f3641db9d5bbf4f3a7fcf890996dfedd3c197bcb7f8c7L739
But that meant every developer would have to enable the app manually before running the test.
So we can also just force enable it on execution, to easy developer experience

.github/workflows/integration-sqlite.yml
```diff
@@ -0,0 +1,171 @@
...
+          - 'sharees_features'
+          - 'sharing_features'
+          - 'videoverification_features'
+
+        php-versions: ['8.2']
+        spreed-versions: ['main']
+
+    services:
+      redis:
+        image: redis
```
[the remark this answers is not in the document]
nickvergessen: Done

AndyScherzinger (approved):

susnux (approved):
:rocket:

ArtificialOwl (approved):

szaimen:
🎉🎉🎉🎉🎉

PR #41155, opened 2023-10-27
"Share with" wording fix in new share panel

Jerome-Herbinet:
## Summary

- Removing space ; so we get "Share with" instead of "Share with " and adding space after translation call (because, IMO a space at the end of a wording is not a good idea)
- Replacing "Share" with "Shared", because, IMO, it's more relevant and consistent (isn't it ?)

**Note :** I require a syntax checking concerning the moved space character. 

## Checklist


apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -292,7 +292,7 @@ export default {
 
 	computed: {
 		title() {
-			let title = t('files_sharing', 'Share with ')
+			let title = t('files_sharing', 'Shared with') + ' ';
```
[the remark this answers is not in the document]
Jerome-Herbinet: It makes sense. Let's wait for other reviewers' opinion.

Fenn-CS, line 295: + 1 on taking out the space.

As for the word `Shared` it would have to be dynamic, because that would not be true for new shares.

nfebe (commented):
Shared would not be true for new shares. It should be `Share` when user has not saved share and `Shared`  otherwise

apps/files_sharing/src/views/SharingDetailsTab.vue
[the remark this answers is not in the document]
susnux: cc @rakekniven

Jerome-Herbinet: You are right but in this case, I won't have the required skills to do such changes ; my skills scope is CSS design and wordings calls change.

jancborchardt: @Fenn-CS could you adjust it so it works in both cases? :)

Jerome-Herbinet:
> Can someone please add a screenshot here? I was unable to find the occurence of given string in web UI using NC v27.1.3 .

@rakekniven :

Fenn-CS:
@jancborchardt So what is the conclusion? So I can implement it.

susnux:
Sorry to tackle your PR but I pushed a commit to show what I mean, because this will allow to translate the sentence independent of sentence order of the language. Feels free to drop that commit if not happy with it :)

Jerome-Herbinet:
If I understand well, this PR hasn't been merged into Nextcloud 29 ; what was the problem ?

Jerome-Herbinet:
@jancborchardt it's fixed in NC 29 (maybe since 28 and maybe in NC 30 I hope = no regression) so I close my PR which is useless now.

jancborchardt:
@Jerome-Herbinet yep, sorry that it didn’t come with your pull request, but let’s be happy it was fixed either way. :)

PR #41193, opened 2023-10-30
ci: Turn off builds for pushes and merges to master/stable* and cancel parallel executions

nickvergessen:


provokateurin (approved):

SystemKeeper (approved):

ChristophWurst (approved):

.github/workflows/node-tests.yml
```diff
@@ -2,10 +2,10 @@ name: Node tests
 
 on:
   pull_request:
-  push:
```
susnux, line 5: This will disable code coverage

susnux (commented):
This is needed and good, but we should keep node-tests and (later) phpunit tests run on push to master & stable* because otherwise the project code coverage can not be calculated.

There will only be partially values for PRs and if they only touch JS then only the JS coverage will be calculated.

But I think we should also calculate a overall project coverage to see whats missing to be tested, therefor we should run tests that count into code coverage on pushes to, at least, master.

.github/workflows/node-tests.yml
SystemKeeper: Could we do this once a day maybe or does it need to happen on push?

nickvergessen: reverted

nickvergessen:
> But I think we should also calculate a overall project coverage to see whats missing to be tested, therefor we should run tests that count into code coverage on pushes to, at least, master.

Reverted for now.

We could also just do it every night. I guess no one will look at it "live", so that should be totally fine and would help to reduce the run times even more.

susnux:
> We could also just do it every night. I guess no one will look at it "live", so that should be totally fine and would help to reduce the run times even more.

This sounds good!

st3iny (approved):

.github/workflows/node-tests.yml
susnux: > Could we do this once a day maybe or does it need to happen on push?

No it just needs to be done on the correct branch

PR #41199, opened 2023-10-30
fix(theming): Add accessible information to app order settings

susnux:
## Summary
1. Add note card with information to `aria-details` of the order selectors instead of the list (has to be an interactive element)
2. Add screen reader information while reordering to let the user know at which position the app currently is sorted.

## Checklist


apps/theming/src/components/AppOrderSelector.vue
```diff
@@ -1,20 +1,33 @@
...
+			class="hidden-visually"
+			role="status">
+			{{ statusInfo }}
+		</div>
+		<ol ref="listElement" data-cy-app-order class="order-selector">
+			<AppOrderSelectorElement v-for="app,index in appList"
+				:key="`${app.id}${renderCount}`"
+				:app="app"
+				:aria-details="ariaDetails"
+				:aria-described-by="statusInfoId"
```
JuliaKirschenheuter, line 14: should be `aria-describedby`, https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby

JuliaKirschenheuter (changes requested):
I think the rest is good, thank you!

apps/theming/src/components/AppOrderSelector.vue
susnux: :facepalm:  sorry

JuliaKirschenheuter (approved):

Pytal (approved):

ShGKme (approved):

PR #41201, opened 2023-10-30
remove comments tab from files

sorbaugh:
## Summary

This is the the second part of a two part Pull Request to move code from the comments app found in server to the activity app.

Here is the other PR: https://github.com/nextcloud/activity/pull/1387

## Checklist


apps/comments/src/comments-app.js
```diff
@@ -20,13 +20,7 @@
  *
  */
 
-import CommentsInstance from './services/CommentsInstance.js'
```
nickvergessen, line 23: The comments feature is actually part of the public API (3 years deprecation required?):
https://github.com/nextcloud/server/blob/4d2217073acf6ed844bd006c102e403e7978e813/lib/private/Comments/Manager.php#L1615-L1623

It allows apps to include comments with:
* 2 lines of PHP:
    *  https://github.com/nextcloud/announcementcenter/blob/main/lib/Controller/PageController.php#L103
    * https://github.com/nextcloud/announcementcenter/blob/main/lib/Listener/CommentsEntityListener.php#L57
* 6 lines of JS/Vue:
    * https://github.com/nextcloud/announcementcenter/blob/main/src/App.vue#L148-L154

---

If we decide not to keep this (which I think is a bummer, but also it's not good documented atm and so only used by 1 app with this way, other apps reimplemented comments manually) at least the comments manager also needs adjusting to not import a non-existing file and also help developers realize the feature was removed without prior notice.

nickvergessen (dismissed):

nickvergessen:
Setting https://github.com/nextcloud/server/labels/pending%20documentation because it's needed in case the feature is being deprecated or removed directly.

apps/comments/src/comments-app.js
skjnldsv: > 3 years deprecation required?

It's an OCA api, it doesn't follow that rule :)
(I used OCA for that exact reason, because this is not an official API)

nickvergessen: https://github.com/nextcloud/server/blob/77355a8a79898a03d4f45e853e3d99c9a633509b/lib/public/Comments/ICommentsManager.php#L478-L483

PHP is `OCP` with "Load the Comments app into the page"
As said, we can violate it, make the function no-op / log a fatal error and continue, so app developers have a better trace.
But then it should at least be documented in the "Removed API" section of https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_28.html#id4

sorbaugh:
Thanks for pointing it out! For context, the comments feature is being moved to the activity repo, so this is one of two pull requests.
https://github.com/nextcloud/activity/pull/1387

lib/private/Comments/Manager.php
```diff
@@ -1613,13 +1613,11 @@ private function sendEvent($eventType, IComment $comment) {
 	}
 
 	/**
-	 * Load the Comments app into the page
+	 * Function is no-op because comments feature was moved
 	 *
 	 * @since 21.0.0
 	 */
 	public function load(): void {
-		$this->initialStateService->provideInitialState('comments', 'max-message-length', IComment::MAX_MESSAGE_LENGTH);
```
nickvergessen, line 1621:
```
		$this->logger->critical('Method ' . __METHOD__ . ' is no-op because comments feature was moved');
```

lib/private/Comments/Manager.php
```diff
@@ -1613,13 +1613,11 @@ private function sendEvent($eventType, IComment $comment) {
 	}
 
 	/**
-	 * Load the Comments app into the page
+	 * Function is no-op because comments feature was moved
```
nickvergessen, line 1616:
```suggestion
	 * Load the Comments app into the page
```

lib/private/Comments/Manager.php
```diff
@@ -1613,13 +1613,11 @@ private function sendEvent($eventType, IComment $comment) {
...
 	 *
 	 * @since 21.0.0
```
nickvergessen, line 1618:
```suggestion
	 * @since 21.0.0
	 * @deprecated 28.0.0 - Function is no-op because comments feature was moved to the activity app
```

nickvergessen: This change must also be applied on `lib/public/Comments/ICommentsManager.php`

lib/private/Comments/Manager.php
sorbaugh: Added :)

lib/private/Comments/Manager.php
sorbaugh: Adjusted :)

nickvergessen: You removed the useful description instead of adding the deprecation

artonge (approved):

apps/comments/src/comments.js
artonge, line 1: Is this file still needed?

apps/comments/src/activitytabviewplugin.js
artonge, line 1: What was the use of this file?

artonge (changes requested):
Actually, most of this need to be kept as:

> if activity app is disabled and comments app enabled:
        show comments tab as-is

susnux:
closing in favor of https://github.com/nextcloud/server/pull/41491
(for reuse code and not duplicate comments code to activity app)

PR #41202, opened 2023-10-30
fix(dav): multiple fixes in usage of webdav library

pulsejet:
1. Refresh token on update
2. Fix some very weird imports
3. Patch fetch instead of request to prevent accessing impl details


apps/comments/src/services/GetComments.ts
```diff
@@ -23,8 +23,8 @@
 import { parseXML, type DAVResult, type FileStat } from 'webdav'
 
 // https://github.com/perry-mitchell/webdav-client/issues/339
-import { processResponsePayload } from '../../../../node_modules/webdav/dist/node/response.js'
-import { prepareFileFromProps } from '../../../../node_modules/webdav/dist/node/tools/dav.js'
+import { processResponsePayload } from 'webdav/dist/node/response.js'
+import { prepareFileFromProps } from 'webdav/dist/node/tools/dav.js'
```
pulsejet, line 27: If we can get rid of these somehow then we can probably also lose the node polyfills (of course `@nextcloud/files` will also have to be fixed up)

pulsejet: https://github.com/nextcloud-libraries/nextcloud-files/pull/821

skjnldsv: Yeah, we had to do that because of the export, if it's fixed, please also remove the comment line above :+1:

skjnldsv:
> 3. Patch fetch instead of request to prevent accessing impl details

Can you clarify what is this about? :)

apps/comments/src/services/GetComments.ts
pulsejet: Not yet. I'll make a separate PR if this can be fixed

pulsejet:
> > 3. Patch fetch instead of request to prevent accessing impl details
> 
> Can you clarify what is this about? :)

To patch `request`, we need
```ts
import { request } from 'webdav/dist/node/request.js'
```

This is an implementation detail of the webdav library and also needs the node polyfills. Instead we can just patch `fetch` which is equivalent since `request` just calls fetch with the same options (and we're just changing the method anyway).

BTW, fixed the style to align with nextcloud-files

artonge (approved):

skjnldsv (approved):

pulsejet:
I want to point out that things _might_ be broken without this patch in 28. When the refresh token changes, the WebDAV clients are never updated in RC.1

pulsejet:
CI failure seems unrelated

PR #41211, opened 2023-10-31
fix(TextToImage): Consistently use the right method to get the preferred providers

marcelklehr:
## Checklist

### DIFF
--- a/lib/private/TextToImage/Manager.php
+++ b/lib/private/TextToImage/Manager.php
@@ -113,20 +113,7 @@ class Manager implements IManager {
 		if (!$this->hasProviders()) {
 			throw new PreConditionNotMetException('No text to image provider is installed that can handle this task');
 		}
-		$providers = $this->getProviders();
-
-		$json = $this->config->getAppValue('core', 'ai.text2image_provider', '');
-		if ($json !== '') {
-			try {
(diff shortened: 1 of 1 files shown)


MB-Finski (approved):

lib/private/TextToImage/Manager.php
```diff
@@ -232,22 +225,9 @@ public function runOrScheduleTask(Task $task) : void {
...
-					$providers = [$provider];
-				}
-			} catch (\JsonException $e) {
-				$this->logger->warning('Failed to decode Text2Image setting `ai.text2image_provider`', ['exception' => $e]);
-			}
-		}
+		$providers = $this->getPreferredProviders();
 		$maxExecutionTime = (int) ini_get('max_execution_time');
-		// Offload the tttttttask to a background job if the expected runtime of the likely provider is longer than 80% of our max execution time
+		// Offload the task to a background job if the expected runtime of the likely provider is longer than 80% of our max execution time
```
julien-nc, line 230: I like the original string better :grin:

julien-nc (approved):

lib/private/TextToImage/Manager.php
marcelklehr: tssss :P

PR #41214, opened 2023-10-31
fix(install): Make installing more verbose

nickvergessen:
## Before

Not really very verbose
```
$ occ maintenance:install --verbose
Nextcloud was successfully installed
```

## After

```
$ occ maintenance:install --verbose
2023-10-31T11:00:35+00:00 Configuring database
2023-10-31T11:00:35+00:00 Run server migrations
2023-10-31T11:00:35+00:00 Migrating schema only
2023-10-31T11:00:35+00:00 - Reading 13000Date20170705121758
2023-10-31T11:00:35+00:00 - Reading 13000Date20170718121200
2023-10-31T11:00:35+00:00 - Reading 13000Date20170814074715
2023-10-31T11:00:35+00:00 - Reading 13000Date20170919121250
2023-10-31T11:00:35+00:00 - Reading 13000Date20170926101637
2023-10-31T11:00:35+00:00 - Reading 14000Date20180129121024
2023-10-31T11:00:35+00:00 - Reading 14000Date20180404140050
2023-10-31T11:00:35+00:00 - Reading 14000Date20180516101403
2023-10-31T11:00:35+00:00 - Reading 14000Date20180518120534
2023-10-31T11:00:35+00:00 - Reading 14000Date20180522074438
2023-10-31T11:00:35+00:00 - Reading 14000Date20180626223656
2023-10-31T11:00:35+00:00 - Reading 14000Date20180710092004
2023-10-31T11:00:35+00:00 - Reading 14000Date20180712153140
2023-10-31T11:00:35+00:00 - Reading 15000Date20180926101451
2023-10-31T11:00:35+00:00 - Reading 15000Date20181015062942
2023-10-31T11:00:35+00:00 - Reading 15000Date20181029084625
2023-10-31T11:00:35+00:00 - Reading 16000Date20190207141427
2023-10-31T11:00:35+00:00 - Reading 16000Date20190212081545
2023-10-31T11:00:35+00:00 - Reading 16000Date20190427105638
2023-10-31T11:00:35+00:00 - Reading 16000Date20190428150708
2023-10-31T11:00:35+00:00 - Reading 17000Date20190514105811
2023-10-31T11:00:35+00:00 - Reading 18000Date20190920085628
2023-10-31T11:00:35+00:00 - Reading 18000Date20191014105105
2023-10-31T11:00:35+00:00 - Reading 18000Date20191204114856
2023-10-31T11:00:35+00:00 - Reading 19000Date20200211083441
[...]

PR #41215, opened 2023-10-31
[stable27] fix(install): Make installing more verbose

nickvergessen:
Backport of #41214

### DIFF
--- a/lib/public/Migration/IOutput.php
+++ b/lib/public/Migration/IOutput.php
@@ -32,6 +32,13 @@ interface IOutput {
 	/**
 	 * @param string $message
 	 * @return void
+	 * @since 28.0.0
+	 */
+	public function debug(string $message): void;
+
+	/**
+	 * @param string $message
+	 * @return void
 	 * @since 9.1.0
 	 */
 	public function info($message);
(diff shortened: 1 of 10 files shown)


lib/public/Migration/IOutput.php
```diff
@@ -29,6 +29,13 @@
  * @since 9.1.0
  */
 interface IOutput {
+	/**
+	 * @param string $message
+	 * @return void
+	 * @since 28.0.0
```
ChristophWurst, line 35: :crying_cat_face:

nickvergessen: Why are you sad, cat? :cuddle:

I guess we are at the point again where a differentiation between "interfaces to help apps what to expect" (adding things would be okay) vs. "interfaces that apps can implement to overwrite services" (adding things is less okay) would be helpful, and this here is the first case.

You want me to deprecate this interface and add a new one instead?

nickvergessen: Ah, it's the backport. Well this seems to be how we roll with this topic atm:
https://github.com/nextcloud/server/blob/stable26/lib/public/OCM/IOCMDiscoveryService.php#L45

But can also set it to 27.1.4

kesselb: Please 27.1.4 :pray:

skjnldsv (approved):

kesselb (approved):

PR #41216, opened 2023-10-31
add descriptions for background pictures

### LINKED ISSUE #40689: [BITV]: Replace `aria-label`'s of background buttons with another name
This is accessibility issue.

For now background buttons have `aria-label`'s which are not really understandable, like "Soft floral (Hannah MacLean, CC0)". The `aria-label` have to be like "Background with abstract ..."

Each name have to unique.

Screenshot from 2023-09-28 18-51-42

szaimen:
Close https://github.com/nextcloud/server/issues/40689

I added translations for the attributions and added descriptions for each foto.


szaimen:
```
<html>
<body>
<!--StartFragment-->

There was 1 error:
--
218 |  
219 | 1) OCA\Theming\Tests\ThemingDefaultsTest::testGetColorPrimaryWithCustomBackground
220 | Undefined array key 2
221 |  
222 | /drone/src/apps/theming/tests/ThemingDefaultsTest.php:478

<!--EndFragment-->
</body>
</html>
```

apps/theming/lib/Service/BackgroundService.php
```diff
@@ -257,4 +158,129 @@ private function getAppDataFolder(): ISimpleFolder {
 			return $rootFolder->newFolder($this->userId);
 		}
 	}
+
+	public function getShippedBackgrounds(): array {
+		return [
+			'hannah-maclean-soft-floral.jpg' => [
+				'attribution' => $this->l10n->t('Soft floral (Hannah MacLean, CC0)'),
+				'description' => $this->l10n->t('Abstract picture in yellow and white color whith a flower on it'),
```
szaimen, line 166: For reviewers: Please review each description carefully. Does the text look fine?

apps/theming/lib/Themes/CommonThemeTrait.php
```diff
@@ -31,6 +31,13 @@
 
 trait CommonThemeTrait {
 	public Util $util;
+	private BackgroundService $backgroundService;
+
+	public function __construct(
+		BackgroundService $backgroundService,
+	) {
+		$this->backgroundService = $backgroundService;
+	}
```
szaimen, line 40: apparently this is not working currently as it is a trait and not a class...

szaimen: I got `"Typed property OCA\\\\\\\\Theming\\\\\\\\Themes\\\\\\\\DefaultTheme::$backgroundService must not be accessed before initialization in file '/var/www/nextcloud/apps/theming/lib/Themes/CommonThemeTrait.php`

szaimen: So apparently the constructor does not work here. Question then how to make it work at all?

ChristophWurst: Pass the dependency as an argument of the methods that need it. Example: https://github.com/nextcloud/server/blob/master/lib/public/AppFramework/Db/TTransactional.php.

The concept of DI and properties for dependencies doesn't really fit traits.

apps/theming/lib/Themes/DefaultTheme.php
```diff
@@ -57,7 +58,8 @@ public function __construct(Util $util,
 								ImageManager $imageManager,
 								IConfig $config,
 								IL10N $l,
-								IAppManager $appManager) {
+								IAppManager $appManager,
+								BackgroundService $backgroundService) {
```
szaimen, line 62: So this is now breaking many tests (34 afaics). @ChristophWurst Any other idea how I can do this without having to refactor everything? Adding backgroundservice to this constructor at least does not seem to work...

ChristophWurst: I don't have any other ideas.

Can you adjust/fix the tests?

szaimen: What if I would move the commonthemetrait? Do you think this would make the logic easier?

ChristophWurst: Try

szaimen: Lets see https://github.com/nextcloud/server/pull/41235

ChristophWurst: Could I ask you to try your changes locally? Using CI for trial and error is really expensive and slows down real builds.

szaimen:
Superceded by https://github.com/nextcloud/server/pull/41256

PR #41234, opened 2023-11-02
fix(OCM): Fix OCM provider discovery for servers without beautiful URLs

nickvergessen:
* Followup to https://github.com/nextcloud/server/pull/39574

## Resolves

* This allows to do federation with instances again that do not support "beautiful" URLs at all.

## Checklist

### DIFF
--- a/lib/private/OCM/OCMDiscoveryService.php
+++ b/lib/private/OCM/OCMDiscoveryService.php
@@ -69,6 +69,9 @@ class OCMDiscoveryService implements IOCMDiscoveryService {
 	 */
 	public function discover(string $remote, bool $skipCache = false): IOCMProvider {
 		$remote = rtrim($remote, '/');
+		if (str_ends_with($remote, '/index.php')) {
+			$remote = substr($remote, 0, -10);
+		}
 
 		if (!$skipCache) {
 			try {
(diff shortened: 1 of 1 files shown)


kesselb:
I'm not familiar with ocm :see_no_evil: 

1) Isn't that against the spec to query index.php/ocm-provider
2) Woudln't that break if the remote instance is not Nextcloud

nickvergessen:
> Isn't that against the spec to query index.php/ocm-provider

If it is, we need to recreate the folder with it's index.php like it was before #39574 

> Wouldn't that break if the remote instance is not Nextcloud

Right. Quick fix suggestion: go back to "first query without `index.php` and if that gives a 404 we try with `index.php`, or implement the folder again.

level420:
@nickvergessen could this also solve the problem that since v 27.1.x (now having 27.1.3) I'm getting warnings about unresolvable "/ocm-provider/" in the  security and warnings section?

My instances are running without pretty urls enabled, and `https://myinstance.com/index.php/ocm-provider/` works, but `https://myinstance.com/ocm-provider/` not.

I'm running a docker stack using the current nextcloud docker image `nextcloud:27.1-fpm` which is behind an apache httpd proxy.

mickenordin:
Since Nextcloud now supports `.well-known/ocm` this PR can be safely closed.

susnux:
> Since Nextcloud now supports .well-known/ocm this PR can be safely closed.

agree :white_check_mark:

PR #41250, opened 2023-11-02
fix(cypress): Add timeout for waiting on Nextcloud server

susnux:
## Summary

Do not wast CI time if docker setup of the Nextcloud server went wrong.

## Checklist

### DIFF
--- a/cypress/dockerNode.ts
+++ b/cypress/dockerNode.ts
@@ -233,7 +233,15 @@ export const getContainerIP = async function(
 // https://github.com/cypress-io/cypress/issues/22676
 export const waitOnNextcloud = async function(ip: string) {
 	console.log('├─ Waiting for Nextcloud to be ready... ⏳')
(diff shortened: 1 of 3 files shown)


ChristophWurst (approved):

SystemKeeper (approved):

susnux:
Timeout after 2m? Something fishy here, I will investigate.

nickvergessen (approved):

susnux:
Looks like the runners were too crowded, but nothing I can do here.
So I reverted all not reviewed changed (only the timeout is added for now to prevent CI being stuck).

I created a follow up that will use GitHub workflow service for the Nextcloud server, which then will save (in best case) ~2 minutes. (best case = when CI is fast, in cases like midday it might save even more CI time).

The follow up: https://github.com/nextcloud/server/pull/41252

PR #41252, opened 2023-11-03
[cypress] Use GitHub action service for Nextcloud server

susnux:
## Summary

Follow up on https://github.com/nextcloud/server/pull/41250 

This changes the Cypress workflow to not start a docker container from Cypress but use a GitHub workflow service as the Nextcloud container.
This saves ~2 minutes CI time per Cypress runner (currently 3).

Also the service is up when the workflow starts meaning cypress does not have to wait for it and should no longer time out.

## Checklist

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -110,10 +110,12 @@ jobs:
       - name: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }} cypress tests
         uses: cypress-io/github-action@ebe8b24c4428922d0f793a5c4c96853a633180e3 # v6.6.0
         with:
-          ci-build-id: ${{ github.sha }}-${{ github.run_number }}
+          # Unavailable without cypress cloud
+          # ci-build-id: ${{ github.sha }}-${{ github.run_number }}
           component: ${{ matrix.containers == 'component' }}
-          group: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }}
-          tag: ${{ github.event_name }}
+          # Unavailable without cypress cloud
+          # group: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }}
+          # tag: ${{ github.event_name }}
         env:
           # Use github service for server
           NEXTCLOUD_HOST: localhost
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -84,6 +84,13 @@ jobs:
 
     name: runner ${{ matrix.containers }}
 
+    services:
+      nextcloud:
+        image: ghcr.io/nextcloud/continuous-integration-shallow-server
+        options: --name nextcloud
+        ports:
+          - 80:80
+
     steps:
       - name: Restore context
         uses: buildjet/cache/restore@e376f15c6ec6dc595375c78633174c7e5f92dc0e # v3
@@ -103,12 +110,14 @@ jobs:
       - name: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }} cypress tests
         uses: cypress-io/github-action@ebe8b24c4428922d0f793a5c4c96853a633180e3 # v6.6.0
         with:
+          ci-build-id: ${{ github.sha }}-${{ github.run_number }}
           component: ${{ matrix.containers == 'component' }}
-          group: ${{ matrix.use-cypress-cloud && matrix.containers == 'component' && 'Run component' || matrix.use-cypress-cloud && 'Run E2E' || '' }}
-          # cypress env
-          ci-build-id: ${{ matrix.use-cypress-cloud && format('{0}-{1}', github.sha, github.run_number) || '' }}
-          tag: ${{ matrix.use-cypress-cloud && github.event_name || '' }}
+          group: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }}
+          tag: ${{ github.event_name }}
         env:
+          # Use github service for server
+          NEXTCLOUD_HOST: localhost
+          NEXTCLOUD_CONTAINER: nextcloud
           # Needs to be prefixed with CYPRESS_
           CYPRESS_BRANCH: ${{ env.BRANCH }}
           # https://github.com/cypress-io/github-action/issues/124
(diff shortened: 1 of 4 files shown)


.github/workflows/cypress.yml
```diff
@@ -84,6 +84,13 @@ jobs:
 
     name: runner ${{ matrix.containers }}
 
+    services:
+      nextcloud:
+        image: ghcr.io/nextcloud/continuous-integration-shallow-server
+        options: --name nextcloud
+        ports:
+          - 80:80
```
skjnldsv, line 92: That means component will also have a server started

.github/workflows/cypress.yml
```diff
@@ -104,11 +111,10 @@ jobs:
         uses: cypress-io/github-action@59810ebfa5a5ac6fcfdcfdf036d1cd4d083a88f2 # v6.5.0
         with:
           component: ${{ matrix.containers == 'component' }}
-          group: ${{ matrix.use-cypress-cloud && matrix.containers == 'component' && 'Run component' || matrix.use-cypress-cloud && 'Run E2E' || '' }}
-          # cypress env
-          ci-build-id: ${{ matrix.use-cypress-cloud && format('{0}-{1}', github.sha, github.run_number) || '' }}
-          tag: ${{ matrix.use-cypress-cloud && github.event_name || '' }}
```
skjnldsv, line 110: This is not related

skjnldsv:
> This saves ~2 minutes CI time per Cypress runner (currently 3).

Do you know why? I mean, both are running docker? This look like a crazy difference :exploding_head:

.github/workflows/cypress.yml
susnux: But invalid as there is no `matrix.use-cypress-cloud`

susnux:
> Do you know why? I mean, both are running docker? This look like a crazy difference 🤯

I guess because services are cached and created on the beginn while manually starting docker happens on step evaluation

.github/workflows/cypress.yml
susnux: Yes, sadly github does not allow conditional services

.github/workflows/cypress.yml
skjnldsv: Right, but the group, ci build and tag are still mandatory and relevant to have

susnux: Well as the variable does not exist all of them are set to ''
So it makes no difference

max-nextcloud (approved):
Interesting approach.

I'd be happy to save 2 min per cypress runner.

I'm wondering if there are any downsides to this but can't think of any. The only thing that comes to my mind is that we may need to maintain multiple images at some point so once this lands on stable branches they still have a container that works for them. But i guess we'll need the images anyway and can just configure them in the cypress.yml accordingly.

.github/workflows/cypress.yml
skjnldsv: Doesn't mean we should dro pit instead of fixing it :thinking: 

- https://github.com/nextcloud/server/pull/40736 changed the behaviour
- https://github.com/nextcloud/server/pull/41039 changed it too but broke the group/tag 

I'd suggest we revert back to

```suggestion
          group: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }}
          # cypress env
          ci-build-id: ${{ github.sha }}-${{ github.run_number }}
          tag: ${{ github.event_name }}
```

susnux: makes sense

skjnldsv (approved):

.github/workflows/cypress.yml
susnux: Well this breaks the CI, as those flags are only allowed when using the cypress cloud which we do not.

ref: https://docs.cypress.io/guides/references/error-messages#The---ci-build-id---group---tag---parallel-or---auto-cancel-after-failures-flags-can-only-be-used-when-recording

skjnldsv: Time for our custom cloud :cry:

.github/workflows/cypress.yml
```diff
@@ -103,12 +110,14 @@ jobs:
       - name: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }} cypress tests
         uses: cypress-io/github-action@ebe8b24c4428922d0f793a5c4c96853a633180e3 # v6.6.0
         with:
+          ci-build-id: ${{ github.sha }}-${{ github.run_number }}
```
skjnldsv, line 113:
```suggestion
          # Unavailable without cypress cloud
          # ci-build-id: ${{ github.sha }}-${{ github.run_number }}
```

.github/workflows/cypress.yml
```diff
@@ -103,12 +110,14 @@ jobs:
...
           component: ${{ matrix.containers == 'component' }}
-          group: ${{ matrix.use-cypress-cloud && matrix.containers == 'component' && 'Run component' || matrix.use-cypress-cloud && 'Run E2E' || '' }}
-          # cypress env
-          ci-build-id: ${{ matrix.use-cypress-cloud && format('{0}-{1}', github.sha, github.run_number) || '' }}
-          tag: ${{ matrix.use-cypress-cloud && github.event_name || '' }}
+          group: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }}
+          tag: ${{ github.event_name }}
```
skjnldsv, line 116:
```suggestion
          # Unavailable without cypress cloud
          # group: Run ${{ matrix.containers == 'component' && 'component' || 'E2E' }}
          # tag: ${{ github.event_name }}
```

.github/workflows/cypress.yml
skjnldsv: https://github.com/nextcloud/server/commits/39158c0a52c478df989c955603c643f3905a7343

skjnldsv:
Seems like it's failing much more randomly with that PR  :thinking:

skjnldsv:
`navigation-bar-settings.cy.ts` is really failing

skjnldsv:
`Error response from daemon: No such container: nextcloud-cypress-tests-server`

PR #41256, opened 2023-11-03
add descriptions for background pictures

### LINKED ISSUE #40689: [BITV]: Replace `aria-label`'s of background buttons with another name

szaimen:
Close https://github.com/nextcloud/server/issues/40689


apps/theming/lib/Service/BackgroundService.php
```diff
@@ -57,101 +57,120 @@ class BackgroundService {
 	public const SHIPPED_BACKGROUNDS = [
 		'hannah-maclean-soft-floral.jpg' => [
 			'attribution' => 'Soft floral (Hannah MacLean, CC0)',
+			'description' => 'Abstract picture in yellow and white color whith a flower on it',
```
szaimen, line 60: Please review these descriptions carefully. Are they good or should they rather be something like `Abstract background picture ...` ? (So should I add the word `background` everywhere?

JuliaKirschenheuter: I would vote for `Abstract background picture ...`

szaimen: okay adding the word background then everywhere

JuliaKirschenheuter (approved):
Looks good, thank you!

apps/theming/lib/Service/BackgroundService.php
szaimen: done

susnux: Well I would have said the opposite as background image is no description of what you see on it.
Moreover you should avoid terms like "picture of" or "image of" as it is already clear that it is an image. 

But I am pretty bad at describing images so fine for me - better than before:)

susnux (approved):

PR #41264, opened 2023-11-03
[stable26] Show proper warning on 0 quota

juliusknorr:
- fix(files): url used to retrive storage stats (backport of https://github.com/nextcloud/server/pull/39588)
- fix: Apply directory permissions for new api response

## Steps to reproduce

- Have 0 quota configured
- Open the file list

## before

The file create menu was shown

## After

A proper info is shown as it was on Nextcloud 25

Screenshot 2023-11-03 at 11 57 28

### DIFF
--- a/apps/files/js/files.js
+++ b/apps/files/js/files.js
@@ -25,7 +25,7 @@
 				state.call.abort();
 			}
 			state.dir = currentDir;
-			state.call = $.getJSON(OC.generateUrl('apps/files/ajax/getstoragestats?dir={dir}', {
+			state.call = $.getJSON(OC.generateUrl('apps/files/api/v1/stats?dir={dir}', {
 				dir: currentDir,
 			}), function(response) {
 				state.dir = null;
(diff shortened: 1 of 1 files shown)


apps/files/js/files.js
```diff
@@ -70,8 +70,13 @@
 			if (response === undefined) {
 				return;
 			}
+
+			if (response.data !== undefined && response.data.free !== undefined) {
+				$('#free_space').val(response.data.free);
+				OCA.Files.App.fileList._updateDirectoryPermissions();
+			}
+
 			if (response.data !== undefined && response.data.uploadMaxFilesize !== undefined) {
```
juliushaertl, line 79: The following parts still don't work with the new API so there might be further bugs but this PR at least fixes the reported one.

@skjnldsv I assume this is handled separately anways in the files to vue rewrite?

juliushaertl:
f63fa8cac616484bf0bc6d85b429740adfe0c801 would be worth to be ported to 27 as well

apps/files/js/files.js
skjnldsv: yes, this code is gone on nc28

solracsf:
Why the message is not `You don't have either permission or quota to upload or create files here` ?

juliushaertl:
Something to discuss separately in an issue I'd say since we would want to change this in the master branch first anyways.

kesselb (approved):
Tested and works :+1:

blizzz (approved):

PR #41270, opened 2023-11-03
Fix styles for input fields on focus state

### LINKED ISSUE #41267: [BITV]: Contrast of focused "Search" and "Search contacts" is not enough
This is accessibility issue.

The contrast between focused and unfocused state on "Search" and "Search contacts" have to be 3:1. Now: unfocused : grey, focused: blue. That is not enough. See https://github.com/nextcloud/server/issues/40711

Screenshot from 2023-10-30 11-46-31
Screenshot from 2023-10-30 11-46-24

JuliaKirschenheuter:
* Resolves: https://github.com/nextcloud/server/issues/41267

### 🖼️ Screenshots

| image | Screenshot from 2023-11-03 15-03-42 |-------------|---------------------------------------------------------------|
| image | Screenshot from 2023-11-03 15-02-15

## Checklist


core/src/views/ContactsMenu.vue
```diff
@@ -170,6 +174,12 @@ export default {
 			width: 100%;
 			height: 34px;
 			margin: 8px 0;
+
+			&:focus,
+			&:active {
```
szaimen, line 179: what about focus-visible?

susnux (approved):

core/src/views/UnifiedSearch.vue
```diff
@@ -762,6 +762,12 @@ $input-padding: 6px;
 
 	&__form-input {
 		margin: 0 !important;
+		&:focus,
+		&:focus-visible,
+		&:active {
+			border-color: 2px solid var(--color-main-text) !important;
+			box-shadow: 0 0 0 2px var(--color-main-background) !important;
+		}
```
ShGKme, line 770: IMO, if possible, we should migrate these components to `NcTextField` so we won't need to duplicate all fixes here.

JuliaKirschenheuter: I would like to make this improvement after all other important issues are solved. Is ok for you?

ShGKme (approved):

marcoambrosini (approved):

PR #41271, opened 2023-11-03
enh(TextProcessing): Add two new provider interfaces

marcelklehr:
## Summary
Since we added the TextProcessing API in 27.1 some shortcomings have presented themselves, which I seek to improve with this PR.

Additions in this PR:

- allow providers to obtain current task's userId
- allow providers to expose average task runtime
- IManager now also exposes a runOrSchedule() method which takes into account this runtime estimation
- the schedule OCS API endpoint now uses runOrSchedule to avoid always scheduling a background job which has a max-delay of 5min even if the provider would be done in a few seconds, like the OpenAI provider for example.

## Todo

- [x] Check and possible extend tests

## Checklist


core/Migrations/Version28000Date20231103104802.php
```diff
@@ -0,0 +1,60 @@
...
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+		if ($schema->hasTable('textprocessing_tasks')) {
+			$table = $schema->getTable('textprocessing_tasks');
+
+			$table->addColumn('completion_expected_at', Types::DATETIME, [
+				'notnull' => false,
+			]);
```
julien-nc, line 53: We usually check if the column is there already.

julien-nc (changes requested):

core/Controller/TextProcessingApiController.php
```diff
@@ -102,7 +103,7 @@ public function taskTypes(): DataResponse {
 	 * @param string $appId ID of the app that will execute the task
 	 * @param string $identifier An arbitrary identifier for the task
 	 *
-	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTextProcessingTask}, array{}>|DataResponse<Http::STATUS_BAD_REQUEST|Http::STATUS_PRECONDITION_FAILED, array{message: string}, array{}>
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTextProcessingTask}, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_BAD_REQUEST|Http::STATUS_PRECONDITION_FAILED, array{message: string}, array{}>
```
come-nc, line 106: The new error code is not documented below

core/Controller/TextProcessingApiController.php
```diff
@@ -118,7 +119,11 @@ public function schedule(string $input, string $type, string $appId, string $ide
 			return new DataResponse(['message' => $this->l->t('Requested task type does not exist')], Http::STATUS_BAD_REQUEST);
 		}
 		try {
-			$this->textProcessingManager->scheduleTask($task);
+			try {
+				$this->textProcessingManager->runOrScheduleTask($task);
+			} catch(\RuntimeException) {
```
come-nc, line 124: Same thing as in the other PR, should be a custom exception if it has a special meaning.

lib/private/TextProcessing/Manager.php
```diff
@@ -158,18 +152,25 @@ public function runTask(OCPTask $task): string {
...
 	 * @inheritDoc
-	 * @throws Exception
 	 */
 	public function scheduleTask(OCPTask $task): void {
 		if (!$this->canHandleTask($task)) {
 			throw new PreConditionNotMetException('No LanguageModel provider is installed that can handle this task');
 		}
 		$task->setStatus(OCPTask::STATUS_SCHEDULED);
+		[$provider, ] = $this->getPreferredProviders($task);
+		if ($provider instanceof IProvider2) {
```
come-nc, line 169: But do we know if this task will be run by this specific provider? It seems they are not bound together.

come-nc: Also canUseProvider is not checked

lib/public/TextProcessing/IProvider2.php
```diff
@@ -0,0 +1,48 @@
...
+namespace OCP\TextProcessing;
+
+/**
+ * This interface supersedes IProvider. It allows the system to learn
+ * the provider's expected runtime and lets the provider know which user is running a task
+ * @since 28.0.0
+ * @template T of ITaskType
+ * @template-extends IProvider<T>
+ */
+interface IProvider2 extends IProvider {
```
come-nc, line 36:
```suggestion
interface IProviderWithExpectedRuntime extends IProvider {
```
Or something descriptive like that

core/Controller/TextProcessingApiController.php
marcelklehr: We never document 500 I believe

provokateurin: Nope, those don't need a description. The openapi-extractor (or the CI job that runs it) will tell you if there is something wrong.

lib/public/TextProcessing/IProvider2.php
marcelklehr: IProviderWithExpectedRuntimeAndUserId :thinking: ?

lib/private/TextProcessing/Manager.php
marcelklehr: Good point!

marcelklehr: runTask also calls getPreferredProviders, so unless the settings were changed in the meantime it's fairly certain that we'll get the same provider, I'd say

lib/public/TextProcessing/IProvider2.php
marcelklehr: Personally I'm in favor of IProvider2 because it supersedes IProvider and is not just an addendum

come-nc: But will you ever be able to discard the previous one? For user backends we use addendums, and here you could split into `IExpectedRuntimeAwareProvider` and `IUserAwareProvider`.
But yeah fine with `IProvider2` if you know what you’re doing.

lib/private/TextProcessing/Manager.php
come-nc: Yeah it’s this whole «likely provider» thing which is scaring me out. But I do not have the big picture here, what is it doing?
Why is it not taking the task as parameter to return expected runtime?

marcelklehr: A provider already knows which task type it is used for and I deemed that information good enough to determine the expected runtime. We could pass in the full task with the input text as well, but the basic idea is that we want to know: Will the task be done within 10 seconds or will it take 50 mins?

come-nc: It’s not possible for a provider to register for several task types?

marcelklehr: Nope. One provider per task type

lib/public/TextProcessing/IProvider2.php
marcelklehr: I decided to split IProvider2 into `IProviderWithUserId` and `IProviderWithExpectedRuntime`

lib/public/TextProcessing/IProvider2.php
```diff
@@ -0,0 +1,48 @@
...
+	 * @param ?string $userId the current user's id
+	 * @since 28.0.0
+	 */
+	public function setUserId(?string $userId): string;
+
+	/**
+	 * @return int The expected average runtime of a task in seconds
+	 * @since 28.0.0
+	 */
+	public function getExpectedRuntime(): int;
```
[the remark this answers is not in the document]
marcelklehr: I think Runtime is quite salient, but I'm not particular about this. Any other opinions?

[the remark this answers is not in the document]
julien-nc: What about `getExpectedRunTime` with a capital T for Time?
I'm fine with `getExpectedRunTime` or `getExpectedProcessingDuration`.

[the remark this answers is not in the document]
marcelklehr: It's also called getExpectedRuntime in the TextToImage API already

lib/private/TextProcessing/Manager.php
```diff
@@ -178,6 +172,25 @@ public function scheduleTask(OCPTask $task): void {
 		]);
 	}
 
+	/**
+	 * @inheritDoc
+	 */
+	public function runOrScheduleTask(OCPTask $task) : bool {
```
julien-nc, line 178:
```suggestion
	public function runOrScheduleTask(OCPTask $task): bool {
```
:grin:

lib/private/TextProcessing/Manager.php
```diff
@@ -145,31 +137,33 @@ public function runTask(OCPTask $task): string {
...
 	/**
 	 * @inheritDoc
-	 * @throws Exception
 	 */
 	public function scheduleTask(OCPTask $task): void {
 		if (!$this->canHandleTask($task)) {
 			throw new PreConditionNotMetException('No LanguageModel provider is installed that can handle this task');
 		}
 		$task->setStatus(OCPTask::STATUS_SCHEDULED);
+		[$provider, ] = $this->getPreferredProviders($task);
```
julien-nc, line 161:
```suggestion
		[$provider,] = $this->getPreferredProviders($task);
```

lib/public/TextProcessing/IProviderWithUserId.php
```diff
@@ -0,0 +1,41 @@
...
+ * @since 28.0.0
+ * @template T of ITaskType
+ * @template-extends IProvider<T>
+ */
+interface IProviderWithUserId extends IProvider {
+	/**
+	 * @param ?string $userId the current user's id
+	 * @since 28.0.0
+	 */
+	public function setUserId(?string $userId): string;
```
julien-nc, line 40:
```suggestion
	public function setUserId(?string $userId): void;
```
Right?

julien-nc (changes requested):
LGTM. Small remarks.

julien-nc (approved):

come-nc (approved):

marcelklehr:
Docs are here: https://github.com/nextcloud/documentation/pull/11273

PR #41272, opened 2023-11-03
fix: reduce memory consumption of scans

### LINKED ISSUE #40919: [Bug]: occ files:scan-app-data consumes too much memory

### Bug description

When running occ files:scan-app-data, the memory used by occ quickly rises until it crashes when the 4GB memory limit is exhausted:
```
[14-Oct-2023 19:54:06 UTC] PHP Fatal error:  Allowed memory size of 4294967296 bytes exhausted (tried to allocate 16777224 bytes) in /var/www/html/lib/private/Files/Cache/Cache.php on line 602
```

I would not expect a file scanner to use so much memory. My Nextcloud data directory contains 235k files, of which 17k in appdata and 100k in updater.

### Steps to reproduce

1. Run `occ files:scan-app-data`

### Expected behavior

The memory usage of the file scanner should probably not increase with the number of files

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

27

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Nginx

### Database engine version

MariaDB

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

solracsf:
* Resolves: https://github.com/nextcloud/server/issues/40919

## Summary

`array_filter` is pretty memory intensive, specially in the context of `appdata_` folder where it can have thousands of folders. Construct the array in a loop instead.

The `array_filter` creates a new array and populates it with elements that meet the filter condition. This means it creates a new array in memory to hold the filtered elements, which can be memory-intensive, especially if the original array `$children` is large.

On the other hand, a foreach loop initializes an empty array `$childFolders` and then iterates through the `$children` array, only adding elements to `$childFolders` if they meet the filtering condition. This approach doesn't create an additional array for the filtered elements, so it consumes less memory.

### Before
old

### After
new

### Test system
```
+---------+--------+--------------+
| Folders | Files  | Elapsed time |
+---------+--------+--------------+
| 217688  | 155452 | 00:09:46     |
+---------+--------+--------------+
```

## Checklist

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -599,9 +599,12 @@ class Cache implements ICache {
 			}
 
 			/** @var ICacheEntry[] $childFolders */
-			$childFolders = array_filter($children, function ($child) {
-				return $child->getMimeType() == FileInfo::MIMETYPE_FOLDER;
-			});
+			$childFolders = [];
+			foreach ($children as $child) {
+				if ($child->getMimeType() == FileInfo::MIMETYPE_FOLDER) {
+					$childFolders[] = $child;
+				}
+			}
 			foreach ($childFolders as $folder) {
 				$parentIds[] = $folder->getId();
 				$queue[] = $folder->getId();


ChristophWurst (approved):

kesselb:
To benchmark it myself, I need to upload a good number of images, pre generate all previews, delete the previews again and run occ files:scan-app-data?

come-nc:
> ## Summary
> `array_filter` is pretty memory intensive, specially in the context of `appdata_` folder where it can have thousands of folders. Construct the array in a loop instead.
> 
> The `array_filter` creates a new array and populates it with elements that meet the filter condition. This means it creates a new array in memory to hold the filtered elements, which can be memory-intensive, especially if the original array `$children` is large.
> 
> On the other hand, a foreach loop initializes an empty array `$childFolders` and then iterates through the `$children` array, only adding elements to `$childFolders` if they meet the filtering condition. This approach doesn't create an additional array for the filtered elements, so it consumes less memory.

I do not understand the difference, you wrote twice the same thing from what I understand.

If array_filter is worse than a loop doing the same thing an issue should be opened with the PHP project.
Did you test on several PHP version for a difference?
Also, maybe it uses more memory because it prioritizes speed?

come-nc:
https://externals.io/message/100452
Interesting read, 6 years old so not sure if all still apply but it seems a callback function is always expensive speedwise.

solracsf:
Im my tests, on different php projects, the foreach loop was always less **memory** intensive (i'm not analysing speed/performance here) once the array is big (>hundreds). In this specific case, I've tested with v8.1 only.

blizzz (approved):

blizzz:
foreach() has a lower memory footprint the most of the array_* functions, so when you can expect to get a few more entries in your arrays, use foreach.

mrtumnus:
Ok, but this made it nearly 2x as slow, so was this really a good tradeoff?
(stumbled across this as I wait for my `files:scan-app-data` to complete :laughing: )

PR #41295, opened 2023-11-06
fix(JobList#hasReservedJob): Break tie when multiple jobs are reserved

marcelklehr:
## Summary
Currently, when multiple jobs of the same class are reserved that have all set allowParallelRuns to false, none of them will run. This changes the check to only consider the jobs reserved in the past 6h as currently running.

## Checklist


miaulalala (approved):

nickvergessen (approved):
Worth to put into the comment of the method/interface (public api)

marcelklehr:
Done. CI failures seem unrelated. Please merge.

PR #41298, opened 2023-11-06
fix description of sharing toggle

szaimen:
| Before | After |
|---|---|
| image | image |


apps/files_sharing/src/actions/sharingStatusAction.ts
```diff
@@ -54,7 +54,7 @@ export const action = new FileAction({
 			return t('files_sharing', 'Shared')
 		}
 
-		return ''
+		return t('files_sharing', 'Show sharing options')
```
szaimen, line 57: @skjnldsv this on one hand makes sure that the item has a desciption in the dropdown:
image
However it also displays inline which we probably do not want:
image
Do you have an idea how to fix both at the same time?

skjnldsv (approved):

artonge (approved):

szaimen:
Unfortunately it does not work in gird view:
image

szaimen:
I guess this is the problem?
image

susnux (approved):

PR #41300, opened 2023-11-06
Add timezone getter to ITimeFactory

miaulalala:
## Summary
Add factory method to generate a \DateTimeZone object

Better testability

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/public/AppFramework/Utility/ITimeFactory.php
+++ b/lib/public/AppFramework/Utility/ITimeFactory.php
@@ -58,4 +58,12 @@ interface ITimeFactory extends ClockInterface {
 	 * @since 26.0.0
 	 */
 	public function withTimeZone(\DateTimeZone $timezone): static;
+
+	/**
+	 * @param string|null $timezone
+	 * @return \DateTimeZone Requested timezone if provided, UTC otherwise
+	 * @throws \Exception
+	 * @since 29.0.0
+	 */
+	public function getTimeZone(?string $timezone = null): \DateTimeZone;
 }
--- a/lib/private/AppFramework/Utility/TimeFactory.php
+++ b/lib/private/AppFramework/Utility/TimeFactory.php
@@ -73,4 +73,11 @@ class TimeFactory implements ITimeFactory {
 
 		return $clone;
 	}
+
+	public function getTimeZone(?string $timezone = null): \DateTimeZone {
+		if ($timezone !== null) {
+			return new \DateTimeZone($timezone);
+		}
+		return $this->timezone;
+	}
 }
--- a/tests/lib/AppFramework/Utility/TimeFactoryTest.php
+++ b/tests/lib/AppFramework/Utility/TimeFactoryTest.php
@@ -46,4 +46,21 @@ class TimeFactoryTest extends \Test\TestCase {
 		$now = $withTimeZone->now();
 		self::assertSame('Europe/Berlin', $now->getTimezone()->getName());
 	}
+
+	public function testGetTimeZone(): void {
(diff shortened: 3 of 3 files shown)


miaulalala:
Got some additional ideas, adding a DateTimeImmutable factory method with params would be nice too.

lib/public/AppFramework/Utility/ITimeFactory.php
```diff
@@ -58,4 +58,12 @@ public function getDateTime(string $time = 'now', \DateTimeZone $timezone = null
...
 	 */
 	public function withTimeZone(\DateTimeZone $timezone): static;
+
+	/**
+	 * @param string|null $timezone
+	 * @return \DateTimeZone
+	 * @throws \Exception
+	 * @since 28.0.0
+	 */
+	public function getTimezone(?string $timezone = null): \DateTimeZone;
```
nickvergessen, line 68: To be more inline with above method lets uppercase Z?

```suggestion
	public function getTimeZone(?string $timezone = null): \DateTimeZone;
```

AndyScherzinger (approved):

nickvergessen (approved):

lib/public/AppFramework/Utility/ITimeFactory.php
```diff
@@ -58,4 +58,12 @@ public function getDateTime(string $time = 'now', \DateTimeZone $timezone = null
 	 * @since 26.0.0
 	 */
 	public function withTimeZone(\DateTimeZone $timezone): static;
+
+	/**
+	 * @param string|null $timezone
+	 * @return \DateTimeZone
```
come-nc, line 64:
```suggestion
	 * @return \DateTimeZone Requested timezone if provided, UTC otherwise
```
You need to document what passing null does. Looking at the implementation it seems to always return UTC?

lib/public/AppFramework/Utility/ITimeFactory.php
```diff
@@ -58,4 +58,12 @@ public function getDateTime(string $time = 'now', \DateTimeZone $timezone = null
...
+	 * @throws \Exception
```
come-nc, line 65:
```suggestion
	 * @throws \DateInvalidTimeZoneException
```
Unless it’s expected to be able to throw other stuff later?

ChristophWurst:
needs dev docs

lib/public/AppFramework/Utility/ITimeFactory.php
nickvergessen: > Looking at the implementation it seems to always return UTC?

Only if `withTimeZone()` was not called?

come-nc: Indeed. So my suggestion is not the right one, but there should be an explanation as to what passing null means.

lib/public/AppFramework/Utility/ITimeFactory.php
miaulalala: But new Timezone throws a simple \Exception?

come-nc: Ouuuh you’re right, I got tricked by documentation, this changed in PHP 8.3, so we have to keep general `\Exception` until 8.2 is EOL ^^.

come-nc (approved):

lib/public/AppFramework/Utility/ITimeFactory.php
nickvergessen: > until 8.2 is EOL

*me cries in 8.0*

Issue #40242, opened 2023-09-04
🖼️ iOS live photos support


AndyScherzinger:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Support iOS Live Photos

This is splitted into 2 part, supporting them regarding DAV operations on the server (i.e. when the `jpeg` is moved, also move the `mov` file.

- [x] #40243
- [ ] #40244

**Describe alternatives you've considered**

**Additional context**

sorbaugh:
cc @tobiasKaminsky this might also need additional work in the clients?

AndyScherzinger:
> cc @tobiasKaminsky this might also need additional work in the clients

Step 1 would be iOS client sending/setting the meta data.

Step 2 would be the clients marking live photos as such combining the image and video on the visualization, specifics to be discussed with @nextcloud/designers, other platforms just display the image and add a 'live' icon to it. 

Step 3 is to combine dav operations to be applied to both files.

Step 4 display live photos on all clients.

Step 4 would be the stretch goal

tobiasKaminsky:
For Step 2: https://support.apple.com/guide/iphone/edit-live-photos-iphd8dbb3291/ios

tobiasKaminsky:
server: https://livephotos.ltd.nextcloud.com
user: ios

Webdav property: <nc:metadata-files-live-photo/>

On server:
sample.jpg: <nc:metadata-files-live-photo>sample.mov</nc:metadata-files-live-photo>
sample.mov: <nc:metadata-files-live-photo>sample.jpg</nc:metadata-files-live-photo>

AndyScherzinger:
closing, left open is sub-task to display live photos on web, tracked in the remaining task

PR #41308, opened 2023-11-06
Customize rendering for live photos

artonge:
- Step 2 of https://github.com/nextcloud/server/issues/40242 

### Changes

- [x] Hide `.mov` file when it has a `hidden=true` property.
- [x] Show a play icon on `.png` file when it has a `live-photo` metadata.

### Visuals

Screenshot from 2023-11-09 10-48-48
Screenshot from 2023-11-09 10-48-40


skjnldsv:
> We should probably improve the contrast of the overlay icon like for the star icon

Do it with css

apps/files/src/views/FilesList.vue
```diff
@@ -256,7 +256,10 @@ export default Vue.extend({
 		},
 
 		dirContents(): Node[] {
-			return (this.currentFolder?._children || []).map(this.getNode).filter(file => file)
+			return (this.currentFolder?._children || [])
+				.map(this.getNode)
+				.filter(file => file)
+				.filter(({ attributes: { hidden } }) => hidden !== true)
```
skjnldsv, line 262: Sorry, fancy but unreadable.
```suggestion
				.filter(file => file?.attributes?.hidden !== true)
```

apps/files/src/init.ts
```diff
@@ -63,3 +65,7 @@ registerRecentView()
 
 // Register preview service worker
 registerPreviewServiceWorker()
+
+registerDavProperty('nc:hidden', { nc: 'http://nextcloud.org/ns' })
+
+initLivePhotos()
```
skjnldsv, line 71: Why are we not registering the `registerDavProperty('nc:metadata-files-live-photo')` here too?
We migrated away from side-effect action import :)

skjnldsv (dismissed):

apps/files/src/init.ts
artonge: It is not side effect import, the registration is done in `initLivePhotos()`.
This was just to not "pollute" the init script with feature specific code, and keep it all in one file.

skjnldsv: Indeed! No preference then, I see both points :)

ArtificialOwl (approved):
back-end part is 👍

apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -87,6 +87,7 @@ class FilesPlugin extends ServerPlugin {
 	public const SUBFOLDER_COUNT_PROPERTYNAME = '{http://nextcloud.org/ns}contained-folder-count';
 	public const SUBFILE_COUNT_PROPERTYNAME = '{http://nextcloud.org/ns}contained-file-count';
 	public const FILE_METADATA_PREFIX = '{http://nextcloud.org/ns}metadata-';
+	public const HIDDEN_PROPERTYNAME = '{http://nextcloud.org/ns}hidden';
```
susnux, line 90: Probably needs to go in here: https://github.com/nextcloud/documentation/blob/cd26ad08b1713d78750e1789aa4b992e34343d5d/developer_manual/client_apis/WebDAV/basic.rst#supported-properties

apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -50,6 +50,10 @@
 			:aria-label="t('files', 'Favorite')">
 			<FavoriteIcon v-once />
 		</span>
+
+		<OverlayIcon :is="fileOverlay"
```
susnux, line 54: This will break once we migrate to vue3 so you can already prepare by using `<component>` instead:
```suggestion
		<component :is="fileOverlay"
```

apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -50,6 +50,10 @@
...
+			v-if="fileOverlay"
+			class="files-list__row-icon-overlay" />
```
susnux, line 56: Alternatively:
```suggestion
		<PlayCircleIcon v-if="isLivePhoto" class="files-list__row-icon-overlay" />
```

Then drop `fileOverlay` and add a computed:
```js
// computed: {
isLivePhoto() {
    return isLivePhoto(this.source)
}
```

apps/files/src/views/FilesList.vue
```diff
@@ -256,7 +256,10 @@ export default Vue.extend({
 		},
 
 		dirContents(): Node[] {
-			return (this.currentFolder?._children || []).map(this.getNode).filter(file => file)
+			return (this.currentFolder?._children || [])
+				.map(this.getNode)
+				.filter(file => file)
```
susnux, line 261:
```suggestion
```
Not needed to call filter twice, this is already checked in the next one

susnux (approved):
some nits

apps/files/src/components/FileEntry/FileEntryPreview.vue
artonge: No, as we want a generic solution, like for folders

apps/files/src/views/FilesList.vue
artonge: The first one checks if a file is `undefined` or null, the second one checks the content of the hidden property. So not a duplicate check, or I misunderstood your comment?

apps/dav/lib/Connector/Sabre/FilesPlugin.php
artonge: I added the documentation tag on the PR

artonge:
Cypress failure unrelated

artonge:
Cypress failure unrelated

apps/files/src/views/FilesList.vue
susnux: Never mind I thought it would be `===` but in this case it makes no sense

susnux (approved):

artonge:
CI failure unrelated

PR #41327, opened 2023-11-07
dont reuse etag for folders marked explicitly unscanned

icewind1991:
Speculated issue:

- An external smb storage is setup where the folder mtime doesn't change when new files get created
- `files_external:notify` is running marking the parent folder of any created file with a size of -1
- the background scanner triggers for the parent folder, finding the new file
- because the mtime of the folder didn't change, the old etag for the parent folder is reused
- the sync client doesn't pickup the change because the etag of the parent folder didn't change.

A use case that might trigger this is an external smb storage where the underlying fs doesn't change folder mtime as we expect but "files_external:notify" is being used to process external file changes.

I was able to reproduce this with manual `touch`ing to "unchange" the folder mtime a few times but then *something* changed and I can't anymore...

### DIFF
--- a/lib/private/Files/Cache/Scanner.php
+++ b/lib/private/Files/Cache/Scanner.php
@@ -203,7 +203,8 @@ class Scanner extends BasicEmitter implements IScanner {
 						$fileId = $cacheData['fileid'];
 						$data['fileid'] = $fileId;
 						// only reuse data if the file hasn't explicitly changed
-						if (isset($data['storage_mtime']) && isset($cacheData['storage_mtime']) && $data['storage_mtime'] === $cacheData['storage_mtime']) {
+						$mtimeUnchanged = isset($data['storage_mtime']) && isset($cacheData['storage_mtime']) && $data['storage_mtime'] === $cacheData['storage_mtime'];
+						if ($mtimeUnchanged && $cacheData['size'] !== -1) {
 							$data['mtime'] = $cacheData['mtime'];
 							if (($reuseExisting & self::REUSE_SIZE) && ($data['size'] === -1)) {
 								$data['size'] = $cacheData['size'];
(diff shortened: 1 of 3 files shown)


icewind1991:
Fix confirmed working by customer

come-nc (approved):

come-nc:
@icewind1991 backport to stable28 as well no?

lib/private/Files/Cache/Scanner.php
```diff
@@ -203,7 +203,8 @@ public function scanFile($file, $reuseExisting = 0, $parentId = -1, $cacheData =
 						$fileId = $cacheData['fileid'];
 						$data['fileid'] = $fileId;
 						// only reuse data if the file hasn't explicitly changed
-						if (isset($data['storage_mtime']) && isset($cacheData['storage_mtime']) && $data['storage_mtime'] === $cacheData['storage_mtime']) {
+						$mtimeUnchanged = isset($data['storage_mtime']) && isset($cacheData['storage_mtime']) && $data['storage_mtime'] === $cacheData['storage_mtime'];
+						if ($mtimeUnchanged && $cacheData['size'] !== -1) {
```
Fenn-CS, line 207: I guess `$cacheData['size'] !== -1` is what makes the difference here! Code looks okay, but it would be nice if there is comment that states what `$cacheData['size']  === -1` means` either around here or somewhere at the file since it's used in multiple places.

It would also be nice to have that kind of explanation in the commit body, so it can help reviewers and future commit readers.

lib/private/Files/Cache/Scanner.php
```diff
@@ -389,16 +394,20 @@ protected function getExistingChildren($folderId) {
 	 * @param int|float $oldSize the size of the folder before (re)scanning the children
 	 * @return int|float the size of the scanned folder or -1 if the size is unknown at this stage
 	 */
-	protected function scanChildren(string $path, $recursive, int $reuse, int $folderId, bool $lock, int|float $oldSize) {
+	protected function scanChildren(string $path, $recursive, int $reuse, int $folderId, bool $lock, int|float $oldSize, &$etagChanged = false) {
```
Fenn-CS, line 397: Is it intentional to pass a reference variable here?

Fenn-CS: Oh seen

> // "etag changed" propagates up, but not down, so we pass `false` to the children even if we already know that the etag of the current folder changed

nfebe (approved):

PR #41329, opened 2023-11-07
fix(files): Do not encode the path twice (and do not use undefined method)

susnux:
## Summary

Fixes files being not able to be moved, currently the error is "encodePath not defined" but adding `encodePath` will break as well because then file names are encoded twice, e.g. in the following example it would fail because "Neuer Ordner" would be encoded to "Neuer%20Ordner" and then to "Neuer%2520Ordner".

### Before

https://github.com/nextcloud/server/assets/1855448/cf0b9894-4bba-4568-9623-408006ebb45b

### After

https://github.com/nextcloud/server/assets/1855448/7958afea-3100-4658-b8e0-ceb68ea17fd8

## Checklist


skjnldsv:
Hum, weird.
We had to do it because it was misbehaving in the first place :thinking:

skjnldsv (approved):

Pytal (approved):

PR #41334, opened 2023-11-07
fix(component): utilized NcDateTime component

### LINKED ISSUE #41261: Replace mtime render with `NcDateTime`
https://nextcloud-vue-components.netlify.app/#/Components/NcDateTime
https://github.com/nextcloud/server/blob/4db03884a566bc637e859cb83c806201b326bb3c/apps/files/src/components/FileEntry.vue#L80-L87

emoral435:
Resolves #41261
Replaces span element with custom Nc Vue Component
Signed-off-by: Eduardo Morales <emoral435@gmail.com>

* Resolves: #41261

## BEFORE
chrome_c0cPiphzOz
## AFTER
chrome_c0cPiphzOz

## Summary
Replaces span element with custom Nc Vue Component. Frontend looks the same, but the actual code uses the Nextcloud Component

## TODO

- [x] utilize prebuilt NcDateTime component from @nextcloud/vue library

## Checklist


apps/files/src/components/FileEntry.vue
```diff
@@ -20,7 +20,7 @@
   -
   -->
 
-<template>
+  <template>
```
skjnldsv, line 23:
```suggestion
<template>
```

apps/files/src/components/FileEntry.vue
```diff
@@ -262,6 +264,12 @@ export default Vue.extend({
 			}
 			return t('files_trashbin', 'A long time ago')
 		},
+		mDateTime() {
+			if (this.source.mtime) {
+				return moment(this.source.mtime)
```
skjnldsv, line 269: Actually, I think `NcDateTime` requires a `Date` object, not `moment`
https://nextcloud-vue-components.netlify.app/#/Components/NcDateTime
image

apps/files/src/components/FileEntry.vue
```diff
@@ -262,6 +264,12 @@ export default Vue.extend({
 			}
 			return t('files_trashbin', 'A long time ago')
```
skjnldsv, line 265:
```suggestion
			return t('files', 'A long time ago')
```

apps/files/src/components/FileEntry.vue
```diff
@@ -262,6 +264,12 @@ export default Vue.extend({
...
 		},
+		mDateTime() {
+			if (this.source.mtime) {
+				return moment(this.source.mtime)
+			}
+			return t('files_trashbin', 'A long time ago')
```
skjnldsv, line 271:
```suggestion
			return t('files', 'A long time ago')
```

apps/files/src/components/FileEntry.vue
skjnldsv, line 265: I think the entire`mtime` computed can be removed since it's unused ?

apps/files/src/components/FileEntry.vue
```diff
@@ -83,7 +83,7 @@
 			class="files-list__row-mtime"
 			data-cy-files-list-row-mtime
 			@click="openDetailsIfAvailable">
-			<span>{{ mtime }}</span>
+			<NcDateTime :timestamp="mDateTime"/>
```
skjnldsv, line 86: I would suggest to remove the seconds with this:

```suggestion
			<NcDateTime :timestamp="mDateTime" :ignore-seconds="true" />
```

skjnldsv (dismissed):
Nice!!
See inline comments :)

apps/files/src/components/FileEntry.vue
susnux: Yes `Date` object means the JS native `Date`

emoral435:
I removed the mtime computed section because yeah, it wasn't really being used! Nice catch!
And because the mtime member from the Node class returns a date object of when it was modified, I used that

apps/files/src/components/FileEntry.vue
```diff
@@ -255,13 +257,6 @@ export default Vue.extend({
 				color: `color-mix(in srgb, var(--color-main-text) ${ratio}%, var(--color-text-maxcontrast))`,
 			}
 		},
-
-		mtime() {
-			if (this.source.mtime) {
```
susnux, line 260: Can this be empty? In this case NcDateTime will throw an error as the value must be int or Date. If not everything should be fine.

susnux (approved):

apps/files/src/components/FileEntry.vue
susnux: >  And because the mtime member from the Node class returns a date object of when it was modified, I used that

Ah did not see that, then it should be fine

emoral435: I believe it can't be empty under most circumstances, but to comply with the rest of the checks in place with the rest of the file, I'll add in some checks so that if the mtime is empty, we get a default translatable string to display instead

emoral435: Actually, because this is only conditionally rendered when the mtime is not empty, then this wont throw an error, so your right! It should be fine

apps/files/src/components/FileEntry.vue
emoral435: Moved this back in by a bit, I must have hit a space by accident! Whoops

apps/files/src/components/FileEntry.vue
emoral435: Agreed, removed it 👍

apps/files/src/components/FileEntry.vue
```diff
@@ -83,7 +83,7 @@
 			class="files-list__row-mtime"
 			data-cy-files-list-row-mtime
 			@click="openDetailsIfAvailable">
-			<span>{{ mtime }}</span>
+			<NcDateTime :timestamp="this.source.mtime" :ignore-seconds="true" />
```
Pytal, line 86:
```suggestion
			<NcDateTime :timestamp="source.mtime" :ignore-seconds="true" />
```
`this` is not needed in template

emoral435: changing this right now, thank you!

Pytal (approved):

PR #41338, opened 2023-11-08
fix(files): make sure we focus the parent actions submenu entry

skjnldsv:
Best with https://github.com/nextcloud-libraries/nextcloud-vue/pull/4775


apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -317,6 +318,21 @@ export default Vue.extend({
...
+			this.openedSubmenu = null
+			// Wait for first render
+			await this.$nextTick()
+
+			// Focus the previous menu action button
+			this.$nextTick(() => {
+				// Focus the action button
+				const menuAction = this.$refs[`action-${action.id}`][0]
+				if (menuAction) {
+					menuAction.$el.querySelector('button')?.focus()
```
susnux, line 331: For performance I would not recommend `querySelector` if not needed but use `getElementsByClassName` or `getElementByTagName`.

(In this case probably not so important but I learned this couple of days ago, if you do multiple queries the `querySelector` is quite slow because of additional logic).

susnux (approved):

susnux:
I like this solution :)

apps/files/src/components/FileEntry/FileEntryActions.vue
skjnldsv: Yeah, selecting have various arrays of perf gaps. 
But to be fair, the difference will have very little impact for 99% of use cases
https://www.measurethat.net/Benchmarks/Show/1144/1/getelementbyid-vs-queryselector-vs-getelementsbyclassna 
image


And it's different than the css performances too
https://benfrain.com/css-performance-revisited-selectors-bloat-expensive-styles/

emoral435 (approved):
Looks good!

PR #41340, opened 2023-11-08
feat(caldav): Create personal event for out-of-office messages

### LINKED ISSUE #41337: Create personal event for out of office messages
### How to use GitHub

## Is your feature request related to a problem? Please describe.

As a Nextcloud user I want my calendar to reflect when I am out of office.

## Describe the solution you'd like

Use the data of https://github.com/nextcloud/server/issues/40616 and create an event in my personal calendar with status busy. Replace event when absence info changes. Delete event when absence ends or is disabled.

## Describe alternatives you've considered

N/a

## Additional context

N/a

ChristophWurst:
* Resolves: https://github.com/nextcloud/server/issues/41337

## Summary

Create and maintain an event in a user's personal calendar when they configure an out of office message.

## TODO

- [x] Create event when OOO is set
- [x] Update event when OOO is updated
- [x] Delete event when OOO is deleted before it ends
- [x] Bug when OOO event is deleted by user and causes a trash bin conflict when deleted again by the system

## Checklist


apps/dav/lib/Listener/OutOfOfficeListener.php
```diff
@@ -0,0 +1,179 @@
...
+			$this->logger->warning('Personal calendar has been deleted');
+			return null;
+		}
+
+		return $calendarNode;
+	}
+
+	private function getEventFileName(string $token): string {
+		return "out_of_office_$token.ics";
+	}
```
ChristophWurst, line 152: I want to write an event and find the same object again when the OOO data changes or gets cleared. Forcing a stable file name is not great because this only works once. After the first deletion there is an entry in the trash bin and the second event can not be moved.

Are there good alternatives or should I just try to disable the trash bin for this automatic event? @miaulalala @tcitworld?

ChristophWurst: I could adjust the emitted event so it also transports the previous OOO values. That allows me to find the old event searching by date range.

tcitworld: In that case could you use the object UID instead of it's URI ?

ChristophWurst: What do you mean?

tcitworld: Finding the old event through it's UID instead of it's date range should be faster. Sorry, last message wasn't clear.

tcitworld: There's little constraints on UID (alphanumeric + `-`), so you can build your own UID based on deterministic things (user, timestamp).

ChristophWurst: I think that's more or less what I tried with `"out_of_office_$token.ics"`. The ID is unique and stable.

It gets problematic if
1) OOO info is created -> integration creates event
2) OOO is changed -> integration deletes previous event, creates new
2) OOO is changed again -> integration deletes previous event :boom: because there would be two "out_of_office_123.ics" in the trash bin now

ChristophWurst: Found a better way. I'm just replacing the contents :)

tcitworld: Make sure to create an entry into `calendar_changes` too then.

ChristophWurst: I'm going through `\Sabre\CalDAV\CalendarObject::put`, so that happens automagically :)

apps/dav/lib/Listener/OutOfOfficeListener.php
```diff
@@ -0,0 +1,218 @@
...
+			->setTime(0, 0);
+		if ($timeZoneData !== null) {
+			/** @var VCalendar $vtimezoneObj */
+			$vtimezoneObj = Reader::read($timeZoneData);
+			/** @var VTimeZone $vtimezone */
+			$vtimezone = $vtimezoneObj->VTIMEZONE;
+			$calendarTimeZone = $vtimezone->getTimeZone();
+			$vCalendar = new VCalendar([
+				'CALSCALE' => 'GREGORIAN',
+				'VERSION' => '2.0',
```
miaulalala, line 191: If you initialise the object with `new VCalendar()` you don't need to set the defaults manually. You're also missing the ['prodid' which is REQUIRED but MUST NOT occur more than once](https://datatracker.ietf.org/doc/html/rfc5545#section-3.6).

apps/dav/lib/Listener/OutOfOfficeListener.php
```diff
@@ -0,0 +1,218 @@
...
+				'VTIMEZONE' => $vtimezone,
+				'VEVENT' => [
+					'SUMMARY' => $shortMessage,
+					'DESCRIPTION' => $longMessage,
+					'STATUS' => 'CONFIRMED',
+					'DTSTART' => $start->setTimezone($calendarTimeZone)->setTime(0, 0),
+					'DTEND' => $end->setTimezone($calendarTimeZone)->setTime(0, 0),
+					'X-NEXTCLOUD-OUT-OF-OFFICE' => $data->getId(),
+				],
+			]);
```
miaulalala, line 201:
```suggestion
			$vCalendar = new VCalendar();			
			$vCalendar->add($vtimezone);
			// I think this works, but otherwise intialise a new VEvent with the $vCalendar as root.
			$vCalendar->add(
				'VEVENT' => [
					'SUMMARY' => $shortMessage,
					'DESCRIPTION' => $longMessage,
					'STATUS' => 'CONFIRMED',
					'DTSTART' => $start->setTimezone($calendarTimeZone)->setTime(0, 0),
					'DTEND' => $end->setTimezone($calendarTimeZone)->setTime(0, 0),
					'X-NEXTCLOUD-OUT-OF-OFFICE' => $data->getId(),
				],
			]);
```

apps/dav/lib/Listener/OutOfOfficeListener.php
```diff
@@ -0,0 +1,218 @@
...
+		} else {
```
miaulalala, line 202: early return is possibe here too.

miaulalala (changes requested):

miaulalala:
You can leave as is except for the PRODID.

apps/dav/lib/Listener/OutOfOfficeListener.php
ChristophWurst: not a fan of this tbh because this structure doesn't eliminate any indentation and is more prone to bugs because one has to remember to add early returns everywhere

apps/dav/lib/Listener/OutOfOfficeListener.php
ChristophWurst: fixed!

apps/dav/lib/Listener/OutOfOfficeListener.php
```diff
@@ -0,0 +1,207 @@
...
+			'DTEND' => $end,
+			'X-NEXTCLOUD-OUT-OF-OFFICE' => $data->getId(),
+		]);
+		if ($timeZoneData !== null) {
+			/** @var VCalendar $vtimezoneObj */
+			$vtimezoneObj = Reader::read($timeZoneData);
+			/** @var VTimeZone $vtimezone */
+			$vtimezone = $vtimezoneObj->VTIMEZONE;
+			$calendarTimeZone = $vtimezone->getTimeZone();
+			$vCalendar->add($vtimezone);
```
miaulalala, line 198: Did you mean this?

```suggestion
			$vtimezone = $vtimezoneObj->VTIMEZONE;
			$calendarTimeZone = $vtimezone->getTimeZone();
			$vCalendar->add($calendarTimeZone);
```

miaulalala (changes requested):

apps/dav/lib/Listener/OutOfOfficeListener.php
ChristophWurst: No because I want to add the full VTIMEZONE, not the TZID

ChristophWurst: Test it

miaulalala (approved):

AndyScherzinger (approved):

tcitworld:
@ChristophWurst  #42142 made me notice I had never submitted my review. Added two comments directly on commit itself since I can't post it anymore.
https://github.com/nextcloud/server/commit/5b7a5474b9fe9bc08f65f48c115ea08ecac5a73c

PR #41341, opened 2023-11-08
fix: Allow to set app order on navigation entries added by closures

susnux:
## Summary
Some apps, e.g. talk, use the navigation manager to inject the navigation entry instead of using the appinfo.xml
This caused issues as the app property was not set.

This moves some logic from init of the NavigationManager to the add function and refactors the app order config to use navigation IDs rather than app names + index of navigation from appinfo.xml

Also adjusted frontend code for that change.

## Checklist


juliusknorr (approved):

lib/private/NavigationManager.php
```diff
@@ -102,6 +108,11 @@ public function add($entry) {
 
 		$id = $entry['id'];
 		$entry['unread'] = $this->unreadCounters[$id] ?? 0;
+		// This is the default app that will always be shown first
+		if ($entry['type'] === 'link') {
+			$entry['default'] = ($entry['app'] ?? false) === $this->defaultApp;
+		}
+		$entry['order'] = $this->customAppOrder[$id]['order'] ?? $entry['order'] ?? 100;
```
come-nc, line 115: I’d put that into the if to avoid messing with non-link navigation entries

tests/lib/App/AppManagerTest.php
```diff
@@ -680,23 +680,23 @@ public function provideDefaultApps(): array {
 			[
 				'unexist,settings',
 				'files',
-				'{"settings":[1],"files":[2]}',
+				'{"settings":{"app":"settings","order":1},"files":{"app":"files","order":2]}',
```
come-nc, line 683:
```suggestion
				'{"settings":{"app":"settings","order":1},"files":{"app":"files","order":2}}',
```

Also, where is app key used?

susnux: In AppManager to figure out the default app for a user. Otherwise you would have to somehow map an ID to a app which is not possible as the ID can be chosen without any restrictions on closures.

lib/private/NavigationManager.php
susnux: Done :heavy_check_mark:

come-nc (approved):

come-nc (changes requested):

lib/private/App/AppManager.php
```diff
@@ -838,8 +838,9 @@ public function getDefaultAppForUser(?IUser $user = null, bool $withFallbacks =
 				/* Fallback on user defined apporder */
 				$customOrders = json_decode($this->config->getUserValue($user->getUID(), 'core', 'apporder', '[]'), true, flags:JSON_THROW_ON_ERROR);
 				if (!empty($customOrders)) {
-					$customOrders = array_map('min', $customOrders);
-					asort($customOrders);
+					uasort($customOrders, function ($a, $b) {
+						return $a['order'] - $b['order'];
+					});
 					$defaultApps = array_keys($customOrders);
```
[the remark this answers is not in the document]
susnux: :face_exhaling:

lib/private/NavigationManager.php
```diff
@@ -102,6 +108,12 @@ public function add($entry) {
 
 		$id = $entry['id'];
 		$entry['unread'] = $this->unreadCounters[$id] ?? 0;
+		if ($entry['type'] === 'link') {
+			// This is the default app that will always be shown first
+			$entry['default'] = ($entry['app'] ?? false) === $this->defaultApp;
```
come-nc, line 113: What does having default to true does? Because it will be true for all links of the default app this way. If an app has several links, maybe only the first one should have default to true?

tests/lib/App/AppManagerTest.php
```diff
@@ -680,23 +680,23 @@ public function provideDefaultApps(): array {
 			[
 				'unexist,settings',
 				'files',
-				'{"settings":[1],"files":[2]}',
+				'{"settings":{"app":"settings","order":1},"files":{"app":"files","order":2}}',
```
come-nc, line 683: We should have a test with several links for a same app, and a test with a closure thing

lib/private/NavigationManager.php
susnux: Thats the problem of "defaultApp" as we do not have a default app but a default navigation route.

This way we set all routes of that app to default, so all routes of that app are sorted first (with their respective order) and then all other apps follow.

So app1 has `a: order 2` and `b: order 3` as navigation entries, app2 has `c: order 1`.
If you set app1 as default it will look like:
`a, b, c`

If you only set the first route as `default=true`, it looks like:
`a, c, b`

tests/lib/App/AppManagerTest.php
susnux: Done

lib/private/NavigationManager.php
come-nc: This is not my experience.
If I add a Photo2 to photos info.xml and I set it first, with Photo third, I get this:
Screenshot_20231109_175153

So the order I set in the settings is respected even if I separate a and b.

come-nc (approved):
The `default` situation is not clear to me but it’s working as expected. (Also I do not know of any application actually registering several links?)

lib/private/App/AppManager.php
```diff
@@ -838,9 +838,12 @@ public function getDefaultAppForUser(?IUser $user = null, bool $withFallbacks =
 				/* Fallback on user defined apporder */
 				$customOrders = json_decode($this->config->getUserValue($user->getUID(), 'core', 'apporder', '[]'), true, flags:JSON_THROW_ON_ERROR);
 				if (!empty($customOrders)) {
-					$customOrders = array_map('min', $customOrders);
-					asort($customOrders);
-					$defaultApps = array_keys($customOrders);
+					// filter only entries with app key (when added using closures or NavigationManager::add the app is not guranteed to be set)
```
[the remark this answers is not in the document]
susnux: Yes but the "app" is not guaranteed to be set, because for closures / entries added not from appinfo.xml the caller has to set it and we can not set it if not set as we do not know the app.
We can also not enforce it as that would be breaking.

lib/private/NavigationManager.php
susnux: Yes but that is for the case where you set default app but not the apporder:
Having a second entry for files:
Screen Shot 2023-11-10 at 20 26 51
Setting files as the default app:
Screen Shot 2023-11-10 at 20 31 12

lib/private/App/AppManager.php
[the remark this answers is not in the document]
AndyScherzinger: unresolved for the moment as marked to check again

Issue #38884, opened 2023-06-19
[Bug]: Redirected to "All files" when sorting favourite files


Jerome-Herbinet:
Occurs in NC 26.0.2

### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

[Capture vidéo du 19-06-2023 10:21:07.webm](https://github.com/nextcloud/server/assets/33763786/9263d1e4-688f-43c6-91bd-2f874093f8ee)

### Steps to reproduce

1. Go to all files (sorted by name)
2. Then go to favourites
3. Try to sort by date
4. See that you are redirected to "All files" and that the sorting action didn't happen.

### Expected behavior

We should have stayed in the favourites list and the sorting action should have happened.

### Installation method

Community Web installer on a VPS or web space

### Nextcloud Server version

26

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MySQL

### Is this bug present after an update or on a fresh install?

Updated from a MINOR version (ex. 22.1 to 22.2)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "instanceid": "***REMOVED SENSITIVE VALUE***",
    "passwordsalt": "***REMOVED SENSITIVE VALUE***",
    "secret": "***REMOVED SENSITIVE VALUE***",
    "trusted_domains": [
        "nextcloud.mydomain.fr"
    ],
    "datadirectory": "***REMOVED SENSITIVE VALUE***",
    "dbtype": "mysql",
    "version": "26.0.2.1",
    "updatechecker": true,
[...]

PR #41356, opened 2023-11-09
[stable27] fix(files): prevent redirect on heading column sort

skjnldsv:
Fix https://github.com/nextcloud/server/issues/38884

Seems like the old inline `onclick="event.preventDefault()"` isn't working after all


emoral435 (approved):
Looks good to me!

ShGKme (approved):

PR #41119, opened 2023-10-25
optimize builtin storage wrappers

icewind1991:
Mostly replacing `$storage->instanceOfStorage` with `$mount instance` which is significantly faster

PR #41363, opened 2023-11-09
[27] optimize builtin storage wrappers

icewind1991:
Backport of https://github.com/nextcloud/server/pull/41119

### DIFF
--- a/lib/private/Files/FileInfo.php
+++ b/lib/private/Files/FileInfo.php
@@ -32,6 +32,8 @@
  */
 namespace OC\Files;
 
+use OC\Files\Mount\HomeMountPoint;
+use OCA\Files_Sharing\ISharedMountPoint;
 use OCP\Files\Cache\ICacheEntry;
 use OCP\Files\Mount\IMountPoint;
 use OCP\IUser;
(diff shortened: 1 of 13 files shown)


lib/private/Files/FileInfo.php
```diff
@@ -309,27 +311,12 @@ public function isShareable() {
...
 	 */
 	public function isShared() {
-		$sid = $this->getStorage()->getId();
-		if (!is_null($sid)) {
-			$sid = explode(':', $sid);
-			return ($sid[0] === 'shared');
-		}
-
-		return false;
+		return $this->mount instanceof ISharedMountPoint;
```
blizzz, line 314: will it not break when files_sharing is disabled?

juliushaertl: No, `::class` usage does not require the class to actually exist.

https://www.php.net/manual/en/language.oop5.basic.php#language.oop5.basic.class.class

blizzz: it's not `::class` though, I was doubtful about `instanceof`, but it is fine indeed.

juliusknorr (approved):

juliushaertl:
Failure unrelated (drone job timeout)

Altahrim (approved):

PR #41387, opened 2023-11-10
fix(cypress): Do not install calendar app from app store for testing

susnux:
## Summary

Instead use locally provided test app for the navigation bar.

## TODO

- We might instead of provide the test app adjust the qa testing app to have navigation entries, not sure if that would be better? I do not know if it is ok to add stub navigation entries to that app?

## Checklist


nickvergessen (approved):

susnux:
The other issue with Cypress is fixed in https://github.com/nextcloud/server/pull/41398

artonge (approved):

PR #41390, opened 2023-11-11
Metadata must be set a editable for PROPPATCH

ArtificialOwl:
With the implementation of metadata and its compatibility with webdav PROPPATCH request, and to avoid unexpected behavior, 2 things needs to be implemented:

- [x] manage a list of remotely editable metadata,
- [x] compare edit permissions on metadata and actual rights on file for initiator

_throws exception on fail, might be better to just log a warning but keep the process running ?_

### DIFF
--- a/lib/public/FilesMetadata/Model/IMetadataValueWrapper.php
+++ b/lib/public/FilesMetadata/Model/IMetadataValueWrapper.php
@@ -37,16 +37,30 @@ use OCP\FilesMetadata\Exceptions\FilesMetadataTypeException;
  * @since 28.0.0
  */
 interface IMetadataValueWrapper extends JsonSerializable {
-	/**
-	 * @since 28.0.0
-	 */
+	/** @since 28.0.0 */
 	public const TYPE_STRING = 'string';
+	/** @since 28.0.0 */
 	public const TYPE_INT = 'int';
+	/** @since 28.0.0 */
 	public const TYPE_FLOAT = 'float';
+	/** @since 28.0.0 */
 	public const TYPE_BOOL = 'bool';
+	/** @since 28.0.0 */
 	public const TYPE_ARRAY = 'array';
+	/** @since 28.0.0 */
 	public const TYPE_STRING_LIST = 'string[]';
+	/** @since 28.0.0 */
 	public const TYPE_INT_LIST = 'int[]';
+
+	/** @since 28.0.0 */
+	public const EDIT_FORBIDDEN = 0;
+	/** @since 28.0.0 */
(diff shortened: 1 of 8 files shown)


ArtificialOwl:
_cypress error not related_

ArtificialOwl:
_cypress error does not seems related_

lib/public/FilesMetadata/Model/IMetadataQuery.php
```diff
@@ -58,16 +58,28 @@ public function extractMetadata(array $row): IFilesMetadata;
...
 	 * If a link to the metadataKey already exists, returns known alias.
 	 *
+	 * <b>WARNING: $enforce set to TRUE is not supported by SQLite.
+	 * The solution is to set it to FALSE and call self::enforceMetadataKey() afterward</b>
+	 *
 	 * TODO: investigate how to support a search done on multiple values for same key (AND).
 	 *
 	 * @param string $metadataKey metadata key
-	 * @param bool $enforce limit the request only to existing metadata
+	 * @param bool $enforce limit the request only to existing metadata (<b>not supported by SQLite</b>)
```
artonge, line 67: Then maybe we should remove it?

artonge (approved):

lib/public/FilesMetadata/Model/IMetadataQuery.php
ArtificialOwl: I could make a patch post beta3 that comment the innerJoin (with an explanation) and call the enforceMetadataKey() instead

ArtificialOwl: or having DBAL doing that conversion ?

artonge: Or replace the implementation of `$enforce` with `WHERE NOT NULL` ?

come-nc (approved):
Small improvements needed on phpdoc for the EDIT_* enum

AndyScherzinger (approved):

lib/public/FilesMetadata/Model/IMetadataValueWrapper.php
```diff
@@ -287,6 +301,26 @@ public function setIndexed(bool $indexed): self;
...
+	 * @return self
+	 * @since 28.0.0
+	 */
+	public function setEditPermission(int $permission): self;
+
+	/**
+	 * get remote edit permission
+	 * (Webdav PROPPATCH)
+	 *
+	 * @return int edit permission
```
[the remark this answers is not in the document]
ArtificialOwl: let me do this on a separated PR, post merge

PR #41395, opened 2023-11-12
feat: add preview provider for emf files based on office

kesselb:
For https://github.com/nextcloud/files_emfviewer/issues/1

Needs https://github.com/nextcloud/viewer/pull/2065 to display the preview in viewer. 

## Summary

Add support for emf files. 

Chrome and Firefox do not render emf files and therefore we generate a png preview through libreoffice.

## TODO

- [x] CI
- [x] Testing
- [x] Review

## Checklist


kesselb:
@st3iny made me aware that existing emf files have the wrong mime type. 

Pushed a change for `OC\Repair\RepairMimeTypes` to set image/emf when upgrading to 28.

st3iny (approved):
Tested and works.

Weirdly, some previews are not generated when going to a folder with many emf files. This seems to be random as the set of failing files is different each time when reuploading all files.

Uploading files one by one works just fine. Could this be an issue with too many parallel instances of `libreoffice`?

kesselb:
LibreOffice only allows one invocation per user profile.[^1]

https://github.com/nextcloud/server/pull/41395/commits/b5241d52204e5d88e2fba388041e02e0bd56aea7 changed the implementation to always use a different user profile. 


This has some drawbacks:

- Overload protection: If you upload 100 emf files, you may end up with 100 LibreOffice invocations. Though, you can use preview_concurrency_new to limit the number of previews that can be generated concurrently when php-sysvsem is available.
- New profile: I assume it takes a few bits to generate a fresh LibreOffice user profile. It appears that there is no way to ask LibreOffice to not create a profile and just work with the defaults. The profile will be cleaned after use by our temp manager.
- Remove the configuration option preview_office_cl_parameters:  This is not strictly necessary yet, but if you set the configuration option, the generated path for the user profile is also missing. The configuration option is not well documented (e.g., it's unclear that the last option needs to be --outdir) and actually, there should be no reason to change it after all.

https://github.com/nextcloud/server/pull/41395/commits/475dd6035435e326b0ffab6e74f961757896b97d is a follow-up to https://github.com/nextcloud/server/pull/10198 to remove the ImageMagick code. 

I'm a bit unhappy. We are very close to 28 and my plan was to just add a new preview provider based on the office one and not rework the office provider. However, without the different user profiles, the generation does not work reliable. 

If the changes (especially to Office provider) are too late, it should also work to only release the first commit with 28. It will work somehow unless you upload plenty of documents. This issue has been around for quite a while for the other office documents and maybe acceptable. 





[^1]: https://wiki.documentfoundation.org/UserProfile

ChristophWurst:
Do I need to add any config to turn on libreoffice previews? The EML preview doesn't work for me

kesselb:
> Do I need to add any config to turn on libreoffice previews? The EML preview doesn't work for me

Yes, the provider for libreoffice is not enabled by default. 

For config.php:

```
  'enabledPreviewProviders' => [
    'OC\Preview\BMP',
    'OC\Preview\GIF',
    'OC\Preview\JPEG',
    'OC\Preview\Krita',
    'OC\Preview\MarkDown',
    'OC\Preview\MP3',
    'OC\Preview\OpenDocument',
    'OC\Preview\PNG',
    'OC\Preview\TXT',
    'OC\Preview\XBitmap',
    'OC\Preview\EMF',
    'OC\Preview\MSOfficeDoc',
    'OC\Preview\MSOffice2003',
    'OC\Preview\MSOffice2007',
  ],
```

ChristophWurst:
I think it created a preview for me but since I'm using a self-signed certificate the service worker doesn't run and I see an error on the console. AFAIK the requests go through the sw for caching.

ChristophWurst (approved):
Had to install a trustworthy SSL cert to get a working service worker

Conversion works. I get a preview for my test eml file :heart_eyes_cat: 

Code looks sane. The performance tradeoff with separate profiles is fine if else libreoffice fails hard on parallelism

kesselb:
Pending documentation: 
- Configuration flag
- Add a hint about preview_concurrency_new

kesselb:
Documentation: https://github.com/nextcloud/documentation/pull/11292

PR #41406, opened 2023-11-13
fix(log): Log exception class when message is empty

nickvergessen:
* Resolves:
*  Before is first line
* After is last line

grafik

## Checklist


tcitworld (approved):
Maybe add an `"Exception thrown:"` prefix for context?

ChristophWurst (approved):
Can we also find out where an empty log message is logged?

nickvergessen:
> Can we also find out where an empty log message is logged?

The error from production was in logged via this: https://github.com/nextcloud/server/blob/bb2c2bb7f0aa57ad7e77a914fb341e2535801da9/index.php#L81

nickvergessen:
Added the prefix

PR #41413, opened 2023-11-13
Fix apps search provider order

Altahrim:
Related to advanced search

Application search provider should be at the end, except if we are already in app list


jancborchardt (approved):

nfebe (approved):

Altahrim:
CI failure unrelated

PR #41432, opened 2023-11-13
Fix search providers icons

Altahrim:
Related to advanced search

* Use white icon for user search provider
* Add missing icons for settings search providers


Altahrim:
CI failure unrelated

lib/private/Search/SearchComposer.php
```diff
@@ -221,6 +221,10 @@ private function fetchIcon(string $appId, string $providerId): string {
 			[$appId, 'app.svg'],
 			['core', 'places/default-app-icon.svg'],
 		];
+		if ($appId === 'settings' && $providerId === 'users') {
+			// The file /apps/settings/users.svg if already used in black version by user menu
```
come-nc, line 225:
```suggestion
			// The file /apps/settings/users.svg is already used in black version by user menu
```
Typo maybe? Or I do not understand the comment.
But still I do not understand the comment I think.

Altahrim: Yes, typo…
We are using `/apps/settings/users.svg` as an icon in the user menu on top right. In the menu, we need a black icon.
But for search, we need a white version.

Fenn-CS: Must these icons, have the same name? Can't both colors be present? to avoid this?

Altahrim: > Must these icons have the same name?

My hack avoid this. The new icon is white and we keep the old one

come-nc (approved):

PR #41434, opened 2023-11-13
don't return a 'sparse' array from getMountsForFileId

icewind1991:
Not all callers are setup to deal with non-continuous arrays here. Easier to fix it here than to verify all callers


skjnldsv:
Conflicts

Altahrim (approved):

skjnldsv (approved):

icewind1991:
was fixed with https://github.com/nextcloud/server/pull/46886

PR #41436, opened 2023-11-13
Add `inAppSearch` for advanced search providers

Altahrim:
Add information about in-app search like `mail` or `talk` applications

### DIFF
--- a/lib/public/Search/IInAppSearch.php
+++ b/lib/public/Search/IInAppSearch.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2023 Benjamin Gaussorgues <benjamin.gaussorgues@nextcloud.com>
+ *
+ * @author Benjamin Gaussorgues <benjamin.gaussorgues@nextcloud.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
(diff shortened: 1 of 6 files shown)


provokateurin (approved):
OpenAPI is ok

ChristophWurst:
Non-breaking alternative: add a tagging interface with no methods.

Added benefit is that the majority of search providers don't have to implement this new method just to return false :)

ChristophWurst (commented):
Commit the new file ;)

Then run `composer dump-autoload` and commit the changes that add the new class to the class map.

lib/public/Search/IInAppSearch.php
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+namespace OCP\Search;
+
+/**
+ * Interface for search providers supporting in-app search
+ *
+ * @since 28.0.0
+ */
+interface IInAppSearch {
```
ChristophWurst, line 33:
```suggestion
interface IInAppSearch extends IProvider {
```

I *think* that the is-a relation applies here. IInAppSearch should only be used by classes that are a search provider.

But I'm okay without. Just dumping my thoughts :)

ChristophWurst (approved):

lib/public/Search/IInAppSearch.php
Altahrim: I am agree. It makes no sense without it :)

Altahrim:
Unrelated Cypress failure

PR #41437, opened 2023-11-13
don't show shares from disabled users

icewind1991:
This makes it so shares owned by disabled users no longer show up.

- [ ] Is the behavior change desired, these shares currently mostly work?
- [ ] Is there a way to reduce the performance overhead of checking the user availability?


come-nc:
See https://github.com/nextcloud/server/pull/39699 , hiding shares by disabled users was supposed to be an option in 28, not a default.

PR #41426, opened 2023-11-13
Use background to increase overlay contrast

artonge:
And limit it to files.

Part of step 2 of https://github.com/nextcloud/server/issues/40242

image
Screenshot from 2023-11-13 14-50-33

PR #41443, opened 2023-11-13
fix(files): Show Collectives icon in Collectives folders

mejo-:
## Summary

Brings back the Collectives icon to Collectives folders.

Requires #41426.

| Screenshot |
|---|
| 2023-11-13T18:32:00,950104663+01:00 |

## Checklist


artonge (approved):
I don't think it needs https://github.com/nextcloud/server/pull/41426. But should be merged by then :)

mejo-:
> I don't think it needs https://github.com/nextcloud/server/pull/41426. But should be merged by then :)

Not technically, but without it looks pretty ugly :wink:

juliusknorr (approved):

PR #41455, opened 2023-11-14
fix(files): Use `color-primary-element-text` for files/folders overlay icons

mejo-:
## Screenshots

:derelict_house: Before bright | :house_with_garden: After bright
--- | ---
grafik | grafik

:derelict_house: Before dark | :house_with_garden: After dark
--- | ---
grafik | grafik


apps/files/src/components/FilesListVirtual.vue
```diff
@@ -515,16 +515,9 @@ export default Vue.extend({
...
+				color: var(--color-primary-element-text);
 				// better alignment with the folder icon
 				margin-top: 2px;
-
-				// Improve icon contrast with a background for files
-				&--file {
-					color: var(--color-main-text);
-					background: var(--color-main-background);
-					border-radius: 100%;
-				}
```
artonge, line 527: Why revert that part ?

jancborchardt (approved):
Good stuff, looks clean and as intended. :)

artonge (changes requested):

apps/files/src/components/FilesListVirtual.vue
mejo-: I thought it would be the cleaner solution. But thinking about it again, it's probably still needed - given that files often have preview images and the overlay needs to be visible on them, right? Will revert the revert :zany_face:

mejo-: Done now.

szaimen (approved):

AndyScherzinger (approved):

PR #41459, opened 2023-11-14
implements search on null/notnull metadata

ArtificialOwl:
Add the possiblity to search on null/not null value on metadata.
This is a test-feature for the Photos App.

This PR requires to add this line
```
      '{DAV:}is-defined'    => Operator::class,
```
in https://github.com/nextcloud/3rdparty/blob/master/icewind/searchdav/src/DAV/QueryParser.php#L78

```
[...]
     <d:where>
        <d:and>
          <d:not>
          <d:is-defined>
            <d:prop><nc:metadata-test/></d:prop>
            </d:is-defined>
          </d:not>
        </d:and>
[...]
```

Need:
- [x] https://github.com/icewind1991/SearchDAV/pull/13
- [x] https://github.com/nextcloud/3rdparty/pull/1643

### DIFF
--- a/apps/dav/lib/Files/FileSearchBackend.php
+++ b/apps/dav/lib/Files/FileSearchBackend.php
@@ -386,7 +386,7 @@ class FileSearchBackend implements ISearchBackend {
 				}
 
 				return new SearchComparison(
-					ISearchComparison::COMPARE_DEFINED,
+					$trimmedType,
 					$field,
 					$this->castValue($property, $value ?? ''),
 					$extra ?? ''
--- a/apps/dav/lib/Files/FileSearchBackend.php
+++ b/apps/dav/lib/Files/FileSearchBackend.php
@@ -39,6 +39,7 @@ use OCP\Files\Cache\ICacheEntry;
 use OCP\Files\Folder;
 use OCP\Files\IRootFolder;
 use OCP\Files\Node;
+use OCP\Files\Search\ISearchComparison;
 use OCP\Files\Search\ISearchOperator;
 use OCP\Files\Search\ISearchOrder;
 use OCP\Files\Search\ISearchQuery;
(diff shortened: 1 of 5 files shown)


lib/private/Files/Cache/SearchBuilder.php
```diff
@@ -57,6 +58,7 @@ class SearchBuilder {
 		ISearchComparison::COMPARE_GREATER_THAN_EQUAL => 'lt',
 		ISearchComparison::COMPARE_LESS_THAN => 'gte',
 		ISearchComparison::COMPARE_LESS_THAN_EQUAL => 'gt',
+		ISearchComparison::COMPARE_NULL => 'isNotNull',
```
artonge, line 61:
```suggestion
		ISearchComparison::COMPARE_NOT_NULL => 'isNotNull',
```

apps/dav/lib/Files/FileSearchBackend.php
```diff
@@ -382,7 +382,9 @@ private function transformSearchOperation(Operator $operator) {
 					return new SearchComparison($trimmedType, $this->mapPropertyNameToColumn($property), $value);
 				}
 
-				// no break
+			case Operator::OPERATION_IS_DEFINED:
+				$property = $operator->arguments[0];
+				return new SearchComparison('null', substr($property->name, strlen(FilesPlugin::FILE_METADATA_PREFIX)), '', IMetadataQuery::EXTRA);
```
artonge, line 387: Maybe it should not be limited to the metadata?

artonge (commented):
Nice

apps/dav/lib/Files/FileSearchBackend.php
ArtificialOwl: let's check if it is working with metadata first !

lib/private/Files/Cache/SearchBuilder.php
ArtificialOwl: check the name of the array :]

icewind1991:
might be cleaner to just transform the dav operator to `field == null` `!(field == null)` instead of adding a new search operator

artonge:
> might be cleaner to just transform the dav operator to `field == null` `!(field == null)` instead of adding a new search operator

Do you mean in the DAV request ? The issue is that we are casting the type of the property here: https://github.com/nextcloud/server/blob/b79a21d29d575fe42ae74aa21bc1a7d08546367c/apps/dav/lib/Files/FileSearchBackend.php#L418-L436
And as the property is used in another comparison in the request, it is assumed to be an integer. So using the appropriate operator would be cleaner I think.

artonge:
Pushed 3rdparty changes

ChristophWurst (approved):
:+1: for the caldav changes

AndyScherzinger (approved):

ChristophWurst:
adds new public API -> needs docs

PR #41481, opened 2023-11-15
Enh/beta improvements

### LINKED ISSUE #41450: Beta design feedback
Discussed with @karlitschek and @marcoambrosini. cc @AndyScherzinger for awareness. Sorted by priority of visual impact.

Since we are in the beta stage and already beyond feature freeze, we need to focus on easy fixes.

1. [x] **Sidebar tabs too present and conflict with primary buttons, especially in Talk:** Style will be changed from full primary background highlight to underline. → FIX at https://github.com/nextcloud-libraries/nextcloud-vue/pull/4815
2. [x] **Share button too present:** Style change to tertiary non-bold button (background will be removed and color adjusted to color-primary-element). → FIX at https://github.com/nextcloud/server/pull/41453
3. [x] **Groupfolder and other folder icons are off:** There is a dark icon with white stroke on color-primary icon. Instead icon should simply be color-primary-element-text on the color-primary-element icon – same logic as the primary button. → FIX at https://github.com/nextcloud/server/pull/41455
4. [x] **Favorites button in sidebar overlaps icon/preview too much:** "Favorite" can move to actions menu, and indicator to start of subline with size and modified date. https://github.com/nextcloud/server/pull/41481
5. [x] **Comments layout improvements:** Vertically align name with avatar, remove "" quotes so subline does not wrap. https://github.com/nextcloud/server/pull/41481

Files

Additional second set of feedback:
1. [ ] **Black border in "New" menu in Safari:** There is a black border on the first entry of the "New" menu. It should only appear on tab navigation, not on initial opening or interacting with the mouse. → Bug with Safari, does not appear with Chrome. @ShGKme can reproduce
   Issue: https://github.com/nextcloud-libraries/nextcloud-vue/issues/4874
2. [ ] **Some action menus have scrollbars:** Issue exists at https://github.com/nextcloud-libraries/nextcloud-vue/issues/3920 → According to @marcoambrosini @szaimen very complicated as the component would need logic to know where it is opened. Currently always 50vh to work in all circumstances.
3. [x] **Time of calendar notification is based on first of a recurring meeting:** The time in the notification should be based on the next meeting and not the first of a recurring event. → Christoph opened issue at https://github.com/nextcloud/server/issues/41615
4. [x] **External storage settings layout looks off:** Should be on top of the page, or properly integrated into emptycontent view → @marcoambrosini @szaimen https://github.com/nextcloud/server/issues/41614
5. [x] **"Shared" label takes a lot of space in Files list for internal shares:** → Proposals at https://github.com/nextcloud/server/issues/41620 https://github.com/nextcloud/server/pull/41656

Some more feedback from Jan:
- [x] https://github.com/nextcloud-libraries/nextcloud-vue/issues/4872
- [x] https://github.com/nextcloud-libraries/nextcloud-vue/issues/4870 → pull request at https://github.com/nextcloud-libraries/nextcloud-vue/pull/4875
- [x] https://github.com/nextcloud-libraries/nextcloud-vue/issues/4871

marcoambrosini:
* Resolves: #41450

Screenshot 2023-11-15 at 14 58 04


marcoambrosini:
There's a problem with the starring method, it does not work. Nor in master nor in this pr. Investigating the issue

marcoambrosini:
Those unquoted symbols don't look right @jancborchardt

jancborchardt (commented):
- The favorite indicator needs to show left in the subline still, no? Or is this related to your comment about the favoriting not working?
- I think the unquoted symbols are fine, the quotes are much too busy. If possible the symbols could be color-main-text (not bold) to make them stand out a bit.
- Not keeping a left gap for the avatars (and not indenting the messages) just looks off and too busy. If you look at Discord mobile for example, they also habe a gap on the left. Additionally in Talk, we sticky the avatar so on long messages you still have the context, which is great. So we should keep that. As discussed, the issue in Talk is more the space reserved for the timestamp on the right. So let's look at that separately.

marcoambrosini:
> favorite indicator needs to show left in the subline still, no

yes, done!

szaimen:
Do we display the favorited state now anywhere in the sidebar? If not, would that be a regression?

marcoambrosini:
Yes @simison, in the subline 

Screenshot 2023-11-15 at 18 27 25

szaimen:
> Yes @szaimen, in the subline

ah nice! Then looks good to me from the screenshots 👍

szaimen (approved):

apps/files/src/views/Sidebar.vue
```diff
@@ -246,7 +261,6 @@ export default {
 					},
 					compact: this.hasLowHeight || !this.fileInfo.hasPreview || this.isFullScreen,
 					loading: this.loading,
-					starred: this.fileInfo.isFavourited,
```
susnux, line 249: We should then decide if we simply want to drop this in the library as well.

apps/files/src/views/Sidebar.vue
```diff
@@ -189,7 +203,8 @@ export default {
 		 * @return {string}
 		 */
 		subtitle() {
-			return `${this.size}, ${this.time}`
+			const starredIndicator = this.fileInfo.isFavourited ? '★ ' : ''
```
susnux, line 206:
```suggestion
			const starredIndicator = this.fileInfo.isFavourited ? '⭐ ' : ''
```

If we want to make it more discoverable / prominent that the file is a favorite we might want to use this instead.

susnux (approved):
LGTM

apps/files/src/views/Sidebar.vue
AndyScherzinger: Is this accessible? The grey one would meet the contrast requirements I would hope.

CC @JuliaKirschenheuter

jancborchardt: I think we would have to use an icon with alt text / aria-label, right @JuliaKirschenheuter @marcoambrosini?

susnux: Well we have `--icon-starred-yellow` as an icon that fulfills accessibility requirements for bright and dark mode.

marcoambrosini: The problem is that the NcAppSidebar only accepts a string in that subline. I suggest we add the text starred beside the star, so a screen reader would read out the info too. 
Contrast is ok, yes.

jancborchardt: @marcoambrosini would that work space-wise and with translations? Otherwise it might be better to have the indicator as an icon in the top right of the file preview/icon as it was before, just not as big (and with an outline). Then it matches how it looks like in the file list.

marcoambrosini: I will add the text only in the aria label and not in the paragraph. We can't add just an icon to the sidebar preview. It's either the toggle button or nothing.

marcoambrosini: Also not possible, there it's either the switch or nothing. I suggest we go ahead with this and we modify the sidebar component so that we can slot in a proper icon there as @susnux suggests

JuliaKirschenheuter: sorry for a stupid question, but why we can't just use an normal material icon with `aria-label` in that place?

jancborchardt: > Also not possible, there it's either the switch or nothing.

Why is this not possible? Since the current solution is not accessible ("star" is not understandable as "favorite", and we can not add an aria-label) we need to do it otherwise.

Can we please do:
- Just use an icon on the top right of the filetype icon, non-interactive like in the file list
- The action to toggle it would be in the action menu, like you did here

@marcoambrosini @szaimen

JuliaKirschenheuter: > Just use an icon on the top right of the filetype icon, non-interactive like in the file list

Screenshot from 2023-11-22 13-44-55

and we have already everything we need

ShGKme: > Since the current solution is not accessible ("star" is not understandable as "favorite", and we can not add an aria-label) we need to do it otherwise.

I don't think that this is not accessible.

Emoji is read by a screen reader. It could be unclear that "star" means "favorite". But it is the same for all users. Any user needs to understand that "star" is "a starred file" which means "favorite".

This would be good to have a description here. But not required if we appeal to a common user experience of "starring".

jancborchardt: Ok, thanks for the assessment @ShGKme – then @AndyScherzinger and I would agree we should merge this as-is for now.

In the long term however it could possibly be better to have an image with aria-label which is more explicit, which we could also place on the top right of the filetype icon so it is symmetric to the files list.

jancborchardt (approved):

AndyScherzinger (approved):

PR #41488, opened 2023-11-15
fix(dav): Remove shares of inexistent sharers

ChristophWurst:
* Resolves: # 

## Summary

`occ dav:remove-invalid-shares` removes shares when the share recipient does not exist (anymore). This also adds a check for the sharers. If they don't exist, shares will be dropped.

## TODO

- [x] Do

## Checklist

### DIFF
--- a/apps/dav/lib/Command/RemoveInvalidShares.php
+++ b/apps/dav/lib/Command/RemoveInvalidShares.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace OCA\DAV\Command;
 
 use OCA\DAV\Connector\Sabre\Principal;
+use OCP\Cache\CappedMemoryCache;
 use OCP\IDBConnection;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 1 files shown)


apps/dav/lib/Command/RemoveInvalidShares.php
```diff
@@ -37,44 +38,95 @@
...
-			$p = $this->principalBackend->getPrincipalByPath($principaluri);
-			if ($p === null) {
-				$this->deleteSharesForPrincipal($principaluri);
+			if (!isset($this->cache[$principaluri])) {
+				$exists = $this->cache[$principaluri] = $this->principalBackend->getPrincipalByPath($principaluri) !== null;
+			} else {
+				$exists = $this->cache[$principaluri];
+			}
+			if (!$exists) {
+				$output->writeln("Principal URI $principaluri is an invalid share recipient", OutputInterface::VERBOSITY_VERBOSE);
```
tcitworld, line 73: If the `dry-run` option is given it should probably be set to `VERBOSITY_NORMAL`.

tcitworld (commented):
Is using a cache worth it rather than just using the class properties to keep track of key/values booleans ?

blizzz:

ChristophWurst:
This was not needed in the end

PR #41491, opened 2023-11-15
feat(comments): Plug in comments into activity sidebar tab if available

### LINKED ISSUE #658: 📑 Sidebar: combine file Activity and Comments into »Activity« timeline tab (possibly also "Versions")
Basically as written up and with mockups at https://github.com/owncloud/core/issues/20259
We need to finally streamline the sidebar instead of putting more and more stuff in there.
sidebar-mockup

cc @nextcloud/javascript @rullzer @MorrisJobke @schiessle

susnux:
* Resolves: #658 
* Replaces: #41483 
* Requires: https://github.com/nextcloud/activity/pull/1416

## Summary
This is a more stable approach than #41483, if activity is enabled we plug into their stream and provide the comments.

### Screenshot

https://github.com/nextcloud/server/assets/1855448/d414c961-a7c6-4b73-bdc8-f5724e2e7aa0

## Checklist


AndyScherzinger (approved):

artonge (approved):

nickvergessen (changes requested):
This breaks the comments API when used by other apps again, as pointed out in the previous PR

nickvergessen:
> Did you test this?

yes, something is hardcoding the object type to `files` and try tomorrow again (or you install https://github.com/nextcloud/announcementcenter and try it yourself)

susnux:
OK I know whats going on here... Have I ever said I hate mixins?
I had some weird issues were the type was not set, so I set it to files, well turns out the comments API sets it using mixins syntax from hell. I will fix this.

susnux:
Should be fixed now @nickvergessen 

<details>
I really do not like code that hides what happens, in this case the comment type was injected only in one of the entry points as as a Mixin (**!**) instead of declaring it a prop so you know that it is there are need to be set / can be set.
</details>

apps/comments/src/comments-tab.js
```diff
@@ -22,40 +22,53 @@
...
-	id: 'comments',
-	name: t('comments', 'Comments'),
-	iconSvg: MessageReplyText,
+// @ts-expect-error __webpack_nonce__ is injected by webpack
+__webpack_nonce__ = btoa(getRequestToken())
 
-	async mount(el, fileInfo, context) {
-		if (TabInstance) {
+if (loadState('comments', 'activityEnabled', false) && OCA?.Activity?.registerSidebarAction !== undefined) {
+	// Do mount own tab but mount into activity
```
come-nc, line 33:
```suggestion
	// Do not mount own tab but mount into activity
```
Typo?

artonge:
Rebased and fixed come's comment

nickvergessen:
Public API still doesn't work here. But now we have a different error:
grafik

apps/comments/src/mixins/CommentMixin.js
```diff
@@ -39,6 +40,10 @@ export default {
 			type: [String, Number],
 			required: true,
 		},
+		commentsType: {
```
nickvergessen, line 43: ressourceId is the counter part/ID to this, so ressourceType would be more matching?

apps/comments/src/mixins/CommentView.ts
```diff
@@ -0,0 +1,72 @@
...
+import { generateOcsUrl } from '@nextcloud/router'
+import { defineComponent } from 'vue'
+
+export default defineComponent({
+	props: {
+		ressourceId: {
+			type: Number,
+			required: true,
+		},
+		commentsType: {
```
nickvergessen, line 13:
```suggestion
		ressourceType: {
```

apps/comments/src/mixins/CommentView.ts
```diff
@@ -0,0 +1,72 @@
...
+		 * Autocomplete @mentions
+		 *
+		 * @param {string} search the query
+		 * @param {Function} callback the callback to process the results with
+		 */
+		async autoComplete(search, callback) {
+			const { data } = await axios.get(generateOcsUrl('core/autocomplete/get'), {
+				params: {
+					search,
+					itemType: 'files',
```
nickvergessen, line 39: can we use the `ressourceType` here?

apps/comments/src/services/CommentsInstance.js
```diff
@@ -51,15 +51,14 @@ export default class CommentInstance {
...
 	constructor(commentsType = 'files', options) {
-		// Add comments type as a global mixin
-		Vue.mixin({
-			data() {
-				return {
-					commentsType,
-				}
+		// Merge options and set `commentsType` property
+		options = {
+			...options,
```
nickvergessen, line 56: undefined is not handled here.

nickvergessen: When modifying my app from
```js
 new OCA.Comments.View('announcement')
```
to
```js
 new OCA.Comments.View('announcement', {})
```

it works, but maybe we can avoid this breaking

apps/comments/src/views/Comments.vue
```diff
@@ -28,8 +28,9 @@
 		<!-- Editor -->
 		<Comment v-bind="editorData"
 			:auto-complete="autoComplete"
-			:user-data="userData"
+			:comments-type="commentsType"
```
nickvergessen, line 31: also resource-type and resourceType everywhere here?

nickvergessen (changes requested):

apps/comments/src/mixins/CommentView.ts
susnux: Did not change it, the name was used before

susnux: But I can change it if you like

apps/comments/src/mixins/CommentMixin.js
artonge: `commentsType` was already used in the code. Let's keep it like that

apps/comments/src/mixins/CommentView.ts
artonge: Let's keep it like that

apps/comments/src/services/CommentsInstance.js
susnux: Well this only worked by luck as the "API" of comments never had that part optional
`* @param  {object} options the vue options (propsData, parent, el...)`
instead of 
`* @param  {object|undefined} options the vue options (propsData, parent, el...)`

But yes I can make this optional.

artonge: Will push a fix

apps/comments/src/views/Comments.vue
artonge: As said, outside of the scope of this PR

artonge:
@nickvergessen I tested with annoncement center. Should be fine now :)

apps/comments/src/mixins/CommentMixin.js
susnux: Did not change any naming there but yes can do so

susnux:
@nickvergessen resolved you comments should be fine now

nickvergessen:
> Oh we did this simultaneously :/

All hail to `git push --force-with-lease` I guess

nickvergessen:
Now the hardcoded `files` is back? 
`/remote.php/dav/comments/files/1`

nickvergessen:
> Now the hardcoded files is back?

Only on posting, loading old comments works properly

susnux:
> Now the hardcoded files is back?

There is no hardcoded 'files', only if you do not set the appropriate `ressourceType` it falls back to files by default.

Here:
https://github.com/nextcloud/announcementcenter/blob/91188b48845c3b2175eefd4fd4ecff2036a7d70e/src/App.vue#L150C35-L150C62

No resource type is set, what ressource type do you expect it to use?

nickvergessen:
> No resource type is set, what ressource type do you expect it to use?

The first argument is the resource type!?

nickvergessen:
There were some mixing of the property name making it fallback to the default `'files'`
I added a commit (and merged the recompile into the previous commit)

But now it works in files and announcements.

nickvergessen (approved):

susnux:
Cypress is unrelated (flaky user-creation test)

Issue #41381, opened 2023-11-10
Global search beta follow up


Fenn-CS:
Follow up issues from https://github.com/nextcloud/server/pull/40823

### TODO

**High priority:**
- [x] Typing in the user search filter doesn't show all the users, it seems that it shows only the preloaded ones
- [x] When filtering the local view (e.g. users) it's impossible to reset the view. When reopening the search modal, it should be possible to reset the search.
- [x] Search result links that should be improved to be actual links for a11y reasons

**Other:** 
- [x] Icon inversion which does not work well now since icons come in white or black instead of one consistent color. (Black icons turn white and white icons turn black, leading to to white on white background issues)
- [x] Use the new people filter (selector) that @marcoambrosini is working on (Current people filter not completed) - DONE, however, NcPopover which is used has issues with closing triggers.
- [x] Update back-end API (if ability not already present) to show "Search in App" dynamically.
- [x] Remove person filter
- [x] BUG: Clicking outside the search modal needs two clicks on the magnifier icon to get the modal back.
- [x] BUG: Some app icons are white on white.
- [x] BUG: Modified timestamp bubbles change size on hover, have no pointing cursor.
- [x] BUG: Generally no feedback for no results.
- [x] BUG: Search people has white text on white background.
- [x] BUG: Search people has a tiny user bubble, barely wider than the avatar.
- [x] BUG: Search people No result goes to the top, overlaps other input.
- [x] BUG: Search people has no effect at all. The search string doesn't contain a person identifier.
- [x] Avatars are square. I guess that's because all previews are square so this makes sense *technically*. For a user, this is inconsistent with *avatars*, though
    ![image](https://github.com/nextcloud/server/assets/1374172/5dfcef83-ae9b-47cf-b861-aa79f65c34e2)
- [x] When previews do not exist (endpoint returns 404) the browser renders a placeholder for the broken image
    ![Bildschirmfoto vom 2023-11-14 14-23-13](https://github.com/nextcloud/server/assets/1374172/c677f3c0-38a4-42e3-8faa-a632326d2e2d)
- [x] *x* on apps/person selection changes size on hover
- [x] No space for icons in modified (Fenn-CS : Best handled at component level, migrate to vue lib repo? "Do not reserve space for icon in NcActionButton if no icon is provided"???)
- [x] *x* on apps/person selection should have pointing cursor

[...]

PR #41507, opened 2023-11-15
Enhancements: Fix and updates to most recent global search UI

### LINKED ISSUE #41492: Global search filter for "since" and "until" are broken
### How to use GitHub

---

* Today filters with since=1700048245 (NOW) and until=1700048245 (NOW) (same timestamp, should range from 0:00 this day until 23:59 this day) - basically will never find something at the moment
* Last 7 days filters with since=1699443555 (NOW - 7\*24h) and until=1700048355 (NOW), the range is 7\*24h instead of the beginning of that day (is at least unexpected for me)
* This year  filters with since=1672527600 (1.1. 0:00 - Ok) and until=1700048599 (NOW), which might be problematic for things like a calendar where you would also like to find future events of THIS year.
* Custom range, I select "yesterday and today"  with since=1699916400 (14.11. 0:00 - Ok) and until=1700002800 (15.11. 0:00) - I would expect today to be included, but it actually searches only yesterday at all
 grafik

nfebe:
- Allow future dates in custom date range
- Fix date ranges for example, before now, today range was roughly between `now()` to `now() + a few milli seconds`

Resolves : https://github.com/nextcloud/server/issues/41492, https://github.com/nextcloud/server/issues/41381

## High
- [x] Date filter logic broken #41492 
- [x] Mail previews not right 
- [x] Broken folder icon in files results
- [x] Messages: avatar in circles not squares
- [x] “Search in app” is only available for a few apps (remove links)
- [x] Search field is not focused automatically when the modal is opened

## Medium
- [x] Contacts filter shouldn’t be available if contacts app is not available 
- [ ] No space for icons in modified

## Low
- [x] Apply range primary to the right 
- [x] Date range filter -> custom date range
- [x] Apply range -> Search in date range
- [ ] "No matching results" emptycontent icon should be a search icon instead of the current one
- [ ] Text in People" menu regular not bold
- [ ] Close icon material design icon "close" instead of "close-thick"
- [x] Cursor pointer instead on hovering close icon
- [x] Avatar in filter chips shouldn't be clickable and size should be 24px


Fenn-CS:
This should get reviewed and merged, so smaller follow up prs get in faster, it makes things faster and gets as much as possible into the regular build schedule.

artonge (approved):

core/src/components/GlobalSearch/CustomDateRangeModal.vue
```diff
@@ -1,33 +1,33 @@
 <template>
 	<NcModal v-if="isModalOpen"
 		id="global-search"
-		:name="t('core', 'Date range filter')"
+		:name="t('core', 'Custom date range')"
 		:show.sync="isModalOpen"
 		:size="'small'"
 		:clear-view-delay="0"
-		:title="t('Date range filter')"
+		:title="t('Custom date range')"
```
susnux, line 8:
```suggestion
		:title="t('core', 'Custom date range')"
```

susnux (changes requested):
missing appkey for translation, this will not work :/

core/src/components/GlobalSearch/SearchResult.vue
```diff
@@ -0,0 +1,162 @@
+<template>
+	<NcListItem class="result-items__item"
+		:name="title ?? ''"
+		:bold="false"
+		@click="openResult(result)">
+		<template #icon>
+			<div class="result-items__item-icon"
```
susnux, line 7: Either this or set `alt=""` on the `` tag. But I think hiding it is correct here as this is pure representation (images).

```suggestion
			<div aria-hidden="true"
				class="result-items__item-icon"
```

core/src/components/GlobalSearch/SearchResult.vue
```diff
@@ -0,0 +1,162 @@
...
+		 * enter still opens the first result
+		 */
+		focused: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	data() {
+		return {
+			thumbnailHasError: false,
```
susnux, line 78: Maybe we need a:
```js
watch: {
    thumbnailUrl() {
        this.thumbnailHasError = false
    }
}
```

Because if this component is recycled because of changing results and the thumbnail has an error, the new result will also show the thumbnail error.
If this happens depends on whether the component will be recycled by `GlobalSearchModal` or not (depending on the `key` value in the `v-for` look).

susnux (approved):

PR #41511, opened 2023-11-15
Enhancements for advanced search API

Altahrim:
- `/` isn't allowed in UID, it will avoid conflicts
- undeprecate `getTerm` on SearchQuery
- introduce constants for builtin filters

### DIFF
--- a/lib/private/Search/FilterFactory.php
+++ b/lib/private/Search/FilterFactory.php
@@ -32,6 +32,8 @@ use OCP\IUserManager;
 use RuntimeException;
 
 final class FilterFactory {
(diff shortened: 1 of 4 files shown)


nickvergessen (approved):
👍 but where is it generated like this? Need a second place that should use the constant?

Altahrim:
> but where is it generated like this?

That will be in frontend. For now it can be used with the API

nickvergessen (approved):

AndyScherzinger (approved):

come-nc (approved):

PR #41512, opened 2023-11-15
fix(userstatus): Only track message timestamp for values

ChristophWurst:
* Resolves: # 

## Summary

Do not track status messages that are empty, e.g. DND outside of availability.

## How to test

1) Set availability and DND automation, make sure current time is *outside* working hours
2) Run background jobs in case the status isn't propagated right away
3) Open the dashboard with another user

master: ", seconds ago" shows for the main user
here: no status update is shown on the dashboard because the user didn't change their status.

## Checklist

### DIFF
--- a/apps/user_status/lib/Db/UserStatus.php
+++ b/apps/user_status/lib/Db/UserStatus.php
@@ -50,7 +50,7 @@ use OCP\AppFramework\Db\Entity;
  * @method void setCustomMessage(string|null $customMessage)
  * @method int|null getClearAt()
  * @method void setClearAt(int|null $clearAt)
- * @method setIsBackup(bool $true): void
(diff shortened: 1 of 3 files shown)


apps/user_status/lib/Db/UserStatus.php
```diff
@@ -50,7 +50,7 @@
  * @method void setCustomMessage(string|null $customMessage)
  * @method int|null getClearAt()
  * @method void setClearAt(int|null $clearAt)
- * @method setIsBackup(bool $true): void
+ * @method setIsBackup(bool $isBackup): void
```
nickvergessen, line 53: opsy

nickvergessen (approved):

Antreesy (approved):

Issue #41364, opened 2023-11-09
Migrate remaining setup checks to new API


come-nc:
Remaining checks not using the new API:

- [x] isBruteforceThrottled (Your remote address was identified as "{remoteAddress}" and is bruteforce throttled at the moment slowing down the performance of various requests…) https://github.com/nextcloud/server/pull/41435
- [x] forwardedForHeadersWorking (The reverse proxy header configuration is incorrect, or you are accessing Nextcloud from a trusted proxy. If not, this is a security issue and can allow an attacker…) https://github.com/nextcloud/server/pull/41438
- [x] OpcacheSetupRecommendations (The PHP OPcache module is not properly configured…) https://github.com/nextcloud/server/pull/41547
- [x] missingIndexes (The database is missing some indexes…) https://github.com/nextcloud/server/pull/41460
- [x] missingPrimaryKeys (The database is missing some primary keys…) https://github.com/nextcloud/server/pull/41460
- [x] missingColumns (The database is missing some optional columns…) https://github.com/nextcloud/server/pull/41460
- [x] isImagickEnabled (The PHP module "imagick" is not enabled although the theming app is) https://github.com/nextcloud/server/pull/42177
- [x] imageMagickLacksSVGSupport (Module php-imagick in this instance has no SVG support) https://github.com/nextcloud/server/pull/42177
- [x] pendingBigIntConversionColumns (Some columns in the database are missing a conversion to big int…) https://github.com/nextcloud/server/pull/41468
- [x] isSqliteUsed (SQLite is currently being used as the backend database…) ->  https://github.com/nextcloud/server/pull/41535
- [x] isMemoryLimitSufficient https://github.com/nextcloud/server/pull/41086
- [x] appDirsWithDifferentOwner (Some app directories are owned by a different user than…) https://github.com/nextcloud/server/pull/42176
- [x] hasPassedCodeIntegrityCheck https://github.com/nextcloud/server/pull/41312
- [x] suggestedOverwriteCliURL (Please make sure to set the "overwrite.cli.url" option in your config.php file to the URL that your users mainly use to access this Nextcloud…) https://github.com/nextcloud/server/pull/42796
- [x] cronErrors (It was not possible to execute the cron job via CLI. The following technical errors have appeared:…) https://github.com/nextcloud/server/pull/42802
- [x] isUsedTlsLibOutdated https://github.com/nextcloud/server/pull/42807
[...]

PR #41535, opened 2023-11-16
Merge sqlite check to database check

come-nc:
* See https://github.com/nextcloud/server/issues/41364

## Summary

Merge the SQlite setup check into the already existing supported database check.

## Checklist

### DIFF
--- a/apps/settings/lib/SetupChecks/SupportedDatabase.php
+++ b/apps/settings/lib/SetupChecks/SupportedDatabase.php
@@ -33,12 +33,14 @@ use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
 use OCP\IDBConnection;
 use OCP\IL10N;
+use OCP\IURLGenerator;
 use OCP\SetupCheck\ISetupCheck;
 use OCP\SetupCheck\SetupResult;
 
 class SupportedDatabase implements ISetupCheck {
 	public function __construct(
 		private IL10N $l10n,
+		private IURLGenerator $urlGenerator,
 		private IDBConnection $connection,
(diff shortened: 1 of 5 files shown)


apps/settings/lib/SetupChecks/SupportedDatabase.php
```diff
@@ -81,7 +83,10 @@ public function run(): SetupResult {
 		} elseif ($databasePlatform instanceof OraclePlatform) {
 			$version = 'Oracle';
 		} elseif ($databasePlatform instanceof SqlitePlatform) {
-			$version = 'Sqlite';
+			return SetupResult::warning(
+				$this->l10n->t('SQLite is currently being used as the backend database. For larger installations we recommend that you switch to a different database backend. This is particularly recommended when using the desktop client for file synchronisation. To migrate to another database use the command line tool: "occ db:convert-type".'),
```
Altahrim, line 87: Can we add this warning only if the instance have more than 5 users?
I am thinking about a small arbitrary limit so an admin of a very small instance won't have this warning.

Altahrim (approved):

apps/settings/lib/SetupChecks/SupportedDatabase.php
nickvergessen: https://docs.nextcloud.com/server/latest/admin_manual/installation/system_requirements.html
> SQLite (only recommended for testing and minimal-instances)

I think warning is totally fine for "production" instances. It always can have negative performance impact.
We also warn about int32 when you are alone, not having a memcache (info only), missing indexes, etc.

I think warning against SQLite is totally fine as it's really recommended to go with anything else.

nickvergessen (approved):

PR #41540, opened 2023-11-16
Add metadata generation job

artonge:


ArtificialOwl (approved):

core/BackgroundJobs/GenerateMetadataJob.php
```diff
@@ -0,0 +1,111 @@
...
+		if ($folder->getMountPoint() instanceof \OC\Files\Mount\MoveableMount) {
+			return;
+		}
+
+		foreach ($folder->getDirectoryListing() as $node) {
+			if ($node instanceof Folder) {
+				$this->scanFolder($node);
+				continue;
+			}
+
```
ArtificialOwl, line 104: what if you `getMetadata()` first and see if it exists or not, and `continue;` if it exists ?

artonge: It could be partial ?

ArtificialOwl: partially generated ? it's fine, the script is not about updating metadata but creating

artonge: The script is about filling the metadata with useful content. If the metadata are partial, then it is not as useful.

come-nc (approved):

PR #41544, opened 2023-11-16
do not write htaccess file if disk space is too low

szaimen:
I copied the logic from https://github.com/nextcloud/server/blob/1b18b976b45b155f005676640e0d49ca7830b8b0/lib/private/Config.php#L288-L295

### DIFF
--- a/lib/private/Setup.php
+++ b/lib/private/Setup.php
@@ -560,6 +560,14 @@ class Setup {
 		}
 
 		if ($content !== '') {
+			// Never write file back if disk space should be too low
+			if (function_exists('disk_free_space')) {
+				$df = disk_free_space(\OC::$SERVERROOT);
+				$size = strlen($content) + 10240;
+				if ($df !== false && $df < (float)$size) {
+					throw new \Exception(\OC::$SERVERROOT . " does not have enough space for writing the htaccess file! Not writing it back!");
+				}
+			}
 			//suppress errors in case we don't have permissions for it
 			return (bool)@file_put_contents($setupHelper->pathToHtaccess(), $htaccessContent . $content . "\n");
 		}
--- a/core/Command/Maintenance/UpdateHtaccess.php
+++ b/core/Command/Maintenance/UpdateHtaccess.php
@@ -39,7 +39,7 @@ class UpdateHtaccess extends Command {
 			$output->writeln('.htaccess has been updated');
 			return 0;
 		} else {
-			$output->writeln('<error>Error updating .htaccess file, not enough permissions or "overwrite.cli.url" set to an invalid URL?</error>');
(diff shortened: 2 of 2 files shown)


kesselb:
Why?

lib/private/Setup.php
```diff
@@ -560,6 +560,14 @@ public static function updateHtaccess() {
 		}
 
 		if ($content !== '') {
+			// Never write file back if disk space should be too low
+			if (function_exists('disk_free_space')) {
+				$df = disk_free_space(\OC::$SERVERROOT);
+				$size = strlen($content) + 10240;
+				if ($df !== false && $df < (float)$size) {
+					throw new \Exception(\OC::$SERVERROOT . " does not have enough space for writing the htaccess file! Not writing it back!");
```
kesselb, line 568: Either document the exception or change to return false. I would prefer the latter and extend the warning in https://github.com/nextcloud/server/blob/master/core/Command/Maintenance/UpdateHtaccess.php

szaimen:
> Why?

Because the file got truncated on @tobiasKaminsky's instance due to low disk space...

lib/private/Setup.php
kesselb: Hmm. The code is used for install (web, cli), updater (web, cli) and own cli command. If you throw the exception here, it will abort the install process. The updater case will catch the exception and show it, the command does not catch it. 

 I assume that is more or less expected / good? If so please keep the exception but document it. Thanks.

szaimen: > I assume that is more or less expected / good? 

yes

> If so please keep the exception but document it. Thanks.

done

tobiasKaminsky (approved):

lib/private/Setup.php
```diff
@@ -560,6 +560,14 @@ public static function updateHtaccess() {
...
+				}
+			}
 			//suppress errors in case we don't have permissions for it
 			return (bool)@file_put_contents($setupHelper->pathToHtaccess(), $htaccessContent . $content . "\n");
```
Altahrim, line 572: You can also check if all bytes have been written. Should be something like this:
```suggestion
			$fullContent = $htaccessContent . $content . "\n";
			return @file_put_contents($setupHelper->pathToHtaccess(), $$fullContent) === strlen($fullContent);
```

szaimen: This is not going to help here. The prpblem is that file_pu_content truncates the file if there is not enough space which should never happen as the file is very important to work...

szaimen: We had the same prpblem with config.php before and fixed in in the same way.

Altahrim: It's another security in case something happen during the copy.
The file will already be erased, but at least you'll have the info

Altahrim (approved):

PR #41547, opened 2023-11-16
Migrate opcache check to new SetupCheck API

come-nc:
* See https://github.com/nextcloud/server/issues/41364

## Summary

Migrate opcache check to new SetupCheck API

## TODO

- [x] Decide what to do with `<code>` tags.

## Checklist

### DIFF
--- a/apps/settings/lib/SetupChecks/PhpOpcacheSetup.php
+++ b/apps/settings/lib/SetupChecks/PhpOpcacheSetup.php
@@ -67,20 +67,20 @@ class PhpOpcacheSetup implements ISetupCheck {
 		}
 
 		if (!$this->iniGetWrapper->getBool('opcache.enable')) {
-			$recommendations[] = $this->l10n->t('OPcache is disabled. For better performance, it is recommended to apply <code>opcache.enable=1</code> to your PHP configuration.');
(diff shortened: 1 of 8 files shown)


apps/settings/lib/SetupChecks/PhpOpcacheSetup.php
```diff
@@ -0,0 +1,147 @@
...
+			}
+		}
+
+		return $recommendations;
+	}
+
+	public function run(): SetupResult {
+		$recommendations = $this->getOpcacheSetupRecommendations();
+		if (!empty($recommendations)) {
+			return SetupResult::error(
```
Altahrim, line 126: I think recommendations about cache size can be warning only. 
`save_comment=0` seems to be the only real error here, but I would also include disabled opcache.

come-nc: I just used the same level as before.
Can you comment inline on recommendations which you think should be warning or error?
Then I can adapt this to a two-level check.

Altahrim (approved):

artonge (approved):
Anything changed beside moving code and adapting tests?

come-nc:
Failure not related, see https://github.com/nextcloud/server/pull/42633

come-nc:
> Anything changed beside moving code and adapting tests?

Yes, level is error or warninig depending on which recommendation is triggered, saw that with Benjamin.

PR #41565, opened 2023-11-17
fix(ObjectStore): Make copying behavior consistent with local storage

### LINKED ISSUE #41564: [Bug]: Inconsistent copy behavior between Object storage and local storage

### Bug description

If you copy a shared file on local storage the copied file has dropped all permissions, meaning you are the owner and have all permissions.
On object storage if you copy a shared file, the permissions are preserved, meaning if you could only view a file you can also only view the copied file.

### Steps to reproduce

1. Share a file and set permissions to view only
2. As the person with who you shared the file: Copy the file
3. See that on local storage you can reshare the copied file and on object storage you can not.

### Expected behavior

Consistent behavior (one of both).

### Installation method

None

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

susnux:
* Resolves: #41564 

## Summary
Drop file permissions on copy like we do on local storage.

## Checklist

### DIFF
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -68,6 +68,8 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common implements IChunkedFil
 
 	private $logger;
 
+	private bool $handleCopiesAsOwned;
+
 	/** @var bool */
 	protected $validateWrites = true;
 
@@ -88,6 +90,7 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common implements IChunkedFil
 		if (isset($params['validateWrites'])) {
 			$this->validateWrites = (bool)$params['validateWrites'];
 		}
+		$this->handleCopiesAsOwned = (bool)($params['handleCopiesAsOwned'] ?? false);
 
 		$this->logger = \OC::$server->getLogger();
 	}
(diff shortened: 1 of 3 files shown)


Altahrim (approved):

lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -88,6 +90,7 @@ public function __construct($params) {
 		if (isset($params['validateWrites'])) {
 			$this->validateWrites = (bool)$params['validateWrites'];
 		}
+		$this->handleCopiesAsOwned = (bool)($params['handleCopiesAsOwned'] ?? false);
```
juliushaertl, line 93: Do we need a parameter for that actually? If we keep it it should be documented

juliusknorr (approved):
Small comment, but looks good otherwise

joshtrichards (commented):
From your inline comments it seems you think the local  (non-object) policy is the broken one.  I'm not sure local's behavior is incorrect (seems object is), but I appreciate the conservativeness to add a config option to avoid changing existing behavior for object (in case someone is relying on it).

But I think it should just outright be made consistent with local as-is w/o config option if at all possible. There's just no justification for having different behavior (or a switch for one but not the other/etc)

Do we have any idea when this inconsistent behavior was introduced or has it seemingly "always" been this way?

AndyScherzinger:
> Do we have any idea when this inconsistent behavior was introduced or has it seemingly "always" been this way?

@joshtrichards Has always been like this, since native S3 copying respects the original file's permissions. One could argue that it does so for a reason because the permission had been the decision of the original owner of the file. At the same time, like you mentioned you could also argue the other way around that the copy belongs to the copying party and hence all permission-limitations are dropped.

icewind1991:
I think the new behavior is wrong, the permissions on object storages are not related to the permissions of the objects in the objectstore, but only the permissions from things like sharing/acls.

See https://github.com/nextcloud/server/pull/53733 for making the local storage behave as the object store used to.

PR #41571, opened 2023-11-17
Allow setting `strict-dynamic` on `strict-src-elem` and set it by default

susnux:
## Summary

Add a function to set 'strict-dynamic' to `script-src-elem` only which allows setting it with less weaken the CSP.
This is required for modern JS code that uses `import` which does not allow to use nonces (there is simply no way to set a nonce on import). Chrome then enforces the nonce rule and fails because there is none.
So instead of setting 'strict-dynamic' on every script source we only trust scripts provided with `<script>` tags by default and only if they have the nonce set.

## Checklist


susnux:
> [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

We added support for this with NC27 but I do not think we should backport the default change. Opinions?

juliusknorr (approved):

AndyScherzinger (approved):

susnux:
Fixed one test I forgot to update.

susnux:
Documentation: https://github.com/nextcloud/documentation/pull/11291

PR #41573, opened 2023-11-17
fix(files): avoid sending thumbnail URL for directory as none exists

Altahrim:
Avoid sending a thumbnail URL for directories since the URL ends up in HTTP 404


ChristophWurst (approved):

susnux (approved):

susnux:
drone unrelated

PR #41578, opened 2023-11-17
Enable AppFramework dispatcher to enforce integer ranges

blizzz:
## Summary

Psalm allows to specify an int range for integer parameters. We can consider them in the Http Dispatcher so that incoming int values are checked for their range directly in the app framework. Example use case for dashboard added.

Set 28 as milestone, but could go into 29 only. 

## Checklist

### DIFF
--- a/lib/private/AppFramework/Utility/ControllerMethodReflector.php
+++ b/lib/private/AppFramework/Utility/ControllerMethodReflector.php
@@ -42,6 +42,7 @@ class ControllerMethodReflector implements IControllerMethodReflector {
 	public $annotations = [];
 	private $types = [];
 	private $parameters = [];
+	private array $ranges = [];
 
 	/**
 	 * @param object $object an object or classname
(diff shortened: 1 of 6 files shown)


lib/private/AppFramework/Utility/ControllerMethodReflector.php
```diff
@@ -54,26 +55,38 @@ public function reflect($object, string $method) {
...
 				}
 
-				$this->annotations[$annontation] = [$annotationValue];
+				$this->annotations[$annotation] = [$annotationValue];
 			}
 
 			// extract type parameter information
 			preg_match_all('/@param\h+(?P<type>\w+)\h+\$(?P<var>\w+)/', $docs, $matches);
 			$this->types = array_combine($matches['var'], $matches['type']);
+			preg_match_all('/@psalm-param\h+(?P<type>\w+)<(?P<rangeMin>(-?\d+|min)),\h*(?P<rangeMax>(-?\d+|max))>\h+\$(?P<var>\w+)/', $docs, $matches);
```
provokateurin, line 78: This also has to work on the normal `@param`

provokateurin (changes requested):

lib/private/AppFramework/Utility/ControllerMethodReflector.php
provokateurin: Well psalm happily parses it. I'm not sure which standard we try to follow

provokateurin: But you are right, we should implement it in openapi-extractor to have it actually usable.

blizzz: > Well psalm happily parses it. I'm not sure which standard we try to follow

psalm should not hijack it ;)

st3iny (approved):
Tested and works. Fancy stuff :sparkles:

nickvergessen:
>  ```diff
> -                        "description": "Limit number of result items per widget",
> +                        "description": "Limit number of result items per widget, not more than 30 are allowed",
> ```

Need to rebuild openapi with `composer run openapi`

nickvergessen (approved):

blizzz:
> > ```diff
> > -                        "description": "Limit number of result items per widget",
> > +                        "description": "Limit number of result items per widget, not more than 30 are allowed",
> > ```
> 
> Need to rebuild openapi with `composer run openapi`

```
$ composer i && composer run openapi
…
Script "openapi" is not defined in this package
```

:panda_face:

blizzz:
```build/openapi-checker.sh``` it shall be

blizzz:
@provokateurin yeah or no?

provokateurin (approved):
Sorry, I thought this was solved already. I opened an issue in openapi-extractor to handle the `@psalm-*` annotations and there is another one already for supporting these integer ranges. So at some point this will also properly show up in the openapi docs :)

provokateurin:
(Although I think it is debatable if 30 is a good limit since it is still quite high :woman_shrugging:)

provokateurin:
And 0 as lower limit is also not useful, maybe set it to at least 1

blizzz:
Checked by with @st3iny, he also thinks 30 is a good value.

Either way, the numbers are mostly gut estimations, if necessary we can always tune them.

The 0  I increase to 1.

PR #41583, opened 2023-11-17
fix(CSP): Only add `strict-dynamic` when using nonce

susnux:
## Summary

Using `strict-dynamic` will invalidate every `self` allowed source, so it only makes sense together with nonce.
Otherwise browsers not supporting CSP 3.0 will break.

## Checklist


SystemKeeper:
Test with Talk on iOS and Talk on Android and both work with this PR again!

AndyScherzinger (approved):

juliusknorr (approved):

PR #41584, opened 2023-11-17
chore: Update `@nextcloud/vue` to 8.2.0

susnux:
## Summary

Required for accessibility.

## Checklist


ShGKme (approved):

AndyScherzinger (approved):

cypress/e2e/settings/users_groups.cy.ts
```diff
@@ -39,15 +39,15 @@ describe('Settings: Create groups', () => {
...
+			cy.get('input').should('exist').and('have.value', '')
 			// set the group name to foo
-			cy.get('input[placeholder="Group name"]').type(groupName)
+			cy.get('input').type(groupName)
 			// see that the group name is foo
-			cy.get('input[placeholder="Group name"]').should('have.value', groupName)
+			cy.get('input').should('have.value', groupName)
 			// submit the group name
-			cy.get('input[placeholder="Group name"] ~ button').click()
+			cy.get('input ~ button').click()
```
ShGKme, line 50: To fix the test, only removing `[role="menu"]` seems to be required. Then there is no need in `data-cy` attribute, which uses test less semantic and more coupled on implementation (manually marked place).

susnux: It would depend on internal implementation (class `action-item__popper`) which is bad.
So this flags the action where to look on a non implementation dependent way (we set the attribute in our code so we do not rely on the implementation of the action, except for the li element but this is required due to a bug with actions (the attribute will be assigned on the li tag and also forwarded to the input tag).

PR #41601, opened 2023-11-18
add named metadata event

ArtificialOwl:
This adds:

- a new event `MetadataNamedEvent`,
- an optional parameter to the existing `occ files:scan --generate-metadata [name]` command.

The idea is to provide a last param to the `IFilesMetadataManager::refreshMetadata()` that assign a name to the event.
Any app listening to the `MetadataNameEvent` will use `getName()` to confirm the event is related to itself and manage the refresh of metadata the same way it would manage the `Live` and/or `Background` Event.

This way, we can ask a refresh of the metadata from all files' from the cloud without loosing time refreshing all metadata provided by all apps.

### DIFF
--- a/apps/files/lib/Command/Scan.php
+++ b/apps/files/lib/Command/Scan.php
@@ -96,8 +96,9 @@ class Scan extends Base {
 			->addOption(
 				'generate-metadata',
 				null,
-				InputOption::VALUE_NONE,
-				'Generate metadata for all scanned files'
+				InputOption::VALUE_OPTIONAL,
+				'Generate metadata for all scanned files; if specified only generate for named value',
+				''
 			)
 			->addOption(
 				'all',
@@ -122,7 +123,7 @@ class Scan extends Base {
 			);
 	}
 
-	protected function scanFiles(string $user, string $path, bool $scanMetadata, OutputInterface $output, bool $backgroundScan = false, bool $recursive = true, bool $homeOnly = false): void {
+	protected function scanFiles(string $user, string $path, ?string $scanMetadata, OutputInterface $output, bool $backgroundScan = false, bool $recursive = true, bool $homeOnly = false): void {
 		$connection = $this->reconnectToDatabase($output);
 		$scanner = new \OC\Files\Utils\Scanner(
 			$user,
@@ -136,11 +137,12 @@ class Scan extends Base {
 			$output->writeln("\tFile\t<info>$path</info>", OutputInterface::VERBOSITY_VERBOSE);
 			++$this->filesCounter;
 			$this->abortIfInterrupted();
-			if ($scanMetadata) {
+			if ($scanMetadata !== null) {
 				$node = $this->rootFolder->get($path);
 				$this->filesMetadataManager->refreshMetadata(
 					$node,
-					IFilesMetadataManager::PROCESS_LIVE | IFilesMetadataManager::PROCESS_BACKGROUND
+					($scanMetadata !== '') ? IFilesMetadataManager::PROCESS_NAMED : IFilesMetadataManager::PROCESS_LIVE | IFilesMetadataManager::PROCESS_BACKGROUND,
+					$scanMetadata
 				);
 			}
 		});
@@ -221,6 +223,12 @@ class Scan extends Base {
 
 		$this->initTools($output);
 
+		// getOption() logic on VALUE_OPTIONAL
+		$metadata = null; // null if --generate-metadata is not set, empty if option have no value, value if set
+		if ($input->getOption('generate-metadata') !== '') {
+			$metadata = $input->getOption('generate-metadata') ?? '';
+		}
+
 		$user_count = 0;
 		foreach ($users as $user) {
 			if (is_object($user)) {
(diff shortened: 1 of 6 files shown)


ArtificialOwl:
One of its purpose would be to use the scan+metadata to replace `files_fulltextsearch`

artonge:
So existing providers need to all listen to this new event?

lib/private/FilesMetadata/FilesMetadataManager.php
```diff
@@ -98,8 +100,12 @@ public function refreshMetadata(
 		}
 
 		// if $process is LIVE, we enforce LIVE
+		// if $process is NAMED and a named is set, we go NAMED
+		// else BACKGROUND
 		if ((self::PROCESS_LIVE & $process) !== 0) {
 			$event = new MetadataLiveEvent($node, $metadata);
+		} elseif ((self::PROCESS_NAMED & $process) !== 0 && $namedEvent !== '') {
+			$event = new MetadataNamedEvent($node, $metadata, $namedEvent);
```
come-nc, line 108: So named event is always live, or always backgrounded?
Why the name is not an optional parameter for both `MetadataLiveEvent` and `MetadataBackgroundEvent`?

ArtificialOwl:
> So existing providers need to all listen to this new event?

no, this is purely optionnal; only if they want to called via a specific occ command. 
It should only be implemented if they want to provide a faster way to only add their own metadata on a scan

lib/private/FilesMetadata/FilesMetadataManager.php
ArtificialOwl: Because it would require a check condition on every Live/Background listeners

come-nc:
So the passed name does not need to match a metadata name, it’s like a tag used by some providers?

ArtificialOwl:
> So the passed name does not need to match a metadata name, it’s like a tag used by some providers?

yes, as we don't communicate via provider we cannot enforce provider_id

ArtificialOwl:
The idea behind this is mostly to remove the custom files scan from files_fulltextsearch and use metadata to store the status of the index of a document there

apps/files/lib/Command/Scan.php
```diff
@@ -140,7 +140,8 @@ protected function scanFiles(string $user, string $path, bool $scanMetadata, Out
 				$node = $this->rootFolder->get($path);
 				$this->filesMetadataManager->refreshMetadata(
 					$node,
-					IFilesMetadataManager::PROCESS_LIVE | IFilesMetadataManager::PROCESS_BACKGROUND
+					($scanMetadata !== '_LIVE_') ? IFilesMetadataManager::PROCESS_NAMED : IFilesMetadataManager::PROCESS_LIVE | IFilesMetadataManager::PROCESS_BACKGROUND,
+					$scanMetadata
```
come-nc, line 144: I’d feel a bit better if magic value `_LIVE_` was not passed there. It should also be moved to a const for clean code.

lib/private/FilesMetadata/FilesMetadataManager.php
```diff
@@ -98,8 +100,12 @@ public function refreshMetadata(
 		}
 
 		// if $process is LIVE, we enforce LIVE
+		// if $process is NAMED and a named is set, we go NAMED
+		// else BACKGROUND
 		if ((self::PROCESS_LIVE & $process) !== 0) {
 			$event = new MetadataLiveEvent($node, $metadata);
+		} elseif ((self::PROCESS_NAMED & $process) !== 0 && $namedEvent !== '') {
```
come-nc, line 107: If PROCESS_NAMED was passed with an empty name, it feels weird to emit `MetadataBackgroundEvent`.
It should either error out or emit `MetadataNamedEvent` with the empty name.

lib/public/FilesMetadata/Event/MetadataNamedEvent.php
```diff
@@ -0,0 +1,69 @@
...
+namespace OCP\FilesMetadata\Event;
+
+use OCP\Files\Node;
+use OCP\FilesMetadata\AMetadataEvent;
+use OCP\FilesMetadata\Model\IFilesMetadata;
+
+/**
+ * MetadataNamedEvent is an event similar to MetadataBackgroundEvent completed with a target name,
+ * used to identify the designed app to refresh file's metadata.
+ * Meaning that when using this event, your app should implement a filter on the event's registered name
```
come-nc, line 35: It should be explained that this is specific to the occ command

come-nc (approved):
I get the idea, I’m not entirely sure this won’t be cause for trouble.
Especially there is nothing to prevent name clash between applications.

lib/public/FilesMetadata/Event/MetadataNamedEvent.php
ArtificialOwl: comment rewritten !

lib/private/FilesMetadata/FilesMetadataManager.php
ArtificialOwl: right, I choose the latter

apps/files/lib/Command/Scan.php
ArtificialOwl: ok, I have made it cleaner, not using a placeholder

artonge (approved):
Feels highly specific, but I do not see any other solution.

PR #41609, opened 2023-11-20
feat: In app search

marcoambrosini:
## Summary
This is an attempt to reconcile app search and global search into a single component and user flow.

### Advantages of this approach:
* One search box in the UI is less confusing than 2;
* The search-box is on top of the content, no overlapping between search box container and search results

[Screencast from 2023-11-20 17-13-37.webm](https://github.com/nextcloud/server/assets/26852655/f9ad3921-5c72-4a37-a203-406a2811b2c1)

## On mobile

| Collapsed | Expanded |
|----------|----------|
| Screenshot from 2023-11-20 17-56-55 | Screenshot from 2023-11-20 17-56-50|

## TODO

- [x] Limit this to apps that accept filtering
- [ ] Use proper app name for placeholder
- [x] Expand transition for search box


marcoambrosini:
@szaimen we'll have to address proper in app search strategy this for the next major. @jancborchardt had usability concerns about this solution. I think it's a pretty solid starting point compared to other alternatives, but of course I'm happy to iterate over this or explore other solutions and see how they compare.

szaimen:
> @jancborchardt had usability concerns about this solution. 

What kind of usability concerns?

> I think it's a pretty solid starting point compared to other alternatives

Indeed, I think it is much better in terms of usability than the current situation.

marcoambrosini:
Closing as there's no need for a last minute solution anymore, and it seems that we all favor proper in app search that's tailored to the respective apps and thus slightly different in each case.

jancborchardt:
@marcoambrosini I still think this could be a good combination of in-app and global search. It would also move us closer to the usual "big search field in the top center" that people know from Google and Microsoft et al. This was also a big point why you favored this approach, no @marcoambrosini?

Also cc @andyscherzinger since we recently talked about search.

szaimen:
> @marcoambrosini I still think this could be a good combination of in-app and global search. It would also move us closer to the usual "big search field in the top center" that people know from Google and Microsoft et al. This was also a big point why you favored this approach, no @marcoambrosini?

I agree 👍

szaimen:
I just tested this branch locally and it doesnt seem to work... @marcoambrosini would you mind updating it? :)

susnux:
@szaimen there were some "heavy" adjustments needed to bring this API wise to a consistent state (not 100% related to this feature but we basically needed to split out the event handling to the main view and only handle the unified global search in the modal).

I tested it locally and it seems to work as expected, some styles are broken due to the clickable area refactoring, but will be fixed soon I think with new nextcloud-vue version.

[Bildschirmaufnahme_20240625_211909.webm](https://github.com/nextcloud/server/assets/1855448/5b1b71b8-51b8-4423-a504-f58ca0b21671)

szaimen:
Thanks Ferdinand! Looks already much better. Only two comments:
- It should take the whole header on mobile: 
image 
image
- Can the button to search globally expand on desktop with the wording `Search everywhere` like it was initially done by Marco? :)
image
image

susnux:
> It should take the whole header on mobile

This are only styles so yes possible.

> Can the button to search globally expand on desktop with the wording Search everywhere like it was initially done by Marco?

The button is currently hacked into the component, or NcInputField does not support more than one trailing button, so while this is possible we would need to adjust the width programmatically (icons only is always `default-clickable-area` but with text the width is unknown).

szaimen:
> > It should take the whole header on mobile
> 
> This are only styles so yes possible.

Cool :)

> > Can the button to search globally expand on desktop with the wording Search everywhere like it was initially done by Marco?
> 
> The button is currently hacked into the component, or NcInputField does not support more than one trailing button, so while this is possible we would need to adjust the width programmatically (icons only is always `default-clickable-area` but with text the width is unknown).

Okay, lets do this in a follow-up then :)

susnux:
@szaimen I tackled both points now, what do you think?

[Bildschirmaufnahme_20240626_111050.webm](https://github.com/nextcloud/server/assets/1855448/b1bcf219-04ff-403f-9be1-f0eb1f33d162)

szaimen (approved):
Tested and looks great! 

Only small thing I found is that apparently 
- if you do not have the notifiction app enabled, it seems to break the width on mobile 
- (and also if you type someting in, the button to search everywhere seems to vanish):
image

susnux:
> (and also if you type someting in, the button to search everywhere seems to vanish):

This is a library bug, the button stays there but somehow the slot vanishes from beeing rendered

szaimen:
> > (and also if you type someting in, the button to search everywhere seems to vanish):
> 
> This is a library bug, the button stays there but somehow the slot vanishes from beeing rendered

All right, sounds like a follow-up then :)

susnux:
> if you do not have the notifiction app enabled, it seems to break the width on mobile

Fixed, was missing a `inset-inline: 0` for mobile view so that is properly aligned.

szaimen (approved):
Tested and works. Ready to go 🚀🎉☺️

susnux (approved):

jancborchardt (approved):
Looks generally fine with me, of course some polishing opportunities:
- [ ] Better icon for "Search everywhere" [Travel Explore](https://fonts.google.com/icons?icon.set=Material+Symbols&icon.size=24&icon.color=%23000000&selected=Material+Symbols+Outlined:travel_explore:FILL@1;wght@400;GRAD@0;opsz@24&icon.query=search)
- [ ] Vertical alignment of clear x button
- [ ] Possibly the "Search everywhere" button should be secondary? Right now it is a bit strange to have bold text in the field which is not your input

But generally yeah, now finally people can filter locally again, and quickly. @AndyScherzinger what do you think?

jancborchardt:
@susnux sounds good with cloud-search!

@szaimen @susnux very important too: Does this then also work in apps management, account management, and other apps with existing in-app search like Talk, Contacts, Mail, etc?

And as @AndyScherzinger brought up: What to do with apps which do _not_ have in-app search? If e.g. in Dashboard or Activity you click the icon and the unified search modal would open directly, that would be different from how it works in the other places.

(The alternative still for this would be to bring a search field like Talk, Contacts and Mail has to Files, apps mgmt and account mgmt, but then we still have 2 searches which is a bit much.)

core/src/views/UnifiedSearch.vue
```diff
@@ -4,47 +4,164 @@
...
+		debouncedQueryUpdate() {
+			return debounce(this.emitUpdatedQuery, 250)
+		},
+
+		/**
+		 * Current page (app) supports local in-app search
+		 */
+		supportsLocalSearch() {
+			// TODO: Make this an API
+			const providerPaths = ['/settings/users', '/apps/files', '/apps/deck']
```
szaimen, line 83: I guess this needs to be adjusted here...
`settings/apps` seems to be indeed missing and other apps as well if they should be supported...

marcoambrosini:
Could the input be always shown and centered if there's enough space? 
By enough space I mean showing a few apps and then clipping the list

marcoambrosini:
> very important too: Does this then also work in apps management, account management, and other apps with existing in-app search like Talk, Contacts, Mail, etc?

We would need to remove local search from those apps if we go with this imo

core/src/views/UnifiedSearch.vue
susnux: Probably yes, this is the currently implemented list of apps that support "filter in current view"

susnux:
> sounds good with cloud-search!

Bildschirmfoto_20240626_204122

jancborchardt (commented):
@susnux @szaimen just before you put more time into this – we talked about this again among @AndyScherzinger @karlitschek and me, and in addition to the considerations [in the previous comment](https://github.com/nextcloud/server/pull/41609#issuecomment-2191683800) we think that in the long term it would be better to have 2 searches:
- One present search **field** inside the app, like it is in Talk, Contacts, Mail, Collectives, etc.
- One search **icon** on the top right for the Unified search

And to do that, we would "simply" need to add in-app searches in:
- Files: As per mockups in https://github.com/nextcloud/server/issues/45437
- Apps management: Top of the left navigation
- Account management: Top of the left navigation

This would also resolve the confusion with the bit strange interaction pattern of having an input field with a button that does something else, and with internal search not existing in specific apps and due to that the flow being different.

This is very comparable to how for example macOS or GNOME does it with every app having in-app search, and there being a global search (like Spotlight) in case you don’t find it.

@szaimen @susnux sorry for the back and forth – it just turns out it is way more complex than thought. :\

szaimen:
Hi @susnux this is not working on daily.ltd2.nextcloud.com anymore. Was it removed?

PR #41631, opened 2023-11-21
Remove empty user search provider

Altahrim:
User search provider is currently unimplemented and we'll stick to current behaviour of filter within front


artonge:
You sure? This was added by @sorbaugh a month ago: https://github.com/nextcloud/server/pull/40413

PR #41632, opened 2023-11-21
Fix adding x-requested-with header for relative URLs in some cases

julien-nc:
When NC is at the root path of the web server, `getRootUrl()` returns an empty string and `isNextcloudUrl()` always returned true for relative URLs.

The condition to know if a URL points to NC is now:
* the URL starts with the absolute NC base URL (this implicitly checks if the URL is absolute)
* OR the URL is relative and start with NC's root URL (the sub path)

This broke the OpenStreetMap integration in instances that are accessible at the root of the webserver.


emoral435 (approved):
Looks good to me!

juliusknorr (approved):

julien-nc:
@juliushaertl @emoral435 @susnux Cypress test failure seems unrelated (theming app). Can someone force merge this? :pray:

PR #41635, opened 2023-11-21
fix(theming): Adjust config listener to validate `apporder` config also for closure navigation

susnux:
## Summary

The `app` key is optional in the settings, so we only need to check it if set.

## Checklist


AndyScherzinger (approved):

come-nc (approved):

susnux:
Drone fails due to the sharing tests (unrelated)

Issue #41467, opened 2023-11-14
[Bug]: shared file link opens folder, not file


gerbrent:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

opening a shared internal link that is pointing to a file, i.e. https://mynextcloud.url/f/1234567 in a browser opens the folder, not the file itself.

This is seen as a regression.

### Steps to reproduce

1. copy/share/receive an internal link to a file eg. https://mynextcloud.url/f/1234567
2. click on / browse to said link in browser
3. the folder that contains the file is indeed opened, but the file itself is not opened in eg. Files editor for .md

### Expected behavior

expecting the file that was linked to be opened in Files editor for .md, etc.

### Installation method

None

### Nextcloud Server version

28.0.0 beta 3

### Operating system

Other

### PHP engine version

None

### Web server

None

### Database engine version

None

### Is this bug present after an update or on a fresh install?

Updated from a MINOR version (ex. 22.1 to 22.2)

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Additional info

Nextcloud Server version 28.0.0 beta 2
Firefox & Chromium 
openSUSE Tumbleweed

AndyScherzinger:
https://github.com/nextcloud/server/pull/27102

PR #41636, opened 2023-11-21
[stable27] fix: Avoid triggering a defered sidebar open if openFile is already handling that

juliusknorr:
In the office integration we don't want to show the sidebar after opening a file through a direct link as it takes away too much space. We just call a sidebar close within richdocuments but the deferred call to show the details view on the scrollTo action will trigger later. In this case we can actually avoid trigger the details view that way as it is already called when the file action is triggered a few lines below https://github.com/nextcloud/server/blob/18034204e6dbb8bbecafbf2bb750832c004b21a0/apps/files/js/filelist.js#L440

This fix is targeting 27 only as master might have changed behaviour due to the files2vue migration (needs https://github.com/nextcloud/server/issues/41467 to be re-evaluated)

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/files/js/filelist.js
+++ b/apps/files/js/filelist.js
@@ -410,7 +410,7 @@
 
 			if (options.scrollTo) {
 				this.$fileList.one('updated', function() {
-					self.scrollTo(options.scrollTo);
+					self.scrollTo(options.scrollTo, !options.openFile);
 				});
 			}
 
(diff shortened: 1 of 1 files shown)


Pytal (approved):

Darshan-upadhyay1110:
Tested https://github.com/nextcloud/server/pull/41636.patch on stable27. 

Cases i have tested :


  -  Opening internal shared document from Recommended files => OK ( Before patch => NOK )
  -  Open Documents created by me in my own Documents folder => OK
  -  Open internal shared document from files view => OK
  -  Open internal shared document via url => OK ( Before patch => NOK )
  - Open public shared document via url => OK


Note: Sidebar is now closed for all the file open cases mentioned above.

NC version : Nextcloud Hub 6 (27.1.3)
Richdocument : '8.2.3'

emoral435 (approved):

PR #41642, opened 2023-11-21
Correct spelling of Brute Force and other fixes

Valdnet:
1. Correcting the spelling of "Bruteforce" to "Brute Force".
2. Adding quotation marks.
3. Fixed a typo.

### DIFF
--- a/apps/settings/lib/SetupChecks/BruteForceThrottler.php
+++ b/apps/settings/lib/SetupChecks/BruteForceThrottler.php
@@ -47,7 +47,7 @@ class BruteForceThrottler implements ISetupCheck {
 	}
 
 	public function getName(): string {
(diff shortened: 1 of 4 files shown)


come-nc (approved):
Not an english native speaker so I do not know about the change itself but the PR won’t break anything :+1:

rakekniven (changes requested):
"brute-force" is correct from my pov. Can you double check?

Valdnet:
"brute-force" is an alternative spelling of this word.

Valdnet:
I propose "Brute force" 😁.

rakekniven (approved):

Valdnet:
So we change it to "Brute-force" 😉.

AndyScherzinger (approved):

Issue #41484, opened 2023-11-15
[Bug]: Cannot search in users management page with global search enabled


blizzz:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

With global search enabled and on the users page, the list of users cannot be filtered anymore. When clicking "people" on the global search, and looking for a users, they are not found either. Also the duplicated input box is confusing.

[global search vs user mgmt.webm](https://github.com/nextcloud/server/assets/2184312/7eca8aa1-d53a-4b59-a712-d6e024f6f247)

### Steps to reproduce

1. Have global search enabled
2. go to users page
3. try to search for a user

### Expected behavior

A search for the entered string is performed and only matching users are shown. As it happens with old search.

### Installation method

None

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

blizzz:
cc @Fenn-CS @Altahrim

Altahrim:
It seems the user search provider is a bit incomplete: https://github.com/nextcloud/server/blob/master/apps/settings/lib/Search/UserSearch.php#L72-L75

blizzz:
> It seems the user search provider is a bit incomplete: https://github.com/nextcloud/server/blob/master/apps/settings/lib/Search/UserSearch.php#L72-L75

Related comment: https://github.com/nextcloud/server/issues/41381#issuecomment-1812127225

blizzz:
[...]

PR #41646, opened 2023-11-21
feat(search): hide search provider from list

Altahrim:
- Some search providers can be hidden if they send `null` in `getOrder`
 - Hide user search provider if not on user list (#41484)

### DIFF
--- a/apps/settings/lib/Search/UserSearch.php
+++ b/apps/settings/lib/Search/UserSearch.php
@@ -6,7 +6,7 @@ declare(strict_types=1);
  * @copyright Copyright (c) 2023 Stephan Orbaugh <stephan.orbaugh@nextcloud.com>
  *
  * @author Stephan Orbaugh <stephan.orbaugh@nextcloud.com>
- *
+ * @author Benjamin Gaussorgues <benjamin.gaussorgues@nextcloud.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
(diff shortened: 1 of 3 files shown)


artonge (approved):

apps/settings/lib/Search/UserSearch.php
```diff
@@ -33,45 +33,26 @@
...
 		return $this->l->t('Users');
 	}
 
-	/**
-	 * @inheritDoc
-	 */
-	public function getOrder(string $route, array $routeParameters): int {
-		return 300;
+	public function getOrder(string $route, array $routeParameters): ?int {
+		return $route === 'settings.Users.usersList'
```
Fenn-CS, line 50: Please, why are we doing this in this getter?  Is is because the `route` is passed? I mean for example why was it not done in `getName`? 

I am thinking it could be handled elsewhere.... But I am not sure what the bottleneck could be, just throwing it here.

nfebe (approved):
Apart from one in-line worry, this works.

apps/settings/lib/Search/UserSearch.php
nickvergessen: Yeah definitely need the route e.g. for the collective search and talks "current conversation chat messages" etc.
So would also do it in this method.
Also has the least impact api wise

lib/public/Search/IProvider.php
```diff
@@ -68,15 +68,16 @@ public function getName(): string;
 	/**
 	 * Get the search provider order
 	 * The lower the int, the higher it will be sorted (0 will be before 10)
+	 * If null, the search provider will be hidden
```
nickvergessen, line 71:
```suggestion
	 * If null, the search provider will be hidden in the UI and the API not called
```

Otherwise could be read as I'm still being called/invoked

lib/public/Search/IProvider.php
```diff
@@ -68,15 +68,16 @@ public function getName(): string;
...
 	 *
 	 * @param string $route the route the user is currently at, e.g. files.view.index
 	 * @param array $routeParameters the parameters of the route the user is currently at, e.g. [fileId = 982, dir = "/"]
 	 *
-	 * @return int
+	 * @return int|null
 	 *
 	 * @since 20.0.0
```
nickvergessen, line 78:
```suggestion
	 * @since 20.0.0
 	 * @since 28.0.0 Can return null
```

PR #41649, opened 2023-11-21
fix: lower threshold for system address book sync

kesselb:
## Summary

- Switch back to countUsers to have the actual value. countSeenUsers is bad if Nextcloud is connected to a larger directory, but only a part is using Nextcloud and therefore the seen count is much lower because the sync is done for all users.
- Lower the threshold to 100

## TODO

- [ ] CI

## Checklist

### DIFF
--- a/apps/dav/lib/Migration/Version1027Date20230504122946.php
+++ b/apps/dav/lib/Migration/Version1027Date20230504122946.php
@@ -49,7 +49,7 @@ class Version1027Date20230504122946 extends SimpleMigrationStep {
 	 * @param array $options
 	 */
 	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
-		if($this->userManager->countSeenUsers() > 1000) {
+		if($this->userManager->countSeenUsers() > 100 || array_sum($this->userManager->countUsers()) > 100) {
 			$this->config->setAppValue('dav', 'needs_system_address_book_sync', 'yes');
 			$output->info('Could not sync system address books during update - too many user records have been found. Please call occ dav:sync-system-addressbook manually.');
 			return;


miaulalala (approved):

apps/dav/lib/Migration/Version1027Date20230504122946.php
```diff
@@ -49,7 +49,7 @@ public function __construct(private SyncService $syncService,
 	 * @param array $options
 	 */
 	public function postSchemaChange(IOutput $output, Closure $schemaClosure, array $options): void {
-		if($this->userManager->countSeenUsers() > 1000) {
+		if(array_sum($this->userManager->countUsers()) > 100) {
```
ChristophWurst, line 52: if countSeenUsers() is faster you can do a combined condition. something like `if (countSeenUsers() > 100 || countUsers() > 100)` 

this means countUsers is called only for user backends that don't support counting, or where seen users is lower 100 while actual users is above

@miaulalala anything I'm missing? this is what we defined a few weeks ago

kesselb: Yep, that's cool :+1: 

**countSeenUsers**
 `select count(userid) from oc_preferences where appid = 'login' and configkey = 'lastLogin';`

**countUsers**

Database: https://github.com/nextcloud/server/blob/1cd4d4e018687513fa6e14d3d1af5a88d18a4760/lib/private/User/Database.php#L468-L476
SAML: https://github.com/nextcloud/user_saml/blob/b8d0ed551752d2e1ef23a8fac26424b1d067b91c/lib/UserBackend.php#L674-L677
LDAP: Is usually an external request for the ldap directory.

ChristophWurst (approved):

AndyScherzinger (approved):

PR #41650, opened 2023-11-21
feat(LDAP): implement IIsAdmin interface

### LINKED ISSUE #6428: Give admin rights to a specific LDAP group
My users are stored in an LDAP directory and I want to give admin permissions to some of them. I could manually put them in the default admin group of nextcloud but it would be more efficient to be able to select a group from LDAP as admin group. Could that be implemented?

blizzz:
* Resolves: #6428 

## Summary

Promotes an LDAP group (per LDAP configuration) to an admin group.

A group can either be promoted via a dedicates occ call whres the group parameter can be a nextcloud group ID or a group name that will be search against – an exact match is required in that case.:

```
$ php occ ldap:promote-group  --help 2>/dev/null
Description:
  declares the specified group as admin group (only one is possible per LDAP configuration)

Usage:
  ldap:promote-group [options] [--] <group>

Arguments:
  group                 the group ID in Nextcloud or a group name

Options:
  -y, --yes             do not ask for confirmation
…

# Example
$ php occ ldap:promote-group  "Nextcloud Admins"
Promote Nextcloud Admins to the admin group (y|N)? y
Group Nextcloud Admins was promoted

$ php occ ldap:promote-group  "Paramount Court"
Promote Nextcloud Admins to the admin group and demote Nextcloud Admins (Group ID: nextcloud_admins) (y|N)? y
Group Paramount Court was promoted

$ php occ ldap:promote-group  "Paramount Court"
The specified group is already promoted
```

It is also possible to set the admin group mapping using `occ ldap:set-config $configId ldapAdminGroup $groupId` but as the Nextcloud group ID might not be known (yet) it is especially recommnded for automatized setups to use this command, that would also pull in the group and determine the group ID.

In order to demote or reset a promotion, an empty string should be set against to the targeted config's ldapAdminGroup:

```
# Reset an admin group mapping via set-config
occ ldap:set-config $configId ldapAdminGroup ""
# Example
occ ldap:set-config s01 ldapAdminGroup ""
```

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Command/PromoteGroup.php
+++ b/apps/user_ldap/lib/Command/PromoteGroup.php
@@ -1,0 +1,113 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2023 Arthur Schiwon <blizzz@arthur-schiwon.de>
+ *
+ * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCA\User_LDAP\Command;
+
+use OCA\User_LDAP\Group_Proxy;
+use OCP\IGroup;
+use OCP\IGroupManager;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Helper\QuestionHelper;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\Question;
+
+class PromoteGroup extends Command {
+
+	public function __construct(
+		private IGroupManager $groupManager,
+		private Group_Proxy $backend
+	) {
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('ldap:promote-group')
+			->setDescription('declares the specified group as admin group (only one is possible per LDAP configuration)')
+			->addArgument(
+				'group',
+				InputArgument::REQUIRED,
+				'the group ID in Nextcloud or a group name'
+			)
+			->addOption(
+				'yes',
+				'y',
+				InputOption::VALUE_NONE,
+				'do not ask for confirmation'
+			);
+	}
+
+	protected function promoteGroup(IGroup $group, InputInterface $input, OutputInterface $output): void {
+		if ($input->getOption('yes') === false) {
+			/** @var QuestionHelper $helper */
+			$helper = $this->getHelper('question');
+
+			$idLabel = '';
+			if ($group->getGID() !== $group->getDisplayName()) {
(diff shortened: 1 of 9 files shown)


blizzz:
Need to check the integration tests, the positive one is failing :thinking: 

User should be in the target group
https://github.com/nextcloud/docker-ci/blob/master/openldap/LDIFs/ordinary_small.ldif#L160-L163

apps/user_ldap/lib/Command/PromoteGroup.php
```diff
@@ -0,0 +1,118 @@
...
+	protected function configure(): void {
+		$this
+			->setName('ldap:promote-group')
+			->setDescription('declares the specified group as admin group (only one is possible per LDAP configuration)')
+			->addArgument(
+				'group',
+				InputArgument::REQUIRED,
+				'the group ID in Nextcloud or a group name'
+			)
+			->addOption(
```
juliushaertl, line 61: Optional: We could have a option to remove the group again or set an empty config.

blizzz: That is what slipped my mind yesterday, yes.

blizzz: The tricky part here is that we do not specify the config prefix, but detect it depending on the group ownership. So, only resetting any group would need the config prefix. 

Resetting by a group name or id could fail if the group is deleted, but we may try to guess by comparing stored config – not full proof. I tend to go with documenting how to do with with set-config instead (does not require another command).

Apropos group deletion, this is something we could take into account as well and remove the promotion when a group disappears – although this is a bit atypical from what we usually do (like keeping users in place), but actually safer than keeping the entry. Questionable upon unassigned the group id (clearing the mappings), which should not be done in prod though (would keep it in that case).

blizzz:
> Apropos group deletion, this is something we could take into account as well and remove the promotion when a group disappears – although this is a bit atypical from what we usually do (like keeping users in place), but actually safer than keeping the entry. Questionable upon unassigned the group id (clearing the mappings), which should not be done in prod though (would keep it in that case).

An addition for later for we do not delete groups yet.

blizzz:
> Resetting by a group name or id could fail if the group is deleted, but we may try to guess by comparing stored config – not full proof. I tend to go with documenting how to do with with set-config instead (does not require another command).

Updated the description, that will go into documentation eventually.

apps/user_ldap/lib/Command/PromoteGroup.php
juliushaertl: Ah ok, I misread the changeset then in regards how this works with multiple ldap configurations. Thanks for explaining, makes sense then.

juliusknorr (approved):

apps/user_ldap/lib/Command/PromoteGroup.php
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('ldap:promote-group')
+			->setDescription('declares the specified group as admin group (only one is possible per LDAP configuration)')
+			->addArgument(
+				'group',
+				InputArgument::REQUIRED,
+				'the group ID in Nextcloud or a group name'
```
come-nc, line 54: LDAP dn does not work?

apps/user_ldap/lib/Command/PromoteGroup.php
```diff
@@ -0,0 +1,113 @@
...
+		if ($input->getOption('yes') === false) {
+			/** @var QuestionHelper $helper */
+			$helper = $this->getHelper('question');
+
+			$idLabel = '';
+			if ($group->getGID() !== $group->getDisplayName()) {
+				$idLabel = sprintf(' (Group ID: %s)', $group->getGID());
+			}
+
+			$q = new Question(sprintf('Promote %s%s to the admin group (y|N)? ', $group->getDisplayName(), $idLabel));
```
come-nc, line 74: The question should also inform if this is unpromoting another group which was previously set

apps/user_ldap/lib/Command/PromoteGroup.php
blizzz: not unless you specific it in the search attributes

apps/user_ldap/lib/Command/PromoteGroup.php
blizzz: done, added a commit with this handling

come-nc (approved):

blizzz:
addressed the php-cs test complaint.

PR #41652, opened 2023-11-21
Focus global search input on open

### LINKED ISSUE #41651: Left over polishing for global search from
Left over polishing for global search from https://github.com/nextcloud/server/pull/41507

## High
- [x] Search field is not focused automatically when the modal is opened

## Medium
- [ ] Contacts filter shouldn’t be available if contacts app is not available 
- [ ] No space for icons in modified

## Low
- [ ] "No matching results" emptycontent icon should be a search icon instead of the current one
- [ ] Text in People" menu regular not bold
- [ ] Close icon material design icon "close" instead of "close-thick"

nfebe:
For better accesibility, the search input in the global search modal should be focused when the search modal is opened.

Resolves: https://github.com/nextcloud/server/issues/41651


artonge (approved):

marcoambrosini (changes requested):
Tested and doesn't work for me. 
Wouldn't it be better to watch `isVisible` and call `focusInput` whenever the value flips to `true`?

Fenn-CS:
@marcoambrosini I don't think that's the issue, I linked this issue in the chat earlier where you sent a similar implementation.

https://github.com/nextcloud/server/pull/41609/files#diff-bf6af3705e31866549e767f82756d2c3a26b3af00931f66e127712a7361b3978R124

The actual issue is that, the search reference is undefined.

Fenn-CS:
> Tested and doesn't work for me.
> Wouldn't it be better to watch isVisible and call focusInput whenever the value flips to true?

Resolved @marcoambrosini

marcoambrosini (approved):
Works now :)

AndyScherzinger (approved):

blizzz:
What about the failing cypress tests?

AndyScherzinger:
> What about the failing cypress tests?

* theming/navigation-bar-settings fail everywhere
* settings/apps only here (I'll restart it)

PR #41656, opened 2023-11-22
Show shared label only if the current user have shared the file.

### LINKED ISSUE #41620: "Shared" label takes a lot of space in Files list
There are some drawbacks to the recent change for the "Shared" label that came with the Vue rewrite:
- The "Shared" label takes a lot of space in the files list and makes it look very busy
- Now a lot of entries have the "Shared" label, making it actually more difficult to distinguish your "own" files and files of others at a glance
- Somehow the avatars got smaller

Now we shouldn’t remove the "Shared" label altogether since it is an important indicator of state and it has also been there in prior versions.

Here are 2 other options we could go for:
1. **Medium with avatars + icons** For internal shares coming from others, we use the proposal by @Clementine46 from https://github.com/nextcloud/server/issues/38156
2. **Remove label for shares from others:** This is basically back to how it was before, but keeping the layout and color improvement of the "Shared" label.

Current | Medium with avatars + icons | Remove label for shares from others
-|-|-
Sharing current|sharing medium|Sharing plain

cc @nextcloud/designers and @karlitschek since you reported the original issue.

marcoambrosini:
The presence of the text "Sharing label" in all shared files caused the view to be too loaded, so it's been decided to show it only when the current user is the owner of the share.

* Resolves: #41620 

## Summary
Screenshot from 2023-11-22 15-42-42

## TODO

- [ ] ...

## Checklist


jancborchardt (approved):
Looks good to me, as this is just the previous state with the slight color improvement and switch of icon/text, without any further modification. @karlitschek for review as well since you reported it.

szaimen (commented):
This will basically re-introduce https://github.com/nextcloud/server/issues/38156. So not sure if I agree with this change.

jancborchardt:
@szaimen see @marcoambrosini’s comment at https://github.com/nextcloud/server/issues/41620#issuecomment-1822194264 – it’s either this or the full-blown "Shared" label everywhere which @karlitschek also reported as being too present.

So for now I think it’s better if we go back to a slightly improved version of the original state.

AndyScherzinger (approved):

szaimen:
> @szaimen see @marcoambrosini’s comment at [#41620 (comment)](https://github.com/nextcloud/server/issues/41620#issuecomment-1822194264) – it’s either this or the full-blown "Shared" label everywhere which @karlitschek also reported as being too present.

yes, I saw that but still do not like this change. Anyway, feel free to approve the PR yourself. I am not blocking it. :)

marcoambrosini:
For what is worth, I also think it's better to always show the shared label at all times. I think it's the 2nd most important thing after the file name. Let's figure out a better strategy to make things clear soon!

PR #41659, opened 2023-11-22
feat(UserMenu): add target property to usermenu component

### LINKED ISSUE #41050: Please add 'target = "_blank"' to the link 'List of invalid files' in Settings > Admin > Overview
Due to some causes in updating Nextcloud, some files do not passed the integrity check often and the link to view the list of those files is displayed on the Overviewe - Admin Settings. Clicking the link will bring you to the simple page displaying the list. The page is without the normal theme/style of the other Nextcloud pages. I always feel I do something wrong or there is something wrong at the Nextcloud server so it brings me out of the main flow of working to the point of totally server failure. Also I always want to see the list along with working on the Nextcloud at the same time.

I would like to suggest adding target = "_blank" to the link to keep us on the main flow of work together with looking the list and to make us feel safe with this small error too.

I find that the place to add it is at the line 320 in core/js/setupchecks.js.

Thanks

### How to use GitHub

fitrahmunir:
* Resolves: #41050

## Summary
Add target property to UserMenu and UserMenuEntry components and add target "_blank" to Administration settings

## TODO

- [x] Add target property to UserMenu and UserMenuEntry components
- [x] Add target "_blank" to admin_settings

## Checklist

### DIFF
--- a/core/src/components/UserMenu/UserMenuEntry.vue
+++ b/core/src/components/UserMenu/UserMenuEntry.vue
@@ -25,6 +25,7 @@
 		class="menu-entry">
 		<a v-if="href"
 			:href="href"
+			:target="target"
 			:class="{ active }"
 			@click.exact="handleClick">
 			<NcLoadingIcon v-if="loading"
@@ -70,6 +71,10 @@ export default {
 		active: {
 			type: Boolean,
 			required: true,
+		},
+		target: {
+			type: String,
+			required: false,
 		},
 		icon: {
 			type: String,
--- a/core/src/views/UserMenu.vue
+++ b/core/src/views/UserMenu.vue
@@ -44,6 +44,7 @@
 				:key="entry.id"
 				:name="entry.name"
 				:href="entry.href"
+				:target="entry.target"
 				:active="entry.active"
 				:icon="entry.icon" />
 		</ul>
--- a/lib/private/NavigationManager.php
+++ b/lib/private/NavigationManager.php
@@ -294,6 +294,7 @@ class NavigationManager implements INavigationManager {
 					'id' => 'admin_settings',
 					'order' => 4,
 					'href' => $this->urlGenerator->linkToRoute('settings.AdminSettings.index', ['section' => 'overview']),
+					'target' => '_blank',
 					'name' => $l->t('Administration settings'),
 					'icon' => $this->urlGenerator->imagePath('settings', 'admin.svg'),
 				]);
--- a/tests/lib/NavigationManagerTest.php
+++ b/tests/lib/NavigationManagerTest.php
@@ -343,7 +343,8 @@ class NavigationManagerTest extends TestCase {
 				'active' => false,
 				'type' => 'settings',
 				'classes' => '',
-				'unread' => 0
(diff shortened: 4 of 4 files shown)


nickvergessen:
Failing CI is related:
image

nickvergessen:
I don't think the admin settings should open in a new tab all the time?
Neither the user menu links...

cc @nextcloud/designers

marcoambrosini:
This is an old one.. I agree with the proposed change. I think that admin settings and personal settings, and all apps for that matter,  should open in another tab. This solution:

- **Reduces friction**: once I'm done with the admin settings work, I don't have to remember where I was and what I was doing. I just close the tab and the first thing I see is my work exactly where I left it. 
 Moreover, had open settings in the same tab, the state of the app that I navigated away from is not saved, and will likely never be saved:
It has been suggested in the past that we  could "remember" the state of each app, but this is hard to implement and maybe not even desirable in most cases. In any event, we'd be reinventing something that's already there at the browser level: **multitasking**. So in my opinion it's not a wise allocation of resources.

- **Cuts on lading times**: I don't have to load the app that I was using again.


**So by insisting on loading different things in the same tab in pursuit of the big single page application feel, we're effectively discouraging users from using the multitasking feature that comes with a web browser.**

@nextcloud/designers please go and take a look at other (multi-app) projects, the usual suspects. All of them open apps and global settings in new tabs, despite the fact that their loading times are in general faster than ours. (Testing their loading times vs. my local docker loading times).

szaimen:
Hm... I think I don't like this idea. New tabs usually should only be opened for links to external services or if it is really necessary IIRC...

szaimen:
> And our friends at G & M, that have all the UX research resources that they want, reached that conclusion too.

Fine by me then

fitrahmunir:
@nickvergessen please review this PR

emoral435 (commented):
I agree with the designers, I think that not having the user exit out of the page is always convenient. IMO - I would rather use more RAM for the tab than have to navigate back to my page xD But I like the code changes as well, just a minor request

come-nc (commented):
👎 I do not think this is a good idea. It forces the user to open those pages in new tabs while it’s usually not what I want.
See https://sensiolabs.com/blog/2025/the-tab-trap-why-forcing-new-tabs-is-bad-ux for why this is a bad practice.

ShGKme:
I agree with szaimen and come-nc. 

If a user wants to open admin settings in a new tab, the user can user CTRL+Click, middle mouse button click, or the context menu while the default is opening in the current tab. We do not prevent the choice here.

If we change the default, keeping it open in the same tab is not so straightforward, and we are giving no choice to the user.

Especially if a user initially opened the instance to navigate to the admin settings.

ShGKme:
Maybe it was changed in the meanwhile, but the linked issue seems about a different problem.

susnux:
@fitrahmunir thank you very much for your contribution!
As outlined by the others opening every user menu entry in a new tab is not good UX.
Moreover the bug report was about the setup checks for the integrity results where such behavior probably makes sense because it redirects to a plain text file and not a webpage.

PR #41662, opened 2023-11-22
chore(settings): Cleanup IManager and Manager type annotations

provokateurin:
## Summary

Use correct types everywhere in IManager and Manager of Settings.

## Checklist

### DIFF
--- a/lib/private/Settings/Manager.php
+++ b/lib/private/Settings/Manager.php
@@ -12,6 +12,7 @@
  * @author Roeland Jago Douma <roeland@famdouma.nl>
  * @author sualko <klaus@jsxc.org>
  * @author Carl Schwan <carl@carlschwan.eu>
+ * @author Kate Döen <kate.doeen@nextcloud.com>
  *
  * @license GNU AGPL version 3 or any later version
  *
@@ -90,17 +91,14 @@ class Manager implements IManager {
 		$this->subAdmin = $subAdmin;
 	}
 
-	/** @var array */
+	/** @var array<self::SETTINGS_*, list<class-string<IIconSection>>> */
 	protected $sectionClasses = [];
 
-	/** @var array */
+	/** @var array<self::SETTINGS_*, array<string, IIconSection>> */
 	protected $sections = [];
 
 	/**
-	 * @param string $type 'admin' or 'personal'
-	 * @param string $section Class must implement OCP\Settings\IIconSection
-	 *
-	 * @return void
+	 * @inheritdoc
 	 */
 	public function registerSection(string $type, string $section) {
 		if (!isset($this->sectionClasses[$type])) {
@@ -111,7 +109,7 @@ class Manager implements IManager {
 	}
 
 	/**
-	 * @param string $type 'admin' or 'personal'
+	 * @psalm-param self::SETTINGS_* $type
 	 *
 	 * @return IIconSection[]
 	 */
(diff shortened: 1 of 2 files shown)


lib/public/Settings/IManager.php
```diff
@@ -103,7 +104,7 @@ public function getAllowedAdminSettings(string $section, IUser $user): array;
 	/**
 	 * Returns a list of admin settings that the given user can use.
 	 *
-	 * @return array<int, list<ISettings>> The array of admin settings there admin delegation is allowed.
+	 * @return list<ISettings> The array of admin settings there admin delegation is allowed.
```
provokateurin, line 107: This interface was wrong from the beginning. The implementation never returned the nested array

ChristophWurst (approved):

lib/public/Settings/IManager.php
nickvergessen: 🙈

nickvergessen (approved):

lib/public/Settings/IManager.php
```diff
@@ -53,31 +54,31 @@ interface IManager {
 	public const KEY_PERSONAL_SECTION = 'personal-section';
 
 	/**
-	 * @param string $type 'admin-section' or 'personal-section'
-	 * @param string $section Class must implement OCP\Settings\ISection
+	 * @psalm-param 'admin'|'personal' $type
```
ChristophWurst, line 57: is this a breaking change or did we never allow admin-section and personal-section?

provokateurin: The method is never called with any other values and the code can only handle those two types.

lib/private/Settings/Manager.php
```diff
@@ -111,7 +109,7 @@ public function registerSection(string $type, string $section) {
 	}
 
 	/**
-	 * @param string $type 'admin' or 'personal'
+	 * @psalm-param 'admin'|'personal' $type
```
nickvergessen, line 112: Does that work in psalm:
```suggestion
	 * @psalm-param self::KEY_*_SECTION $type
```

lib/public/Settings/IManager.php
nickvergessen: https://github.com/nextcloud/server/blob/046d5451b15d46581cde846676e0b6d631fedbb7/lib/public/Settings/IManager.php#L35-L53 should deprecate the wrong ones then?

nickvergessen: But yeah was done wrongly here:
https://github.com/nextcloud/server/blob/927e7fcf67e4e1216e1551160919cd018f12b963/lib/private/App/AppManager.php#L459-L478

provokateurin: Ok what should I do about this?

lib/private/Settings/Manager.php
provokateurin: No, seems like the asterisk is only allowed at the end

provokateurin:
Ugh looking at the implementations more closely I found some more wrong types in the interface :/

lib/public/Settings/IManager.php
nickvergessen: I would say deprecate the `_SECTION` constants and in the deprecation mention that they were never used and instead the `_SETTINGS` sibling should be used.

provokateurin: We could also add new constants like `SETTINGS_ADMIN` and `SETTINGS_PERSONAL` and then use `self::SETTINGS_*` throughout the interface and implementation (like you suggested below).

lib/public/Settings/IManager.php
```diff
@@ -34,50 +35,64 @@
 interface IManager {
 	/**
 	 * @since 9.1.0
+	 * @depreacted 29.0.0 Use SETTINGS_ADMIN instead
```
nickvergessen, line 38:
```suggestion
	 * @depreacted 29.0.0 Use {@see self::SETTINGS_ADMIN} instead
```
Makes it clickable in IDEs

nickvergessen (approved):

PR #41669, opened 2023-11-22
Add app-based filtering to global search

nfebe:
Emit search queries from the global search modal that would trigger list filtering in various apps that support it.


AndyScherzinger (approved):

Altahrim (approved):

juliushaertl:
From a UX perspective i feel this is the wrong approach to filter the app from within the modal, so I hope https://github.com/nextcloud/server/pull/41609 is also considered soonish.

Fenn-CS:
> From a UX perspective i feel this is the wrong approach to filter the app from within the modal, so I hope https://github.com/nextcloud/server/pull/41609 is also considered soonish.

Well, I have already over-stated this point, especially after implementing and interacting with this. The priority communicated is that we first restore the "original" behavior which is what this PR is about.

So I hope, we undo this asap and improve on what Marco has proposed already. I think is way more decent and at least we have the toggle for users to just go back to the old search.

core/src/views/GlobalSearchModal.vue
```diff
@@ -227,6 +238,9 @@ export default {
 				this.results = []
 				return
 			}
+			if (this.supportFiltering()) {
+				emit('nextcloud:unified-search.search', { query })
```
Altahrim, line 242: On second thought, why restrict?
If the app doesn't subscribe to the event, it will not do anything right?

Fenn-CS: 1. For the very reason you stated, emitting it does not do anything so, no need to emit it when it's not needed.
2. Code clarity, it makes it clear why, where and when the event needs to be emitted, otherwise it could pass it's just another step in the search procedure

Altahrim: > no need to emit it when it's not needed.

I think we don't know that. Other applications out there could use this mechanism.

juliushaertl: Yep, would agree, especially for community apps or developers working on that I'd vote for always emitting.

Fenn-CS: Would open a follow up PR then!

Fenn-CS:
This partially solves https://github.com/nextcloud/server/issues/41484

PR #41670, opened 2023-11-22
fix(papercut) - rerendered submenu actions for time sensitive actions

### LINKED ISSUE #41336: Move reminder conditional rendering into action `enabled`
https://github.com/nextcloud/server/blob/32c1aebaae07b9b4d7e2dcdfca6b2cbd36a6cfc3/apps/files_reminders/src/actions/setReminderSuggestionActions.ts#L86

Since the presence of the action should be evaluated ideally on click instead of on script execution I think the `FileAction` enabled callback needs to be set to something like `() => Boolean(getDateTime(option.dateTimePreset))` to be more reactive

_Originally posted by @Pytal in https://github.com/nextcloud/server/pull/41010#discussion_r1385269899_

emoral435:
* Resolves #41336

## Summary
Changes submenu loading to re-render the actions in the submenu, as the items could potentially be time dependent. As of now, the items within the submenu are only loaded when dependencies change, or upon initial loading of the page, but by loading it whenever the user clicked the button, they wont have to reload the page, saving them an extra click ✅

## BEFORE
piM9LcEWjP

## AFTER
jFihBodbD6

## TODO

- [x] Since the presence of the action should be evaluated ideally on click instead of on script execution I think the FileAction enabled callback needs to be set to something like () => Boolean(getDateTime(option.dateTimePreset)) to be more reactive
- [x] Made it so that when an action within the dropdown changes the openedSubmenu dependency, reloads only the submenu action and doesn't change other computed dependencies, which would slow down the loading of the page

## Checklist


emoral435:
will compile assets once reviews look good!

apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -108,6 +108,7 @@ import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 
 import CustomElementRender from '../CustomElementRender.vue'
 import logger from '../../logger.js'
+import type { LogLevel } from '@nextcloud/logger'
```
Pytal, line 111:
```suggestion
```

apps/files_reminders/src/actions/setReminderSuggestionActions.ts
```diff
@@ -69,7 +69,7 @@ const nextWeek: ReminderOption = {
  * @param option The option to generate the action for
  * @return The file action or null if the option should not be shown
  */
-const generateFileAction = (option): FileAction|null => {
+const generateFileAction = (option: ReminderOption): FileAction|null => {
 	const dateTime = getDateTime(option.dateTimePreset)
 	if (!dateTime) {
 		return null
```
Pytal, line 75:
```suggestion
```
This check should be removed

Now `getDateTime()` should be called and the returned Date used in the displayName, title, and exec methods instead

emoral435: I see what you mean, I am updating any area with the current dateTime variable with a call to getDateTime()

susnux (dismissed):
Why is 3rdparty changed?

Pytal (approved):
Thanks! Just gotta remove 3rdparty changes 🙈

Pytal:
Let's squash these commits into one @emoral435 :)

apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -270,6 +270,32 @@ export default Vue.extend({
 			return action.displayName([this.source], this.currentView)
 		},
 
+		// same as enabledSubmenuActions computed function, but instead as a method to be able to
```
skjnldsv, line 273: Really not sure about this, the implementation looks very unusual.
Why would only sub-menus allowed to auto refresh?

I think this is something that should be fixed in a different way maybe?
Have you tried returning a reference for the displayName and have that reference change? Is the menu not already updating as is? I think ""in theory"" having dynamic string updated on render should already work.
The issue is that we're returning a string from  `displayName` or `title` instead of a `reference` to a string, so vue doesn't see any change :thinking:

emoral435: Yep, as of right now, I am trying to think of a clever way to have the submenu actions refresh so that time dependent actions are loaded depending on the time when the user clicked, and not when the page got loaded. 

I am trying to see ways around the caching such that we do not recompute too much so that the clients side does not slow down, so I will give what you said and other methods as well a shot

skjnldsv (commented):
Please also remove the last `.filter(Boolean) as FileAction[]` then :)

apps/files/src/components/FileEntry/FileEntryActions.vue
Pytal: Seems to work so fine with me ⏲

emoral435: I tried this out on my local dev, its fine with me as well! I like this solution, ties back to the original issue discussing how refreshing the submenu on every click may slow down performance :) submitting my pr changes right now

apps/files_reminders/src/actions/setReminderSuggestionActions.ts
```diff
@@ -110,7 +115,27 @@ const generateFileAction = (option): FileAction|null => {
...
 
+[laterToday, tomorrow, thisWeekend, nextWeek].forEach((option) => {
+	// Generate the initial date string
+	const dateTime = getDateTime(option.dateTimePreset)
+	if (!dateTime) {
+		return
+	}
+	option.dateString = getDateString(dateTime)
+	option.verboseDateString = getVerboseDateString(dateTime)
+	+
```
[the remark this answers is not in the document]
emoral435: Yep, embarrassed that got caught LOL, thank you!

skjnldsv (approved):

PR #41684, opened 2023-11-23
Add mimetype for ReStructuredText

kyteinsky:
## Summary
Mimetype for re structured text. It is aliased to text.  
There is no official mimetype for `.rst` files so `text/plain` can also work here.

## Checklist


kesselb:
If you want to update the mime type for files already existing in oc_filecache an additional change for `lib/private/Repair/RepairMimeTypes.php` is necessary.

Example: https://github.com/nextcloud/server/pull/41395/files#diff-f89544ac76d4f0e2cd70b4979d625b3992a72fa26c6f0b5fe92ec0364b9716e9

kesselb:
> I had assumed repair was a file preview thing. Added it now, thanks for the explanation. Do the changes in `tests/data/integritycheck/mimetypeListModified/core/js/mimetypelist.js` also need to be pushed?

Yep.

There should be a brief inline description in CheckerTest how to update it properly.

kyteinsky:
The CheckerTest did not update the `core/js/mimetypelist.js` file, which was updated with `./occ maintenance:mimetype:update-js` as mentioned in the file.

core/js/mimetypelist.js
```diff
@@ -13,6 +13,7 @@ OC.MimeTypeList={
     "application/font-sfnt": "font",
     "application/font-woff": "font",
     "application/gpx+xml": "location",
+    "application/gzip": "package/x-generic",
```
kesselb, line 16: Would you mind double-checking this alias?
It looks wrong/unrelated to me.

kyteinsky: It was put there by `./occ maintenance:mimetype:update-js`. It still is the case. Maybe someone forgot to run the update command or removed it manually?  
It looks in sync with `mimetypealiases.dist.json` file.

kesselb: Thanks for checking :+1:

kesselb (approved):

version.php
```diff
@@ -30,7 +30,7 @@
 // between betas, final and RCs. This is _not_ the public version number. Reset minor/patch level
 // when updating major/minor version number.
 
-$OC_Version = [29, 0, 0, 5];
+$OC_Version = [29, 0, 0, 6];
```
emoral435, line 33: Is this new version update really required?

emoral435 (commented):
AFAIK (and I very well can be wrong, so correct me if I am), this should not cause any problems with how we handle mimetyping in other parts of our frontend 👍 So if this version upgrade is really needed, then this looks good to me!

version.php
skjnldsv: I think it does, in any case, since we're releasing 29, we can leave the upgrade in i'd say :thinking:

skjnldsv:
Conflicts @kyteinsky

version.php
kyteinsky: yes, it is needed for the repair step to re-label the mimetypes of `.rst` files

skjnldsv (approved):

PR #41668, opened 2023-11-22
test boolean favorite

ArtificialOwl:
follow up from #41655

PR #41705, opened 2023-11-23
confirm content on dav-v2 test

ArtificialOwl:
As requested there https://github.com/nextcloud/server/pull/41668#issuecomment-1823277567 ...

This check if content returned by the SEARCH request to get favorites fit some expectation.
It also fix the image search

### DIFF
--- a/build/integration/features/bootstrap/WebDav.php
+++ b/build/integration/features/bootstrap/WebDav.php
@@ -7,6 +7,7 @@
  * @author Joas Schilling <coding@schilljs.com>
  * @author John Molakvoæ <skjnldsv@protonmail.com>
  * @author Lukas Reschke <lukas@statuscode.ch>
+ * @author Maxence Lange <maxence@artificial-owl.com>
  * @author Morris Jobke <hey@morrisjobke.de>
  * @author Robin Appelman <robin@icewind.nl>
  * @author Roeland Jago Douma <roeland@famdouma.nl>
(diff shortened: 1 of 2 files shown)


ArtificialOwl:
As requested there https://github.com/nextcloud/server/pull/41668#issuecomment-1823277567 ...

This check if content returned by the SEARCH request to get favorites fit some expectation.
It also fix the image search

build/integration/features/bootstrap/WebDav.php
```diff
@@ -1112,4 +1124,32 @@ public function theUploadShouldFailOnObjectStorage() {
...
 		}
 	}
+
+	/**
+	 * @return array
+	 * @throws Exception
+	 */
+	private function convertResponseToDavSingleEntry(): array {
+		$results = $this->convertResponseToDavEntries();
+		if (count($results) <> 1) {
```
juliushaertl, line 1134: To match the exception case, otherwise I'd prefer to use `!==` 
```suggestion
		if (count($results) > 1) {
```

juliusknorr (approved):
Nitpick inline, otherwise looks good 👍

build/integration/features/bootstrap/WebDav.php
ArtificialOwl: the next array_shift will fail :(

ArtificialOwl: I changed the exception message, thanks for the review !

AndyScherzinger (approved):

PR #41718, opened 2023-11-24
Improve personal settings layout

marcoambrosini:
* Resolves: # 

## Summary
One column layout regardless of the width, like in other settings sections. Currently, it's centered, but it can also be moved to the left like in other sections. I just thought it looked better centered.
I think that this is a massive improvement in clarity and usability of this page, and a very low-hanging fruit too.
There's still a lot that can be done, but I think this is a step in the right direction.

**Before**

[Screencast from 2023-11-24 10-45-10.webm](https://github.com/nextcloud/server/assets/26852655/cf17fa95-20d9-476b-8f39-00eb7f732d99)

**After**

[Screencast from 2023-11-24 11-58-26.webm](https://github.com/nextcloud/server/assets/26852655/354af987-9589-4834-9851-d3c4eaf5105c)

## Checklist


szaimen (dismissed):
I like it and looks good from the screenrecording 👍
But didnt test

jancborchardt:
This seems a bit extreme towards the other side, no? The information density is very low and the input fields / selects are uncharacteristically wide.

It would make sense to have a 2-column interface, with avatar, profile and groups in the left column (like now) and all the input field things in the right column. That would be more condensed while at the same time not strangely wide.

(Another issue with the layout or rather the UX is this whole "Profile visibility" section on the bottom which basically duplicates all the fields above. It would be great if we can combine that as well – but should be a separate concern of course.)

jancborchardt (dismissed):
This seems a bit extreme towards the other side, no? The information density is very low and the input fields / selects are uncharacteristically wide.

It would make sense to have a 2-column interface, with avatar, profile and groups in the left column (like now) and all the input field things in the right column. That would be more condensed while at the same time not strangely wide.

(Another issue with the layout or rather the UX is this whole "Profile visibility" section on the bottom which basically duplicates all the fields above. It would be great if we can combine that as well – but should be a separate concern of course.)

marcoambrosini:
@jancborchardt how would the left column behave on scroll? 

If the concern is the width of the input, we can narrow the columns. Most other settings pages are one column too.

marcoambrosini:
I like the idea of having everything in one line, but I think that since this already improves things a lot, we should merge and then iterate. We can arbitrarily reduce the width as we see fit without having to justify the empty space as long as the layout improves the user experience.

Ref: google settings
Group 1

jancborchardt:
> I think that since this already improves things a lot, we should merge and then iterate over the actual design.

@marcoambrosini I disagree on this – as said this goes too far in the other extreme direction. 2 columns would be the proper fix.

> how would the left column behave on scroll?

It can just scroll up and out of view, no need for stickyness there for now.

@nimishavijay nice on the mockup indeed! Great also since this is the start of a new cycle, this could go in as an immediate follow-up for this.

marcoambrosini:
@jan do you mean like this?  I think it's worse than 1 column, especially given that when we scrolled down it's still 1 column, just awkwardly placed on the screen.
 
[Screencast from 2023-12-19 10-59-57.webm](https://github.com/nextcloud/server/assets/26852655/6de2887c-79cc-4050-8383-7d14c42b489e)

marcoambrosini:
--> invert and align to the left

jancborchardt:
@marcoambrosini another point: "Language" and "Locale" can move to the other column as they are personal (more like group memberships and quota usage) and don’t really fit with the others which are profile fields.

marcoambrosini:
@Fenn-CS it needs to scroll because with everything we added it's unlikely to fit in the height of the screen. The profile picture is on the right because otherwise when that column is scrolled away, the other column would be in an odd position: not at the center of the screen and not left aligned either.

szaimen:
Great, looks much better layout-wise :)
Though I wonder is it expected that each section has a different width and aligning now?
See 
localhost_8443_index php_settings_user (3)

jancborchardt (commented):
@marcoambrosini @kra-mo this seems to be a nice thing to update once the forms/settings style improvements are in btw: https://github.com/nextcloud-libraries/nextcloud-vue/issues/7124

jancborchardt:
Closing this as duplicate of or rather being superseded by https://github.com/nextcloud/server/issues/57008 as we meanwhile also have new components. (Just FYI @marcoambrosini)

PR #41729, opened 2023-11-24
Enable new global search by default

nfebe:
The new global search would now be activated by default and users, have the option to revert to the old unified search by setting `unified_search.enabled` to true in the NC `config.php` file.


Altahrim (approved):

artonge (approved):

Fenn-CS:
PHP change, cypress failures unrelated

PR #41755, opened 2023-11-26
Lazy grouping within IAppConfig

ArtificialOwl:
### Added features to the `IAppConfig`:

- **lazy loading** to avoid loading all app configs on each request. It split the configs in 2 groups; the normal (fast) config values that are loaded in all requests and the lazy config that contains values that are loaded only when one config key of this type is needed.
- **typed value** so that app knows what to expect when setting/getting config values,
- **sensitivity** is now set directly when storing a new config value.

### limitation

- _sensitivity_ status and _type_ of a value cannot be removed/changed by mistake.
- While it is easy to configure value as _sensitive_ on set, to switch back the status to non-sensitive the app must call `updateSensitive()`
- Once typed, a value must be deleted in order to change its type.

### `oc_appconfig`

New fields in the `appconfig` table: 

- **lazy** (bool): 0 is the normal/default way to store a config value,
- **type** (int): will store the type and the sensitive status of a config value. This is a bit flag using those const:

```
	VALUE_SENSITIVE = 1;
	VALUE_MIXED = 2;
	VALUE_STRING = 4;
	VALUE_INT = 8;
	VALUE_FLOAT = 16;
	VALUE_BOOL = 32;
	VALUE_ARRAY = 64;
```

### occ

**occ config:app:get**

It adds the `--details` option to the `config:app:get`:

```
$ ./occ config:app:get myapp mykey --details
{
    "app": "myapp",
    "key": "mykey",
    "value": "ouila",
    "lazy": true,
    "type": 4,
    "typeString": "string",
    "sensitive": false
}
```

**occ config:app:set**

New options:

- `--lazy`, `--no-lazy` to change the way to store/retrieve the config value,
- `--sensitive`, `--no-sensitive` to change the sensitivity of a config value,
- `--type` to set/change the type of a value,
- Unless using `--no-interaction` a confirmation from the admin is needed when editing the type/status of a config value.

### Note(s)

_PR is big and adds huge edits. It should stays compatible with previous call to AppConfig._


lib/private/LazyConfig.php
```diff
@@ -0,0 +1,566 @@
...
+
+		$qb = $this->connection->getQueryBuilder();
+		$qb->select('config_key', 'config_value')
+			->from('appconfig_lazy')
+			->where($qb->expr()->eq('app_id', $qb->createNamedParameter($appId)));
+
+		if (!$includeExtras) {
+			$qb->andWhere($qb->expr()->eq('extra', $qb->createNamedParameter($extra)));
+		}
+		$result = $qb->execute();
```
come-nc, line 501:
```suggestion
		$result = $qb->executeQuery();
```

core/Migrations/Version29000Date20231126110901.php
```diff
@@ -0,0 +1,60 @@
...
+			$table->addIndex(['app_id', 'extra'], 'lazy_app_id_i');
+			$table->addIndex(['app_id', 'extra', 'config_key'], 'lazy_app_id_key_i');
+			$updated = true;
+		}
+
+		if (!$updated) {
+			return null;
+		}
+
+		return $schema;
```
come-nc, line 58:
```suggestion
		if (!$schema->hasTable('appconfig_lazy')) {
			$table = $schema->createTable('appconfig_lazy');
			$table->addColumn('app_id', Types::STRING, ['length' => 32]);
			$table->addColumn('extra', Types::STRING, ['length' => 32]);
			$table->addColumn('config_key', Types::STRING, ['length' => 64]);
			$table->addColumn('config_value', Types::TEXT);

			$table->setPrimaryKey(['app_id', 'config_key'], 'lazy_app_prim');
			$table->addIndex(['app_id', 'extra'], 'lazy_app_id_i');
			$table->addIndex(['app_id', 'extra', 'config_key'], 'lazy_app_id_key_i');
			return $schema;
		}

		return null;
```

core/Migrations/Version29000Date20231126110901.php
```diff
@@ -0,0 +1,62 @@
...
+			return null;
+		}
+
+		$table = $schema->getTable('appconfig');
+		if ($table->hasColumn('lazy_group')) {
+			return null;
+		}
+
+		$table->addColumn('lazy_group', Types::STRING, ['length' => 32, 'notnull' => true, 'default' => '']);
+		$table->addColumn('sensitive', Types::SMALLINT, ['length' => 1, 'notnull' => true, 'default' => '0']);
```
nickvergessen, line 50:
```suggestion
		$table->addColumn('lazy_group', Types::STRING, ['length' => 32, 'notnull' => false, 'default' => '']);
		$table->addColumn('sensitive', Types::SMALLINT, ['length' => 1, 'notnull' => true, 'default' => 0]);
```

lib/public/IConfig.php
```diff
@@ -126,6 +126,7 @@ public function deleteSystemValue($key);
 	 * @param string $appName the appName that we stored the value under
 	 * @return string[] the keys stored for the app
 	 * @since 8.0.0
+	 * @deprecated - load IAppConfig directly
```
nickvergessen, line 129:
```suggestion
	 * @deprecated 29.0.0 Use {@see IAppConfig} directly
```

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
 	 * get multiply values, either the app or key can be used as wildcard by setting it to false
+	 * @deprecated use getAllValues()
 	 *
 	 * @param string|false $key
 	 * @param string|false $app
+	 *
 	 * @return array|false
 	 * @since 7.0.0
+	 * @deprecated use getAllValues()
+	 * @see self::getAllValues()
```
nickvergessen, line 406:
```suggestion
	 * @deprecated 29.0.0 Use {@see getAllValues()}
```

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * #######################################################################
+	 * # Below this mark are the method deprecated and replaced since 29.0.0 #
+	 * #######################################################################
+	 *
+	 */
 
 	/**
 	 * get multiply values, either the app or key can be used as wildcard by setting it to false
+	 * @deprecated use getAllValues()
```
nickvergessen, line 398:
```suggestion
```

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
 	 * @return array|false
 	 * @since 7.0.0
+	 * @deprecated use getAllValues()
+	 * @see self::getAllValues()
 	 */
 	public function getValues($app, $key);
 
 	/**
 	 * get all values of the app or and filters out sensitive data
+	 * @deprecated use getAllValues()
```
nickvergessen, line 412:
```suggestion
```

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
 	 *
 	 * @param string $app
+	 *
 	 * @return array
 	 * @since 12.0.0
+	 * @see self::getAllValues()
```
nickvergessen, line 418:
```suggestion
	 * @deprecated 29.0.0 Use {@see getAllValues()}
```

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
-
-	/**
-	 * Get all apps using the config
-	 * @return string[] an array of app ids
-	 *
-	 * This function returns a list of all apps that have at least one
-	 * entry in the appconfig table.
-	 * @since 7.0.0
-	 */
-	public function getApps();
```
nickvergessen, line 71: Can not break public API, needs to be deprecated for 3 years.

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * Clearing cache consist of emptying the internal cache and the distributed cache.
+	 * The cache will be rebuilt only the next time a config value is requested.
+	 *
+	 * @since 29.0.0
+	 */
+	public function clearCache(): void;
+
+	/**
+	 * For debug purpose.
```
nickvergessen, line 380: Is this meant to be used by apps? I guess no?
I would not put it into the public API

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	public function setValueArray(string $app, string $key, array $value, bool $lazy = false, ?bool $sensitive = null): bool;
+
+	/**
+	 * Delete single config key from database.
+	 *
+	 * @param string $app id of the app
+	 * @param string $key config key
+	 * @since 29.0.0
+	 */
+	public function unsetKey(string $app, string $key): void;
```
nickvergessen, line 359: Let's stay with delete as name?!

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+
+	/**
+	 * delete all config keys linked to an app
+	 *
+	 * @param string $app id of the app
+	 * @since 29.0.0
+	 */
+	public function unsetAppKeys(string $app): void;
```
nickvergessen, line 367: Let's stay with delete as name?!

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * @return bool TRUE if value was different, therefor updated in database
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy grouping
+	 * @see self::setValueInt()
+	 * @see self::setValueFloat()
+	 * @see self::setValueBool()
+	 * @see self::setValueArray()
+	 */
+	public function setValueString(string $app, string $key, string $value, bool $lazy = false, ?bool $sensitive = null): bool;
```
nickvergessen, line 266: New method, let's force developers to do it properly.
Sensitive can not really have 3 states. Either it is sensitive or not.

```suggestion
	public function setValueString(string $app, string $key, string $value, bool $lazy = false, bool $sensitive = false): bool;
```

I would even argue that security should not be optional and neither should be the sensitive flag?

```suggestion
	public function setValueString(string $app, string $key, string $value, bool $sensitive, bool $lazy = false): bool;
```

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * @return bool TRUE if value was different, therefor updated in database
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy grouping
+	 * @see self::setValueString()
+	 * @see self::setValueFloat()
+	 * @see self::setValueBool()
+	 * @see self::setValueArray()
+	 */
+	public function setValueInt(string $app, string $key, int $value, bool $lazy = false, ?bool $sensitive = null): bool;
```
nickvergessen, line 287: See comment on setValueString

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * @return bool TRUE if value was different, therefor updated in database
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy grouping
+	 * @see self::setValueString()
+	 * @see self::setValueInt()
+	 * @see self::setValueBool()
+	 * @see self::setValueArray()
+	 */
+	public function setValueFloat(string $app, string $key, float $value, bool $lazy = false, ?bool $sensitive = null): bool;
```
nickvergessen, line 308: See comment on setValueString

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * @return bool TRUE if value was different, therefor updated in database
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy grouping
+	 * @see self::setValueString()
+	 * @see self::setValueInt()
+	 * @see self::setValueFloat()
+	 * @see self::setValueArray()
+	 */
+	public function setValueBool(string $app, string $key, bool $value, bool $lazy = false, ?bool $sensitive = null): bool;
```
nickvergessen, line 329: See comment on setValueString

nickvergessen: That being said, can a boolean be sensitive?

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * @return bool TRUE if value was different, therefor updated in database
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy grouping
+	 * @see self::setValueString()
+	 * @see self::setValueInt()
+	 * @see self::setValueFloat()
+	 * @see self::setValueBool()
+	 */
+	public function setValueArray(string $app, string $key, array $value, bool $lazy = false, ?bool $sensitive = null): bool;
```
nickvergessen, line 350: See comment on setValueString

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	/**
+	 * Store a config key and its value in database
+	 * If config key is already known with the exact same config value, the database is not updated.
+	 * If config key is not supposed to be read during the boot of the cloud, it is advised to set it as lazy loaded.
+	 *
+	 * @param string $app id of the app
+	 * @param string $key config key
+	 * @param string $value config value
+	 * @param bool|null $sensitive value should be hidden when needed. if NULL sensitive flag is not changed in database
+	 * @param string $lazy name of the lazy group
```
nickvergessen, line 256:
```suggestion
	 * @param bool $lazy set config as lazy loaded
```

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 */
+	public function statusCache(): array;
+
+	/*
+	 *
+	 * #######################################################################
+	 * # Below this mark are the method deprecated and replaced since 29.0.0 #
+	 * #######################################################################
+	 *
+	 */
```
nickvergessen, line 394:
```suggestion
```

Each method needs to indicate this anyway

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 * @see self::getValueInt()
+	 * @see self::getValueFloat()
+	 * @see self::getValueArray()
+	 */
+	public function getValueBool(string $app, string $key, bool $default = false, bool $lazy = false): bool;
+
+	/**
+	 * Get config value assigned to a config key.
+	 * If config key is not found in database, default value is returned.
+	 * If config key belongs to a lazy group, the name of the lazy group needs to be specified.
```
nickvergessen, line 230: Lazy is not a string/name anymore

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 * @see self::getValueInt()
+	 * @see self::getValueBool()
+	 * @see self::getValueArray()
+	 */
+	public function getValueFloat(string $app, string $key, float $default = 0, bool $lazy = false): float;
+
+	/**
+	 * Get config value assigned to a config key.
+	 * If config key is not found in database, default value is returned.
+	 * If config key belongs to a lazy group, the name of the lazy group needs to be specified.
```
nickvergessen, line 210: Lazy is not a string/name anymore

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 * @see self::getValueFloat()
+	 * @see self::getValueBool()
+	 * @see self::getValueArray()
+	 */
+	public function getValueInt(string $app, string $key, int $default = 0, bool $lazy = false): int;
+
+	/**
+	 * Get config value assigned to a config key.
+	 * If config key is not found in database, default value is returned.
+	 * If config key belongs to a lazy group, the name of the lazy group needs to be specified.
```
nickvergessen, line 190: Lazy is not a string/name anymore

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 * @see self::getValueFloat()
+	 * @see self::getValueBool()
+	 * @see self::getValueArray()
+	 */
+	public function getValueString(string $app, string $key, string $default = '', bool $lazy = false): string;
+
+	/**
+	 * Get config value assigned to a config key.
+	 * If config key is not found in database, default value is returned.
+	 * If config key belongs to a lazy group, the name of the lazy group needs to be specified.
```
nickvergessen, line 170: Lazy is not a string/name anymore

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+	 *
+	 * @return array<string, string> [appId => configValue]
+	 * @since 29.0.0
+	 */
+	public function searchValues(string $key, bool $lazy = false): array;
+
+	/**
+	 * Get config value assigned to a config key.
+	 * If config key is not found in database, default value is returned.
+	 * If config key belongs to a lazy group, the name of the lazy group needs to be specified.
```
nickvergessen, line 150: Lazy is not a string/name anymore

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
...
+ * In order to avoid loading useless config values in memory for each request on
+ * the cloud, it has been made possible to group your config keys.
+ * Each group, called _lazy group_, is only loaded in memory when one its config
+ * keys is retrieved.
+ *
+ * It is advised to only use the default lazy group, named '' (empty string), for
+ * config keys used in the registered part of your code that is called even when
+ * your app is not boot (as in event listeners, ...)
+ *
+ * **Note:** Lazy group are not linked to app ids. Multiple app can share the same
```
nickvergessen, line 48: Lazy is not a string/name anymore

lib/public/IAppConfig.php
```diff
@@ -29,44 +32,390 @@
 /**
  * This class provides an easy way for apps to store config values in the
  * database.
+ *
+ * **Note:** since 29.0.0, it supports **lazy grouping**
+ *
+ * ### What is lazy grouping ?
+ * In order to avoid loading useless config values in memory for each request on
+ * the cloud, it has been made possible to group your config keys.
+ * Each group, called _lazy group_, is only loaded in memory when one its config
```
nickvergessen, line 41: Lazy is not a string/name anymore

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
+	 * @param string $key config key
+	 * @param bool $lazy search within lazy config (since 29.0.0)
+	 *
+	 * @return bool TRUE if key exists
+	 * @since 7.0.0
+	 */
+	public function hasKey(string $app, string $key, bool $lazy = false): bool {
+		$this->assertParams($app, $key);
+		$this->loadConfig($lazy);
+		($lazy) ? $cache = &$this->lazyCache : $cache = &$this->fastCache;
```
nickvergessen, line 162: Inline assignment of hell 🙈 
You even need to silence psalm for it …

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
+	 *
+	 * @return array<string, string> [appId => configValue]
+	 * @since 29.0.0
+	 */
+	public function searchValues(string $key, bool $lazy = false): array {
+		$this->assertParams('', $key, true);
+		$this->loadConfig($lazy);
+		$values = [];
+		/** @var array<array-key, array<array-key, mixed>> $cache */
+		($lazy) ? $cache = &$this->lazyCache : $cache = &$this->fastCache;
```
nickvergessen, line 261: Also here

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
+	 * @see self::getValueArray()
 	 */
-	public function getValue($app, $key, $default = null) {
-		$this->loadConfigValues();
+	public function getValueString(string $app, string $key, string $default = '', bool $lazy = false): string {
+		$this->assertParams($app, $key);
+		$this->loadConfig($lazy);
+		($lazy) ? $cache = &$this->lazyCache : $cache = &$this->fastCache;
+
+		/** @psalm-suppress UndefinedVariable */
```
nickvergessen, line 294: Instead of surpressing, fix the code and do it properly:
```suggestion
		if ($lazy) {
			$cache = &$this->lazyCache;
		} else {
			$cache = &$this->fastCache;
		}
```

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy grouping
+	 * @see self::getValueString()
+	 * @see self::getValueInt()
+	 * @see self::getValueFloat()
+	 * @see self::getValueArray()
+	 */
+	public function getValueBool(string $app, string $key, bool $default = false, bool $lazy = false): bool {
+		return in_array(
+			$this->getValueString($app, $key, $default ? 'true' : 'false'), ['1', 'true', 'yes', 'on'], $lazy
```
nickvergessen, line 356:
```suggestion
			$this->getValueString($app, $key, $default ? 'true' : 'false', $lazy), ['1', 'true', 'yes', 'on']
```

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
+	 * @see self::getValueBool()
+	 */
+	public function getValueArray(string $app, string $key, array $default = [], bool $lazy = false): array {
+		try {
+			$defaultJson = json_encode($default, JSON_THROW_ON_ERROR);
+			$value = json_decode(
+				$this->getValueString($app, $key, $defaultJson, $lazy), true, JSON_THROW_ON_ERROR
+			);
+
+			return (is_array($value)) ? $value : [$value];
```
nickvergessen, line 383: Clean code in new methods is better then allowing bad patterns from the start:
```suggestion
			return is_array($value) ? $value : [];
```

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
+	 * if $lazy is:
+	 * - false: set fast config as loaded
+	 * - true : set lazy config as loaded
+	 * - null : set both config as loaded
+	 *
+	 * @param bool $lazy
+	 */
+	private function setAsLoaded(?bool $lazy): void {
+		if ($lazy !== null) {
+			$lazy ? $this->lazyLoaded = true : $this->fastLoaded = true;
```
nickvergessen, line 795: no inline assignments

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
 
 	/**
-	 * Get the available keys for an app
+	 * @inheritDoc
 	 *
-	 * @param string $app the app we are looking for
-	 * @return array an array of key names
+	 * @param string $app id of the app
+	 * @param string $key config key
+	 * @param bool $lazy search within lazy config (since 29.0.0)
```
nickvergessen, line 154: Add as `@since` tag?

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1041 @@
...
+		   ->where($qb->expr()->eq('appid', $qb->createNamedParameter($app)))
+		   ->andWhere($qb->expr()->eq('configkey', $qb->createNamedParameter($key)));
+		$qb->executeStatement();
 
+		// we really want to delete that key
+		unset($this->lazyCache[$app][$key]);
 		if ($this->hasKey($app, $key)) {
-			return $this->cache[$app][$key];
+			unset($this->fastCache[$app][$key]);
+			$this->storeDistributedCache();
```
nickvergessen, line 604: The data in your cache can be really old,
could be multiple minutes if this is done at the last jobs of a cron run, and you would then store previous old appconfigs again.

nickvergessen (changes requested):
Really not a fan of the caching strategy.
It sounds really risky to get database and cache out of sync easily with a long running task in OCC or request, etc.

As per chat however I like the general idea of having a lazy "block" and even more integrating the sensitive part directly.

Speaking of sensitive some apps use the `ICrypto::encrypt()` and `ICrypto::dectrypt()`. We could include that into the methods and help developers again to not have security optionally?

lib/private/AppConfig.php
ArtificialOwl: i'll do that, but I am more inclined to think that this is an issue with psalm, not my code :-]

lib/private/AppConfig.php
ArtificialOwl: should we compare the local/internal cached value with the one in database when doing operation like update/delete ? or store a random string/token in database to implement some versioning ?

lib/public/IAppConfig.php
ArtificialOwl: I am okay as long as we know for sure that replacing the `deleteKey()` and `deleteApp()` wont break anything

lib/public/IAppConfig.php
ArtificialOwl: i am fine with public(), Circles needs it

lib/private/AppConfig.php
nickvergessen: Multiple sinces are allowed

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1061 @@
...
+		string $app,
+		string $key,
+		array $default = [],
+		bool $lazy = false
+	): array {
+		try {
+			$defaultJson = json_encode($default, JSON_THROW_ON_ERROR);
+			$value = json_decode(
+				$this->getValueString($app, $key, $defaultJson, $lazy), true, JSON_THROW_ON_ERROR
+			);
```
nickvergessen, line 388:
```suggestion
			$value = json_decode(
				$this->getValueString($app, $key, $defaultJson, $lazy), true, 512, JSON_THROW_ON_ERROR
			);
```

or

```suggestion
			$value = json_decode(
				$this->getValueString($app, $key, $defaultJson, $lazy), true, flags: JSON_THROW_ON_ERROR
			);
```

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1061 @@
...
+				$key,
+				json_encode($value, JSON_THROW_ON_ERROR),
+				$lazy,
+				self::VALUE_ARRAY | ($sensitive ? self::VALUE_SENSITIVE : 0)
+			);
+		} catch (JsonException $e) {
+			$this->logger->warning(
+				'could not setValueArray', ['app' => $app, 'key' => $key, 'exception' => $e]
+			);
 		}
```
nickvergessen, line 562:
```suggestion
		} catch (JsonException $e) {
			$this->logger->error(
				'could not setValueArray', ['app' => $app, 'key' => $key, 'exception' => $e]
			);
			throw $e;
		}
```

This is too important. It basically means the function did not operate at all. This should not go silent. App authors need to change their code immediately, so this should log as error and throw hard.

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1061 @@
...
+		 * We ignore if mixed type is request.
+		 * If type of stored value is set as mixed, we don't filter.
+		 * If type of stored value is defined, we compare with the one requested.
+		 */
+		$knownType = $this->valueTypes[$app][$key] ?? 0;
+		if (!$this->isTyped(self::VALUE_MIXED, $type)
+			&& $knownType > 0
+			&& !$this->isTyped(self::VALUE_MIXED, $knownType)
+			&& !$this->isTyped($type, $knownType)) {
+			throw new AppConfigIncorrectTypeException('conflict with value type from database');
```
nickvergessen, line 663: Let's log this as an error (because it again means app authors need to do something, or administrators manipulated data wrongly) with the information which key was expected which type and what type it actually was

lib/private/AppConfig.php
nickvergessen, line 388:
```suggestion
			$value = json_decode(
				$this->getValueString($app, $key, $defaultJson, $lazy), true, 512, JSON_THROW_ON_ERROR
			);
```

or

```suggestion
			$value = json_decode(
				$this->getValueString($app, $key, $defaultJson, $lazy), true, flags: JSON_THROW_ON_ERROR
			);
```

nickvergessen (commented):
We should add a type option to `occ config:app:set` similar to `occ config:system:set`, so you can change an app config and retain the type info.

lib/private/AppConfig.php
ArtificialOwl: nice catch

lib/private/AppConfig.php
ArtificialOwl: @nickvergessen we keep the exception, or no exception thrown ?

ArtificialOwl:
> We should add a type option to `occ config:app:set` similar to `occ config:system:set`, so you can change an app config and retain the type info.

in this PR ?

ArtificialOwl:
anyone willing to confirm the fail test 

```
--

There was 1 failure:

1) Test\DB\QueryBuilder\FunctionBuilderTest::testConcatString with data set "2 columns: string real and int literal" (Closure Object (...))
--- Expected
+++ Actual
@@ @@
-'12'
+'22'

/home/runner/work/server/server/tests/lib/DB/QueryBuilder/FunctionBuilderTest.php:62
```
as it is working locally

icewind1991:
Not a fan of using bitflags for the type and sensitive flags. Rather just have a separate bool for the sensitive flag and int/enum for the type.

ArtificialOwl:
> Not a fan of using bitflags for the type and sensitive flags. Rather just have a separate bool for the sensitive flag and int/enum for the type.

Might not be the best but I really wanted to keep it as less invasive as possible in database. 
This is readonly (no index/search will be done on the value itself), and might allow us to add small other boolean bitflag in the future without creating a new field to the table

lib/private/SystemConfig.php
```diff
@@ -123,11 +123,7 @@ class SystemConfig {
 		],
 	];
 
-	/** @var Config */
-	private $config;
-
-	public function __construct(Config $config) {
-		$this->config = $config;
+	public function __construct(private Config $config) {
```
come-nc, line 126:
```suggestion
	public function __construct(
		private Config $config,
	) {
```

lib/public/IAppConfig.php
```diff
@@ -26,47 +29,500 @@
...
+	 * @return string[] list of app ids
+	 * @since 7.0.0
+	 */
+	public function getApps(): array;
+
+	/**
+	 * Returns all keys stored in database, related to an app.
+	 * Please note that the values are not returned.
+	 *
+	 * **WARNING:** ignore lazy filtering
```
come-nc, line 82: Not clear, does that means lazy keys are always returned or never, or something else?

lib/public/IAppConfig.php
```diff
@@ -26,47 +29,500 @@
...
+	 *
+	 * @param string $app id of the app
+	 *
+	 * @return string[] list of stored config keys
+	 * @since 7.0.0
```
come-nc, line 87: `@since` should be 29.0.0 for all new functions on this interface

lib/public/IAppConfig.php
come-nc: circles need it for debug or normal usage?

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1378 @@
...
+	 * @see getValueArray()
 	 */
-	public function getValue($app, $key, $default = null) {
-		$this->loadConfigValues();
+	public function getValueInt(
+		string $app,
+		string $key,
+		int $default = 0,
+		bool $lazy = false
+	): int {
```
come-nc, line 316:
```suggestion
	): int|float {
```
If you want bigint support on 32bit

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1378 @@
...
+				'mixed' => IAppConfig::VALUE_MIXED,
+				'string' => IAppConfig::VALUE_STRING,
+				'integer' => IAppConfig::VALUE_INT,
+				'float' => IAppConfig::VALUE_FLOAT,
+				'boolean' => IAppConfig::VALUE_BOOL,
+				'array' => IAppConfig::VALUE_ARRAY
+			};
+		} catch (UnhandledMatchError) {
+			throw new AppConfigIncorrectTypeException('Unknown type ' . $type);
+		}
```
come-nc, line 972:
```suggestion
		return match (strtolower($type)) {
			'mixed' => IAppConfig::VALUE_MIXED,
			'string' => IAppConfig::VALUE_STRING,
			'integer' => IAppConfig::VALUE_INT,
			'float' => IAppConfig::VALUE_FLOAT,
			'boolean' => IAppConfig::VALUE_BOOL,
			'array' => IAppConfig::VALUE_ARRAY,
			default => throw new AppConfigIncorrectTypeException('Unknown type ' . $type),
		};
```
Same thing but simpler no?

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1378 @@
...
-					$values[$sensitiveKey] = IConfig::SENSITIVE_VALUE;
-				}
+	/**
+	 * @param int $needle bitflag to search
+	 * @param int $type known value
+	 *
+	 * @return bool TRUE if bitflag $needle is set in $type
+	 */
+	private function isTyped(int $needle, int $type): bool {
+		return (($needle & $type) !== 0);
```
come-nc, line 1073:
```suggestion
		return (($needle & $type) === $needle);
```
This would support testing several flags at the same time I think

lib/private/AppConfig.php
```diff
@@ -30,413 +33,1378 @@
...
+	 *
+	 * @return string[] list of app ids
+	 * @since 7.0.0
+	 */
+	public function getApps(): array {
+		$this->loadConfigAll();
+		$apps = array_merge(array_keys($this->fastCache), array_keys($this->lazyCache));
+		sort($apps);
+
+		return $apps;
```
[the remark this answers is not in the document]
come-nc: No https://www.php.net/manual/en/function.array-merge.php

lib/private/AppConfig.php
ArtificialOwl: no needed

come-nc (approved):

artonge (approved):

lib/private/AppConfig.php
```diff
@@ -30,416 +33,1504 @@
...
-			$result = array_combine($appIds, $values);
+			$cache = &$this->fastCache;
+		}
 
-			return array_filter($result);
+		$type = $this->getValueType($app, $key);
+		try {
+			$typeString = $this->convertTypeToString($type);
+		} catch (AppConfigIncorrectTypeException $e) {
+			$this->logger->warning('type stored in database is not correct', ['exception' => $e, 'type' => $type]);
```
nickvergessen, line 1061: This spams `Unknown numeric type 0` very heavily for me.

lib/public/IAppConfig.php
```diff
@@ -26,47 +29,548 @@
...
+ * is requested.
+ *
+ * It is advised to set a config key as lazy when its value is only needed during
+ * really specific request, in part of code that is not called frequently.
+ *
+ * **Note:** some methods from this class are marked with a warning about ignoring
+ * lazy filtering, meaning it will load in memory all apps config values. use them
+ * wisely and only in part of code called during specific request/action.
+ *
+ * @since 29.0.0 supports lazy loading
```
nickvergessen, line 56: this differs from the actual implementation. Use either and make sure it doesn't mention the lazy groups :P

lib/public/IAppConfig.php
```diff
@@ -26,47 +29,548 @@
...
  * @since 7.0.0
  */
 interface IAppConfig {
+	public const VALUE_SENSITIVE = 1;
+	public const VALUE_MIXED = 2;
+	public const VALUE_STRING = 4;
+	public const VALUE_INT = 8;
+	public const VALUE_FLOAT = 16;
+	public const VALUE_BOOL = 32;
+	public const VALUE_ARRAY = 64;
```
nickvergessen, line 66: please add the `@since`

lib/public/IAppConfig.php
```diff
@@ -26,47 +29,548 @@
...
+	 * @return int|float stored config value or $default if not set in database
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy loading
+	 * @see getValueString()
+	 * @see getValueInt()
+	 * @see getValueFloat()
+	 * @see getValueBool()
+	 * @see getValueArray()
+	 */
+	public function getValueBigInt(string $app, string $key, int|float $default = 0, bool $lazy = false): int|float;
```
nickvergessen, line 220: Still not sure when this should be used, also the docs don't help to decide.
When I store e.g. a "file id" which could be BIGINT on DB level, should I use this, or MUST I use this?

lib/public/IAppConfig.php
```diff
@@ -26,47 +29,548 @@
...
+	 * @param bool $lazy search within lazy loaded config
+	 *
+	 * @return string stored config value or $default if not set in database
+	 * @since 29.0.0
+	 * @see IAppConfig for explanation about lazy loading
+	 * @see getValueInt()
+	 * @see getValueBigInt()
+	 * @see getValueFloat()
+	 * @see getValueBool()
+	 * @see getValueArray()
```
nickvergessen, line 176: I would remove all those duplicate references, just "bloats" the doc block from my pov.
```suggestion
```

nickvergessen:
❌ Drone failures are related!

lib/public/IAppConfig.php
ArtificialOwl: `setValueBigInt`/`getValueBigInt` is supposed to help supports huge integer above 2,147,483,647 (on 32bits) or 9,223,372,036,854,775,807, as php converts them to float (mostly for 32 bits support ;p)



if I comment this exception
https://github.com/nextcloud/server/pull/41755/files#diff-3921e92ccf03a5dc1400f57821a7bc829b242e7691ff1991cda3a3c431fabc6aR201-R203

This is the behavior:
```
$ ./occ config:app:set toto1 toto2 --value 9223372036854775808 --type integer
Config value 'toto2' for app 'toto1' is now set to '9223372036854775807', stored as integer in fast cache
```

As you can see, the `int` value is cap, using an int|float would avoid this by switching to float (only) when needed.
The other solution would be to type `setValueInt()`/`getValueInt()` with `float|int`, but it also means that value is not `int` at 100% and I don't think we want this.

This is currently not implemented, but if we allow the use of bigint type in the occ command, this would be the behavior
```
$ ./occ config:app:set toto1 toto2 --value 9223372036854775808 --type bigint
Config value 'toto2' for app 'toto1' is now set to '9.2233720368548E+18', stored as integer in fast cache
$ ./occ config:app:set toto1 toto2 --value 9223372036854775807 --type integer
Config value 'toto2' for app 'toto1' is now set to '9223372036854775807', stored as integer in fast cache
```

lib/private/AppConfig.php
ArtificialOwl: I am interested how you came to this as type should not be 0 in database

nickvergessen: Mostlikely from `occ config:app:set -vvv ${dir##*/} enabled --value="no"` which I have in my set up script to force disable all apps as an option.
The point is more that there will be a lot of "dirty" instances already out there and spaming their logs with warnings when the admin can not know what to do about it is meh.

lib/private/AppConfig.php
```diff
@@ -30,416 +33,1397 @@
...
+	 */
+	public function setValueInt(
+		string $app,
+		string $key,
+		int $value,
+		bool $lazy = false,
+		bool $sensitive = false
+	): bool {
+		if ($value > 2147400000) {
+			$this->logger->debug('You are trying to store an integer value around/above 2,147,483,647. This is a reminder that reaching this theoretical limit on 32 bits system will result to an exception.');
```
nickvergessen, line 579:
```suggestion
			$this->logger->debug('You are trying to store an integer value around/above 2,147,483,647. This is a reminder that reaching this theoretical limit on 32 bits system will throw an exception.');
```

core/Command/Config/App/SetConfig.php
```diff
@@ -69,15 +95,176 @@
...
+			 * If --type is specified in the command line, we upgrade the type in database
+			 * after a confirmation from admin.
+			 * If not we get the type from current stored value or VALUE_MIXED as default.
+			 */
+			try {
+				$currType = $this->appConfig->getValueType($appName, $configName);
+				if ($typeString === null || $type === $currType || !$this->ask($input, $output, $typeString)) {
+					$type = $currType;
+				} else {
+					$updated = $this->appConfig->updateType($appName, $configName, $type);
```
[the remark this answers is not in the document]
nickvergessen: > Argument 3 of OC\AppConfig::updateType cannot be null, possibly null value provided

nickvergessen (commented):
Okay otherwise

lib/public/IAppConfig.php
ArtificialOwl: fixed by removing `setValueBigInt`/`getValueBigInt`. Also, using setValueInt() with huge integer value, close to the 32bits limit will generate a debug log entry.

lib/private/AppConfig.php
```diff
@@ -30,416 +33,1504 @@
...
+			&& !$this->isTyped(self::VALUE_MIXED, $knownType)
+			&& !$this->isTyped($type, $knownType)) {
+			$this->logger->warning('conflict with value type from database', ['app' => $app, 'key' => $key, 'type' => $type, 'knownType' => $knownType]);
+			throw new AppConfigTypeConflictException('conflict with value type from database');
+		}
+
+		if ($lazy) {
+			$cache = &$this->lazyCache;
+		} else {
+			$cache = &$this->fastCache;
```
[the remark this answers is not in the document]
ArtificialOwl: fixed in all places

lib/private/AppConfig.php
ArtificialOwl: I cannot reproduce this issue and will merge as it is right now.
By default, `occ config:app:set` set value as mixed. 

The only issue that I am seeing when using the `occ config:app:set` is that during the upgrading process, between the files are changed and the upgrade process is run.

Maybe we can edit the database (adding type+lazy field) already in 28 so we don't have error about missing field during the upgrade steps

nickvergessen (approved):
At least needs an issue in the guests app to make them aware of the breakage that is coming their way:
https://github.com/nextcloud/guests/blob/master/lib/AppConfigOverwrite.php

juliushaertl:
Just noticed we also have two different IAppConfig classes (one generic and one as a app framework service). The app framework service should probably also get adapted to cover the new API as currently it is just a wrapper for the deprecated methods: https://github.com/nextcloud/server/blob/215aef3cbdc1963be1bb6bca5218ee0a4b7f1665/lib/public/AppFramework/Services/IAppConfig.php

nickvergessen:
Should we just deprecate it? we shouldn't have 2 similar interfaces like this?

juliushaertl:
I still like the service approach to simplify and not need to pass in the app id which makes the api a bit easier for app developers and and more restrictive to not mess with other apps configs

nickvergessen:
Ah, missed that bit. It's basically what I abstracted away in talk by using my own config class :D

core/Command/Config/App/SetConfig.php
```diff
@@ -56,6 +63,25 @@ protected function configure() {
 				InputOption::VALUE_REQUIRED,
 				'The new value of the config'
 			)
+			->addOption(
+				'type',
+				null,
+				InputOption::VALUE_REQUIRED,
+				'Value type [string, integer, float, boolean, array]',
+				'string'
+			)
```
ChristophWurst, line 72: This would be important for https://docs.nextcloud.com/server/stable/admin_manual/configuration_server/occ_command.html#setting-a-single-configuration-value. @ArtificialOwl @sorbaugh please document this soon :pray:

joshtrichards: Array handling still needs to be documented. Particularly since it's diverged from `system`: https://docs.nextcloud.com/server/latest/admin_manual/occ_command.html#setting-an-array-configuration-value

PR #41758, opened 2023-11-27
Delete dot

Valdnet:


susnux:
Is this needed? I mean I am fine with this, but it makes the sentence a lot longer and thus harder for some people to read.

PR #41765, opened 2023-11-27
Synchronize operation on live photo files

### LINKED ISSUE #40243: 🖼️🔄️ Synchronization of the two files
Implement storing of the metadata linking iOS Live Photos files and respect them on DAV operations applying the same DAV operation to the respective other file.

- [x] Allow clients to set a `livephoto` metadata in the `OC\Metadata` API through WebDAV. This metadata is set for the two files and contains the `fileid` of the other file. Another solution would be to use a header, but not a fan of multiplying custom headers. Also, allowing clients to write the `livephotos` metadata allows writing it for already uploaded files
- [ ] Add listeners to react to `Node*Event` - 
    - [x] `NodeDeletedEvent` to remove the second file
    - [x] `NodeRenamedEvent` to rename the second file
        - [x] `BeforeNodeRenameEvent` to block the operation if one of the two files cannot be renamed (ex: file with the same name at the location, or quota limitation, or … ?)
        - [ ] Test when renamed file is a share
    - [ ] `NodeCopiedEvent` to copy the second file
        - [ ] `BeforeNodeCopiedEvent` to block the operation if one of the two files cannot be copied (ex: file with the same name at the location, or quota limitation, or … ?)
        - [ ] Test when renamed file is a share
- [ ] Optional: block write permissions on the `.jpg` to prevent inconsistencies?

artonge:
* Resolves: https://github.com/nextcloud/server/issues/40243

This PR adds automatic management of live photos. This means that both file of a live photo will be kept in sync, if they are renamed, moved, or deleted.

A few extra changes were needed:

- I create a `BeforeNodeRestoredEvent` event to react to restore action from the trashbin
- I created `NodeRestoredEvent`, for completeness.
- Both are emitted from `apps/files_trashbin/lib/Trashbin.php`
- I added an `abortOperation` method in both `BeforeNodeDeletedEvent` and `BeforeNodeRenamedEvent` to allow an event listener to abort the operation.


AndyScherzinger (approved):

apps/files/lib/Listener/SyncLivePhotosListener.php
```diff
@@ -0,0 +1,252 @@
...
+			if ($peerFile !== null) {
+				return $peerFile;
+			}
+		}
+
+		$metadata->unset('files-live-photo');
+		return null;
+	}
+
+	private function getTrashItem(array $trashFolder, string $path): ?ITrashItem {
```
artonge, line 234: TODO: improve by adding an API to the trashbinManager

apps/files/lib/Listener/SyncLivePhotosListener.php
artonge, line 1: Split handle into sub methods

artonge, line 1: Add cypress tests

artonge, line 1: Validate logic with what's expected by the user, and add it as a comment

ArtificialOwl (approved):
🐘

blizzz:
fixed milestone. master is 29.

PR #30368, opened 2021-12-21
Allow DAV Object properties

tcitworld:
The current implementation only saves them as string. It seems they can be more complex than that, and that objects were saved directly.

You may find such objects saved in some production databases by executing:
```sql
SELECT * from oc_properties where propertyvalue = 'Object';
```

This seems to cause issues in some particular cases (https://github.com/nextcloud/calendar/issues/3551) like the calendar trashbin because of things like this and (apparently ?) Apple devices:
https://github.com/nextcloud/server/blob/0b01103137aea94c65f784980dfec645958faaf7/apps/dav/lib/CalDAV/CalDavBackend.php#L3208-L3211

This commit adds a repair job to clean all of these "broken" properties values, adds a new database column to save the type of the property, and handles converting from and to correct values.

Implementation is very similar to SabreDAV's own PDO backend: https://github.com/nextcloud/3rdparty/blob/4921806dfb1c5c309eac60195ed34e2749baf3c1/sabre/dav/lib/DAV/PropertyStorage/Backend/PDO.php

Fixes https://github.com/nextcloud/calendar/issues/3858#issuecomment-1003335709

PR #41775, opened 2023-11-27
[stable24] fix(dav): Name properties table columns explicitly

ChristophWurst:
* Ref https://github.com/nextcloud/server/pull/30368#issuecomment-1152462916

## Summary

And ignore any optional columns. This allows instances to run the expensive query for the new column in 25 in the background, while keeping the 24 installation live.

## TODO

- [x] Make changes
- [ ] Test changes

## Checklist

### DIFF
--- a/apps/dav/lib/DAV/CustomPropertiesBackend.php
+++ b/apps/dav/lib/DAV/CustomPropertiesBackend.php
@@ -234,7 +234,7 @@ class CustomPropertiesBackend implements BackendInterface {
 		}
 
 		$qb = $this->connection->getQueryBuilder();
-		$qb->select('*')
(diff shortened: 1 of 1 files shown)


juliushaertl:
Looking at the usages more closely i think it should be safe to even run without this patch as we continue only using propertyname and propertyvalue:

https://github.com/nextcloud/server/blob/fix/dav/explicit-properties-column-names/apps/dav/lib/DAV/CustomPropertiesBackend.php#L243
https://github.com/nextcloud/server/blob/fix/dav/explicit-properties-column-names/apps/dav/lib/DAV/CustomPropertiesBackend.php#L286

ChristophWurst:
Good catch. That seems to be the case indeed!

ChristophWurst:
Obsolete

PR #41785, opened 2023-11-28
Migrate openResult method to SearchResult component

nfebe:
Ref https://github.com/nextcloud/server/issues/41381#issuecomment-1825643655

After moving the `SearchResult` to its own component and leaving out the `openResult` method, the search results do not lead anywhere but but error out.


ChristophWurst (dismissed):
https://github.com/nextcloud/server/issues/41381#issuecomment-1812127225

this element should be a real link

Fenn-CS:
> https://github.com/nextcloud/server/issues/41381#issuecomment-1812127225
>
> this element should be a real link

@ChristophWurst using the `href` attribute of `NcListItem` would force the search result to open in a new tab, is this the desire behavior? 

https://github.com/nextcloud-libraries/nextcloud-vue/blob/master/src/components/NcListItem/NcListItem.vue#L319

ChristophWurst:
Then the component needs a new prop to influence this behavior

artonge (approved):

Fenn-CS:
@ChristophWurst that makes sense, I have created these issue for that here : https://github.com/nextcloud-libraries/nextcloud-vue/issues/4880 and here https://github.com/nextcloud/server/issues/41800

Because this breaks search entirely, so this PR should restore the old initial before of using JavaScript to trigger the routing.

AndyScherzinger (approved):

ChristophWurst (approved):
Works for simple clicks again

The approach of window.location is still there and problematic

PR #41792, opened 2023-11-28
Handle close GlobalSearchModal gracefully

nfebe:
Sorts out re-opening errors, such as needing two clicks after clicking outside, or modal flashes.

-------------------

The current close infrastructure modifies a prop which has no real effect aside bugs.

In addition, calling the `NcModal.close()` as the primary way to close the search modal instead of using the states defined in `GlobalSearch` view causing re-open bugs (Modal cannot open, needs to click twice, and other weird stuff).

Contributes to https://github.com/nextcloud/server/issues/41381


ChristophWurst (approved):
Code looks good. Didn't test

artonge (approved):
I think that using the `.sync` modifier can help in such cases: https://v2.vuejs.org/v2/guide/components-custom-events.html#sync-Modifier
But code looks good like that

AndyScherzinger (approved):

susnux (approved):

susnux:
drone error unrelated

Issue #11323, opened 2018-09-22
Option to display deleted files in directory tree


m1cm1c:
In Dropbox's web interface's directory tree, there is this neat option called "show deleted files":

![](https://i.imgur.com/pziNKSS.png)

When you click it, it shows the files that were deleted in the location you are right now. You can then restore them or look at their versions. This would be an awesome feature for Nextcloud because right now it's very difficult to restore deleted files.

m1cm1c:
This is not a duplicate of any of the mentioned issues.

RafalSkolasinski:
+1 on this - it seems like an awesome things to have!

szaimen:
cc @nextcloud/server is this feasible?

m1cm1c:
Would still be nice. I don't see how this would have gotten stale.

PR #41793, opened 2023-11-28
Add additional user profile attributes from LDAP

### LINKED ISSUE #40194: More LDAP User Profile Attributes
### How to use GitHub

No.

It would be fantastic if you could integrate more LDAP user profile attributes, such as birthday or timezone. This would make it easier to fully integrate LDAP user management with Nextcloud Contacts. Ideally any contact card field can be passed from LDAP. 

None, this only concerns LDAP and Nextcloud integration.

**Additional context**
None.

slapcat:
* Resolves: #40194 

## Summary
- Extend user_ldap User Profile Attributes to include birthday and anniversary.
- Update CardDAV converter to include these attributes in contacts for the Accounts address book.
- Add required constants to AccountManager.

## Checklist

- [x] Screenshots before/after for front-end changes:
before
after

**EDIT:** Documentation PR [#11323](https://github.com/nextcloud/documentation/pull/11323)


apps/dav/lib/CardDAV/Converter.php
```diff
@@ -134,6 +134,15 @@ public function createCardFromUser(IUser $user): ?VCard {
 				case IAccountManager::PROPERTY_ROLE:
 					$vCard->add(new Text($vCard, 'TITLE', $property->getValue(), ['X-NC-SCOPE' => $scope]));
 					break;
+				case IAccountManager::PROPERTY_BIRTHDATE:
+					$vCard->add(new Text($vCard, 'BDAY;VALUE=DATE', substr($property->getValue(), 0, 8), ['X-NC-SCOPE' => $scope]));
```
ChristophWurst, line 138: what value will $scope have?

apps/dav/lib/CardDAV/Converter.php
```diff
@@ -134,6 +134,15 @@ public function createCardFromUser(IUser $user): ?VCard {
...
+					break;
+				case IAccountManager::PROPERTY_ANNIVERSARYDATE:
+					$vCard->add(new Text($vCard, 'ANNIVERSARY;VALUE=DATE', substr($property->getValue(), 0, 8), ['X-NC-SCOPE' => $scope]));
+					/* iOS compatibility */
+					$vCard->add(new Text($vCard, 'ITEM3.X-ABDATE;TYPE=pref', substr($property->getValue(), 0, 8), ['X-NC-SCOPE' => $scope]));
```
kesselb, line 143: This "ITEM3." thing is a bit tricky: https://github.com/nextcloud/contacts/issues/223

apps/dav/lib/CardDAV/Converter.php
slapcat: In my test environment, $scope is `v2-local`. I only included it to match how the previous properties get set. It can be removed if not required.

apps/dav/lib/CardDAV/Converter.php
slapcat: IIUC, this issue and another like it I found (https://github.com/nextcloud/contacts/issues/2560) concern importing vCards with these ITEMx keys into NC and displaying them in the profile. This is not an issue regarding the LDAP => NC sync because the `ANNIVERSARY` key is also added, which NC understands and displays in the profile. I don't have an Android, but when I export from NC and import to Google Contacts in the browser, the anniversary date displays correctly. If Android *requires* the `X-ANDROID-CUSTOM` key to display the anniversary, I can add this also to the code.

apps/dav/lib/CardDAV/Converter.php
ChristophWurst: I'm asking because it's not clear to me if this property is always fine to expose to other users. Users can not change the scope and influence who sees their birth date.

apps/dav/lib/CardDAV/Converter.php
kesselb: I would like to know if it still works on iOS if you drop `ITEM3.X-ABDATE`? 

ANNIVERSARY is the only attribute we should write: https://www.rfc-editor.org/rfc/rfc6350.html#section-6.2.6

Everything else is off standard.

slapcat: It does not work on iOS at all without the `ITEM3.X-ABDATE`. When you create a contact on iOS natively, it uses this key as well instead of following the standard.

apps/dav/lib/CardDAV/Converter.php
slapcat: Since they are not necessary for this feature anyhow, I have removed the scope from those lines.

ChristophWurst: No scope has the same semantic as scope private, so the property will only show for the user

apps/dav/lib/CardDAV/Converter.php
kesselb: Hey @slapcat, thanks for checking! 

I guess we don't have much of a choice then ;) 

I created the attached vcard using the online version of icloud contacts. 
You can add many anniversary dates and thus the itemX prefix. 

As we only have one anniversary date right now, it should be fine to use item1.

image


[Jane Doe.vcf.txt](https://github.com/nextcloud/server/files/13532083/Jane.Doe.vcf.txt)

kesselb:
Thank you :+1: 

Will try to give it a test next week. 

I see a few failing tests. 
The following patch should fix one of them. 

```
Index: tests/lib/Accounts/AccountManagerTest.php
IDEA additional info:
<+>UTF-8
diff --git a/tests/lib/Accounts/AccountManagerTest.php b/tests/lib/Accounts/AccountManagerTest.php
--- a/tests/lib/Accounts/AccountManagerTest.php	(revision 7eaa573bac0f82b5b2b837104a508dab37c6e231)
+++ b/tests/lib/Accounts/AccountManagerTest.php	(date 1701547210963)
@@ -632,6 +632,18 @@
 				'scope' => IAccountManager::SCOPE_LOCAL,
 			],
 
+			[
+				'name' => IAccountManager::PROPERTY_BIRTHDATE,
+				'value' => '',
+				'scope' => IAccountManager::SCOPE_LOCAL,
+			],
+
+			[
+				'name' => IAccountManager::PROPERTY_ANNIVERSARYDATE,
+				'value' => '',
+				'scope' => IAccountManager::SCOPE_LOCAL,
+			],
+
 			[
 				'name' => IAccountManager::PROPERTY_PROFILE_ENABLED,
 				'value' => '1',

```

apps/settings/lib/Settings/Personal/PersonalInfo.php
```diff
@@ -167,6 +167,8 @@ public function getForm(): TemplateResponse {
 			'role' => $this->getProperty($account, IAccountManager::PROPERTY_ROLE),
 			'headline' => $this->getProperty($account, IAccountManager::PROPERTY_HEADLINE),
 			'biography' => $this->getProperty($account, IAccountManager::PROPERTY_BIOGRAPHY),
+			'birthDate' => $this->getProperty($account, IAccountManager::PROPERTY_BIRTHDATE),
```
kesselb, line 170: Seems unused because the view was not updated to show the data in the user profile. 
I'd say we should remove it here for now.

kesselb:
This should fix the second failing tests by updating the expected data. 

```diff
Index: apps/settings/tests/UserMigration/assets/account.json
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/apps/settings/tests/UserMigration/assets/account.json b/apps/settings/tests/UserMigration/assets/account.json
--- a/apps/settings/tests/UserMigration/assets/account.json	(revision 7eaa573bac0f82b5b2b837104a508dab37c6e231)
+++ b/apps/settings/tests/UserMigration/assets/account.json	(date 1701548042464)
@@ -1,1 +1,0 @@
-{"displayname":{"name":"displayname","value":"Emma Jones","scope":"v2-federated","verified":"0","verificationData":""},"address":{"name":"address","value":"920 Grass St","scope":"v2-local","verified":"0","verificationData":""},"website":{"name":"website","value":"","scope":"v2-local","verified":"0","verificationData":""},"email":{"name":"email","value":"","scope":"v2-federated","verified":"0","verificationData":""},"avatar":{"name":"avatar","value":"","scope":"v2-federated","verified":"0","verificationData":""},"phone":{"name":"phone","value":"","scope":"v2-local","verified":"0","verificationData":""},"twitter":{"name":"twitter","value":"","scope":"v2-local","verified":"0","verificationData":""},"fediverse":{"name":"fediverse","value":"","scope":"v2-local","verified":"0","verificationData":""},"organisation":{"name":"organisation","value":"","scope":"v2-local","verified":"0","verificationData":""},"role":{"name":"role","value":"","scope":"v2-local","verified":"0","verificationData":""},"headline":{"name":"headline","value":"","scope":"v2-local","verified":"0","verificationData":""},"biography":{"name":"biography","value":"","scope":"v2-local","verified":"0","verificationData":""},"profile_enabled":{"name":"profile_enabled","value":"1","scope":"v2-local","verified":"0","verificationData":""},"additional_mail":[]}
\ No newline at end of file
Index: apps/settings/tests/UserMigration/assets/account-complex.json
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/apps/settings/tests/UserMigration/assets/account-complex.json b/apps/settings/tests/UserMigration/assets/account-complex.json
--- a/apps/settings/tests/UserMigration/assets/account-complex.json	(revision 7eaa573bac0f82b5b2b837104a508dab37c6e231)
+++ b/apps/settings/tests/UserMigration/assets/account-complex.json	(date 1701548067344)
@@ -1,1 +1,0 @@
-{"displayname":{"name":"displayname","value":"Steve Smith","scope":"v2-local","verified":"0","verificationData":""},"address":{"name":"address","value":"123 Water St","scope":"v2-local","verified":"0","verificationData":""},"website":{"name":"website","value":"https:\/\/example.org","scope":"v2-local","verified":"0","verificationData":""},"email":{"name":"email","value":"steve@example.org","scope":"v2-federated","verified":"0","verificationData":""},"avatar":{"name":"avatar","value":"","scope":"v2-local","verified":"0","verificationData":""},"phone":{"name":"phone","value":"+12178515387","scope":"v2-private","verified":"0","verificationData":""},"twitter":{"name":"twitter","value":"steve","scope":"v2-federated","verified":"0","verificationData":""},"fediverse":{"name":"fediverse","value":"@steve@floss.social","scope":"v2-federated","verified":"0","verificationData":""},"organisation":{"name":"organisation","value":"Mytery Machine","scope":"v2-private","verified":"0","verificationData":""},"role":{"name":"role","value":"Manager","scope":"v2-private","verified":"0","verificationData":""},"headline":{"name":"headline","value":"I am Steve","scope":"v2-local","verified":"0","verificationData":""},"biography":{"name":"biography","value":"Lorem ipsum dolor sit amet, consectetur adipiscing elit. Mauris porttitor ullamcorper dictum. Sed fermentum ut ligula scelerisque semper. Aliquam interdum convallis tellus eu dapibus. Integer in justo sollicitudin, hendrerit ligula sit amet, blandit sem.\n\nSuspendisse consectetur ultrices accumsan. Quisque sagittis bibendum lectus ut placerat. Mauris tincidunt ornare neque, et pulvinar tortor porttitor eu.","scope":"v2-local","verified":"0","verificationData":""},"profile_enabled":{"name":"profile_enabled","value":"1","scope":"v2-local","verified":"0","verificationData":""},"additional_mail":[{"name":"additional_mail","value":"steve@example.com","scope":"v2-published","verified":"0","verificationData":""},{"name":"additional_mail","value":"steve@earth.world","scope":"v2-local","verified":"0","verificationData":""}]}
\ No newline at end of file

```

kesselb:
> Side note: I thought the `biography` field would be converted to the carddav `notes`, but see now it is not and has no correlated field for the contact card. @kesselb would it be okay if I add this to this PR, or is this kept separate for a reason?

Good catch, sounds good to me. 

@miaulalala @ChristophWurst mapping the about field from our profile, which should be populated from the ldap biography field if available, to the notes field in vcard. Wdyt?

ChristophWurst:
I think we should map LDAP attributes to profile properties, and profile properties to the system contact. Then users can set the scope of the birthday and anniversary fields.

Without a scope setting we either
1) Have to make the property private, which means nobody will see it so it's useless.
2) Have to make the property local, which means all other users can see your birthday, even if you don't want to expose this.

slapcat:
I added back the scoping as suggested to the code I already wrote. I'm not as familiar with the frontend and would find it difficult to add these features correctly to the profile page. Can someone assist with that portion? Or can we split that into a separate PR?

st3iny:
Hey @slapcat, a colleague of mine and me hacked together a birthday field and added it to the profile a while ago. Perhaps the frontend code could be reused here to allow configuring the birthday and set the scope.

There is also working backend code for saving a user's birthday property.

The frontend is mostly in:
https://github.com/nextcloud/server/blob/feat/account/add-birthday-prop/apps/settings/src/components/PersonalInfo/BirthdaySection.vue

The full PR is at: https://github.com/nextcloud/server/pull/39592

st3iny:
The alternative PR including a documentation PR are merged.

Ref https://github.com/nextcloud/server/pull/45512
Ref https://github.com/nextcloud/documentation/pull/11323

slapcat:
Thanks @st3iny !

PR #41808, opened 2023-11-28
accessibility fix - changed the dir attribute to have valid value

### LINKED ISSUE #41780: [BITV] Files list breadcrumbs contain not valid attributes
The breadcrumbs of the file list contain a `dir` attribute on the `li` element.
But the `dir` attribute is meant for text direction and might only have the values `ltr` `rtl` or `auto`.

This is an accessibility issue as the attribute probably should be either a `data-dir` attribute or is a vue property not correctly set.

emoral435:
* Resolves #41780

## Summary
As the dir attribute works with ltr, rtl, and auto values, the current values (being the path to the files) was an accessibility issue.
Changed the dir attribute to be auto, as with different language support, this will enable future languages having the right directionality of their language. Kept the path stored as a data- attribute for future support.

## TODO

- [x] Fixed dir accessibility issue

## Images

| Before  | After |
| ------------- | ------------- |
| image|chrome_VWozBXci0c  |

## Checklist


apps/files/src/components/BreadCrumbs.vue
```diff
@@ -6,6 +6,8 @@
 			:aria-label="ariaLabel(section)"
 			:title="ariaLabel(section)"
 			v-bind="section"
+			dir="auto"
+			:data-dir="section.to?.query?.dir"
```
susnux, line 10: Do we somewhere depend on this? Meaning we could maybe get rid of it altogether.
But I think we need the `to` property:

```suggestion
			:data-dir="section.to?.query?.dir"
			:to="section.to"
```

emoral435: just checked, and it does not seem as if there really are any places that use this data attribute - so i'll remove it - it is easy to add otherwise

susnux (approved):

Pytal (approved):

PR #9425, opened 2018-05-08
Check connection status and attempt to reconnect each each minute

icewind1991:
An attempt to prevent "mysql has gone away" errors when there is a long time between database requests

PR #41819, opened 2023-11-29
Regular db pings to reconnect after timed out connections

juliusknorr:
Re-implementation of https://github.com/nextcloud/server/pull/9425 to regularly check if the connection to the db is still active using a dummy select query (which is what the previous implementation used the ping method for which got removed from doctrine).

### DIFF
--- a/lib/private/DB/Connection.php
+++ b/lib/private/DB/Connection.php
@@ -41,6 +41,7 @@ use Doctrine\DBAL\Configuration;
 use Doctrine\DBAL\Connections\PrimaryReadReplicaConnection;
 use Doctrine\DBAL\Driver;
 use Doctrine\DBAL\Exception;
+use Doctrine\DBAL\Exception\ConnectionLost;
 use Doctrine\DBAL\Platforms\MySQLPlatform;
 use Doctrine\DBAL\Platforms\OraclePlatform;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

ChristophWurst (approved):

icewind1991:
We should probably add a check if a transaction is active and don't do automatic re-connections if so since the pending updates will be lost.

icewind1991 (approved):

lib/private/DB/Connection.php
```diff
@@ -127,10 +129,13 @@ public function __construct(
 	public function connect($connectionName = null) {
 		try {
 			if ($this->_conn) {
+				$this->reconnectIfNeeded();
 				/** @psalm-suppress InternalMethod */
 				return parent::connect();
 			}
 
+			$this->lastConnectionCheck[$this->getConnectionName()] = time();
```
artonge, line 137: `$this->getConnectionName()` will return `replica` when the `$this->_conn` is `null`.
So, on initialization, we will always save:

```php
$this->lastConnectionCheck =  [
  'replica' => time(),
],
```

lib/private/DB/Connection.php
```diff
@@ -679,4 +684,26 @@ public function rollBack() {
 		}
 		return $result;
 	}
+
+	private function reconnectIfNeeded(): void {
+		if (
+			!isset($this->lastConnectionCheck[$this->getConnectionName()]) ||
+			$this->lastConnectionCheck[$this->getConnectionName()] + 30 >= time() ||
+			$this->isTransactionActive()
```
artonge, line 692: Regarding this condition: `!isset($this->lastConnectionCheck[$this->getConnectionName()])`

This means that if we never saved the last connection check for a connection, we are skipping the reconnect logic?

Practically, this seems like for single DB server setups, we are always skipping the reconnect logic, as `$this->getConnectionName()` returns `replica` before initialization, and `primarry` afterward.

PR #41844, opened 2023-11-29
fix(TextProcessing): Add task type template param to IManager and registerTPProvider

marcelklehr:
otherwise we cannot use them reasonably

## Checklist

### DIFF
--- a/lib/public/TextProcessing/IProvider.php
+++ b/lib/public/TextProcessing/IProvider.php
@@ -31,7 +31,7 @@ use RuntimeException;
 /**
  * This is the interface that is implemented by apps that
  * implement a text processing provider
- * @template T of ITaskType
+ * @psalm-template-covariant  T of ITaskType
  * @since 27.1.0
  */
 interface IProvider {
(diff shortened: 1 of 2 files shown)


provokateurin:
Where and why is it failing?

marcelklehr:
It is failing when using implementations of IProvider, ie. when registering a provider or when passing a Task object to IManager#runTask or IManager#scheduleTask

provokateurin (approved):

come-nc:
Can you give a minimal example on what did not work which will now work, ideally on psalm.dev?

marcelklehr:
and https://psalm.dev/r/62efc9b86a

come-nc (approved):
Thanks for the details :+1:

marcelklehr:
@come-nc  Do you think we should do the covariant or add new template  to the methods in questions as in my latest comment?

come-nc:
> @come-nc Do you think we should do the covariant or add new template to the methods in questions as in my latest comment?

The covariant thing, because you might have a function that accept both template<B> and template<C> if both inherit A.
Then if you have a function which only accepts template<SubType> you should also reflect it in phpdoc typing but that’s another story.

PR #36001, opened 2023-01-04
Validate user timezone given from login data before saving it

tcitworld:
Don't bother saving invalid data. 

Follow-up to #36000

PR #41866, opened 2023-11-29
fix(dav): Prioritize timezone from core/login

ChristophWurst:
## Summary

The DAV app tries to guess a user's timezone from the calendar timezones and the availability settings. It turns out Nextcloud already tracks a timezone, set when a user logs in. Let's use that as primary source, then fall back to calendar timezones and availability timezone.

## How to test

1) Set your calendar timezones to your home timezone
2) Travel to a different timezone
3) Log in
4) Use code that accesses the timezone server

Master: calendar timezone is used
Here: login timezone is used

In the long run we should probably only *set*, not *update* the timezone at login, and make the timezone configurable via the personal settings.

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/TimezoneService.php
+++ b/apps/dav/lib/CalDAV/TimezoneService.php
@@ -42,6 +42,15 @@ class TimezoneService {
 	}
 
 	public function getUserTimezone(string $userId): ?string {
+		$fromConfig = $this->config->getUserValue(
+			$userId,
+			'core',
+			'timezone',
+		);
+		if ($fromConfig !== '') {
+			return $fromConfig;
+		}
+
 		$availabilityPropPath = 'calendars/' . $userId . '/inbox';
 		$availabilityProp = '{' . Plugin::NS_CALDAV . '}calendar-availability';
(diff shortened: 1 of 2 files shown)


miaulalala (approved):

apps/dav/lib/CalDAV/TimezoneService.php
```diff
@@ -42,6 +42,15 @@ public function __construct(private IConfig $config,
 	}
 
 	public function getUserTimezone(string $userId): ?string {
+		$fromConfig = $this->config->getUserValue(
+			$userId,
+			'core',
+			'timezone',
+		);
+		if ($fromConfig !== '') {
```
tcitworld, line 50: I wanted to say that the value might not be a valid one, but *someone* fixed that a couple of months ago. https://github.com/nextcloud/server/pull/36001 :see_no_evil: 

Older values might still be bad though, as I didn't add a repair step.

tcitworld (approved):
Nice

apps/dav/lib/CalDAV/TimezoneService.php
ChristophWurst: Good enough IMO. The value is set at every login.

PR #41885, opened 2023-11-29
Improve share logic for enforced password & expiry date

nfebe:
* It's possible for the admin to enforce and expiry date after, some shares have been created. This commit makes possible to update the share with the new admin constraints.

* This commit would users to modify enforced expiry to anything within range and less than the enforced limit in the pre-create dialog for public shares.

* This commit fixes, unable to update share without updating password.


AndyScherzinger (approved):

Pytal (approved):

Fenn-CS:
Cypress errors in user management unrelated.

PR #40577, opened 2023-09-22
[27] use multipart copy for s3

icewind1991:
Backport of https://github.com/nextcloud/server/pull/40531

PR #41914, opened 2023-11-29
only do a multipart s3 copy when above the regular copy limit

icewind1991:
Multipart copy seems problematic with some backends so avoid it when possible.

### DIFF
--- a/apps/files_external/tests/Storage/Amazons3MultiPartTest.php
+++ b/apps/files_external/tests/Storage/Amazons3MultiPartTest.php
@@ -1,0 +1,68 @@
+<?php
+/**
+ * @copyright Copyright (c) 2023 Robin Appelman <robin@icewind.nl>
+ *
+ * @author Maxence Lange <maxence@artificial-owl.com>
+ * @author Robin Appelman <robin@icewind.nl>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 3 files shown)


solracsf (commented):
Why don't test if MultiPart is supported? Something like:

```
if ($this->getObjectStore() instanceof IObjectStoreMultiPartUpload) {
   // multipart
} else {
  // not multipart
}
```

apps/files_external/tests/Storage/Amazons3MultiPartTest.php
```diff
@@ -0,0 +1,68 @@
...
+		parent::tearDown();
+	}
+
+	public function testStat() {
+		$this->markTestSkipped('S3 doesn\'t update the parents folder mtime');
+	}
+
+	public function testHashInFileName() {
+		$this->markTestSkipped('Localstack has a bug with hashes in filename');
+	}
```
come-nc, line 67: This adds skipped tests? Why?
Output of phpunit is full of skipped tests and it’s making it hard to see failures. It’s best to use groups and exclude non-relevant tests.

come-nc (approved):

apps/files_external/tests/Storage/Amazons3MultiPartTest.php
icewind1991: These tests overwrite the tests from the parent class

juliusknorr (approved):
Seems good, do you know what s3 providers/solutions enforce such a limit?

juliushaertl:
> Why don't test if MultiPart is supported?

It might need a different interface, I'm not sure if all providers that support multipart **upload** also support multipart **copy**

Ziyann:
For reference, I'm also affected by this issue, using Google Cloud Storage (but haven't tested this patch yet). Some built-in solution would be greatly appreciated, either a configurable limit like `copySizeLimit` here (which we could set to a very high value, thus never triggering multipart copy), or even better, some enable/disable switches for the various operations, because right now I'm manually reverting https://github.com/nextcloud/server/pull/40577 on production after upgrades.
There are many S3 providers with different features supported, so I think it would be great to have as many configuration options as possible (avoiding manual files changes on production).

juliushaertl:
Pushed an additional commit to be able to fully turn off multipart copy for s3 providers not supporting it at all.

PR #41916, opened 2023-11-29
fix(files): add visible label to webdav URL in settings

### LINKED ISSUE #41835: [BITV] 9.1.3.1h/3.1 - Modal dialog "File settings" - The input field for the WebDAV URL does not have a label associated with it programmatically.  (2)
It is recommended to implement a visual label (e.g., "WebDAV Url") using the "label" element, which would be linked to the field using the "for" and "id" attributes.

https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-fa5c8d46f9-v3-n1

9.1.3.1h/11.1 - https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-fa5c8d46f9-v11-n1

ShGKme:
* Resolves: #41835

Before | After
---|---
image | image

## Checklist


susnux (approved):

apps/files/src/views/Settings.vue
```diff
@@ -56,6 +56,7 @@
 		<!-- Webdav URL-->
 		<NcAppSettingsSection id="webdav" :name="t('files', 'WebDAV')">
 			<NcInputField id="webdav-url-input"
```
susnux, line 58: Should it not be NcTextField?

Pytal (approved):
Other than the NcTextField suggestion

apps/files/src/views/Settings.vue
ShGKme: Then we need to allow passing any content to the trailing button slot.

https://github.com/nextcloud-libraries/nextcloud-vue/blob/79e68856175859975dc6b7233eb8d503d37f5a6a/src/components/NcTextField/NcTextField.vue#L141-L145

ShGKme:
Rebased onto master to resolve conflict

marcoambrosini (approved):

apps/files/src/views/Settings.vue
susnux: Yes I also misuse NcInputField for some projects because of that limitation.

PR #41927, opened 2023-11-30
perf: Use more performant way to obtain and check the email as a login name with token login

juliusknorr:
[UserManager::getByEmail](https://github.com/nextcloud/server/blob/bcc1d57a038e8431b63eb9e1e5f76d0616a474eb/lib/private/User/Manager.php#L712-L723) is slow on large userbases so we should avoid it where possible.

- [x] When logging in with a token we first try to verify against the login name passed but then do a second attempt using the email address. As we already know the user id that we need to look up the email for we can use the stored uid from the token and compare the email instead of searching by mail
- [x] Revert change to always use an sql cast that was a behavioral change from https://github.com/nextcloud/server/commit/8bbdd9ea0ce9678720459037e30239dc176b9ed8 between 24 and 25
- [x] ~Add a partial index to the configvalue to ensure that most email checks can use an indexed query~ -> Moved to https://github.com/nextcloud/server/pull/42022
  ```  KEY `tmp_configvalue_index` (`configvalue`(80)),```
  - [x] OCI doesn't like that index `cannot create index on expression with datatype LOB`

TODO:

- [ ] Come up with a good approach for adding a test case to avoid future regressions in performance

### DIFF
--- a/core/Application.php
+++ b/core/Application.php
@@ -44,6 +44,7 @@ use OC\Authentication\Listeners\UserDeletedWebAuthnCleanupListener;
 use OC\Authentication\Notifications\Notifier as AuthenticationNotifier;
 use OC\Core\Listener\BeforeTemplateRenderedListener;
 use OC\Core\Notification\CoreNotifier;
+use OC\SystemConfig;
 use OC\TagManager;
 use OCP\AppFramework\App;
 use OCP\AppFramework\Http\Events\BeforeLoginTemplateRenderedEvent;
@@ -81,6 +82,7 @@ class Application extends App {
 		$notificationManager->registerNotifierService(AuthenticationNotifier::class);
 
 		$eventDispatcher->addListener(AddMissingIndicesEvent::class, function (AddMissingIndicesEvent $event) {
+			$dbType = $this->getContainer()->get(SystemConfig::class)->getSystemValue('dbtype', 'sqlite');
 			$event->addMissingIndex(
 				'share',
 				'share_with_index',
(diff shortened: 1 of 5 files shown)


juliushaertl:
There was 1 error:

1) Test\User\SessionTest::testLogClientInThrottlerEmail
Error: Call to a member function getEMailAddress() on null

/home/runner/work/server/server/lib/private/User/Session.php:463
/home/runner/work/server/server/tests/lib/User/SessionTest.php:1147

--

juliushaertl:
@nickvergessen As discussed via chat, I split out the additional index and moved it to https://github.com/nextcloud/server/pull/42022 to not forget about that.

nfebe (approved):
So far, so good.

_By the way, I think it's helpful to attach the point explanations in the pull request description in the various commit bodies._

nickvergessen (approved):

Fenn-CS:
Automatic backports have failed...

juliushaertl:
/backport a3a343ce413d0f86f70d5b47ab28e40d6dd763aa,3cd1d74a815531c9b7ae25ebbf8e7c45fa566e74 to stable28

juliushaertl:
/backport a3a343ce413d0f86f70d5b47ab28e40d6dd763aa,3cd1d74a815531c9b7ae25ebbf8e7c45fa566e74 to stable27

juliushaertl:
/backport a3a343ce413d0f86f70d5b47ab28e40d6dd763aa,3cd1d74a815531c9b7ae25ebbf8e7c45fa566e74 to stable25

PR #41944, opened 2023-11-30
fix: Properly fetch version from shared files by accessing the owner storage version

juliusknorr:
File versions of shared files are located in the owners root, so we need special handling to obtain the version file.

This fixes version previews for shared files and in combination with https://github.com/nextcloud/richdocuments/pull/3330 version browsing of shared office files

### DIFF
--- a/apps/files_versions/lib/Versions/LegacyVersionsBackend.php
+++ b/apps/files_versions/lib/Versions/LegacyVersionsBackend.php
@@ -27,6 +27,7 @@ declare(strict_types=1);
 namespace OCA\Files_Versions\Versions;
 
 use OC\Files\View;
+use OCA\Files_Sharing\ISharedStorage;
 use OCA\Files_Sharing\SharedStorage;
 use OCA\Files_Versions\Db\VersionEntity;
 use OCA\Files_Versions\Db\VersionsMapper;
@@ -196,6 +197,21 @@ class LegacyVersionsBackend implements IVersionBackend, INameableVersionBackend,
 
 	public function getVersionFile(IUser $user, FileInfo $sourceFile, $revision): File {
 		$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+		$owner = $sourceFile->getOwner();
+		$storage = $sourceFile->getStorage();
+
+		// Shared files have their versions in the owners root folder so we need to obtain them from there
+		if ($storage->instanceOfStorage(ISharedStorage::class) && $owner) {
+			/** @var SharedStorage $storage */
+			$userFolder = $this->rootFolder->getUserFolder($owner->getUID());
+			$user = $owner;
+			$ownerPathInStorage = $sourceFile->getInternalPath();
+			$sourceFile = $storage->getShare()->getNode();
+			if ($sourceFile instanceof Folder) {
+				$sourceFile = $sourceFile->get($ownerPathInStorage);
+			}
+		}
+
 		$versionFolder = $this->getVersionFolder($user);
 		/** @var File $file */
 		$file = $versionFolder->get($userFolder->getRelativePath($sourceFile->getPath()) . '.v' . $revision);


AndyScherzinger (approved):

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
```diff
@@ -196,6 +197,21 @@ public function read(IVersion $version) {
...
 	public function getVersionFile(IUser $user, FileInfo $sourceFile, $revision): File {
 		$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+		$ownerUserId = $sourceFile->getOwner()?->getUID();
+		$storage = $sourceFile->getStorage();
+
+		if ($storage->instanceOfStorage(ISharedStorage::class) && $ownerUserId) {
+			/** @var SharedStorage $storage */
+			// Shared files have their versions in the owners root folder
+			$userFolder = $this->rootFolder->getUserFolder($ownerUserId);
+			$user = $this->userManager->get($ownerUserId);
```
artonge, line 207: Isn't it the same as `$sourceFile->getOwner()`?

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
```diff
@@ -196,6 +197,21 @@ public function read(IVersion $version) {
 
 	public function getVersionFile(IUser $user, FileInfo $sourceFile, $revision): File {
 		$userFolder = $this->rootFolder->getUserFolder($user->getUID());
+		$ownerUserId = $sourceFile->getOwner()?->getUID();
+		$storage = $sourceFile->getStorage();
+
+		if ($storage->instanceOfStorage(ISharedStorage::class) && $ownerUserId) {
+			/** @var SharedStorage $storage */
+			// Shared files have their versions in the owners root folder
+			$userFolder = $this->rootFolder->getUserFolder($ownerUserId);
```
artonge, line 206: Could we simply do that for all files?

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
```diff
@@ -196,6 +197,21 @@ public function read(IVersion $version) {
...
+			$user = $this->userManager->get($ownerUserId);
+			$ownerPathInStorage = $sourceFile->getInternalPath();
+			$sourceFile = $storage->getShare()->getNode();
+			if ($sourceFile instanceof Folder) {
+				$sourceFile = $sourceFile->get($ownerPathInStorage);
+			}
```
artonge, line 212: I don't get this part of the code, why the distinction?

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
juliushaertl: I would rather leave it as it is as in theory the owner could be null/empty e.g. for external storage where where the current user is always taken as the owner, but this might behave differently when running in background jobs for example.

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
juliushaertl: Yep, good catch.

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
juliushaertl: I'm unsure how the existing other code parts handle that, but assuming that the version is always stored and obtained from the owner file system for the LegacyVersionsBackend we could unify that logic, yes.

juliushaertl: Sorry I cannot find of a nicer way to rewrite that with shared logic as with shares the paths are just different. I adapted the owner user change from above, should be a bit cleaner now at least.

apps/files_versions/lib/Versions/LegacyVersionsBackend.php
artonge: Ok, but now I do not understand why we are overriding `$sourceFile`.

If we want to have a `Node` object relative to the owner's folder, would that work?

```suggestion
			$sourceFile = $userStorage->get($sourceFile->getId());
```

juliushaertl: Getting the file by id might be slower here as it would require setting up mountpoints until the file with id is found in the provided user home. Getting by path can take more shortcuts there by only setting up mountpoints in the path that is requested.

artonge (approved):

PR #41947, opened 2023-11-30
optionally hide hidden files if accessed via public share

wronny:
* Resolves: # n/a

## Summary
optionally hide hidden files if accessed via public share
config.php setting: hide_hidden_files_via_public_share_access true|false (default = false)

Nextcloud users are able to show/hide hidden files (starting with dot "."). However if a folder containing hidden files is accessed via a public share hidden files are shown. As far as I can see there is currently no option available to hide hidden files for public share access.

I'm not sure if this is a valid or the best way to implement this. It's just a way I figured out to implement it. By default this change should not have any impact. Only if enabled via config.php setting 'hide_hidden_files_via_public_share_access' => true hidden files are excluded from file list if a folder is accessed via a public share link.

Before (example using a nextcloud maps share)
before

After and enabled via config.php
after_and_enabled

## TODO

- [ ] 

## Checklist

- Code is [properly formatted] (https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -258,6 +258,34 @@ class Cache implements ICache {
 	}
 
 	/**
+	 * get the metadata of all files stored in $folder except hidden files
+	 *
+	 * @param int $fileId the file id of the folder
+	 * @return ICacheEntry[]
+	 */
+	public function getFolderContentsByIdExceptHidden($fileId) {
+		if ($fileId > -1) {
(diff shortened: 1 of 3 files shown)


susnux:
Thank you for your PR, with Nextcloud 32 we completely reworked the public shares UI.
While this is still a nice feature it should rather be implemented on the frontend similar to the user config.

PR #41949, opened 2023-12-01
[stable28] enh(settings): Load from disabled users endpoint

Pytal:
- Fix https://github.com/nextcloud/server/issues/41342

Backport https://github.com/nextcloud/server/pull/40169


AndyScherzinger (approved):

lib/public/User/Backend/IProvideEnabledStateBackend.php
```diff
@@ -52,5 +52,5 @@ public function setUserEnabled(string $uid, bool $enabled, callable $queryDataba
 	 *
 	 * @return string[]
 	 */
-	public function getDisabledUserList(int $offset = 0, ?int $limit = null): array;
+	public function getDisabledUserList(?int $limit = null, int $offset = 0): array;
```
nickvergessen, line 55: Order of arguments changing after 2 RCs? really?

nickvergessen (approved):
🙈 
But didn't check if pagination works properly when you have 100 disabled users and start searching for a user starting with `z`

lib/public/User/Backend/IProvideEnabledStateBackend.php
Pytal: 🙈 seems necessary at least from @come-nc's explanation https://github.com/nextcloud/server/pull/41949/commits/9a6437e56cec5a078425412fd06b485b6a63fde8

artonge (approved):

PR #41950, opened 2023-12-01
fix(files): Do not use aria-label on wrong element but use hidden label

### LINKED ISSUE #41894: [BITV] 9.4.1.2/3.6 - All checkboxes in the first column of the table use the "aria-label" attribute, which has been incorrectly applied in this case on the "span" element, which is not allowed. (2)
The "aria-label" attribute should be used together with the "input" element or, alternatively, an existing "label" element should be used instead of ARIA. 

image

https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-bfbf07e5c3-v3-n6

9.4.1.2/11.6 - https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-bfbf07e5c3-v11-n6

susnux:
* Resolves: #41894

## Summary

The aria labels were put on the wrapper of the checkbox instead of the input, instead use a visually hidden label.

## Checklist


Pytal (approved):

ShGKme (commented):
Shouldn't we fix it on `NcCheckboxRadioSwitch` side allowing to create a switch without a text with an aria-label?

ShGKme:
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/4903

PR #41952, opened 2023-12-01
fix(files): Fix type issues of mixing string and number

susnux:
## Summary

Fix some type issues in the files list where numbers were mixed with strings for fileids and thus the strict compare will fail (`===`).

## Checklist

### DIFF
--- a/apps/files/src/components/FileEntry.vue
+++ b/apps/files/src/components/FileEntry.vue
@@ -104,13 +104,13 @@
 import type { PropType } from 'vue'
 
 import { extname, join } from 'path'
-import { FileType, formatFileSize, Permission, Folder, File as NcFile, NodeStatus, Node, View } from '@nextcloud/files'
+import { FileType, formatFileSize, Permission, Folder, File as NcFile, NodeStatus, Node as NcNode, View, Navigation } from '@nextcloud/files'
 import { getUploader } from '@nextcloud/upload'
 import { showError } from '@nextcloud/dialogs'
 import { translate as t } from '@nextcloud/l10n'
 import { vOnClickOutside } from '@vueuse/components'
 import moment from '@nextcloud/moment'
-import Vue from 'vue'
+import Vue, { defineComponent } from 'vue'
 
 import { action as sidebarAction } from '../actions/sidebarAction.ts'
 import { getDragAndDropPreview } from '../utils/dragUtils.ts'
@@ -132,7 +132,7 @@ import logger from '../logger.js'
 
 Vue.directive('onClickOutside', vOnClickOutside)
 
-export default Vue.extend({
+export default defineComponent({
 	name: 'FileEntry',
 
 	components: {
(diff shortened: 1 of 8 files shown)


apps/files/src/components/FileEntry.vue
```diff
@@ -194,8 +194,8 @@ export default Vue.extend({
 	},
 
 	computed: {
-		currentView(): View {
```
artonge, line 197: Why remove the typing?

apps/files/src/components/FileEntry/FileEntryCheckbox.vue
```diff
@@ -103,8 +110,9 @@ export default Vue.extend({
 
 				const lastSelection = this.selectionStore.lastSelection
 				const filesToSelect = this.nodes
-					.map(file => file.fileid?.toString?.())
+					.map(file => file.fileid)
 					.slice(start, end + 1)
+					.filter(isNumber)
```
artonge, line 115: Is it in case the file ID is a string? If so, then this is just hiding the issue, no?

apps/files/src/components/FilesListTableHeaderActions.vue
```diff
@@ -115,13 +119,17 @@ export default Vue.extend({
 		},
 
 		nodes() {
+			// eslint-disable-next-line jsdoc/require-jsdoc
```
artonge, line 122: Is there a way to disable requiring jsdoc for ts files?

apps/files/src/views/FilesList.vue
```diff
@@ -486,7 +492,9 @@ export default defineComponent({
 			if (window?.OCA?.Files?.Sidebar?.setActiveTab) {
 				window.OCA.Files.Sidebar.setActiveTab('sharing')
 			}
-			sidebarAction.exec(this.currentFolder, this.currentView, this.currentFolder.path)
+			if (this.currentFolder !== undefined) {
```
skjnldsv, line 495: Seems like there is a lot of unrelated changes?
How would `this.currentFolder` be undefined?

The FilesList gets rendered only if we have files, which we have only if `this.currentFolder` is defined, no? :thinking:

apps/files/src/components/FileEntry.vue
```diff
@@ -431,14 +435,16 @@ export default Vue.extend({
 
 			// Dragging set of files, if we're dragging a file
 			// that is already selected, we use the entire selection
-			if (this.selectedFiles.includes(this.fileid)) {
-				this.draggingStore.set(this.selectedFiles)
-			} else {
-				this.draggingStore.set([this.fileid])
+			if (this.fileid !== undefined) {
```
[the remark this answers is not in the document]
susnux: In case `this.source` is undefined / has no fileid, see definition:

```ts
fileid() {
	return this.source?.fileid // this can also be undefined
}
```

apps/files/src/components/FileEntry/FileEntryCheckbox.vue
susnux: This is for the case fileid is undefined as the typing of `file.fileid` is `number|undefined`

apps/files/src/components/FileEntry.vue
susnux: Because it was wrong, the correct type would be `View | null` as this is the type of `Navigation.active`

apps/files/src/components/FileEntry.vue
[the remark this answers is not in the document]
susnux: Makes sense! Changed it and added a comment about the assertion

skjnldsv:
Too much conflicts, closing :see_no_evil:

PR #41962, opened 2023-12-01
fix(dav): Make current ooo info time-dependent

### LINKED ISSUE #11059: Only show out of office message when applicable
### How to use GitHub

---

## Steps to reproduce
1. Set up OOO for upcoming (not current!) vacation 24th Dec to 8th Jan
2. Have DND user status (availability or manual)

### Expected behaviour
Nothing happens

### Actual behaviour
Bildschirmfoto vom 2023-11-30 22-34-39

THe OOO api endpoint returns start and end time. We should check it before displaying.

## Talk app

**Talk app version:** main

ChristophWurst:
* Helps resolve https://github.com/nextcloud/spreed/issues/11059

## Summary

* If there is an out of office absence info and it happens now -> return data
* Else: return no data

## Checklist

### DIFF
--- a/apps/dav/appinfo/routes.php
+++ b/apps/dav/appinfo/routes.php
@@ -33,7 +33,8 @@ return [
 	],
 	'ocs' => [
 		['name' => 'direct#getUrl', 'url' => '/api/v1/direct', 'verb' => 'POST'],
-		['name' => 'out_of_office#getCurrentOutOfOfficeData', 'url' => '/api/v1/outOfOffice/{userId}', 'verb' => 'GET'],
+		['name' => 'out_of_office#getCurrentOutOfOfficeData', 'url' => '/api/v1/outOfOffice/{userId}/now', 'verb' => 'GET'],
+		['name' => 'out_of_office#getOutOfOffice', 'url' => '/api/v1/outOfOffice/{userId}', 'verb' => 'GET'],
 		['name' => 'out_of_office#setOutOfOffice', 'url' => '/api/v1/outOfOffice/{userId}', 'verb' => 'POST'],
 		['name' => 'out_of_office#clearOutOfOffice', 'url' => '/api/v1/outOfOffice/{userId}', 'verb' => 'DELETE'],
 	],
--- a/apps/dav/lib/Controller/OutOfOfficeController.php
+++ b/apps/dav/lib/Controller/OutOfOfficeController.php
@@ -27,7 +27,6 @@ declare(strict_types=1);
 namespace OCA\DAV\Controller;
 
 use DateTimeImmutable;
-use OCA\DAV\Db\AbsenceMapper;
 use OCA\DAV\ResponseDefinitions;
 use OCA\DAV\Service\AbsenceService;
 use OCP\AppFramework\Db\DoesNotExistException;
@@ -36,18 +35,20 @@ use OCP\AppFramework\Http\Attribute\NoAdminRequired;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCSController;
 use OCP\IRequest;
+use OCP\IUserManager;
 use OCP\IUserSession;
 use OCP\User\IAvailabilityCoordinator;
 
 /**
  * @psalm-import-type DAVOutOfOfficeData from ResponseDefinitions
+ * @psalm-import-type DAVCurrentOutOfOfficeData from ResponseDefinitions
  */
 class OutOfOfficeController extends OCSController {
 
 	public function __construct(
 		string $appName,
 		IRequest $request,
-		private AbsenceMapper $absenceMapper,
+		private IUserManager $userManager,
 		private ?IUserSession $userSession,
 		private AbsenceService $absenceService,
 		private IAvailabilityCoordinator $coordinator,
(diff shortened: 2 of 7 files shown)


nickvergessen (changes requested):
This will break the option for clients or potential admin features in the feature to show OOO in the form, etc.

Please let's not do this 🙈 
We should check the applicable time in talk, it's on the way.

apps/dav/lib/Controller/OutOfOfficeController.php
```diff
@@ -66,8 +66,12 @@ public function __construct(
 	 */
 	#[NoAdminRequired]
 	public function getCurrentOutOfOfficeData(string $userId): DataResponse {
```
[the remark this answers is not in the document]
nickvergessen: I'm fine with that change, but we should always return the data on the API so it can be used for multiple use cases (like settings + showing in chat, etc).

ChristophWurst:
OOF. But the endpoint was specifically designed to only return data if it's currently active.

Do we need two endpoints?

Antreesy (approved):
Tested, works as expected, returning 404, if OoO is not in the effect yet (or expired)

Also wanted to ask about second endpoint. I understand that it might be useful for clients to get that data beforehand, but that's not the general case, I'd say

ChristophWurst:
> I would say no, a simple date check on the returned date easily tells if it's applicable or not?

No, it's not, because you don't know the timezone of the other user.

Pushed another version. The old endpoint to get the absence data independent of time is back. In addition there is a timezone-aware endpoint that only returns data if the absence happens right now.

apps/dav/appinfo/routes.php
```diff
@@ -33,7 +33,8 @@
 	],
 	'ocs' => [
 		['name' => 'direct#getUrl', 'url' => '/api/v1/direct', 'verb' => 'POST'],
-		['name' => 'out_of_office#getCurrentOutOfOfficeData', 'url' => '/api/v1/outOfOffice/{userId}', 'verb' => 'GET'],
+		['name' => 'out_of_office#getCurrentOutOfOfficeData', 'url' => '/api/v1/outOfOffice/{userId}/now', 'verb' => 'GET'],
+		['name' => 'out_of_office#getOutOfOffice', 'url' => '/api/v1/outOfOffice/{userId}', 'verb' => 'GET'],
```
nickvergessen, line 37: requires openapi.json update `bash build/openapi-checker.sh` (or something alike)

ChristophWurst:
API changed another time for the new return type. Please let me know if it's usable.

provokateurin (approved):
OpenAPI looks good

lib/public/User/IOutOfOfficeData.php
```diff
@@ -74,4 +75,18 @@ public function getShortMessage(): string;
...
+
+	/**
+	 * @return array {
+	 *      id: int,
+	 *      userId: string,
+	 *      startDate: int,
+	 *      endDate: int,
+	 *      shortMessage: string,
+	 *      message: string,
+	 *  }
```
nickvergessen, line 87: This needs to go to ResponseDefinitions.php and then be used as type in the annotation

ChristophWurst: This doesn't feel right. Then the public API is based on private API, no? I'm trying to get the definitions into this interface, then use them from the controller.

apps/dav/lib/Controller/OutOfOfficeController.php
```diff
@@ -59,15 +61,45 @@ public function __construct(
 	 * Get the currently configured out-of-office data of a user.
 	 *
 	 * @param string $userId The user id to get out-of-office data for.
-	 * @return DataResponse<Http::STATUS_OK, DAVOutOfOfficeData, array{}>|DataResponse<Http::STATUS_NOT_FOUND, null, array{}>
+	 * @return DataResponse<Http::STATUS_OK, OutOfOfficeData, array{}>|DataResponse<Http::STATUS_NOT_FOUND, null, array{}>
```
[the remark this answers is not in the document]
nickvergessen: It's reading the ResponseDefinitions.php and not actually eating/evaluating PHP code.

So temporary fix duplicate the data format to ResponseDefinitions.php

lib/public/User/IOutOfOfficeData.php
nickvergessen: > Then the public API is based on private API, no?

Right, should duplicate in that case.

apps/dav/lib/Controller/OutOfOfficeController.php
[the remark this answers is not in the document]
ChristophWurst: oof

apps/dav/lib/Controller/OutOfOfficeController.php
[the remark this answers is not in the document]
ChristophWurst: Included in the latest push

apps/dav/lib/ResponseDefinitions.php
```diff
@@ -35,6 +35,16 @@
  *     status: string,
  *     message: string,
  * }
+ *
+ * @todo this is a copy of \OCP\User\IOutOfOfficeData
+ * @psalm-type DAVOutOfOfficeData2 = array{
```
provokateurin, line 40: I'd prefer if they extend a common type alias. Also, why are there two definitions that seem very similar?

ChristophWurst: Because of https://github.com/nextcloud/server/pull/41962#pullrequestreview-1759433208

ChristophWurst: What is a common type alias?

ChristophWurst: Done

apps/dav/lib/ResponseDefinitions.php
```diff
@@ -27,13 +27,24 @@
...
- *     message: string,
+ * }
+ *
+ * @todo this is a copy of \OCP\User\IOutOfOfficeData
+ * @psalm-type DAVOutOfOfficeData2 = DAVOutOfOfficeDataCommon&array{
+ *     id: string,
+ *     startDate: int,
+ *     endDate: int,
+ *     shortMessage: string,
  * }
```
provokateurin, line 48: I have no idea how the backend works, but is there no way to align these a bit more? Why is the ID an int for one endpoint and a string for another? `status` and `shortMessage` also sound similar, are they different things?

nickvergessen (approved):

apps/dav/lib/ResponseDefinitions.php
```diff
@@ -27,13 +27,24 @@
...
  *     id: int,
- *     userId: string,
  *     firstDay: string,
  *     lastDay: string,
  *     status: string,
- *     message: string,
+ * }
+ *
+ * @todo this is a copy of \OCP\User\IOutOfOfficeData
+ * @psalm-type DAVOutOfOfficeData2 = DAVOutOfOfficeDataCommon&array{
```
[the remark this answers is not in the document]
provokateurin: Ok @nickvergessen already said it as well https://github.com/nextcloud/server/pull/41962#discussion_r1412073831

apps/dav/lib/ResponseDefinitions.php
ChristophWurst: I agree but we are in RC3 and I don't want to cause explosions right now.

provokateurin: maybe a simple cast to int here is ok for now and it can be fixed later.

ChristophWurst: maybe it's ok, maybe it's not. that is the risk I don't want to take at this stage

provokateurin: Ok then keep it :woman_shrugging:

apps/dav/lib/Controller/OutOfOfficeController.php
```diff
@@ -59,15 +61,45 @@ public function __construct(
 	 * Get the currently configured out-of-office data of a user.
 	 *
 	 * @param string $userId The user id to get out-of-office data for.
-	 * @return DataResponse<Http::STATUS_OK, DAVOutOfOfficeData, array{}>|DataResponse<Http::STATUS_NOT_FOUND, null, array{}>
+	 * @return DataResponse<Http::STATUS_OK, DAVCurentOutOfOfficeData, array{}>|DataResponse<Http::STATUS_NOT_FOUND, null, array{}>
```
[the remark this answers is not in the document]
ChristophWurst: it's called *consistency*

nickvergessen:
OpenAPI needs rebuild 🙈

ChristophWurst:
CI is happy now. @nickvergessen are you, too?

nickvergessen (approved):

PR #41972, opened 2023-12-01
fix(files): move actions for selected file out from table header

### LINKED ISSUE #41834: [BITV] 9.1.3.1e/3.2 - When a file is selected, the table changes - the column headers at the top disappear and are replaced by buttons. (2)
You get a table with 5 columns but only two column headers. When a data cell in column 2 is focused with the screen reader, the names of all buttons like "Remove from favorites" are read out together as column header. It is recommended to place the buttons outside the table and keep the correct column headers instead of replacing them

image

https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-3f0ff0bc0e-v3-n2

9.1.3.1e/11.2 - https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-3f0ff0bc0e-v11-n2

ShGKme:
* Resolves: #41834 

## Summary

Rendering actions button inside the table column header still counts as the table header. As the result:
- actions buttons are used as column name which is not correct
- other columns has no name

In this PR actions are moved out from the table and table header is always rendered.

## TODO

- [x] `FilesListTableHeaderActions` is now a `<div>`, not a table header cell `<th>`
- [x] `VirtualList` has a new slot `#header-overlay` to display content with absolute position above the header
  - It has the same styles as a table row and full width
- [x] `FilesListVirtual` passes `<FilesListTableHeaderActions>` to the new `#header-overlay` slot
- [x] `FilesListTableHeader` doesn't render `<FilesListTabelHeaderActions>` and always render table header (
  - Table header should always be rendered, otherwise screen reader looses column names
  - When **actions** are shown, they overlays the header, and buttons are not visible
- [x] Remove styles override, unneeded after https://github.com/nextcloud/nextcloud-vue/pull/3936

## Screen Reader NVDA

Column | Before | After
---|---|---
**Name** | Add to favorite Move or copy Download Actions column2 clickable link Templates | Name column 2 clickable link Templates
**Size** | column 4 clickable 0 KB | Size column 4 clickable 0 KB

## Screenshots - no visual changes

Before | After
---|---
image | image 
image | image 
image | image

## Checklist


apps/files/src/components/FilesListTableHeaderActions.vue
```diff
@@ -220,7 +220,7 @@ export default Vue.extend({
 	flex: 1 1 100% !important;
 
 	// Remove when https://github.com/nextcloud/nextcloud-vue/pull/3936 is merged
-	::v-deep .button-vue__wrapper {
+	:deep(.button-vue__wrapper) {
```
Pytal, line 223: Can be removed?

Pytal (approved):

apps/files/src/components/FilesListTableHeaderActions.vue
ShGKme: Done

skjnldsv (approved):
Great work!!

ShGKme:
Rebased onto `master`

PR #41979, opened 2023-12-02
fix(files_sharing): fix sharing status indicator side after component update

skjnldsv:
| Before | After |
|:---------:|:------:|
|||


szaimen (approved):
Fine by me but wondering why this is needed

nfebe (approved):

skjnldsv:
> Fine by me but wondering why this is needed

It is in the commit comments :)

> Seems like the components now have max-height/width set too

skjnldsv:
`files/files_copy-move.cy.ts` also failing on master

skjnldsv:
Backport in https://github.com/nextcloud/server/pull/42735

PR #41995, opened 2023-12-04
fix(TemplateManager): Bail early if templatedirectory is empty

### LINKED ISSUE #39266: Possibility to disable "Templates" directory creation in the config file
If a custom skeleton is used in the config file,

`'skeletondirectory' => '/home/nextcloud/skeleton',`

then an empty "Templates" folder is created for every new user.
This behavior is (somewhat) described in the documentation, see [#39227](https://github.com/nextcloud/server/issues/39227#issuecomment-1628630194)

I found counter-intuitive to create an empty folder, non-wanted by the administrators. I receive constant tickets from my users "For what we should use this "Templates" folder? Can I delete it?". They don't understand why they have this empty folder.

From my point of view, if we have in the config file:
`'templatedirectory' => '',` or similar,
then no "Templates" folder should be created.

Thank you!

marcelklehr:
* Resolves: #39266

## Checklist

### DIFF
--- a/lib/private/Files/Template/TemplateManager.php
+++ b/lib/private/Files/Template/TemplateManager.php
@@ -275,6 +275,11 @@ class TemplateManager implements ITemplateManager {
 		$isDefaultTemplates = $skeletonTemplatePath === $defaultTemplateDirectory;
 		$userLang = $this->l10nFactory->getUserLanguage($this->userManager->get($this->userId));
 
+		if ($skeletonTemplatePath === '') {
+			$this->setTemplatePath('');
+			return '';
+		}
+
 		try {
 			$l10n = $this->l10nFactory->get('lib', $userLang);
 			$userFolder = $this->rootFolder->getUserFolder($this->userId);


juliusknorr (approved):

lib/private/Files/Template/TemplateManager.php
```diff
@@ -275,6 +275,11 @@ public function initializeTemplateDirectory(string $path = null, string $userId
 		$isDefaultTemplates = $skeletonTemplatePath === $defaultTemplateDirectory;
 		$userLang = $this->l10nFactory->getUserLanguage($this->userManager->get($this->userId));
 
+		if (empty($skeletonTemplatePath)) {
```
nickvergessen, line 278:
```suggestion
		if ($skeletonTemplatePath === '') {
```

Since you use `getSystemValueString` it only returns string, then you can explicitly check for the empty string

ChristophWurst (approved):
Please squash your commits into one

nickvergessen (approved):

PR #42008, opened 2023-12-04
fix(metadata): Allow to load metadata of multiple files at once

nickvergessen:
* Would be helpful for Talk to avoid a query per file: https://github.com/nextcloud/spreed/pull/11093
* Similarly other apps like photos could benefit from it, but instead it currently uses exposure/manipulation via the query builder: https://github.com/nextcloud/photos/blob/aa82646be1493e74a00513757153e1759a8e3a69/lib/DB/Place/PlaceMapper.php#L59-L67

## Checklist

### DIFF
--- a/lib/private/FilesMetadata/FilesMetadataManager.php
+++ b/lib/private/FilesMetadata/FilesMetadataManager.php
@@ -148,6 +148,19 @@ class FilesMetadataManager implements IFilesMetadataManager {
 	}
 
 	/**
+	 * returns metadata of multiple file ids
+	 *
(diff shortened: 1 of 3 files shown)


bigcat88 (approved):

ArtificialOwl (approved):

tobiasKaminsky:
Is there any documentation for it? 
I think this might be also useful for Files clients.

nickvergessen:
I did not add an DAV/API endpoint.
That could/should be done by the initial feature developer if it's not there yet

PR #42022, opened 2023-12-04
perf: Add partial index on configvalue of preferences table

juliusknorr:
To be reconsidered, there are some instances using such an index, but having diverging schema between different db backends is probably something to avoid.

Other ideas:
- Add a hash column to access by value with an index (might also need a hash for lower cased form)


juliushaertl:
Filed as https://github.com/nextcloud/server/issues/42505 for tracking further

PR #42025, opened 2023-12-05
fix(webdav): multiple headers check

datenangebot:
## Summary
Some customers had the issue that public shares can't be opened. They result in a reload loop. This is caused by an unauthenticated dav request. I turns out that the `XMLHttpRequest` could be set twice and the logic checks for exact one match.

We provided this patch as a quick fix and maybe it's a good idea to bring this to the server to make it more robust besides fixing the main cause.

Customers were on nextcloud 24.0.12.9.

## TODO

- [ ] check if useful
- [ ] add backports

## Checklist


artonge (approved):

skjnldsv:
Fixed already

PR #42030, opened 2023-12-05
Replace input filed with password field and added password error message

### LINKED ISSUE #41882: [BITV] 9.3.3.1/3.1 - Section "Details" / tab "Sharing": When the field with the password is left empty, no error message is displayed, which is necessary. The field only receives a red border, and an error icon appears. (2)
image

https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-c3b56e67ab-v3-n1

9.3.3.1/11.1 - https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-c3b56e67ab-v11-n1

JuliaKirschenheuter:
* Resolves: https://github.com/nextcloud/server/issues/41882

## Summary

* Replace input filed with password field and added password error message
* Remove unneeded NcDateTimePicker

Before:

Screenshot from 2023-12-05 11-12-07

After:

Peek 2023-12-05 10-47

## Checklist


apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -105,10 +105,11 @@
 						<NcCheckboxRadioSwitch :checked.sync="isPasswordProtected" :disabled="isPasswordEnforced">
 							{{ t('files_sharing', 'Set password') }}
 						</NcCheckboxRadioSwitch>
-						<NcInputField v-if="isPasswordProtected"
+						<NcPasswordField v-if="isPasswordProtected"
 							:type="hasUnsavedPassword ? 'text' : 'password'"
 							:value="hasUnsavedPassword ? share.newPassword : '***************'"
 							:error="passwordError"
```
ShGKme, line 111: `type` doesn't work on `NcPasswordField`. It is always `password`.

We also should not hide password with `***************` in `value` here.

ShGKme: I assume, `NcInputField` was used here because the generated password should have been initially visible before saving. Now the user always needs to click on "Show password" first to see or copy the generated password.

susnux (approved):

apps/files_sharing/src/views/SharingDetailsTab.vue
Pytal: > We also should not hide password with `***************` in `value` here.

I believe this is intentional @jancborchardt @szaimen?

ShGKme: > @ShGKme have i resolved a problem?

Not completely.

This field is not a truly "Password field" in the current usage.

It has 3 jobs:
1. On add password: show the generated password as a plain text to copy
2. After p1: to enter another password to be saved
3. On edit: to change the password without showing it.

And here we have 2 issues:
1. With `NcPasswordField` a user always needs one more click on "Show password" to see and copy or edit the password. Always +1 click. Binding `:type` doesn't work on `NcPasswordField`. 
2. On edit you are editing the current password. So you need to edit this `***************`. And it is pronounced as `"fifteen stars"`. I guess in some screen-readers it may even be pronounced as `"star star star star star star star star star star star star star star star"`

---

But I need to think about it to propose a solution...

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -105,10 +105,10 @@
 						<NcCheckboxRadioSwitch :checked.sync="isPasswordProtected" :disabled="isPasswordEnforced">
 							{{ t('files_sharing', 'Set password') }}
 						</NcCheckboxRadioSwitch>
-						<NcInputField v-if="isPasswordProtected"
-							:type="hasUnsavedPassword ? 'text' : 'password'"
-							:value="hasUnsavedPassword ? share.newPassword : '***************'"
+						<NcPasswordField v-if="isPasswordProtected"
+							:value="hasUnsavedPassword ? share.newPassword : ''"
```
ShGKme, line 109: Same as
```suggestion
							:value="share.newPassword"
```

ShGKme (approved):
Fixes the issue, but has 2 drawbacks:
- Now a user always needs +1 more click to see the generated password
- On edit of saved password, this might not be clear that an empty password field is a new password field and there is a password.

@nextcloud/designers What do you think?

jancborchardt:
> Now a user always needs +1 more click to see the generated password

This I’d say is less of an issue

> On edit of saved password, this might not be clear that an empty password field is a new password field and there is a password.

@ShGKme I’m not sure I fully understand, especially the part "[…] is a new password field and there is a password" – it looks fine from your posted video above, but it only shows one flow.

ShGKme:
> This I’d say is less of an issue

Just to be clear, a user always needs it because checking "set password" generates a random password.

password

> @ShGKme I’m not sure I fully understand, especially the part "[…] is a new password field and there is a password" – it looks fine from your posted video above, but it only shows one flow.

When you open an existing share with a password, you see:

image

I'm not sure that it is clear that it is "enter a new password to change it" and not "you have set a password and it is an empty password".

jancborchardt:
@ShGKme thanks for explaining. Well both of those cases are confusing, the second one even more than the first. In both cases, ideally the password should be shown in plain text from a UX point of view.

If that is a security concern, instead of the plain input field we could instead show feedback text like:
- New: Password generated [ Copy ] [ Edit ]
- Existing: Password set [ Copy ] [ Edit ]

That way it would also be a 2-step process, but way clearer what it is about, and clicking on "Edit" should also directly show the password in plain text.

ShGKme:
> In both cases, ideally the password should be shown in plain text from a UX point of view.

We can show the password only when a user checks "Set password" and a new password is generated.

Here the only issue could be that "Show password" is `false` by default in `NcPasswordPicker`.

But for security concerns, we don't actually store the password. So when users edit an existing entry, they can set a new password, but can neither see nor edit the previously entered password.

jancborchardt (approved):
Error text is fine, so not blocked from that side @JuliaKirschenheuter :)

Fenn-CS:
> We can do all improvements after certification 🙏. Do you have any considerations regarding this error text? If not, please approve this PR.

@JuliaKirschenheuter  please, you should track a new issue for the improvements, so we ensure it gets done. Thank you...

Issue #41895, opened 2023-11-29
[BITV] 9.4.1.2/3.7 - In the breadcrumb navigation, there are "aria-label" attributes on "li" elements, which is not allowed, such as aria-label="Go to the "/" directory". (2)


JuliaKirschenheuter:
https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-bfbf07e5c3-v3-n7

9.4.1.2/11.7 - https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-bfbf07e5c3-v11-n7

JuliaKirschenheuter:
@szaimen not resolved?

szaimen:
> @szaimen not resolved?

I may have moved it by mistake...

szaimen:
We have this here now:
![image](https://github.com/nextcloud/server/assets/42591237/7d97dc54-45e2-4d63-8eed-62eddbddc812)
Is this the correct implementation @susnux @ShGKme @emoral435 @Pytal @JuliaKirschenheuter ?

JuliaKirschenheuter:
semantically it is correct but `aria-label` seems to be not translated. @emoral435 what was the last stand 🤔?

emoral435:
<img width="609" alt="image" src="https://github.com/nextcloud/server/assets/110193237/471f3eee-34e0-495d-aee1-651be4b2caef">

This is what we came up as our decision (copied and pasted it from my messages with skjnldsv - won't tag him because I do not want to bog him down with another ping, LOL).

So as far as implementation, it is correct, however, upon testing right now, it seems as if the aria label for the home section seems to be untranslated, probably due to it being not a component that we made. It is a component supplied by Vue itself. Will be correcting this!

szaimen:
Seems like aria-label="Home" is still not translated?
![image](https://github.com/nextcloud/server/assets/42591237/86cade6b-c28a-4ebb-9939-f3e1ced73641)

ShGKme:
> Seems like aria-label="Home" is still not translated?

Is it informal or formal German?

szaimen:
formal German

emoral435:
@ShGKme Would you be able to handle the translations here? I unfortunately do not know formal German. However, I also would not mind doing it myself, I am just now familiar with how our translation system works. A pointer in the right direction would help me lots :)

ShGKme:
[...]

PR #42035, opened 2023-12-05
Remove unneeded characters form `aria-label` and `title` on BreadCrumbs

JuliaKirschenheuter:
* Part of https://github.com/nextcloud/server/issues/41895

## Summary

Remove unneeded characters like `&quot` form `aria-label` and `title` on BreadCrumbs

🏚️ Before | 🏡 After
---|---
Screenshot from 2023-12-05 14-46-13 | Screenshot from 2023-12-05 14-43-17

## Checklist


susnux (approved):

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -110,7 +110,7 @@ export default Vue.extend({
 			if (section?.to?.query?.dir === this.$route.query.dir) {
 				return t('files', 'Reload current directory')
 			}
-			return t('files', 'Go to the "{dir}" directory', section)
+			return t('files', 'Go to the {dir} directory', section)
```
ShGKme, line 113: During the feature and string freeze, we should not create strings for translators.

ShGKme (changes requested):
It seems to me that the issue was only about `aria-label` on the `<li>` element. I check and screen-readers work fine with the escaped HTML quote in the attribute value.

JuliaKirschenheuter:
not relevant, closing PR

PR #42039, opened 2023-12-05
Add UserConfigChangedEvent to fire whenever a user config value is changed

### LINKED ISSUE #41938: Can we fire UserChangedEvent when setUserValue method is used?
### How to use GitHub

- I have some apps in nextcloud that set user values using [`setUserValue`](https://github.com/nextcloud/server/blob/ef3b1a86ebac1b116740a815d2b8ebace53b3980/lib/private/AllConfig.php#L251)
- I would like to develop a new app that can check when certain such values are updated and develop some functionality on top of this

- Fire the `UserChangedEvent` when [`setUserValue`](https://github.com/nextcloud/server/blob/ef3b1a86ebac1b116740a815d2b8ebace53b3980/lib/private/AllConfig.php#L251) is called 
- Perhaps use the [`$user->triggerChange()`](https://github.com/nextcloud/server/blob/ef3b1a86ebac1b116740a815d2b8ebace53b3980/lib/private/User/User.php#L615) to fire this

- Patch the server code to fire this event

**Additional context**
- My understanding from checking the code and by logging locally is that this event is not fired. Please close this issue if it is indeed fired and there is something going wrong in my setup.

akhil1508:
* Closes #41938

### DIFF
--- a/lib/composer/composer/autoload_static.php
+++ b/lib/composer/composer/autoload_static.php
@@ -1022,6 +1022,7 @@ class ComposerStaticInit749170dad3f5e7f9ca158f5a9f04f6a2
         'OCP\\User\\Events\\PasswordUpdatedEvent' => __DIR__ . '/../../..' . '/lib/public/User/Events/PasswordUpdatedEvent.php',
         'OCP\\User\\Events\\PostLoginEvent' => __DIR__ . '/../../..' . '/lib/public/User/Events/PostLoginEvent.php',
         'OCP\\User\\Events\\UserChangedEvent' => __DIR__ . '/../../..' . '/lib/public/User/Events/UserChangedEvent.php',
+        'OCP\\User\\Events\\UserConfigChangedEvent' => __DIR__ . '/../../..' . '/lib/public/User/Events/UserConfigChangedEvent.php',
         'OCP\\User\\Events\\UserCreatedEvent' => __DIR__ . '/../../..' . '/lib/public/User/Events/UserCreatedEvent.php',
         'OCP\\User\\Events\\UserDeletedEvent' => __DIR__ . '/../../..' . '/lib/public/User/Events/UserDeletedEvent.php',
         'OCP\\User\\Events\\UserFirstTimeLoggedInEvent' => __DIR__ . '/../../..' . '/lib/public/User/Events/UserFirstTimeLoggedInEvent.php',
--- a/lib/private/Config/UserConfig.php
+++ b/lib/private/Config/UserConfig.php
@@ -23,9 +23,11 @@ use OCP\Config\ValueType;
 use OCP\DB\Exception as DBException;
 use OCP\DB\IResult;
 use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\EventDispatcher\IEventDispatcher;
 use OCP\IConfig;
 use OCP\IDBConnection;
 use OCP\Security\ICrypto;
+use OCP\User\Events\UserConfigChangedEvent;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -75,6 +77,7 @@ class UserConfig implements IUserConfig {
 		private readonly PresetManager $presetManager,
 		protected LoggerInterface $logger,
 		protected ICrypto $crypto,
+		protected IEventDispatcher $dispatcher,
 	) {
 	}
 
@@ -1123,12 +1126,14 @@ class UserConfig implements IUserConfig {
 			}
 		}
 
+		$oldValue = null;
 		if ($this->hasKey($userId, $app, $key, $lazy)) {
 			/**
 			 * no update if key is already known with set lazy status and value is
 			 * not different, unless sensitivity is switched from false to true.
 			 */
-			if ($origValue === $this->getTypedValue($userId, $app, $key, $value, $lazy, $type)
+			$oldValue = $this->getTypedValue($userId, $app, $key, $value, $lazy, $type);
+			if ($origValue === $oldValue
 				&& (!$sensitive || $this->isSensitive($userId, $app, $key, $lazy))) {
 				return false;
 			}
@@ -1209,6 +1214,8 @@ class UserConfig implements IUserConfig {
 
 			$update->executeStatement();
 		}
+
+		$this->dispatcher->dispatchTyped(new UserConfigChangedEvent($userId, $app, $key, $value, $oldValue));
 
 		if ($refreshCache) {
 			$this->clearCache($userId);
--- a/lib/public/User/Events/UserConfigChangedEvent.php
+++ b/lib/public/User/Events/UserConfigChangedEvent.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2025 Murena SAS <akhil.potukuchi.ext@murena.com>
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCP\User\Events;
+
+use OCP\AppFramework\Attribute\Listenable;
+use OCP\EventDispatcher\Event;
+
+/** @since 33.0.0 */
+#[Listenable(since: '33.0.0')]
+class UserConfigChangedEvent extends Event {
+	/**
+	 * @since 33.0.0
+	 */
+	public function __construct(
+		private string $userId,
+		private string $appId,
+		private string $key,
+		private mixed $value,
+		private mixed $oldValue = null,
+	) {
+		parent::__construct();
+	}
+
+	/**
+	 * @since 33.0.0
+	 */
+	public function getUserId(): string {
+		return $this->userId;
+	}
+
+	/**
+	 * @since 33.0.0
+	 */
+	public function getAppId(): string {
+		return $this->appId;
+	}
+
+	/**
+	 * @since 33.0.0
+	 */
+	public function getKey(): string {
+		return $this->key;
+	}
+
+	/**
+	 * @since 33.0.0
+	 */
+	public function getValue(): mixed {
+		return $this->value;
+	}
+
+	/**
+	 * @since 33.0.0
+	 */
+	public function getOldValue(): mixed {
+		return $this->oldValue;
+	}
+}
--- a/lib/composer/composer/autoload_classmap.php
+++ b/lib/composer/composer/autoload_classmap.php
@@ -981,6 +981,7 @@ return array(
     'OCP\\User\\Events\\PasswordUpdatedEvent' => $baseDir . '/lib/public/User/Events/PasswordUpdatedEvent.php',
     'OCP\\User\\Events\\PostLoginEvent' => $baseDir . '/lib/public/User/Events/PostLoginEvent.php',
     'OCP\\User\\Events\\UserChangedEvent' => $baseDir . '/lib/public/User/Events/UserChangedEvent.php',
+    'OCP\\User\\Events\\UserConfigChangedEvent' => $baseDir . '/lib/public/User/Events/UserConfigChangedEvent.php',
     'OCP\\User\\Events\\UserCreatedEvent' => $baseDir . '/lib/public/User/Events/UserCreatedEvent.php',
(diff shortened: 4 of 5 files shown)


lib/private/AllConfig.php
```diff
@@ -304,6 +307,14 @@ public function setUserValue($userId, $appName, $key, $value, $preCondition = nu
 			}
 			$this->userCache[$userId][$appName][$key] = (string)$value;
 		}
+		$this->triggerUserValueChange($userId, $appName, $key, $value, $prevValue);
+	}
+
+	private function triggerUserValueChange($userId, $appId, $key, $value, $oldValue = null) {
+		if (\OC::$server instanceof \OCP\IServerContainer) {
+			$dispatcher = \OC::$server->get(IEventDispatcher::class);
```
akhil1508, line 315: Cannot use DI for this as it is used in `base.php` over [here](https://github.com/nextcloud/server/blob/fd06f0c72de958855d535230b79583b30b9763b1/lib/base.php#L168)

skjnldsv:
Hey @akhil1508 
Thanks for the rebase! Let me ask for reviews, and see how we can help you move thi sforward :+1:

lib/composer/composer/autoload_static.php
```diff
@@ -20,15 +20,15 @@ class ComposerStaticInit749170dad3f5e7f9ca158f5a9f04f6a2
...
         array (
             0 => __DIR__ . '/../../..' . '/core',
         ),
-        'OC\\' => 
+        'OC\\' =>
         array (
             0 => __DIR__ . '/../../..' . '/lib/private',
         ),
-        'OCP\\' => 
+        'OCP\\' =>
```
come-nc, line 31: I guess you opened and save the file with an editor which removed trailing spaces, no? Please make sure the file in git matches the result of composer install --no-dev

lib/public/User/Events/UserConfigChangedEvent.php
```diff
@@ -0,0 +1,98 @@
...
+
+class UserConfigChangedEvent extends Event {
+	private string $userId;
+	private string $appId;
+	private string $key;
+	private mixed $value;
+	private mixed $oldValue;
+
+	/**
+	 * @since 30.0.0
```
come-nc, line 43:
```suggestion
	 * @since 31.0.0
```
You have to up those to 31 now :-/

lib/private/AllConfig.php
come-nc: Is that the right link? The Allconfig constructor does not call setUserValue.

lib/public/User/Events/UserConfigChangedEvent.php
akhil1508: Done

lib/composer/composer/autoload_static.php
akhil1508: > I guess you opened and save the file with an editor which removed trailing spaces, no?

@come-nc Yes :| 

> Please make sure the file in git matches the result of composer install --no-dev

Done. So I ran `composer install --no-dev` and then discarded changes for the other files:

-  `composer/ClassLoader.php`
- `composer/installed.php`
- `composer/autoload_real.php`
- `composer/InstalledVersions.php`
- `autoload.php`

Is that okay?

lib/private/AllConfig.php
akhil1508: @come-nc I believe I meant to say:

1. `AllConfig` is instantiated in `base.php` as `new \OC\AllConfig(new \OC\SystemConfig(self::$config))`
2. If I add `IEventListener $eventListener` in the constructor for DI in this class (the usual way), it could be problematic. I remember it causing errors when I first wrote this code, but maybe  I should re-test? 🤔

come-nc: Yes please, the event listener is available early so I would be surprised if this is called even earlier. Maybe for login time :thinking: 
If some value change call do not trigger the event we need to document that.

skjnldsv:
Still some comments to address :)
@akhil1508 to you need help?

akhil1508:
@skjnldsv Apologies for the delay, I look into finishing up this PR now :)

lib/private/AllConfig.php
akhil1508: @come-nc so if I use DI to set the event dispatcher, I will have to replace https://github.com/nextcloud/server/blob/fd06f0c72de958855d535230b79583b30b9763b1/lib/base.php#L168

with 

```php
new \OC\AllConfig(new \OC\SystemConfig(self::$config), new EventDispatcher(new SymfonyDispatcher(...), ))
```
This doesn't look like the best solution to me :|


I did test `new \OC\AllConfig(new \OC\SystemConfig(self::$config), Server::get(IEventDispatcher::class))` and this gets me the error: 

```
Error: Typed static property OC::$server must not be accessed before initialization in /var/www/html/lib/public/Server.php:55
Stack trace:
#0 /var/www/html/lib/base.php(174): OCP\Server::get('OCP\\EventDispat...')
```

Could you advise? I could set  `$this->dispatcher` in `fixDIInit` just like `$this->connection` but dispatcher is only used in one method

come-nc: My bad, AllConfig really is used really early, before the dispatcher can be injected.
So keep your solution, simple use the non-deprecated version:
```suggestion
			$dispatcher = \OCP\Server::get(IEventDispatcher::class);
```

come-nc (changes requested):
Wait, but actually this needs to got into UserConfig, because AllConfig delegates to UserConfig, and some code path will use directly UserConfig, so if you do not want to miss events you need to dispatch from `OC\Config\UserConfig`.
(that was changed in 31 which is why we missed it when first reviewing this I think)

come-nc:
(And UserConfig supports DI :tada: so the dispatcher can be injected there)

akhil1508:
@come-nc  indeed! I made the required changes to `UserConfig.php` now, let me know if the approach looks okay

lib/public/User/Events/UserConfigChangedEvent.php
```diff
@@ -0,0 +1,98 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\User\Events;
+
+use OCP\EventDispatcher\Event;
+
+/**
+ * @since 31.0.0
```
come-nc:
```suggestion
 * @since 32.0.0
```
I’m sorry about the running gag, but this needs to be bumped to 32 now :see_no_evil: (and all the other `@since` as well).
Also, it would be good to add the new attribute `Listenable`, see https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/api.html#php-public-api

lib/public/User/Events/UserConfigChangedEvent.php
```diff
@@ -0,0 +1,98 @@
...
+		string $key,
+		mixed $value,
+		mixed $oldValue = null) {
+		parent::__construct();
+		$this->userId = $userId;
+		$this->appId = $appId;
+		$this->key = $key;
+		$this->value = $value;
+		$this->oldValue = $oldValue;
+	}
```
come-nc:
```suggestion
	/**
	 * @since 32.0.0
	 */
	public function __construct(
		private string $userId,
		private string $appId,
		private string $key,
		private mixed $value,
		private mixed $oldValue = null,
	) {
		parent::__construct();
	}
```
Constructor property promotion makes it pretty :sparkles:

lib/public/User/Events/UserConfigChangedEvent.php
```diff
@@ -0,0 +1,98 @@
...
+
+	/**
+	 * @return string
```
come-nc:
```suggestion
```
Strong type is enough, same for the others.

akhil1508:
@come-nc thanks for the review, resolved your comments

lib/private/Config/UserConfig.php
```diff
@@ -75,6 +77,7 @@ public function __construct(
 		protected IConfig $config,
 		protected LoggerInterface $logger,
 		protected ICrypto $crypto,
+		protected IEventDispatcher $dispatcher
```
come-nc:
```suggestion
		protected IEventDispatcher $dispatcher,
```
I think `composer run cs:fix` would fix it as well.

lib/public/User/Events/UserConfigChangedEvent.php
```diff
@@ -0,0 +1,81 @@
...
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCP\User\Events;
+
+use OCP\EventDispatcher\Event;
+use OCP\AppFramework\Attribute\Listenable;
+
+#[Listenable(since: '32.0.0')]
```
come-nc: Not sure what the tool will say but you may have to still add the `@since` phpdoc tag.
Ignore this comment if CI is green.

lib/public/User/Events/UserConfigChangedEvent.php
```diff
@@ -0,0 +1,81 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
come-nc, line 8:
```suggestion
/**
 * SPDX-FileCopyrightText: 2025 Murena SAS <akhil.potukuchi.ext@murena.com>
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```
We use SPDX headers now for licensing. Changing to this should turn the REUSE CI green.

lib/public/User/Events/UserConfigChangedEvent.php
come-nc: Yep: Error: lib/public/User/Events/UserConfigChangedEvent.php:32:1: InvalidDocblock: PHPDoc is required for classes/interfaces in OCP. (see https://psalm.dev/008)
```suggestion
/** @since 32.0.0 */
#[Listenable(since: '32.0.0')]
```

come-nc (approved):
Tiny nitpick to pass CI, but code looks good now.
Can you also squash/edit commits to please the conventionnal commits CI?

come-nc:
`composer run cs:fix` is needed for the order of imports as well

akhil1508:
>  Can you also squash/edit commits to please the conventionnal commits CI?

@come-nc Squashed

come-nc:
Tests needs to be adapted

Example error:
```
365) Test\lib\Config\UserConfigTest::testClearCacheAll
ArgumentCountError: Too few arguments to function OC\Config\UserConfig::__construct(), 6 passed in /home/runner/actions-runner/_work/server/server/tests/lib/Config/UserConfigTest.php on line 294 and exactly 7 expected

**Comment by akhil1508** [comment]
> Tests needs to be adapted

@come-nc sorry, missed that. Fixed now

**Review by AndyScherzinger — APPROVED** [review]

File: lib/public/User/Events/UserConfigChangedEvent.php [hunk 10]
```diff
@@ -0,0 +1,65 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCP\User\Events;
+
+use OCP\AppFramework\Attribute\Listenable;
+use OCP\EventDispatcher\Event;
+
+/** @since 32.0.0 */
+#[Listenable(since: '32.0.0')]
```
susnux: Also for the tags below


```suggestion
/** @since 33.0.0 */
#[Listenable(since: '33.0.0')]
```

susnux (approved):
Looks good, but the versions need to be adjusted

akhil1508:
> Looks good, but the versions need to be adjusted

@susnux done, also rebased and squashed commits

susnux (approved):

provokateurin (approved):

susnux:
Needs documentation here:
https://github.com/nextcloud/documentation/blob/1a0415c8bf1541e90ae0b4da487a362a51a1cfe2/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_33.rst?plain=1#L182

Issue #37208, opened 2023-03-14
Design improvements to "Shared" indicator in files list


jancborchardt:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
2 issues with the "Shared" indicator in the files list:
- It is not super present
- The icons are not aligned with the icons of non-shared files

**Describe the solution you'd like**
- The text and icon of the "Shared" indicator could be colorized in `var(--color-primary-element)`
- The icons would look better if to the right of the "Shared" text, so they align with the other icons in the list

**Additional context**
Current:
![image](https://user-images.githubusercontent.com/925062/224966680-db5c41d3-597d-4603-8cfe-c9039ee9b89b.png)

Proposed:
![Sharing colorized](https://user-images.githubusercontent.com/925062/224966705-ca66773f-b6b1-4abe-a7a3-b170873e9db4.png)

gennaios:
I think the share action could be moved to the action menu. I don’t think users are constantly sharing files such that it needs a separate icon. Then the current icon could be shown only for shared files. Such would make it I think much more visible plus there would be more space for longer file names.

skjnldsv:
https://github.com/nextcloud/server/issues/38156

PR #42041, opened 2023-12-05
Revert "Improve shared status button"

skjnldsv:
This reverts commit c2c5994d7069464b9e69a54ce2c660a1a340c711.

Sorry @marcoambrosini, I cannot approve the previous PR:
1. using deep selectors is not acceptable to me
2. you're changing this for all actions, while only the sharing got request
3. a comment was written by a lead with another lead reaction and got completely ignored

I'll have a look with the design team to make sure we keep the previous issue fixed too

### DIFF
--- a/apps/files/src/components/FileEntry/FileEntryActions.vue
+++ b/apps/files/src/components/FileEntry/FileEntryActions.vue
@@ -37,7 +37,6 @@
 			:container="getBoundariesElement"
 			:disabled="isLoading || loading !== ''"
 			:force-name="true"
-			type="tertiary"
 			:force-menu="enabledInlineActions.length === 0 /* forceMenu only if no inline actions */"
 			:inline="enabledInlineActions.length"
 			:open.sync="openedMenu"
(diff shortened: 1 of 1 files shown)


artonge (approved):

Pytal (approved):

susnux (approved):

skjnldsv:
I'll have a look with the design team to make sure we keep the previous issue fixed too

marcoambrosini:
Aside from the execution, admittedly poor and rushed, I just want to register that I did this reluctantly and didn't agree much with the set of changes proposed for this button. Nor with how fast they were rushed in without exploring other solutions.
So I thought of this as a temporary hack, while waiting to have have shared separate shared and owner columns.

skjnldsv:
All good, Marco :)
if the button style must go, then it must go! But without global styling

jancborchardt (changes requested):
> I'll have a look with the design team to make sure we keep the previous issue fixed too

Could we make sure that we don’t revert it without addressing that part here already?
The specification / mockup exists at https://github.com/nextcloud/server/issues/37208, and as long as design-wise it looks like that, you are expert regarding implementation. :)

blizzz:
What about this PR, it appears stale?

PR #42055, opened 2023-12-06
enh(NavigationManager): Use ID as fallback for `app` property of entries

susnux:
## Summary
Some apps like Talk do not use the appinfo for registering the navigation entries but using the NavigationManager directly. Those apps do not always set the app property of the navigation entry, in this cases it is currently not possible to use that entry for the default app.

To fix 99% of the cases the ID is now checked to be a valid and enabled app and if yes then use it as fallback.

## Checklist

### DIFF
--- a/lib/private/NavigationManager.php
+++ b/lib/private/NavigationManager.php
@@ -95,7 +95,10 @@ class NavigationManager implements INavigationManager {
 			return;
 		}
 
+		$id = $entry['id'];
+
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

juliusknorr (approved):

nickvergessen:
❌  no-db tests fail and are related

PR #42065, opened 2023-12-06
fix: Let git ignore composer dev dependencies

susnux:
## Summary

Do not trash the state when running composer install with dev dependencies.

## Checklist

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,6 +11,7 @@
 		}
 	},
 	"autoload": {
+		"exclude-from-classmap": ["**/bamarni/composer-bin-plugin/**"],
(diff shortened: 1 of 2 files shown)


ChristophWurst (approved):

composer.json
```diff
@@ -11,6 +11,7 @@
 		}
 	},
 	"autoload": {
+		"exclude-from-classmap": ["**/bamarni/composer-bin-plugin/**"],
```
nickvergessen, line 14: that's fancy and as the biggest pain so far

nickvergessen (approved):

PR #42094, opened 2023-12-07
Rename "global search" to "unified search"

nfebe:
- Changes appearances of "Global search" to "Unified search" in UI
- Refactors code, to remove usage of term "GlobalSearch" in files and code structure
- Rename old unified search to `legacy-unified-search`


AndyScherzinger (approved):

core/src/views/UnifiedSearchModal.vue
```diff
@@ -1,25 +1,24 @@
...
+	<NcModal id="unified-search"
+		ref="unifiedSearchModal"
+		:name="t('core', 'Unified search')"
 		:show.sync="internalIsVisible"
 		:clear-view-delay="0"
 		:title="t('Unified search')"
 		@close="closeModal">
 		<CustomDateRangeModal :is-open="showDateRangeModal"
-			:class="'global-search__date-range'"
+			:class="'unified-search__date-range'"
```
susnux, line 10:
```suggestion
			class="unified-search__date-range"
```

susnux (changes requested):
Tested and no search is loaded at all because of this error:

```
Uncaught Error: Could not find initial state limit-default of unified-search
    r index.es.mjs:13
    54201 UnifiedSearchService.js:30
    Webpack 5
        o
        c
        O
        <anonymous>
        <anonymous>
index.es.mjs:13:14

```

Fenn-CS:
> Tested and no search is loaded at all because of this error:
> 
> Uncaught Error: Could not find initial state limit-default of unified-search
>     r index.es.mjs:13
>     54201 UnifiedSearchService.js:30
>     Webpack 5
>         o
>         c
>         O
>         <anonymous>
>         <anonymous>
> index.es.mjs:13:14

This change does not contain the build, did you watch or build? I can't see this?

emoral435 (approved):
Tested this out in my dev setup and works, also looked over the code and it looks good to me 👍

emoral435:
> Tested and no search is loaded at all because of this error:

I think it is because of not running npm build or watch, as @Fenn-CS mentioned. After running watch, the app works as expected on my dev setup with no error

Fenn-CS:
@susnux Since, you left a change request on this, I would like you to take a look one more time, thank you!

susnux (approved):

Fenn-CS:
backport to stable28

Issue #41996, opened 2023-12-04
System address book creation errors on concurrent user creation


juliusknorr:
So this sounds like an edge case I just trigger with automatically configuring my dev setup, but still worth to report

Steps to reproduce:

- Install a new Nextcloud instance through occ
- call multiple occ user:add in parallel

An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'principals/system/system-system' for key 'addressbook_index'

@ChristophWurst Do I read the trace correctly that this happens only because the system address book was not created initially and is then trying to get created by each of the occ user:add calls? 

```
{
  "reqId": "VpkednLd2tMVnO4SA1QG",
  "level": 3,
  "time": "2023-12-04T09:52:36+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "no app in context",
  "method": "",
  "url": "--",
  "message": "Exception thrown: Doctrine\\DBAL\\Exception\\UniqueConstraintViolationException",
  "userAgent": "--",
  "version": "29.0.0.2",
  "exception": {
    "Exception": "Doctrine\\DBAL\\Exception\\UniqueConstraintViolationException",
    "Message": "An exception occurred while executing a query: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'principals/system/system-system' for key 'addressbook_index'",
    "Code": 1062,
    "Trace": [
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1938,
        "function": "convert",
        "class": "Doctrine\\DBAL\\Driver\\API\\MySQL\\ExceptionConverter",
        "type": "->",
        "args": [
          [
            "Doctrine\\DBAL\\Driver\\PDO\\Exception"
          ],
          [
            "Doctrine\\DBAL\\Query"
          ]
        ]
      },
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1880,
        "function": "handleDriverException",
        "class": "Doctrine\\DBAL\\Connection",
        "type": "->",
        "args": [
          [
            "Doctrine\\DBAL\\Driver\\PDO\\Exception"
          ],
          [
            "Doctrine\\DBAL\\Query"
          ]
        ]
      },
      {
        "file": "/var/www/html/3rdparty/doctrine/dbal/src/Connection.php",
        "line": 1208,
        "function": "convertExceptionDuringQuery",
[...]

PR #42102, opened 2023-12-07
fix(carddav): don't read after write for address book creation

miaulalala:
## Summary

In clustered setups, a read after write will sporadically fail. We have all the data, so we don't need to query the backend again.

## Checklist


apps/dav/lib/BackgroundJob/CalendarStatusAutomation.php
```diff
@@ -0,0 +1,26 @@
+<?php
+/*
+ * *
+ *  * {$app} App
+ *  *
+ *  * @copyright 2022 Anna Larch <anna.larch@gmx.net>
```
ChristophWurst, line 6: please fix your ide and these license headers

ChristophWurst (commented):
:+1: makes a lot of sense

miaulalala:
> Does this address #41996?

I'm not sure, but it might help.

ChristophWurst (commented):
On a second thought I'm not sure this does help. The INSERT+SELECT is already wrapped in a transaction. A read/write split should still execute both queries on a primary node. Read after write is only problematic outside of transactions.

ChristophWurst:
See https://github.com/nextcloud/groupware/issues/24. Mariadb experts have told me to either
1) Avoid the read
2) Wrap write+read in a transaction

The address book creation does the latter since https://github.com/nextcloud/server/pull/34412.

Did you see this error on a productive instance? Does it run 25?

PR #42103, opened 2023-12-07
[stable28]  fix: Adjust reference picker code for the vue based FilePicker

susnux:
Manual backport of #https://github.com/nextcloud/server/pull/42104


apps/files/src/views/FileReferencePickerElement.vue
```diff
@@ -40,74 +49,55 @@ export default {
...
-				false, // modal
-				FilePickerType.Choose, // type
-				'',
-				{
-					target: this.$refs.picker,
-				},
-			)
+
+		onClose(nodes?: NcNode[]) {
+			if (!nodes) {
```
Pytal, line 88:
```suggestion
			if (nodes.length === 0) {
```
Arrays are always truthy

Pytal (approved):

apps/files/src/views/FileReferencePickerElement.vue
blizzz: Mh, then it should be fixed (also on master), for then it must be true and the other condition could never trigger? Or do I misinterpret something, as you'd wave it through with the checkmark?

Pytal: A prioritized RC3 checkmark rather in anticipation of a quick fix 🙈

blizzz: But it's the last RC before the final, there will be no possibility for a fix anymore, unless something is terribly broken. So either we can live with that, or fix it before merging.

Pytal: Quick fixed :)

blizzz: Thank you! =)

Pytal: Actually the previous quick fix is also not 100% correct as the code reveals some unintuitive behaviour https://github.com/search?q=repo%3Anextcloud-libraries%2Fnextcloud-dialogs%20%22emit(%27close%27%22&type=code which should be handled explicitly

Pushed the real fix

Pytal: In general I'd say we should use `null` rather than `undefined` for explicitness @susnux

blizzz: and the fixes should go to the master PR as well

blizzz (approved):

PR #42109, opened 2023-12-07
Clean-up and standardization of "App password"

rakekniven:
Reported at Transifex.

We have only two occurences at Transifex matching "App token".
We have 21 matching "App password".

After looking at the UI, code and discussion at community chat I propose to align the wording.

### DIFF
--- a/core/templates/loginflow/authpicker.php
+++ b/core/templates/loginflow/authpicker.php
@@ -57,7 +57,7 @@ $urlGenerator = $_['urlGenerator'];
 			<label for="user" class="infield"><?php p($l->t('Username')) ?></label>
 		</p>
 		<p class="groupbottom">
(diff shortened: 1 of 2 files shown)


susnux (approved):

nfebe (commented):
Please can you fix up (or squash) the commits? There are all the same change?

Fenn-CS:
> How to do this using the GitHub web UI?

Oh, in that case, I would just squash or recommend that anyone who does squashes before merge. While the changes look okay, I do not know anything about this word alignment so I would tag @AndyScherzinger and once he gives a green light, we would get this in.

Thanks for the pr @rakekniven (no action required from your end)

juliusknorr (approved):

juliushaertl:
Squashed the commits into one since squash-merge is disabled here

PR #42132, opened 2023-12-09
enh(oauth2): allowed toggling of aria label

### LINKED ISSUE #41824: [BITV] 9.1.1.1a/10.1 - In the "OAuth 2.0 clients" table, there is a button labeled "Show client secret." This button consistently displays the same alt text regardless of whether the "client secret" has been revealed or hidden.  (1)
This should be modified. It is best to update the alt text to correspond with "Show client secret" and "Hide client secret."

`<button data-v-8f4e0666="" data-v-7be1d638="" 
aria-label="Show client secret" type="button" 
class="button-vue button-vue--icon-only button-vue--vue-tertiary-no-background">`

https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-cd239865a5-v10-n1

emoral435:
* Resolves: #41824

## Summary
The button in the table for OAuth 2.0 previously only had the aria label of "Show client secret", even when the button had toggled the rendering of the client secret. Switched aria label to "Hide client secret" when the rendered secret is visible!

## TODO
- [x] add a way to toggle between the states and correctly assign the aria label

## Images

| Before | After |
| ------------- | ------------- |
|image|(image|

## Checklist


apps/oauth2/src/components/OAuthItem.vue
```diff
@@ -28,7 +28,7 @@
 			<div class="action-secret">
 				<code>{{ renderedSecret }}</code>
 				<NcButton type="tertiary-no-background"
-					:aria-label="t('oauth2', 'Show client secret')"
+					:aria-label="t('oauth2', `${renderSecret ? 'Hide client secret' : 'Show client secret' }`)"
```
ShGKme, line 31: The value of the `t` function should be a constant string so the translation key can be extract

```suggestion
					:aria-label="renderSecret ? t('oauth2', 'Hide client secret') : t('oauth2', 'Show client secret')"
```

In this case, I'd move it to a computed to not overload the template.

ShGKme (dismissed):

ShGKme (changes requested):
The icon on this button also has a fixed title. 
```
:title="t('oauth2', 'Show client secret')"
```

Using title on a span is not correct, it should be on the button and it has same dynamic value with "Hide".

But in this case, I think we can just remove the `title` attr here.

emoral435:
nice catch! gonna delete that part and then submit the new version : )

apps/oauth2/src/components/OAuthItem.vue
```diff
@@ -87,6 +86,18 @@ export default {
 				return '****'
 			}
 		},
+		
+		/**
+		 * toggles the aria-label for button to show the secret key for OAuth 2.0 Clients
+		 */
+		toggleAriaLabel() {
+			// if the secret key is not rendered (for examples, !this.rendered == '*******'), then suggest to show the secret key
```
ShGKme, line 94: Not sure if we need so detailed comments here, `toggleAriaLabel` describes the meaning for me

ShGKme (approved):
Works and seems fine now

apps/oauth2/src/components/OAuthItem.vue
emoral435: Gotcha : ) I will keep the second comment but remove the header comment for consistancy

Pytal: Still present 🙈

emoral435: Decided to delete the second comment too in the most recent pr :)

emoral435:
@Pytal removed all comments 👍, agree with Grigorii that the name itself is self explanatory!

apps/oauth2/src/components/OAuthItem.vue
```diff
@@ -19,7 +19,7 @@
   - along with this program. If not, see <http://www.gnu.org/licenses/>.
   -
   -->
-<template>
+  <template>
```
ShGKme, line 22: There are unneeded extra spaces

ShGKme:
```suggestion
<template>
```

apps/oauth2/src/components/OAuthItem.vue
```diff
@@ -87,6 +86,14 @@ export default {
 				return '****'
 			}
 		},
+
```
ShGKme, line 89: Same here:

```suggestion

```

ShGKme (commented):
It should have been catched by linter...

apps/oauth2/src/components/OAuthItem.vue
emoral435: changed!

apps/oauth2/src/components/OAuthItem.vue
```diff
@@ -87,6 +86,13 @@ export default {
 				return '****'
 			}
 		},
+		toggleAriaLabel() {
+			if (!this.renderSecret) {
+				return t('oauth2', 'Show client secret')
+			} else {
+				return t('oauth2', 'Hide client secret')
+			}
```
Pytal, line 94:
```suggestion
			}
			return t('oauth2', 'Hide client secret')
```
Early return pattern instead of if-else :)

Pytal (approved):

PR #42142, opened 2023-12-11
fix(dav): Prevent out-of-office event time drifts

### LINKED ISSUE #42139: [Bug]: Absence event starts one day too early and is always set to UTC

### Bug description

Despite https://github.com/nextcloud/server/pull/41703, setting an out of office with 28 RC3 leads to a calendar event that starts one day too early and is drifting for timezones with an offset due to the event being set to UTC.

### Steps to reproduce

1. Log into Nextcloud
2. Configure an absence
3. Open Calendar

### Expected behavior

Absence events starts 00:00 of my first day of absence and ends 00:00 of the first day I'm back.

### Installation method

None

### Nextcloud Server version

28

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

ChristophWurst:
* Resolves: https://github.com/nextcloud/server/issues/42139

## Summary

This fixes both symptoms of https://github.com/nextcloud/server/issues/42139

Users whose calendar has no `timezone` property had wrong calendar start dates when their actual timezone has an offset from UTC. This happens for CET for example. The fix is to use the new *timezone service* that has more sources for a timezone and therefore higher changing of finding data.

The known downside of this fix is that in contrast to the calendar timezone, the timezone service only returns the *timezone id*, not the full *VTIMEZONE* object. Therefore the ooo events have DTSTART and DTEND with TZID=x but no VTIMEZONE component. This is not ideal but IMO a better solution than the current drifts.

## How to test

0. Clear your absence if you already had one
1. Set your default calendar `timezone` to NULL in oc_calendars
2. Schedule a new absence
3. Open the calendar

master: event is created, but at UTC and it starts one day before absence if you live in UTC+1
here: event is created, has DTSTART and DTEND set to your timezone id and starts at the first day of absence.

## Checklist

### DIFF
--- a/apps/dav/lib/Listener/OutOfOfficeListener.php
+++ b/apps/dav/lib/Listener/OutOfOfficeListener.php
@@ -26,9 +26,11 @@ declare(strict_types=1);
 namespace OCA\DAV\Listener;
 
 use DateTimeImmutable;
+use DateTimeZone;
 use OCA\DAV\CalDAV\CalDavBackend;
 use OCA\DAV\CalDAV\Calendar;
 use OCA\DAV\CalDAV\CalendarHome;
+use OCA\DAV\CalDAV\TimezoneService;
 use OCA\DAV\ServerFactory;
 use OCP\EventDispatcher\Event;
 use OCP\EventDispatcher\IEventListener;
@@ -40,9 +42,6 @@ use OCP\User\IOutOfOfficeData;
 use Psr\Log\LoggerInterface;
 use Sabre\DAV\Exception\NotFound;
 use Sabre\VObject\Component\VCalendar;
-use Sabre\VObject\Component\VEvent;
-use Sabre\VObject\Component\VTimeZone;
-use Sabre\VObject\Reader;
 use function fclose;
 use function fopen;
 use function fwrite;
@@ -55,6 +54,7 @@ class OutOfOfficeListener implements IEventListener {
 	public function __construct(
 		private ServerFactory $serverFactory,
 		private IConfig $appConfig,
+		private TimezoneService $timezoneService,
 		private LoggerInterface $logger
 	) {
 	}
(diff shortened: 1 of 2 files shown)


apps/dav/lib/Listener/OutOfOfficeListener.php
```diff
@@ -188,21 +190,6 @@ private function createVCalendarEvent(IOutOfOfficeData $data, ?string $timeZoneD
...
-		if ($timeZoneData !== null) {
-			/** @var VCalendar $vtimezoneObj */
-			$vtimezoneObj = Reader::read($timeZoneData);
-			/** @var VTimeZone $vtimezone */
-			$vtimezone = $vtimezoneObj->VTIMEZONE;
-			$calendarTimeZone = $vtimezone->getTimeZone();
-			$vCalendar->add($vtimezone);
-
-			/** @psalm-suppress UndefinedMethod */
-			$vEvent->DTSTART->setDateTime($start->setTimezone($calendarTimeZone)->setTime(0, 0));
```
ChristophWurst, line 202: the day drift happened here because we set the time above without a timezone. if the actual time is on UTC+1 and you set to 00:00 the date moves by one day. So even if we set the time again here, it's one day too early.

setting timezone first, then the time ensures that the day does not change unexpectedly

tcitworld:
> Therefore the ooo events have DTSTART and DTEND with TZID=x but no VTIMEZONE component. This is not ideal but IMO a better solution than the current drifts.

Want to create an issue to have such a service that generates VTIMEZONE data? There's probably a lot of places it could be used.

tcitworld (approved):

ChristophWurst:
> Want to create an issue to have such a service that generates VTIMEZONE data? There's probably a lot of places it could be used.

https://github.com/nextcloud/server/issues/42145

miaulalala (approved):
Looks reasonable

ChristophWurst:
@miaulalala could you please test it, too?

nickvergessen (approved):
Makes sense

miaulalala:
Tested it, works.

blizzz:
accidentally merged the wrong PR, sorry. ^ recreating the backport.

Issue #42143, opened 2023-12-06
Subscription Calendar Events are not returned by REPORT dav request


Juergen-Vollmer:
### Steps to reproduce

1. Add a new calender from an ical link 
2. the dates does not show anymore

### Expected behavior

The dates shpoud show, it workded until approx 2 weeks ago.

### Actual behaviour

the dates does not show anymore in the webinterface of nextcloud

### Calendar app version

4.6.0

### CalDAV-clients used

davx5

### Browser

firefox, chromium actual versions

### Client operating system

opensuse 15.5

### Server operating system

opensuse 15.5

### Web server

Apache

### Database engine version

MariaDB

### PHP engine version

None

### Nextcloud version

28.0.0 rc2

### Updated from an older installed version or fresh install

Updated from an older version

### Nextcloud configuration

```Text
Enabled:
- activity: 2.20.0
- admin_audit: 1.18.0
- calendar: 4.6.0
- circles: 28.0.0-dev
- cloud_federation_api: 1.11.0
- comments: 1.18.0
- contacts: 5.5.0-rc.2
- cookbook: 0.10.4
- dav: 1.29.1
- deck: 1.12.0-beta.4
- federatedfilesharing: 1.18.0
- federation: 1.18.0
- files: 2.0.0
- files_external: 1.20.0
- files_markdown: 2.4.1
- files_mindmap: 0.0.30
- files_pdfviewer: 2.9.0
- files_reminders: 1.1.0
- files_sharing: 1.20.0
- files_trashbin: 1.18.0
- files_versions: 1.21.0
- firstrunwizard: 2.17.0
- forms: 4.0.0-alpha.2
- groupfolders: 16.0.0
- keeweb: 0.6.13
- logreader: 2.13.0
- lookup_server_connector: 1.16.0
- mail: 3.5.0-rc.2
- nextcloud_announcements: 1.17.0
- notes: 4.9.0-beta.3
- notifications: 2.16.0
- oauth2: 1.16.3
- password_policy: 1.18.0
- photos: 2.4.0
- polls: 6.0.0-rc1
- previewgenerator: 5.4.0
- privacy: 1.12.0
- provisioning_api: 1.18.0
- recommendations: 2.0.0
- related_resources: 1.3.0
- richdocuments: 8.3.0-beta.1
- serverinfo: 1.18.0
- settings: 1.10.0
- sharebymail: 1.18.0
- spreed: 18.0.0-rc.2
- support: 1.11.0
[...]

PR #42144, opened 2023-12-11
fix(caldav): don't reuse query builder objects

miaulalala:
* Resolves: 

## Summary
Reusing query builder objects causes a `Doctrine\DBAL\Query\QueryException`

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/CalDavBackend.php
+++ b/apps/dav/lib/CalDAV/CalDavBackend.php
@@ -3092,11 +3092,13 @@ class CalDavBackend extends AbstractBackend implements SyncSupport, Subscription
 				->andWhere($query->expr()->eq('calendartype', $query->createNamedParameter(self::CALENDAR_TYPE_SUBSCRIPTION)))
 				->executeStatement();
 
+			$query = $this->db->getQueryBuilder();
 			$query->delete('calendarchanges')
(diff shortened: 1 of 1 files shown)


ChristophWurst (commented):
Makes sense! 

What I find suspicious is that this is described as a regression, like the feature broke recently. But the QB recyling has happened since the feature was added in https://github.com/nextcloud/server/commit/30d13bb760b233eb4b55b46066deb4bf05d41ce4.

ChristophWurst:
https://github.com/nextcloud/server/commit/7f0a54c5581b3bde0f271d6e6a549076f14354dc only logs a warning but doesn't abort.
While the qb recyling is really ugly here, I think it worked because discarding the previous WHEREs for every recycle is fine :thinking:

tcitworld (approved):

tcitworld:
Yeah, I think it's unlikely this will fix https://github.com/nextcloud/server/issues/42143, but they can try to apply the simple patch.

ChristophWurst (approved):
Fine by me but won't help with https://github.com/nextcloud/server/issues/42143

X-Raph-X:
Hi, I've patched my NC28.0.0 because I have the issue #42143 . I confirm the issue is **NOT** solved by this commit.

PR #42167, opened 2023-12-11
Unified search improvements

### LINKED ISSUE #41381: Global search beta follow up

### LINKED ISSUE #41939: Can not search for "own" search results
### How to use GitHub

### Steps

* Write chat messages (or any thing else that has a search provider supporting the person/people filter)
* Try to search for things from yourself

### Expected

✅ 

### Actual

grafik

### Problem

The global search uses the `/index.php/contactsmenu/contacts` endpoint.

This is problematic in multiple ways:
1. Own user is not returned
2. Endpoint can not be used by desktop/mobile clients (needs OCS)

### Potential quick workarounds

Hard code the own user into the options because we can generate all the details from existing JS data.

nfebe:
Resolves : https://github.com/nextcloud/server/issues/41381, https://github.com/nextcloud/server/issues/41939


Fenn-CS:
Users can now see themselves in the people filter as "Me" once, they start searching the filtering would exclude the authenticated user from the results.

Screenshot from 2023-12-12 19-45-15

susnux:
@Fenn-CS conflicts

Fenn-CS:
@susnux That was the result of compiling and merge : https://github.com/nextcloud/server/pull/42094, now fixed!

susnux (approved):

sorbaugh (approved):

core/src/services/UnifiedSearchService.js
```diff
@@ -103,5 +104,20 @@ export async function getContacts({ searchTerm }) {
...
 	})
+	/*
+	 * Add authenticated user to list of contacts for search filter
+	 * If authtenicated user is searching/filtering, do not add them to the list
+	 */
+	if (!searchTerm) {
+		let authenticatedUser = getCurrentUser()
+		authenticatedUser = {
+			id: authenticatedUser.uid,
+			fullName: 'Me',
```
nickvergessen, line 115: Should use the displayname (like e.g. github does):
```
fullName: authenticatedUser.displayName,
```

Otherwise we need to translate it.

Fenn-CS: Thanks for catching that @nickvergessen here : https://github.com/nextcloud/server/pull/42205

PR #42174, opened 2023-12-12
chore(comments): Drop dead code from comments app

susnux:
## Summary
Drop code that is not used anymore from repository - also drop "comments" script / entry point which is unused.
## Checklist


emoral435 (approved):

apps/comments/.l10nignore
```diff
@@ -1,3 +0,0 @@
-# compiled vue templates
-src/templates.js
-js/
```
artonge, line 3: Isn't that line needed?

artonge (approved):
What was the use of this script?

apps/comments/.l10nignore
susnux: `src/templates.js` is removed
`js/` does not exist as compiled code is provided in `../../dist`

artonge: Sure :see_no_evil:

susnux:
Drone (samba) unrelated

susnux:
> What was the use of this script?

Old API that was dropped long time ago it seems. The only API used by other apps is the `OCA.Comments.View` used by e.g. announcement center

artonge:
I remember @nickvergessen was also afraid to lose an API for Talk

nickvergessen:
> Old API that was dropped long time ago it seems

It was not dropped. I prevent removal just last month....

This is public API:
https://github.com/nextcloud/server/blob/aa5f037af71c915424c6dcfd5ad2dc82797dc0d6/lib/private/Comments/Manager.php#L1595-L1603
Can we just keep it for now?

We can deprecate and drop in 3 years if there are reasons apart from "just because"

nickvergessen:
See https://github.com/nextcloud/server/pull/41491#pullrequestreview-1732075803 and further

nickvergessen:
Reverting in https://github.com/nextcloud/server/pull/42207 to restore the nice functionality and public API

susnux:
@nickvergessen that is a different script that is still provided. `comments-app` is still provided.

PR #42184, opened 2023-12-13
enh(UnifiedSearch): Keep the searchbar on top of the modal

susnux:
## Summary
For me it is a bit inconvenient that I have to scroll all the way up to change the search query if I was scrolling through the results of the new unified search modal.
Therefor I changed the modal to keep the search bar and filter on top while scrolling (except for very small screens where this would be an accessibility problem).

*Note* The diff looks worse than it is due to white space changes...

### After :sparkles: 

https://github.com/nextcloud/server/assets/1855448/3b31bdeb-c64f-4f01-8dd7-cd0176a65306

### Before

https://github.com/nextcloud/server/assets/1855448/ec4b9e31-03df-4d86-9fc0-4fd7aaac1669

## Checklist


nfebe (commented):
Why is there a lot of whites pace changes? When this already passes out lint check? Unless the white space changes are necessitated by the changes required for this work, I think it's better we have commit that does exactly what the commit says and another one for the white space changes.

It might be worthwhile to introduce and editor-config file to make everyone always ends up with the same set of white space changes. 


The sticky search bar look better though....🚀🚀

susnux:
@Fenn-CS because I needed to indent the search bar one level meaning there is a wrapper div to keep all sticky parts in place (see the "...__header" class).

nfebe (approved):

jancborchardt (changes requested):
Great improvement! Just one design improvement: The scroll container does not extend all the way to the right and bottom, leading entries to be cut off at the bottom and the scrollbar floating instead of being on the right edge of the modal. Could this be fixed? :)

susnux:
Thank you!

@jancborchardt 
> The scroll container does not extend all the way to the right and bottom, leading entries to be cut off at the bottom and the scrollbar floating instead of being on the right edge of the modal. Could this be fixed? :)

This is done now: 

https://github.com/nextcloud/server/assets/1855448/7c6bebe1-c440-4edf-8afa-15c4b319ad0d

jancborchardt (changes requested):
Nice on the scrollbar fix – the cut-off on the bottom is still there though?

image


Compare with current and intended design, nicely cut off only at bottom edge of the modal :)
image

jancborchardt (changes requested):
@susnux now there’s no whitespace on the right of the hover/focus box – it should have the same amount on the right as it has on the left:
image

Sorry for all the little feedback – is this actually an issue with the component that other modals would have to fix on their own also? These are things which need to work out of the box, so might be better to fix in the component directly?

susnux:
@jancborchardt this is not related and an issue with the vue library  - it will be fixed with the next release - ref: https://github.com/nextcloud-libraries/nextcloud-vue/pull/4956

With that it will look like this:

https://github.com/nextcloud/server/assets/1855448/9b638f48-d4c9-48c8-ab12-7ffdb35d72b3

jancborchardt (approved):
Thanks for the info @susnux – looks great now. :)

PR #42205, opened 2023-12-13
Translate me-text in unified search people filter

nfebe:
Follow up : https://github.com/nextcloud/server/pull/42167


artonge (approved):

core/src/services/UnifiedSearchService.js
```diff
@@ -112,7 +112,7 @@ export async function getContacts({ searchTerm }) {
 		let authenticatedUser = getCurrentUser()
 		authenticatedUser = {
 			id: authenticatedUser.uid,
-			fullName: 'Me',
+			fullName: t('core', 'Me'),
```
nickvergessen, line 115:
```suggestion
			fullName: authenticatedUser.displayName,
```
I think the own display name is better? Also removes the question if it should be "me" or "myself"

Cc @jancborchardt

nickvergessen (changes requested):

core/src/services/UnifiedSearchService.js
nimishavijay: Yes, generally we try to avoid using any variant of "Me" in the UI, so display name will work better :)

nickvergessen: grafik

nickvergessen (approved):

nickvergessen:
I had pushed and compiled it already -.-

PR #42207, opened 2023-12-13
Revert "chore(comments): Drop dead code from comments app"

nickvergessen:
Reverts nextcloud/server#42174

API is public and not dead.


susnux (changes requested):
Which API is provided? The API exposed by server ([server/lib/private/Comments/Manager.php](https://github.com/nextcloud/server/blob/aa5f037af71c915424c6dcfd5ad2dc82797dc0d6/lib/private/Comments/Manager.php#L1615-L1623)) is still provided (`comments-app.js`).

nickvergessen:
Ah, I thought that was `comments` but yeah just tested and it still seems to work.

Too many files with similar names. Sorry about the confusion then.

susnux:
> Sorry about the confusion then.

No problem

> Too many files with similar names.
Thats why I cleaned up :wink:

PR #42209, opened 2023-12-13
feat(comments): Add a meta data column for comments

nickvergessen:
* Required for https://github.com/nextcloud/spreed/issues/11206

## Summary

Allows apps to store meta-data information into the comments table. A usecase that exists since a longer time is being able to indicate that a message was sent silently.
For the upcoming "Edit message" we want to store the actor and time information of the last person that edited a chat message, to avoid another database query on every chat request.

While changing the table anyway, I enforced the currently optional reference_id column which talk also relies on.

## Checklist

### DIFF
--- a/lib/public/Comments/IComment.php
+++ b/lib/public/Comments/IComment.php
@@ -280,6 +280,25 @@ interface IComment {
 	public function setReferenceId(?string $referenceId): IComment;
 
 	/**
+	 * Returns the metadata of the comment
+	 *
+	 * @return array|null
+	 * @since 29.0.0
+	 */
+	public function getMetaData(): ?array;
+
+	/**
+	 * Sets (overwrites) the metadata of the comment
+	 * Data as a json encoded array
+	 *
+	 * @param array|null $metaData
+	 * @return IComment
+	 * @throws \JsonException When the metadata can not be converted to a json encoded string
+	 * @since 29.0.0
+	 */
+	public function setMetaData(?array $metaData): IComment;
+
+	/**
 	 * Returns the reactions array if exists
 	 *
 	 * The keys is the emoji of reaction and the value is the total.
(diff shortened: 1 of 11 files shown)


lib/public/Comments/IComment.php
```diff
@@ -279,6 +279,24 @@ public function getReferenceId(): ?string;
...
+	 * returns the meta data of the comment
+	 *
+	 * @return string|null
+	 * @since 29.0.0
+	 */
+	public function getMetaData(): ?string;
+
+	/**
+	 * Sets (overwrites) the meta data of the comment
+	 * It is recommended to store the data as a json encoded array
```
marcelklehr, line 292: Mmh. Simply exposing this as a string and recommending the use of json will likely prevent actors to interact on this field. Can we Enforce the json somehow? Ie. take an array and internally json_encode?

marcelklehr (changes requested):
Psalm complains about the usage of updateQuery in the update method

lib/public/Comments/IComment.php
nickvergessen: Right, makes sense.

nickvergessen: Hmm since comments are exposed via DAV, arrays are problematic (even when all other known consumers of comments don't use metadata):
```
There was 1 failure:

1) OCA\DAV\Tests\unit\Comments\CommentsNodeTest::testGetProperties
Failed asserting that an array has the key '{http://owncloud.org/ns}metaData'.

/__w/server/server/apps/dav/tests/unit/Comments/CommentsNodeTest.php:497
```

Should we simply skip the field on the DAV API until a DAV app makes use of it?

nickvergessen: Or we json_encode it there (in XML content 🙈 )

marcelklehr: I think json encoded in XML would be nice, even if it's weird

nickvergessen: Nevermind, it seems to work. The problem is that the assert has the expected array on the actual. So it got me confused. Will revert my change and add a test for the metaData

marcelklehr (approved):

come-nc (approved):

PR #42285, opened 2023-12-14
fix(theming): Adjust theming util to calculate primary element color based on WCAG color contrast

susnux:
## Summary
This adjusts the theming utils for calculating the primary color:
1. Adjust luma calculation to match with the algorithm used by WCAG 2
2. Add function for calculating color contrast according to WCAG 2
3. Fix the `invertTextColor` by not simply checking the luma channel of the HSL color but the real color contrast
4. Improved primary element color calculation by providing the background color so an element color with high enough contrast can be calculated

Also keep blurry background and hover states in mind while calculating the element color.

Moreover I added a unit test for accessible colors.

### Screenshots
Only the hover state has changed:
before | after
---|---
image|image

## Checklist

### DIFF
--- a/apps/theming/css/default.css
+++ b/apps/theming/css/default.css
@@ -77,7 +77,7 @@
   --color-primary-light-text: #00293f;
   --color-primary-light-hover: #dbe4ea;
   --color-primary-element: #00679e;
-  --color-primary-element-hover: #1674a6;
+  --color-primary-element-hover: #1070a4;
   --color-primary-element-text: #ffffff;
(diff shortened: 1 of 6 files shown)


skjnldsv (approved):

Altahrim (approved):

susnux:
Had to skip the test if the base class was run as a test case on its own as - of cause - the `$theme` was not set there. So in that case the test need to be skipped and not failed.

ShGKme:
Is it intended change?

Dark theme:

Before | After
---|---
image | image

PR #42314, opened 2023-12-15
[stable27] fix(files): Set download property and encode source

Pytal:
- Fix part of https://github.com/nextcloud/groupfolders/issues/2680

## Summary

- Fix download on trashbin files originally in a groupfolder

## Checklist

### DIFF
--- a/apps/files/src/components/FileEntry.vue
+++ b/apps/files/src/components/FileEntry.vue
@@ -120,6 +120,7 @@
 <script lang='ts'>
 import { debounce } from 'debounce'
 import { formatFileSize } from '@nextcloud/files'
+import { encodePath } from '@nextcloud/paths'
 import { Fragment } from 'vue-frag'
 import { join, extname } from 'path'
 import { showError, showSuccess } from '@nextcloud/dialogs'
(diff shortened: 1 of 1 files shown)


Pytal:
Download will trigger but the files will be 0 bytes as the URL is incorrect due to groupfolder trash being located in `<data_dir>/__groupfolders/trash/`

Not sure if downloading from groupfolders trash was possible before @icewind1991? If we want to allow downloads like normal files then this will need adjustments

blizzz:
Still results in a network error for me, and the download url did not really change: `https://my.nc.srv/remote.php/dav/trashbin/$UID/trash/$FILENAME.$EXT.d1675427849`

Pytal:
> Still results in a network error for me, and the download url did not really change: `https://my.nc.srv/remote.php/dav/trashbin/$UID/trash/$FILENAME.$EXT.d1675427849`

Tried after compile?
URL not changed in this PR, only the download property is set which should trigger the 0 byte download

Will need to know from @icewind1991 how groupfolder trash downloads worked before

blizzz:
> > Still results in a network error for me, and the download url did not really change: `https://my.nc.srv/remote.php/dav/trashbin/$UID/trash/$FILENAME.$EXT.d1675427849`
> 
> Tried after compile? URL not changed in this PR, only the download property is set which should trigger the 0 byte download

… and hard reload, yes. Even twice for good measure ;)

Pytal:
Superseded by https://github.com/nextcloud/groupfolders/pull/2716

PR #42345, opened 2023-12-18
feat: Track dirty table writes and long transactions

### LINKED ISSUE #33542: Primary/Secondary choosing querybuilder
In case of write operations mark table as dirty and if read request comes in _and_ table is in a "primary-enforced reads after writes list", so primary-forced reads only happen if write happen and table is configured in the static list where this behavior needs to happen.

juliusknorr:
- feat: First attempt to track dirty tables after writes and switch back to replicas if reads go to other tables
- feat: Add logging for transaction time

* Based on: #41998 

## Summary

The main benefit of this approach is that we can switch back to read replicas for queries that are not towards dirty tables (that were written before), which makes #41998 more performant as less reads go to the primary. In addition the logging can be used to more easily track down read after write by using the debug log.

In addition logging long transactions can be a good indicator for code paths that might need rework to avoid long locks or could cause potential deadlock scenarios.

- [x] Pending documentation as we should cover this in the developer docs https://github.com/nextcloud/documentation/pull/11492
  - [x] Mention log messages to look out for https://github.com/nextcloud/documentation/pull/11492

### DIFF
--- a/lib/private/DB/Connection.php
+++ b/lib/private/DB/Connection.php
@@ -81,6 +81,8 @@ class Connection extends PrimaryReadReplicaConnection {
 
 	protected ?float $transactionActiveSince = null;
 
+	protected $tableDirtyWrites = [];
+
 	/**
 	 * Initializes a new instance of the Connection class.
 	 *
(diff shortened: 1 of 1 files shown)


lib/private/DB/Connection.php
```diff
@@ -618,4 +648,31 @@ protected function performConnect(?string $connectionName = null): bool {
...
+		}
+		return parent::beginTransaction();
+	}
+
+	public function commit() {
+		$result = parent::commit();
+		if ($this->getTransactionNestingLevel() === 0) {
+			$timeTook = microtime(true) - $this->transactionActiveSince;
+			$this->transactionActiveSince = null;
+			if ($timeTook > 1) {
```
ChristophWurst, line 664: this will be true very often. should we use a lower threshold like 10ms?

juliushaertl: microtime will actually return seconds if true is passed on as a parameter so a float is returned. So this would only log if greater 1 second

ChristophWurst (approved):
Nice

ChristophWurst (commented):
typo :)

ChristophWurst (approved):

lib/private/DB/Connection.php
```diff
@@ -618,4 +648,31 @@ protected function performConnect(?string $connectionName = null): bool {
...
+				$this->logger->warning('Transaction took longer than 1s: ' . $timeTook, ['exception' => new \Exception('Long running transaction')]);
+			}
+		}
+		return $result;
+	}
+
+	public function rollBack() {
+		$result = parent::rollBack();
+		if ($this->getTransactionNestingLevel() === 0) {
```
Altahrim, line 673: Not sure about this one. 
First rollback is supposed to cancel all nested transactions right (at least at database level)?

And, maybe it's also interesting to log long rollbacked transactions because they also consume resources.

Altahrim (approved):

lib/private/DB/Connection.php
juliushaertl: Yeah we only track the outer transaction, nested transactions are a bit more complex https://github.com/nextcloud/3rdparty/blob/a71bd8af76fdcfad78c865d1c60f6dde6e24f1dd/doctrine/dbal/src/Connection.php#L1347-L1384

Thanks for the additional log suggestion, that makes a lot of sense. Added that.

Issue #42346, opened 2023-12-18
File metadata performance issues


juliushaertl:
When checking db queries on a propfind with 28 I noticed there are files_metadata related queries that could be a performance issue. Note I only noticed those two from a quick sight, might be worth to use the profiler again to have a closer look a the queries with some more background knowledge:

- Some queries seems to occur once per file in the propfind. It would be much more sensible to prefetch them in batch with an ` IN () ` query:

![image](https://github.com/nextcloud/server/assets/3404133/bd88d73f-b349-4a76-a090-9bfa4d071b4f)

- In addition i noticed we query the custom properties backend for the metadata that is actually handled in a Sabre plugin. Either those should be manually added to the ignore list https://github.com/nextcloud/server/blob/aa5dcb095256d6352de758361798de86c3e2d0d0/apps/dav/lib/DAV/CustomPropertiesBackend.php#L65 or if metadata should be generally not handled maybe we can skip anything with the metadata prefix.

![image](https://github.com/nextcloud/server/assets/3404133/374ea059-1890-4bad-af3c-0ff2ed3cb32c)

This might have some bad impact on larger instances. cc @AndyScherzinger @sorbaugh

AndyScherzinger:
lopping in @sorbaugh @ArtificialOwl @artonge

artonge:
> Some queries seems to occur once per file in the propfind. It would be much more sensible to prefetch them in batch with an IN () query:

I should have used the already available metadata from the `FileInfo`. Here is a fix: https://github.com/nextcloud/server/pull/42349

> In addition i noticed we query the custom properties backend for the metadata that is actually handled in a Sabre plugin. Either those should be manually added to the ignore list 

Can you expand on that, I am not familiar enough with the `IGNORED_PROPERTIES` logic.

juliushaertl:
> Can you expand on that? I am not familiar enough with the IGNORED_PROPERTIES logic.

Sure, the CustomPropertiesBackend is used to store and retrieve any unknown dav property that WebDAV clients may pass in the oc_properties table. One example is windows WebDAV mount storing additional metadata in those properties for the files. For properties that are handled by Nextcloud itself we do not need this logic and can just ignore the property in this plugin, which will save us additional queries. The IGNORED_PROPERTIES is a block list for such properties which will limit the query to only happen when there are actual "custom" properties requested

[...]

PR #42350, opened 2023-12-18
Don't query oc_properties for metadata props

artonge:
This prevents a SQL request on `oc_properies` on each requested files.

Part of https://github.com/nextcloud/server/issues/42346


juliusknorr (approved):

come-nc (approved):

solracsf:
@artonge any objection backporting to 27?

artonge:
It would need to be adapted as the metadata API was not exactly the same

PR #42352, opened 2023-12-18
[Fix] Abort Streamer-processing when user-connection aborted

a-bran:
* Resolves: #8161

## Summary
Streamer, which builds compressed archives to stream to the browser (e.g. for downloading folders), didnt have checks when the browser aborts the connection (e.g. download cancelled).

This PR adds checks at certain positions:
- before file or folder is processed in `addDirRecursive` to avoid opening files or further traversing directories, even though writing is impossible
- before file is given to streamer in `addFileFromStream`
- FileStream is wrapped to close file-Stream when connection aborted while reading
  - This happens after 1-3 stream-read-calls as PHP `connection_status()` only updates when data is attempted to be send to the browser (e.g. with `flush()`)

## Todo
- How the abort is handled while another system is processing streams could maybe be a bit more elegant or moved somewhere else?

### DIFF
--- a/lib/private/Streamer.php
+++ b/lib/private/Streamer.php
@@ -28,7 +28,7 @@
  */
 namespace OC;
 
-use Icewind\Streams\CallBackWrapper;
+use Icewind\Streams\CallbackWrapper;
 use OC\Files\Filesystem;
 use OCP\Files\File;
 use OCP\Files\Folder;
--- a/lib/private/Streamer.php
+++ b/lib/private/Streamer.php
@@ -28,6 +28,7 @@
  */
 namespace OC;
 
+use Icewind\Streams\CallBackWrapper;
 use OC\Files\Filesystem;
 use OCP\Files\File;
 use OCP\Files\Folder;
@@ -126,6 +127,7 @@ class Streamer {
 		/** @var LoggerInterface $logger */
 		$logger = \OC::$server->query(LoggerInterface::class);
 		foreach ($files as $file) {
+			if(connection_status() !== CONNECTION_NORMAL) return;
 			if ($file instanceof File) {
 				try {
 					$fh = $file->fopen('r');
@@ -161,6 +163,15 @@ class Streamer {
 	 * @return bool $success
 	 */
 	public function addFileFromStream($stream, string $internalName, int|float $size, $time): bool {
+		if(connection_status() !== CONNECTION_NORMAL) return false;
+		// Close file-stream when user-connection closed
+		$stream = CallbackWrapper::wrap($stream,
+			function ($count) use ($stream) {
+				if (connection_status() !== CONNECTION_NORMAL) {
+					fclose($stream);
+				}
+			});
+
 		$options = [];
 		if ($time) {
 			$options = [


lib/private/Streamer.php
```diff
@@ -126,6 +127,7 @@ public function addDirRecursive(string $dir, string $internalDir = ''): void {
 		/** @var LoggerInterface $logger */
 		$logger = \OC::$server->query(LoggerInterface::class);
 		foreach ($files as $file) {
+			if(connection_status() !== CONNECTION_NORMAL) return;
```
kesselb, line 130:
```suggestion
			if(connection_aborted() === 1) {
				break;
			}
```

lib/private/Streamer.php
```diff
@@ -161,6 +163,15 @@ public function addDirRecursive(string $dir, string $internalDir = ''): void {
 	 * @return bool $success
 	 */
 	public function addFileFromStream($stream, string $internalName, int|float $size, $time): bool {
+		if(connection_status() !== CONNECTION_NORMAL) return false;
+		// Close file-stream when user-connection closed
+		$stream = CallbackWrapper::wrap($stream,
+			function ($count) use ($stream) {
+				if (connection_status() !== CONNECTION_NORMAL) {
+					fclose($stream);
```
kesselb, line 171: I'm unsure if closing the stream inside the read callback is a good idea. 

I suggested using exceptions to bubble up the state properly here: https://github.com/nextcloud/server/issues/8161#issuecomment-1646853594

lib/private/Streamer.php
```diff
@@ -161,6 +163,15 @@ public function addDirRecursive(string $dir, string $internalDir = ''): void {
 	 * @return bool $success
 	 */
 	public function addFileFromStream($stream, string $internalName, int|float $size, $time): bool {
+		if(connection_status() !== CONNECTION_NORMAL) return false;
```
kesselb, line 166:
```suggestion
		if(connection_aborted() === 1) {
				return false;
		}
```

kesselb (commented):
Thank you :+1:

lib/private/Streamer.php
a-bran: mhh, I thought we can handle this gracefully.
The reading system has to check `fefo` or if `fread` succeeded anyways and can then finalize.

In this case e.g. ZipStreamer doesnt catch exceptions so they just bubble to the top and then after everything finished the file is closed by PHP garbage collection.
On theother hand, ZipStreamer does finalize the Zip stream which is a waste of processing.

If we want an exception (so we force cleanup, if any, to execute) we should be consistent and do this in any other case aswell.

a-bran:
@kesselb Hey, some time has passed, sorry if i am bothing you by pinging.
Do you have some inputs on the comments i left in response?

lib/private/Streamer.php
Altahrim: I think we can even skip the `=== 1`:
```suggestion
			if (connection_aborted()) {
				break;
			}
```
Same for the following calls to `connection_status`

PR #42353, opened 2023-12-18
Add global accessibility switch

### LINKED ISSUE #42222: [BITV]: create a server side switch for activation / deactivation a Grid View
Grid view in files (core app) is not accessible. A good decision would be to create a configurable parameter in a `config` file which can be enabled or disabled for every instance individually.

Please create a comment above this parameter in `config` file with a text: _this feature is not accessible and is not BITV confirm._

Thanks a lot!

artonge:
Adds new global config switch to disable non-accessible features: `enable_non-accessible_features`. (config.php)

Fix https://github.com/nextcloud/server/issues/42222 https://github.com/nextcloud/server/issues/42262

PR for the image editor: https://github.com/nextcloud/viewer/pull/2100


nfebe (commented):
Code looks okay, but this mean, users cannot isolate weather to use new unified search from using the grid view...

Is this okay @AndyScherzinger

susnux:
There are two switches:
1. Enable new unified search or keep the legacy one
2. Optionally Disable **all** not accessible features (new unified search, grid view, image editor). As soon as one of the features is accessible we can remove it from the list of not accessible features.

artonge:
> There are two switches:

In the current state, there is only one switch, the accessibility one.

I thought the switch to toggle between unified and legacy search was only due to accessibility concern. Was I wrong?

Fenn-CS:
> I thought the switch to toggle between unified and legacy search was only due to accessibility concern. Was I wrong?

For Unified search, it could be a thing that there are accessibility issues, but the initial idea was to allow users who don't like new unified search to switch back, until we think it's stable enough.

For example issues with unified search and filtering apps, or some user running into a problem with the new search then they can disable it.

AndyScherzinger:
> Is this okay @AndyScherzinger

Yes this is okay, since we want to make things accessible anyway, where-ever possible. In case a system needs to be accessible by law, than like it is designed it is a yes/no question, hence a single config switch.

AndyScherzinger:
> For Unified search, it could be a thing that there are accessibility issues, but the initial idea was to allow users who don't like new unified search to switch back, until we think it's stable enough.

That switch is also still there. So all good.

config/config.sample.php
```diff
@@ -2400,9 +2400,9 @@
 'reference_opengraph' => true,
 
 /**
- * Enable use of old unified search
+ * Disable features that are do respect accessibility standards yet.
```
ShGKme, line 2403: To be aligned with the property name (not `enable_*` for "disabling").

```suggestion
 * Enable features that don't respect accessibility standards yet.
```

ShGKme: However, IMO, `'disable_non-accessible_features' => false"` config by default would fit better, so default while would be the same as false and the focus is on disabling.

AndyScherzinger: Well I would say from a language perspective:

* `true` should mean `ON`
* `false` should mean `OFF`

So a description of disable+false translates to "enable" and that is hard to process for humans.

artonge: Yes, thanks for the correction, I started with `disable...` and switched to `enable..` because of @AndyScherzinger explication, but I forgot to adjust the comment.

lib/private/TemplateLayout.php
```diff
@@ -108,8 +108,8 @@ public function __construct($renderAs, $appId = '') {
 
 			$this->initialState->provideInitialState('core', 'active-app', $this->navigationManager->getActiveEntry());
 			$this->initialState->provideInitialState('core', 'apps', $this->navigationManager->getAll());
-			
-			if ($this->config->getSystemValueBool('unified_search.enabled', false)) {
+
+			if ($this->config->getSystemValueBool('unified_search.enabled', false) || !$this->config->getSystemValueBool('enable_non-accessible_features', true)) {
```
artonge, line 112: @Fenn-CS I've added the former switch back. However, I have some doubt about how it is used. Can you double-check and confirm that the condition is reversed?

Shouldn't it be:

```suggestion
			if (!$this->config->getSystemValueBool('unified_search.enabled', false) || !$this->config->getSystemValueBool('enable_non-accessible_features', true)) {
```

Fenn-CS: @artonge It looks okay, the confusion is stemming form the name, the new search was called "global search" so, this was to turn on the former search called "unified search" ... However all of them are now called unified search so that seems off...

it could be renamed to `unified_search.enabled`.

All that it does it toggle between showing the new search modal or the former search tooltip!

nfebe (approved):

skjnldsv (approved):

AndyScherzinger (approved):

PR #42361, opened 2023-12-18
fix(manager): removed dead code

### LINKED ISSUE #32373: Dead code?
It seems that this test is always false

https://github.com/nextcloud/server/blob/fe33e9c08cbbc80738660d2d78838f04d24e0e2e/lib/private/Share20/Manager.php#L1571

emoral435:
## 🚧 Summary
After looking through the codebase, what the piece of code was doing before was just checking the type of sharing that the file had enabled. This leaves up to question - does it directly impact the password settings that the other checks do not do?

So, here's what the check did - and why I removed it
- Checked if it was a link share, email share, or circle share

As the type of shares being used lets the user dictate if there is a password or not, the other checks in place are more useful towards checking if there is a password on the shared object.

## Checklist

- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required (NA)

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -1573,14 +1573,8 @@ class Manager implements IManager {
 	 * @return bool
 	 */
 	public function checkPassword(IShare $share, $password) {
-		$passwordProtected = $share->getShareType() !== IShare::TYPE_LINK
-			|| $share->getShareType() !== IShare::TYPE_EMAIL
-			|| $share->getShareType() !== IShare::TYPE_CIRCLE;
-		if (!$passwordProtected) {
-			//TODO maybe exception?
-			return false;
-		}
 
+		// if there is no password on the share object / passsword is null, there is nothing to check
 		if ($password === null || $share->getPassword() === null) {
 			return false;
 		}


lib/private/Share20/Manager.php
```diff
@@ -1573,14 +1573,8 @@ protected function checkShare(IShare $share): void {
 	 * @return bool
 	 */
 	public function checkPassword(IShare $share, $password) {
-		$passwordProtected = $share->getShareType() !== IShare::TYPE_LINK
-			|| $share->getShareType() !== IShare::TYPE_EMAIL
-			|| $share->getShareType() !== IShare::TYPE_CIRCLE;
```
nickvergessen, line 1578: What a weird code section. The value can never be all 3 of the constants... so this is always true?
I tracked it back until https://github.com/nextcloud/server/commit/c191173d5914f77ebac425e315d2b962077bd654#diff-6c825986578290fe355bffe05c5232629a36e554d48f5d889bb2c1ec73645265R1113-R1114 and there it was already wrong.

So, nice find.

nickvergessen (approved):

come-nc (approved):

PR #42379, opened 2023-12-19
enh(breadcrumbs): removed unnecessary aria label

### LINKED ISSUE #42265: 9.2.5.3/3.1 In the breadcrumb navigation, the directory name is being overridden by the use of "aria-label" attributes. This could cause issues when using voice control programs.
9.2.5.3/3.1 In the breadcrumb navigation, the directory name is being overridden by the use of "aria-label" attributes. This could cause issues when using voice control programs.

image

emoral435:
## ☑️ This fix depends on nextcloud-libraries/nextcloud-vue#4973 to be merged first! Resolves #42265 

##🚧 Summary
This is a quick review, for a small change due to the changes made in a nextcloud component. As the aria label is being set on the component, this is redundant, as the section that needs the aria label is set on the correct element in the new version of the component.

## Checklist

- [ ] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required (NA)


apps/files/src/components/BreadCrumbs.vue
ShGKme, line 1: As the method is only used for `title`, I'd rename it to `titleForSection(section)`

ShGKme (commented):
I think we should also specify `aria-label` on `NcBreadcrumbs` so it will be clear that this navigation is the current directory path.

But I'm not sure we should add the title for each button.

It may make sense for the first button because it has only an icon. And maybe for the last one because it has a different value with `Reload current directory`.

But it may overload a user giving to much information. I'd even consider removing `Reload current directory`, because via screen reader it is known that this button is a link for the current directory.

skjnldsv (approved):
:elephant:

apps/files/src/components/BreadCrumbs.vue
emoral435: changed!

emoral435:
@ShGKme So, I changed the method name to `titleForSection`, added an aria label to the nav wrapper that is the NcBreadCrumbs component, and removed the title attribute from all the breadcrumbs except the home icon!

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -1,16 +1,16 @@
...
 			:key="section.dir"
-			:aria-label="ariaLabel(section)"
-			:title="ariaLabel(section)"
 			v-bind="section"
 			dir="auto"
 			:to="section.to"
 			@click.native="onClick(section.to)">
 			<template v-if="index === 0" #icon>
-				<Home :size="20" />
+				<Home :size="20" :title="titleForSection(section)" />
```
ShGKme, line 13: I apologize for nitpicks here again =D

But `title` attribute should be kept on the `NcBreadcrumb` as an interactive element (button/link), it should not be moved to icon only.

- Icon is hidden for accessibility via `aria-hidden="true"`
- On icon title works only if the pointer is on the icon, not just on the button
- In general, `title` should be set on interactive elements

emoral435: Its no problem! Better to keep the codebase as regulated as possible 👍 Will be adding back the titleForSection back to the title attribute for `NcBreadCrumb`!

ShGKme (approved):
Need to check how it is read with other than NVDA screen readers when there as both Text content and Title in navigation links.

With NVDA either title or text content is read and works fine.

JuliaKirschenheuter (changes requested):
Screenshot from 2023-12-22 16-22-04

please check the console

emoral435:
Should be fixed now! Just added t as a method, forgot to do so before :)

JuliaKirschenheuter (approved):
Works now, thank you!

Sorry, i know that it is a bit too late but i think it would be better to remove native tooltips from breadcrumbs (all `title` attrs) because then we will face with a problem again: tooltips are there but there is no other way to discover this information besides mouse over. But i can create an issue for that

JuliaKirschenheuter:
https://github.com/nextcloud/server/pull/42379#pullrequestreview-1794663900 -> https://github.com/nextcloud/server/issues/42460

PR #42382, opened 2023-12-19
Refresh FileList on current folder update

### LINKED ISSUE #845: Refresh current folder view
What would be the way to refresh the current directory in the files app?

In my app I dont know if/what files have changed so I just need to do a full refresh.

Previously I would use this 

```js
const fileList = OCA.Files?.App?.fileList ?? OCA?.Sharing?.PublicApp?.fileList
if (!fileList) {
  return
}
fileList.changeDirectory(fileList.getCurrentDirectory(), true, true)
```
https://github.com/Raudius/files_scripts/blob/v3.0.0/src/files.ts#L3

Raudius:
* Resolves: https://github.com/nextcloud-libraries/nextcloud-files/issues/845

## Summary
This PR subscribes the `FileList` view to the file-updated event. If the file updated matches the current folder it re-fetches the current view.

## Checklist


apps/files/src/views/FilesList.vue
```diff
@@ -556,6 +557,17 @@ export default defineComponent({
 			showError(this.t('files', 'Unknown error during upload'))
 		},
 
+		/**
+		 * Refreshes the current folder on update.
+		 *
+		 * @param {Node} node is the file/folder being updated.
+ 		 */
+		onUpdatedNode(node) {
+			if (node.fileid === this.currentFolder.fileid) {
```
skjnldsv, line 566: To be safe?

```suggestion
			if (node?.fileid === this.currentFolder?.fileid) {
```

skjnldsv (approved):

hschletz:
This PR won't work because the function names mismatch: subscribe() uses "onUpdateNode", but actual name is "onUpdatedNode".

Raudius:
Nice catch @hschletz! I have now fixed it and tested again using my app which already integrates the changes:

https://github.com/Raudius/files_scripts/blob/master/src/files.ts#L10

apps/files/src/views/FilesList.vue
```diff
@@ -436,6 +436,7 @@ export default defineComponent({
 
 	mounted() {
 		this.fetchContent()
+		subscribe('files:node:updated', this.onUpdatedNode)
```
artonge, line 439: No need to unsubscribe?

artonge: If not, then a comment would be appreciated :)

Raudius: Thanks for the tip, I am not familiar with the event bus.

Should be fixed now

artonge (approved):

artonge:
@Raudius please rebase and compile again, then it will be good to merge :)

PR #42391, opened 2023-12-20
feat(cron): Warn on excessive memory consumption in background jobs

nickvergessen:
* Resolves: # 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -148,7 +148,8 @@ try {
 				break;
 			}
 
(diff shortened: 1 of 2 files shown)


ChristophWurst:
would be helpful if you could log the arguments too so we can pinpoint the job instance, e.g. for mail and the account synced in the job

maybe as logging context argument?

ChristophWurst (approved):

come-nc (approved):

PR #42400, opened 2023-12-20
[stable28] Update crl after revoke rides.csr

mgallien:
Backport of https://github.com/nextcloud/server/pull/42140


nickvergessen (approved):

miaulalala (approved):

mgallien:
`integration-dav-v2` test failed in drone but I would bet this is unrelated
I would prefer not to restart it to not add extra load on drone

mgallien:
@miaulalala @nickvergessen what should I do for `Block merges during freezes / Block merges during freezes (pull_request)` check ?

PR #42419, opened 2023-12-21
fix(files): Adjust ID for skip content buttons

### LINKED ISSUE #41696: [BITV]: Skip links are not working anymore
Try to navigate via skip links to "main content" or "navigation". Nothing happens.

Peek 2023-11-23 15-44

susnux:
* Resolves: #41696 for the files app

## Summary
As we now use Vue the ID needs to be adjusted.

## Checklist

### DIFF
--- a/apps/files/lib/Controller/ViewController.php
+++ b/apps/files/lib/Controller/ViewController.php
@@ -280,7 +280,9 @@ class ViewController extends Controller {
 		$this->initialState->provideInitialState('templates', $this->templateManager->listCreators());
 
 		$params = [
-			'fileNotFound' => $fileNotFound ? 1 : 0
+			'fileNotFound' => $fileNotFound ? 1 : 0,
+			'id-app-content' => '#app-content-vue',
+			'id-app-navigation' => '#app-navigation-vue',
 		];
 
 		$response = new TemplateResponse(
--- a/apps/files/tests/Controller/ViewControllerTest.php
+++ b/apps/files/tests/Controller/ViewControllerTest.php
@@ -186,6 +186,8 @@ class ViewControllerTest extends TestCase {
 			'index',
 			[
 				'fileNotFound' => 0,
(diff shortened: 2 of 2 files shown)


artonge (approved):

ShGKme (changes requested):
Main content may have ids:
1. `#app-content-vue` in Vue apps from rendered as [NcAppContent](https://github.com/nextcloud-libraries/nextcloud-vue/blob/c91077681f78471a09618f82190fdaa83285ba42/src/components/NcAppContent/NcAppContent.vue#L74), usually used in full Vue apps with rendered as `NcContent` with navigation and sidebar (Talk, Photos, Activity, Users settings, Apps settings) and in Files (there is `NcAppContent` but no `NcContent`).
2. Set in Template via `id-app-content`. By default, it is `#app-content` for user and `#content` for public, but can be custom set via `id-app-content` property. Used by not full Vue apps (not `NcContent` apps), for example, any page in Personal or Admin settings, Help, Public share.

**Skip links** always uses the ID from p.2.

This PR fixes the issue for Files app but doesn't fix the issue in general. For example, Photos app, Activity app.

I think we should have a solution that fixes **Skip links** in a common way, otherwise we need to fix it in each app separately and have no control over it.

susnux:
> think we should have a solution that fixes **Skip links** in a common way, otherwise we need to fix it in each app separately and have no control over it.

I know this is just the first quick fix. I currently prepare a more general fix on the vue library.

ShGKme (approved):
A common solution from `@nextcloud/vue` will work for general apps like Photos, Activity or Talk.

This PR is needed because Files app has more complex mounting with `NcAppContent` but without `NcContent` by manual mounting of content, navigation, and sidebar.

JuliaKirschenheuter (approved):

susnux:
/backport 59c982f31fc123bf4039b326ecc1ec999e52474b to stable28

PR #42424, opened 2023-12-21
fix(theming): Add adjusted color versions of default background

### LINKED ISSUE #42196: [BITV]: Replace default background image with darker one to pass a11y contrasts on menu
Right now contrast between foreground and background colors is 4.17:1. This is not enough.

**_Have to be min. 4:5 to 1 in dark theme and light theme._**

Screenshot from 2023-12-13 09-17-52

See: 9.1.4.3/0.1 - Note: For text (including in focused state) under 18 points (or for bold text under 14 points), the contrast ratio between the foreground and background colors should be at least 4.5:1

susnux:
* Resolves: #42196

## Summary
The color was adjusted by @jancborchardt for better / required contrast of the app menu

## Checklist


ShGKme (approved):

emoral435 (approved):

ShGKme:
I wonder if we should change `attribution` here, as it is not an original image anymore.

https://github.com/nextcloud/server/blob/7d8741c53f7be7e77139f124e441e2ba10e66bf2/apps/theming/lib/Service/BackgroundService.php#L133-L138

PR #42436, opened 2023-12-21
enh(settings): Add SetupCheck to ensure webserver correctly serves `.mjs` files

susnux:
## Summary
Even if it was noted in the upgrade guides for Nextcloud 27 some instances are not correctly configured to server `.mjs` files as JavaScript. This breaks apps like logreader.
So this adds a setup check for validating the server is correctly configured.

## Checklist

### DIFF
--- a/apps/settings/lib/SetupChecks/JavaScriptModules.php
+++ b/apps/settings/lib/SetupChecks/JavaScriptModules.php
@@ -1,0 +1,77 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2023 Ferdinand Thiessen <opensource@fthiessen.de>
+ *
+ * @author Ferdinand Thiessen <opensource@fthiessen.de>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OCA\Settings\SetupChecks;
+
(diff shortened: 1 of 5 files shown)


apps/settings/lib/SetupChecks/JavaScriptModules.php
```diff
@@ -0,0 +1,77 @@
...
+		private LoggerInterface $logger,
+	) {
+	}
+
+	public function getCategory(): string {
+		return 'network';
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('Supports JavaScript modules');
```
come-nc, line 54:
```suggestion
		return $this->l10n->t('JavaScript modules support');
```
I *think*?

apps/settings/lib/SetupChecks/JavaScriptModules.php
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	public function run(): SetupResult {
+		$testFile = $this->urlGenerator->linkTo('settings', 'js/esm-test.mjs');
+		$testURLs = array_merge(
+			[$this->urlGenerator->getAbsoluteURL($testFile)],
+			array_map(fn (string $host): string => $host . $testFile, $this->config->getSystemValue('trusted_domains', []))
+		);
```
come-nc, line 62: Can you explain why you test several URLs and why only one of them working is enough?

susnux: Not all of them are accessible from the webserver, like a simple test setup running in docker might have `192.168....` and `172.17.0.2`.
So it might that one of it is not resolved by the local webserver.

come-nc (approved):

apps/settings/lib/SetupChecks/JavaScriptModules.php
```diff
@@ -0,0 +1,77 @@
...
+
+		foreach ($testURLs as $testURL) {
+			try {
+				$client = $this->clientService->newClient();
+				$response = $client->head($testURL, ['connect_timeout' => 10]);
+				if (preg_match('/(text|application)\/javascript/i', $response->getHeader('Content-Type'))) {
+					return SetupResult::success();
```
artonge, line 69: So, the test returns on the first successful request?
Shouldn't the test be a success if all requests are successful?
Also, could the error message be customized with the failing domains?

susnux: See https://github.com/nextcloud/server/pull/42436#discussion_r1446153004

Basically we need to test all available domains but we can not expect all to work as some might not even be resolvable from our local server.

artonge (approved):

susnux:
cypress unrelated

Issue #40866, opened 2023-10-09
Folder Smart Picker broken


jospoortvliet:
**Describe the bug**
The folder/file smart picker in Office seems broken.
First, it forgets its last used location, every time I open it I have to navigate 6 layers deep again. second, the links it then adds are broken.

Also, the top navigation bar where you pick the smart picker doesn't stick with what I choose, adding extra mouse clicks as well.

**To Reproduce**
1. Open a document, go to insert
2. choose smart picker and go for folder
3. insert a file
4. see a broken link

**Expected behavior**
A working link

**Screenshots**
How about a video:
https://cloud.nextcloud.com/f/7915124
(internally shared to office chat room, sorry for external ppl...)

**Client details:**
Mac OS, Firefox, both in Writer and Spreadsheet

## Server details

**Nextcloud version:**
* 
* 27.2.2
* COOLWSD version: 23.05.3.1 git hash:  a719c0f
* LOKit version: built on Collabora Office 23.05.3.1 git hash: 403de15
* Served by: Ubuntu 22.04.3 LTS 
* 9002c172

juliushaertl:
Relevant code part https://github.com/nextcloud/server/blob/73f3c0aef73dcb08d3e3ac8c6e0f670eb769a8d7/apps/files/src/views/FileReferencePickerElement.vue#L57-L73

small1:
I noice this in the text app. If we pick form a group folder we get a broken link

susnux:
> the dialog is missing thumbnails for me

Fixed with #42057

PR #42437, opened 2023-12-21
fix(files): Fix reference picker not working

susnux:
This resolves https://github.com/nextcloud/server/issues/40866 for stable 27.

The nodes are an array of all selected nodes even if multiselect is false, so we need to use the first element.


apps/files/src/views/FileReferencePickerElement.vue
```diff
@@ -52,7 +52,7 @@ export default {
 				{
 					label: t('files', 'Choose'),
 					type: 'primary',
-					callback: (node) => this.submit(node.fileid),
+					callback: (nodes) => { console.warn('got', nodes); this.submit(nodes[0].fileid) },
```
juliushaertl, line 55:
```suggestion
					callback: (nodes) => { this.submit(nodes[0].fileid) },
```

or log at debug level but with a more clear message?

emoral435: Agreed, just change the log with a clearer message and then it LGTM :)

susnux: Done

juliusknorr (approved):

artonge (approved):

skjnldsv (approved):

AndyScherzinger (approved):

emoral435 (approved):

PR #27613, opened 2021-06-22
Check datadirectory owner, not config owner.

Dreamsorcerer:
Changes permission checks to use datadirectory, rather than config file.
Step 1 towards root-owned configs (#11075).

For root-owned configs, the only thing that needs to be writeable is the data directory. For non-root configs, this check shouldn't make any difference, it's very likely the same user is used for both config and data dir.

Currently, the user shouldn't see any difference in behaviour, as the check here would still disallow a root-owned config:
https://github.com/nextcloud/server/blob/master/lib/base.php#L247
Once the rest of the work to support root-owned configs is complete, then this check should be removed (`|| !$configFileWritable && \OCP\Util::needUpgrade()`).

PR #42449, opened 2023-12-22
fix(cron): Respect config_is_read_only config in in cron and console commands

helmo:
* Resolves: https://help.nextcloud.com/t/running-cron-with-read-only-config/177221

## Summary

I’d like to keep the config and application code read-only for the webserver as an extra security measure.
We have the `config_is_read_only` config boolean for that, but still occ and cron complain about not being the config file owner.

Even if they do have write access, which I think atleast cron should be able to do without.

## Checklist

### DIFF
--- a/console.php
+++ b/console.php
@@ -63,15 +63,17 @@ try {
 		echo "The posix extensions are required - see https://www.php.net/manual/en/book.posix.php" . PHP_EOL;
 		exit(1);
 	}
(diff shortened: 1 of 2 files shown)


helmo:
This has a merge conflict since https://github.com/nextcloud/server/pull/27613 was merged ... and I think that solved the same issue.

One minor detail remains, I've posted the string change in  #44094

PR #42497, opened 2023-12-27
Ignore shared Collective mounts when scanning files

### LINKED ISSUE #42496: [Bug]: Shared Collectives cause "Path not found" when scannig files with "occ files:scan --all"

### Bug description

When a Collective is shared with one or more users and you start a manual file scan using `occ files:scan --all`, this leads to the following error message for each user who is member of a Collective circle:

```
Path not found: /username
```

Where `username` is the name of the affected user.

### Steps to reproduce

1. Create a collective
2. Share it with at least one other user
3. Run `occ files:scan --all` in a terminal

### Expected behavior

`occ files:scan --all` runs without "Path not found" error messages

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

28

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

There is a workaround to fix this. In `/server/lib/private/Files/Utils/Scanner.php` the following code needs to be added, so Collectives mounts will be ignored during a file scan:

```
            // don't scan received local shares, these can be scanned when scanning the owner's storage
            if ($storage->instanceOfStorage(SharedStorage::class)
                || $storage->instanceOfStorage(CollectiveStorage::class)
            ) {
                continue;
            }
```

Also see https://help.nextcloud.com/t/occ-files-scan-outputs-path-not-found-for-some-users/177507/12 and https://github.com/nextcloud/server/blob/5125c34f011651821d3a1d6e3ce2d0337ab75fe9/lib/private/Files/Utils/Scanner.php#L237-L239

arnowelzel:
* Resolves: #42496

## Summary

When scanning files using `occ files:scan`, shared Collective mounts trigger "Path not found" error messages. This change will make sure that that Collective mounts will be ignored.

## Checklist

### DIFF
--- a/lib/private/Files/Utils/Scanner.php
+++ b/lib/private/Files/Utils/Scanner.php
@@ -239,7 +239,7 @@ class Scanner extends PublicEmitter {
 				}
 			}
 
-			// don't scan received local shares, these can be scanned when scanning the owner's storage
+			// don't scan received local shares or collectives, these can be scanned when scanning the owner's storage
 			if ($storage->instanceOfStorage(SharedStorage::class) || $storage->instanceOfStorage(CollectiveStorage::class)) {
 				continue;
 			}
--- a/lib/private/Files/Utils/Scanner.php
+++ b/lib/private/Files/Utils/Scanner.php
@@ -240,7 +240,7 @@ class Scanner extends PublicEmitter {
 			}
 
 			// don't scan received local shares or collectives, these can be scanned when scanning the owner's storage
-			if ($storage->instanceOfStorage(SharedStorage::class) || $storage->instanceOfStorage(CollectiveStorage::class)) {
+			if (substr($storage->getId(), 0, 6) !== 'home::') {
 				continue;
 			}
 
--- a/lib/private/Files/Utils/Scanner.php
+++ b/lib/private/Files/Utils/Scanner.php
@@ -37,6 +37,7 @@ use OC\Files\Storage\Home;
 use OC\ForbiddenException;
 use OC\Hooks\PublicEmitter;
 use OC\Lock\DBLockingProvider;
+use OCA\Collectives\Mount\CollectiveStorage;
 use OCA\Files_Sharing\SharedStorage;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Files\Events\BeforeFileScannedEvent;
@@ -239,9 +240,10 @@ class Scanner extends PublicEmitter {
 			}
 
 			// don't scan received local shares, these can be scanned when scanning the owner's storage
-			if ($storage->instanceOfStorage(SharedStorage::class)) {
+			if ($storage->instanceOfStorage(SharedStorage::class) || $storage->instanceOfStorage(CollectiveStorage::class)) {
 				continue;
 			}
+
 			$relativePath = $mount->getInternalPath($dir);
 			$scanner = $storage->getScanner();
 			$scanner->setUseTransactions(false);


lib/private/Files/Utils/Scanner.php
```diff
@@ -235,9 +236,10 @@ public function scan($dir = '', $recursive = \OC\Files\Cache\Scanner::SCAN_RECUR
 			}
 
 			// don't scan received local shares, these can be scanned when scanning the owner's storage
```
emoral435, line 238: If you have the time, it would be really appreciated to add the new collective check to the comment :) not needed though

emoral435 (approved):
Nice change, looks good to me :)

emoral435:
Fast review ~!

lib/private/Files/Utils/Scanner.php
arnowelzel: Done ;-)

artonge (approved):
Looks good, but let's wait for @max-nextcloud or @mejo- input :)

max-nextcloud:
@mejo- is the better person when it comes to backend changes.

I have some uninformed questions though...
* What will happen if collectives is not installed and the used class is not available?
  Will php be happy with that?
* You are talking about a `shared collective` - are you referring to a collective that has been explicitly shared by link or a normal collective with multiple members?

In general it seems not ideal to me to have collectives specific code in server. I wonder if we could instead make `CollectiveStorage` a `SharedStorage` as that could happen on the collectives side of things.

max-nextcloud:
I remember that collectives storage was designed closely to what groupfolders does. Does this issue not occur with groupfolders ? How is the groupfolders implementation different from collectives?

arnowelzel:
> @mejo- is the better person when it comes to backend changes.
> 
> I have some uninformed questions though...
> 
> * What will happen if collectives is not installed and the used class is not available?
>   Will php be happy with that?

I tested this case and had no problem without Collectives installed, since `instanceOfStorage()` will only check the class name and it should not matter, if the class actually exists.

But I agree it would be better not to rely on multiple classes here but instead have a common method which can be used. Maybe adding a method `isShared()` to the `Storage` interface may help. This sould return `true` by `SharedStorage` or `CollectiveStorage` and otherwise `false` so the scanner knows when to ignore the storage:

```
			// don't scan shared storages, these can be scanned when scanning the owner's storage
			if ($storage->isShared()) {
				continue;
			}
``` 

However this would involve substantial refactoring of the backend code.

> * You are talking about a `shared collective` - are you referring to a collective that has been explicitly shared by link or a normal collective with multiple members?

I refer to a normal collective which is shared with multiple members of a circle.

skjnldsv:
@max-nextcloud @mejo- any update?

lib/private/Files/Utils/Scanner.php
```diff
@@ -234,10 +235,11 @@ public function scan($dir = '', $recursive = \OC\Files\Cache\Scanner::SCAN_RECUR
 				}
 			}
 
-			// don't scan received local shares, these can be scanned when scanning the owner's storage
-			if ($storage->instanceOfStorage(SharedStorage::class)) {
+			// don't scan received local shares or collectives, these can be scanned when scanning the owner's storage
+			if ($storage->instanceOfStorage(SharedStorage::class) || $storage->instanceOfStorage(CollectiveStorage::class)) {
```
skjnldsv, line 239: I'm wondering if we don't need to check if the class exists before ? :thinking: 
@come-nc @icewind1991

skjnldsv:
@arnowelzel psalm said no

lib/private/Files/Utils/Scanner.php
```diff
@@ -238,10 +239,11 @@
...
```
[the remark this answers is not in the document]
arnowelzel: Strange - it works without problems for me. Maybe then the more thorough refactoring implementing an additional interface method as suggested, is a better. I'll have a look into this.

arnowelzel:
Since using the class name of the Collective storage is not possible in any way without triggering a psalm error I changed the test as following:

```php
			// don't scan received local shares or collectives, these can be scanned when scanning the owner's storage
			if (substr($storage->getId(), 0, 6) !== 'home::') {
				continue;
			}
```

The idea here is, that only home storages should get scanned. Recieved shared storage start with `shared::` and Collectives, which trigger the error, report IDs as `local::` followed by the physical path. However I don't know if any other storages which should indeed be scanned, also report an ID starting with something different from `home::`.

I also noticed the following code in `\OCA\Collectives\Mount\CollectiveStorage` - maybe in the past this problem was mitigated by using `NoopScanner` which does not exist any longer?

```php
	public function getScanner($path = '', $storage = null): Scanner {
		if (!$storage) {
			$storage = $this;
		}
		if ($storage->instanceOfStorage(ObjectStoreStorage::class)) {
			// NoopScanner is private API and kept here for compatibility with older releases
			$storage->scanner = class_exists(NoopScanner::class) ? new NoopScanner($storage) : new ObjectStoreScanner($storage);
		} elseif (!isset($storage->scanner)) {
			$storage->scanner = new Scanner($storage);
		}
		return $storage->scanner;
	}
```

mejo-:
Probably @icewind1991 is the best to decide whether it's a good idea to limit scanning to `home` storages.

lib/private/Files/Utils/Scanner.php
```diff
@@ -238,10 +239,11 @@ public function scan($dir = '', $recursive = \OC\Files\Cache\Scanner::SCAN_RECUR
 				}
 			}
 
-			// don't scan received local shares, these can be scanned when scanning the owner's storage
-			if ($storage->instanceOfStorage(SharedStorage::class)) {
+			// don't scan received local shares or collectives, these can be scanned when scanning the owner's storage
+			if (substr($storage->getId(), 0, 6) !== 'home::') {
```
icewind1991, line 243: This will break scanning of external storages.

icewind1991 (changes requested):

icewind1991:
What is stopping the collectives from being scanned correctly in the first place?

lib/private/Files/Utils/Scanner.php
arnowelzel: What ID do external storages have?

arnowelzel:
> What is stopping the collectives from being scanned correctly in the first place?

They are shared storages like `SharedStorage` and thus don't provide a physical path which can be scanned.

lib/private/Files/Utils/Scanner.php
icewind1991: they all have different prefixes, in general attempts to filter by storage id is imprecise at best.

icewind1991:
If they can't be scanned at all then having them use ` NoopScanner` is probably the best (shared storage should also move to that really)

arnowelzel:
> If they can't be scanned at all then having them use ` NoopScanner` is probably the best (shared storage should also move to that really)

Well - `NoopScanner` does not exist any longer.

icewind1991:
Right, it got moved to `ObjectStoreScanner` when it got slightly less noop, should be good to still use that one or just create your own

arnowelzel:
> Right, it got moved to `ObjectStoreScanner` when it got slightly less noop, should be good to still use that one or just create your own

It seems, this is what Collectives is using right now - but it does not work. See https://github.com/nextcloud/collectives/blob/main/lib/Mount/CollectiveStorage.php#L62-L74:

```php
public function getScanner($path = '', $storage = null): Scanner {
		if (!$storage) {
			$storage = $this;
		}
		if ($storage->instanceOfStorage(ObjectStoreStorage::class)) {
			// NoopScanner is private API and kept here for compatibility with older releases
			$storage->scanner = class_exists(NoopScanner::class) ? new NoopScanner($storage) : new ObjectStoreScanner($storage);
		} elseif (!isset($storage->scanner)) {
			$storage->scanner = new Scanner($storage);
		}
		return $storage->scanner;
	}
```

arnowelzel:
Since no one has an idea about this I suggest to close this. The code is way outdated anyway.

PR #42502, opened 2023-12-28
Fix sharing icon

### LINKED ISSUE #40202: adjust icon of shared state in broadcrumb
> Shared
> image

I am not sure about this icon. Shouldn't we better use a different icon, e.g. this one?
image
https://fonts.google.com/icons?icon.style=Outlined&icon.set=Material+Symbols&selected=Material+Symbols+Outlined:folder_shared:FILL@0;wght@400;GRAD@0;opsz@24&icon.query=shar

_Originally posted by @szaimen in https://github.com/nextcloud/server/issues/40192#issuecomment-1702890351_

cc @nextcloud/designers

marcoambrosini:
* Resolves: #40202

## shared state
Screenshot from 2023-12-28 14-13-36

## unshared state
Screenshot from 2023-12-28 14-13-59

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/files/src/views/FilesList.vue
+++ b/apps/files/src/views/FilesList.vue
@@ -34,8 +34,8 @@
 						type="tertiary"
 						@click="openSharingSidebar">
 						<template #icon>
-							<LinkIcon v-if="shareButtonType === Type.SHARE_TYPE_LINK" />
-							<ShareVariantIcon v-else :size="20" />
+							<FolderAccountIcon v-if="shareButtonType !== null" />
+							<AccountPlusIcon v-else :size="20" />
 						</template>
 					</NcButton>
 
@@ -135,7 +135,7 @@ import { UploadPicker } from '@nextcloud/upload'
 import { loadState } from '@nextcloud/initial-state'
 import { defineComponent } from 'vue'
 
-import LinkIcon from 'vue-material-design-icons/Link.vue'
+import FolderAccountIcon from 'vue-material-design-icons/FolderAccount.vue'
 import ListViewIcon from 'vue-material-design-icons/FormatListBulletedSquare.vue'
 import NcAppContent from '@nextcloud/vue/dist/Components/NcAppContent.js'
 import NcButton from '@nextcloud/vue/dist/Components/NcButton.js'
@@ -143,7 +143,7 @@ import NcEmptyContent from '@nextcloud/vue/dist/Components/NcEmptyContent.js'
 import NcIconSvgWrapper from '@nextcloud/vue/dist/Components/NcIconSvgWrapper.js'
 import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 import PlusIcon from 'vue-material-design-icons/Plus.vue'
-import ShareVariantIcon from 'vue-material-design-icons/ShareVariant.vue'
+import AccountPlusIcon from 'vue-material-design-icons/AccountPlus.vue'
 import ViewGridIcon from 'vue-material-design-icons/ViewGrid.vue'
 
 import { action as sidebarAction } from '../actions/sidebarAction.ts'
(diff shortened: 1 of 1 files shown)


skjnldsv:
Seems like the first icon on your screenshot isn't what is being discussed on the mentioned issue: #40202
I fear we don't use that icon anywhere else :)
image

apps/files/src/views/FilesList.vue
```diff
@@ -34,8 +34,8 @@
 						type="tertiary"
 						@click="openSharingSidebar">
 						<template #icon>
-							<LinkIcon v-if="shareButtonType === Type.SHARE_TYPE_LINK" />
-							<ShareVariantIcon v-else :size="20" />
+							<FolderAccountIcon v-if="shareButtonType !== null" />
```
skjnldsv, line 37: @jancborchardt shall we lose the share link icon too?

para28711:
Stephen f Balogh

marcoambrosini:
> Seems like the first icon on your screenshot isn't what is being discussed on the mentioned issue: https://github.com/nextcloud/server/issues/40202

I see now that @jancborchardt refers to the folder with the link in it. I think we should use the folder-share material design icon more appropriate and I think that we should use that one in the files list too.

apps/files/src/views/FilesList.vue
jancborchardt: If it’s shared by link, that should be reflected?

jancborchardt:
@marcoambrosini @skjnldsv let’s indeed focus on the issue at hand here to prevent creating deadlocks like these. :)

This icon https://fonts.google.com/icons?selected=Material%20Icons%3Aperson_add_alt_1%3A communicates both "Share" and "Add to share", so it can be used in both unshared and shared state, which we currently do not separate in the breadcrumbs either.

apps/files/src/views/FilesList.vue
skjnldsv: I agree, but this PR removes that

apps/files/src/views/FilesList.vue
```diff
@@ -34,8 +34,8 @@
 						type="tertiary"
 						@click="openSharingSidebar">
 						<template #icon>
-							<LinkIcon v-if="shareButtonType === Type.SHARE_TYPE_LINK" />
```
szaimen, line 37: this should probably be restored?

apps/files/src/views/FilesList.vue
marcoambrosini: @jancborchardt we said in yesterday's design review that we're not going to use this icon in iOS in favor of the native one. I think that in line with that, we should use the material folder-shared icon that @szaimen proposed in the issue both in the web interface and in the android apps.

marcoambrosini:
> let’s indeed focus on the issue at hand 

The issue at hand proposes the usage of this new icon @jancborchardt 

> which we currently do not separate in the breadcrumbs either.

We currently have separate states in the breadcrumbs icon, just with a worse logic. A different icon is shown only if the folder is shared by link. No distinction is made between unshared and shared without link. 
I think it's just better to distinguish between shared and not shared.
Ref https://github.com/nextcloud/server/pull/40192

jancborchardt:
So as @skjnldsv said, we don’t use the proposed icon anywhere. What we do use is these ones:
image
image

The core issue is mainly that in the breadcrumbs we still use the "triangle" share icon, and that needs to be replaced. If we discuss more than that then it will of course lead to unnecessary discussion and block a simple fix as this → so let’s keep the matters separate.

apps/files/src/views/FilesList.vue
skjnldsv: https://github.com/nextcloud/server/pull/42502#discussion_r1437444971

jancborchardt:
Moving assignment from @marcoambrosini to @szaimen as discussed in the design team call. :)

szaimen:
Superceded by https://github.com/nextcloud/server/pull/43433

PR #42507, opened 2023-12-28
Fix/files/42460/breadcrumb conditional rendering

### LINKED ISSUE #42460: [BITV]: remove all native tooltips from breadcrumbs
I think it would be better to remove native tooltips from breadcrumbs (all `title` attrs) because then we will face with a problem again: tooltips are there but there is no other way to discover this information besides mouse over.

Screenshot from 2023-12-22 16-53-30

emoral435:
## ☑️ Resolves #42460

🖼️ Screenshots

🏡 After
firefox_7h4YBek9ip

🚧 Summary
Removed the title attribute from all breadcrumb headers from all but the first and last breadcrumbs - the first having an aria label, and the last being differently titled than the rest

## Checklist


JuliaKirschenheuter (dismissed):
good but not completely right 

The last item should not have any `title`, it is just the same item as previous one

ShGKme (approved):

skjnldsv (changes requested):
https://github.com/nextcloud/server/issues/42460#issuecomment-1874951161

emoral435:
Will wait to see what we choose to do going forward before making any more changes :)

skjnldsv (approved):

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -30,7 +30,8 @@
 			v-bind="section"
 			dir="auto"
 			:to="section.to"
-			:title="titleForSection(section)"
+			:title="(index === 0 || index === sections.length - 1) ? titleForSection(section) : null"
+			:aria-description="(index === sections.length - 1) ? 'Reload current directory' : null"
```
ShGKme, line 34: This string is not translated

ShGKme (changes requested):

emoral435:
Cleaned up the component a bit, I was making it a bit sloppier than it needed to be, to be honest!

ShGKme (approved):

emoral435:
/ backport 255441980538bf46b7307a2aac257c4cd3c8a266 to stable28

PR #42513, opened 2023-12-28
enh(files): made breadcrumb component lang=ts

### LINKED ISSUE #42508: Enh(files): make Breacrumbs.vue's script TS and add copyright header
As of now, I just realized that the breadcrumb component we use is not yet using lang="ts", and there is no header - will add this myself!

emoral435:
## 🚧 Summary
Changes the language for the Files Breadcrumb component TS - there may be more that needs to be added for better  typing functionality, but this is just to get the ball rolling!


skjnldsv (approved):

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -63,12 +87,12 @@ export default Vue.extend({
 			return this.$navigation.active
 		},
 
-		dirs() {
-			const cumulativePath = (acc) => (value) => (acc += `${value}/`)
+		dirs(): Array<string> {
```
susnux, line 90:
```suggestion
		dirs(): string[] {
```

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -86,19 +110,19 @@ export default Vue.extend({
...
-		getNodeFromId(id) {
+		getNodeFromId(id: number): Node | undefined {
 			return this.filesStore.getNode(id)
 		},
-		getFileIdFromPath(path) {
+		getFileIdFromPath(path: string): number | undefined {
 			return this.pathsStore.getPath(this.currentView?.id, path)
 		},
-		getDirDisplayName(path) {
+		getDirDisplayName(path: string): string | TypeError {
```
susnux, line 119:
```suggestion
		getDirDisplayName(path: string): string {
```

There is no TypeError returned but only strings. TypeError might be thrown.

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -86,19 +110,19 @@ export default Vue.extend({
...
 			if (path === '/') {
 				return t('files', 'Home')
 			}
 
-			const fileId = this.getFileIdFromPath(path)
-			const node = this.getNodeFromId(fileId)
+			const fileId: number | undefined = this.getFileIdFromPath(path)
```
susnux, line 124: You assert in the next line that the fileid is not undefined so you are sure this returns a valid number.
So I would say you should then assert on this line or be defensive on the next line.

emoral435: I decided to be defensive on the next line. The changes in the updated commit is

const fileId: number | undefined = this.getFileIdFromPath(path)

const node: Node | undefined = (fileId) ? this.getNodeFromId(fileId) : undefined

Using getNodeFromId() will just access an array from the filestore, so accessing array[undefined] will throw undefined anyways, but its good to be declarative with the ternary operator. If you believe there is a better, more intuitive way to display this LMK!! :)

emoral435:
Cherry-picked the wrong commit (sorry Julia) by accident! Should be all good now

skjnldsv (approved):

Pytal (approved):

PR #42516, opened 2023-12-29
Update it.json

SMH17:
Improved localization


juliushaertl:
Thanks for your contribution. Nextcloud’s translation system is powered by [Transifex](https://www.transifex.com/nextcloud/). To start translating sign up and enter a group. If your community app should be translated by the [Nextcloud community on Transifex](https://www.transifex.com/nextcloud/nextcloud/dashboard/) just follow the setup section below.

We cannot merge those changes directly because they would be reverted by the state of transifex on the next sync, so please contribute those to the translation system directly.

https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#adding-translations

PR #42534, opened 2023-12-30
feat: generate avatars for display name with chinese characters

kesselb:
* Resolves: https://help.nextcloud.com/t/the-problem-of-automatically-generating-an-avatar-for-a-chinese-user-name/177750

## Summary

We are using NotoSans-Regular for generating avatars. 

If the character is not supported by NotoSans-Regular the users end up with an avatar like below:

image

The pull request here adds another font and uses a poor regex to flip out the font when chinese characters are detected. The font actually should cover even more characters. 

@danxuliu did a deep dive into the topic a while ago and wrote down everything you need to know: https://github.com/nextcloud/server/pull/25529

## TODO

- [ ] Testing Testing
- [ ] Do we want to ship another font
- [ ] Extend regex to match supported characters for NotoSC font

## Checklist

### DIFF
--- a/lib/private/Avatar/Avatar.php
+++ b/lib/private/Avatar/Avatar.php
@@ -100,19 +100,7 @@ abstract class Avatar implements IAvatar {
 		return $avatar;
 	}
 
-	/**
-	 * {size} = 500
-	 * {fill} = hex color to fill
-	 * {letter} = Letter to display
-	 *
-	 * Generate SVG avatar
-	 *
-	 * @param int $size The requested image size in pixel
-	 * @return string
-	 *
-	 */
-	protected function getAvatarVector(int $size, bool $darkTheme): string {
-		$userDisplayName = $this->getDisplayName();
+	protected function getAvatarVector(string $userDisplayName, int $size, bool $darkTheme): string {
 		$fgRGB = $this->avatarBackgroundColor($userDisplayName);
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@kesselb can you tell a bit about the state of the PR? 😊

Phreeman33:
Tested and worked fine.

As traditional Chinese users, we've been overriding fonts with Noto Sans TC to display Chinese characters properly. Here's the catch: Noto Sans SC uses some Simplified Chinese glyphs (probably GB18030 standard), but the same character actually has different strokes in:

1. Traditional Chinese (Taiwan)
2. Traditional Chinese (Hong Kong/Macau)
3. Japanese
4. Korean

e.g.:  for `\u89d2`
In TC:
角-TC
In SC:
角-SC

I recommend we just merge this commint. Since:

1. This PR already fixes avatar display for most cases in Chinese characters.
2. The glyph differences only affect a small number of characters.

Maybe we could handle the glyph differences with languange/location setting for CJK fonts matching later.

Trivia: Lots of websites/devices secretly use GB18030 glyphs for all Chinese text. The stroke differences are super subtle (mostly like 1-2 pen strokes!), which is why most traditional Chinese users never even notice! 😉

kesselb:
Hi @Phreeman33,

Many thanks for your feedback—it's much appreciated! 🙏

Thanks for taking the initiative and sending a reworked version of the PR to improve the avatar situation for users with Chinese characters.

skjnldsv:
Overrules by https://github.com/nextcloud/server/pull/51855

PR #42456, opened 2023-12-22
fix(theming): adjust default `--color-background-plain` to new background

ShGKme:
* Continue of https://github.com/nextcloud/server/pull/42424

Before `--color-background-plain` was the same as the default background image color. Now the image has changed.

It is noticeable on slow connection.

On the image for me color is `#01679d` while we use as primary `#00679e`, but 1/256 is not notable.

Probably there are more places that we need to adjust.

## Screenshots

Before | After
---|---
bg-before | bg-after

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #42552, opened 2024-01-02
fix(theming): change color button contrast

### LINKED ISSUE #41849: [BITV] 9.1.4.3/7.1 - The label of the "Change color" button in the "Background" section does not meet the contrast requirements for text - it is 4,16:1. (1)
https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-6fe46b78da-v7-n1

ShGKme:
* Resolves: #41849

## Summary

Use `NcButton` as a trigger for changing color to make the text always contrast. 

Note: there is no pair of text colors that is contrast enough on any background. E.g. in some gray of blue backgrounds, both white and black text is not contrast enough. So we desided to use fixed background for the text.

## TODO

- [x] **User Settings:** Use `NcButton` on the color preview instead of plain text to ensure it is well-visible
- [x] **Admin Settings:** Separate `NcButton` and the color preview
- [x] Adjust tests and use more semantic selectors

## Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
image | image

### Other visual changes

Popover is now shown on the button

image

There is no interactivity on the colored rectangle. Making it interactive would need to get rid of using `NcButton` and copy its styles.

change-color-hover

## Checklist


apps/theming/src/components/BackgroundSettings.vue
```diff
@@ -368,11 +380,6 @@ export default {
 			border-color: var(--color-border);
 		}
 
-		&__color {
-			color: var(--color-primary-text);
```
szaimen, line 372: I was just wondering why the variable does not work? In prinviple it should be automatically correctly calculated or is the calculation not working?

ShGKme: Because this element may have any background

> Note: there is no pair of text colors that is contrast enough on any background. E.g. in some gray of blue backgrounds, both white and black text is not contrast enough. So we decided to use fixed background for the text.

But maybe the primary issue is the default `--color-primary` is not the same as the default `--color-primary-element`...

ShGKme: I'm a little confused with our new colors...

szaimen: I think `--color-primary-text` should be used with `--color-primary`...

jancborchardt (approved):
Nice and simple solution, looks good design-wise.

apps/theming/src/components/BackgroundSettings.vue
jancborchardt: > Because this element may have any background
> 
> > Note: there is no pair of text colors that is contrast enough on any background. E.g. in some gray of blue backgrounds, both white and black text is not contrast enough. So we decided to use fixed background for the text.
> 
> But maybe the primary issue is that the default `--color-primary` is not the same as the default `--color-primary-element`...

Actually, this was/is only the case with our default blue, right? And since we now had to change the background image for accessibility, we could deprecate `--color-primary-element` in favor of `--color-primary` since the primary color always needs to have enough contrast.

Or am I missing something? Also cc @juliushaertl @nextcloud/designers (and @susnux @JuliaKirschenheuter since you switched out the background)

juliushaertl: From my perspective `-color-primary-element` is still needed for cases where we want elements to stick out from the background color, we cannot ensure that just with primary as the admin might set e.g. to white.

ShGKme: I feel confused.

Is there any case for usage of `--color-primary` except page's background color and places where it is wrongly used instead of `--color-primary-element`?

ShGKme: > Actually, this was/is only the case with our default blue, right?

Practically — yes.

Theoretically — no, because you can choose any possible color =D

ShGKme: > since we now had to change the background image

By the way. Does it mean we don't use Nextcloud Blue color on UI anymore?

marcoambrosini (approved):

ShGKme:
Seems that this change is actually required for a11y only if there is still **Nextcloud Blue** in the palette. Which we don't use anywhere anymore as a primary color...

@nextcloud/designers do you think this is fine to merge regardless of a11y requirements, or better to avoid it, if we decide to remove the problematic color from the palette?

szaimen:
I think it is fine to merge it. This should probably also solve cases with custom colors...

apps/theming/src/components/BackgroundSettings.vue
susnux: > By the way. Does it mean we don't use Nextcloud Blue color on UI anymore?

It is still used if you remove the default background image -> in this case you again get accessibility issues with the app names in the main navigation because the primary-text color was forced to white for design but white on default blue is only 4.1:1 contrast. Thus it would need to be black font color for default black.

But to prevent that design change we darkened the default background image so white text keep working.

ShGKme: > except for the main background if no image was selected

> It is still used if you remove the default background image -> in this case you again get accessibility issues with the app names in the main navigation because the primary-text color was forced to white for design but white on default blue is only 4.1:1 contrast. Thus it would need to be black font color for default black.

> But to prevent that design change we darkened the default background image so white text keep working

This is also what I wanted to change in [this PR](https://github.com/nextcloud/server/pull/42456)... If the background image is not loaded (or removed), the background should be new dark blue, not an old Nextcloud Blue which brings an a11y issue.

May we change the default `--color-primary` to the new dark blue?

> `--color-primary` is the color the user selected. `--color-primary-element` is that color adjusted to be accessible.

But it is not automatically adjusted, right? Only predefined.

ShGKme:
Rebased onto main, fixed Cypress e2e tests.

apps/theming/src/components/BackgroundSettings.vue
susnux: > But it is not automatically adjusted, right? Only predefined.

It is `--color-primary-element` is calculated automatically based on the background color and the primary color.

susnux: So we can deprecate and move `--color-primary-element` to `--color-primary` and use `--color-primary-default` or similar for the originally selected user color.

ShGKme: So, you don't think we should change the default palette in the picker? @susnux

szaimen: > So we can deprecate and move `--color-primary-element` to `--color-primary` 

I wouldnt do this. I would rather deprecate and remove `--color-primary` in the long run...

ShGKme:
I forgot that we have the same button with the same issue also on Admin settings.

I separated the button and color view and added icon to be inline with other buttons here.
Re-requesting designers to re-approve. Sorry for the double work 👉👈

szaimen (approved):
LGTM from design perspective :)

susnux (approved):

susnux:
@ShGKme can you resolve the conflicts?

Pytal (approved):

ShGKme:
> @ShGKme can you resolve the conflicts?

There is a Cypress issue that I haven't resolved yet.

cypress/e2e/theming/themingUtils.ts
```diff
@@ -71,16 +71,18 @@ export const validateUserThemingDefaultCss = function(expectedColor = defaultPri
...
+	// Click on random color
+	cy.get('.color-picker__simple-color-circle').eq(randColour).click()
 
-	// Return selected colour
-	return cy.get(pickerSelector).get('.color-picker__simple-color-circle').eq(randColour).then(($el) => {
-		$el.trigger('click')
-		return $el.css('background-color')
-	})
+	// Get the selected color from the color preview block
+	return cy.get(colorPreviewSelector).then(($el) => $el.css('background-color'))
```
susnux, line 87: This will flaky return the wrong color, because the background color is updated on the next tick and this selector might be faster than vue.

Instead use the backgroudn color of `.color-picker__simple-color-circle`

susnux: Alternative would be to wait for the network request

ShGKme: Hmm.. Would `cy.wait` also be a solution here?

susnux: I would not wait for arbitrary time here. Either cy.wait for some webrequest or simply take the color from the picker. You never know the time needed for all requests etc.

susnux: Or a wait for color change

ShGKme: I meant not timeout but waitUntil

susnux: Well thats what I changed in the push ;-)

PR #42558, opened 2024-01-03
use deletion time instead of mtime in trashbin handling

icewind1991:
The mtime of the deleted file and the deletion time are the same with the standard trashbin backend, but this isn't necessarily the case for other backends.


icewind1991:
https://github.com/nextcloud/groupfolders/pull/2714 works around this for the groupfolder trash backend

Pytal (approved):

Altahrim (approved):

Issue #41699, opened 2023-11-23
[BITV]: change styles for focused items in sidebar


JuliaKirschenheuter:
For now it is not really possible to recognize that current item is focused ;(

![Screenshot from 2023-11-23 16-45-46](https://github.com/nextcloud/server/assets/6078378/2e541d79-fe50-4819-868b-0a017d010c3b)
![Screenshot from 2023-11-23 16-45-23](https://github.com/nextcloud/server/assets/6078378/63a778c5-2b36-4641-97b6-966caf80bc26)

@jancborchardt could we do it like in settings? 

![Screenshot from 2023-11-23 16-48-14](https://github.com/nextcloud/server/assets/6078378/6d0820d9-a3d3-4426-a14b-696f6c1cee8c)

szaimen:
It looks like that currently so I guess it is indeed missing a dependency update in server and photos?
![image](https://github.com/nextcloud/server/assets/42591237/1f3eef17-b48b-45cd-83db-e488f572cb30)
![image](https://github.com/nextcloud/server/assets/42591237/654ae6a5-9903-417c-8c5a-fda19eaa8026)

szaimen:
In activity and files dependency update seems to be still misssing

PR #42559, opened 2024-01-03
Remove unneeded styles: everything is handled by NcVue

JuliaKirschenheuter:
* For https://github.com/nextcloud/server/issues/41699
* Blocked by https://github.com/nextcloud-libraries/nextcloud-vue/pull/5022

## Summary

Vue-navigation | Unchanged non-vue navigation
---|---
Screenshot from 2024-01-03 15-31-17 | Screenshot from 2024-01-03 15-31-12

## Checklist


ShGKme:
This PR changes styles for non-`@nextcloud-vue` navigation while linked PR in nextcloud-vue changes `@nextcloud/vue` navigation. Are they related?

emoral435 (approved):
I checked this on my dev env and saw that removing this did not cause anything big to stylistically change - it looks good to me. 

I do not know if there are any other apps that somehow depend on this, but to my understanding this only lets the elements that are for app navigation within a page have their border color be the primary color. When the changes to the `NcAppNavigationItem.scss` gets merged, this will be all good 👍

Pytal (approved):

ShGKme:
> This PR changes styles for non-`@nextcloud-vue` navigation while linked PR in nextcloud-vue changes `@nextcloud/vue` navigation. Are they related?

So, these styles are not supposed to affect `@nextcloud/vue/NcAppNavigation`. But it does, because the layout of the `NcAppNavigation` was changed and the selector here is not actual anymore.

So the best could be to fix the selector.

but probably these styles are not actual anymore. Need to check if we still have any non-NcAppNavigation navigation.

ShGKme:
I haven't found non-vue navigation anywhere. Should be save to remove.

ShGKme (approved):

PR #42457, opened 2023-12-22
fix(files): Allow Ctrl/Command click on file entries to open in a new tab

juliusknorr:
Bring back opening files in a new tab by Ctrl-clicking on the file list entry. metaKey is needed for macOS support.

PR #42566, opened 2024-01-03
fix(files): Fix unresponsive file click to download

Pytal:
## Summary

With https://github.com/nextcloud/server/pull/42457 a regression is introduced due to the `event.preventDefault()` which prevents link activation

To preserve ctrl-click functionality and fix the non-triggering download we need to set the download action as default so it executes

## Checklist


juliusknorr (approved):

Pytal:
Pushed simpler alternative fixups which moves the `event.preventDefault()` instead so download does not depend on the file action as before, what do you think @skjnldsv?

skjnldsv (approved):

skjnldsv:
> what do you think @skjnldsv?

Code make sense :)

PR #42570, opened 2024-01-03
enh(settings): removed background-gradient to ensure contrast ratio a11y requirements

### LINKED ISSUE #42440: [BITV] Preview card in personal settings might not satisfy contrast requirements
The preview card in the personal settings uses `primary-element` as background color but overlays it with a `primary-gradient` which is using `primary-color` and then uses `primary-element-text` for the username.
But this combination is not always matching meaning either `primary-element-*` **or** `primary` should be used. See example where black is used but does not provide enough contrast:

Screenshot 2023-12-22 at 01-35-12 Persönliche Informationen - Persönliche Einstellungen - Nextcloud

emoral435:
## 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
image|image

Extra SS to check a random colors color contrast ratio
Colour_Contrast_Analyser_xNREkpPDYl

## 🚧 Summary
Removed the background gradient on the profile preview section of the settings to increase accessibility

## Checklist


Pytal (approved):

nimishavijay (commented):
Nice! Looks good, additionally: we don't use `#000000` for text, we only use `#222222`, so that should be the color of the text and the color that the contrast is checked against :)

JuliaKirschenheuter (approved):
Looks good!

> Extra SS to check a random colors color contrast ratio

we don't need to pay attention on it, please test everything only on default background without changing color. We certificate only default background and everything else is not relevant for us ;)

PR #42584, opened 2024-01-04
fix(files): fix unshare action and merge duplicate shares

### LINKED ISSUE #14437: Share overview should allow to unshare
In the right hand sidebar I have the "Shared with other", "Shared with you" and "Shared by link" views.

In "Shared with you" the '...' menu of each file/folder shows a unshare action which performs a unshare from self operation.

For "Shared with others" and "Shared by link" the '...' menu only has a delete action which deletes the file completely. I think, like for "Shared with you" the delete action should be replaced with a unshare actions which allows me to unshare the file again from all recipients as a easy way to cleanup old shares. Having always to open the details few is really cumbersome.

skjnldsv:
## Fix unshare action
The delete action now properly adjust itself depending on the selectcion
  - `Delete`
  - `Unshare`
  - `Delete and unshare` if shares and files are mixed in the selection

cc @nextcloud/designers for wording validation
image

Fix https://github.com/nextcloud/server/issues/14437

## Merge duplicates shares in the views

The sharing views will group the same nodes coming from the same source :warning: 
This can happen if a user share a folder with different groups and you're a member of them. The same file will appear multiple times in the `Shared with you` section.
We ONLY MERGE entries with the same location on the user file system, which means, if you have two different shares but they are in different folders (moved) and therefore are matching different shares, we will NOT group them, and their actions will both target different shares.

image

cc @nextcloud/designers for wording validation and icon choice


apps/files_sharing/src/services/SharingService.ts
```diff
@@ -144,6 +148,13 @@ const getDeletedShares = function(): AxiosPromise<OCSResponse<any>> {
 	})
 }
 
+const groupBy = function(xs: (Folder | File)[], key: string) {
+	return Object.values(xs.reduce(function(rv, x) {
+		(rv[x[key]] = rv[x[key]] || []).push(x)
+		return rv
+	}, {})) as (Folder | File)[][]
+}
+
```
artonge, line 157: :robot: 
Hard to read, a comment would help, a rewriting with clearer variable names and multiple steps would be more appreciated :).

apps/files_sharing/src/services/SharingService.ts
```diff
@@ -162,12 +173,21 @@ export const getContents = async (sharedWithYou = true, sharedWithOthers = true,
...
 
 	if (filterTypes.length > 0) {
 		contents = contents.filter((node) => filterTypes.includes(node.attributes?.share_type))
 	}
 
+	// Merge duplicate shares and group their attributes
+	// Also check the sharingStatusAction.ts code
+	contents = groupBy(contents, 'source').map((nodes) => {
+		const node = nodes[0]
+		node.attributes['share-types'] = nodes.map(node => node.attributes['share-types'])
```
artonge, line 187: Unsure about that one, we mess with the content return by the server, which is unexpected misleading.
Shouldn't we use another property?

apps/files/src/actions/deleteAction.ts
```diff
@@ -24,17 +24,48 @@ import { Permission, Node, View, FileAction } from '@nextcloud/files'
...
 
 import logger from '../logger.js'
+import { getCurrentUser } from '@nextcloud/auth'
+
+// A file which the owner is NOT the current user
+// is considered shared. Other methods like group folders
+// will always use the current user as the owner.
+// It will therefore not be considered shared.
+const isAllUnshare = (nodes: Node[]) => {
+	return !nodes.some(node => node.owner === getCurrentUser()?.uid)
```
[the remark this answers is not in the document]
skjnldsv: Seems safer to detect different owner as thi sis what proves this is shared to you, no matter what mount type it is :)

apps/files_sharing/src/services/SharingService.ts
skjnldsv: I'll add more comments, you're right :)

apps/files_sharing/src/services/SharingService.ts
skjnldsv: @artonge the sharing API already messes with the content returned by server Properties are completely different and it even returns duplicates nodes
This is the simplest way to get closer to our single point of truth: webdav :)

It was already kind of designed like this, as I've seen `share-types` returned as an empty array sometimes
This is why we are using `Object.values(node?.attributes?.['share-types'] || {}).flat() as number[]` as [a safety measure since a while ago](https://github.com/nextcloud/server/blob/a360f4868887748b586ad4346ff4075fb1dfcff0/apps/files_sharing/src/actions/sharingStatusAction.ts#L49)

artonge (approved):

skjnldsv:
Found a bug in https://github.com/nextcloud-libraries/nextcloud-files: https://github.com/nextcloud-libraries/nextcloud-files/pull/862

nimishavijay:
> Delete and unshare if shares and files are mixed in the selection

Would all the files be deleted? or would the not-shared files be deleted and shared files be unshared? I'm wondering if it makes sense to just show `Delete` and `Unshare` as 2 separate actions if the files are mixed. ccing @jancborchardt because he was involved in the issue :)

> Merge duplicates shares in the views
> Shared multiple times with different users

We are moving away from "users" in the wording. So `Shared multiple times with different people` or `Shared multiple times with different accounts` (no strong preference there, if I had to pick I'd go with "people")

szaimen (approved):
LGTM from design perspective but didnt test

skjnldsv:
Failure not related

PR #42597, opened 2024-01-05
perf: Avoid updating the folder size if we know the size difference

juliusknorr:
When using object storage the scanner does not return any results so `correctFolderSize` was always called even for operations where we don't need to update the parent sizes like `mkdir`.

This PR introduces a shortcut to skip this update if we know the size difference for sure so it can be propagated through https://github.com/nextcloud/server/blob/7db5bcc40920f45337056b86b24207a94a7c6e61/lib/private/Files/Cache/Updater.php#L144

### DIFF
--- a/lib/private/Files/Cache/Updater.php
+++ b/lib/private/Files/Cache/Updater.php
@@ -119,7 +119,7 @@ class Updater implements IUpdater {
 	 * @param string $path
 	 * @param int $time
 	 */
-	public function update($path, $time = null) {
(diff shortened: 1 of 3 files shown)


juliushaertl:
Tests passing except for unrelated failure that also happens on master

Altahrim (approved):

juliushaertl:
Failures seem consistent with this change though I have no idea why those would occur. Will look into

come-nc (approved):

PR #42607, opened 2024-01-06
fix(session): Avoid useless authtoken DB queries for anonymous requests

### LINKED ISSUE #42589: [Bug]: Two useless authtoken database queries for every anonymous request

### Bug description

Reading query logs I noticed

```
 primary SELECT * FROM `oc_appconfig`
 replica SELECT * FROM `oc_authtoken` WHERE (`token` = :dcValue1) AND (`version` = :dcValue2)
 replica SELECT * FROM `oc_authtoken` WHERE (`token` = :dcValue1) AND (`version` = :dcValue2)
```

for every anonymous request.

The problem is that `\OC\User\Session::tryTokenLogin` tries to find a token for the current PHP session. `\OC\Authentication\Token\PublicKeyTokenProvider::getToken` does up to two lookups when the instance has an instance secret set. **There will never be a hit**. The only exception would be a hash collision of the new session ID and a previous one.

The solution would be to check if the request had sent a cookie with the instance id as name. Those are used for the PHP session. If there is no cookie, this is a new session, and there won't be a token.

### Steps to reproduce

1.  ``curl https://localhost/login``

### Expected behavior

```
 primary SELECT * FROM `oc_appconfig`
```

### Installation method

None

### Nextcloud Server version

26

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

solracsf:
* Resolves: https://github.com/nextcloud/server/issues/42589

## Checklist

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -842,13 +842,16 @@ class Session implements IUserSession, Emitter {
 		$authHeader = $request->getHeader('Authorization');
 		if (str_starts_with($authHeader, 'Bearer ')) {
 			$token = substr($authHeader, 7);
-		} else {
-			// No auth header, let's try session id
+		} elseif ($request->getCookie($this->config->getSystemValueString('instanceid')) !== null) {
+			// No auth header, let's try session id, but only if this is an existing
+			// session and the request has a session cookie
 			try {
 				$token = $this->session->getId();
 			} catch (SessionNotAvailableException $ex) {
 				return false;
 			}
+		} else {
+			return false;
 		}
 
 		if (!$this->loginWithToken($token)) {
--- a/tests/lib/User/SessionTest.php
+++ b/tests/lib/User/SessionTest.php
@@ -479,6 +479,56 @@ class SessionTest extends \Test\TestCase {
 		$userSession->logClientIn('john', 'doe', $request, $this->throttler);
 	}
 
+	public function testTryTokenLoginNoHeaderNoSessionCookie(): void {
+		$request = $this->createMock(IRequest::class);
+		$this->config->expects(self::once())
+			->method('getSystemValueString')
+			->with('instanceid')
+			->willReturn('abc123');
+		$request->method('getHeader')->with('Authorization')->willReturn('');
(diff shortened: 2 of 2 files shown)


lib/private/User/Session.php
```diff
@@ -858,6 +858,13 @@ public function tryTokenLogin(IRequest $request) {
 			return false;
 		}
 
+		// Check if the request had sent a cookie with the instance id as name
+		// If there is no cookie, this is a new session
+		$instanceId = $this->config->getSystemValueString('instanceid');
+		if (is_null($request->getCookie($instanceId))) {
+			return false;
+		}
```
juliushaertl, line 866: Without looking in depth, just from the following code lines I think that this could have unexpected side effects when just authenticating with an app password which is also a token but not necessarily has cookies if clients send it.

ChristophWurst: Adjusted so that the token lookup by session id only happens if there is neither an Authrorization header nor session cookie

ChristophWurst (approved):
Bildschirmfoto vom 2024-01-16 19-19-41

:+1: as expected

tested with 100 samples of `GET /login`

juliusknorr (approved):

lib/private/User/Session.php
```diff
@@ -842,13 +842,16 @@ public function tryTokenLogin(IRequest $request) {
 		$authHeader = $request->getHeader('Authorization');
 		if (str_starts_with($authHeader, 'Bearer ')) {
 			$token = substr($authHeader, 7);
-		} else {
-			// No auth header, let's try session id
+		} else if ($request->getCookie($this->config->getSystemValueString('instanceid')) !== null) {
```
solracsf, line 845:
```suggestion
		} elseif ($request->getCookie($this->config->getSystemValueString('instanceid')) !== null) {
```

PR #42610, opened 2024-01-06
(config.php.sample) Clarify the purpose of `session_keepalive` parameter

### LINKED ISSUE #7244: Description of auto_logout could be more clear.
`auto_logout` is described as:
> Enable or disable the automatic logout after session_lifetime, even if session keepalive is enabled. This will make sure that an inactive browser will be logged out even if requests to the server might extend the session lifetime.

I initially read that as saying the server would invalidate/delete the session after `session_lifetime`, even if there are requests/activity that would otherwise extend the session lifetime. It looks like it just affects what the client does though, not the server. Especially given that there doesn't seem to be another way for the server to limit the duration of a session in the face of a compromised client, it would be nice if the documentation made it clear that `auto_logout` does not provide that security.

hellodarkness:
## Summary

Comments to clarify the purpose of session_keepalive. 

Fixes nextcloud/documentation#7244

## Checklist

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -342,7 +342,7 @@ $CONFIG = [
  * keepalive is enabled. This will make sure that an inactive browser will log itself out
  * even if requests to the server might extend the session lifetime. Note: the logout is   
  * handled on the client side. This is not a way to limit the duration of potentially
- * compromized sessions.
+ * compromised sessions.
  *
  * Defaults to ``false``
  */
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -339,8 +339,10 @@ $CONFIG = [
 
 /**
  * Enable or disable the automatic logout after session_lifetime, even if session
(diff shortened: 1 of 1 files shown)


config/config.sample.php
```diff
@@ -339,8 +339,10 @@
 
 /**
  * Enable or disable the automatic logout after session_lifetime, even if session
- * keepalive is enabled. This will make sure that an inactive browser will be logged out
- * even if requests to the server might extend the session lifetime.
+ * keepalive is enabled. This will make sure that an inactive browser will log itself out
+ * even if requests to the server might extend the session lifetime. Note: the logout is   
+ * handled on the client side. This is not a way to limit the duration of potentially
+ * compromized sessions.
```
provokateurin, line 345:
```suggestion
 * compromised sessions.
```

provokateurin (approved):

mejo- (approved):

PR #42638, opened 2024-01-08
also filter by storage when getting shares in folder

icewind1991:
Found while looking through the code.

Only filtering by `path` (when doing the non-shallow search) could otherwise return results from other storages/users.

The filter by `parent` doesn't strictly require the `storage` filter but I've added it anyway for consistency.


nickvergessen:
Do you know, does this improve/decrease performance of the queries due to index changes?

icewind1991:
I would expect filtering by path to be faster with the added filter, filtering by parent should be the same

nickvergessen (approved):

come-nc (approved):

PR #42643, opened 2024-01-09
fix(core): enhanced contrast when UserMenuItem is focused

### LINKED ISSUE #42565: Focus for user-menu nav is not accessible
# Issue Summary
Whenever a user tabs onto the selected view for their user-menu navbar, the contrast ratio may be too low.

EG. : 

Image

This may be to hard for many to see

emoral435:
* Resolves: #42565 

## Summary
Makes the focus and active state of the user-menu for the chosen view have better contrast

## 🖼️ Screenshots

🏚️ Before | 🏡 After
---|---
firefox_KMsTYveI03 | firefox_NRbRmYvb39

## Checklist


Pytal (approved):

ShGKme (changes requested):
The problem here is that we have a conflict between `.active` and `:focus-visible` styles.

This solution fixes it by manually defining `.active:focus-visible` styles. But it is more risky as we need to manually manage consistency between all 3 cases. For example, icon is not light on dark mode now: 

As alternative, we can disable active styles from focus visible by adding `:not(:focus-visible)` selector on active styles (on both `a` and `img`. For some reason it's 2 blocks, but we can do it in 1)

Then we won't need to duplicate styles and can be sure that they are the same.

image

emoral435:
@ShGKme your suggestion worked perfectly :) thank you! This is the proposed changes outcome:

firefox_rYQCZhzvcR

ShGKme (approved):

PR #42649, opened 2024-01-09
enh(OCP\Translation): Add ITranslationProviderWithId

marcelklehr:
## Summary
In order to allow External Apps to register machine translation providers, we need to allow declaring a getId() method which is used instead of the class name, as app_api will use an anonymous class to register these and all providers would have the same class name. 

## Checklist


bigcat88 (approved):

andrey18106 (approved):

marcelklehr:
Drone failure seems unrelated

nickvergessen (approved):

marcelklehr:
Added ITranslationProviderWithUserId

lib/private/Translation/TranslationManager.php
```diff
@@ -73,19 +77,26 @@ public function translate(string $text, ?string &$fromLanguage, string $toLangua
...
+				($a instanceof ITranslationProviderWithId ? $a->getId() : $a::class) <=> ($b instanceof ITranslationProviderWithId ? $b->getId() : $b::class)
+			);
 			$providers = $newProviders;
 		}
 
 		if ($fromLanguage === null) {
 			foreach ($providers as $provider) {
 				if ($provider instanceof IDetectLanguageProvider) {
+					if ($provider instanceof  ITranslationProviderWithUserId) {
+						$provider->setUserId($this->userSession->getUser()?->getUID());
```
nickvergessen, line 98: this inserts null if `$this->userSession->getUser()` is null, but `setUserId` does not allow null

lib/private/Translation/TranslationManager.php
```diff
@@ -73,19 +77,26 @@ public function translate(string $text, ?string &$fromLanguage, string $toLangua
...
+			$newProviders += array_udiff($providers, $newProviders, fn ($a, $b) =>
+				($a instanceof ITranslationProviderWithId ? $a->getId() : $a::class) <=> ($b instanceof ITranslationProviderWithId ? $b->getId() : $b::class)
+			);
 			$providers = $newProviders;
 		}
 
 		if ($fromLanguage === null) {
 			foreach ($providers as $provider) {
 				if ($provider instanceof IDetectLanguageProvider) {
+					if ($provider instanceof  ITranslationProviderWithUserId) {
```
nickvergessen, line 97:
```suggestion
					if ($provider instanceof ITranslationProviderWithUserId) {
```

lib/private/Translation/TranslationManager.php
```diff
@@ -73,19 +77,26 @@ public function translate(string $text, ?string &$fromLanguage, string $toLangua
 			$precedence = json_decode($json, true);
 			$newProviders = [];
 			foreach ($precedence as $className) {
-				$provider = current(array_filter($providers, fn ($provider) => $provider::class === $className));
+				$provider = current(array_filter($providers, fn ($provider) =>
+				  $provider instanceof ITranslationProviderWithId ? $provider->getId() === $className : $provider::class === $className))
+				;
```
nickvergessen, line 82: if you are not staying on a single line, I would recommend to use `function() { return  … }` instead of `fn() => …` to avoid issues when someone adds a line in the future.

nickvergessen (approved):

bigcat88:
Can someone use super power to force merge this?

bigcat88:
Thank you, Julius.

PR #42662, opened 2024-01-09
enh(core): move all contacts menu features to unified settings

### LINKED ISSUE #42386: Move all features within the contacts menu button on the header to unified search
This feature has been duplicated since the global search has been there. Now that the global search is even better, we should remove it.

Screenshot from 2023-12-20 09-24-44

emoral435:
* Resolves: #42386

## Summary
Because the unified search has the same functionality as the contacts menu in the header, it makes more sense to put it in the unified search section of 'People'. This got cleared by the designers too! So I added the same functionality seen in the contacts menu. It essentially just 'moved' the functionality to the unified search :)

There are some caveats, 'sins of the father' per say that happened when moving the feature. 
For example, as shown in the gif below, on scroll, the 'local time' popup can be scrolled in such a way that makes it go to high up in the view. Perhaps a scroll event can be used to stop this?
The action buttons are also not as highlighted on hover, as the whole button has to be hovered! There also may be some accessibility issues that are present, but in the last check, Micheal did not seem to find any, but if there are any, I will fix them as so.

As this is a medium sized commit, should I also include tests? New to this, just let me know :)

## GIF to show the new changes!
firefox_CgoHkqwaVy

## Checklist


emoral435:
Any idea on how to change the Cypress issues failing? They seem related to me removing the ContactsMenu.vue file :)

marcoambrosini (commented):
@emoral435 the functionality should be moved to the contacts search results, not to the search filters. I'm available tomorrow for a quick call to clarify this if you'd like :)

emoral435:
See OG issue to see resolution!

Issue #42489, opened 2023-12-27
[Bug]: S3 excessive egress traffic caused by background GenerateMetadataJob


motoride:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

NC tries to reindex metadata on my entire external storage mount, causing high data transfer costs.

### Steps to reproduce

1. Attach external S3 compatible storage
2. Wait for cron to trigger GenerateMetadataJob 
3. Observe every single file on external storage being called through the API on a daily basis

### Expected behavior

Files should be indexed once, and if possible shouldn’t be downloaded from the external storage which causes high data transfer costs. 

### Installation method

Other Community project

### Nextcloud Server version

28

### Operating system

Other

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database engine version

MySQL

### Is this bug present after an update or on a fresh install?

Upgraded to a MAJOR version (ex. 22 to 23)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "***REMOVED SENSITIVE VALUE***"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "dbtype": "mysql",
        "version": "28.0.0.11",
        "overwrite.cli.url": "***REMOVED SENSITIVE VALUE***",
        "dbname": "***REMOVED SENSITIVE VALUE***",
[...]

PR #42670, opened 2024-01-10
occ background-job:delete

ArtificialOwl:
allow an admin to remove a background job entry from the database

### DIFF
--- a/core/Command/Background/Delete.php
+++ b/core/Command/Background/Delete.php
@@ -1,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2024, Maxence Lange <maxence@artificial-owl.com>
+ *
+ * @author Maxence Lange <maxence@artificial-owl.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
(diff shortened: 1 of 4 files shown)


core/Command/Background/DeleteCommand.php
```diff
@@ -0,0 +1,80 @@
...
+namespace OC\Core\Command\Background;
+
+use OC\Core\Command\Base;
+use OCP\BackgroundJob\IJobList;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Question\ConfirmationQuestion;
+
+class DeleteCommand extends Base {
```
nickvergessen, line 35:
```suggestion
class DeleteJob extends Base {
```
?

nickvergessen (commented):
This is potentially very dangerous and at the same time some apps might regenerate their jobs.
So far I never had the case where an admin would have needed to delete a job? Can you give an example?

ArtificialOwl:
> This is potentially very dangerous and at the same time some apps might regenerate their jobs. So far I never had the case where an admin would have needed to delete a job? Can you give an example?

The only one I have in mind right now is this issue with a background job that could not resolve itself:
https://github.com/nextcloud/server/issues/42489

It is a unique job that remove itself when its purpose is over.
While this job should be fixed by https://github.com/nextcloud/server/pull/42667 it also made us realize there is no easy way to cancel it

core/Command/Background/DeleteJob.php
```diff
@@ -0,0 +1,80 @@
...
+			$output->writeln('<error>Job with ID ' . $jobId . ' could not be found in the database</error>');
+			return 1;
+		}
+
+		$output->writeln('Job class: ' . get_class($job));
+		$output->writeln('Arguments: ' . json_encode($job->getArgument()));
+		$output->writeln('');
+
+		$question = new ConfirmationQuestion(
+			'<comment>Do you really want to delete this background job ?</comment> (y/N) ', false,
```
artonge, line 67:
```suggestion
			'<comment>Do you really want to delete this background job ? It could create some misbehaviours in Nextcloud.</comment> (y/N) ', false,
```

artonge (approved):

come-nc:
Please just name the file Delete.php and the class Delete if the command is :delete.

ListCommmand is called ListCommand because list is a reserved keyword, delete is not.

joshtrichards:
Could "hide" being `NC_debug` but the built-in confirmation might also be enough warning.

come-nc (approved):

PR #42678, opened 2024-01-10
refactor(App): Remove registerRoutes method

provokateurin:
## Summary

Removes the deprecated App::registerRoutes method.
The RouteConfig class is no longer used so I deleted it as well including it's tests.
For some reason much of the implementation was the same as RouteParser, but that one has no tests :thinking:.

## Checklist

  - [ ] 29 upgrade docs for app devs


provokateurin:
OpenAPI is failing because it previously failed to discover all the routes in user_ldap so they are missing now and need to be documented.

lib/private/AppFramework/Routing/RouteConfig.php
```diff
@@ -1,294 +0,0 @@
...
-	/** @var array */
-	private $routes;
-
-	/** @var string */
-	private $appName;
-
-	/** @var string[] */
-	private $controllerNameCache = [];
-
-	protected $rootUrlApps = [
```
nickvergessen, line 56: ???

How is this implemented now?

nickvergessen:
> The RouteConfig class is no longer used so I deleted it as well including it's tests.
For some reason much of the implementation was the same as RouteParser, but that one has no tests 🤔.

Time to migrate the tests over then?

lib/private/AppFramework/Routing/RouteConfig.php
provokateurin: Check RouteParser, it has exactly the same apps listed and the same logic :woman_shrugging:. Think someone just copied parts of the implementation which lead to this logic duplication.

nickvergessen: 🙈

lib/public/AppFramework/App.php
```diff
@@ -34,10 +35,7 @@
  */
 namespace OCP\AppFramework;
 
-use OC\AppFramework\Routing\RouteConfig;
-use OC\Route\Router;
 use OC\ServerContainer;
-use OCP\Route\IRouter;
```
come-nc, line 40: Delete IRouter as well maybe? It’s marked deprecated since 9.0.0.

provokateurin: Sure.

I found the IRoute class right next to it. It's not deprecated, but I also couldn't find any usage within server, so it could be deleted as well or do we first need to deprecate it too?

provokateurin: But Router which implements IRouter is still used and can not simply be removed. I don't know if it makes sense to delete the interface in that case.

come-nc: https://github.com/nextcloud/server/commit/9bf01283e72716d4f94ca6893ff3d6cda6680d44 Not clear to me why it was deprecated, and what should be used instead.
Feel free to ignore this.

provokateurin:
> Time to migrate the tests over then?

Yes, but I took a look and I'm not really sure what the tests are trying to achieve. Are they testing that the routes are correctly parsed or are they checking that the routes are added correctly to the Router?

come-nc:
@provokateurin  Can you extract the `apps/` part that can be merged right away without any breaking I think?

susnux:
Rewrote the tests.

tests/lib/AppFramework/Routing/RouteParserTest.php
```diff
@@ -0,0 +1,348 @@
+<?php
+/**
```
come-nc, line 2: strict types

ChristophWurst (approved):
:broom: :sparkles:

lib/private/Route/Router.php
```diff
@@ -168,8 +168,7 @@ public function loadRoutes($app = null) {
 		if (!isset($this->loadedApps['core'])) {
 			$this->loadedApps['core'] = true;
 			$this->useCollection('root');
-			$this->setupRoutes($this->getAttributeRoutes('core'), 'core');
-			require __DIR__ . '/../../../core/routes.php';
+			$this->requireRouteFile(__DIR__ . '/../../../core/routes.php', 'core');
```
susnux, line 171:
```suggestion
			$this->setupRoutes($this->getAttributeRoutes('core'), 'core');
			$this->requireRouteFile(__DIR__ . '/../../../core/routes.php', 'core');
```

susnux (approved):

come-nc (approved):

provokateurin:
Cool, can't approve my own PR unfortunately :)

PR #42679, opened 2024-01-10
fix(quota): automatically detect chunk quota with `OC-Total-Length` header

### LINKED ISSUE #40702: Enhance our of quota check
TODO
- implement optional filesize header for chunked upload
- use it on clients
- until then clients will do a propfind with <d:quota-available-bytes />

- given an user with quota and groupfolders and shared folders
- client uploads a file and receives 507 "quota_exceeded"
- Android would now retry to upload file upon next app start
- instead it would be better to first check if enough quota is available

Problem:
We cannot user user info, as this is only user's quota, but not taking groupfolders or shared folders into account

@kesselb (as we talked about it on chat)

skjnldsv:
## Problem

When uploading chunks, legacy old methods had a dedicated `oc-chunked` header which is NOT used anymore
There was some specific name pattern to follow and the new chunk uploading is much simpler

## Solution

Because we also send the `OC-Total-Length` header, we can actually check the quota from the first uploaded chunk.

### Results

With https://github.com/nextcloud-libraries/nextcloud-upload/pull/1042

| Before | After |
|:---------:|:------:|
|||

Fix https://github.com/nextcloud/server/issues/40702

### DIFF
--- a/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/QuotaPlugin.php
@@ -31,6 +31,7 @@
 namespace OCA\DAV\Connector\Sabre;
 
 use OCA\DAV\Upload\FutureFile;
+use OCA\DAV\Upload\UploadFolder;
 use OCP\Files\StorageNotAvailableException;
 use Sabre\DAV\Exception\InsufficientStorage;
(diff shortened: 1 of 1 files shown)


come-nc (approved):

juliusknorr (approved):

skjnldsv:
Docs update https://github.com/nextcloud/documentation/pull/11432

PR #42680, opened 2024-01-10
Improve view only error message

artonge:
When a user tries to download a shared file that do not have the download permissions, this plugin will throw an error, but the message is misleading.

Here it is adjusted.


skjnldsv (approved):

ArtificialOwl (approved):

ArtificialOwl:
I have been here

PR #42668, opened 2024-01-09
enh(profile): Refactor code to use vue components and styles where possible

susnux:
## Summary
Refactor profile view to make use of standard vue components. This reduces code complexity and duplicated code.
Moreover some visual bugs are resolved, e.g. the profile action buttons now work correctly (see below).

Also a lot of custom styling is removed in favor of our component styles from the nextcloud vue library.

actions before | actions after
---|---
Screenshot_20240109_194640|Screenshot_20240109_203810

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #42705, opened 2024-01-11
fix(openapi): Make OpenAPI CI green again

nickvergessen:
* Resolves regression from #42668


susnux (approved):
Ok, but then we should not recommend everyone to use the new way :no_mouth:

bigcat88 (approved):

juliusknorr (approved):

szaimen:
Makes openAPI green 👍

Issue #42674, opened 2024-01-10
[Bug]: The counting language is corrupted


nhanlx89:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

![Untitled](https://github.com/nextcloud/server/assets/90684647/c6822b4c-3d02-4524-9816-54872837784b)
There is a problem with the Vietnamese language error as shown in the picture I sent
For English, it appears 1 folder and 2 files. but for Vietnamese, it does "{đếm} tệp và {đếm} thư mục"
/

### Steps to reproduce

1.Login to the website
2.And see the error
3.Notify the manufacturer

### Expected behavior

I need it to work properly and not have language errors anymore

### Installation method

None

### Nextcloud Server version

28

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Updated from a MINOR version (ex. 22.1 to 22.2)

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [X] Other

### Configuration report

```shell
{
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "files.angiang.dcs.vn"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "dbtype": "mysql",
        "version": "28.0.1.1",
        "overwriteprotocol": "https",
        "overwrite.cli.url": "http:\/\/files.angiang.dcs.vn",
[...]

PR #42716, opened 2024-01-11
fix(files): use `getSummaryFor` for better files summary translations

skjnldsv:
Because the old way was not l10n compliant, I created a dedicated more complete summary generation, for drag and drop, but the main files list was not using it

Discovered with https://github.com/nextcloud/server/issues/42674


artonge (approved):

ShGKme (approved):

skjnldsv:
https://github.com/nextcloud/server/pull/42734

PR #42746, opened 2024-01-12
Update IRC URL

j-carl:
* Resolves: N/A

## Summary

Update IRC URL to use libera instead of freenode

## Checklist

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -4,7 +4,7 @@
 -->
 ## Submitting issues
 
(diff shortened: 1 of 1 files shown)


AndyScherzinger:
@blizzz since tagged on v30. Shall we just remove it completely since we do not use IRC in any way anymore?

blizzz:
> @blizzz since tagged on v30. Shall we just remove it completely since we do not use IRC in any way anymore?

There are ~360 in the channel at this moment. By skimming, I do not see any employee though. So, unsure about completely removing it. Maybe adding some "inofficial" to it?

PR #42801, opened 2024-01-15
feat(AppFramework): Add Route attribute

provokateurin:
## Summary

Depends on openapi-extractor: https://github.com/nextcloud/openapi-extractor/pull/73
Documentation: https://github.com/nextcloud/documentation/pull/11442

A new routing method that should make it easier for new developers. It avoids having to manually verify the connection between the layers of the routes and the controllers.

The changes to the openapi.json file are annoying but required because the order of the routes changed. You can verify there were no changes using http://play.jd-tool.io (with unordered set option). Since there is no diff we can also be sure I migrated all routes correctly.

I did some benchmarking and would consider the slight drop in performance to be acceptable. It is about 2-5ms which is not noticable and there are many other factors in a remote HTTP connection that cause way more delay (tests were done on localhost). I have some ideas which could be tried to improve the performance a bit (but there is no guarantee that they will).

Only the 99th percentile is really relevant as the longest request can also be a fluke (see in the baseline testing for the capabilities endpoint). All tests were done with 10k requests on the standalone PHP webserver with SQLite and without any caching configured.

The core test is to check for changes in parsing the routes and the test with Talk (which has not been migrated) is to check the impact on apps with many routes. The total number of enabled apps has no performance impact on this new mechanism.

Before:
```log
$ ab -n 10000 -H "OCS-APIRequest: true" http://localhost:8080/ocs/v2.php/cloud/capabilities
This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:
Server Hostname:        localhost
Server Port:            8080

Document Path:          /ocs/v2.php/cloud/capabilities
Document Length:        1220 bytes

Concurrency Level:      1
Time taken for tests:   143.307 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      25310186 bytes
HTML transferred:       12200000 bytes
Requests per second:    69.78 [#/sec] (mean)
Time per request:       14.331 [ms] (mean)
Time per request:       14.331 [ms] (mean, across all concurrent requests)
Transfer rate:          172.48 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    12   14   4.0     13     354
Waiting:       12   14   4.0     13     353
Total:         12   14   4.0     13     354

Percentage of the requests served within a certain time (ms)
  50%     13
  66%     14
  75%     14
  80%     15
  90%     17
  95%     19
  98%     21
  99%     22
 100%    354 (longest request)

$ ab -n 10000 -H "OCS-APIRequest: true" -H "Authorization: Bearer bgN7T-A4q9p-QTT5C-QcFEK-57wfc" http://localhost:8080/ocs/v2.php/apps/spreed/api/v4/room
This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:
Server Hostname:        localhost
Server Port:            8080

Document Path:          /ocs/v2.php/apps/spreed/api/v4/room
Document Length:        7612 bytes

Concurrency Level:      1
Time taken for tests:   260.035 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      89919180 bytes
HTML transferred:       76120000 bytes
Requests per second:    38.46 [#/sec] (mean)
Time per request:       26.003 [ms] (mean)
Time per request:       26.003 [ms] (mean, across all concurrent requests)
Transfer rate:          337.69 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    23   26   4.2     24     269
Waiting:       23   26   4.2     24     267
Total:         23   26   4.2     24     269

Percentage of the requests served within a certain time (ms)
  50%     24
  66%     26
  75%     27
  80%     28
  90%     31
  95%     33
  98%     35
  99%     37
 100%    269 (longest request)
```

Now:
```log
$ ab -n 10000 -H "OCS-APIRequest: true" http://localhost:8080/ocs/v2.php/cloud/capabilities
This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:
Server Hostname:        localhost
Server Port:            8080

Document Path:          /ocs/v2.php/cloud/capabilities
Document Length:        1220 bytes

Concurrency Level:      1
Time taken for tests:   147.590 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      25309812 bytes
HTML transferred:       12200000 bytes
Requests per second:    67.76 [#/sec] (mean)
Time per request:       14.759 [ms] (mean)
Time per request:       14.759 [ms] (mean, across all concurrent requests)
Transfer rate:          167.47 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    13   15   2.5     14      42
Waiting:       12   14   2.5     13      41
Total:         13   15   2.5     14      42

Percentage of the requests served within a certain time (ms)
  50%     14
  66%     14
  75%     15
  80%     16
  90%     18
  95%     20
  98%     22
  99%     24
 100%     42 (longest request)

$ ab -n 10000 -H "OCS-APIRequest: true" -H "Authorization: Bearer bgN7T-A4q9p-QTT5C-QcFEK-57wfc" http://localhost:8080/ocs/v2.php/apps/spreed/api/v4/room
This is ApacheBench, Version 2.3 <$Revision: 1903618 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking localhost (be patient)
Completed 1000 requests
Completed 2000 requests
Completed 3000 requests
Completed 4000 requests
Completed 5000 requests
Completed 6000 requests
Completed 7000 requests
Completed 8000 requests
Completed 9000 requests
Completed 10000 requests
Finished 10000 requests


Server Software:
Server Hostname:        localhost
Server Port:            8080

Document Path:          /ocs/v2.php/apps/spreed/api/v4/room
Document Length:        7612 bytes

Concurrency Level:      1
Time taken for tests:   279.000 seconds
Complete requests:      10000
Failed requests:        0
Total transferred:      89920016 bytes
HTML transferred:       76120000 bytes
Requests per second:    35.84 [#/sec] (mean)
Time per request:       27.900 [ms] (mean)
Time per request:       27.900 [ms] (mean, across all concurrent requests)
Transfer rate:          314.74 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.0      0       0
Processing:    24   28   5.3     26     360
Waiting:       24   27   5.2     26     359
Total:         24   28   5.3     27     360

Percentage of the requests served within a certain time (ms)
  50%     27
  66%     27
  75%     28
  80%     29
  90%     34
  95%     38
  98%     39
  99%     42
 100%    360 (longest request)
```

## Checklist


provokateurin:
No idea why OpenAPI changes (seems to use the old version?), but the openapi-extractor PR needs to be merged first anyway.

lib/public/AppFramework/Http/Attribute/Route.php
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @since 29.0.0
+ */
+#[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+class Route {
+
+	/**
+	 * Corresponds to the `ocs` key in routes.php
+	 */
+	public const TYPE_OCS = 'ocs';
```
nickvergessen, line 42:
```suggestion
	 * Corresponds to the `ocs` key in routes.php
	 * @since 29.0.0
	 */
	public const TYPE_API = 'ocs';
```

lib/public/AppFramework/Http/Attribute/Route.php
```diff
@@ -0,0 +1,153 @@
...
+
+	/**
+	 * Corresponds to the `routes` key in routes.php
+	 */
+	public const TYPE_INDEX = 'routes';
```
nickvergessen, line 47:
```suggestion
	 * Corresponds to the `routes` key in routes.php
	 * @since 29.0.0
	 */
	public const TYPE_FRONTPAGE = 'routes';
```

nickvergessen (commented):
We discussed at another point recently that all future API work should be using OCS routes so APIs can also be used by mobile/desktop clients.

Could make this more clear by biasing the constant names.

---

Other than that, as mentioned I love the idea as it will reduce conflicts, amount of file and bring the 2 things that depend on each other closer together.

nickvergessen:
Since we use the Symfony router under the hood. Would there be an issue extending/using it's attribute as a base? We ship it already: https://github.com/nextcloud/3rdparty/blob/8a78b41c955bb3530a5c25f187e3dc96091dadd7/symfony/routing/Annotation/Route.php

Ref https://symfony.com/doc/current/routing.html

That could even allow us to remove more routing code in the future?

provokateurin:
> Would there be an issue extending/using it's attribute as a base?

I didn't know about this. Removing routing code sounds good, but this attribute also supports a lot of stuff that we don't support in our routing yet. I don't think that is a good idea :/

provokateurin:
(I only migrated core and one other app to ensure it works for both since they are handled a bit different in the router)

lib/public/AppFramework/Http/Attribute/Route.php
come-nc: `Route(Route::TYPE_OCS, ` is quite long, can we have `OcsRoute` and `FrontPageRoute` instead?
Also, I think there is a resource or collection feature in routes.php where you give an endpoint and it creates routes for post/put/delete/get, no? With the attribute we lose that? (May not be important though)

provokateurin: I thought about having separate attributes, but I find it more ugly and it duplicates all everything since only the class name would be different. The only way I see is to add attributes that extend the existing attribute.

You are talking about resources, but with this PR we lose absolutely nothing. It's just an additional way to define routes and it just doesn't support resources. We could add it in the future with another attribute, but it's a feature that is rarely used so I doubt the effort is worth it.

come-nc: > You are talking about resources, but with this PR we lose absolutely nothing. It's just an additional way to define routes and it just doesn't support resources. We could add it in the future with another attribute, but it's a feature that is rarely used so I doubt the effort is worth it.

I’d really like if we use only one way of declaring routes in the end, so for me if we add this we should deprecate routes.php, otherwise it’s confusing.
But yeah I’m not that concerned for `resources` I never used it myself I just thought it was worth mentionning.

provokateurin: There are also other cases where routes need to be defined that can not be done using attributes (e.g. PublicShareController or CORS).
I tend to agree that we should deprecate routes.php, but we should probably never remove support for it.

juliushaertl: Would also prefer two separate attributes for readability

juliushaertl: Or (not sure if that works with attributes) we could make the the type optional with a default to OCS, so that developers only need to specify it for frontpage contollers

provokateurin: Default value feels dangerous, I would rather not go with that option

provokateurin: I'll add two attributes that simply extend the current attribute

ChristophWurst:
Blackfire analysis:
* /login https://blackfire.io/profiles/compare/25ecda33-f01e-46d8-a45f-774c8db2a460/graph
* /ocs/v2.php/cloud/capabilities https://blackfire.io/profiles/compare/9093963f-b6d3-43bb-b695-18a97090dadb/graph

Every comparison is primed with a simple curl to warm caches.
Every test is run with 10 samples to get more accurate numbers.

provokateurin:
@ChristophWurst thanks for the profiling. When I did my testing I found that even with 1k requests I got very different results every time. At 10k it was quite stable.

To everyone: Do you think the performance hit is acceptable or not?

julien-nc (changes requested):
Code looks good and works fine.

The performance hit seems acceptable to me.

provokateurin:
This branch is currently broken if navigate to the root of the instance. Any other URL will work fine. I'm looking into it already.

lib/private/AppFramework/Routing/RouteConfig.php
```diff
@@ -136,7 +136,13 @@ protected function processRoute(array $route, string $routeNamePrefix = ''): voi
 		$controllerName = $this->buildControllerName($controller);
 		$actionName = $this->buildActionName($action);
 
-		$routeName = $routeNamePrefix . $this->appName . '.' . $controller . '.' . $action . $postfix;
+		/*
+		 * The route name has to be lowercase, for symfony to match it correctly.
+		 * This is required because smyfony allows mixed casing for controller names in the routes.
+		 * To avoid breaking all the existing route names, registering and matching will only use the lowercase names.
+		 * This is also save on the PHP side because class and method names collide regardless of the casing.
```
come-nc, line 143:
```suggestion
		 * This is also safe on the PHP side because class and method names collide regardless of the casing.
```

lib/private/AppFramework/Routing/RouteParser.php
```diff
@@ -100,7 +100,13 @@ private function processRoute(array $route, string $appName, string $routeNamePr
 		$controllerName = $this->buildControllerName($controller);
 		$actionName = $this->buildActionName($action);
 
-		$routeName = $routeNamePrefix . $appName . '.' . $controller . '.' . $action . $postfix;
+		/*
+		 * The route name has to be lowercase, for symfony to match it correctly.
+		 * This is required because smyfony allows mixed casing for controller names in the routes.
+		 * To avoid breaking all the existing route names, registering and matching will only use the lowercase names.
+		 * This is also save on the PHP side because class and method names collide regardless of the casing.
```
come-nc, line 107:
```suggestion
		 * This is also safe on the PHP side because class and method names collide regardless of the casing.
```

lib/private/Route/Router.php
```diff
@@ -412,6 +441,58 @@ protected function fixLegacyRootName(string $routeName): string {
...
+			}
+
+			$class = new ReflectionClass($appNameSpace . '\\Controller\\' . basename($file->getPathname(), '.php'));
+
+			foreach ($class->getMethods() as $method) {
+				foreach ($method->getAttributes() as $attribute) {
+					if ($attribute->getName() !== 'OCP\AppFramework\Http\Attribute\Route' &&
+						$attribute->getName() !== 'OCP\AppFramework\Http\Attribute\ApiRoute' &&
+						$attribute->getName() !== 'OCP\AppFramework\Http\Attribute\FrontpageRoute') {
+						continue;
```
come-nc, line 475: https://www.php.net/manual/en/reflectionfunctionabstract.getattributes.php

getAttributes could be used with `ReflectionAttribute::IS_INSTANCEOF` to avoid filtering ourselves?

lib/private/Route/Router.php
```diff
@@ -412,6 +441,58 @@ protected function fixLegacyRootName(string $routeName): string {
...
+					}
+
+					/** @var RouteAttribute $route */
+					$route = $attribute->newInstance();
+
+					$serializedRoute = $route->toArray();
+					// Remove 'Controller' suffix
+					$serializedRoute['name'] = substr($class->getShortName(), 0, -10) . '#' . $method->getName();
```
come-nc, line 483: No strtolower here?

lib/public/AppFramework/Http/Attribute/Route.php
```diff
@@ -0,0 +1,157 @@
...
+	 * @since 29.0.0
+	 */
+	public function getType(): string {
+		return $this->type;
+	}
+
+	/**
+	 * @since 29.0.0
+	 */
+	public function getVerb(): ?string {
```
come-nc, line 119:
```suggestion
	public function getVerb(): string {
```

lib/private/Route/Router.php
provokateurin: No, those routes will be passed on to the parseRoute method which will convert it to lowercase.

lib/private/Route/Router.php
provokateurin: Works, very nice improvement :+1:

provokateurin:
CI failure is fixed, it was just another case of the casing problem with matching legacy routes.

julien-nc (approved):
:+1:

lib/public/AppFramework/Http/Attribute/ApiRoute.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+/**
```
come-nc, line 3:
```suggestion
<?php

declare(strict_types=1);

/**
```

lib/public/AppFramework/Http/Attribute/FrontpageRoute.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+/**
```
come-nc, line 3:
```suggestion
<?php

declare(strict_types=1);

/**
```

lib/public/AppFramework/Http/Attribute/Route.php
```diff
@@ -0,0 +1,157 @@
+<?php
+
+/**
```
come-nc, line 3:
```suggestion
<?php

declare(strict_types=1);

/**
```

come-nc (approved):

apps/user_status/lib/Controller/HeartbeatController.php
```diff
@@ -81,6 +82,7 @@ public function __construct(string $appName,
 	 * 204: User has no status to keep alive
 	 * 400: Invalid status to update
 	 */
+	#[ApiRoute(verb: 'PUT', url: '/api/v1/heartbeat')]
```
nickvergessen, line 85: it was OCS?

lib/public/AppFramework/Http/Attribute/Route.php
```diff
@@ -0,0 +1,159 @@
...
+ */
+#[Attribute(Attribute::TARGET_METHOD | Attribute::IS_REPEATABLE)]
+class Route {
+
+	/**
+	 * Corresponds to the `ocs` key in routes.php
+	 *
+	 * @see ApiRoute
+	 */
+	public const TYPE_API = 'ocs';
```
nickvergessen, line 46: missing `@since` (Ref  https://github.com/nextcloud/server/pull/43581/files#diff-1701be546807fc8bd1d573dd5f48a53aa14647d1d8212c9f6db671f7813a304eR45-R46 )

lib/public/AppFramework/Http/Attribute/Route.php
```diff
@@ -0,0 +1,159 @@
...
+
+	/**
+	 * Corresponds to the `routes` key in routes.php
+	 *
+	 * @see FrontpageRoute
+	 */
+	public const TYPE_FRONTPAGE = 'routes';
```
nickvergessen, line 53: missing `@since`

PR #42811, opened 2024-01-15
make failure to load app commands non-fatal

icewind1991:
Currently if a command fails to load the occ commands fails with a fatal error. This also prevents using `occ app:disable` to disable the offending command.

This instead catches the error, logs it and continues loading the commands from other apps.

### DIFF
--- a/lib/private/Console/Application.php
+++ b/lib/private/Console/Application.php
@@ -121,7 +121,14 @@ class Application {
 						// load commands using info.xml
 						$info = $appManager->getAppInfo($app);
 						if (isset($info['commands'])) {
-							$this->loadCommandsFromInfoXml($info['commands']);
+							try {
+								$this->loadCommandsFromInfoXml($info['commands']);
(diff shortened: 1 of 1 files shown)


lib/private/Console/Application.php
```diff
@@ -121,7 +121,14 @@ public function loadCommands(
 						// load commands using info.xml
 						$info = $appManager->getAppInfo($app);
 						if (isset($info['commands'])) {
-							$this->loadCommandsFromInfoXml($info['commands']);
+							try {
+								$this->loadCommandsFromInfoXml($info['commands']);
+							} catch (\Exception $e) {
```
blizzz, line 126: maybe even go for Throwable?

blizzz (approved):

sorbaugh (approved):

come-nc (approved):

artonge (approved):

PR #42820, opened 2024-01-15
[stable28] prepare migration to lazy config

ArtificialOwl:
With the changes on `AppConfig` in NC29 (https://github.com/nextcloud/server/pull/41755), this patch will only prep the database for a smoother migration to NC29's.

This patch will avoid error if trying to set/get app config values (also disable/enable apps) between the moment the file are updated and the upgrade is started.

### DIFF
--- a/core/Migrations/Version28000Date20231126110901.php
+++ b/core/Migrations/Version28000Date20231126110901.php
@@ -1,0 +1,57 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright 2023 Maxence Lange <maxence@artificial-owl.com>
+ *
+ * @author Maxence Lange <maxence@artificial-owl.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\Core\Migrations;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+/**
+ * Create new fields for type and lazy loading in appconfig for the new IAppConfig API.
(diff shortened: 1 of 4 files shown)


core/Migrations/Version29000Date20231126110901.php
```diff
@@ -0,0 +1,65 @@
...
+		// type=2 means value is typed as MIXED
+		$table->addColumn('type', Types::INTEGER, ['notnull' => true, 'default' => 2]);
+		$table->addColumn('lazy', Types::BOOLEAN, ['notnull' => false, 'default' => false]);
+
+		if ($table->hasIndex('appconfig_config_key_index')) {
+			$table->dropIndex('appconfig_config_key_index');
+		}
+
+		$table->addIndex(['lazy'], 'ac_lazy_i');
+		$table->addIndex(['appid', 'lazy'], 'ac_app_lazy_i');
```
Altahrim, line 60: I am not sure this one is necessary. 
At least for MySQL like,  `ac_app_lazy_key_i` is enough.

core/Migrations/Version29000Date20231126110901.php
```diff
@@ -0,0 +1,65 @@
...
+		if ($table->hasColumn('lazy')) {
+			return null;
+		}
+
+		// type=2 means value is typed as MIXED
+		$table->addColumn('type', Types::INTEGER, ['notnull' => true, 'default' => 2]);
+		$table->addColumn('lazy', Types::BOOLEAN, ['notnull' => false, 'default' => false]);
+
+		if ($table->hasIndex('appconfig_config_key_index')) {
+			$table->dropIndex('appconfig_config_key_index');
```
artonge, line 56: So no more query on `configkey` only?

artonge (approved):

core/Migrations/Version29000Date20231126110901.php
ArtificialOwl: would not it been faster to have a specific index just on a single field (lazy) if I only run a search based on one field (lazy) ?

Altahrim: It's not possible to use `ac_app_lazy_i` for a search on `lazy` only. The first index `ac_lazy_i` will be used for that.

Maybe the best solution is to add only one index: `lazy, appid, configkey`, it will speedup searches on:
 - `lazy`
 - `lazy` + `appid` 
 - `lazy` + `appid` + `configkey` 

It depends of queries you need to speedup of course :)

nickvergessen:
Needs a `version.php` bump so the migration is triggered.
This is also missing in master.

core/Migrations/Version29000Date20231126110901.php
ArtificialOwl: yes, useless as never used

ArtificialOwl:
- removing the generation of indexes
- the migration will also be changed on master: #42903

Altahrim (approved):

nickvergessen:
> Block merges during freezes 

🙈

szaimen:
This is breaking updates from 28.0.0 and 28.0.1 to 29.0.x 🙈

PR #42839, opened 2024-01-16
Migrate available temp space check to new SetupCheck API

come-nc:
See #41364 

## Summary

Migrate available temp space check to new SetupCheck API.
Also put info in the check and not in the log file.

- [x] See whether is makes sense to display available temp space even if not s3 based?

## Checklist


come-nc:
`isEnoughTempSpaceAvailableIfS3PrimaryStorageIsUsed` uses `sys_get_temp_dir` but `temporaryDirectoryWritable` uses the `ITempManager`, which has a more complex way of find temp directory path.

I think it would make sense to merge both checks and always use path from `ITempManager` as that is what Nextcloud and apps should use. But we need to be sure this is what is used for uploads with an s3 backend.

nickvergessen (approved):

susnux (approved):

nickvergessen (approved):

PR #42850, opened 2024-01-16
fix(CalDAV): set acls for DeletedCalendarObjectsCollection

JohannesGGE:
* Resolves: # 

## TODO

- [x] ...

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/Trashbin/DeletedCalendarObjectsCollection.php
+++ b/apps/dav/lib/CalDAV/Trashbin/DeletedCalendarObjectsCollection.php
@@ -31,11 +31,15 @@ use Sabre\DAV\Exception\BadRequest;
 use Sabre\DAV\Exception\Forbidden;
 use Sabre\DAV\Exception\NotFound;
 use Sabre\DAV\Exception\NotImplemented;
+use Sabre\DAVACL\ACLTrait;
+use Sabre\DAVACL\IACL;
 use function array_map;
 use function implode;
 use function preg_match;
 
-class DeletedCalendarObjectsCollection implements ICalendarObjectContainer {
+class DeletedCalendarObjectsCollection implements ICalendarObjectContainer, IACL {
+	use ACLTrait;
+
 	public const NAME = 'objects';
 
 	/** @var CalDavBackend */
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
Makes sense

miaulalala (approved):

apps/dav/lib/CalDAV/Trashbin/DeletedCalendarObjectsCollection.php
```diff
@@ -31,11 +31,15 @@
...
 use Sabre\DAV\Exception\NotImplemented;
+use Sabre\DAVACL\ACLTrait;
+use Sabre\DAVACL\IACL;
 use function array_map;
 use function implode;
 use function preg_match;
 
-class DeletedCalendarObjectsCollection implements ICalendarObjectContainer {
+class DeletedCalendarObjectsCollection implements ICalendarObjectContainer, IACL {
+	use ACLTrait;
```
tcitworld, line 41: This brings 
```
'privilege' => '{DAV:}all',
```

I think `{DAV:}read` should be enough, as the collection shouldn't be changed.

tcitworld (dismissed):

apps/dav/lib/CalDAV/Trashbin/DeletedCalendarObjectsCollection.php
JohannesGGE: You mean by adding:
```php
public function getACL(): array {
	return [
		[
			'privilege' => '{DAV:}read',
			'principal' => $this->getOwner(),
			'protected' => true,
		],
		[
			'privilege' => '{DAV:}unbind',
			'principal' => '{DAV:}owner',
			'protected' => true,
		]
	];
}
```
?

tcitworld: Do you need `{DAV:}unbind` ?

tcitworld: But the owner should be able to delete the elements in the collection, not the collection itself, right?

`DeletedCalendarObject` already has `{DAV:}unbind`.

JohannesGGE: The deletion does not work either via cadaver or via web interface without `{DAV:}unbind` for the owner. At least in my setup.

ChristophWurst: > But the owner should be able to delete the elements in the collection, not the collection itself, right?
> 
> `DeletedCalendarObject` already has `{DAV:}unbind`.
> 
> And in any case the `delete` method here throws `Forbidden`.

It's a bit strange. If you drop unbind from the collection, objects in the collection can't be deleted neither. Sabre returns a `Node with name 'objects' could not be found`.

Since deletion is indeed protected by the Forbidden, I would like to move forward with this.

ChristophWurst: `\Sabre\DAVACL\Plugin::beforeUnbind` checks the parent for unbind before a node is unbound.

PR #42886, opened 2024-01-17
Creates a AllUserStatusWidget to show the currrent status of all users.

StCyr:
## Summary

Adds a "All users status" dashboard widget.

## TODO

Don't know. Please tell me...

- [ ] ...

## Checklist


nickvergessen:
But it's not all statuses anyway? Maybe you can explain in a ticket first what you are trying to achieve?

StCyr:
> But it's not all statuses anyway? Maybe you can explain in a ticket first what you are trying to achieve?

What do you mean by "not all statuses"? Do you mean "because it's limited to the first 10 users"?

I can create a ticket if you want but there's already an explanation here: https://help.nextcloud.com/t/dashboard-widget-to-show-status-of-all-users/178830

If your remark is about the fact that this widget is limited to the first 10 users, I will gladly implement any improvement to this.

nickvergessen:
> I can create a ticket if you want but there's already an explanation here: https://help.nextcloud.com/t/dashboard-widget-to-show-status-of-all-users/178830

That's good enough. Still not sure this should be part of the default as it would be quite confusing for people to have 2 status widgets not knowing the real difference.

Also that one does not really help the person in the forum, as they will still only see 7 people by default and not all 10.

StCyr:
> Also that one does not really help the person in the forum, as they will still only see 7 people by default and not all 10.

I've increased the default to 9 for the AllUserStatusWidget widget.

Also, the title is "User statuses" for this widget (should maybe be "Users status" btw) while it's "Recent statuses" for the UserStatusWidget widget.

Here's a screenshot of what it looks like in my dev environment:

image

johncarterofmars:
I am the originator of this request. The reason I suggested it is that there isn't a way to see a whole team at once. The recent statuses only show obviously the most recent changes. We were wanting something where a team member could see everyone all at once. For a team spread across time zones, this would be very helpful, especially if one is responsible for transferring calls to different people. That was my goal anyway. If it isn't useful or doesn't belong in main, I completely understand. Our team uses NC as our central hub, and one thing miss from the old Skype days is the ability to see everyone's statuses easily.

StCyr:
> I am the originator of this request. The reason I suggested it is that there isn't a way to see a whole team at once. The recent statuses only show obviously the most recent changes. We were wanting something where a team member could see everyone all at once. For a team spread across time zones, this would be very helpful, especially if one is responsible for transferring calls to different people. That was my goal anyway. If it isn't useful or doesn't belong in main, I completely understand. Our team uses NC as our central hub, and one thing miss from the old Skype days is the ability to see everyone's statuses easily.

I can implement a personal setting page if you want. But there's currently no setting page for this app, so I'm not sure you NC guys would accept that. Could you please give your opinion on this?

nickvergessen:
I would put it as a normal page (not settings) so it also gets a header entry and you can bookmark it.

But that sounds like there would be quite some UI overhead coming in as you would have to build it manually instead of just providing widget data.

So my suggestion would be, if this code here works, wrap it into a new app and repository, and you are good to go.

You can then extend it from there.

---

Too much overkill, but just as a note. When you use Talk, you can also see all user statuses if you have a room with everyone.

mejo-:
Just wanted to mention that I heard requests for this feature ("easily list status of all users") a few times already, so thanks @StCyr for looking into it. I don't think that it's enough to list only 10 users though. This doesn't scale for instances with more users.

My ideas would be to either make the dashboard widget interactive (with a search field), or to have a dedicated page with a (paginated) list of all user statuses - like @nickvergessen suggested.

I also wonder whether the search for people in our new unified search modal should show the status of users. @marcoambrosini and @Altahrim as the ones who worked on unified search recently.

blizzz:
FWIW the PR should go against the master branch, not stable28.

blizzz:
I am closing this, since NC 28 entered EOL and there is no recent activity.

PR #10825, opened 2018-08-23
Copy a file to the same directory

greenido:
1. Added the logic to change the new file name to 'xxxx_1' (or 2,3,4, when there are already copies of the file).
2. After the copy there was a need to reload the list of files in order to show the user the changes.

Fixes #9931

PR #42889, opened 2024-01-17
enh(files): Allow to copy files into same directory

### LINKED ISSUE #10441: Allow to duplicate a file by copying it to the same directory
Related to issue #9931.

It is not currently possible, but I'd love to be able to duplicate a text file + re-name it to .md at the same time while using the copy menu.  System notice that appears to the user is:

> Could not copy "Markdown Examples.txt"; target exists.

but it could become

> "Markdown Examples.txt" already exists; Do you want to Duplicate or Rename the copy?

My request is to allow users to duplicate (copy) and re-name a file within the Copy and Move menu when target already exists.
duplicate

### LINKED ISSUE #18846: Copy whole folder fails because of 'target exists'
related to #15785
when copying a whole folder, there is no way to paste that folder in the current parent.

### Steps to reproduce
1. create a folder `parent`
2. create a folder inside `parent` called `test`
3. select and make action `Move or copy`
4. click `copy` in the dialogue that appears

outcome is the error message: `could not copy "test", target exists`

### Expected behaviour
If there is an attempt to copy a folder into a target where the same name exists, an option should be provided to rename the copied folder.  Alternatively, a safe name could be suggested like `test_copy`.

### Actual behaviour
Copy of folder fails with no option to change the destination name.

### Server configuration

**Operating system:**
Debian 9

**Web server:**
Apache2.4

**Database:
MariaDB

**PHP version:**
7.2.26

**Nextcloud version:** 
17.0.2

**Updated from an older Nextcloud/ownCloud or fresh install:**
updated

### Client configuration
**Browser:**
Firefox (latest)

**Operating system:**
OS X

### LINKED ISSUE #41768: [Bug]: Not able to create a copy of a file in the same directory

### Bug description

I have a folder in Files with a file template for a meeting protocol. Every week I copy the file to the same directory to get a new copy. This is no longer possible in 28.

### Steps to reproduce

1. Create a folder
2. Create a file in the folder
3. Click the *Move or copy*

### Expected behavior

The file picker lets me pick the source directory for *Copy*.

### Installation method

None

### Nextcloud Server version

28

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

susnux:
* Bring back: #10825 
* Resolves https://github.com/nextcloud/server/issues/18846
* Resolves https://github.com/nextcloud/server/issues/10441
* Resolves: https://github.com/nextcloud/server/issues/41768

## Summary
Bring back copy to current directory behavior which will clone the file and add a "copy" suffix on the file name.

## Checklist


apps/files/src/actions/moveOrCopyAction.ts
```diff
@@ -69,6 +67,30 @@ const getActionForNodes = (nodes: Node[]): MoveCopyAction => {
  * @return {Promise<void>} A promise that resolves when the copy/move is done
  */
 export const handleCopyMoveNodeTo = async (node: Node, destination: Folder, method: MoveCopyAction.COPY | MoveCopyAction.MOVE, overwrite = false) => {
+	/**
+	 * Create an unique name for a node
+	 * @param node Node that is copied
+	 * @param otherNodes Other nodes in the target directory to check for unique name
+	 * @return Either the node basename, if unique, or the name with a `(copy N)` suffix that is unique
+	 */
+	const makeUniqueName = (node: Node, otherNodes: Node[]|FileStat[]) => {
```
skjnldsv, line 76: We already have a method and pattern for that :)
https://github.com/nextcloud/server/blob/9f157d7f8030a1f7a9455be10810ee11f1cb15f0/apps/files/src/newMenu/newFolder.ts#L59-L67

skjnldsv (changes requested):

apps/files/src/actions/moveOrCopyAction.ts
skjnldsv: Ah, I have no preferences, I used the old behavior, which was just the number! Please adjust the method then so we have whatever the design team approved! :+1:

apps/files/src/actions/moveOrCopyAction.ts
```diff
@@ -86,33 +109,43 @@ export const handleCopyMoveNodeTo = async (node: Node, destination: Folder, meth
 	 * node: /foo/bar/file.txt -> path = /foo/bar
 	 * destination: /foo
 	 * Allow move of /foo does not start with /foo/bar so allow
+	 * But allow copy a file to the same directory
 	 */
-	if (destination.path.startsWith(node.path)) {
+	if (destination.path.startsWith(node.path) && destination.path !== node.path) {
```
artonge, line 114: Does it prevent moving a folder into itself? Not sure how it could happen, but still.
```js
destination.path = "/a/b"
node.path = "/a/b"

destination.path.startsWith(node.path) === true
(destination.path !== node.path) === false
```

I think the following would work:
```suggestion
	if (destination.path.startsWith(node.path) && (destination.path !== node.path || node.type === Folder)) {
```

But warning, I might simply be confused.

apps/files/src/actions/moveOrCopyAction.ts
```diff
@@ -86,33 +109,43 @@ export const handleCopyMoveNodeTo = async (node: Node, destination: Folder, meth
...
+			const currentPath = join(davRootPath, node.path)
+			const destinationPath = join(davRootPath, destination.path)
 
-			// If we're moving, update the node
-			// if we're copying, we don't need to update the node
-			// the view will refresh itself
-			if (method === MoveCopyAction.MOVE) {
+			if (method === MoveCopyAction.COPY) {
+				let target = node.basename
+				// If we do not allow overwriting than find an unique name
```
artonge, line 130:
```suggestion
				// If we do not allow overwriting then find an unique name
```

apps/files/src/actions/moveOrCopyAction.ts
susnux: No the whole `&& ...` can be dropped.

Because if your target folder is `/foo/bar` then ofcourse I can not move `/foo` into it, but I can also not move `/foo/bar` into it, and I am even not allowed to copy it into itself, because if both have the same path I try:
`/foo/bar` -> `/foo/bar` leading to `/foo/bar/bar` which is also not allowed.

To clone that directory you would need to pass `/foo` as the destination and then the check is ok as it was.

apps/files/src/actions/moveOrCopyAction.ts
skjnldsv: Perfect!!  Thanks @susnux :)

artonge (approved):

apps/files/src/utils/fileUtils.ts
```diff
@@ -21,6 +21,25 @@
...
+ */
+export const getUniqueName = (name: string, otherNames: string[], suffix = (n: number) => `(${n})`): string => {
+	let newName = name
+	let i = 1
+	while (otherNames.includes(newName)) {
+		const ext = extname(name)
+		newName = `${basename(name, ext)} ${suffix(i++)}${ext}`
+	}
+	return newName
+}
```
emoral435, line 42: Seems like if you move folders with no extension name, getUniqueName clips the last letter from the file when you append the suffix to the end of the new file name:

image

Probably due to files not having extensions, so might have to add a check for that for files

emoral435 (changes requested):
Just needs the file part and it LGTM!!

susnux:
@emoral435 can you check again if that still happens for you?

emoral435 (approved):
Checked, and it works perfectly for me now :) sorry for being nitpicky!

emoral435:
@skjnldsv before we merge this in, it still says that you have requested changes. I'm 99% sure your fine with this, but just to make sure, all the changes now are fine by you, correct? :)

skjnldsv (approved):

PR #42896, opened 2024-01-17
fix(userstatus): CALL status should overwrite MEETING status

### LINKED ISSUE #42451: [Bug]: userstatus - "In a call" should take precedence over "In a meeting"

### Bug description

When a user is in a call, the "in a call message" should be written. If the meeting automation sets its status earlier, then it might not be overwritten by in a call

### Steps to reproduce

1. Have meeting in calendar
2. start call
3. Meeting status is not overwritten

### Expected behavior

Meeting status is overwritten.

It needs a rollback of a possible "In a meeting" status in the "In a call" automation.

### Installation method

None

### Nextcloud Server version

28

### Operating system

Other

### PHP engine version

Other

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

miaulalala:
* Resolves: https://github.com/nextcloud/server/issues/42451

## Summary

When a meeting status is set, the call status will not be written if a backup exists. The code now shortcuts this single scenario and overwrites the existing status "In a meeting" with the "In a call" status irregardless of a backup status, since both are automated status and will revert to the backup after their respective automation has run.

This also fixes the issue of having a call while "in a meeting" is set, leaving the call, and the meeting status not coming back after.

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/Status/StatusService.php
+++ b/apps/dav/lib/CalDAV/Status/StatusService.php
@@ -77,12 +77,13 @@ class StatusService {
 			return;
 		}
 
-		$userStatusTimestamp = null;
-		$currentStatus = null;
 		try {
 			$currentStatus = $this->userStatusService->findByUserId($userId);
-			$userStatusTimestamp = $currentStatus->getIsUserDefined() ? $currentStatus->getStatusTimestamp() : null;
+			// Was the status set by anything other than the calendar automation?
+			$userStatusTimestamp = $currentStatus->getIsUserDefined() && $currentStatus->getMessageId() !== IUserStatus::MESSAGE_CALENDAR_BUSY ? $currentStatus->getStatusTimestamp() : null;
 		} catch (DoesNotExistException) {
+			$userStatusTimestamp = null;
+			$currentStatus = null;
 		}
 
 		if($currentStatus !== null && $currentStatus->getMessageId() === IUserStatus::MESSAGE_CALL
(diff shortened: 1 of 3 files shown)


apps/dav/lib/CalDAV/Status/StatusService.php
```diff
@@ -123,19 +124,21 @@ public function processCalendarStatus(string $userId): void {
...
-		$this->logger->debug("Found $count applicable event(s), changing user status", ['user' => $userId]);
-		$this->userStatusService->setUserStatus(
-			$userId,
-			IUserStatus::AWAY,
-			IUserStatus::MESSAGE_CALENDAR_BUSY,
-			true
-		);
-
+		// Only update the status if it's neccesary otherwise we mess up the timestamp
+		if($currentStatus !== null && $currentStatus->getMessageId() !== IUserStatus::MESSAGE_CALENDAR_BUSY) {
```
ChristophWurst, line 128: what would be messed up?

miaulalala: I've had that in my local env. The timestamp is updated, thus the DTSTART is lower than the timestamp - which leads to a reset because no event is found (which is wrong).

apps/user_status/lib/Service/StatusService.php
```diff
@@ -257,21 +255,38 @@ public function setUserStatus(string $userId,
...
+			$userStatus->setStatusTimestamp($this->timeFactory->getTime());
+			$userStatus->setIsUserDefined(true);
+			$userStatus->setIsBackup(false);
+			$userStatus->setMessageId($messageId);
+			$userStatus->setCustomIcon(null);
+			$userStatus->setCustomMessage($customMessage);
+			$userStatus->setClearAt(null);
+			$userStatus->setStatusMessageTimestamp($this->timeFactory->now()->getTimestamp());
+			return $this->mapper->update($userStatus);
+		}
```
ChristophWurst, line 278: :+1: makes sense

apps/dav/lib/CalDAV/Status/StatusService.php
nickvergessen: One of the unit tests fails, so I wonder if the test is actually right and the code here wrong?
Should it be:
```suggestion
		// Only update the status if it's neccesary otherwise we mess up the timestamp
		if ($currentStatus === null || $currentStatus->getMessageId() !== IUserStatus::MESSAGE_CALENDAR_BUSY) {
```

Otherwise the event status will only be set, when you had any status before?

nickvergessen (changes requested):

nickvergessen (approved):

ChristophWurst (approved):

nickvergessen:
Already done in https://github.com/nextcloud/server/pull/43107

PR #42912, opened 2024-01-18
fix(i18n): Change cli to uppercase

rakekniven:
Like all other occurences.

Reported at Transifex.


nickvergessen (approved):

come-nc (approved):

nickvergessen:
Also should not merge now, to avoid breaking translations with todays RCs

PR #42913, opened 2024-01-18
fix(settings): stop reloading after force-enabling an app

skjnldsv:
## Steps
1. have an incompatible app
2. Click `Allow untested app`
3. The page reloads

## Fixed:
The page does NOT reload anymore

https://github.com/nextcloud/server/assets/14975046/7624798d-956c-422f-9a7d-bfb878788f36


artonge (approved):
:rofl: Not sure why, but this makes me laugh

skjnldsv:
> 🤣 Not sure why, but this makes me laugh

It's the little things™ :sparkles: :sparkles: :sparkles:  :see_no_evil:

szaimen (approved):
Tested and seems to work.

Maybe the idea of a reload was that a potential icon is directly shown in the header which it might not be now?

skjnldsv:
> Maybe the idea of a reload was that a potential icon is directly shown in the header which it might not be now?

That's what happens when you enable an app, not force-enable :)

PR #42933, opened 2024-01-18
fix(db): Execute dirty reads on the primary node

ChristophWurst:
* Follow-up to https://github.com/nextcloud/server/pull/42345

## Summary

By introducing read replicas in 29+ we open the door for new read after write problems. That is, when we write on the primary and read on the replica and the replica hasn't received the data yet. This condition is logged for developer awareness. Until the dirty read is fixed by the developer, we should force the dirty read to happen on the primary. This guarantees consistency, while hurting performance due to more load on the primary.

## Checklist

### DIFF
--- a/lib/private/DB/Connection.php
+++ b/lib/private/DB/Connection.php
@@ -285,6 +285,10 @@ class Connection extends PrimaryReadReplicaConnection {
 					'exception' => new \Exception(),
 				],
 			);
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):
Makes sense as discussed

nickvergessen (approved):

Issue #42683, opened 2024-01-10
[BITV]: adapt `aria-label`'s text for checkboxes in files list


JuliaKirschenheuter:
`aria-label`'s text stays unchanged while checking in- and out checkboxes like "Select the row for Media". 

- description should be more specific like "Select folder "New folder"" or "Select "Mycat.jpeg" file"
- ~description should be changed depending on a chackbox state like "Select folder "New folder"" -> "Deselect folder "New folder""~

![checkbox](https://github.com/nextcloud/server/assets/6078378/63490eb4-24c8-40ab-82f0-8d70c3ac13c7)

Pytal:
As the checked and indeterminate states already provide this information to screen readers, instead of adding this info to the label itself we should switch it to "toggle" and have the screen reader announce the correct states

szaimen:
I think this is missing a dependency update in stable28

szaimen:
For this one

PR #42949, opened 2024-01-18
fix(files): Fix checkbox state semantics

Pytal:
- For https://github.com/nextcloud/server/issues/42683

- Needs https://github.com/nextcloud-libraries/nextcloud-vue/pull/5096

> Opening for review in advance

## Summary

- To fix the semantics of checkboxes we remove the "select" / "unselect" wording from the labels
- Switch it to "toggle" and make use of the native checked and indeterminate `input` states

## Checklist


apps/files/src/components/FilesListTableHeader.vue
```diff
@@ -142,9 +142,7 @@ export default Vue.extend({
 		},
 
 		selectAllBind() {
-			const label = this.isNoneSelected || this.isSomeSelected
-				? this.t('files', 'Select all')
-				: this.t('files', 'Unselect all')
+			const label = t('files', 'Toggle selection for all files and folders')
```
skjnldsv, line 145: Aren't we loosing context here ?

susnux: I guess not because it will be announced like `checkbox - checked - Toggle selection for all files and folders` or `checkbox - unchecked - Toggle selection for all files and folders`

Pytal: Accessibility best practices suggest not doubling information that is already available e.g. a popup menu label should not say "Close menu" when opened or "Open menu" when closed when the state is already communicated with aria-expanded

emoral435 (approved):
After reading through the issue title change, I agree with the choice of using the wording 'Toggle', etc. instead of select/deselect and using the other native accessibility tools to aid the user. This LGTM!

emoral435:
image

I will say though, on a side-note: the value for the input label isnt being shown on the DOM. I could be wrong, and there could be a different way telling us which checkboxes are input, but isn't this also an a11y issue we should work on? Outside of the aria-label, are there any other native tools that let the user know they checked the box?

Pytal:
Thanks for the review @emoral435!

> I will say though, on a side-note: the value for the input label isnt being shown on the DOM. I could be wrong, and there could be a different way telling us which checkboxes are input, but isn't this also an a11y issue we should work on? Outside of the aria-label, are there any other native tools that let the user know they checked the box?

Don't know where it is in Firefox devtools but in Chrome you can see checked state in the Properties tab or just access the el.checked property in the console

emoral435:
No problem, @Pytal! Thank you for the MDN reference! Your PR is still great, just wanted to bring up the topic because I'm not all too familiar with this, but I checked this on Chrome and I see the checked property! Sweet :)

skjnldsv (approved):

PR #42950, opened 2024-01-18
fix(files): fixed aria-desciption logic for last breadcrumb

### LINKED ISSUE #41895: [BITV] 9.4.1.2/3.7 - In the breadcrumb navigation, there are "aria-label" attributes on "li" elements, which is not allowed, such as aria-label="Go to the "/" directory". (2)

emoral435:
* Resolves: #41895 
## Summary
Cleaned up the aria-description logic a bit!

## Screenshots for the new changes
image

## Checklist


apps/files/src/components/BreadCrumbs.vue
```diff
@@ -31,10 +31,10 @@
 			dir="auto"
 			:to="section.to"
 			:title="titleForSection(index, section)"
-			:aria-description="ariaForSection(index, section)"
+			:aria-description="ariaForSection(section)"
 			@click.native="onClick(section.to)">
 			<template v-if="index === 0" #icon>
-				<Home :size="20"/>
+				<Home :size="20" :aria-label="t('files', 'Home')"/>
```
ShGKme, line 37: The label should be on the interactive element itself — on the button. `NcBreadbrumb` sets it via the `name` prop. So the issue is probably in the data, where `name` "Home" is not translated here, in the `section` object.

ShGKme (changes requested):

emoral435:
Woops - pushed too soon. Will erase this comment once I find it, but the translation might be needed from another file etc. Will look into this!

apps/files/src/components/BreadCrumbs.vue
emoral435: That's odd. I checked, and when the `name` member in the `sections` object always will have the 'home' section translated:
image
So the fact that we even see the name 'Home' means that it should be translated, no?

Pytal:
I just checked on master and it is already translated correctly 💡

image

Pytal:
Seems @szaimen checked with German where it is not translated so no issue code-wise

szaimen:
> Seems @szaimen checked with German where it is not translated so no issue code-wise

indeed, looks like it...

emoral435:
@Pytal @ShGKme I still propose that this PR get merged. Upon checking the master branch, it appears that the last breadcrumb doesn't have that aria-description that we proposed to keep :/

image

This PR still fixes that! Updated the main summary for the PR though

Pytal (approved):

apps/files/src/components/BreadCrumbs.vue
emoral435: Turns out it was just due to the fact that the German translation is still untranslated, but the PR still holds value :)

emoral435:
/skjnldsv-backport ea765cfdd35d7f935b82b9d4dd438015a1d9e7e0 to stable28

skjnldsv (approved):

PR #42963, opened 2024-01-19
fix header inversion bug

### LINKED ISSUE #42962: Notifications menu and contact-search menu broken with certain predefined backgrounds
### How to use GitHub

### Steps

1. Go to go to index.php/settings/user/theming
2. Select one of the following default offered backgrounds: All that make the font color go black
  - Pink bubbles
  - Yellow bricks
  - Lion rock hill
  - "the ducks"
  - soft floral
  - morning fog
4. Open the notification menu => Inverted color ❌ 
5. Open the contact search => menu opens in place of the button instead of a popover menu 💥 
Bildschirmfoto vom 2024-01-19 14-43-44

How it looks with any other background:

Bildschirmfoto vom 2024-01-19 14-43-35

szaimen:
Should fix https://github.com/nextcloud/server/issues/42962

I am not really satisfied with the solution but it is what it is...


core/css/header.scss
```diff
@@ -172,7 +172,20 @@
 	/* Right header standard */
 	.header-right {
 		> .header-menu:not(.user-menu) {
-			filter: var(--background-image-invert-if-bright);
+			// For general
+			> .header-menu__trigger {
```
susnux, line 176:
```suggestion
			> button {
```

this should work for all I think

szaimen: I fear the assistant button is not a button but a link...

szaimen: Also I would rather not risk using the button tag here...

susnux (approved):
works!

core/css/header.scss
susnux: Then we should probably adjust the two other that do not use `.header-menu__trigger`.
But for the moment your solution works fine!

ShGKme: > Also I would rather not risk using the button tag here...

```css
button,
a {}
```

?

szaimen: Even more risky...

susnux: or even better only apply to `img` tag. Everything else should handle it correctly and we prevent future issues

ShGKme (approved):

blizzz:
rebased

AndyScherzinger (approved):

szaimen:
CI failure unrleated afaics

PR #42973, opened 2024-01-19
Expose dashboard layout and statuses API

provokateurin:
## Summary

I cleaned up the format that was used by the API, migrated the existing endpoints to OCS (the LayoutApiController never had a route so it was not possible to update the layout or statuses before) and added endpoints to get the current layout and statuses.

## Checklist


apps/dashboard/appinfo/routes.php
```diff
@@ -28,12 +28,14 @@
...
-		['name' => 'dashboard#updateStatuses', 'url' => '/statuses', 'verb' => 'POST'],
 	],
 	'ocs' => [
 		['name' => 'dashboardApi#getWidgets', 'url' => '/api/v1/widgets', 'verb' => 'GET'],
 		['name' => 'dashboardApi#getWidgetItems', 'url' => '/api/v1/widget-items', 'verb' => 'GET'],
 		['name' => 'dashboardApi#getWidgetItemsV2', 'url' => '/api/v2/widget-items', 'verb' => 'GET'],
+		['name' => 'dashboardApi#getLayout', 'url' => '/layout', 'verb' => 'GET'],
+		['name' => 'dashboardApi#updateLayout', 'url' => '/layout', 'verb' => 'POST'],
+		['name' => 'dashboardApi#getStatuses', 'url' => '/statuses', 'verb' => 'GET'],
+		['name' => 'dashboardApi#updateStatuses', 'url' => '/statuses', 'verb' => 'POST'],
```
julien-nc, line 39: Don't you wanna use the new routing system here?

apps/dashboard/src/DashboardApp.vue
```diff
@@ -394,13 +394,16 @@ export default {
 			}
 		},
 		enableStatus(app) {
-			this.enabledStatuses[app] = true
+			this.enabledStatuses.push(app);
```
julien-nc, line 397:
```suggestion
			this.enabledStatuses.push(app)
```

apps/dashboard/src/DashboardApp.vue
```diff
@@ -394,13 +394,16 @@ export default {
...
 			this.saveStatuses()
 		},
 		disableStatus(app) {
-			this.enabledStatuses[app] = false
-			const i = this.registeredStatus.findIndex((s) => s === app)
+			let i = this.enabledStatuses.findIndex((s) => s === app)
+			if (i !== -1) {
+				this.enabledStatuses.splice(i, 1)
+			}
+			i = this.registeredStatus.findIndex((s) => s === app)
```
julien-nc, line 406:
```suggestion
			const i = this.enabledStatuses.findIndex((s) => s === app)
			if (i !== -1) {
				this.enabledStatuses.splice(i, 1)
			}
			const j = this.registeredStatus.findIndex((s) => s === app)
```
And adjust the rest of the method.
I had to read it twice to realize why i was not a const. Slightly clearer like this IMO. We can afford declaring one more integer.

julien-nc (changes requested):
Looks good. A few minor change suggestions.

apps/dashboard/appinfo/routes.php
provokateurin: Sure why not, this PR was written before it was in place.

julien-nc (approved):
:+1:

st3iny (approved):
Tested and works.

Nit: I would prefer the new routes to be `/api/v1/...` as they are the first versions of the respective APIs.

provokateurin:
Yeah I first had them as v1, but I'm not sure what makes most sense here. To me it would feel a bit weird that older versions support v2 but not all of the updated v1, so I went with v3 instead :woman_shrugging:

provokateurin:
Just a rebase to fix CI

PR #42977, opened 2024-01-20
theming: Separate `primary` and `background` colors - fix the header menu colors

susnux:
* Fixes #42954
* Fixes #35482
* Fixes #44325

## Summary
This will fix one of the main issues we face with our theming: Currently primary color used to highlight elements is mixed and directly connected to the background color.

So the goal is to have two different colors:
1. `primary` to highlight elements - meaning used as an accent color
2. `background` to be used as background on the body or fallback if images failes to load, moreover this will be used to calculate the header menu text color.

To make this possible this PR is sadly bigger then I hoped, it contains following changes:
1. Add background / primary color for custom images\
Previously we had a primary color predefined for every shipped background image, but not for custom images so that users had to select a color manually. As preparation for the other steps I added a function to automatically calculate the mean color of the image so it can be used as the primary / background color.
2. Separate background and primary color\
While the primary color is intended to highlight elements, it can not always be used as the background color. So now primary is independent from background a user sets, the background color is, if not set directly, calculated as the mean color of the background image. That color is then used as a fallback if the background image could not be loaded and for calculating the color of the text used on the app menu and dashboard (they render directly on the background).
3. Adjust *dashboard* and the *header menu* to use that new color instead of `primary`
4. Adjust theming setting to configure primary and background color
5. Fixed our provided apps that have header menus to use color variables instead of the infamous color-filter.

## TODO

- [x] Tests
- [ ] How to implement migration? IRepairStep? All users need to copy `background_color` to `primary_color` user value

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [ ] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #42978, opened 2024-01-20
fix(unified-search): Refactor header menu code

susnux:
## Summary
1. Adjust styles for the header menu trigger to match the other entries
2. Rename the header menu trigger class so it aligns with other entries
3. Simplify some code (e.g. use `type` on *NcButton* instead of custom styles).

### Screenshots
before|after
---|---
Screenshot_20240120_035700|Screenshot_20240120_040022
Screenshot_20240120_035707|Screenshot_20240120_040029

## Checklist


core/src/views/UnifiedSearchModal.vue
```diff
@@ -542,6 +542,7 @@ export default {
 .unified-search-modal {
 	box-sizing: border-box;
 	height: 100%;
+	min-height: 80vh;
```
emoral435, line 545: is this not already handled by the modal we already have? I do not mind leaving this here just for safe bet, but I do not know if this contributes much, unless I am wrong

emoral435 (approved):

core/src/views/UnifiedSearchModal.vue
susnux: No it is only done on the FilePicker, NcDialog does not set any height.

emoral435: Ahhhh, I see. I thought this was configured by NcDialog, but this is a nice change then! :)

core/src/views/UnifiedSearch.vue
```diff
@@ -62,35 +65,32 @@ export default {
...
+				box-shadow: none !important;
+			}
+
+			&:not(:hover,:focus,:focus-visible) {
+				opacity: .85;
+			}
+
+			&-icon {
+				// ensure the icon has the correct color
+				color: var(--color-background-plain-text) !important;
```
Pytal, line 91: Note for reviewers: new variable in https://github.com/nextcloud/server/pull/42977

Pytal (approved):

nfebe (approved):

core/src/views/UnifiedSearch.vue
susnux: (This was cherry picked from that PR so it gets in here).

But I reverted for current master (use `primary-text`).

PR #42987, opened 2024-01-20
chore(cypress): add 3 more runners

skjnldsv:
Current ci takes 10 min, that should helps


susnux (approved):
Makes sense, we are now at ~4 test files per container.

susnux:
Maybe again look at #41252 how it compares now with this changes here?

PR #42992, opened 2024-01-20
fix(files): Make the navigation reactive to view changes and show also sub routes as active

susnux:
## Summary
This resolves two issues with the files app navigation:

1. The navigation was not reactive as the `currentView` is a computed value but all source objects are **not** reactive (`@nextcloud/files` Navigation). So to ensure changes to the navigation can be detected the object is made observable.
2. Removed the `exact` prop on the app navigation as otherwise it will not show `all files` active if you visited a file like `apps/files/files/123`.
3. Minor type fixes in `Navigation.vue`

#### Before
Pay attention that while the navigation changes the file list does not. Also not that no navigation is active within a subfolder.

https://github.com/nextcloud/server/assets/1855448/2f8f4447-5ca2-45ef-b4c9-1fdd23e9321e

#### After
See that navigation works and the file list is updated + also in subfolders the navigation is kept active.

**Please note: The current version (video) is shown here: https://github.com/nextcloud/server/pull/42992#issuecomment-1902730756 )**

https://github.com/nextcloud/server/assets/1855448/a25aa9fa-7769-477c-9dba-661aa93db49f

## Checklist


apps/files/src/views/Navigation.vue
```diff
@@ -25,9 +25,8 @@
 		<template #list>
 			<NcAppNavigationItem v-for="view in parentViews"
 				:key="view.id"
-				:allow-collapse="true"
+				allow-collapse
```
skjnldsv, line 28: Feels really weird to me :thinking: 
Now it looks like an html attribute instead. I would have kept it as a bool tbh

skjnldsv (dismissed):
> Removed the exact prop on the app navigation as otherwise it will not show all files active if you visited a file like apps/files/files/123.

Now it does this, which is a regression.
I think we need our custom `exact` checker :thinking: 
image

apps/files/src/views/Navigation.vue
ShGKme: IMO, it is a Vue-way, following the HTML standard and being short philosophy.

Like in HTML, Vue proffers us to keep the same HTML-like experience in templates with components.

```html
<input required="required" disabled="disabled" />
<!-- used as -->
<input required disabled />
<!-- same in components -->
<NcInput required disabled />
```

There are even ESLint rules for this, though uncategorized.

https://eslint.vuejs.org/rules/prefer-true-attribute-shorthand.html

> Now it looks like an html attribute instead

I think that there is no such meaning in binding. `v-bind` only means that the value is a dynamic, JS expression, applied to both prop and HTML attributes. And at the same time not having binding like `name="staticString"` doesn't mean that it is an HTML attribute and not a prop.

https://eslint.vuejs.org/rules/no-useless-v-bind.html

susnux:
> Now it does this, which is a regression.
> I think we need our custom `exact` checker 🤔

I do not think we need such a complex solution, I think we can go for the moment with just checking the view we are in.
This fixes the initial issue and will not lead to that regression.

I pushed a fix that only disables `exact` for the files view.

We might later add a `exact` property to the View entry itself to let them decide, but thats up for discussion in a future iteration(?).

susnux:
Looks like this now:

https://github.com/nextcloud/server/assets/1855448/dba75335-405e-44c6-a849-c10e7af14b0a

apps/files/lib/Activity/Helper.php
```diff
@@ -57,26 +58,45 @@ public function getFavoriteFilePaths($user) {
...
+	}
+
+	/**
+	 * Returns an array with the favorites
+	 *
+	 * @param string $user
+	 * @return array
+	 * @throws \RuntimeException when too many or no favorites where found
+	 */
+	public function getFavoriteFilePaths($user) {
```
emoral435, line 87: nitpick: here, for style consistency between functions in this file, shouldn't we have the paremeter (user, in this case) have its type declared? and its return type as well. Bit nitpicky on my part though

`($user)` => `(string $user) : array`

susnux: makes sense :)

apps/files/src/views/favorites.ts
```diff
@@ -31,15 +31,22 @@ import { getContents } from '../services/Favorites'
...
 import logger from '../logger'
 
-export const generateFolderView = function(folder: string, index = 0): View {
+// The return type of the initial state
+interface IFavoriteFolder {
+	fileid: number
+	path: string
+}
+
+export const generateFolderView = function(folder: IFavoriteFolder, index = 0): View {
```
emoral435, line 40: also another nitpick - since this gets the favorite folders, do you think that we should name the function something more descriptive? Something along the lines of `generateFavoriteFolderView`?

apps/files/src/views/Navigation.vue
```diff
@@ -145,27 +147,27 @@ export default {
...
 
-		childViews(): View[] {
+		childViews(): Record<string, View[]> {
 			return this.views
 				// filter parent views
 				.filter(view => !!view.parent)
 				// create a map of parents and their children
 				.reduce((list, view) => {
-					list[view.parent] = [...(list[view.parent] || []), view]
+					list[view.parent!] = [...(list[view.parent!] || []), view]
```
emoral435, line 156: question: ignore if this is not applicable, but are we sure that each view does have a parent? Is it possible that the view could have no parent, for instance, if it was the top level file path (for instance, the view you get when you immediately open `All Files`)?

If this is not applicable, feel free to ignore this :)

apps/files/src/views/Navigation.vue
```diff
@@ -180,6 +182,16 @@ export default {
 	},
 
 	methods: {
+		/**
+		 * Only use exact route matching on routes with child views
+		 * Because if a view does not have children (like the files view) then multiple routes might be matched for it
+		 * Like for the 'files' view this does not work because of optional 'fileid' param so /files and /files/1234 are both in the 'files' view
+		 * @param view The view to check
+		 */
+		useExactRouteMatching(view: View) {
```
emoral435, line 191: addition: since this returns a bool, for code consistancy, `useExactRouteMatching(view: View): bool` makes sense here

emoral435 (commented):
Overall, love this PR. I'm just getting assigned to more files-related issues for the spring sprint, and my advice was to ping you whenever I have a question or draft a PR, so seeing your code genuinely supports my growth here, and I will def ping you to review my PR's because the code quality here is really good. Just a few comments and nitpicks and this is perfect, I tried it on my local dev env and it also works :)

apps/files/src/views/Navigation.vue
skjnldsv: Agree to disagree then :)

susnux: Agree to revert this as unrelated then

skjnldsv:
What I don't get is that it worked before.
v28.0.0 and v28.0.1 doesn't have that reactivity issue.

What broke it in between?

apps/files/src/views/Navigation.vue
susnux: See two lines above: `.filter(view => !!view.parent)` 

After that we are sure :)

apps/files/src/views/Navigation.vue
susnux: It is inferred automatically (your IDE show that it is boolean) but yes be a bit more explicit :)

susnux:
> What I don't get is that it worked before.

I wonder how it worked before because that object was never made reactive?

skjnldsv:
> I wonder how it worked before because that object was never made reactive?

It's a key from an object, which is passed as a reference, it should be reactive by default.
Something changed with the `2.7.16` update, feel free to try it yourself :shrug:

susnux:
> It's a value from an object, which is passed as a reference, it should be reactive by default.

How should it be reactive? As far as I know Vue 2 only the `data` object is made reactive on instantiation by adding `getters` and `setters` for their properties. But not objects which you assign to the vue instance like the `$navigation`?

ShGKme:
> > It's a value from an object, which is passed as a reference, it should be reactive by default.
> 
> How should it be reactive? As far as I know Vue 2 only the `data` object is made reactive on instantiation by adding `getters` and `setters` for their properties. But not objects which you assign to the vue instance like the `$navigation`?

Also any object that was assigned to another reactive object property. So if navigation was assigned to the property of another reactive object anywhere, it also becomes reactive.

A guess - maybe we have components (libs) with different Vue versions built-in. I just found that computed from one component is not reactive in another in Text after provide, but it works in other places.

susnux:
Checked that but our libraries externalize `vue` and in server I checked that there is only one `vue` version used for building.

ShGKme:
> Checked that but our libraries externalize `vue` and in server I checked that there is only one `vue` version used for building.

I always have a lot of `"You are running Vue in development mode."` messages in the console. One Vue package instance makes only one log...

ShGKme:
At least apps like tab apps are separated apps with Vue inside.

emoral435 (approved):
Outside of the weird regression found by @skjnldsv , this PR looks good to me! :)

skjnldsv:
> So if navigation was assigned to the property of another reactive object anywhere, it also becomes reactive.

Yep!
It will make any object reactive if you pass the reference somewhere.
We use it all over the place in sidebar or those old js APIs that worked alongside vue those last few years :shrug:

skjnldsv (approved):

susnux:
Well but does this work with different Vue versions?
Because the `Navigation` is kept globally, so possible different (external) apps can access it and you have no control over the vue version used.

(And I think completely broken when using Vue 3 which uses Proxy thus the original object is never modified).

apps/files/src/main.ts
```diff
@@ -34,7 +34,8 @@ Vue.use(PiniaVuePlugin)
 const pinia = createPinia()
 
 // Init Navigation Service
-const Navigation = getNavigation()
+// This only works with Vue 2 - with Vue 3 this will not modify the source but return just a oberserver
+const Navigation = Vue.observable(getNavigation())
```
ShGKme, line 38: Random thought, wouldn't it be better to return reactive navigation from the `getNavigation`?

apps/files/src/views/favorites.ts
```diff
@@ -118,26 +127,27 @@ export default () => {
...
 				view.order = index
 			}
 		})
 	}
 
 	// Add a folder to the favorites paths array and update the views
-	const addPathToFavorites = function(path: string) {
-		const view = generateFolderView(path)
+	const addToFavorites = function(node: Folder) {
+		const newFavoriteFolder = { path: node.path, fileid: node.fileid! } as IFavoriteFolder
```
ShGKme, line 141:
```suggestion
		const newFavoriteFolder: IFavoriteFolder = { path: node.path, fileid: node.fileid! }
```

ShGKme (commented):
I have `/Foo/Bar/Bar` folders. `Foo` and `Bar` are favorite. before both were listed, now only Bar... Is it intended change?

Before | After
---|---
image | image

susnux:
> I have `/Foo/Bar/Bar` folders. `Foo` and `Bar` are favorite. before both were listed, now only Bar... Is it intended change?

@ShGKme Resolved, a typo in the PHP.

ShGKme (changes requested):
I have `/foo/bar/` both favorite. Initially "All favorites" is open

image

Depending on how I go to the `bar`, I have different result and selected button on the navigation.

If I go from the files list (All favorites is open), I see no content.

Click from the favorites list | Click from the files list
---|---
image | image

susnux:
> Depending on how I go to the `bar`, I have different result and selected button on the navigation.

Resolved this

https://github.com/nextcloud/server/assets/1855448/f273ec43-4cbb-4b8c-81ad-4ddb77726035

ShGKme:
> Resolved this

It still doesn't fully work for me. It fixes the navigation active, but the content is not displayed.

Try to add some content to the "Bar" folder. It must be listed then Favorites/Bar is open from the navigation.

susnux:
I think it is unrelated, but another bug. This also happens on master, the problem here is that `dir` is set to `/bar` while it should be `/foo/bar`

susnux:
@ShGKme pushed a commit that fixes also that issue, you can try again :)

ShGKme (approved):
Works now 🥳

PR #43000, opened 2024-01-21
feat(core): Expose the confirm password endpoint

provokateurin:
## Summary

This allows non-web clients to confirm the password in order to use endpoints that require it.

## Checklist

### DIFF
--- a/core/Controller/LoginController.php
+++ b/core/Controller/LoginController.php
@@ -43,6 +43,7 @@ use OC\User\Session;
 use OC_App;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\NoCSRFRequired;
 use OCP\AppFramework\Http\Attribute\OpenAPI;
 use OCP\AppFramework\Http\Attribute\UseSession;
 use OCP\AppFramework\Http\DataResponse;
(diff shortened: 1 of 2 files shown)


core/Controller/LoginController.php
```diff
@@ -352,13 +354,20 @@ private function createLoginFailedResponse(
 	}
 
 	/**
+	 * Confirm the user password
+	 *
 	 * @NoAdminRequired
 	 * @BruteForceProtection(action=sudo)
 	 *
-	 * @license GNU AGPL version 3 or any later version
```
nickvergessen, line 358: You can not change the license

nickvergessen (changes requested):

core/Controller/LoginController.php
provokateurin: But the license is already in the header? I thought this was a copy paste mistake :woman_shrugging: Does it even make sense to have a specific license for a part of the code in a file (especially since the licenses are the same)?

come-nc (approved):

nickvergessen (changes requested):
Sending auth data should pass the password check:
https://github.com/nextcloud/server/blob/7f2fdd884354a1ab2f07228d7b1c3839139c4378/lib/private/User/Session.php#L605

The problem is you are sending auth + cookies, so we use the cookies only and the line above is never ran. We should fix the PasswordConfirmMiddleware to check if auth data was provided with the request.

provokateurin:
My testing confirms:
- Basic auth without cookies works
- Bearer auth never works (regardless of cookies)

provokateurin:
After the discussions in https://github.com/nextcloud/server/pull/43034 we came to the conclusion that this PR makes more sense and is less susceptible to any additional security holes the other PR could open up.

provokateurin:
Bleh should have re-opened before pushing, will create a new PR

PR #43024, opened 2024-01-22
Consider admin defaults when creating shares

### LINKED ISSUE #42835: [Bug]: Creating new shares doesn't use default admin sharing permissions
### Bug description

Since [Nextcloud 27](https://github.com/nextcloud/server/pull/39472), when you create a new share, the FULL permissions `31` gets used, unless you specifically changes the custom permissions (which then, the default are also NOT used in the pre-checked checkboxes)

### Steps to reproduce

1. Changes default permissions in admin settings
  image
3. Create a user share
  image

### Expected behavior

The new share is created with the default permissions and not `31`

### Additional infos

Issue seems to come from the default permissions set being used here: 
https://github.com/nextcloud/server/blob/e35f6d89a271eede8380e9513811edfe923db4de/apps/files_sharing/src/views/SharingDetailsTab.vue#L755-L762

Historically, we used to sen the permissions as `null` to the Share OCS API and the backend would use the default permissions. Now it's always sent as `31` when untouched by the user.

nfebe:
The current share logic always uses the default `BUNDLED_PERMISSIONS.ALL` which includes everything. This commit updates share creation logic to use `defaultPermissions` if set by admin for the creation of new shares.

Resolves: https://github.com/nextcloud/server/issues/42835


apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -754,9 +754,15 @@ export default {
...
 				if (this.isPublicShare) {
+					// The default sharing permission for public shares should probably
+					// come from the backend via the `this.config`.
 					this.sharingPermission = BUNDLED_PERMISSIONS.READ_ONLY.toString()
 				} else {
 					this.sharingPermission = BUNDLED_PERMISSIONS.ALL.toString()
+					// TEMP COMMENT FOR REVIEW : So apparently, switching to the default permissions
+					// as below is what is required, however, the default permissions does not neccesarily
+					// mean "Allow editing"...
+					// this.sharingPermission = this.config.defaultPermissions
```
Fenn-CS, line 765: So, if we are using default permissions from the API server the words "Allow editing" on the UI becomes false sometimes depending on the default permissions.

Paths:

- Add you option, use default permissions in quick select?
- Update the text based on calculated permissions
- Leave as it is.

cc: @artonge @sorbaugh

nfebe (commented):
Default permissions might not always be same as "Allow editing" (or basically all permissions)

Screenshot from 2024-01-23 12-27-32

apps/files_sharing/src/views/SharingDetailsTab.vue
JSoko: That's right. Why don't rename it to default permissions?

sorbaugh: What do you think @jancborchardt, renaming "Allow editing" to cover what "Default permission" means in the given instance might be the safest option.

Fenn-CS: Just like with "Custom permissions", we could have a sub title to show what is included in the "Default permissions"

Fenn-CS:
Something else that has come up while working on this in relation to maintaining "Allow editing" and "Default permissions"

- It turns out when no permissions are set on the admin, the default permission is "1" which is ready only. So changing "Allow editing" to "Default permissions" which has the probability to be 1, most of the time would make little sense with read only just at the top, that is, "Read only" (first option), second option (Default permissions, still read only)

Screenshot from 2024-01-25 11-29-58

So, the brainstorming is not over it seems, the second path, is simply to add "Default permissions" as a new option IF it is different from READ only (`1`) as well as different from "Allow editing" (`31`)

Fenn-CS:
cc: @jancborchardt

apps/files_sharing/src/views/SharingDetailsTab.vue
jancborchardt: If custom permissions are set by the admin, then we should use the "Custom permissions" entry for that. Adding another "Default permissions" entry would be a bit confusing.

"View only" and "Allow editing" should always stay as is since they are shorthands to quickly reach the 2 most common settings modes.

Fenn-CS: Thanks @jancborchardt what about making it more dynamic? See comment in main thread https://github.com/nextcloud/server/pull/43024#issuecomment-1909859377

jancborchardt:
As far as I can see, the issue reported at https://github.com/nextcloud/server/issues/42835 should be fixed via code and does not need a new entry in this list of presets.
These are merely default share permissions set by the admin and that’s that.

- If the permissions set by the admin are the same as "View only", preselect that
- If the permissions set by the admin are the same as "Allow editing", preselect that
- If the permissions set by the admin are anything else, preselect "Custom permissions" with the relevant subline

But we should not replace an entry of those with "Default permissions" or anything like that.

Fenn-CS:
> As far as I can see, the issue reported at https://github.com/nextcloud/server/issues/42835 should be fixed via code and does not need a new entry in this list of presets.
> These are merely default share permissions set by the admin and that’s that.

Sure, it does not necessarily need a new entry in the list of presets. But definitely needs making changes to how the presets work to avoid confusion. Adding another option or updating language are just other solutions paths suggested. 

So the new implication is that "Allow editing" is not always the selected/default option as is the case now. It would be dynamically selected.

The only difference between this and my last suggestion is that no new entry would be added. To mitigate the issue of clearly understanding what default permissions are, I think it would be nice to have a way for the user to know that  whatever permission was pres-elected for them is the default. This can be done by simply adding something like `Default` in parenthesis. `(Default)`

apps/files_sharing/lib/Listener/LoadSidebarListener.php
```diff
@@ -51,6 +51,7 @@ public function handle(Event $event): void {
 
 		$shareConfig = [
 			'allowPublicUploads' => $this->shareManager->shareApiLinkAllowPublicUpload(),
+			'defaultPermissions' => $this->shareManager->shareApiDefaultPermissions(),
```
skjnldsv, line 54: This is already exposed through capabilities
- https://github.com/nextcloud/server/blob/c6806684d2deee11b64fb8250064115cea291659/apps/files_sharing/lib/Capabilities.php#L179
- https://nextcloud-libraries.github.io/nextcloud-capabilities/functions/getCapabilities.html

skjnldsv: `allowPublicUploads` is too btw https://github.com/nextcloud/server/blob/c6806684d2deee11b64fb8250064115cea291659/apps/files_sharing/lib/Capabilities.php#L162

artonge (approved):

susnux (approved):

susnux:
> /compile

CI is a bit slow at the moment due to 28.0.2 release I guess, you could short cut by compiling locally and submit the assets.

emoral435 (approved):

Fenn-CS:
PHP unit failures unrelated.

PR #43025, opened 2024-01-22
fix: Add command to update re-share if shared-by user has been revoked

### LINKED ISSUE #42407: [Bug]: Sub folder shares were broken on ownership transfer

### Bug description

After an ownership transfer (via the web interface), sub shares can be broken and it's impossible to restore.

### Steps to reproduce

- userA shares a folder D1 with userB
- userB creates and shares a subfolder D2 and shares it with userC
- userA revoke the share of D1 with userB
- userA transfers ownership of D1 to userD

After userD accepts the transfer:
- userC loses access to the folder D2
- if userD tries to share the folder D2 with userC : error message "Node for share not found, fileid: 132472"

### Expected behavior

My suggestion for expected behavior:
Re-share should be deleted if restore failed to not block creating new share for the same user.

### Installation method

Community Docker image

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

luka-nextcloud:
* Resolves: #42407

## Summary
Add command `occ sharing:fix-broken-shares` to fix the shares that were broken on transfer ownership.

## Checklist

### DIFF
--- a/apps/files_sharing/lib/OrphanHelper.php
+++ b/apps/files_sharing/lib/OrphanHelper.php
@@ -10,6 +10,7 @@ namespace OCA\Files_Sharing;
 
 use OC\User\NoUserException;
 use OCP\DB\QueryBuilder\IQueryBuilder;
+use OCP\Files\Config\IUserMountCache;
 use OCP\Files\IRootFolder;
 use OCP\IDBConnection;
 
@@ -17,6 +18,7 @@ class OrphanHelper {
 	public function __construct(
 		private IDBConnection $connection,
 		private IRootFolder $rootFolder,
+		private IUserMountCache $userMountCache,
 	) {
 	}
 
@@ -67,5 +69,27 @@ class OrphanHelper {
 				'target' => (string)$row['file_target'],
 			];
 		}
+	}
+
+	public function findOwner(int $fileId): ?string {
+		$mounts = $this->userMountCache->getMountsForFileId($fileId);
+		if (!$mounts) {
+			return null;
+		}
+		foreach ($mounts as $mount) {
+			$userHomeMountPoint = '/' . $mount->getUser()->getUID() . '/';
+			if ($mount->getMountPoint() === $userHomeMountPoint) {
+				return $mount->getUser()->getUID();
+			}
+		}
+		return null;
+	}
+
+	public function updateShareOwner(int $shareId, string $owner): void {
+		$query = $this->connection->getQueryBuilder();
+		$query->update('share')
+			->set('uid_owner', $query->createNamedParameter($owner))
+			->where($query->expr()->eq('id', $query->createNamedParameter($shareId, IQueryBuilder::PARAM_INT)));
+		$query->executeStatement();
 	}
 }
--- a/apps/files_sharing/appinfo/info.xml
+++ b/apps/files_sharing/appinfo/info.xml
@@ -49,6 +49,7 @@ Turning the feature off removes shared files and folders on the server for all s
 		<command>OCA\Files_Sharing\Command\CleanupRemoteStorages</command>
 		<command>OCA\Files_Sharing\Command\ExiprationNotification</command>
 		<command>OCA\Files_Sharing\Command\DeleteOrphanShares</command>
+		<command>OCA\Files_Sharing\Command\FixShareOwners</command>
 	</commands>
 
 	<settings>
--- a/apps/files_sharing/composer/composer/autoload_classmap.php
+++ b/apps/files_sharing/composer/composer/autoload_classmap.php
@@ -27,6 +27,7 @@ return array(
     'OCA\\Files_Sharing\\Command\\CleanupRemoteStorages' => $baseDir . '/../lib/Command/CleanupRemoteStorages.php',
     'OCA\\Files_Sharing\\Command\\DeleteOrphanShares' => $baseDir . '/../lib/Command/DeleteOrphanShares.php',
     'OCA\\Files_Sharing\\Command\\ExiprationNotification' => $baseDir . '/../lib/Command/ExiprationNotification.php',
+    'OCA\\Files_Sharing\\Command\\FixShareOwners' => $baseDir . '/../lib/Command/FixShareOwners.php',
     'OCA\\Files_Sharing\\Controller\\AcceptController' => $baseDir . '/../lib/Controller/AcceptController.php',
     'OCA\\Files_Sharing\\Controller\\DeletedShareAPIController' => $baseDir . '/../lib/Controller/DeletedShareAPIController.php',
     'OCA\\Files_Sharing\\Controller\\ExternalSharesController' => $baseDir . '/../lib/Controller/ExternalSharesController.php',
--- a/apps/files_sharing/composer/composer/autoload_static.php
+++ b/apps/files_sharing/composer/composer/autoload_static.php
@@ -42,6 +42,7 @@ class ComposerStaticInitFiles_Sharing
         'OCA\\Files_Sharing\\Command\\CleanupRemoteStorages' => __DIR__ . '/..' . '/../lib/Command/CleanupRemoteStorages.php',
         'OCA\\Files_Sharing\\Command\\DeleteOrphanShares' => __DIR__ . '/..' . '/../lib/Command/DeleteOrphanShares.php',
         'OCA\\Files_Sharing\\Command\\ExiprationNotification' => __DIR__ . '/..' . '/../lib/Command/ExiprationNotification.php',
+        'OCA\\Files_Sharing\\Command\\FixShareOwners' => __DIR__ . '/..' . '/../lib/Command/FixShareOwners.php',
         'OCA\\Files_Sharing\\Controller\\AcceptController' => __DIR__ . '/..' . '/../lib/Controller/AcceptController.php',
         'OCA\\Files_Sharing\\Controller\\DeletedShareAPIController' => __DIR__ . '/..' . '/../lib/Controller/DeletedShareAPIController.php',
         'OCA\\Files_Sharing\\Controller\\ExternalSharesController' => __DIR__ . '/..' . '/../lib/Controller/ExternalSharesController.php',
--- a/apps/files_sharing/lib/Command/FixShareOwners.php
+++ b/apps/files_sharing/lib/Command/FixShareOwners.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Files_Sharing\Command;
+
+use OC\Core\Command\Base;
+use OCA\Files_Sharing\OrphanHelper;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class FixShareOwners extends Base {
+	public function __construct(
+		private readonly OrphanHelper $orphanHelper,
+	) {
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('sharing:fix-share-owners')
+			->setDescription('Fix owner of broken shares after transfer ownership on old versions')
+			->addOption(
+				'dry-run',
+				null,
+				InputOption::VALUE_NONE,
+				'only show which shares would be updated'
+			);
+	}
+
+	public function execute(InputInterface $input, OutputInterface $output): int {
+		$shares = $this->orphanHelper->getAllShares();
+		$dryRun = $input->getOption('dry-run');
+		$count = 0;
+
+		foreach ($shares as $share) {
+			if ($this->orphanHelper->isShareValid($share['owner'], $share['fileid']) || !$this->orphanHelper->fileExists($share['fileid'])) {
+				continue;
+			}
+
+			$owner = $this->orphanHelper->findOwner($share['fileid']);
+
+			if ($owner !== null) {
+				if ($dryRun) {
+					$output->writeln("Share with id <info>{$share['id']}</info> (target: <info>{$share['target']}</info>) can be updated to owner <info>$owner</info>");
+				} else {
+					$this->orphanHelper->updateShareOwner($share['id'], $owner);
+					$output->writeln("Share with id <info>{$share['id']}</info> (target: <info>{$share['target']}</info>) updated to owner <info>$owner</info>");
+				}
+				$count++;
+			}
+		}
+
+		if ($count === 0) {
+			$output->writeln('No broken shares detected');
+		}
+
+		return static::SUCCESS;
+	}
+}
--- a/apps/files_sharing/tests/Command/FixShareOwnersTest.php
+++ b/apps/files_sharing/tests/Command/FixShareOwnersTest.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * SPDX-FileCopyrightText: 2017-2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+namespace OCA\Files_Sharing\Tests\Command;
+
+use OCA\Files_Sharing\Command\FixShareOwners;
+use OCA\Files_Sharing\OrphanHelper;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Test\TestCase;
+
+/**
+ * Class FixShareOwnersTest
+ *
+ * @package OCA\Files_Sharing\Tests\Command
+ */
+class FixShareOwnersTest extends TestCase {
+	/**
+	 * @var FixShareOwners
+	 */
+	private $command;
+
+	/**
+	 * @var OrphanHelper|\PHPUnit\Framework\MockObject\MockObject
+	 */
+	private $orphanHelper;
+
+	protected function setUp(): void {
+		parent::setUp();
+
+		$this->orphanHelper = $this->createMock(OrphanHelper::class);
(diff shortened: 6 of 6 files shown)


lib/private/Share20/Manager.php
```diff
@@ -273,6 +273,12 @@ protected function generalCreateChecks(IShare $share) {
 		// And you cannot share your rootfolder
 		if ($this->userManager->userExists($share->getSharedBy())) {
 			$userFolder = $this->rootFolder->getUserFolder($share->getSharedBy());
+
+			// Share record of SharedBy user has been revoked or deleted
+			if (empty($userFolder->getById($share->getNode()->getId()))) {
+				$userFolder = $this->rootFolder->getUserFolder($share->getShareOwner());
+				$share->setSharedBy($share->getShareOwner());
+			}
```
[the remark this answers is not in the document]
luka-nextcloud: I don't think deleting all shares of whole re-share hierarchy is a good approach. It might cause performance issue if sharing item is a folder that has a lot of children. So, I will come up with another fix.

luka-nextcloud:
@artonge I've updated as you requested, please check again.

artonge (commented):
Can you add comments to clarify what each `if` group does?

I am also wondering if we could find this shares with the following query:

```sql
SELECT
	f.fileid,
	f.path,
	f.storage,
	s.id as share_id,
	s.uid_owner as share_owner,
	s.uid_initiator as share_initiator,
	s.share_with as share_recipient
FROM
	oc_filecache f
	JOIN oc_share s ON f.fileid = s.file_source
	AND s.uid_initiator NOT IN (
		SELECT
			user_id
		FROM
			oc_mounts m
		WHERE
			f.storage = m.storage_id
	)
```

If so, then we could have a background job to remove them every hour or so like https://github.com/nextcloud/server/blob/master/apps/files_sharing/lib/DeleteOrphanedSharesJob.php, which might be easier than the current solution.

luka-nextcloud:
> Can you add comments to clarify what each `if` group does?
> 
> I am also wondering if we could find this shares with the following query:
> 
> ```sql
> SELECT
> 	f.fileid,
> 	f.path,
> 	f.storage,
> 	s.id as share_id,
> 	s.uid_owner as share_owner,
> 	s.uid_initiator as share_initiator,
> 	s.share_with as share_recipient
> FROM
> 	oc_filecache f
> 	JOIN oc_share s ON f.fileid = s.file_source
> 	AND s.uid_initiator NOT IN (
> 		SELECT
> 			user_id
> 		FROM
> 			oc_mounts m
> 		WHERE
> 			f.storage = m.storage_id
> 	)
> ```
> 
> If so, then we could have a background job to remove them every hour or so like https://github.com/nextcloud/server/blob/master/apps/files_sharing/lib/DeleteOrphanedSharesJob.php, which might be easier than the current solution.

@artonge It only works if the shared-by user refreshes his files list after his share has revoked. So, I don't think this query would do the job.

apps/files_sharing/lib/Command/FixBrokenShares.php
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
come-nc, line 22:
```suggestion
/**
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```

apps/files_sharing/lib/Command/FixBrokenShares.php
```diff
@@ -0,0 +1,63 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class FixBrokenShares extends Base {
+	private OrphanHelper $orphanHelper;
+
+	public function __construct(OrphanHelper $orphanHelper) {
+		parent::__construct();
+		$this->orphanHelper = $orphanHelper;
+	}
```
come-nc, line 37:
```suggestion
	public function __construct(
		private OrphanHelper $orphanHelper,
	) {
		parent::__construct();
	}
```

apps/files_sharing/lib/Command/FixBrokenShares.php
```diff
@@ -0,0 +1,63 @@
...
+
+			$owner = $this->orphanHelper->findOwner($share['fileid']);
+
+			if (!empty($owner)) {
+				$this->orphanHelper->updateShareOwner($share['id'], $owner);
+				$output->writeln("Share {$share['id']} updated to owner $owner");
+			}
+		}
+
+		return 0;
```
come-nc, line 61:
```suggestion
		return static::SUCCESS;
```

apps/files_sharing/lib/OrphanHelper.php
```diff
@@ -88,4 +92,25 @@ public function getAllShares() {
...
 	}
+
+	public function findOwner(int $fileId): ?string {
+		$mounts = $this->userMountCache->getMountsForFileId($fileId);
+		if (!$mounts) {
+			return null;
+		}
+		foreach ($mounts as $mount) {
+			if ($mount->getInternalPath()) {
+				return $mount->getUser()->getUID();
```
come-nc, line 103: Can you explain the logic in a comment here?
A file can have several owners and only one of them will be returned randomly, right? We may end up with less permissions than expected, no?

apps/files_sharing/lib/OrphanHelper.php
```diff
@@ -88,4 +92,26 @@ public function getAllShares() {
...
+			}
+		}
+		return null;
+	}
+
+	public function updateShareOwner(int $shareId, string $owner): void {
+		$query = $this->connection->getQueryBuilder();
+		$query->update('share')
+			->set('uid_owner', $query->createNamedParameter($owner))
+			->where($query->expr()->eq('id', $query->createNamedParameter($shareId)));
```
come-nc, line 114:
```suggestion
			->where($query->expr()->eq('id', $query->createNamedParameter($shareId, IQueryBuilder::PARAM_INT)));
```

apps/files_sharing/lib/Command/FixBrokenShares.php
```diff
@@ -0,0 +1,59 @@
...
+		$dryRun = $input->getOption('dry-run');
+
+		foreach ($shares as $share) {
+			if ($this->orphanHelper->isShareValid($share['owner'], $share['fileid']) || !$this->orphanHelper->fileExists($share['fileid'])) {
+				continue;
+			}
+
+			$owner = $this->orphanHelper->findOwner($share['fileid']);
+
+			if (!empty($owner)) {
```
come-nc, line 47:
```suggestion
			if ($owner !== null) {
```
`'0'` is a valid userid.

lib/private/Share20/Manager.php
```diff
@@ -1117,6 +1117,71 @@ protected function deleteChildren(IShare $share) {
...
 
+	protected function deleteReshare(IShare $share) {
+		// If the user has another shares, we don't delete the shares by this user
+		if ($share->getShareType() === IShare::TYPE_USER) {
+			$groupShares = $this->getSharedWith($share->getSharedWith(), IShare::TYPE_GROUP, $share->getNode(), -1, 0);
+
+			if (count($groupShares) !== 0) {
+				return;
+			}
+		}
```
come-nc, line 1128: This checks the user does not have access to the same file/folder through another share, right?

But it does not check the same thing for subfiles/subfolders, while still deleting shares of children below.

This PR should contains tests showing behavior in different scenarios where shares should be deleted or should not to validate that it behaves correctly.

come-nc (changes requested):

luka-nextcloud:
@come-nc Could you please review again? All checks have passed now.

apps/files_sharing/lib/OrphanHelper.php
```diff
@@ -10,19 +10,23 @@
...
 class OrphanHelper {
 	private IDBConnection $connection;
 	private IRootFolder $rootFolder;
+	private IUserMountCache $userMountCache;
 
 	public function __construct(
 		IDBConnection $connection,
-		IRootFolder $rootFolder
+		IRootFolder $rootFolder,
+		IUserMountCache $userMountCache
```
come-nc, line 25:
```suggestion
		IUserMountCache $userMountCache,
```

lib/private/Share20/Manager.php
```diff
@@ -1117,6 +1118,108 @@ protected function deleteChildren(IShare $share) {
...
+
+		// Delete re-share records (shared by "share with user") inside folder
+		if ($share->getNodeType() === 'folder' && $share->getShareType() === IShare::TYPE_USER) {
+			$sharesInFolder = $this->getSharesInFolder($share->getSharedWith(), $node, true);
+
+			foreach ($sharesInFolder as $shares) {
+				foreach ($shares as $child) {
+					$this->deleteShare($child);
```
come-nc, line 1160: For subfiles you did not check that the user still has access to it, I think?

lib/private/Share20/Manager.php
```diff
@@ -1117,6 +1118,108 @@ protected function deleteChildren(IShare $share) {
...
+
+					if (count($groupShares) !== 0 || count($userShares) !== 0) {
+						return;
+					}
+
+					$parentNode = $parentNode->getParent();
+				}
+			} catch (NotFoundException) {
+			}
+		}
```
come-nc, line 1152: Instead of doing all this to check if the user still has access to the file or not, does it not make more sense to directly check if the user has access to the file? We must have some tooling for that, no?

luka-nextcloud: @come-nc  I cannot find another way to check this logic. Could you provide some suggestions?

come-nc: Get the user folder of the user and use `getFirstNodeById` would work I think.

come-nc: (Note that this would only work after deleting the share)

come-nc:
@luka-nextcloud After talking with Robin, actually the ownership transfer is supposed to update the owner of the share so there should be no problem in the first place?

luka-nextcloud:
@come-nc Does it mean that these changes are enough for this issue?

image

image

come-nc:
> @come-nc Does it mean that these changes are enough for this issue?
> 
> image
> 
> image

No, the second diff looks like it will cause trouble by arbitrary changing sharedBy of shares.

I lost track of what you are trying to fix, can you explain how to reproduce your problem? Why do you remove the exception handling removing invalid shares in OwnershipTransferService?

luka-nextcloud:
@come-nc
- Reproduce steps:
1. UserA share Folder1 to UserB
2. UserB share Folder1 to UserC
3. UserA deleted sharing to UserB
4. UserA transfer Folder1 to UserD
-> UserC lost access to Folder1. When UserC open sharing tab of Folder1, the error message `Error: Could not get proper share mount for . Failing since else the next calls are called with null` returned.

- Issue occurred because the sharing of UserB to UserC was not deleted or updated after UserA deleted sharing to UserB. Then, when UserA transfering Folder1 to UserD, that re-share record cannot be transferred because it throws `GenericShareException` error.

come-nc:
> @come-nc
> 
> * Reproduce steps:
> 
> 1. UserA share Folder1 to UserB
> 2. UserB share Folder1 to UserC
> 3. UserA deleted sharing to UserB
> 4. UserA transfer Folder1 to UserD
>    -> UserC lost access to Folder1. When UserC open sharing tab of Folder1, the error message `Error: Could not get proper share mount for . Failing since else the next calls are called with null` returned.
> 
> * Issue occurred because the sharing of UserB to UserC was not deleted or updated after UserA deleted sharing to UserB. Then, when UserA transfering Folder1 to UserD, that re-share record cannot be transferred because it throws `GenericShareException` error.

UserC already loses access to Folder1 after step 3, right?
At that point the sharing tab shows no error?

luka-nextcloud:
@come-nc 
> UserC already loses access to Folder1 after step 3, right?

No, UserC didn't lose access to Folder1 after step3. This is the issue we are trying to fix.

> At that point the sharing tab shows no error?

No error.

luka-nextcloud:
@come-nc 
**Original problem:**
Re-share was not deleted when parent share deleted. Then, it caused error while transferring ownership.

**Reproduce steps:**
Mentioned above.

**Solution:**
Delete the re-share and its children when parent share is deleted.

lib/private/Share20/Manager.php
```diff
@@ -1016,6 +1017,110 @@ protected function deleteChildren(IShare $share) {
...
+
+				if ($share->getNodeType() === 'folder') {
+					$sharesInFolder = $this->getSharesInFolder($user->getUID(), $node, true);
+
+					foreach ($sharesInFolder as $shares) {
+						foreach ($shares as $child) {
+							if ($this->hasNodeAccess($user->getUID(), $child->getNode())) {
+								continue;
+							}
+							$this->deleteShare($child);
```
come-nc, line 1093: Why is generalCreateCheck method not directly run on `$child` instead?
I do not understand the logic in hasNodeAccess, it searches for other shares of the same node but only directly to the user?

Also, the code is duplicated between user shares and group shares, please first compute an array of target users and then loop on it.

luka-nextcloud: @come-nc Ok, I have checked again. The logic in `hasNodeAccess` was not really correct. It missed some cases. So, I removed that function and updated approach for `deleteReshare`. First collect related user ids, then collect the re-share by those users, finally do `generalCreateCheck` on those re-share records.

lib/private/Share20/Manager.php
```diff
@@ -1016,6 +1017,71 @@ protected function deleteChildren(IShare $share) {
...
+		}
+
+		$reshareRecords = [];
+		$shareTypes = [
+			IShare::TYPE_GROUP,
+			IShare::TYPE_USER,
+			IShare::TYPE_LINK,
+			IShare::TYPE_REMOTE,
+			IShare::TYPE_EMAIL
+		];
```
come-nc, line 1054: From what I see in `getSharesInFolder` you can use `$providers = $this->factory->getAllProviders();` to directly get all providers without caring for the share type. Unless we explicitely only want to fetch for some type here?

lib/private/Share20/Manager.php
```diff
@@ -1016,6 +1017,71 @@ protected function deleteChildren(IShare $share) {
...
+
+		foreach ($userIds as $userId) {
+			foreach ($shareTypes as $shareType) {
+				$provider = $this->factory->getProviderForType($shareType);
+				$shares = $provider->getSharesBy($userId, $shareType, $node, false, -1, 0);
```
come-nc, line 1059: Why do you pass `false` for `$reshares` flag here but `true` in the call to `getSharesInFolder` below?

apps/files_sharing/lib/OrphanHelper.php
```diff
@@ -73,4 +77,26 @@ public function getAllShares() {
...
 		}
 	}
+
+	public function findOwner(int $fileId): ?string {
+		$mounts = $this->userMountCache->getMountsForFileId($fileId);
+		if (!$mounts) {
+			return null;
+		}
+		foreach ($mounts as $mount) {
+			// Only the mount of owner has the internal path value
```
icewind1991, line 87: I don't think this is true

lib/private/Share20/Manager.php
luka-nextcloud: We still need share type for calling `getSharesBy()` at line 1059.

apps/files_sharing/lib/OrphanHelper.php
luka-nextcloud: @icewind1991 Could you provide a suggestion?

lib/private/Share20/Manager.php
come-nc: Indeed, I missed that.

apps/files_sharing/lib/OrphanHelper.php
icewind1991: You can check if `$mount` is a `HomeMountPoint` and then use `$mount->getUser()`

apps/files_sharing/lib/OrphanHelper.php
```diff
@@ -68,4 +70,26 @@ public function getAllShares() {
...
 	}
+
+	public function findOwner(int $fileId): ?string {
+		$mounts = $this->userMountCache->getMountsForFileId($fileId);
+		if (!$mounts) {
+			return null;
+		}
+		foreach ($mounts as $mount) {
+			$userHomeMountPoint = '/' . $mount->getUser()->getUID() . '/';
+			if ($mount->getMountPoint() === $userHomeMountPoint) {
```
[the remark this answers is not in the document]
come-nc: I was wrong, $mount here is actually an instance of CachedMountFileInfo and it seems there’s no easy way to get a MountPoint/HomeMountPoint instead.

Let’s keep it that way if it works.
I’d like @icewind1991 to approve this just in case there is a corner case we missed.

come-nc (approved):

apps/files_sharing/lib/OrphanHelper.php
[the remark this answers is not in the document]
icewind1991: Didn't realize `$mount` is a `CachedMountInfo`, then checking the mountpoint is probably the easiest.

icewind1991 (approved):

PR #43027, opened 2024-01-22
fix: Use viewBox instead of viewbox in SVGs

provokateurin:
## Summary

The correct attribute name is `viewBox` and not `viewbox`: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox

## Checklist


bigcat88 (approved):

come-nc (approved):
Should we have some kind of lint svg CI workflow?

nickvergessen:
> Should we have some kind of lint svg CI workflow?

We should at least make sure this here is not breaking it again any time soon:
https://github.com/nextcloud/server/blob/master/build/image-optimization.sh#L89-L97

We can of course also run that script on CI and ensure an empty diff. But I would see it as a different task

provokateurin:
> We should at least make sure this here is not breaking it again any time soon

I ran it, it didn't modify the attribute :+1:

PR #43032, opened 2024-01-22
[stable28] fix(theming): Apply same logic on capabilities for primary color as done on themes

susnux:
* Backport of https://github.com/nextcloud/server/pull/43033

## Summary
On themes we replace the default color with the accessible color, so we also need to do this on the capabilities.
This worked previously but as we now enforce contrasts of the text color clients that use the capabilities changes to black text on that color.

## Checklist

### DIFF
--- a/apps/theming/lib/Capabilities.php
+++ b/apps/theming/lib/Capabilities.php
@@ -94,7 +94,11 @@ class Capabilities implements IPublicCapability {
 	 */
 	public function getCapabilities() {
 		$color = $this->theming->getDefaultColorPrimary();
-		$colorText = $this->theming->getDefaultTextColorPrimary();
+		// Same as in DefaultTheme
+		if ($color === BackgroundService::DEFAULT_COLOR) {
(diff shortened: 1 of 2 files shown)


sorbaugh (approved):

SystemKeeper:
For my understanding, why not adjust this in

https://github.com/nextcloud/server/blob/64ccc1ef87f095957bb6d169d7099e8c1976bf67/apps/theming/lib/ThemingDefaults.php#L508-L510

🤔

susnux:
> For my understanding, why not adjust this in

Better would be in `getColorPrimary`, but that would have greater consequences especially for a stable branch I would not change that API and just fix the usage here.

For a general fix I currently prepare this one: #42977

Pytal (approved):

sorbaugh (approved):

blizzz:
Test is still failing unfortunately

susnux:
cypress unrelated

skjnldsv (approved):

PR #43033, opened 2024-01-22
fix(theming): Apply same logic on capabilities for primary color as done on themes

susnux:
## Summary
On themes we replace the default color with the accessible color, so we also need to do this on the capabilities.
This worked previously but as we now enforce contrasts of the text color clients that use the capabilities changes to black text on that color.

## Checklist


Pytal (approved):

skjnldsv (approved):

provokateurin (approved):

susnux:
Maybe we should after all fix this in `OCA\ThemingDefaults` instead. Because that would fix #43043 
(even if that means more tests need to be adjusted, as we basically then will nowhere have the default color anymore but just the default accessible color).

PR #43045, opened 2024-01-23
fix(apps): Fix type information of app manager

nickvergessen:
## Checklist

### DIFF
--- a/lib/private/App/AppManager.php
+++ b/lib/private/App/AppManager.php
@@ -289,7 +289,7 @@ class AppManager implements IAppManager {
 	 * Check if an app is enabled for user
 	 *
 	 * @param string $appId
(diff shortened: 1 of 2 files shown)


provokateurin (changes requested):
Please add the return types to the method head (both interface and implementation).

nickvergessen:
> Please add the return types to the method head (both interface and implementation).

Breaks public API, so not for now

ChristophWurst (commented):
Looks good. Context? Is this a regression?

come-nc (approved):

provokateurin:
> Breaks public API, so not for now

Why would it break something? If the data returned matches the type then it just adds additional information?

provokateurin (approved):

ChristophWurst (approved):

Issue #7995, opened 2018-01-22
Chunked transfers lead to 0 byte files


ylangisc:
We hit the annoying 0 byte files issue when using a `PUT` operation with a chunked transfer encoding to modify an existing 0 byte file. After successfully executing the `PUT` the server replies with a 204 but the file is still empty. Not sure if this is a NC or sabre/dav issue.

```
curl -v -X PUT --header "Transfer-Encoding: chunked" -d @report.csv "http://ubuntu.local/remote.php/webdav/ylatst.txt" -u admin:admin
*   Trying 192.168.48.135...
* TCP_NODELAY set
* Connected to ubuntu.local (192.168.48.135) port 80 (#0)
* Server auth using Basic with user 'admin'
> PUT /remote.php/webdav/ylatst.txt HTTP/1.1
> Host: ubuntu.local
> Authorization: Basic YWRtaW46YWRtaW4=
> User-Agent: curl/7.54.0
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/x-www-form-urlencoded
> Expect: 100-continue
> 
< HTTP/1.1 100 Continue
< HTTP/1.1 204 No Content
< Server: nginx/1.13.8
< Date: Mon, 22 Jan 2018 16:41:33 GMT
< Content-Type: text/html; charset=UTF-8
< Content-Length: 0
< Connection: keep-alive
< Expires: Thu, 19 Nov 1981 08:52:00 GMT
< Cache-Control: no-store, no-cache, must-revalidate
< Pragma: no-cache
< Set-Cookie: oc_sessionPassphrase=xlmQsVcgE%2FcbE8%2FmB8bvsVG3fGrkDGB%2Fg5wHy5sJqrIjfeVotxGf4US%2BnLdGj6xNmzPjH3NzYgm09Qy%2FnkqKQA59LI9qQEmY0mkZZs17wwCvcubybbqiCS83FHYGdvhe; path=/; HttpOnly
< X-Frame-Options: SAMEORIGIN
< X-XSS-Protection: 1; mode=block
< X-Content-Type-Options: nosniff
< X-Robots-Tag: none
< X-Download-Options: noopen
< X-Permitted-Cross-Domain-Policies: none
< Set-Cookie: nc_sameSiteCookielax=true; path=/; httponly;expires=Fri, 31-Dec-2100 23:59:59 GMT; SameSite=lax
[...]

PR #43047, opened 2024-01-23
Buffer chunked requests

artonge:
Clients like DAVx5 on Android, or Cyberduck, use chunked requests.
When FastCGI or FPM is used with apache, requests arrive to Nextcloud without any content.
This leads to the creation of empty files.
The following directive will force the problematic requests to be buffered before being forwarded to Nextcloud.
This way, the "Transfer-Encoding" header is removed, the "Content-Length" header is set, and the request content is proxied to Nextcloud.
Here are more information about the issue:
 - https://docs.cyberduck.io/mountainduck/issues/fastcgi/
 - https://docs.nextcloud.com/server/latest/admin_manual/issues/general_troubleshooting.html#troubleshooting-webdav

Tested for a few weeks by a customer without any issue.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -94,5 +94,19 @@
   RewriteRule ^(?:\.(?!well-known)|autotest|occ|issue|indie|db_|console).* - [R=404,L]
 </IfModule>
 
+# Clients like xDavv5 on Android, or Cyberduck, use chunked requests.
+# When FastCGI or FPM is used with apache, requests arrive to Nextcloud without any content.
+# This leads to the creation of empty files.
+# The following directive will force the problematic requests to be buffered before being forwarded to Nextcloud.
+# This way, the "Transfer-Encoding" header is removed, the "Content-Length" header is set, and the request content is proxied to Nextcloud.
+# Here are more information about the issue:
+#  - https://docs.cyberduck.io/mountainduck/issues/fastcgi/
+#  - https://docs.nextcloud.com/server/latest/admin_manual/issues/general_troubleshooting.html#troubleshooting-webdav
+<IfModule setenvif.c>
+  <Location "/remote.php">
+    SetEnvIf Transfer-Encoding "chunked" proxy-sendcl=1
+  </Location>
+</IfModule>
+
 AddDefaultCharset utf-8
 Options -Indexes


szaimen:
I guess I can remove this from aio then again? :)
https://github.com/nextcloud/all-in-one/blob/b17561d41e575c7efd45fac1a6fb410be79bbe37/Containers/apache/nextcloud.conf#L42

artonge:
> I guess I can remove this from aio then again? :)

I guess yes :) The current PR also more narrow so only changes the behavior for problematic requests.

SystemKeeper:
Would this also need to be added to https://github.com/nextcloud/server/blob/master/lib/private/Setup.php#L506 ?

joshtrichards:
I like the targeted approach! 👍

How does mod_php avoid this? Does it just buffer every transaction?

If I'm understanding this correctly:

- this applies to fpm only
- this could have implications for resource usage on proxies (but since this is narrowly targeted and these transactions failed before I guess it doesn't matter too much?)

Related issues:

- #7995
- #3056

If we proceed with this:

- We should probably document it somewhere, particularly if there's a chance for resource usage trade-offs and/or if mod_php might be preferable in some environments over using this solution
- Probably some work required for parity within the Nginx example config we provide in the Admin Manual since we explicitly disable the `fastcgi_request_buffering` behavior there
- Might be some other touch spots in the docs - e.g. we reference the Sabre URL about 0-byte files currently in the Admin Manual (https://docs.nextcloud.com/server/latest/admin_manual/issues/general_troubleshooting.html#troubleshooting-webdav)

artonge:
> How does mod_php avoid this? Does it just buffer every transaction?

From what I understand, as it is closer to Apache, the communication is better, and mod_php is able to give the file content to Nextcloud. But I am not an expert at all, feel free to correct me :). 

> * this applies to fpm only

And FastCGI, no ?

> * this could have implications for resource usage on proxies (but since this is narrowly targeted and these transactions failed before I guess it doesn't matter too much?)

It does, but as you said, this only targets specific requests, which are currently failing, so better than nothing I guess.

> We should probably document it somewhere, particularly if there's a chance for resource usage trade-offs and/or if mod_php might be preferable in some environments over using this solution

Is it possible to add a condition to exclude mod_php ?

> Probably some work required for parity within the Nginx example config we provide in the Admin Manual since we explicitly disable the fastcgi_request_buffering behavior there

Indeed, the issue also occurs with Nginx.

> Might be some other touch spots in the docs - e.g. we reference the Sabre URL about 0-byte files currently in the Admin Manual (https://docs.nextcloud.com/server/latest/admin_manual/issues/general_troubleshooting.html#troubleshooting-webdav)

Better to keep it, no?

come-nc (approved):
:elephant: First time I hear about those things

icewind1991 (approved):

nursoda:
The PR doesn't fix the issue for me on 28.0.3RC1. Do I also need to tweak my NGINX Nextcloud config to make it work?

PR #43057, opened 2024-01-23
feat(perf): add cache for authtoken lookup

### LINKED ISSUE #42823: Cache authtoken lookup by token
### How to use GitHub

``SELECT * FROM `oc_authtoken` WHERE (`token` = :dcValue1) AND (`version` = :dcValue2)`` is one of the most common queries in Nextcloud executed most requests.

Cache the lookup

1. Positive cache for rows that exist. Keep data in memory for a few seconds to eliminate database lookups for highly concurrent requests of one user.
   1. Keep TTL low enough to kill any connections when the token is deleted within reasonable time (30s?)
   2. For time-critical data, like the password, ``\OC\Authentication\Token\PublicKeyTokenProvider::getPassword`` could still fetch the latest version from the database, to avoid reading an outdated password. This might be debatable and we could accept some failing authentication on external storage etc. for the short period of time where the password changes.
4. Negative cache for rows that do not exist. The tokens are random so it's very unlikely that a token that doesn't exist right now will exist in the next few minutes. Yet if a client retries their token again and again we hit the database for each request.

N/a

**Additional context**

N/a

Altahrim:
* Resolves: #42823

## Summary

Add cache for `authtoken` lookup

## Checklist

### DIFF
--- a/lib/private/Authentication/Token/PublicKeyTokenMapper.php
+++ b/lib/private/Authentication/Token/PublicKeyTokenMapper.php
@@ -29,6 +29,7 @@ namespace OC\Authentication\Token;
 
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\QBMapper;
+use OCP\Authentication\Token\IToken;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
 
@@ -42,8 +43,6 @@ class PublicKeyTokenMapper extends QBMapper {
 
 	/**
 	 * Invalidate (delete) a given token
-	 *
-	 * @param string $token
 	 */
 	public function invalidate(string $token) {
 		/* @var $qb IQueryBuilder */
@@ -150,14 +149,15 @@ class PublicKeyTokenMapper extends QBMapper {
 		return $entities;
 	}
 
-	public function deleteById(string $uid, int $id) {
+	public function getTokenByUserAndId(string $uid, int $id): ?string {
 		/* @var $qb IQueryBuilder */
 		$qb = $this->db->getQueryBuilder();
-		$qb->delete($this->tableName)
+		$qb->select('token')
+			->from($this->tableName)
 			->where($qb->expr()->eq('id', $qb->createNamedParameter($id)))
 			->andWhere($qb->expr()->eq('uid', $qb->createNamedParameter($uid)))
 			->andWhere($qb->expr()->eq('version', $qb->createNamedParameter(PublicKeyToken::VERSION, IQueryBuilder::PARAM_INT)));
-		$qb->execute();
+		return $qb->executeQuery()->fetchOne() ?: null;
 	}
 
 	/**
--- a/lib/private/Authentication/Token/PublicKeyTokenProvider.php
+++ b/lib/private/Authentication/Token/PublicKeyTokenProvider.php
@@ -38,7 +38,8 @@ use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\TTransactional;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Authentication\Token\IToken as OCPIToken;
-use OCP\Cache\CappedMemoryCache;
+use OCP\ICache;
+use OCP\ICacheFactory;
 use OCP\IConfig;
 use OCP\IDBConnection;
 use OCP\IUserManager;
@@ -48,6 +49,8 @@ use Psr\Log\LoggerInterface;
 
 class PublicKeyTokenProvider implements IProvider {
 	public const TOKEN_MIN_LENGTH = 22;
+	/** Token cache TTL in seconds */
+	private const TOKEN_CACHE_TTL = 10;
 
 	use TTransactional;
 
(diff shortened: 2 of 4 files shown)


lib/private/Authentication/Token/PublicKeyTokenMapper.php
```diff
@@ -55,27 +55,39 @@ public function invalidate(string $token) {
...
 			->where($qb->expr()->lt('last_activity', $qb->createNamedParameter($olderThan, IQueryBuilder::PARAM_INT)))
 			->andWhere($qb->expr()->eq('type', $qb->createNamedParameter(IToken::TEMPORARY_TOKEN, IQueryBuilder::PARAM_INT)))
 			->andWhere($qb->expr()->eq('remember', $qb->createNamedParameter($remember, IQueryBuilder::PARAM_INT)))
 			->andWhere($qb->expr()->eq('version', $qb->createNamedParameter(PublicKeyToken::VERSION, IQueryBuilder::PARAM_INT)))
-			->execute();
+			->executeQuery();
+
+		while ($tokenHash = $result->fetchOne()) {
+			yield $tokenHash;
+		}
```
ChristophWurst, line 73: same here

lib/private/Authentication/Token/PublicKeyTokenProvider.php
```diff
@@ -243,55 +239,63 @@ public function renewSessionToken(string $oldSessionId, string $sessionId): OCPI
...
-		$this->mapper->deleteById($uid, $id);
 	}
 
 	public function invalidateOldTokens() {
-		$this->cache->clear();
-
 		$olderThan = $this->time->getTime() - $this->config->getSystemValueInt('session_lifetime', 60 * 60 * 24);
 		$this->logger->debug('Invalidating session tokens older than ' . date('c', $olderThan), ['app' => 'cron']);
-		$this->mapper->invalidateOld($olderThan, OCPIToken::DO_NOT_REMEMBER);
+		foreach ($this->mapper->listOld($olderThan, OCPIToken::DO_NOT_REMEMBER) as $tokenHash) {
```
ChristophWurst, line 271: IMO this is not worth the effort. The cache is short lived. If we delete the tokens in the database only like before, they would only remain usable for 10s. On top of that, the cache is local, so we'd only invalidate the cache for some application servers of clustered setups.

lib/private/Authentication/Token/PublicKeyTokenProvider.php
```diff
@@ -48,47 +49,27 @@
...
-		$this->mapper = $mapper;
-		$this->crypto = $crypto;
-		$this->config = $config;
-		$this->db = $db;
-		$this->logger = $logger;
-		$this->time = $time;
-
-		$this->cache = new CappedMemoryCache();
-		$this->hasher = $hasher;
+	public function __construct(
```
ChristophWurst, line 60: nit: if we want to backport the change I would prefer to do the syntax upgrade in a follow-up and add the cache factory as explicit class property. That keeps the diff at a minimum.

tests/lib/Authentication/Token/PublicKeyTokenMapperTest.php
```diff
@@ -57,9 +57,9 @@ protected function setUp(): void {
 		$this->mapper = new PublicKeyTokenMapper($this->dbConnection);
 	}
 
-	private function resetDatabase() {
+	private function resetDatabase(): void {
 		$qb = $this->dbConnection->getQueryBuilder();
-		$qb->delete('authtoken')->execute();
+		$qb->delete('authtoken')->executeStatement();
```
ChristophWurst, line 62: nit: best done in a separate PR that only goes to master

tests/lib/Authentication/Token/PublicKeyTokenProviderTest.php
```diff
@@ -222,7 +227,7 @@ public function testUpdateTokenDebounce() {
 		$this->tokenProvider->updateTokenActivity($tk);
 	}
 
-	public function testGetTokenByUser() {
+	public function testGetTokenByUser(): void {
```
ChristophWurst, line 230: same with these

ChristophWurst (commented):
Tested & works

Blackfire confirms that 2 requests are shaved off most requests.
Loading my dev env dashboard causes 18 auth token SELECTs on master and 1 here.

:rocket:

lib/private/Authentication/Token/PublicKeyTokenProvider.php
Altahrim: With current setup, we will always have the issue of a token valid on a server for few seconds and invalid on another. All entries will not expire at same time.

But you are right, we can avoid this step and just delete from database :)

lib/private/Authentication/Token/PublicKeyTokenProvider.php
```diff
@@ -275,23 +301,21 @@ public function invalidateOldTokens() {
 	}
 
 	public function invalidateLastUsedBefore(string $uid, int $before): void {
-		$this->cache->clear();
-
-		$this->mapper->invalidateLastUsedBefore($uid, $before);
```
ChristophWurst, line 280: same as https://github.com/nextcloud/server/pull/43057#discussion_r1490659655? we can just delete the db entries and let caches expire automatically in a few seconds

ChristophWurst (approved):

skjnldsv (approved):

solracsf:
Any chance this could be backported?

PR #43082, opened 2024-01-24
fix wrong icon color in app store

### LINKED ISSUE #42998: icon color broken in app-store
See 
image

szaimen:
Fix https://github.com/nextcloud/server/issues/42998

<details>

```
docker run -it --rm ^
--name nextcloud-easy-test1 ^
-p 8444:443 ^
-e SERVER_BRANCH=enh/42998/fix-icon-color ^
-e COMPILE_SERVER=1 ^
--volume="nextcloud_easy_test_npm_cache_volume:/var/www/.npm" ^

```

</details>

### DIFF
--- a/apps/settings/src/views/Apps.vue
+++ b/apps/settings/src/views/Apps.vue
@@ -396,6 +396,28 @@ export default {
 	right: -8px;
 }
 
+// Fix wrong color for active icons
+@media (prefers-color-scheme: dark) { body {
+	.app-navigation-entry.active .app-navigation-entry-icon {
+		filter: var(--primary-invert-if-bright);
+	}
+}}
+@media (prefers-color-scheme: light) { body {
+	.app-navigation-entry.active .app-navigation-entry-icon {
+		filter: var(--primary-invert-if-dark);
+	}
+}}
+[data-themes*=dark] {
+	.app-navigation-entry.active .app-navigation-entry-icon {
+		filter: var(--primary-invert-if-bright);
+	}
+}
+[data-themes*=light] {
+	.app-navigation-entry.active .app-navigation-entry-icon {
+		filter: var(--primary-invert-if-dark);
+	}
+}
+
 .app-sidebar-tabs__release {
 	h2 {
 		border-bottom: 1px solid var(--color-border);
--- a/apps/settings/src/views/Apps.vue
+++ b/apps/settings/src/views/Apps.vue
@@ -398,22 +398,22 @@ export default {
 
 // Fix wrong color for active icons
 @media (prefers-color-scheme: dark) { body {
-	.app-navigation-entry.active .app-navigation-entry-icon {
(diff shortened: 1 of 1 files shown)


apps/settings/src/views/Apps.vue
```diff
@@ -396,6 +396,18 @@ export default {
...
+[data-themes*=dark] {
+	:deep(.app-navigation-entry.active .app-navigation-entry-icon) {
+		filter: var(--primary-invert-if-bright);
+	}
+}
+[data-themes*=light] {
+	:deep(.app-navigation-entry.active .app-navigation-entry-icon) {
+		filter: var(--primary-invert-if-dark);
+	}
+}
```
szaimen, line 409: I tried to fix this but unfortunately I gave up because apparently we need to use `filter: var(--primary-invert-if-bright);` in dark mode and `filter: var(--primary-invert-if-dark);` in light mode. Unfortunately I did not find a working solution for this.

susnux: I think this is missing the default theme case where the system colors are used. Meaning if data-themes is empty you need to use the `prefer-color-schema` media query

skjnldsv, line 409: Here you go @szaimen 

```suggestion
// Fix wrong color for active icons
@media (prefers-color-scheme: dark) { body {
	.app-navigation-entry.active .app-navigation-entry-icon {
		filter: var(--primary-invert-if-bright);
	}
}}

@media (prefers-color-scheme: light) { body {
	.app-navigation-entry.active .app-navigation-entry-icon {
		filter: var(--primary-invert-if-dark);
	}
}}

[data-themes*=dark] {
	.app-navigation-entry.active .app-navigation-entry-icon {
		filter: var(--primary-invert-if-bright);
	}
}
[data-themes*=light] {
	.app-navigation-entry.active .app-navigation-entry-icon {
		filter: var(--primary-invert-if-dark);
	}
}

```

szaimen: Thanks @skjnldsv ! :)

szaimen: I just tested this and it doesnt seem to work. Maybe gecause the scss is scoped? :/

skjnldsv: Yes, you need to keep the `:deep(`, I tested it live, locally, so it worked for me.

szaimen: I see. Where exactly would I need to put the :deep( ?

skjnldsv: Like it was before

szaimen: you mean like that?
image

szaimen: This doesnt seem to make it work unfortunately 😢

PR #43086, opened 2024-01-24
disable the internal s3sdk multi part copy logic

icewind1991:
We have our own logic so just set the threshold in the library to 2^64-1

### DIFF
--- a/lib/private/Files/ObjectStore/S3ObjectTrait.php
+++ b/lib/private/Files/ObjectStore/S3ObjectTrait.php
@@ -212,7 +212,8 @@ trait S3ObjectTrait {
 			$copy->copy();
 		} else {
 			$this->getConnection()->copy($this->getBucket(), $from, $this->getBucket(), $to, 'private', array_merge([
-				'params' => $this->getSSECParameters() + $this->getSSECParameters(true)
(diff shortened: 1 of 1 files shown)


lib/private/Files/ObjectStore/S3ObjectTrait.php
```diff
@@ -212,7 +212,8 @@ public function copyObject($from, $to, array $options = []) {
 			$copy->copy();
 		} else {
 			$this->getConnection()->copy($this->getBucket(), $from, $this->getBucket(), $to, 'private', array_merge([
-				'params' => $this->getSSECParameters() + $this->getSSECParameters(true)
+				'params' => $this->getSSECParameters() + $this->getSSECParameters(true),
+				'mup_threshold' => PHP_INT_MAX,
```
Fenn-CS, line 216: The threshold is set to max? What's the default?

nfebe (approved):
Looks okay. 

_However, if default is max, then it might be unnecessary?_

lib/private/Files/ObjectStore/S3ObjectTrait.php
icewind1991: 5GB I believe

Altahrim (approved):

PR #43098, opened 2024-01-24
fix(settings): federation scope menu a11y

### LINKED ISSUE #41903: [BITV] 9.4.1.2/6.3 - "Change scope" menus - Alle menus have been implemented using the aria-haspopup="menu" attribute, which informs users that a menu is being opened.(1)
> [!IMPORTANT]
> For QA, please just check that PR is merged to stable28

 However, even though the menu items can be navigated using arrow keys, the corresponding ARIA roles role="menu" and role="menuitem" are completely missing. 

https://report.bitvtest.de/default-en/cbedb5ef-4815-4cba-a83d-995599c178c3.html#checkpoint-bfbf07e5c3-v6-n3

ShGKme:
* Resolves: #41903 

## Summary

Use new `radio` behavior feature from `NcActionButton`.
- Remove intermediate `FederationControlAction`
  - `NcActions` doesn't fully support intermediate components
  - Fixes the `menu` role
  - Fixes keyboard navigation - Tab switches to the next UI element
- Use new `type="radio"`
  - Adjusts to `menuitemradio` role
- Remove custom styles to use new defaults
- Slightly fix padding on a small additional button
  Before | After
  ---|---
  image |  image

## Screenshots

Before | After
---|---
image | image

We also have now the default design for radio menus. See discussion in https://github.com/nextcloud-libraries/nextcloud-vue/pull/4744#issuecomment-1790996427

So I also removed the custom design. cc @nextcloud/designers 

Old custom design | New default design
---|---
image | image

## Checklist


apps/settings/src/components/PersonalInfo/shared/FederationControl.vue
```diff
@@ -21,29 +21,31 @@
...
-			:tooltip-disabled="federationScope.tooltipDisabled"
-			:tooltip="federationScope.tooltip" />
+			:close-after-click="true"
+			:disabled="!supportedScopes.includes(federationScope.name)"
+			:icon="federationScope.iconClass"
+			:name="federationScope.displayName"
+			type="radio"
+			:value="federationScope.name"
+			:model-value="scope"
+			@update:modelValue="changeScope">
```
ShGKme, line 38: Cannot use `@update:model-value` here because NcActionButton triggers event in camelCase instead of kebab-case.

Pytal (approved):

apps/settings/src/components/PersonalInfo/shared/FederationControl.vue
susnux: bug?

susnux (approved):

apps/settings/src/components/PersonalInfo/shared/FederationControl.vue
ShGKme: In Vue 3 this would work, but in Vue 2 an event name is case-sensitive. So to be able to use kebab case in template (respecting recommendation and our styles), the event had to be emitted in kebab-case.

What's why in Vue 2 it was recommended to use kebab-case for events.

While for Vue 3 - camelCase 🙈

ShGKme: I'd not call it a bug. `.sync` modifier supports both.

Anyway, we cannot change it without a breaking change. Or emit both names, but I'm not sure it is good idea. (but why not)

ShGKme: image

https://v2.vuejs.org/v2/guide/components-custom-events

susnux: Alright,  thank you for the explanation :)

JuliaKirschenheuter (approved):
🥳

jancborchardt (approved):
Very nice, looks great to me design-wise. :)

PR #43099, opened 2024-01-25
fix(tests): Make PostgreSQL unit tests less flacky by ordering

nickvergessen:
* Fixes e.g. https://drone.nextcloud.com/nextcloud/server/48020/17/5

## Summary

Postgres uses a journal file, so if you select a list from DB without ordering it, the order does not have to be FIFO.
Since the test is not about ordering but selecting distinct, it's okay to change the order to configkey (there was only 1 appid selected anyway so sorting by that doesn't do anything).

## Checklist


nickvergessen:
Why does this happen only now? I can only assume that before due to the 3 columns Postgres did not use a journal, but since recent additions now being 5 columns it started using it.

ChristophWurst (approved):

provokateurin (approved):

PR #42903, opened 2024-01-18
improve index on appconfig

ArtificialOwl:
follow-up of #42820

- [x] only create new fields if not existing yet
- [x] removing `appconfig_configkey_key` index on single field configkey as there is no search on this field
- [x] redefine `lazy` as `Types::INTEGER` (smallint(6)) instead of `Types::BOOLEAN` to be able to set it a `NotNull` (set in 28-rc2)
- [x] index on `lazy`
- [x] removing the use of `$migrationCompleted` that was needed during the migration from 28 to 29 but not anymore since we generate the (previously) missing fields `lazy` and `type` in 28
- [x] do not select `lazy` during loadConfig as we already know its value

_Note: After some discussion, it seems better to switch `lazy` to INTEGER so the field can be set as NotNull (because Oracle)_

Fixes https://github.com/nextcloud/server/issues/43103

PR #43114, opened 2024-01-25
feat(appconfig): Automatically store "sensitive" appconfigs encrypted in the database

nickvergessen:
- [x] Based on #42903 

## Summary

Quite some app configs contain sensitive information that we could prevent to leak, when a backup of the DB is leaked, to reduce the risk of follow up leaks/hacks.
The state of the art is to use `ICrypto::encrypt()` and `ICrypto::decrypt()` for that. While looking into it for some of the talk configs which are arrays, I thought it's a shame that we would loose all the typing as we have to json_encode the config making it strings, in order to be able to encrypt them.

So why not just encrypt all sensitive configs all the time automatically?

## TODO

- [x] Adjust normal getter and setter
- [x] Check `getAllValues` and `getDetails`

## Checklist

### DIFF
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -46,6 +46,7 @@ use OCP\Exceptions\AppConfigUnknownKeyException;
 use OCP\IAppConfig;
 use OCP\IConfig;
 use OCP\IDBConnection;
+use OCP\Security\ICrypto;
 use Psr\Log\LoggerInterface;
 
 /**
@@ -70,6 +71,8 @@ use Psr\Log\LoggerInterface;
 class AppConfig implements IAppConfig {
 	private const APP_MAX_LENGTH = 32;
 	private const KEY_MAX_LENGTH = 64;
+	private const ENCRYPTION_PREFIX = '$AppConfigEncryption$';
+	private const ENCRYPTION_PREFIX_LENGTH = 21; // strlen(self::ENCRYPTION_PREFIX)
 
 	/** @var array<string, array<string, mixed>> ['app_id' => ['config_key' => 'config_value']] */
 	private array $fastCache = [];   // cache for normal config keys
@@ -92,7 +95,8 @@ class AppConfig implements IAppConfig {
 
 	public function __construct(
 		protected IDBConnection $connection,
-		private LoggerInterface $logger,
+		protected LoggerInterface $logger,
+		protected ICrypto $crypto,
 	) {
 	}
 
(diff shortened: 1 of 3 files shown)


lib/private/AppConfig.php
```diff
@@ -462,11 +464,24 @@ private function getTypedValue(
 			throw new AppConfigTypeConflictException('conflict with value type from database');
 		}
 
-		if ($lazy) {
-			return $this->lazyCache[$app][$key] ?? $default;
+		if ($lazy && isset($this->lazyCache[$app][$key])) {
+			$value = $this->lazyCache[$app][$key];
```
ArtificialOwl, line 468: what if you put a ref as `$value = &$this->lazyCache[$app][$key];` ? would it update the value in the global array with decrypted one ?

ArtificialOwl (approved):

lib/private/AppConfig.php
```diff
@@ -462,11 +464,24 @@ private function getTypedValue(
...
+			return $default;
 		}
 
-		return $this->fastCache[$app][$key] ?? $default;
+		$sensitive = $this->isTyped(self::VALUE_SENSITIVE, $knownType);
+		if ($sensitive) {
+			// Only decrypt values that are stored encrypted
+			$sections = substr_count($value, '|');
+			if ($sections === 2 || $sections === 3) {
+				$value = $this->crypto->decrypt($value);
```
ArtificialOwl, line 480: what would happen if unencrypted sensitive string fullfil this condition ?
should we add a prefix to the stored value to help identify encrypted value ?
do we manage/ignore exception on decrypt ?

lib/private/AppConfig.php
nickvergessen: I dislike such side-effects, but anyway I iirc references are not working for scalar types in php

lib/private/AppConfig.php
```diff
@@ -728,12 +743,17 @@ private function setTypedValue(
...
-		if ($this->hasKey($app, $key, $lazy)
+		if (!$sensitive
+			&& $this->hasKey($app, $key, $lazy)
 			&& $value === $this->getTypedValue($app, $key, $value, $lazy, $type)
-			&& (!$sensitive || $this->isSensitive($app, $key, $lazy))) {
+			&& !$this->isSensitive($app, $key, $lazy)) {
 			return false;
 		}
 
+		if ($sensitive) {
```
ArtificialOwl, line 753: firefox crashed and seems suggestion was not attached to my previous review ...

`$sensitive` can be false, but once set as sensitive in database it won't be reverted unless using `updateSensitive()`

```suggestion
		if ($sensitive || $this->isSensitive($app, $key, $lazy)) {
```

lib/private/AppConfig.php
```diff
@@ -728,12 +743,17 @@ private function setTypedValue(
 		 * no update if key is already known with set lazy status, or value is
 		 * different, or sensitivity switched from false to true.
 		 */
-		if ($this->hasKey($app, $key, $lazy)
+		if (!$sensitive
+			&& $this->hasKey($app, $key, $lazy)
 			&& $value === $this->getTypedValue($app, $key, $value, $lazy, $type)
-			&& (!$sensitive || $this->isSensitive($app, $key, $lazy))) {
+			&& !$this->isSensitive($app, $key, $lazy)) {
```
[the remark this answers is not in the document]
nickvergessen: > meaning, while new and old value are the same, we keep update the database if $sensitive is true but value in database is not yet set as sensitive !

That's exactly what I tried to do

lib/private/AppConfig.php
nickvergessen: Shouldn't `$sensitive = $this->isTyped(self::VALUE_SENSITIVE, $type);` above then have this?
Also we would then need to manipulate the type and make sure we store it, even when it was not provided.

All those methods are quite complex and all invoking each others and it's really not straight forward atm for me to see what is right.

nickvergessen: Got it now, and added it in a second commit

ArtificialOwl (approved):

AndyScherzinger (approved):

PR #43116, opened 2024-01-25
fix(files): correct params on Go back link

ShGKme:
## Summary

Before, going back results an invalid URL. Updating the page results in infinity loop.

(Video with sound)

https://github.com/nextcloud/server/assets/25978914/c66a8ddc-fdf7-4f5c-b905-08d91e64fcd3

This issue fixes the link. But probably we should also fix such an invalid link handling.

https://github.com/nextcloud/server/assets/25978914/2d55bf2f-6b3a-44fb-ba29-08b24101e449

Also works on root.

https://github.com/nextcloud/server/assets/25978914/f6877ac5-4f82-46a7-ae1a-c5c890b791a0

## TODO

- [x] Remove `aria-label` - it is not correct to have `aria-label` different from a button text and not needed when there is a text
- [x] Set `fileid` on `"Go back"`

## Checklist


JuliaKirschenheuter (approved):
works, great!

apps/files/src/views/FilesList.vue
```diff
@@ -96,7 +96,6 @@
 			data-cy-files-content-empty>
 			<template #action>
 				<NcButton v-if="dir !== '/'"
-					:aria-label="t('files', 'Go to the previous folder')"
 					type="primary"
 					:to="toPreviousDir">
 					{{ t('files', 'Go back') }}
```
skjnldsv, line 101:
```suggestion
					{{ t('files', 'Go to the parent folder') }}
```

apps/files/src/views/FilesList.vue
```diff
@@ -341,7 +340,9 @@ export default defineComponent({
 		 */
 		toPreviousDir(): Route {
 			const dir = this.dir.split('/').slice(0, -1).join('/') || '/'
-			return { ...this.$route, query: { dir } }
+			const fileid = this.pathsStore.getPath(this.currentView?.id, dir)?.toString()
+			// @ts-expect-error Route expect params to be string, but we need to explicitly set undefined when there is no fileid
+			return { ...this.$route, params: { fileid }, query: { dir } }
```
skjnldsv, line 345: You should not need to provide the fileid.
Maybe making sure we just keep the view?

```suggestion
			return { name: this.$route.name, params: { view: this.$route.params?.view || 'files'  }, query: { dir } }
```

ShGKme: If I understand our files routing correctly, we always have `fileid` in the `path`, except root.

For example, if I open `/Foo/Bar` and `Bar` directory has `fileid=700`, then the current route is:
- In Files: `/apps/files/files/700?dir=/Foo/Bar`
- In favorites: `/apps/files/favorites/700?dir=/Foo/Bar`

Let's say, I'm in the root, then I'm going to Baz and back to Bar.
1. `/apps/files/files?dir=/`
2. `/apps/files/files/699?dir=/Foo`
3. `/apps/files/files/700?dir=/Foo/Bar`
4. `/apps/files/files/701?dir=/Foo/Bar/Baz`

From 4 shouldn't I return to 3 `/apps/files/files/700?dir=/Foo/Bar`?
Why `/apps/files/files/?dir=/Foo/Bar`?

If going back should be without `fileid`, does it mean that while going through directories forward and though breadcrumb there should also be no `fileid`?

1. `/apps/files/files?dir=/`
2. `/apps/files/files?dir=/Foo`
3. `/apps/files/files?dir=/Foo/Bar`
4. `/apps/files/files?dir=/Foo/Bar/Baz`

skjnldsv: > we always have `fileid` in the `path`

no, having fileid in the path is optional, we don't rely on it to open folders or browse.
We only rely on it to scroll to a specific file in the current opened folder or trigger a file action on load

ShGKme: Does it mean that links to folders in the files list should not contain `fileid` at all?
Because there is no such a file in the folder, nothing to scroll to.

For example, if `/Foo` folder has `fileid=700`, then path `/apps/files/files/700?dir=/Foo` doesn't make sense — there is no "700" in the `/Foo`, only in parent `/`.

skjnldsv: You can have it or you can't.
It will have no effect on the loading of the folder.

the fileid, like in older versions, is only used within the Files app to work with a file within a folder (like open the sidebar, etc etc)

ShGKme: > used within the Files app to work with a file within a folder (like open the sidebar, etc etc)

And that is why it makes no sense to me to have `fileid` when opening a folder. Because there is no such a file there.

If I'm in `/apps/files/files/700?dir=/Foo` and 700 is `/Foo` itself, what is supposed to happen?

skjnldsv: `/apps/files/files/226467?dir=/Shared`:
1. Will open the `/Shared` folder 
2. Will trigger the sidebar action for the file id `226467` (that is a children of `/Shared`)

`/apps/files/files/114403?dir=/Shared`:
1. Will open the `/Shared` folder 
2. Will trigger the sidebar action for the file id `114403` which is the `/Shared` folder itself

ShGKme: Hmm, now I'm not sure what it an expected behavior _everywhere_.

Should we have a **folder's** `fileid` by default?
1. Just in the file list when we go to the next folder
2. In the breadcrumbs
3. In the "Go back" link

skjnldsv: > * Just in the file list when we go to the next folder
> * In the breadcrumbs
> * In the "Go back" link

* Yes
* Yes
* aaaaand yes/no, it's good to have it but it's optional

ShGKme: > 1. Yes
> 2. Yes

> will trigger the sidebar action for the file id 

Does it mean that opening any folder is supposed to also open a sidebar after it?

If it is not, we have a inconsistent here:
- Being on a page `/:fileid` doesn't open a sidebar
- Opening a link with `/:fileid` opens a sidebar
- => A fact of reloading a page opens a sidebar o_O

ShGKme: Probably, there should be a query param for sidebar, like `?details`, similar to `openfile`?

skjnldsv: No please :see_no_evil: 
Let's not complicate URLs.

> Being on a page /:fileid doesn't open a sidebar
> Opening a link with /:fileid opens a sidebar
> => A fact of reloading a page opens a sidebar o_O

That's not a inconsistency, this is the feature. It's been long discussed and implemented by the design team.
On load, if you provide a fileid for a FILE, we open the sidebar (if screen size allows)

skjnldsv: Also, please, let's not dive into this, this is not what this issue is about at all.
You have no idea the amount of use cases the Files app supports and I had to match up.

ShGKme:
Closed in flavor of https://github.com/nextcloud/server/pull/43174

PR #43145, opened 2024-01-26
Improve freeprompt task type description

julien-nc:
Any other idea for this description?

This could be backported to 27 and 28.

Extra question: Should we directly rename this task type here? For now, the assistant renames "Free prompt" to "Generate text" but it's a hack.


marcelklehr (commented):
IMO, the free prompt type wasn't meant to be user-facing, but rather a way to allow apps to do custom queries

julien-nc:
@marcelklehr So you would be in favor of artificially changing the description in the assistant UI instead of in the task type?

marcelklehr:
Yes, it's a bit ugly, but seems more correct, imo

PR #43170, opened 2024-01-27
fix(CI): MySQL unit tests, use LTS for code coverage and drop EOL

susnux:
## Summary
1. Use the LTS version of MySQL for the code coverage instead of an EOL version
2. Test LTS (8.0) and the latest version (8.3) (lower + upper bound) and drop EOL version 8.1

## Checklist

### DIFF
--- a/.github/workflows/phpunit-mysql.yml
+++ b/.github/workflows/phpunit-mysql.yml
@@ -53,9 +53,9 @@ jobs:
     strategy:
       matrix:
         php-versions: ['8.0']
-        mysql-versions: ['8.0', '8.1']
+        mysql-versions: ['8.0', '8.3']
         include:
-          - mysql-versions: '8.1'
+          - mysql-versions: '8.0'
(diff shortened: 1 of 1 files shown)


.github/workflows/phpunit-mysql.yml
```diff
@@ -69,7 +69,7 @@ jobs:
         options: --health-cmd="redis-cli ping" --health-interval=10s --health-timeout=5s --health-retries=3
 
       mysql:
-        image: ghcr.io/nextcloud/continuous-integration-mysql-${{ matrix.mysql-versions }}:latest
```
nickvergessen, line 72: We should add it on our CI instead, so we have backups

nickvergessen (changes requested):

nickvergessen (dismissed):

skjnldsv (approved):

nickvergessen (approved):

PR #43189, opened 2024-01-29
Feat: Create filter-plugin architecture for unified search

### LINKED ISSUE #42914: 🔍🗨️ Unified Search in individual Talk room
### How to use GitHub

Currently you can limit searches via unified search to be limited to a talk room or folder

Extend the current unified search implementation to allow for limiting a search to a specific Talk room or a folder in Files.

none

**Current requirements that need to be discussed:**

- [ ] Implementing filter : value in Unified search
- [ ] Filter translation for non-english speakers, e.g. "conversation" -> "raum", etc.

nfebe:
**Description**
--
This commit introduces the mechanism for apps out of the core, to add search filters to the unified search "Places" filter selector.
Resolves : https://github.com/nextcloud/server/issues/42914

**TODO**

- [x] Ensure nothing regarding this feature is triggered if talk is not installed
- [x] Document how to add filters to unified search, somewhere in developer notes? (Transferred to : https://github.com/nextcloud/documentation/issues/11544)

**Manual Testing**

Set spreed/talk branch to : https://github.com/nextcloud/spreed/pull/11575


jancborchardt (commented):
Looks good, 2 details:
- [ ] As @marcoambrosini said: Rename the current "Conversations" filter to "Talk" for better clarity
- [ ] The "In conversation" should be moved further up so it is grouped with the other Talk-related filters

core/src/components/UnifiedSearch/ConversationsListModal.vue
nickvergessen, line 1: This is very problematic and does not belong into the server repository (as mentioned in the chat).
Approaches similarly to existing integrations with Talk and Deck, Maps, Notes, Tasks and all the other things should be taken, so that server just allows a "registration of additional filters/places" and when clicking on it, talk is taking over and just responding with the selected filter value.

nickvergessen (dismissed):

marcoambrosini:
> The reason this was not done straight way is because the providers

I think it's better to do it properly in the backend no?

marcoambrosini (commented):
Looks good! The only thing that I would add is a search box on top of the places filter too. These can be a lot (see our company instance) and it might also get confusing. 
It would be also great to set up search aliases, for example if someone searches  "Chat", they're also presented with the "talk" and "conversation" filters.

Fenn-CS:
The context of the PR has changed and actual UI changes that his PR initially contained are now in : https://github.com/nextcloud/spreed/pull/11575

core/src/store/index.js
```diff
@@ -0,0 +1,11 @@
+import Vue from 'vue';
+import Vuex from 'vuex';
```
Antreesy, line 2: Could you rewrite it to Pinia store, maybe? We'll migrate to it anyway sooner or later, there's no need to increase future efforts

core/src/unified-search.js
```diff
@@ -26,6 +26,7 @@ import { translate as t, translatePlural as n } from '@nextcloud/l10n'
 import Vue from 'vue'
 
 import UnifiedSearch from './views/UnifiedSearch.vue'
+import { useSearchStore } from '../src/store/unified-search-external-filters.js'
```
Antreesy, line 29: You need to register Pinia same way as Vuex:
```js
import { createPinia, PiniaVuePlugin } from 'pinia'
Vue.use(PiniaVuePlugin)
const pinia = createPinia()
export default new Vue({..., pinia, ...})
```

Fenn-CS: Right, sorry about that, looks like I got a cache-false positive when I tested if Pinia would work with the registration.

Antreesy (approved):
As I said, I've nothing against changes. Please align with Design team and Joas based on this API PR: https://github.com/nextcloud/spreed/pull/11456

Maybe you need to use 'talk-message-current' provider?

Fenn-CS:
> Maybe you need to use 'talk-message-current' provider?

I [asked](https://github.com/nextcloud/spreed/pull/11575#discussion_r1489370174) and got confirmation from @nickvergessen that the existing provider is correct.

So mentioning here for further alignment, sending extra parameters to that provider, does nothing.

cc: @Altahrim

nickvergessen:
`talk-message` is the search provider that accepts the filter.
There is also an integration test that the API if called like specified, does the filtering correctly.

Fenn-CS:
Looking at 

https://github.com/nextcloud/spreed/blob/978922ee73e74f15c2ffe3fd83800edb6d2b01af/tests/integration/features/bootstrap/FeatureContext.php#L2452-L2459 it does look like we expect the user to type the search query to include the room? according to regex `'/conversation:ROOM\((?P<name>\w+)\)`?

Or at least that we need the request like that? In the chats what @Altahrim explained was that we simple need to pass `?term=hello&conversation=<token>` for the filtering to work. 

Please, am I missing something? 

Otherwise, If the user is expected to type strings that match `'/conversation:ROOM\((?P<name>\w+)\)` I think it should be parsed at the front-end at still sent simply using the query structure above? 

Looking at same snippet it seems to suggest @Antreesy  is right that we should use `talk-message-current'` instead as `talk-message` has descriptive text "search in other rooms"


cc: @nickvergessen

nickvergessen:
> it does look like we expect the user to type the search query to include the room? according to regex `/conversation:ROOM\((?P<name>\w+)\)`?

This is just because the integration test has only 1 field. But the code you linked to is exactly the important one. It removes the conversation bit from the query and adds `&conversation=<token>`

So yeah:
> `?term=hello&conversation=<token>`

This is exactly what is expected. `/search/providers/talk-message/search?term=hello&conversation=<token>` to be specific

`talk-message-current` only searches in the conversation that is the `from=/call/<token2>`, so no. `talk-message` must be used for the filter plugin.

Fenn-CS:
The URL's being sent look like `/search/providers/talk-message/search?term=uni&from=/apps/spreed/&conversation=qbhbdutu` @nickvergessen kindly test and review at your earliest convenience.

Thank you.

nickvergessen:
Seems to be working after compiling locally (because you didn't commit that yet).

>  The "In conversation" should be moved further up so it is grouped with the other Talk-related filters

This however is currently still missing and impossible for Talk to do.
Maybe we can add a "priority" to the registration object like the normal search providers have?

Another thing I noticed is that `In conversation` is shown as headline instead of `Messages` when no filter is active:
grafik

But that we can also have a look at later. Maybe we can resolve that in Talk itself.

Fenn-CS:
> Seems to be working after compiling locally (because you didn't commit that yet).
> 
> > 
>     The "In conversation" should be moved further up so it is grouped with the other Talk-related filters
> 
> This however is currently still missing and impossible for Talk to do.
> Maybe we can add a "priority" to the registration object like the normal search providers have?


The order is currently not influenced in anyway at the frontend level as that's arrange in the backend. We have some options like :

- Adding priority as suggested : But the question of what is the priority arises? Any app can set a priority that would override the priorities sent from the backend? (As whatever priority we set has to be relative to the other ones, hence to achieve the desire results we need to kinda hard code a priority and do frontend reordering, or copy the  algorithm to calculate priority from the backend to frontend yet making sure the inputs to calculate the priority is/are available)

- We could just arrange the filters in alphabetical order (?)

----

> Another thing I noticed is that In conversation is shown as headline instead of Messages when no filter is active:

- That has to be updated in spreed/talk. 

Brings the question of pluralization, we think figure out how to handle pluralization because ones the filter is not applied it is still considered a provider and in a case such as the one above it should read Messages but not sure whether when a particular conversation is applied is should read the same thing.

Fenn-CS:
@nickvergessen PS, had time to read my comments?

nickvergessen:
> But the question of what is the priority?

Integer, just like with the normal providers on the PHP api

> We could just arrange the filters in alphabetical order 

Sounds like the worst option and still doesn't allow to group Talk options together

Fenn-CS:
> Integer, just like with the normal providers on the PHP api

It is sure an int, but the chosen integer if being set from the front-end needs to carefully picked to make sure it ranks just at the right position in this case grouping with talk. How do we write that algorithm in a way that prevents any app that implements the filter plugin API from distorting the order by simply setting higher values?

What seems easily achievable at this point, is creating a front-end ordering mechanism that simply inserts the new filters immediately after the last filter from a specific provider, without regard to any integer. The result is that, filters would be grouped according to providers, but all filters attached via the plugin API would appear in the order in which they were attached.


@nickvergessen would you consider this a middle-ground compromise?

nickvergessen:
Ah you don't have the priority in the frontend, the backend sorts it and does not expose it.

I guess that sounds like the best option then

core/src/store/unified-search-external-filters.js
```diff
@@ -0,0 +1,36 @@
+/*
+ * @copyright Copyright (c) 2021 Fon E. Noel NFEBE <me@nfebe.com>
```
ChristophWurst, line 2: happy new year

core/src/views/UnifiedSearchModal.vue
```diff
@@ -535,6 +561,42 @@ export default {
...
+					// So, find the provider attached, add apply the extra parameters to those providers only
+					const compatibleProviderIndex = this.providers.findIndex(provider => provider.id === addFilterEvent.id)
+					if (compatibleProviderIndex > -1) {
+						provider.extraParams = addFilterEvent.filterParams
+						this.filteredProviders[i] = provider
+					}
+					break
+				}
+			}
+			console.debug('Search scope set to conversation', addFilterEvent)
```
ChristophWurst, line 579: leftover?

core/src/unified-search.js
```diff
@@ -47,8 +49,21 @@ Vue.mixin({
 	},
 })
 
+// Register the add/register filter action API globally
+window.OCP = window.OCP || {}
+window.OCP.UnifiedSearch = {
```
ChristophWurst, line 54: not sure about the approach. are we returning to globals for public API?

if so, please move this into ``core/src/OCP/index.js``. otherwise `OCP.UnifiedSearch` only exists if `core/src/unified-search.js` is bundled.

ChristophWurst: > not sure about the approach. are we returning to globals for public API?

For comparison see the other exports in core/src/OCP/index.js. We are trying to get away from globals. They make core and app scripts fragile to loading order issues.

Fenn-CS: @ChristophWurst Makes sense, given that, this already has a matching PR in spreed merged and there is existing work going for files would you be up for:

1. Creating a follow up issue to move this to `core/src/OCP/index.js` (I see it makes more sense there, we can even change that to ts and use properly types with the module system used there)....

2. Merge this and do 1, as follow up PR

To meet release schedule for these features?

Fenn-CS: Other minor remarks handled.

ChristophWurst: In both cases we will introduce this global function as public API. That means it has to stay for at least three major releases.

If this has to go in today then we can only add the proper API for 30, deprecate this approach, keep the parallel systems for two more releases and then drop the deprecated function again.

ChristophWurst: @skjnldsv @susnux any other ideas for a pragmatic way forward?

Fenn-CS: > In both cases we will introduce this global function as public API. That means it has to stay for at least three major releases.
> 
> If this has to go in today then we can only add the proper API for 30, deprecate this approach, keep the parallel systems for two more releases and then drop the deprecated function again.

Make sense, we can also deprecate it immediately after and still keep for two releases to allow a good update window. That should not be the problem because we have two apps that currently implement the API and we are the maintainers.

skjnldsv: > we have two apps that currently implement the API

talk and?

core/src/store/unified-search-external-filters.js
```diff
@@ -0,0 +1,36 @@
...
+
+export const useSearchStore = defineStore({
+	id: 'search',
+
+	state: () => ({
+		externalFilters: [],
+	}),
+
+	actions: {
+		registerExternalFilter({ id, appId, label, callback, icon }) {
```
skjnldsv, line 32: Can you write it in typescript and add proper typings so we know what to expect? :)

core/src/unified-search.js
skjnldsv: Agreed with @ChristophWurst, it's a no go from me.
Such API deserves its own handling in a lib package and proper typings so it can be documented and easily used.

core/src/store/unified-search-external-filters.js
Fenn-CS: I already did in a follow up PR that introduces the files filter using same API : https://github.com/nextcloud/server/pull/43963/commits/a996c21d4b9d51961a34b3f71fe35ca134df4509

core/src/views/UnifiedSearchModal.vue
```diff
@@ -258,8 +263,13 @@ export default {
 
 	},
 	mounted() {
+		subscribe('nextcloud:unified-search:add-filter', this.handlePluginFilter)
```
skjnldsv, line 266: That's a no from me, using an event to actually handle this registrations are not a stable way to do so.

I'd suggest using a proper registration like the event bus or the file actions is doing.

core/src/views/UnifiedSearchModal.vue
```diff
@@ -412,12 +423,27 @@ export default {
 		},
 		addProviderFilter(providerFilter, loadMoreResultsForProvider = false) {
 			if (!providerFilter.id) return
+			if (providerFilter.isPluginFilter) {
+				providerFilter.callback()
```
skjnldsv, line 427: What is a callback? What does it do?
A callback that is called without any parameters is odd to me.

skjnldsv (dismissed):

emoral435 (commented):
Overall, good PR!

There are a few places where it gets confusing between the `name` and `label` attributes, and I agree with the API choices that the other reviewers have mentioned. Perhaps for tomorrows team call / Thursday PR chat's, you can schedule something to be aligned with the team, because I understand the time conflicts are not easy ahaha :)

core/src/unified-search.js
```diff
@@ -47,8 +49,21 @@ Vue.mixin({
...
+	registerFilterAction: ({ id, appId, name, label, callback, icon }) => {
+		const searchStore = useSearchStore()
+		searchStore.registerExternalFilter({ id, appId, name, label, callback, icon })
```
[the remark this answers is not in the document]
Fenn-CS: @emoral435 That's because in unified search `name` is used for the display labels, meanwhile in the code base register action functions use labels just as labels. 

However, there is a follow up commit that improves the situation : https://github.com/nextcloud/server/pull/43963/commits/a996c21d4b9d51961a34b3f71fe35ca134df4509

core/src/views/UnifiedSearchModal.vue
```diff
@@ -412,12 +423,27 @@ export default {
...
+			}
 			this.providerResultLimit = loadMoreResultsForProvider ? this.providerResultLimit : 5
 			this.providerActionMenuIsOpen = false
-			const existingFilter = this.filteredProviders.find(existing => existing.id === providerFilter.id)
-			if (!existingFilter) {
-				this.filteredProviders.push({ id: providerFilter.id, name: providerFilter.name, icon: providerFilter.icon, type: 'provider', filters: providerFilter.filters })
+			// With the possibility for other apps to add new filters
+			// Resulting in a possible id/provider collision
+			// If a user tries to apply a filter that seems to already exist, we remove the current one and add the new one.
+			const existingFilterIndex = this.filteredProviders.findIndex(existing => existing.id === providerFilter.id)
```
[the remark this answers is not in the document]
Fenn-CS: I am not sure what you are asking but yes, filter id's can clash and that's why this overwrite check exist.

core/src/unified-search.js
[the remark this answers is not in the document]
emoral435: Ah, I see now! A bit rough, but I see now that the follow up commit will help this out. That should reduce some of the confusion surrounding this :)

core/src/views/UnifiedSearchModal.vue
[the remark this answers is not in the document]
emoral435: Ah, my bad, I did not explain this well. What I am saying is that this check only finds *one* conflicting filter id's, no? What if there where more conflicting id's? 

> However, if you would like, this certainly can be made as another issue, due to time constraints (release coming up, and this is not a major edge case)

I still stick by this though :)

Antreesy:
Clicking on `Load more results` changes the filtering method (or drops some query params at least). Is it expected?

[Screencast from 05.03.2024 17:29:31.webm](https://github.com/nextcloud/server/assets/93392545/53722ab4-8479-4481-ac07-4a61f7b2bc71)

emoral435 (approved):
Outside of cases where to make the code clearer and more maintainable for the future, good PR :) Discussed with others, and the API deprecation can be handled later, just make sure to address https://github.com/nextcloud/server/pull/43189#issuecomment-1979172940 and its all good to me!

Fenn-CS:
> Outside of cases where to make the code clearer and more maintainable for the future, good PR :) Discussed with others, and the API deprecation can be handled later, just make sure to address https://github.com/nextcloud/server/pull/43189#issuecomment-1979172940 and its all good to me!

Sure thanks, would address that in follow up https://github.com/nextcloud/server/pull/43963 which is on its way in, I think it's best this goes in now so at the very least the spreed filter is working by feature freeze.

PR #43191, opened 2024-01-29
fix: Use DI for Setup class and move away from deprecated methods

come-nc:
## Summary

Extracted from https://github.com/nextcloud/server/pull/40633 as it made more sense to clean up `OC\Setup` completely.
Migrate from deprecated methods to `OCP\Server::get` and DI.

## Checklist

### DIFF
--- a/lib/private/Installer.php
+++ b/lib/private/Installer.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types=1);
+
 /**
  * @copyright Copyright (c) 2016, ownCloud, Inc.
  * @copyright Copyright (c) 2016, Lukas Reschke <lukas@statuscode.ch>
--- a/lib/private/Installer.php
+++ b/lib/private/Installer.php
@@ -64,37 +64,17 @@ use Psr\Log\LoggerInterface;
  * This class provides the functionality needed to install, update and remove apps
  */
 class Installer {
-	/** @var AppFetcher */
-	private $appFetcher;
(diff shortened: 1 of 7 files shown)


lib/private/Installer.php
```diff
@@ -401,10 +383,10 @@ public function downloadApp($appId, $allowUnstable = false) {
 	 * @param bool $allowUnstable
 	 * @return string|false false or the version number of the update
 	 */
-	public function isUpdateAvailable($appId, $allowUnstable = false) {
+	public function isUpdateAvailable($appId, $allowUnstable = false): string|false {
 		if ($this->isInstanceReadyForUpdates === null) {
 			$installPath = OC_App::getInstallPath();
-			if ($installPath === false || $installPath === null) {
```
nickvergessen, line 407: but this can return false?!

lib/private/Installer.php
```diff
@@ -401,10 +383,10 @@ public function downloadApp($appId, $allowUnstable = false) {
...
+			if ($installPath === null) {
 				$this->isInstanceReadyForUpdates = false;
 			} else {
 				$this->isInstanceReadyForUpdates = true;
```
Altahrim, line 392: Nit:
```suggestion
			$this->isInstanceReadyForUpdates = $installPath !== null;
```

Altahrim (approved):

ChristophWurst (approved):

PR #43203, opened 2024-01-30
Use Dependency Injection in register_commands.php

come-nc:
## Summary

Use DI for all commands from Core to stop relying on deprecated methods.

## Checklist

### DIFF
--- a/core/Command/Security/ListCertificates.php
+++ b/core/Command/Security/ListCertificates.php
@@ -26,16 +26,20 @@ use OC\Core\Command\Base;
 use OCP\ICertificate;
 use OCP\ICertificateManager;
 use OCP\IL10N;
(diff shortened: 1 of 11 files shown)


come-nc:
There is a service named `IntegrityCodeChecker` which is not a class name and must be used for integrity commands to not mess up with installation.
It’s at https://github.com/nextcloud/server/blob/e22443263f83ac212e69d2e0b3a054a7ca8916b9/lib/private/Server.php#L979
Maybe it can be renamed to the class name so that DI can be used?

ChristophWurst (commented):
:broom::sunglasses:

ChristophWurst (approved):

nickvergessen (approved):

PR #43213, opened 2024-01-30
fix(sharing): migrate QuickShareSelect to NcActions

### LINKED ISSUE #43139: [BITV]: Sharing tab: permissions dropdown is not accessible
Screenshot from 2024-01-26 08-39-29

It is just a `<div>` with buttons inside, not a list ;(. Actually it should be a full `NcActions`, but don't know how to solve it without design changes.

ShGKme:
* ⚠️ Requires `@nextcloud/vue@8.6.0`
* Resolves: https://github.com/nextcloud/server/issues/43139

## Summary

The current custom quick share select implementation has a number of a11y issues:
- The toggle button
  - is not focusable
  - has no button role
  - has no `aria-expanded="false"`
- The menu has popup role, better to have menu or select semantics
- Selected button is marked by `aria-selected` only on `true`, which is not allowed here and should has `false` value
- Keyboard navigation works, but the best is if <kbd>Tab</kbd> moves to the next UI element instead of the next option in the list

Having all of that, it seems that moving to `NcActions` with `menu` + `menuitemradio` semantics and some custom styles for the trigger is the best and easiest way to fix all a11y issues.

## Screenshots

🏚️ Before | 🏡 After
---|---
image | image
image | image
before-sharing-focus | after-sharing-focus

## Checklist


susnux (approved):

JuliaKirschenheuter (approved):
😍 works great!

ShGKme:
Waiting for `@nextcloud/vue@8.6.0`

emoral435 (approved):
LGTM 👍 Tested with NcVue 8.6.1 and works perfectly :)

ShGKme:
Rebased onto master after upgrading `@nextcloud/vue`

PR #43152, opened 2024-01-26
fix(files): disable `Unshare` until proper fix

skjnldsv:
Until we fix it for 28.0.3, it's safer to use the `Delete` wording

fix https://github.com/nextcloud/server/issues/43130

PR #43217, opened 2024-01-30
fix(files): add mount root property and adjust delete wording

### LINKED ISSUE #40940: [F2V] `Disconnect storage` and `Leave this share`
https://github.com/nextcloud/server/blob/dd3cd29224cd93c6168cd42e93e68cdd6ac5e921/apps/files/js/fileactions.js#L787-L791

skjnldsv:
Follow-up of https://github.com/nextcloud/server/pull/43152
Fix https://github.com/nextcloud/server/issues/40940

image
image


apps/files_sharing/src/init.ts
```diff
@@ -29,3 +30,7 @@ import './actions/restoreShareAction'
 import './actions/sharingStatusAction'
 
 registerSharingViews()
+
+registerDavProperty('nc:share-attributes', { nc: 'http://nextcloud.org/ns' })
+registerDavProperty('oc:share-types', { oc: 'http://owncloud.org/ns' })
+registerDavProperty('ocs:share-permissions', { ocs: 'http://open-collaboration-services.org/ns' })
```
skjnldsv, line 36: We should remove them from the default fetch, `files_sharing` might be disabled

apps/files/src/init.ts
```diff
@@ -66,5 +66,6 @@ registerRecentView()
 registerPreviewServiceWorker()
 
 registerDavProperty('nc:hidden', { nc: 'http://nextcloud.org/ns' })
+registerDavProperty('nc:is-mount-root', { nc: 'http://nextcloud.org/ns' })
```
artonge, line 69: Shouldn't we duplicate the line in the share and files_external apps?
We should also ensure that `@nc/files` removes duplicate props

skjnldsv: To gain some perf in the case external AND sharing is disabled?
I'm not sure that is worth it tbh, considering the `files/src/deleteAction` is using it, it's probably safer to leave it there?

artonge: Yes, you're right. It was just to have the code in the correct place. But if `deleteAction` use it, then it is safer to have it here.

artonge (approved):

apps/files/src/init.ts
skjnldsv: But those are great thoughts though! I also thought about it!
But for code maintenance, I don't think that make sense. We would gain a tiny bit for a very specific edge case, but lose much more by loading two more actions that is mostly duplicate code. :see_no_evil:

apps/files/src/actions/deleteAction.ts
```diff
@@ -20,21 +20,99 @@
...
+		if (canDisconnectOnly(nodes)) {
+			return n('files', 'Disconnect storage', 'Disconnect storages', nodes.length)
+		}
+
+		/**
+		 * If we're in the trashbin, we can only delete permanently
+		 */
+		if (view.id === 'trashbin') {
+			return t('files', 'Delete permanently')
+		}
```
come-nc, line 89: Could be higher, no? Things cannot be shared or mounted to trashbin?

come-nc (approved):

apps/files/src/actions/deleteAction.ts
skjnldsv: Absolutely! Will do!

skjnldsv:
/backport 6cf4fe19a021cab9f96ac23e32e52af2ee32218b to stable28

PR #43226, opened 2024-01-31
movie preview bug fix, in some case stream reading functions may block indefinitely

jpsn123:
## Summary
For some reason, I need to run nextcloud as root, and in this case ffmpeg will block the process when it generates preview.

## refencese:
https://www.php.net/manual/en/function.proc-open.php#97012
https://www.php.net/manual/en/function.stream-get-contents.php#106905

### DIFF
--- a/lib/private/Preview/Movie.php
+++ b/lib/private/Preview/Movie.php
@@ -307,6 +307,8 @@ class Movie extends ProviderV2 {
 		$returnCode = -1;
 		$output = '';
 		if (is_resource($proc)) {
+			stream_set_blocking($pipes[1], false);
+			stream_set_blocking($pipes[2], false);
 			$stderr = trim(stream_get_contents($pipes[2]));
 			$stdout = trim(stream_get_contents($pipes[1]));
 			$returnCode = proc_close($proc);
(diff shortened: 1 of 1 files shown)


lib/private/Preview/Movie.php
```diff
@@ -144,6 +144,8 @@ private function generateThumbNail(int $maxX, int $maxY, string $absPath, int $s
 		$returnCode = -1;
 		$output = "";
 		if (is_resource($proc)) {
+			stream_set_blocking($pipes[1], 0);
+			stream_set_blocking($pipes[2], 0);
```
blizzz, line 148:
```suggestion
			stream_set_blocking($pipes[1], false);
			stream_set_blocking($pipes[2], false);
```

Second parameter should be `bool`.

blizzz (commented):
From the documentation:

> In non-blocking mode an [fgets()](https://www.php.net/manual/en/function.fgets.php) call will always return right away while in blocking mode it will wait for data to become available on the stream.

Could be more exhaustive. Is it possible that we only receive incomplete data with this setting?

blizzz:
@jpsn123 thanks for the adaptions! Would you mind rebasing one more time?

jpsn123:
@blizzz  thanks for remind, i amend it, merge to one commit

blizzz (approved):

artonge:
@jpsn123 sorry for the delay here. Can you sign off your commit?

```shell
git rebase HEAD~1 --signoff
git push --force-with-lease origin master
```

Then we should be good :)

susnux (approved):

PR #43231, opened 2024-01-31
feat(files): ask for confirm if deleting 5 items or more 

### LINKED ISSUE #32011: File Delete Confirmation Dialog
When we right-click on the files we have the last two options from the "Download" and "Delete" list. It is very common for our users to click "Delete" instead of "Download" and the files are deleted directly.

When we click on "Delete" a popup window will open asking us to confirm if we really want to delete.

The "Download" and "Delete" buttons are no longer below each other.

**Additional context**
It's not necessary.

skjnldsv:
Fix https://github.com/nextcloud/server/issues/32011
Needs https://github.com/nextcloud/server/pull/43217

image
@nextcloud/designers


apps/files/src/actions/deleteAction.ts
```diff
@@ -136,7 +139,34 @@ export const action = new FileAction({
...
+					},
+				)
+				return
+			}
+			resolve(true)
+		})
+
+		// If the user cancels the deletion, we don't want to do anything
+		if (confirm === false) {
+			showError(t('files', 'Cancelled deletion'))
```
artonge, line 166: No?
```suggestion
			showError(t('files', 'Deletion cancelled'))
```

artonge (approved):

apps/files/src/actions/deleteAction.ts
skjnldsv: Done! :)

susnux (approved):

szaimen (approved):
LGTM from design perspective :)

AndyXheli:
Can't wait for this!!!!! Would this also work if a user moves large amounts of data from one share into another ?

PR #43234, opened 2024-01-31
fix(provisioning_api): Translate exceptions shown in the frontend

### LINKED ISSUE #13820: OCSExption "Cannot remove yourself from the admin group" is not translatable
Reported at forums.

See https://help.nextcloud.com/t/die-deutsche-ubersetzung-von-nextcloud-hier-diskutieren-und-fehler-melden/32139/120

Source line:

https://github.com/nextcloud/server/blob/b8b650aee697e4ab52f1f28e7e2287ac9a27456c/apps/provisioning_api/lib/Controller/UsersController.php#L751

susnux:
* Resolves: https://github.com/nextcloud/server/issues/13820

## Summary
This exception messages are used by the front end to be displayed to the user, using hint exceptions would be quite breaking as hint exceptions are handled as HTTP error 500 for OCS controllers.
But I think it should be ok to simply translate the messages as for programmatically usage you should not use the message but the status code.

Also fixed some minor deprecation issue.

## Checklist

### DIFF
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -63,6 +63,7 @@ use OCP\HintException;
 use OCP\IConfig;
 use OCP\IGroup;
 use OCP\IGroupManager;
+use OCP\IL10N;
 use OCP\IPhoneNumberUtil;
 use OCP\IRequest;
 use OCP\IURLGenerator;
@@ -79,22 +80,8 @@ use Psr\Log\LoggerInterface;
  * @psalm-import-type Provisioning_APIUserDetails from ResponseDefinitions
  */
 class UsersController extends AUserData {
-	/** @var IURLGenerator */
-	protected $urlGenerator;
-	/** @var LoggerInterface */
-	private $logger;
-	/** @var IFactory */
-	protected $l10nFactory;
-	/** @var NewUserMailHelper */
-	private $newUserMailHelper;
-	/** @var ISecureRandom */
-	private $secureRandom;
-	/** @var RemoteWipe */
-	private $remoteWipe;
-	/** @var KnownUserService */
-	private $knownUserService;
-	/** @var IEventDispatcher */
-	private $eventDispatcher;
+
+	private IL10N $l10n;
 
 	public function __construct(
 		string $appName,
@@ -104,14 +91,14 @@ class UsersController extends AUserData {
 		IGroupManager $groupManager,
 		IUserSession $userSession,
 		IAccountManager $accountManager,
-		IURLGenerator $urlGenerator,
-		LoggerInterface $logger,
 		IFactory $l10nFactory,
-		NewUserMailHelper $newUserMailHelper,
-		ISecureRandom $secureRandom,
-		RemoteWipe $remoteWipe,
-		KnownUserService $knownUserService,
-		IEventDispatcher $eventDispatcher,
+		private IURLGenerator $urlGenerator,
+		private LoggerInterface $logger,
+		private NewUserMailHelper $newUserMailHelper,
+		private ISecureRandom $secureRandom,
+		private RemoteWipe $remoteWipe,
+		private KnownUserService $knownUserService,
+		private IEventDispatcher $eventDispatcher,
 		private IPhoneNumberUtil $phoneNumberUtil,
 	) {
 		parent::__construct(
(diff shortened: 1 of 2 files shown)


rakekniven (dismissed):

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -1432,7 +1412,7 @@ public function removeFromGroup(string $userId, string $groupid): DataResponse {
 
 			if (count($userSubAdminGroups) <= 1) {
 				// Subadmin must not be able to remove a user from all their subadmin groups.
-				throw new OCSException('Not viable to remove user from the last group you are SubAdmin of', 105);
+				throw new OCSException($this->l10n->t('Not viable to remove user from the last group you are SubAdmin of'), 105);
```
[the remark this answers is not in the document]
susnux: > Here you use "SubAdmin" or "subadmin". Please align all occurences. My vote goes to "sub-admin"

Done :white_check_mark:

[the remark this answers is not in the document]
susnux: >  btw. I do not get the meaning of this sentence

If I understand this correctly a sub-admin is not allowed to remove a user from all groups that user is sub-admin of, meaning the user must stay sub-admin of at least one groupe (maybe @nickvergessen, as the author, can help)?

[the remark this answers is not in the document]
nickvergessen: yeah I think so, to avoid users that are managed by no one

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -545,7 +525,7 @@ public function addUser(
 
 			// Send new user mail only if a mail is set
 			if ($email !== '') {
-				$newUser->setEMailAddress($email);
+				$newUser->setSystemEMailAddress($email);
```
nickvergessen, line 528: Please don't make functional changes in a PR named "Translate…" Also those are not the same and infact it might be impossible to manage the system email address. So this should stay as `setEMailAddress`

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -1058,7 +1038,7 @@ public function editUser(string $userId, string $key, string $value): DataRespon
 				break;
 			case IAccountManager::PROPERTY_EMAIL:
 				if (filter_var($value, FILTER_VALIDATE_EMAIL) || $value === '') {
-					$targetUser->setEMailAddress($value);
+					$targetUser->setSystemEMailAddress($value);
```
nickvergessen, line 1041:
```suggestion
					$targetUser->setEMailAddress($value);
```

nickvergessen (dismissed):

apps/provisioning_api/lib/Controller/UsersController.php
susnux: Well just wanted to get rid of the deprecation warning and the docs say it is just an alias so one would expect both to be equal:

```php
	/**
	 * set the email address of the user
	 *
	 * It is an alias to setSystemEMailAddress()
	 *
	 * @param string|null $mailAddress
	 * @return void
	 * @since 9.0.0
	 * @deprecated 23.0.0 use setSystemEMailAddress() or setPrimaryEMailAddress()
	 */
	public function setEMailAddress($mailAddress);
```

But I understand this is not the scope of this PR.

susnux: Reverted that change.
But if both functions are really not identical that doc-comment is really misleading.

nickvergessen: Hmm was not aware of the deprecation, maybe I have wrong memory in my head

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -461,15 +441,15 @@ public function addUser(
 				$group = $this->groupManager->get($groupid);
 				// Check if group exists
 				if ($group === null) {
-					throw new OCSException('Subadmin group does not exist', 102);
+					throw new OCSException('Sub-admin group does not exist', 102);
```
Pytal, line 444: Missing the translation?

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -461,15 +441,15 @@ public function addUser(
...
 				}
 				// Check if trying to make subadmin of admin group
 				if ($group->getGID() === 'admin') {
-					throw new OCSException('Cannot create subadmins for admin group', 103);
+					throw new OCSException('Cannot create sub-admins for admin group', 103);
```
Pytal, line 448: Missing translation?

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -461,15 +441,15 @@ public function addUser(
...
 				}
 				// Check if has permission to promote subadmins
 				if (!$subAdminManager->isSubAdminOfGroup($user, $group) && !$isAdmin) {
-					throw new OCSForbiddenException('No permissions to promote subadmins');
+					throw new OCSForbiddenException('No permissions to promote sub-admins');
```
Pytal, line 452: Missing translation?

susnux:
@Pytal also added translations for the missing exceptions - at least where it makes sense.

Pytal (approved):

nickvergessen (approved):

PR #43247, opened 2024-01-31
[IAppConfig] returns non lazy value while searching for lazy one

ArtificialOwl:
There is this possible situation where your part of code is using already existing config value. This config value needs to be switched to _lazy_ to improve performance, but the only way to store a value as `lazy` is:

- set a value using function like `setValueString()` with the argument `lazy: true` and the config value itself.
- using `updateLazy()` on the app/key pair without the need of the content of  the config value.

the `updateLazy()` could be run at one point of the migration, or later on, but there is room for a use case where an app is requesting a _lazy_ config value still stored as _non lazy_ in the database

### DIFF
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -462,11 +462,14 @@ class AppConfig implements IAppConfig {
 			throw new AppConfigTypeConflictException('conflict with value type from database');
 		}
 
-		if ($lazy) {
-			return $this->lazyCache[$app][$key] ?? $default;
-		}
-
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):
Sounds like this can be easily added as a unit test? :)

come-nc (approved):

ArtificialOwl:
> Sounds like this can be easily added as a unit test? :)

https://github.com/nextcloud/server/blob/a1b6357fc1cd1bc5a75b016e3b45f12003e8394b/tests/lib/AppConfigTest.php#L431-L434

PR #40726, opened 2023-09-30
feat: Email option in `occ user:add` command

kyteinsky:
* Resolves: #25319 

## Summary
Builds upon #29368 

This adds an occ command option (`email`) during the addition of a user to auto-generate a password and send a link to the email address to set a new password if:
1. `password-from-env` option was not passed
2. `email` is given and valid

## Docs
* Related: [docs#11160](https://github.com/nextcloud/documentation/pull/11160)

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [ ] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #43250, opened 2024-02-01
enh(users): send welcome email when using occ

### LINKED ISSUE #25319: No "Welcome" email sent when using `occ`.
Using the latest version of NextCloud, we discovered a problem that might be a bug.
We're using a CSV file to import our users into the system, via a small bash script that we created.

```
#!/bin/bash
USERS_CSV="$HOME/nextcloud-users.csv"
OCC="$HOME/public_html/occ"
OLDIFS=$IFS
IFS=","
while read username password displayName email group
 do
    export OC_PASS=$password
    echo "Adding user ${username} (${displayName} | ${password} | ${group})..."
    $OCC user:add --password-from-env --display-name "$displayName" --group "$group" "$username"
    echo "Setting ${username}'s email to ${email}..."
    $OCC user:setting "$username" settings email "$email"
done < $USERS_CSV
 IFS=$OLDIFS
```

This works great, and all users from the CSV file are added. However, none of them receive the "Welcome" email inviting them to set / reset their password.

helmo:
Enhance PR #40726

* Resolves: #25319 

## Summary

This PR tries to solve the remaining issues in #40726
Specifically from the comment https://github.com/nextcloud/server/pull/40726#pullrequestreview-1666910957

## Checklist

### DIFF
--- a/core/Command/User/Add.php
+++ b/core/Command/User/Add.php
@@ -2,6 +2,7 @@
 /**
  * @copyright Copyright (c) 2016, ownCloud, Inc.
  *
+ * @author Anupam Kumar <kyteinsky@gmail.com>
  * @author Arthur Schiwon <blizzz@arthur-schiwon.de>
  * @author Christoph Wurst <christoph@winzerhof-wurst.at>
  * @author Joas Schilling <coding@schilljs.com>
@@ -25,8 +26,6 @@
  */
 namespace OC\Core\Command\User;
 
-use Egulias\EmailValidator\EmailValidator;
-use Egulias\EmailValidator\Validation\RFCValidation;
 use OC\Files\Filesystem;
 use OCA\Settings\Mailer\NewUserMailHelper;
 use OCP\EventDispatcher\IEventDispatcher;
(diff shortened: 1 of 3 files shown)


come-nc:
Changed target to master to see the whole diff, my comment on original PR still applies:

> This needs to work and set email:
> `occ user:add --password-from-env --email bob@example.com bob`

-> This you fixed, I think

> This needs to work and set email, if `newUser.sendEmail` is off it should ask for password same as without `--email`:
> `occ user:add --email bob@example.com bob`

-> This you did not, if `newUser.sendEmail` is off, a temporary password is printed to the output instead of asking for the password in interactive mode. Passing `--email` option when `newUser.sendEmail` is off should not prevent entering a password.

I’m not sure what the correct behavior is here, maybe simply add an other option `--generate-password`?
Because people may want the welcome email to be sent while still setting the password.

come-nc (changes requested):
See above

kyteinsky:
> I’m not sure what the correct behavior is here, maybe simply add an other option --generate-password?

Yeah it would be better to set a generated password when explicitly asked rather than implicitly based on other options.

dropped password printing commit and pushed the requested changes in [`feat/new-user-email` branch](https://github.com/nextcloud/server/pull/40726). Maybe I will add more tests later.

kyteinsky:
merged here: #40726

PR #43090, opened 2024-01-24
fix(files): combine navigation and files list into single Vue app

ShGKme:
* Resolves: https://github.com/nextcloud/server/issues/43085
* See discussion https://github.com/nextcloud-libraries/nextcloud-vue/pull/4983#discussion_r1435029130
  image

## Summary

During `files2vue` migration, files were migrated as two Vue apps:
- Navigation = `<NcAppNavitation>`
- FilesList = `<NcAppContent>`

But Vue app in Nextcloud is supposed to be one NcContent app that contains navigation and content.

After migration has finished there is no need to separate Navigation and FilesList.

## TODO

- [x] Combine Navigation and FilesList into one Files App
- [x] Remove any custom content from PHP template for files
  - There were two hidden inputs, but they are not used anymore
- [x] Check that public share is still working

## Screenshots

No visual changes on both user files and public share

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #43254, opened 2024-02-01
fix(files): right click actions menu

skjnldsv:
After https://github.com/nextcloud/server/pull/43090

The root changed https://github.com/nextcloud/server/blob/1b0eb2bebd6998b2ebfb8ed177640ea5329bd4ca/apps/files/src/components/FileEntry.vue#L402-L407

### DIFF
--- a/apps/files/src/components/FileEntry/FileEntryActions.vue
+++ b/apps/files/src/components/FileEntry/FileEntryActions.vue
@@ -346,7 +346,7 @@ export default Vue.extend({
 <style lang="scss">
 // Allow right click to define the position of the menu
 // only if defined
-.app-content[style*="mouse-pos-x"] .v-popper__popper {
+.content[style*="mouse-pos-x"] .v-popper__popper {
 	transform: translate3d(var(--mouse-pos-x), var(--mouse-pos-y), 0px) !important;
 
 	// If the menu is too close to the bottom, we move it up


ShGKme (approved):

apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -346,7 +346,7 @@ export default Vue.extend({
 <style lang="scss">
 // Allow right click to define the position of the menu
 // only if defined
-.app-content[style*="mouse-pos-x"] .v-popper__popper {
+.content[style*="mouse-pos-x"] .v-popper__popper {
```
ShGKme, line 349: Isn't it the same? Do we need root class here at all?

```suggestion
[style*="mouse-pos-x"] .v-popper__popper {
```

emoral435 (commented):
Overall good - however, it seems that firefox has had yet another regression (😓😓):

left: chrome | right: firefox
firefox_r3qZP3x53L

I remember this being mentioned in a PR before, but I cannot remember for the life of me which PR it was! If you believe this can be ignored for now / chrome is where most people use NC, I can go ahead and submit an approval

apps/files/src/components/FileEntry/FileEntryActions.vue
ShGKme: > Isn't it far less efficient?

Probably it is

susnux:
https://github.com/nextcloud/server/pull/44139

PR #43185, opened 2024-01-29
feat(db): Make dirty query logging available in production

ChristophWurst:
* Resolves: none

## Summary

Makes https://github.com/nextcloud/server/pull/42345 usable in production environments.

:warning: depending on the number of dirty queries detected, this may log very often :warning: 

## How to test

1) Set loglevel to 3
2) Set loglevel_dirty_database_queries to 3
3) Log into Nextcloud with user_status enabled
4) Look into nextcloud.log

master: nothing related to dirty queries
here: https://github.com/nextcloud/server/issues/43109 logged with level 3

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #42922, opened 2024-01-18
fix(db): Remove very verbose dirty query logs

ChristophWurst:
## Summary

* Primary/secondary switching happens in almost every request. Logging this is too much. If one is interested in this mechanism they can turn on query logging. Query logs are prefixed with primary/replica.
* Logging dirty tables for each statement (write) is *a lot*. This doesn't have a real benefit. Removing.

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #42890, opened 2024-01-17
fix(db): Fix replica password example config

ChristophWurst:
* Resolves: # 

## Summary

Replica config is an associative array with password=>password.

Ref https://github.com/nextcloud/server/pull/41998

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #43261, opened 2024-02-01
[stable28] feat(database): Primary/replica split

ChristophWurst:
Manual backports of
* [x] https://github.com/nextcloud/server/pull/41998
* [x] https://github.com/nextcloud/server/pull/42345
* [x] https://github.com/nextcloud/server/pull/42504
* [x] https://github.com/nextcloud/server/pull/42890
* [x] https://github.com/nextcloud/server/pull/42929
* [x] https://github.com/nextcloud/server/pull/43185
* [x] https://github.com/nextcloud/server/pull/42933
* [x] https://github.com/nextcloud/server/pull/42922
* [x] https://github.com/nextcloud/server/pull/43297
* [x] https://github.com/nextcloud/server/pull/43542

Ref https://github.com/nextcloud/server/issues/33542


ChristophWurst:
Tested on a small production instance and this works. The caveat is that it catches *lots* of queries and logs them as dirty. Most of that happens during cron.php execution because the joblist continuously does a `SELECT * FROM oc_jobs … LIMIT 1` mixed with lots of `UPDATE oc_jobs …` so it *looks like* there could be dirty reads. Whether there could be any is subject of a research task.

blizzz:
@ChristophWurst keep or close?

ChristophWurst:
This can be closed

PR #43267, opened 2024-02-01
fix(files): drop to folder path and user feedback

skjnldsv:
Please review two commits separately

## Uploading: addaa9be6b1a0dc2be771ea2b56e6f2f25b99f90
- Better feedback when dragging external files and dropping them to a displayed folder
- Fix the wrong path (currently, it takes the full dropped Folder path, but the uploader destination is always set to the current Folder. Leading uploading failures to paths like `/Photos/Photos/Landscapes`)
- Fixes `t.dataTransfer.files.forEach is not a function`
image

## Cleanup: 23d934f55a14d9e2174f43afeb732902392098a9
On a previous version, we were using the default `CheckboxRadioSwitch` `model` behaviour, that only works with string. With `defineComponent` we now see that there are a lot of typing issues. We switch back to properly using a `number` and not a `string`

## FileEntry mixin: 1d467202f48508b2778acfe485ada4ffcc7bae9f
Much cleaner to not have duplicate code :see_no_evil:


apps/files/src/components/FileEntry/FileEntryCheckbox.vue
```diff
@@ -112,8 +112,9 @@ export default Vue.extend({
 
 				const lastSelection = this.selectionStore.lastSelection
 				const filesToSelect = this.nodes
-					.map(file => file.fileid?.toString?.())
+					.map(file => file.fileid)
 					.slice(start, end + 1)
+					.filter(Boolean) as number[]
```
ShGKme, line 117: `fileid` cannot be 0, right?

apps/files/src/components/FileEntryMixin.ts
```diff
@@ -0,0 +1,409 @@
...
+			type: Array as PropType<Node[]>,
+			required: true,
+		},
+		filesListWidth: {
+			type: Number,
+			default: 0,
+		},
+	},
+
+	setup() {
```
ShGKme, line 65: This seems to work in this case, but in general case it is not valid to have `setup` in `mixin`.
Mixins by design doesn't support `setup`, so the `setup`'s result between several mixins or mixin + the last component won't be merged. 

For example, it is probably not possible to use any composable or another store in those components that uses this mixin.

ShGKme:
I couldn't reproduce the issue from the linked comment...

But fixes bulk upload to `/Photos/Album` for me

Before | After
---|---
image | image

apps/files/src/components/FileEntry/FileEntryCheckbox.vue
skjnldsv: No, unless someone do some questionable stuff :grin: 
It's not possible to insert a 0 to an auto increment field:https://github.com/nextcloud/server/blob/1cfc8a2e0396d97345d3e46260de2b2ce8f17326/core/Migrations/Version13000Date20170718121200.php#L178-L184

apps/files/src/components/FileEntryMixin.ts
skjnldsv: AH! if it's the only thing, I can move them back, it's not too bad :)

skjnldsv:
> I couldn't reproduce the issue from the linked comment...

Part of it has been fixed already, some remains:

## Issue 1
`t.dataTransfer.files.forEach is not a function`

https://github.com/nextcloud/server/assets/14975046/61d0c223-ccb5-4c86-87b0-c12b64fe8c17

## Issue 2
Considering the following path: `/Top folder/Sub folder`
Dragging a file on `Sub folder` while being in the `/Top folder` directory leads to a wrong path being used

https://github.com/nextcloud/server/assets/14975046/67caf300-79d7-4825-9226-e58051acff14

ShGKme (approved):

susnux (approved):
:rocket:

PR #43274, opened 2024-02-01
[IAppConfig] add RepairStep to migrate lazy appconfig

ArtificialOwl:
A SetupCheck to initialize the migration of a list of config value to lazy loading


come-nc (changes requested):
This is not what setup checks are for, this should be a migration.
It’s not even checking anything.

ArtificialOwl:
> This is not what setup checks are for, this should be a migration. It’s not even checking anything.

ok, it is still checking that all lazy config from a list have been migrated but I switched it to a daily background job

ArtificialOwl:
The job is now created during Setup and Repair (upgrade) process.

come-nc (commented):
But why does it need to run daily?

ArtificialOwl:
it's a light process. can be weekly, but the idea is that the list is not final and more elements might be added in the future

come-nc:
Then it should be a repair step, no?
https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/repair.html#repair-steps

PR #43275, opened 2024-02-01
return false on AppConfigUnknownKeyException

ArtificialOwl:
there is no logic in the current behavior of `updateLazy()` and `updateSensitive` that throw an AppConfigUnknownKeyException.

if there is no key in the database, there is no migration and we returns `false`


nickvergessen (approved):
Add unit test 🙊

come-nc (approved):

ArtificialOwl:
> Add unit test 🙊

https://github.com/nextcloud/server/blob/a1b6357fc1cd1bc5a75b016e3b45f12003e8394b/tests/lib/AppConfigTest.php#L1216-L1230

PR #43281, opened 2024-02-02
enh: update desktop client unsupported version (403) error message

### LINKED ISSUE #6273: Display a message to upgrade client if `minimumSupportedDesktopVersion` is thrown
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you want to have the same feature implemented.

### Feature description

When minimumSupportedDesktopVersion is set, it throws a 403 for unsupported client.
Instead, a clear message should be displayed on the client stating that 

> "This version of client is unsupported by the server. Please, upgrade your client."

or similar. Tjis is much more user friendly than a simple `403 Forbidden`.

https://github.com/nextcloud/server/blob/9964bcbfd576e3c0627f9c6bc662b3168d12a80b/config/config.sample.php#L1991-L2004

https://github.com/nextcloud/server/blob/9964bcbfd576e3c0627f9c6bc662b3168d12a80b/apps/dav/lib/Connector/Sabre/BlockLegacyClientPlugin.php#L56-L75

camilasan:
The user should get a more friendly warning when their desktop client version is not supported anymore by the server. See #nextcloud/desktop/issues/6273

### Before:

before

###  After:

after

### Systray message
systray

* Resolves: https://github.com/nextcloud/desktop/issues/6273

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/Connector/Sabre/BlockLegacyClientPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/BlockLegacyClientPlugin.php
@@ -7,6 +7,7 @@
  */
 namespace OCA\DAV\Connector\Sabre;
 
+use OCA\Theming\ThemingDefaults;
 use OCP\IConfig;
 use OCP\IRequest;
 use Sabre\DAV\Server;
@@ -21,10 +22,11 @@ use Sabre\HTTP\RequestInterface;
  */
 class BlockLegacyClientPlugin extends ServerPlugin {
 	protected ?Server $server = null;
-	protected IConfig $config;
 
-	public function __construct(IConfig $config) {
-		$this->config = $config;
+	public function __construct(
+		private IConfig $config,
+		private ThemingDefaults $themingDefaults,
+	) {
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


solracsf (changes requested):
Tests must be fixed.

solracsf:
What about adding a link on the message to `customclient_desktop` link?
This could lead to a better UX after all (upgrade one-click away!) :)

https://github.com/nextcloud/server/blob/ee6a62a8c17ef6df29ab02a2882b254e9918f88c/config/config.sample.php#L1105-L1117

solracsf:
@camilasan I was more thinking about a clickable link (if it can be done).
Otherwise, the full link is too much noise IMO :(

nimishavijay:
Agreed with @solracsf that the link could be hidden. What do you think about adding the link to the text "version 3.0.0 or later"? Would that be possible :)

```
Unable to connect to Nextcloud.
This version of the client is unsupported. Upgrade to [version 3.0.0 or later](https://nextcloud.com/install/#install-clients).
```

solracsf (approved):
Fine by me. 👍 
_Maybe just update the 1st post with an updated screenshot of the final version._

jancborchardt (changes requested):
Seems good if it’s a clickable link like @solracsf suggested. :) 

Another detail: In the settings it says "Unable to connect to Nextcloud" and in the notification "Network error" – both of these are different and neither really seem correct for the issue?
Maybe rather it should say "Connection issue" @camilasan?

jancborchardt (approved):

susnux (approved):

apps/dav/lib/Connector/Sabre/BlockLegacyClientPlugin.php
```diff
@@ -51,7 +53,10 @@ public function beforeHandler(RequestInterface $request) {
 		preg_match(IRequest::USER_AGENT_CLIENT_DESKTOP, $userAgent, $versionMatches);
 		if (isset($versionMatches[1]) &&
 			version_compare($versionMatches[1], $minimumSupportedDesktopVersion) === -1) {
-			throw new \Sabre\DAV\Exception\Forbidden('Unsupported client version.');
+			$customClientDesktopLink = htmlspecialchars($this->themingDefaults->getSyncClientUrl());
+			$minimumSupportedDesktopVersion = htmlspecialchars($minimumSupportedDesktopVersion);
+
+			throw new \Sabre\DAV\Exception\Forbidden("This version of the client is unsupported. Upgrade to <a href=\"$customClientDesktopLink\">version $minimumSupportedDesktopVersion or later</a>.");
```
solracsf, line 59: Ideally, this should be a translated string 😿 
Maybe in a follow-up PR.

PR #43296, opened 2024-02-02
fix kerberos test successfull authentication detection

icewind1991:
user_saml become more strict with the login redirect url that was breaking the login detection


ChristophWurst (approved):

ChristophWurst:
backport?

nickvergessen (approved):

PR #43297, opened 2024-02-02
fix(db): Let dirty writes cool off

ChristophWurst:
## Summary

We can assume that after a few seconds a read will be clean again. This is helpful for false warnings in long running processes.

## TODO

- [x] Track timestamp of table access

## Checklist

### DIFF
--- a/lib/private/DB/Connection.php
+++ b/lib/private/DB/Connection.php
@@ -55,7 +55,9 @@ use OCP\Diagnostics\IEventLogger;
 use OCP\IRequestId;
 use OCP\PreConditionNotMetException;
 use OCP\Profiler\IProfiler;
+use Psr\Clock\ClockInterface;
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):

lib/private/DB/Connection.php
```diff
@@ -55,7 +55,10 @@
 use OCP\IRequestId;
 use OCP\PreConditionNotMetException;
 use OCP\Profiler\IProfiler;
+use Psr\Clock\ClockInterface;
 use Psr\Log\LoggerInterface;
+use function array_filter;
```
nickvergessen, line 60: CS complains it's unused

ChristophWurst: CS is right

nickvergessen (approved):

PR #43352, opened 2024-02-05
Handle reverse proxy setting a port in Forwarded-For

Mika56:
As per RFC 7239, a reverse proxy may add a port to the client's IP address:
> If the server receiving proxied requests requires some
> address-based functionality, this parameter MAY instead contain an IP
> address (and, potentially, a port number)
https://datatracker.ietf.org/doc/html/rfc7239.html#section-5.2:~:text=and%2C%20potentially%2C%20a%20port%20number

This MR removes that port from the IP address, allowing the `Forwarde-For` header to be properly handled.

### DIFF
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -612,6 +612,11 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 							$IP = substr($IP, 1, -1);
 						}
 
+						// remove client port when set
+						if(str_contains($IP, ':')) {
+							$IP = substr($IP, 0, strpos($IP, ':'));
+						}
+
 						if ($this->isTrustedProxy($trustedProxies, $IP)) {
 							continue;
 						}
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -607,13 +607,12 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 					foreach (array_reverse(explode(',', $this->server[$header])) as $IP) {
 						$IP = trim($IP);
 						$colons = substr_count($IP, ':');
-						if($colons > 1) {
+						if ($colons > 1) {
 							// Extract IP from string with brackets and optional port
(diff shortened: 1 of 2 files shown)


lib/private/AppFramework/Http/Request.php
```diff
@@ -612,6 +612,11 @@ public function getRemoteAddress(): string {
 							$IP = substr($IP, 1, -1);
 						}
 
+						// remove client port when set
+						if(str_contains($IP, ':')) {
```
nickvergessen, line 616: this completely breaks IPv6

Mika56: Nice catch. I've updated the PR with new code that should work better

kesselb:
As reference, how Symfony solved the same problem: https://github.com/symfony/symfony/blob/d9d6024a1a663caa7f13360a4a1d9de443e9c954/src/Symfony/Component/HttpFoundation/Request.php#L1986-L1997

I recommend adding some automated tests to make sure the parsing does not break in the future.

Mika56:
And the corresponding tests: https://github.com/symfony/symfony/blob/d9d6024a1a663caa7f13360a4a1d9de443e9c954/src/Symfony/Component/HttpFoundation/Tests/RequestTest.php#L1034
Should we use their test suite? I don't see support for `[IPv6]:port`
Would that be adding a file to `tests/lib/AppFramework/Http` ?

nickvergessen:
You can add your tests in https://github.com/nextcloud/server/blob/216b95f8b1c4d3e5bf48e3b6075b727a71d2e30d/tests/lib/AppFramework/Http/RequestTest.php#L552

nickvergessen:
Do you need more help? Or shall we take over to finish it?

Mika56:
I lack time, if you're willing to take it over, please go for it! :)

nickvergessen:
Due to changes in CI a rebase was required, but I was unable to push that. So I made a new PR at https://github.com/nextcloud/server/pull/43552

Mika56:
No problem, thanks for taking care of it!

PR #43355, opened 2024-02-05
feat: Migrate LinkReferenceProvider to OCP

kyteinsky:
## Summary
This would be useful to eleminate the need for using this OC class
when an app dev wants to implement a custom reference provider
for the web client but wants to fall back on opengraph for mobile
clients.

## Checklist


kyteinsky:
It could've been an abstract class in OCP but that would interfere with `ADiscoverableReferenceProvider.php` so that idea was dropped.  
Please squash before merge. The two commits are separated for ease of review only.

lib/private/Collaboration/Reference/LinkReferenceProvider.php
```diff
@@ -24,160 +25,8 @@
...
-	}
-
-	public function getCachePrefix(string $referenceId): string {
-		return $referenceId;
-	}
-
-	public function getCacheKey(string $referenceId): ?string {
-		return null;
-	}
+/** @deprecated 29.0.0 */
```
julien-nc, line 30:
```suggestion
/** @deprecated 29.0.0 Use OCP\Collaboration\Reference\LinkReferenceProvider instead */
```

julien-nc (changes requested):
Minor comment.

Then :+1: :blue_heart:

julien-nc (approved):

juliusknorr (approved):

juliushaertl:
> The autoloaders are not up to date
> Please run: bash build/autoloaderchecker.sh

kyteinsky:
there are a lot of changed hashes in the apps. Should that be included in the commit?  
```diff
apps/admin_audit/composer/composer/installed.php

-        'reference' => 'b1797842784b250fb01ed5e3bf130705eb94751b',
+        'reference' => '49969705de01aff43cf2e6b662e5c96bd48a3b27',
```

PR #43366, opened 2024-02-05
feature(files) : Give a hint that names starting with dots might be hidden

### LINKED ISSUE #21609: Folders starting with multiple dots not visible
### Describe the bug

When I create a folder starting with multiple dots (at least 3), it gets created but is not visible. 

### Steps to reproduce
1. Create a folder named `... folder with dots`
2. The folder gets created in the file system
3. The folder is not visible in the web ui

### Expected behaviour
The folder should be visible

### Actual behaviour
Folder is not visible

### Server configuration

**Operating system:** Ubuntu 18.04

**Web server:** nginx

**Database:** mariadb

**PHP version:** 7.3

**Nextcloud version:** 18.06

**Updated from an older Nextcloud/ownCloud or fresh install:** Updated

**Where did you install Nextcloud from:**

arublov:
* Resolves: #21609 

## Checklist

Before:
The file name is changed with dots at the beginning without confirmation.

After:
Now when changing a file name with a dot/s at the beginning, a confirmation will pop up first to warn the user that the file can be hidden if the name is changed.

Screenshot 2024-02-05 at 20 57 29

### DIFF
--- a/apps/files/src/components/FileEntry/FileEntryName.vue
+++ b/apps/files/src/components/FileEntry/FileEntryName.vue
@@ -113,6 +113,12 @@ export default Vue.extend({
 		}
 	},
 
+	data() {
+		return {
+			isConfirmationFileCanHidden: false,
+		}
+	},
+
 	computed: {
 		isRenaming() {
 			return this.renamingStore.renamingNode === this.source
@@ -261,7 +267,7 @@ export default Vue.extend({
 			})
 		},
 		stopRenaming() {
-			if (!this.isRenaming) {
+			if (!this.isRenaming || this.isConfirmationFileCanHidden) {
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


susnux:
This will annoy a lot of users because it is shown everytime you rename such a file, especially if you know what you are doing.

So cc @jancborchardt for UX

ChristophWurst:
As an alternative we could use something like a tooltip. It warns but doesn't get into your way. What do you think?

jancborchardt:
Yup, would agree with @ChristophWurst’s suggestion of a tooltip. Does this work @arublov @susnux?

ChristophWurst:
Apparently the dialogue for new files now looks like this:

image

This could be enhanced by a https://nextcloud-vue-components.netlify.app/#/Components/NcNoteCard with type warning right under/over the filename input.

arublov:
> Apparently the dialogue for new files now looks like this:
> 
> image
> 
> This could be enhanced by a https://nextcloud-vue-components.netlify.app/#/Components/NcNoteCard with type warning right under/over the filename input.

Yes, I already tried to implement with NcNoteCard, but if you want to rename a folder/file, it happens:

Screenshot 2024-02-19 at 4 59 13 PM

arublov:
@ChristophWurst / @jancborchardt , If I do like this for renaming folders/files? The NcNoteCard will appear to the right of the input.

Screenshot 2024-02-19 at 5 25 25 PM

marcoambrosini:
I think we just need a never show this again checkbox. This type of warning doesn't work well in a tooltip, and for most users it's actually good to break the flow be very clear about the fact that the file/folder is about to be hidden.

susnux:
> About renaming warning: I think the dialog is the way to go, we just need a never show this again checkbox. This type of warning doesn't work well in a tooltip, and for most users it's actually good to break the flow be very clear about the fact that the file/folder is about to be hidden.


You can take the warn about changed file extension as an example on how to do the "never again" checkbox

PR #43387, opened 2024-02-06
fix!: Migrate jobs away from deprecated interfaces

come-nc:
## Summary

Another nail in the coffin of `ILogger`.
See #32127 

## Checklist


lib/public/Log/ILogFactory.php
```diff
@@ -40,15 +39,6 @@ interface ILogFactory {
 	 */
 	public function get(string $type): IWriter;
 
-	/**
-	 * @param string $path
-	 * @return ILogger
-	 * @since 14.0.0
```
ChristophWurst, line 46: breaking. okay with me but unexpected for a chore commit

ChristophWurst (approved):

ChristophWurst:
* Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required

This does need an entry in the upgrade docs.

lib/public/Log/ILogFactory.php
come-nc: `@deprecated 22.0.0 Use \OCP\Log\ILogFactory::getCustomPsrLogger`

I thought it was okay since it’s been deprecated since 22.
I’m a bit new to this chore/fix/enh thing, I tend to use chore when it’s not something user-facing that I’d want to list in the changelog, is there a better rationale?

ChristophWurst: It's probably debatable if API changes should land in the changelog of server. Since it is a platform with a public API, I would say it should. The tools I've worked with tend to ignore *chore* for changelogs, even if the change is breaking (`chore!`). I'd say a `fix!` would be a good fit here.

https://www.conventionalcommits.org/en/v1.0.0/#summary

apps/dav/lib/BackgroundJob/RefreshWebcalJob.php
```diff
@@ -34,50 +35,26 @@
...
-		$this->timeFactory = $timeFactory;
 	}
 
 	/**
 	 * this function is called at most every hour
 	 *
 	 * @inheritdoc
 	 */
-	public function execute(IJobList $jobList, ILogger $logger = null) {
+	public function start(IJobList $jobList): void {
```
nickvergessen, line 57: 🙈

nickvergessen (approved):

nickvergessen:
above documentation PR looks like it's referencing the wrong thing.
This PR here should still be documented

nickvergessen:
> Can you elaborate on this?

Let me rephrase then:

Even though `OC\BackgroundJob\…` were private, it was the (defacto) standard for quite some time.
I think it's worth to put a one-liner into https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_29.html#id4 for it.

PR #43399, opened 2024-02-06
make storage directory test be a bit less strict about mtimes

icewind1991:
sometimes the clock ticks over during the test

See for example https://github.com/nextcloud/server/actions/runs/7802691175/job/21281023435

Should hopefully reduce false negatives in CI


artonge (approved):
Can't we fix the clock instead?

icewind1991:
ideally everything would use `ClockInterface` so we can control the time during tests. But we don't live in that universe

skjnldsv (approved):

PR #43319, opened 2024-02-03
Fix color of app menu entries and background color

susnux:
* Resolves: https://github.com/nextcloud/server/issues/43298

## Summary
This fixes two issues:
1. a96a7f0c82c1ca22e78c42120e104ed46b9b08aa introduced a regression because it applied the color invert filter on the text which is already colored correctly (`primary-text`).
2. Most of the time the background color is `color-primary` but for some shipped background images the primary color does not lead to the same text color -> app menu names and the right header menu entries have wrong colors.

Especially try following: Set default background -> white text -> click "remove image" -> suddenly the app menu color is black on the same blue.

A better solution will be #42977 but for the moment this one is simpler.

before | after
---|---
Screenshot_20240203_140942|Screenshot_20240203_141051
Screenshot_20240203_141005|Screenshot_20240203_141035

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #43401, opened 2024-02-06
Revert "Fix color of app menu entries and background color"

skjnldsv:
Reverts nextcloud/server#43319

Makes CI green again on master


susnux:
was merged without missing push of cypress checks, will push later.

PR #43407, opened 2024-02-06
perf(dashboard): Stop loading the viewer on the dashboard

susnux:
## Summary

Viewer is ~4MB of JS assets for loading, this is not needed on the dashboard and the sidebar is also not used on the dashboard - there is no sidebar at all.

before | after
---|---
Screenshot_20240206_220934|Screenshot_20240206_221022
Screenshot_20240206_221121|Screenshot_20240206_221059

## Checklist

### DIFF
--- a/apps/dashboard/lib/Controller/DashboardController.php
+++ b/apps/dashboard/lib/Controller/DashboardController.php
@@ -30,8 +30,6 @@ declare(strict_types=1);
  */
 namespace OCA\Dashboard\Controller;
 
-use OCA\Files\Event\LoadSidebar;
-use OCA\Viewer\Event\LoadViewer;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http;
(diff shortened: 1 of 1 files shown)


ShGKme (approved):
BTW, should it be considered a breaking change?

For example, if some community app uses dashboard api v1 and Viewer

nickvergessen:
> BTW, should it be considered a breaking change?
> For example, if some community app uses dashboard api v1 and Viewer

I would say mentioning it in the developer docs doesn't hurt. The fix is also easy (emit the event yourself)

nickvergessen (approved):

skjnldsv (approved):

skjnldsv:
> I would say mentioning it in the developer docs doesn't hurt. The fix is also easy (emit the event yourself)

Which is also not really a fix, they should have done that since the beginning.
never rely on others triggering events for you :)

susnux:
Documentation: https://github.com/nextcloud/documentation/pull/11513

PR #43424, opened 2024-02-07
enh(appconfig): debug logs on lazy loadConfig

ArtificialOwl:
this will generate a debug entry, including a backtrace, in the logs when not-fast appconfig are loaded

### DIFF
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -1160,7 +1160,7 @@ class AppConfig implements IAppConfig {
 		}
 
 		if (($lazy ?? true) !== false) { // if lazy is null or true, we debug log
-			$this->logger->debug('The loading of lazy AppConfig values have been requested', ['trace' => debug_backtrace()]);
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

Altahrim (approved):

lib/private/AppConfig.php
```diff
@@ -1159,6 +1159,10 @@ private function loadConfig(?bool $lazy = false): void {
 			return;
 		}
 
+		if (($lazy ?? true) !== false) { // if lazy is null or true, we debug log
+			$this->logger->debug('The loading of lazy AppConfig values have been requested', ['trace' => debug_backtrace()]);
```
icewind1991, line 1163: logging an exception (but with the debug level) will ensure that the resulting trace can be understood/formatted by the logreader

ArtificialOwl: I tried, but it generate an entry too much looking alike an exception. I will keep it like this as this is purely a developer information.

If you have a valid example, please to not hesitate

nickvergessen (approved):

ArtificialOwl:
_(failing test not related)_

PR #43426, opened 2024-02-07
only get the path for the users cached mount info when we use it

icewind1991:
most of the time we shouldn't need it so we can save joining on the filecache.

Note that for the other "getMountsFor..." calls the path is used more often but those calls are less often done in the first place.

### DIFF
--- a/lib/private/Files/Config/LazyPathCachedMountInfo.php
+++ b/lib/private/Files/Config/LazyPathCachedMountInfo.php
@@ -1,0 +1,63 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022 Robin Appelman <robin@icewind.nl>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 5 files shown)


icewind1991:
some quick browsing a round a test instance with a breakpoint set hasn't resulted in any code path that gets the path so this looks promising

ArtificialOwl (approved):

come-nc (approved):

lib/private/Files/Config/UserMountCache.php
```diff
@@ -204,7 +210,12 @@ private function removeFromCache(ICachedMountInfo $mount) {
 		$query->execute();
 	}
 
-	private function dbRowToMountInfo(array $row) {
+	/**
+	 * @param array $row
+	 * @param (callable(CachedMountInfo): string)|null $pathCallback
+	 * @return CachedMountInfo|null
+	 */
+	private function dbRowToMountInfo(array $row, callable $pathCallback = null): ?ICachedMountInfo {
```
Altahrim, line 218:
```suggestion
	private function dbRowToMountInfo(array $row, ?callable $pathCallback = null): ?ICachedMountInfo {
```

lib/private/Files/Config/LazyPathCachedMountInfo.php
```diff
@@ -0,0 +1,63 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+namespace OC\Files\Config;
+
+use OCP\IUser;
+
+class LazyPathCachedMountInfo extends CachedMountInfo {
+	// we don;t allow \ in paths so it makes a great placeholder
```
Altahrim, line 28: Nit:
```suggestion
	// we don’t allow \ in paths so it makes a great placeholder
```

lib/private/Files/Config/UserMountCache.php
```diff
@@ -243,16 +265,30 @@ public function getMountsForUser(IUser $user) {
 
 			$this->mountsForUsers[$userUID] = [];
```
Altahrim, line 266: Useless now

Altahrim (approved):

PR #43429, opened 2024-02-07
feat: Log large assets that are initially loaded on the frontend

juliusknorr:
Check the offset, currently we have some reportings by default:

- core-common
- viewer-main
- richdocuments-viewer
- richdocuments-reference

### DIFF
--- a/lib/private/Template/ResourceLocator.php
+++ b/lib/private/Template/ResourceLocator.php
@@ -7,6 +7,7 @@
  */
 namespace OC\Template;
 
+use OC\SystemConfig;
 use Psr\Log\LoggerInterface;
 
 abstract class ResourceLocator {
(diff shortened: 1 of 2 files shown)


lib/private/Template/ResourceLocator.php
```diff
@@ -98,6 +101,10 @@ public function find($resources) {
 	 */
 	protected function appendIfExist($root, $file, $webRoot = null) {
 		if ($root !== false && is_file($root.'/'.$file)) {
+			$systemConfig = \OCP\Server::get(SystemConfig::class);
+			if ($systemConfig->getValue('debug', false) && filesize($root.'/'.$file) > self::LOG_LARGE_ASSET_OFFSET) {
```
szaimen, line 105: `filesize($root.'/'.$file)` Could this break on 32-bit?

juliushaertl: Not sure if there are alternatives there, but I'd say it might be ignorable since t his is debug log only and if you have a JS file >4GB there is likely other problems you run into before

PR #26593, opened 2021-04-16
Filter mounts for file id before trying to get user information

juliusknorr:
When fetching the mountpoints for a given fileid the results were previously returning all mountpoints of the given storage id and passing those to dbRowToMountInfo. This is generally ok, but with groupfolders all being located on the same storage id this will cause unnecessary calls for checking the user since we are only interested in the ones that match the path of the fileid.

This especially becomes an issue if there are lots of groupfolders setup and `activity_use_cached_mountpoints` is enabled where the list of mounts is obtained on each file operation for generating activites, see the below blackfire profiling:

image

image

PR #43437, opened 2024-02-07
use lazy user in UserMountCache for getting user for cached mount instead of duplicating logic

icewind1991:
remove some duplication and hopefully make it clearer what the logic is doing

- [x] soft requires https://github.com/nextcloud/server/pull/43426 as they change the same type signature

- reverts https://github.com/nextcloud/server/pull/26593 and uses `LazyUser` instead to achieve the same optimization

### DIFF
--- a/lib/private/Files/Config/UserMountCache.php
+++ b/lib/private/Files/Config/UserMountCache.php
@@ -251,6 +251,9 @@ class UserMountCache implements IUserMountCache {
 	 */
 	public function getMountsForUser(IUser $user) {
 		$userUID = $user->getUID();
+		if (!$this->userManager->userExists($userUID)) {
+			return [];
+		}
 		if (!isset($this->mountsForUsers[$userUID])) {
 			$builder = $this->connection->getQueryBuilder();
 			$query = $builder->select('storage_id', 'root_id', 'user_id', 'mount_point', 'mount_id', 'mount_provider_class')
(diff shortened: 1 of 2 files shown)


juliusknorr (approved):

artonge (approved):

lib/private/Files/Config/UserMountCache.php
```diff
@@ -213,13 +214,10 @@ private function removeFromCache(ICachedMountInfo $mount) {
...
-	 * @return CachedMountInfo|null
+	 * @return CachedMountInfo
 	 */
-	private function dbRowToMountInfo(array $row, ?callable $pathCallback = null): ?ICachedMountInfo {
-		$user = $this->userManager->get($row['user_id']);
-		if (is_null($user)) {
-			return null;
-		}
+	private function dbRowToMountInfo(array $row, ?callable $pathCallback = null): ICachedMountInfo {
+		$user = new LazyUser($row['user_id'], $this->userManager);
```
artonge, line 220: Maybe add a comment explaining the hack?
This is to prevent fetching the full user multiple times?

come-nc (approved):

lib/private/Files/Config/UserMountCache.php
icewind1991: fetching multiple times is not the issue (it's cached), it not having to fetch the user details when nobody is asking for it.

It's the same as with any other use of `LazyUser` so a comment would be redundant imo

PR #43438, opened 2024-02-07
Bump NcVue to `v8.6.2`

JuliaKirschenheuter:
Bump NcVue to `v8.6.2`


package-lock.json
```diff
@@ -4231,6 +4231,18 @@
         "npm": "^9.0.0"
       }
     },
+    "node_modules/@nextcloud/vue/node_modules/@nextcloud/router": {
```
susnux, line 4234: Now doubling the dependencies of router, might be bad. Can we update `@nextcoud/router` for server in the package.json as well? So we only use one version.

ShGKme: Then we need to update it literally everywhere. `@nextcloud/axios`, `@nextcloud/dialogs`, `@nextcloud/files`, `@nextcloud/password-confirmation`, `@nextcloud/upload`

Let's only update the `@nextcloud/vue` first. `@nextcloud/router` is stateless and small, should not be an issue.

ShGKme (approved):

Pytal (approved):

PR #43444, opened 2024-02-08
feat: add `block-unconventional-commits` workflow

skjnldsv:
@nextcloud/server


emoral435 (approved):
I think we can both agree, something about conventional commits are sexy 😍
LGTM 🐘

skjnldsv:
Only `sharing/download-share-disabled.cy.ts` is failing it seems, haven't took the time to fix them for now.

come-nc (commented):
This should be accompanied with a documentation update, maybe here:
https://github.com/nextcloud/server/blob/master/.github/CONTRIBUTING.md
Or/and here:
https://docs.nextcloud.com/server/latest/developer_manual/getting_started/codingguidelines.html

skjnldsv:
Done @come-nc

come-nc (approved):

PR #43446, opened 2024-02-08
feat: add `PATCH` HTTP Verb Support To `IClient` Interface & its HTTP Client Implementation

### LINKED ISSUE #43450: Add Patch Support in the Http Client Implementation
### Scenario 
I am writing a nextcloud app that is using `IClientService` which is using `IClient` interface to communicate with an external api. But I couldn't make a `PATCH` request to that api since the `IClient` Interface didn't implement that.

Check this PR on this regard. https://github.com/nextcloud/server/pull/43446

jithin-space:
## Summary

Added PATCH HTTP Verb Support To IClient Interface & its HTTP Client Implementation

Closes https://github.com/nextcloud/server/issues/43450

## TODO

- [ ] N/A

## Checklist

### DIFF
--- a/lib/public/Http/Client/IClient.php
+++ b/lib/public/Http/Client/IClient.php
@@ -173,6 +173,7 @@ interface IClient {
 	 *              'debug' => true,
 	 * @return IResponse
 	 * @throws \Exception If the request could not get completed
+	 * @since 29.0.0
 	 */
 	public function patch(string $uri, array $options = []): IResponse;
 
--- a/lib/public/Http/Client/IClient.php
+++ b/lib/public/Http/Client/IClient.php
@@ -148,6 +148,35 @@ interface IClient {
 	public function put(string $uri, array $options = []): IResponse;
 
 	/**
+	 * Sends a PATCH request
+	 * @param string $uri
+	 * @param array $options Array such as
+	 *              'body' => [
+	 *                  'field' => 'abc',
+	 *                  'other_field' => '123',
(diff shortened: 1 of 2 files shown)


lib/public/Http/Client/IClient.php
```diff
@@ -147,6 +147,35 @@ public function post(string $uri, array $options = []): IResponse;
...
+	 *                   'max'       => 10,  // allow at most 10 redirects.
+	 *                   'strict'    => true,     // use "strict" RFC compliant redirects.
+	 *                   'referer'   => true,     // add a Referer header
+	 *                   'protocols' => ['https'] // only allow https URLs
+	 *              ],
+	 *              'sink' => '/path/to/file', // save to a file or a stream
+	 *              'verify' => true, // bool or string to CA file
+	 *              'debug' => true,
+	 * @return IResponse
+	 * @throws \Exception If the request could not get completed
```
nickvergessen, line 175:
```suggestion
	 * @throws \Exception If the request could not get completed
	 * @since 29.0.0
```

ChristophWurst (approved):

nickvergessen (approved):

ChristophWurst:
@nickvergessen do we really want to merge after API freeze?

nickvergessen:
It's a pure addition, so I'd be fine.
But yeah following the rules it'd be 30, but that would mean the app can not be released in the meantime or they need to privately access Guzzle skipping all the security measurements we build in, so I'm a bit drawn apart.

Issue #43423, opened 2024-02-07
[Bug]: 'forbidden_chars' Not Enforced During File Uploads


nc-fkl:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

The parameter 'forbidden_chars,' introduced in Nextcloud 28 for setting prohibited characters, applies restrictions only when creating or renaming files and folders through the WebUI. This restriction does not extend to file uploads.

Could you please update the functionality to ensure that this parameter is also applied during file uploads?

The restriction does not apply to file uploads through the WebUI drag and drop or the "Upload files" using the +New button. 

### Steps to reproduce

1. + New
2. Upload files
3. Test#.txt

### Expected behavior

Thest "Test#.txt" should also be declined because of the restricted character like it is already done for folders

### Installation method

None

### Nextcloud Server version

28

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database engine version

MySQL

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

joshtrichards:
Fixed in nextcloud-libraries/nextcloud-upload#1132 (fka: nextcloud-libraries/nextcloud-upload#1086) and 079d026d39565e7b5547741bd62bd69082347ad6

PR #43468, opened 2024-02-08
[Bugfix]: 'forbidden_chars' added the condition to drop file

arublov:
* Resolves: #43423

## TODO

- [ ] ...

## Checklist


apps/files/src/services/DropService.ts
```diff
@@ -67,6 +70,14 @@ export const handleDrop = async (data: DataTransfer): Promise<Upload[]> => {
 const handleFileUpload = async (file: File, path: string = '') => {
 	const uploader = getUploader()
 
+	const toCheck = file.name.split('')
+	toCheck.forEach(char => {
+		if (forbiddenCharacters.indexOf(char) !== -1) {
+			showError(t('files', t('files', '"{char}" is not allowed inside a file name.', { char })))
+			throw '#.. is not allowed inside a file name.'
```
susnux, line 77: Throwing a string is nothing we should do, instead a proper `Error` should be used.

```suggestion
			throw new Error(`${char} is not allowed inside a file name.`).
```

apps/files/src/services/DropService.ts
```diff
@@ -67,6 +70,14 @@ export const handleDrop = async (data: DataTransfer): Promise<Upload[]> => {
 const handleFileUpload = async (file: File, path: string = '') => {
 	const uploader = getUploader()
 
+	const toCheck = file.name.split('')
+	toCheck.forEach(char => {
```
susnux, line 74: This seems to be the wrong way and probably performance wise pretty bad, instead I would do something like:

```js
const forbidden = forbiddenCharacters.split('')
if (forbidden.some((char) => file.name.includes(char))) {
// ...
}
```

why? because file name is probably longer than the list of forbidden chars so we reduce the outer loop.

susnux (changes requested):

apps/files/src/services/DropService.ts
arublov: @susnux thank you for your comments. Yesterday, I submitted that pull request as a draft because I'm still working on it and it's not the final version yet. Today, I'll have the final version for this task. Still, thank you for your comments.

PR #43471, opened 2024-02-09
Cache path by id

icewind1991:
Cache fileid to path mapping. 

The fact that a single fileid can have multiple paths complicates attampts to cache this information, however since most usages of `getById` only use a single item of the result, and always the same one (whether or not this is correct for that usage is another question :see_no_evil:). We can provide a new api that only returns a single result which makes caching easier.

We can get away with using a local cache without doing any active cache invalidation by checking that the node at the path we cached still has the correct fileid.

- add `getFirstNodeById` to get a single node for id regardless of how many nodes with that id the user has
- switch over all calls to `getById` that always use the first node to the new api
- [x] add caching to `getFirstNodeById`

I would recommend reviewing this PR commit-by-commit as most of the changes by LOC are mechanical changes to adjust to the new api.

### DIFF
--- a/lib/private/Files/Node/LazyFolder.php
+++ b/lib/private/Files/Node/LazyFolder.php
@@ -492,7 +492,11 @@ class LazyFolder implements Folder {
 	 * @inheritDoc
 	 */
 	public function getById($id) {
-		return $this->__call(__FUNCTION__, func_get_args());
+		return $this->getRootFolder()->getByIdInPath((int)$id, $this->getPath());
+	}
+
+	public function getFirstNodeById(int $id): ?\OCP\Files\Node {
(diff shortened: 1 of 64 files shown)


lib/private/Files/Node/LazyFolder.php
```diff
@@ -112,7 +112,7 @@ public function emit($scope, $method, $arguments = []) {
 		$this->__call(__FUNCTION__, func_get_args());
 	}
 
-	/**
+	/**'
```
come-nc, line 115:
```suggestion
	/**
```

come-nc (approved):
:+1:

Altahrim (approved):

PR #43527, opened 2024-02-12
fix(user_ldap): Do not block access to configuration page upon bad configuration

### LINKED ISSUE #43524: [Bug]:  Nextcloud inaccessible with erroneous LDAP configuration

### Bug description

With an erroneous LDAP configuration Nextcloud becomes inaccessible even to local admin 

### Steps to reproduce

1. Have an instance with an ldap configuration
2. Remove %uid from ldapLoginFilter   
3. Nextcloud becomes inaccessible 

### Expected behavior

Nextcloud is still accessible for local users even with an erroneous LDAP configuration

### Installation method

None

### Nextcloud Server version

28

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

come-nc:
- Resolve #43524

## Summary

It should be possible to access configuration page with the local admin
 user even when LDAP configuration is not valid.
This prevent ldap backend from throwing in countUsers to allow this.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Connection.php
+++ b/apps/user_ldap/lib/Connection.php
@@ -241,6 +241,7 @@ class Connection extends LDAPUtility {
 	}
 
 	/**
+	 * @throws ServerNotAvailableException When connection failed or configuration is invalid
 	 * @return resource|\LDAP\Connection The LDAP resource
 	 */
 	public function getConnectionResource() {
--- a/apps/user_ldap/lib/User_LDAP.php
+++ b/apps/user_ldap/lib/User_LDAP.php
@@ -300,7 +300,7 @@ class User_LDAP extends BackendUtility implements IUserBackend, UserInterface, I
 	 * @param string|\OCA\User_LDAP\User\User $user either the Nextcloud user
 	 * name or an instance of that user
 	 * @throws \Exception
-	 * @throws \OC\ServerNotAvailableException
+	 * @throws ServerNotAvailableException
 	 */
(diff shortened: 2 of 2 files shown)


susnux (approved):

blizzz (commented):
Needs a comment in code, for one would not be able to tell why the Exception is caught here and nowhere else.

Without having a better idea, it also feels weird to solve it this way, because
- the countUsers() method is not hard wired to the settings page
- a change there can have the side effect to make this solutions fail as side effect
- for completeness, side effects from this are also possible, but unlikely to be problematic

Not sure if it is possible to catch is somewhere higher, more to the configuration logic?

come-nc:
Well the method was not documented to throw anything, but *is* documented to return false if an error happens.

blizzz:
It is not a strong no, but it would cause me stomachaches. Unfortunately without looking into the matter I have no better idea at the moment either, and it is unlikely to change in the near future. That said a pragmatic approach is to go ahead with this solution, but opening an issue to have an improvement in mind at least. And add some comments into the code to clearly state the intention and meaning of the change, for by looking at it you would not have the slightest idea about it's actual purpose.

PR #35970, opened 2023-01-03
Improve exception serializing

juliusknorr:
## Summary

Small improvement to not start serializing an exception during logging if the log level would skip it anyways. Noticed when seeing errors due to `Nesting level too deep - recursive dependency? at /var/www/html/lib/private/Log/ExceptionSerializer.php#215` while there was no actual error logged when commenting out the argument cleanup

The second commit avoids such cases by handling the arguments first (which limits to a recursion depth of 5) before iterating over them for filtering for sensitive parameters.

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #43529, opened 2024-02-12
enh(log): exit log() earlier if no crashreport registered

ArtificialOwl:
follow-up on the previous improvement from @juliushaertl : https://github.com/nextcloud/server/pull/35970

The idea is to avoid some formatting of the payload as soon as we know for sure that it will not be used

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -62,24 +62,22 @@ use function strtr;
  * MonoLog is an example implementing this interface.
  */
 class Log implements ILogger, IDataLogger {
-	private IWriter $logger;
 	private ?SystemConfig $config;
 	private ?bool $logConditionSatisfied = null;
 	private ?Normalizer $normalizer;
-	private ?IRegistry $crashReporters;
 	private ?IEventDispatcher $eventDispatcher;
 
 	/**
 	 * @param IWriter $logger The logger that should be used
-	 * @param SystemConfig $config the system config object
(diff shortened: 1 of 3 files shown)


ArtificialOwl:
_(failing test not related)_

icewind1991 (changes requested):
This would break `BeforeMessageLoggedEvent`

ArtificialOwl:
> This would break `BeforeMessageLoggedEvent`

I added a `hasListeners()` and exit the method if no `BeforeMessageLoggedEvent` is registered. 
Not sure the performance gain anymore. 
Your thoughts ?

lib/private/Log.php
```diff
@@ -329,8 +326,8 @@ public function logException(Throwable $exception, array $context = []) {
 		$level = $context['level'] ?? ILogger::ERROR;
 
 		$minLevel = $this->getLogLevel($context);
-		if ($level < $minLevel && ($this->crashReporters === null || !$this->crashReporters->hasReporters())) {
-			return;
+		if ($level < $minLevel && (($this->crashReporters?->hasReporters() ?? false) === false)) {
```
icewind1991, line 329: missing the `hasListeners` check here

ArtificialOwl: this fix a known issue ?

icewind1991 (approved):
log events still work

skjnldsv (approved):

PR #43531, opened 2024-02-12
get child ids for folder in a separate query during move

icewind1991:
This should help reduce lock contention and hopefully resolve some deadlock issues with concurrent moves.

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -666,6 +666,11 @@ class Cache implements ICache {
 			if ($sourceData['mimetype'] === 'httpd/unix-directory') {
 				//update all child entries
 				$sourceLength = mb_strlen($sourcePath);
+
+				$childIds = $this->getChildIds($sourceStorageId, $sourcePath);
+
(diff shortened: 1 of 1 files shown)


lib/private/Files/Cache/Cache.php
```diff
@@ -711,19 +715,25 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
+					foreach ($childChunks as $chunk) {
+						$query->setParameter('files', $chunk, IQueryBuilder::PARAM_INT_ARRAY);
+						$query->execute();
+					}
 				} catch (\OC\DatabaseException $e) {
 					$this->connection->rollBack();
 					throw $e;
 				}
+			} else {
+				$this->connection->beginTransaction();
```
blizzz, line 736: It does not have a catch and rollback in the current implementation, it might be worth to add it? (optionally)

blizzz (approved):
did not test, is reasonable.

ChristophWurst (approved):

ChristophWurst:
Conflicts

blizzz:
conflicts

PR #43534, opened 2024-02-12
minor syntax update

ArtificialOwl:
Some small minor syntax update.

I have currently no idea how we can manage this FIXME:

https://github.com/nextcloud/server/blob/b8272ea51916cb8178edc55afbc3c51d4b48fa9d/lib/private/Log.php#L74-L77

It comes from a 9 years old commit https://github.com/nextcloud/server/commit/e79c255cabecb768061bb58ea508a38011ef6284
🫂 @MorrisJobke 

2 solutions: 
- removing the code and the null type of `$config`
- or removing the FIXME comment line.

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -62,38 +62,26 @@ use function strtr;
  * MonoLog is an example implementing this interface.
  */
 class Log implements ILogger, IDataLogger {
-	private ?SystemConfig $config;
 	private ?bool $logConditionSatisfied = null;
-	private ?Normalizer $normalizer;
-	private ?IEventDispatcher $eventDispatcher;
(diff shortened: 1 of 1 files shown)


MorrisJobke:
> * or removing the FIXME comment line.

That's a nice one 😆 




> * removing the code and the null type of `$config`

That makes sense, but check if some initializing code passes in `null` as well. I guess that was my lazy self back then. I didn't wanted to check all the callers 🙈

ArtificialOwl:
Hello Sir ! 

Hardcoded initialisation in core are fine, i'll remove your code :)

emoral435 (approved):
Good ol' cleaning commit, love to see this 🥳

Altahrim (approved):

PR #43573, opened 2024-02-14
Fix for ignored `CSP_NONCE` in ContentSecurity Header

HolgerHees:
We should use 'cspNonceManager' for requesting the NONCE value, because it is doing the same as before, except that it honors a CPS_NONCE environment variable if available.

If a CSP_NONCE env var is defined, it is used nearly everywhere, except for setting the correct ContentSecurityPolicy Header.

This commit fixes this.

### DIFF
--- a/lib/private/AppFramework/Middleware/Security/CSPMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/CSPMiddleware.php
@@ -53,7 +53,7 @@ class CSPMiddleware extends Middleware {
 		$defaultPolicy = $this->contentSecurityPolicyManager->mergePolicies($defaultPolicy, $policy);
 
 		if ($this->cspNonceManager->browserSupportsCspV3()) {
-			$defaultPolicy->useJsNonce($this->csrfTokenManager->getToken()->getEncryptedValue());
+			$defaultPolicy->useJsNonce($this->cspNonceManager->getNonce());
 		}
 
 		$response->setContentSecurityPolicy($defaultPolicy);
--- a/core/templates/layout.base.php
+++ b/core/templates/layout.base.php
@@ -14,6 +14,7 @@
 		</title>
 		<meta name="viewport" content="width=device-width, initial-scale=1.0, minimum-scale=1.0">
 		<meta name="theme-color" content="<?php p($theme->getColorPrimary()); ?>">
+		<meta name="csp-nonce" nonce="<?php p($_['cspNonce']); /* Do not pass into "content" to prevent exfiltration */ ?>">
 		<link rel="icon" href="<?php print_unescaped(image_path('core', 'favicon.ico')); /* IE11+ supports png */ ?>">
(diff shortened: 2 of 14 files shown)


susnux (approved):
Looks good!
But then we can remove `csrfTokenManager` from the class as it is no longer used.

susnux:
There might be a problem because all our webpack loaded scripts use the csrf token as nonce -> will break if you use a custom nonce. But we should fix that.

HolgerHees:
> There might be a problem because all our webpack loaded scripts use the csrf token as nonce -> will break if you use a custom nonce. But we should fix that.

@susnux it looks like my change will be merged in version 30.0.0. Is the problem, you mentioned solved?

lib/private/AppFramework/Middleware/Security/CSPMiddleware.php
```diff
@@ -71,7 +71,7 @@ public function afterController($controller, $methodName, Response $response): R
 		$defaultPolicy = $this->contentSecurityPolicyManager->mergePolicies($defaultPolicy, $policy);
 
 		if ($this->cspNonceManager->browserSupportsCspV3()) {
-			$defaultPolicy->useJsNonce($this->csrfTokenManager->getToken()->getEncryptedValue());
+			$defaultPolicy->useJsNonce($this->cspNonceManager->getNonce());
```
nickvergessen, line 74: the value is also `base64_encode(` now which it was not before.
I'm find if @susnux confirms this is not an issue

susnux: We should remove that `base64_encode` it double encodes it, as the token from the csrf manager is already base64 encoded, so this is not a real problem but base64 encode makes the string longer and we should keep the string as short as possible (with same entropy).

Reason: the CSP header is limited in length, so a long nonce can truncate the header value / make browsers ignore CSP.

And as a longer token in this case does not mean better randomness (entropy is the same), we should remove the second encode.

But I think we can do in a separate PR, however you like.

susnux: Nope: `:` is included in the "getEncrypted" return value, this is not allowed for a nonce (strictly following the standard only base64 characters are allowed).

susnux:
> There might be a problem because all our webpack loaded scripts use the csrf token as nonce -> will break if you use a custom nonce. But we should fix that.

For this issue I pushed a commit to provide the CSP nonce as `<meta name="csp-nonce">` important to notice is that we can not use `content` because this can lead to side-channel exfiltration attacks (e.g. using CSS selectors like `meta[content~=...]`) but we need to use the `nonce` attribute as this can not be queried by selectors and only accessed using JS like `element.nonce`.

Ref: https://html.spec.whatwg.org/multipage/urls-and-fetching.html#nonce-attributes

susnux:
On my ToDo: migrate all dynamic scripts to use this.

nickvergessen (approved):

PR #43574, opened 2024-02-14
enh(metadata): migrate metadata to lazy appconfig

ArtificialOwl:
- store the 'known metadata' object, used to store type for each eventual keys, as a lazy appconfig value,
- remove part of code that was only need for 28 migration
- some changes in `IMetadataQuery::__construct()` to pass `IFilesMetadataManager` instead of `IFilesMetadata` to not load lazy appconfig values out of webdav search.
- keeping `__construct()` compatible while generating a debug log when used with deprecated argument

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -170,7 +170,7 @@ class Cache implements ICache {
 		} elseif (!$data) {
 			return $data;
 		} else {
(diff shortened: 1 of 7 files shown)


come-nc (approved):
When are metadata loaded, only for specific queries?

nfebe (approved):

PR #43581, opened 2024-02-14
fix(settings): Deprecate old things for real

nickvergessen:
Psalm output before:
```
ERROR: InvalidDocblock - lib/public/Settings/IManager.php:40:2 - Typo in @deprecated for const in OCP. (see https://psalm.dev/008)
	/**
	 * @since 9.1.0
	 * @depreacted 29.0.0 Use {@see self::SETTINGS_ADMIN} instead
	 */
	public const KEY_ADMIN_SETTINGS = 'admin';


ERROR: InvalidDocblock - lib/public/Settings/IManager.php:46:2 - Typo in @deprecated for const in OCP. (see https://psalm.dev/008)
	/**
	 * @since 9.1.0
	 * @depreacted 29.0.0 Use {@see self::SETTINGS_ADMIN} instead
	 */
	public const KEY_ADMIN_SECTION = 'admin-section';


ERROR: InvalidDocblock - lib/public/Settings/IManager.php:52:2 - Typo in @deprecated for const in OCP. (see https://psalm.dev/008)
	/**
	 * @since 13.0.0
	 * @depreacted 29.0.0 Use {@see self::SETTINGS_PERSONAL} instead
	 */
	public const KEY_PERSONAL_SETTINGS = 'personal';


ERROR: InvalidDocblock - lib/public/Settings/IManager.php:58:2 - Typo in @deprecated for const in OCP. (see https://psalm.dev/008)
	/**
	 * @since 13.0.0
	 * @depreacted 29.0.0 Use {@see self::SETTINGS_PERSONAL} instead
	 */
	public const KEY_PERSONAL_SECTION = 'personal-section';


------------------------------
4 errors found
------------------------------
```

## Follow up

- Also enforce the `@since` for constants in OCP => Follow up PR as there are 223 failures 😱 

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
[...]

PR #43595, opened 2024-02-15
fix(OCP): Add since tag to all constants

nickvergessen:
Red CI due to parallel merges of
- https://github.com/nextcloud/server/pull/43355
- https://github.com/nextcloud/server/pull/43581

PS: I find private constants weird on public API. Any problem with exposing them? Would allowing apps to create their own providers on top of it.


SystemKeeper (approved):

juliushaertl:
I don't see a problem exposing them, but seems a bit weird that private visibility still requires a since annotation 🤷

juliusknorr (approved):

nickvergessen:
> but seems a bit weird that private visibility still requires a since annotation 🤷

Same is the case for "private" methods that you could add atm.

PR #43601, opened 2024-02-15
Enable HMR depending on debug mode

onny:
## Summary

Currently Nextlouc needs to get patched to use Vue Devtools due to "hot module reloading" is restricted, see
- https://github.com/nextcloud-libraries/nextcloud-vue?tab=readme-ov-file#using-vue-devtools-in-firefox
- https://github.com/nextcloud/hmr_enabler

This PR loosens security on debug mode and makes Vue Devtools usable

## Checklist


onny:
ready for review 😊 @skjnldsv @artonge

lib/private/AppFramework/Middleware/Security/CSPMiddleware.php
```diff
@@ -74,6 +74,17 @@ public function afterController($controller, $methodName, Response $response): R
 			$defaultPolicy->useJsNonce($this->csrfTokenManager->getToken()->getEncryptedValue());
 		}
 
+		// Loosen security presets in debug mode to enable development
+		// tools functionality
+		$debugging = \OC::$server->getConfig()->getSystemValue('debug', false);
```
skjnldsv, line 79: please use dependency injection in the construct :)
`\OC::$server->getConfig()` is deprecated

ChristophWurst (changes requested):
I am strongly against this. Most developers have their dev setups in debug mode and never test without. So we'll likely run into features breaking in production because nobody notices CSP errors locally.

onny:
> I am strongly against this. Most developers have their dev setups in debug mode and never test without. So we'll likely run into features breaking in production because nobody notices CSP errors locally.

would it be possible to introduce an extra system setting variable like "enable_hmr" which can be only enabled if debug mode is also enabled?

ShGKme:
HMR requires connection to only one server on a specific host. A proper solution would be to fix only required restrictions for a specific dev server and not to disable CSP completely.

ShGKme:
I'd personally prefer to have a solution in a separated app, not in the server.

PR #43605, opened 2024-02-15
only cleanup orphaned shared daily

icewind1991:
Since we filter out the shares with no matching files when loading them already there is no real downsides to leaving them in the db a bit longer.

Reducing the interval should decrease the db load a bit


juliusknorr (approved):

joshtrichards (approved):
> Since we filter out the shares with no matching files when loading them already

You had me at this. :-) 
Makes sense.

narcisgarcia:
I've seen with Nextcloud 31, cron task does not remove orphan shares.
In my case it was a reshared _External storage_.

PR #43613, opened 2024-02-16
fix: fallback from guessing the owner from path in versioning

icewind1991:
Fixes version hook for files without owner (groupfolders, external storage) when the write is from a context without user (cli, cron).

### DIFF
--- a/apps/files_versions/lib/Listener/FileEventsListener.php
+++ b/apps/files_versions/lib/Listener/FileEventsListener.php
@@ -367,6 +367,16 @@ class FileEventsListener implements IEventListener {
 		}
 
 		$owner = $node->getOwner()?->getUid();
+
+		// If no owner, extract it from the path.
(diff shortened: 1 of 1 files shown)


solracsf (approved):

apps/files_versions/lib/Listener/FileEventsListener.php
```diff
@@ -361,6 +361,14 @@ private function getPathForNode(Node $node): ?string {
 		}
 
 		$owner = $node->getOwner()?->getUid();
+
+		if (!$owner) {
+			$parts = explode('/', $node->getPath(), 4);
+			if (count($parts) === 4) {
+				$owner = $parts[1];
+			}
+		}
```
skjnldsv, line 370: Would be nice if we had comments to explain this :see_no_evil:

skjnldsv (approved):

PR #43625, opened 2024-02-16
Revert "enh: breadcrumbs icons are now more visually descriptive"

emoral435:
Reverts nextcloud/server#43604

We still need to get nextcloud-libraries/nextcloud-vue#5216 merged first before we get the previous PR merged 🥹

Without having the newest NcVue library version, it will introduce a regression as such:

firefox_0G03fcmqBJ

Once the new library gets merged, will add the changes back!


Pytal (approved):

skjnldsv:
We can also just fast forward to a new NC/Vue release. This is 29 as alpha state, that happens :)

emoral435:
Ah, okay! Truthfully, I was a bit nervous about making the breaking visual change on the developer instance, but the buttons still work, and having to submit the same PR again would be a lot. Closing PR unless otherwise told to open again.

Issue #10086, opened 2018-07-03
Allow plus (+) character in username


skroczek:
### Steps to reproduce
1. Create user with username containing plus char

### Expected behaviour
User should be created

### Actual behaviour
An error message with text "Only the following characters are allowed in a username: "a-z", "A-Z", "0-9", and "_.@-'"" occured.

MorrisJobke:
cc @blizzz @nickvergessen @rullzer Opinion on this?

nickvergessen:
causes behaviour problems, because " " is replaced with "+" on urlencode, so we would need to change this everywhere with the full trail of problems.

MorrisJobke:
> causes behaviour problems, because " " is replaced with "+" on urlencode, so we would need to change this everywhere with the full trail of problems.

Full ack.

skroczek:
urlencode encodes " " as +, that's true. But it encodes '+' as '%2B', so I'm not sure if it would be such a big problem. A real problem exists, if you mix `urldecode` and `rawurlencode`. In this case, the already allowed white space leads to a problem:

```php
<?php
// Result: foo+bar@foobar.de
echo rawurldecode(urlencode('foo bar@foobar.de'));
```
All other combinations of rawurldecode and urlencode with '+' and '  ' are fine. And this problem already exists in the current code, so allowing the plus character is not that bad.

blizzz:
👎 No fan of adding strange characters there. It's not supposed to be a user facing value anyhow.

skroczek:
@blizzz I understand, that for commonly used usernames plus must sound as a strange character. But if you allow email addresses as usernames, what you are doing right now, it isn't. Even gmail allows + as a delimiter to tag mail addresses.
But I think it would actually be the best if the characters could be configured. However, I fear that it would take a much greater effort to implement this.

nickvergessen:
> But I think it would actually be the best if the characters could be configured.

No, that does not help at all. Apps need to know what they have to expect. So they would always have to implement for the biggest result set anyway. The thing is if we all of a sudden allow new characters that might cause undefined big problems for apps.

Counting all the Nops in here I see 4 so far, so I'm closing this for now.

nickvergessen:
Workaround: `$uid = str_replace('+', '-', $uid);` in the right place.
If you have 2 people with the same email just one using + and the other one - that's bad luck

Escubaer:
[...]

PR #43634, opened 2024-02-17
Allow + character for usernames as allowed in mail addresses via rfc5233

beposec:
## Summary
This should allow usage of '+' character in usernames as allowed for mail addresses via rfc5233.


solracsf:
See https://github.com/nextcloud/server/issues/10086

beposec:
@solracsf  do you mean I have to fix it in all places, like it happened in owncloud? Or is this customisation not desired for Nextcloud?

solracsf:
In all cases, PR is incomplete; there aare other places to reconsider, and tests to be added or changed.
But wait for an opinion from maintainers.

PR #43646, opened 2024-02-18
Add user:keys:verify command

SystemKeeper:
## Summary

This occ command checks if the stored public key belongs to the stored private key. In case of a mismatch, some nextcloud functionality is broken (e.g. push proxy registration fails). While this should not happen, it is hard to debug when we end up in such a situation. The command is for verification only, in a next step we could add a second command to update the stored public key.

## Checklist

### DIFF
--- a/core/Command/User/Keys/Verify.php
+++ b/core/Command/User/Keys/Verify.php
@@ -1,0 +1,100 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright Copyright (c) 2024, Marcel Müller <marcel.mueller@nextcloud.com>
+ *
+ * @author Marcel Müller <marcel.mueller@nextcloud.com>
+ *
+ * @license AGPL-3.0-or-later
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OC\Core\Command\User\Keys;
+
+use OC\Security\IdentityProof\Manager;
+use OCP\IUser;
+use OCP\IUserManager;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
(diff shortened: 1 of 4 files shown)


SystemKeeper:
I saw that on a production instance for a single user. Is it possible that two requests end up generating a new keypair and one wins to store the private key and one the public key? Otherwise I am not really sure how to end up in that situation.
Ref https://github.com/nextcloud/server/blob/1d8d43f159635b9ef7493bd7d81a8329ec758c87/lib/private/Security/IdentityProof/Manager.php#L99-L103

core/Command/User/Keys/Test.php
```diff
@@ -0,0 +1,100 @@
...
+		protected IUserManager $userManager,
+		protected Manager $keyManager,
+	) {
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('user:keys:test')
+			->setDescription('Test if the stored public key belongs to the stored private key')
```
come-nc, line 48:
```suggestion
			->setDescription('Test if the stored public key matches the stored private key')
```
Maybe it’s me but I find the wording "belongs to" weird here. Please ignore if it’s the correct way to say things and I just do not know about it.

core/Command/User/Keys/Test.php
```diff
@@ -0,0 +1,100 @@
...
+		$output->writeln('');
+
+		$output->writeln('Stored public key:');
+		$output->writeln($publicKey);
+		$output->writeln('Derived public key:');
+		$output->writeln($publicKeyDerived);
+
+		if ($publicKey != $publicKeyDerived) {
+			$output->writeln('<error>Stored public key does not belong to stored private key</error>');
+			return 1;
```
come-nc, line 93:
```suggestion
			return static::FAILURE;
```

core/Command/User/Keys/Test.php
```diff
@@ -0,0 +1,100 @@
...
+		}
+
+		$output->writeln('<info>Stored public key belongs to stored private key</info>');
+
+		return 0;
```
come-nc, line 98:
```suggestion
		return static::SUCCESS;
```

come-nc (commented):
Good apart from nitpick

core/Command/User/Keys/Test.php
SystemKeeper: Good point, yea. Could be my german mind translating stuff 😅 Thanks!

SystemKeeper:
> Good apart from nitpick

Thanks! Changes as suggested and updated the documentation.

come-nc (commented):
Shouldn’t that be part of the encryption application actually?

come-nc (approved):

core/Command/User/Keys/Test.php
```diff
@@ -0,0 +1,100 @@
...
+	public function __construct(
+		protected IUserManager $userManager,
+		protected Manager $keyManager,
+	) {
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('user:keys:test')
```
nickvergessen, line 47: I'd say it should be
```suggestion
			->setName('user:keys:verify')
```

nickvergessen (approved):

PR #43650, opened 2024-02-19
Log host/ip for access violation

Commifreak:
## Summary

During checking some `Host violates local access rules` I noticed, that there is no value being logged. This PR adds the affected Host/IP to the Exception.

## Checklist

### DIFF
--- a/lib/private/Http/Client/Client.php
+++ b/lib/private/Http/Client/Client.php
@@ -192,7 +192,7 @@ class Client implements IClient {
 			throw new LocalServerException('Could not detect any host');
 		}
 		if (!$this->remoteHostValidator->isValid($host)) {
-			throw new LocalServerException('Host violates local access rules');
+			throw new LocalServerException('Host "'.$host.'" violates local access rules');
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


ChristophWurst (approved):
Makes sense

Commifreak:
> Makes sense

I also added the hostName and port to the message.

ChristophWurst (approved):

Altahrim (approved):

skjnldsv:
```
There were 3 failures:

1) lib\Http\Client\DnsPinMiddlewareTest::testRejectIPv4
Failed asserting that exception message 'Host "192.168.0.1" (www.example.com:80) violates local access rules' contains 'Host violates local access rules'.

2) lib\Http\Client\DnsPinMiddlewareTest::testRejectIPv6
Failed asserting that exception message 'Host "fd12:3456:789a:1::1" (ipv6.example.com:80) violates local access rules' contains 'Host violates local access rules'.

3) lib\Http\Client\DnsPinMiddlewareTest::testRejectCanonicalName
Failed asserting that exception message 'Host "192.168.0.2" (www.example.com:80) violates local access rules' contains 'Host violates local access rules'.
```

Commifreak:
Uhm... Thanks for bringing that up. Iam not well experienced with testst and it feels wrong to hard-code the to-test IPs to the `expectExceptionMessage` function. Is there a way to define that the expected exception message contains not the exact string but `violates local access rules`?

PR #43652, opened 2024-02-19
Fixes for getting the filename in the FileInfo class

hopleus:
## Summary
Fixes for getting the filename in the case when the file name key is received, but it is empty.

P.S. Faced with such a problem with ObjectStore enabled in S3 mode

## Checklist


hopleus:
@solracsf Hi. Could you run the tests again?

P.S. Last time cypress crashed with 503 errors and I didn't understand why

icewind1991 (approved):

lib/private/Files/FileInfo.php
```diff
@@ -181,7 +181,11 @@ public function getMimePart() {
 	 * @return string
 	 */
 	public function getName() {
-		return isset($this->data['name']) ? $this->data['name'] : basename($this->getPath());
+		if (isset($this->data['name']) && !empty($this->data['name'])) {
+			return $this->data['name'];
+		}
+
+		return basename($this->getPath());
```
Altahrim, line 188: `empty` should be enough:
```suggestion
		return empty($this->data['name'])
			? basename($this->getPath())
			: $this->data['name'];
```

hopleus: > `empty` should be enough:

I agree, isset is redundant. I corrected the code

Altahrim (approved):

PR #43655, opened 2024-02-19
fix: ignore error when removing "is writable" test file

icewind1991:
the error is safe to ignore


blizzz:
How does it error then?

PR #43668, opened 2024-02-20
feat(core): Expose the confirm password endpoint

provokateurin:
Second try of https://github.com/nextcloud/server/pull/43000

## Summary

This allows non-web clients to confirm the password in order to use endpoints that require it.

## Checklist

### DIFF
--- a/core/Controller/AppPasswordController.php
+++ b/core/Controller/AppPasswordController.php
@@ -31,7 +31,9 @@ namespace OC\Core\Controller;
 use OC\Authentication\Events\AppPasswordCreatedEvent;
 use OC\Authentication\Token\IProvider;
 use OC\Authentication\Token\IToken;
+use OC\User\Session;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\UseSession;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\OCS\OCSForbiddenException;
 use OCP\Authentication\Exceptions\CredentialsUnavailableException;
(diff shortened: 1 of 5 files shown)


provokateurin:
Ah of course this stills needs the frontend changes to work with the new endpoint.

nickvergessen (approved):
Fine by me, but I would prefer to add/duplicate it into https://github.com/nextcloud/server/blob/master/core/Controller/AppPasswordController.php to have it available as proper OCS API

nickvergessen:
> Ah of course this stills needs the frontend changes to work with the new endpoint.

Currently this is non breaking and nothing needs changes? Or am I missing something?

bigcat88 (approved):
The only change is adding `NoCSRFRequired` - right?

provokateurin:
You're both right, I thought I had converted it to OCS which is not true :sweat_smile:

core/Controller/AppPasswordController.php
```diff
@@ -165,4 +172,35 @@ public function rotateAppPassword(): DataResponse {
...
 		]);
 	}
+
+	/**
+	 * Confirm the user password
+	 *
+	 * @NoAdminRequired
+	 * @BruteForceProtection(action=sudo)
+	 *
+	 * @license GNU AGPL version 3 or any later version
```
nickvergessen, line 182: Can be removed as it's the same as the rest of this file:

```suggestion
```

nickvergessen (approved):

Altahrim (approved):

PR #19934, opened 2020-03-13
Simple but effective oauth discovery and userinfo endpoint

jlallana:
After continuing with the changes in this application, I found that for openid connect to work it also needs changes in core. I gave up for the moment and I'm trying to make an application from scratch to give full support to openid connect.

PR #43684, opened 2024-02-20
feat(oauth2): simple userinfo endpoint

hardviper:
## Summary
This solves the oAuth2 authorization issue in openproject when openproject is the client.
Added endpoint that returns a minimum set of user information.
This is very necessary in my openproject integration with nextcloud.
I hope this helps someone else.
Thanks to [@jlallana](https://www.github.com/jlallana) for suggesting this solution in PR #19934.

## Checklist

  - [x] `\OCA\OAuth2\Tests\Controller\OauthApiControllerTest`


hardviper:
@skjnldsv This adds the user_info endpoint, but does not add the discovery. 
Now I have fixed the problem noted by the bot, and also added some more userinfo fields.

skjnldsv:
There was an issue with the rebase, please fix it

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -226,4 +230,32 @@ public function getToken(
...
 		);
 	}
+
+	/**
+	 * @PublicPage
+	 * @NoCSRFRequired
+	 *
+	 * @return JSONResponse
+	 */
+	public function getUserInfo() {
```
ChristophWurst, line 240:
```suggestion
	public function getUserInfo(): JSONResponse {
```

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -226,4 +230,32 @@ public function getToken(
...
+		$user = $this->userSession->getUser();
+		if ($user) {
+			$displayName = $user->getDisplayName();
+			$partedName = explode(' ', $displayName);
+			$userId = $user->getUID();
+			$response = new JSONResponse([
+				'sub' => $userId,
+				'name' => $displayName,
+				'given_name' => $partedName[0],
+				'family_name' => $partedName[1] ?? $partedName[0],
```
ChristophWurst, line 250: does this work internationally?

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -226,4 +230,32 @@ public function getToken(
...
+				'email' => $user->getEMailAddress(),
+				'picture' => $this->urlGenerator->getAbsoluteURL("index.php/avatar/$userId/512"),
```
ChristophWurst, line 252: replace the hard coded URL with a named route

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -226,4 +230,32 @@ public function getToken(
...
+			]);
+		}else{
```
ChristophWurst, line 254:
```suggestion
		} else {
```

coding style

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -226,4 +230,32 @@ public function getToken(
...
+
+	/**
+	 * @PublicPage
+	 * @NoCSRFRequired
+	 *
+	 * @return JSONResponse
+	 */
+	public function getUserInfo() {
+		$user = $this->userSession->getUser();
+		if ($user) {
```
ChristophWurst, line 242:
```suggestion
		if ($user === null) {
		    return new JSONResponse([
				'error' => 'user_not_found',
			], Http::STATUS_NOT_FOUND);
```

make this an early return

ChristophWurst (changes requested):
I'm fine with the endpoint. Some instructions for testing would be appreciated.

The name split logic probably doesn't work for all names. If we can I would strongly favor removing it and only exposing the full name.

hardviper:
@ChristophWurst Thanks for the review! I have committed your suggestions. I also changed the code based on your comments. I kindly ask you to accept proposals for partedname. I really need this and it might help other users when setting up oauth2.

apps/oauth2/lib/Controller/OauthApiController.php
hardviper: Not sure if this is international. I started working on this PR because these are the fields I need in userinfo. Without them, it is not possible to implement authorization into an openproject via nextcloud. I added a locale check so that these fields are only returned if the locale is ru.

ChristophWurst:
Are given_name and family_name required attributes? Are there any specs (RFCs) for the userinfo endpoint?

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -226,4 +232,42 @@ public function getToken(
...
+	 *
+	 * @return JSONResponse
+	 */
+	public function getUserInfo(): JSONResponse {
+		$user = $this->userSession->getUser();
+		if ($user === null) {
+		    return new JSONResponse([
+				'error' => 'user_not_found',
+			], Http::STATUS_NOT_FOUND);
+		} else {
```
ChristophWurst, line 248: the idea with the early return is to that you don't need the else block. remove it please.

hardviper:
@ChristophWurst 
> Are given_name and family_name required attributes? Are there any specs (RFCs) for the userinfo endpoint?

[Standard claims are described here.](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)

Paragraph [5.6.  Claim Types](https://openid.net/specs/openid-connect-core-1_0.html#ClaimTypes) says

> Normal Claims MUST be supported. Support for Aggregated Claims and Distributed Claims is OPTIONAL.

[5.6.1.  Normal Claims](https://openid.net/specs/openid-connect-core-1_0.html#NormalClaims
)
```
  {
   "sub": "248289761001",
   "name": "Jane Doe",
   "given_name": "Jane",
   "family_name": "Doe",
   "email": "janedoe@example.com",
   "picture": "http://example.com/janedoe/me.jpg"
  }
```

ChristophWurst:
Thanks for the specification. Standard claims are what is exposed here: a simple JSON object with no nesting (aggregation). The specification does not say that the standard claims given_name or family_name MUST be exposed. So let's remove them.

ChristophWurst:
@julien-nc do you have more insights? Can we have the userinfo endpoint without the error prone name split?

hardviper:
@ChristophWurst 
Can I implement an additional name separation feature? For example, what if I implement the necessity and name separation parameters in the config?

For example
```
 'oauth2' =>
  [
    'separate_name' => true,
    'separator' => ' ',
    'idx_given_name' => 0,
    'idx_family_name' => 1,
  ],
```

julien-nc:
> do you have more insights? Can we have the userinfo endpoint without the error prone name split?

@ChristophWurst The only viable solution would be to separate first name and last name in NC. I agree with you that since it's not safe to split a display name to deduce first and last name, we could just drop those from the userinfo endpoint response.

hardviper:
@julien-nc What do you say about [this](https://github.com/nextcloud/server/pull/43684#issuecomment-1969146921)?

julien-nc:
@hardviper The main problem with this approach is when names are formatted differently between users. For example if some users have `FIRST_NAME LAST_NAME` and some others `LAST_NAME FIRST_NAME`. Also if there are spaces in the first or last name...

I'm not against your solution if it's optional. Suggestion for the config values:
```
'oauth2' =>
  [
    'process_name' => true,
    'separator' => ' ',
    'first_name_position' => 0,
    'family_name_position' => 1,
  ],
```

apps/oauth2/lib/Controller/OauthApiController.php
hardviper: resolved in [feat(oauth2): simple userinfo endpoint #43684](https://github.com/nextcloud/server/pull/43684/commits/1e264ac59b828d715e78dcf8843f8fdfe931086e)

hardviper:
@julien-nc @ChristophWurst Added optional name separation, it is disabled by default. Can you do a review? What else is needed to merge and backport?

hardviper:
@ChristophWurst Added a test, please check if I did everything right?

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -226,4 +232,46 @@ public function getToken(
...
+		$oauth_conf = $this->config->getSystemValue('oauth2', ['process_name' => false]);
+		if ($oauth_conf["process_name"] === true &&
+			key_exists("separator", $oauth_conf) &&
+			key_exists("first_name_position", $oauth_conf) &&
+			key_exists("family_name_position", $oauth_conf) &&
+			$oauth_conf["separator"] !== ""
+		) {
+			$partedName = explode($oauth_conf["separator"], $displayName);
+			$userInfo['given_name'] = $partedName[$oauth_conf["first_name_position"]];
+			$userInfo['family_name'] = $partedName[$oauth_conf["family_name_position"]] ?? $partedName[0];
```
ChristophWurst, line 272: this path is not covered by tests

hardviper: I added this. Thank you so much for helping me with my first PR.

ChristophWurst (approved):

apps/oauth2/lib/Controller/OauthApiController.php
```diff
@@ -260,7 +260,7 @@ public function getUserInfo(): JSONResponse {
 				]))
 		];
 
-		$oauth_conf = $this->config->getSystemValue('oauth2', ['process_name' => false]);
+		$oauthConf = $this->config->getSystemValue('oauth2', ['process_name' => false]);
 		if ($oauth_conf["process_name"] === true &&
```
ChristophWurst, line 264: the usage os the var need to change too

hardviper: I should be ashamed of that 😳🙂

kesselb:
The approach to split the full name by a given character assumes that every name follows the same pattern. 

Please look at https://en.wikipedia.org/wiki/Personal_name to get a better understanding how different personal names are. 

To solve this problem, we need our own fields to store first name/last name and make the full name a getter from them. 

I assume you can still implement this endpoint via your own Nextcloud app.

hardviper:
> To solve this problem, we need our own fields to store first name/last name and make the full name a getter from them.

Is there an addition of last name and first name fields in the roadmap?

kesselb:
> Is there an addition of last name and first name fields in the roadmap?

Forwarded your question.
Please ping me if we haven't replied within the next 4 weeks.

hardviper:
@kesselb Do you have an answer for the roadmap?

Issue #15192, opened 2019-04-22
Consistent handling of last modified dates when copying/moving files


despens:
**Is your feature request related to a problem? Please describe.**

Nextcloud handles last-modified metadata inconsistently when copying and moving files and folders:

| Activity | Source | Destination | last modified preserved |
| -- | -- | -- | -- |
| file upload > 10 MB via browser | Desktop | Primary Storage | ✔️ |
| file upload < 10 MB via browser | Desktop | Primary Storage | ⛔ |
| file upload via browser | Desktop | External Storage Storage | ⛔ |
| sync files with Desktop client | Desktop | Primary Storage | ✔️ |
| moving files and folders within nextcloud | Primary Storage | Primary Storage | ✔️ |
| moving files and folders within nextcloud | Primary Storage | External Storage | ⛔ |
| moving files and folders within nextcloud | External Storage | Primary Storage | ⛔ |
| copying files via cloud federation | remote nextcloud | local nextcloud | ⛔ |

(Primary storage refers to object storage directly managed by nextcloud, External Storage refers to externally mounted object storage.)

**Describe the solution you'd like**

Last-modification metadata should always be preserved, no matter where files and folders are coming from or moving to.

**Describe alternatives you've considered**

This should just work :wink: 

**Additional context**

For my use case, the date of last modification of of files and folders is critical for comparing different versions of the same item at different locations. It also is important when moving files from active usage into an archival storage (nearline bucket mounted externally).

(related #15177)

codiflow:
I can add that uploading files into nextcloud by using **WebDAV** also doesn't preserve the original timestamps.

despens:
Thx @codiflow, added info to the table, please check if this is correct. Did you try with several WebDAV clients?

codiflow:
> Thx @codiflow, added info to the table, please check if this is correct. Did you try with several WebDAV clients?

I use Krusader on Manjaro Linux and Windows Explorer on Windows 8.1 Pro. With both the current timestamp will set for any upload. With both it doesn't matter if the file is smaller or larger than 10 MB.

CherAlban:
I can confirm the <>10MB behavior with Firefox 66.0.3 (64-bit).
Using WebDav in BeyondCompare the timestamp always gets set to now on upload. (which is very annoying for my use case)

despens:
Using WebDAV via Gnome Shell, can confirm that last-modified dates are not preserved when moving items to nextcloud.

despens:
[...]

Issue #2031, opened 2016-11-06
Folder mtime is not being preserved


Juma7C9:
Hi,
This is one of the original ownCloud issues, that was still debated prior the fork ([Reference](https://github.com/owncloud/core/issues/14008)), and if I am correct, is still not resolved.

--------

As stated before, this problem may affect many user's workflow (as it becomes impossible - for instance - to sort directories according to their creation/modification date), and as it was previously dismissed on performance/bandwidth usage grounds, it was also proposed to introduce this possibility as an optional switch.

### Steps to reproduce
1. Copy some files and folders inside the Nextcloud local folder
2. Wait the folder to be synced with the server
3. Check modification time on the web interface

*Or*

1. Upload some files and folders using the web interface
2. Let the files to be synced to the local client(s)
3. Check local modification time

*Or*

1. Copy some files and directories inside the `data/<User>/files` directory preserving attributes
2. Run `occ files:scan --all`
3. Check modification time on web interface

### Expected behaviour
Files and folders show both the correct original modification time

### Actual behaviour
In all the three scenarios, while files show the original modification time, the directories' mtime becomes the time it was last synced in the first two cases, and the last time it was modified in the latter (regardless of the mtime of the directory on disk. This means that if some directory is being `touch`'ed  and then `occ files:scan --all` is run twice, the mtime is updated only the first time, and it reflects the time of the first run.)

### Server configuration
Debian 8.6 "Jessie"

**Web server:**
Apache 2.4.10

**Database:**
MySQL 5.5.52

**PHP version:**
PHP 5.6.27

**Nextcloud version:**
10.0.1

**Updated from an older Nextcloud/ownCloud or fresh install:**
Fresh install

ckujau:
Why is this marked `stale`? The `mtime` is still not being preserved with the `nextcloud-14.0.4`. I know it's even [documented](https://docs.nextcloud.com/desktop/2.3/navigating.html "Using the Synchronization Client"), but it's still a [bug](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) from a user's POV :-\ Is there any way to get this resolved? Or is there a config toggle involved to switch from tracking "last uploaded" to "last modified"?

kesselb:
@rullzer Is that fixed by filecache_extended?

chainria:
[...]

PR #43689, opened 2024-02-20
feat(tests): Test that mtime is the same after move or rename

come-nc:
## Summary

There is no clear consensus on whether copy should preserve mtime, but rename or move should definetly preserve it.
So adding some tests for this, and if failure are found I may fix some of them in the PR as well.

## Checklist

### DIFF
--- a/apps/files_external/lib/Lib/Storage/AmazonS3.php
+++ b/apps/files_external/lib/Lib/Storage/AmazonS3.php
@@ -573,7 +573,7 @@ class AmazonS3 extends \OC\Files\Storage\Common {
 		return true;
 	}
 
-	public function copy($source, $target, $isFile = null) {
+	public function copy($source, $target, bool $preserveMtime = false, ?bool $isFile = null): bool {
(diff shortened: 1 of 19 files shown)


joshtrichards:
Related open issues: 
- #15192
- #44389 
- #2031

trendco:
At the moment I find it terrible! If you sync with the Windows client, the date is retained. If you add files via WebDav or upload them via the browser, the date is changed to the current date. If you copy within the NC via the browser, it is also set to the current date. If you move it in the browser, it remains the same, but only if you move it to the internal memory! If you move from an externally mounted storage (e.g. smb/cifs), it is also changed to the current date. What a mess

Some want it this way, others that way. Why don't you make this variable, i.e. customisable?

You could set a default via the admin settings, but users could customise this in their own settings.

And you could, for example, insert a checkbox ‘Keep file date’ when copying/moving via the browser, then everyone can decide for themselves what suits them.

It would then be very flexible and everyone could set it as it suits them.

PR #43696, opened 2024-02-20
fix(db): Remove duplicate missing index check for `cards`

joshtrichards:
## Summary

Just looks like it was duplicated during #39487.

Probably not backport worthy since I'm not aware of any real side effects.

## Checklist

### DIFF
--- a/core/Application.php
+++ b/core/Application.php
@@ -165,13 +165,6 @@ class Application extends App {
 
 			$event->addMissingIndex(
 				'cards',
-				'cards_abiduri',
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):

nickvergessen (approved):
Hmm had a quick check, but couldn't find something missing from https://github.com/nextcloud/server/commit/86310a35b25d22ff81aa6ca4ee0c52c769e3d359

So just a double code block it seems

PR #43699, opened 2024-02-20
fix(AppFramework): Correctly implement CSRF checks

provokateurin:
## Summary

These changes ensure that actually all the CSRF protection methods can be correctly used.
I improved the existing codes and allowed requests with OCS-APIRequest headers to pass which was not possible before and a bug.
As far as I can see there were no security problems before and this does not open any new ones :tm:

## Checklist

### DIFF
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -451,8 +451,10 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 
 
 	/**
-	 * Checks if the CSRF check was correct
-	 * @return bool true if CSRF check passed
+	 * Checks if the request passes the CSRF checks.
+	 *
+	 * A request must always pass the strict cookie check, unless it has the OCS-APIRequest set or no session (@see cookieCheckRequired).
+	 * If the OCS-APIRequest is set or a valid CSRF token is sent the request will to pass.
 	 */
 	public function passesCSRFCheck(): bool {
 		if ($this->csrfTokenManager === null) {
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -463,6 +463,10 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 			return false;
 		}
 
+		if ($this->getHeader('OCS-APIRequest') !== '') {
+			return true;
+		}
+
 		if (isset($this->items['get']['requesttoken'])) {
 			$token = $this->items['get']['requesttoken'];
 		} elseif (isset($this->items['post']['requesttoken'])) {
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -481,17 +481,12 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 	/**
 	 * Whether the cookie checks are required
 	 *
+	 * In case the OCS-APIRequest header is set or the user has no session, we don't to check the cookies because the client is not a browser and thus doesn't need CSRF checks.
+	 *
 	 * @return bool
 	 */
 	private function cookieCheckRequired(): bool {
-		if ($this->getHeader('OCS-APIREQUEST')) {
-			return false;
-		}
-		if ($this->getCookie(session_name()) === null && $this->getCookie('nc_token') === null) {
-			return false;
-		}
-
-		return true;
+		return $this->getHeader('OCS-APIRequest') === '' && ($this->getCookie(session_name()) !== null || $this->getCookie('nc_token') !== null);
 	}
 
 	/**
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -487,7 +487,7 @@ class Request implements \ArrayAccess, \Countable, IRequest {
 	/**
 	 * Whether the cookie checks are required
 	 *
-	 * In case the OCS-APIRequest header is set or the user has no session, we don't to check the cookies because the client is not a browser and thus doesn't need CSRF checks.
+	 * In case the OCS-APIRequest header is set or the user has no session, we don't need to check the cookies because the client is not a browser and thus doesn't need CSRF checks.
 	 *
 	 * @return bool
 	 */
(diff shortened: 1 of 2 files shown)


nickvergessen (changes requested):
Seems to break cypress?

lib/private/AppFramework/Http/Request.php
```diff
@@ -451,47 +451,31 @@ protected function decodeContent() {
...
-		} elseif (isset($this->items['server']['HTTP_REQUESTTOKEN'])) {
-			$token = $this->items['server']['HTTP_REQUESTTOKEN'];
-		} else {
-			//no token found.
-			return false;
-		}
-		$token = new CsrfToken($token);
-
-		return $this->csrfTokenManager->isTokenValid($token);
+		return $this->getHeader('OCS-APIREQUEST') || $this->passesStrictCookieCheck() || $this->passesCSRFTokenCheck();
```
ChristophWurst, line 462: this means the check passes if strict cookies are sent but no or an invalid no CSRF token. should it be `&&`?

provokateurin: From my understanding either of the methods are fine, so this should be correct.

come-nc:
Code looks good but I do not know the situation well enough to say if it’s okay to require only one method to pass.

lib/private/AppFramework/Http/Request.php
provokateurin: I started reading some literature about this topic and you are right.
By allowing the request if it passes the cookie check an evil actor that is already on your server could send you a compromised link (e.g. as a Talk message) and it would execute just fine because you are already on the correct host and send the cookies.

To my understanding the cookie check is actually not doing anything at all and doesn't give us any additional protection, because it will not protect against the attack mentioned above and the other methods already provide sufficient protection (especially since very old browsers (7+ years) treat same-site cookies as normal cookies).

The header is a sufficient way to protect against CSRF attacks because there is no way for an attacker to set it in a malicious request.

ChristophWurst: Makes sense!

provokateurin:
I rewrote this PR from scratch with the new information I had. This should be much easier to review. The main problem before was concerning the cookie checks which I didn't understand correctly, but now I also added documentation for the next people who will look at it...

come-nc (commented):
In the end what is the behavior change brought by this PR?

provokateurin:
It allows using the OCS-APIRequest header to bypass CSRF checks. For clients this is necessary, so they don't have to deal with CSRF tokens. Both methods are safe, but the former was not possible until now.

nickvergessen:
So nothing for 29 during freeze :speak_no_evil:

provokateurin:
I guess so :/

provokateurin:
Since the branch-off is done, can I get reviews to get it into Nextcloud 30?

come-nc (approved):

provokateurin:
@nickvergessen @ChristophWurst Can I get your reviews?

lib/private/AppFramework/Http/Request.php
```diff
@@ -451,8 +451,10 @@ protected function decodeContent() {
 
 
 	/**
-	 * Checks if the CSRF check was correct
-	 * @return bool true if CSRF check passed
+	 * Checks if the request passes the CSRF checks.
+	 *
+	 * A request must always pass the strict cookie check, unless it has the OCS-APIRequest set or no session (@see cookieCheckRequired).
+	 * If the OCS-APIRequest is set or a valid CSRF token is sent the request will to pass.
```
Altahrim, line 457:
```suggestion
	 * If the OCS-APIRequest is set or a valid CSRF token is sent the check will succeed.
```

lib/private/AppFramework/Http/Request.php
```diff
@@ -463,6 +465,10 @@ public function passesCSRFCheck(): bool {
 			return false;
 		}
 
+		if ($this->getHeader('OCS-APIRequest') !== '') {
+			return true;
+		}
```
Altahrim, line 470: I think we can move this condition before the `passesStrictCookieCheck`. It saves a few calls.

lib/private/AppFramework/Http/Request.php
```diff
@@ -481,17 +487,12 @@ public function passesCSRFCheck(): bool {
 	/**
 	 * Whether the cookie checks are required
 	 *
+	 * In case the OCS-APIRequest header is set or the user has no session, we don't to check the cookies because the client is not a browser and thus doesn't need CSRF checks.
```
Altahrim, line 490:
```suggestion
	 * In case the OCS-APIRequest header is set or the user has no session, we don't need to check the cookies because the client is not a browser and thus doesn't need CSRF checks.
```

lib/private/AppFramework/Http/Request.php
provokateurin: I agree, although I think it makes sense to always validate the cookies. If invalid cookies are sent they should be rejected regardless of the header.

tests/lib/AppFramework/Http/RequestTest.php
```diff
@@ -2219,7 +2219,7 @@ public function invalidTokenDataProvider() {
 	 * @dataProvider invalidTokenDataProvider
 	 * @param string $invalidToken
 	 */
-	public function testPassesCSRFCheckWithInvalidToken($invalidToken) {
+	public function testFailsCSRFCheckWithInvalidToken($invalidToken) {
```
nickvergessen, line 2222: We usually try to follow the pattern `test...MethodName...ScenarioDescription()`, since the method to test is called `passesCSRFCheck()` the old name matches the pattern, but I also see how it can be confusing in this case. The problem is the actual method name, but we should also not change that... So maybe just stick with the old name of test cases.

provokateurin: Ah I see, I thought the `passes` was from the it passes/fails and not from the method name. I agree that the method name is not good :+1:

lib/private/AppFramework/Http/Request.php
Altahrim: According to the *current* code, `passesStrictCookieCheck` basically return `true` if `OCS-APIRequest`  is not empty. It saves two function call.
Your solution is more future proof ^^

Altahrim (approved):

bigcat88 (approved):

provokateurin:
I made a minimal version of these changes that only include the actual fix and leaves out all the refactoring and documentation: https://github.com/nextcloud/server/pull/46760

PR #43701, opened 2024-02-20
Stop the auto-logout loop after redirecting to the logout page

julien-nc:
Linked with https://github.com/nextcloud/user_oidc/issues/743

If redirecting to the logout page leads to a long series of redirect, the current page's scripts are still active until a new page is actually loaded. So `session-heartbeat.js` will continue redirecting again and again to the logout page.

Solution: Clear the interval that triggers the redirect right after having redirected to `/logout`.

One problematic scenario:
When auto-logout is triggered in the frontend, it is possible that the session has already expired so the `requesttoken` GET param in the logout URL is not valid anymore. In this case, we get redirected to the login page. If user_oidc is configured to provide SSO login, the login flow will immediately starts. This flow can take long enough before a new page is actually loaded so `session-heartbeat.js` has time to redirect again to `/logout` which will start the login flow again. So we enter a weird loop in which the login flow never has time to load any page before the next redirection to `/logout`.

@juliushaertl As far as I could see, logging in when the login page is accessed via `/login?redirect_url=/logout?requesttoken%3DfO5%252FSJDIa873djX6f` is not problematic, there is no redirection to `/logout` after logging in. So I don't think there is no need for a backend fix to avoid a redirection to `/logout` right after logging in.

If this goes in, let's backport it to stable28, stable27 and stable26.


skjnldsv (approved):

skjnldsv:
backports?

juliusknorr (approved):

PR #43747, opened 2024-02-22
feat: Add download limit

### LINKED ISSUE #43566: Move the files_downloadlimit setting to new location
Subtask to #6841

Pytal:
- Resolve https://github.com/nextcloud/server/issues/43566

## TODO

- [ ] ...

## Checklist


apps/files_sharing/src/services/DownloadLimitService.ts
```diff
@@ -0,0 +1,49 @@
...
+	const response = await axios.put(generateOcsUrl('/apps/files_downloadlimit/api/v1/{token}/limit', { token }), {
+		limit,
+	})
+	return response.data.ocs.data
+}
+
+export const deleteDownloadLimit = async (token: string) => {
+	const response = await axios.delete(generateOcsUrl('/apps/files_downloadlimit/api/v1/{token}/limit', { token }))
+	return response.data.ocs.data
+}
```
skjnldsv, line 49: Scoped app APIs calls into another app is something we should avoid :(

Pytal: What alternative would you suggest? Merging the files_downloadlimit into files_sharing?

skjnldsv: > Merging the files_downloadlimit app into files_sharing?

Seeing how closely it starts to be to the core functionalities, could actually be a simpler option yeah :/
More importantly, we already have an API for this, so if the app suddenly step away from this API, what is going to happen of it?

For me there is only two acceptable options
1. (preferably), the current API gets improved so the designers can also work their magic
2. another API gets implemented and we deprecate the previous one (can quickly be messy, not fond of that one)
3. we merge the files_downloadlimit into files_sharing (and we deprecate the API?)

Pytal: Going with 1 as discussed

PR #43727, opened 2024-02-21
Check share status when touching versions

artonge:

PR #43754, opened 2024-02-22
chore(files_versions): improve and fix technical debt on `Version.vue`

skjnldsv:
After https://github.com/nextcloud/server/pull/43727/files#r1497933392

- Do not import some methods that were not designed for the same purpose
- Simplify code 
- Migrate to typescript
- Cleanup and added comments


apps/files_versions/src/components/Version.vue
```diff
@@ -203,59 +208,53 @@ export default {
...
-		enableDeletion() {
+		enableDeletion(): boolean {
 			return this.capabilities.files.version_deletion === true
 		},
 
-		/** @return {boolean} */
-		hasDeletePermissions() {
-			return hasPermissions(this.fileInfo.permissions, Permission.DELETE)
+		hasDeletePermissions(): boolean {
+			return (this.fileInfo.permissions & Permission.DELETE) !== 0
```
artonge, line 232: I honestly don't understand how this is better, this kind of condition is famous for being hard to read and error-prone.

artonge (changes requested):

apps/files_versions/src/components/Version.vue
skjnldsv: We should refrain from importing methods that are from other apps

artonge: Then you can create a new function local to `files_versions` if you want, but the current change is not better in my opinion.

skjnldsv: If you prefer :+1:

artonge (approved):
Thanks for the upgrade :heart:

susnux (approved):
LGTM

PR #43762, opened 2024-02-22
fix(cypress): more stable `navigation-bar-settings` and `versions` cypress tests

skjnldsv:
After https://github.com/nextcloud/server/pull/43360

`testapp` DO have to be moved up twice
image


Altahrim (approved):

susnux (approved):
Thank you also for the comments, really help there!

PR #43768, opened 2024-02-22
Add tests for versions actions

artonge:
The PR initial goal was to tests that permissions were properly checked when triggering versions actions on received shared files. But I wrote some tests utils on the way, and encountered a few bugs.

## Tests changes

- [x] Add utils to e2e tests `files_sharing`
- [x] Add move and copy utils for e2e tests of `files`
- [x] Add test selectors on `files_sharing` and `files_versions` components

## Code changes

- [x] Prevent error in `Version.vue` from `@nc/vue` when deleting an `NcListItem` from one of its actions
- [x] Remove legacy share permissions editor
- [x] Init `ViewOnlyPlugin` after auth to prevent error when using Basic auth


skjnldsv (approved):
:heart_eyes:

artonge:
/backport 51377570530fa299cf96cc60eaa26dbf14cf0872 to stable27

artonge:
/backport 51377570530fa299cf96cc60eaa26dbf14cf0872 to stable26

artonge:
/backport 51377570530fa299cf96cc60eaa26dbf14cf0872 to stable25

apps/files_versions/src/components/Version.vue
```diff
@@ -266,7 +271,11 @@ export default defineComponent({
 			this.$emit('restore', this.version)
 		},
 
-		deleteVersion() {
+		async deleteVersion() {
+			// Let @nc-vue properly remove the popover before we delete the version.
+			// This prevents @nc-vue from throwing a error.
+			await this.$nextTick()
+			await this.$nextTick()
```
skjnldsv, line 278: I feel like we could have something better here :thinking:

artonge: I guess we should prevent the property access in `@nc/vue`, but I was too lazy.

susnux (approved):
Awesome!

I still do not like the data-test... selectors, because that is testing implementation instead of user flow (a user does not look for selectors but for semantic like labels or roles). But I think this is the best we can get here :)

skjnldsv:
> I still do not like the data-test... selectors,

Seconded, didn't we talked about this :see_no_evil:

artonge:
/backport https://github.com/nextcloud/server/commit/51377570530fa299cf96cc60eaa26dbf14cf0872 to stable24

PR #43775, opened 2024-02-23
fix: Enforce forbidden filename characters

### LINKED ISSUE #43603: Implementing usage of forbidden_chars for file/folder uploads via WebDAV
### How to use GitHub

### Feature description

Hi Nextcloud Team,

there is a forbidden_characters=[....] feature in the WebUI that disallows the usage for renaming/uploading/moving of files & folders with specific characters that are forbidden.
Could we also implement this for WebDAV so that it is in sync with the frontend?

Thank you for your effort! :)

Updating the documentation that is clear that forbidden_chars is only used in the WebUI and not respected by Desktop Client / WebDav. ( https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html#forbidden-chars )

**Additional context**

Desktop Client:
https://github.com/nextcloud/desktop/issues/6447

WebUI:
https://github.com/nextcloud/server/pull/40585
https://github.com/nextcloud/server/issues/38614
https://github.com/nextcloud/server/issues/43423

Screenshot Frontend:
+new_folder (6)

susnux:
* Resolves https://github.com/nextcloud/server/issues/43603

## Summary
This is a follow up on #40585 to enforce this also from the backend side.

## Checklist

### DIFF
--- a/lib/private/legacy/OC_Util.php
+++ b/lib/private/legacy/OC_Util.php
@@ -1112,8 +1112,8 @@ class OC_Util {
 			return false;
 		}
 
-		foreach (str_split($trimmed) as $char) {
-			if (str_contains(\OCP\Constants::FILENAME_INVALID_CHARS, $char)) {
+		foreach (\OCP\Util::getForbiddenFileNameChars() as $char) {
+			if (str_contains($trimmed, $char)) {
 				return false;
 			}
 		}
--- a/lib/public/Util.php
+++ b/lib/public/Util.php
@@ -51,6 +51,7 @@ use OC\AppScriptDependency;
 use OC\AppScriptSort;
 use OCP\Share\IManager;
 use Psr\Container\ContainerExceptionInterface;
+use Psr\Log\LoggerInterface;
 
 /**
  * This class provides different helper functions to make the life of a developer easier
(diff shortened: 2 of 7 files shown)


lib/private/legacy/OC_Util.php
```diff
@@ -1091,6 +1091,26 @@ public static function getHumanVersion() {
...
+		}
+
+		// Get admin defined invalid characters
+		$additionalChars = \OC::$server->getConfig()->getSystemValue('forbidden_chars', []);
+		if (!is_array($additionalChars)) {
+			\OC::$server->getLogger()->error('Invalid system config value for "forbidden_chars" is ignored.');
+			$additionalChars = [];
+		}
+		return array_merge($invalidChars, $additionalChars);
+	}
```
skjnldsv, line 1112: Why in this class?

I feel like we should have somewhere else more suitable for this method? :thinking:

susnux: Because `isValidFileName` below is using it. Putting it somewhere else will require this file to import that other place too. Might be cyclic dependency then?

skjnldsv: As we're cleaning and deprecating `OC_Util`, I'd say it would be good to move them :)

skjnldsv (approved):

lib/public/Util.php
```diff
@@ -520,11 +520,32 @@ public static function uploadLimit(): int|float {
...
+		// Get always forbidden characters
+		$invalidChars = str_split(\OCP\Constants::FILENAME_INVALID_CHARS);
+		if ($invalidChars === false) {
+			$invalidChars = [];
+		}
+
+		// Get admin defined invalid characters
+		$additionalChars = \OC::$server->getConfig()->getSystemValue('forbidden_chars', []);
+		if (!is_array($additionalChars)) {
+			\OC::$server->getLogger()->error('Invalid system config value for "forbidden_chars" is ignored.');
```
blizzz, line 538: maybe use the public `\OCP\Server::get->()`

blizzz (approved):
One nitpick, looks good, did not test

unteem:
Nice this can avoid some frustration ! Thanks for this. 

Any reasons that it is not backported to v27 & v28 ?

PR #43776, opened 2024-02-23
[stable27] Update psalm-baseline.xml

nextcloud-command:
Auto-generated update psalm-baseline.xml with fixed psalm warnings


blizzz (approved):

miaulalala (approved):

kesselb:
Rebased and pushed a commit to add the error, about the wrong return type, to the baseline. 
The problem is fixed in newer version of sabre/dav already.

Issue #10207, opened 2018-07-11
CSP nonce by default


rullzer:
In https://github.com/nextcloud/server/pull/10205 I tried to always add the CSP nonce.

As most browsers if unsafe-inline is there and a nonce will ignore the unsafe-inline. This is the CSPv3 backwards compatibility idea.

However this seems to break on Edge. As edge doesn't properly parse the nonce on external resources.

We should investigate how to enable the nonce on more browsers.

MorrisJobke:
> We should investigate how to enable the nonce on more browsers.

Safari: #11966

MichaIng:
@rullzer 
Not sure, if it's related, but since I updated to NC15.0.0 Beta 2, many images (various icons on all NC pages) are blocked by CSP with: `Header always set Content-Security-Policy "default-src 'self' 'unsafe-inline' 'unsafe-eval';"` (Apache)

I guess it's related to this topic, but I don't understand enough to know how to fix best. Simply removing the CSP header works, which is fine for me now, but would be great, after work on this is done/NC15 release to have e.g. some page on docs about the minimum CSP entries required to show all Nextcloud core content.

This is on Opera developer browser v58, but should be not related. I am open for testing/log reports, in case.

rullzer:
You should not set your own csp headers. Nextcloud does this all automatically.

MichaIng:
@rullzer 
Ah okay, makes sense then.

But if I want to set my own CSP headers globally for other/own websites, is it possible to do this without affecting the ones from Nextcloud? Perhaps `Header always set Content-Security-Policy ""` within Nextcloud vhost/location config to unset global headers first?

szaimen:
I'd vote for closing this now that Edge was superseded by Chromium Edge.

MichaIng:
I guess this means to re-open this issue for discussion, doesn't it?

Currently the nonce is still only used based on a whitelist:
- https://github.com/nextcloud/server/blob/215aef3/lib/private/AppFramework/Middleware/Security/CSPMiddleware.php#L74-L76
- https://github.com/nextcloud/server/blob/215aef3/lib/private/Security/CSP/ContentSecurityPolicyNonceManager.php#L80-L95

If compatibility with old Microsoft Edge is not required, then #10205 could be reattempted, switching to a blacklist. But there are a lot of browsers out there, including the mobile variants, so much to test 🙂.

szaimen:
cc @nextcloud/security

susnux:
[...]

PR #43778, opened 2024-02-23
Fix CSP for script-src with nonce on edge

juliusknorr:
This fixes issues that may occur using MS Edge where the usage of a nonce for script-src was not indicated in the CSP headers.

https://learn.microsoft.com/en-us/microsoft-edge/web-platform/user-agent-guidance

Short summary for the user agent pattern, `Edg/<Version>` is the new way to detect instead of `Edge/<Version>`

- fix: Adjust user agent pattern for Edge
- fix: Add edge as supported user agent for CSPv3 nonces
- fix: Allow nonce in csp header also if no other reasons are given

## Checklist

### DIFF
--- a/lib/private/AppFramework/Http/Request.php
+++ b/lib/private/AppFramework/Http/Request.php
@@ -68,7 +68,7 @@ use Symfony\Component\HttpFoundation\IpUtils;
 class Request implements \ArrayAccess, \Countable, IRequest {
 	public const USER_AGENT_IE = '/(MSIE)|(Trident)/';
 	// Microsoft Edge User Agent from https://msdn.microsoft.com/en-us/library/hh869301(v=vs.85).aspx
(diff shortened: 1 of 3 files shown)


susnux:
Alternative would be to throw away `browserSupportsCspV3` and instead always use the nonce #10207 

It seems like all our supported browsers support CSP nonce, if you find a browser that does not it probably will also not be able to parse our JS code.

nickvergessen (approved):

susnux (approved):

susnux:
> Makes sense, I'd still like to keep the commits here to be able to backport them to 28 and only do the full removal for master then.

Sounds good, so just fixup the commits :)

Orchal:
Hi, it should, the pull request has been merged! I did not test yet

PR #43804, opened 2024-02-24
fix(settings): Split group lists

Pytal:
## Summary

- Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/5302
- Requires https://github.com/nextcloud-libraries/nextcloud-vue/pull/5347
- Requires NcVue release
- Adjust app navigation to use multiple lists

Before | After
--- | ---
image | image

## Checklist


JuliaKirschenheuter (approved):
Great! Works ;)

emoral435 (approved):
This looks good to me 😀 is this potentially linked to one of the issues within the board?

ShGKme:
I don't know if it is `@nextcloud/vue` or local issue, but scrolling is only applyed to the `NcAppNavigation` itself

scrolling

It makes groups least unreachable on a11y viewport size

image

Pytal:
Good point @ShGKme :)

Removed the scrolling on purpose because `height: fit-content` with `overflow: auto` still makes it scroll and to keep the same design but I think adding a separate container which scrolls like `ul.app-navigation__list` when using default slot makes sense (in this case NcAppNavigationList)

image

JuliaKirschenheuter:
I've tested it and it looks ok and fixes initial problem:

Screenshot from 2024-03-06 18-11-08

apps/files/src/views/Navigation.vue
```diff
@@ -20,63 +20,67 @@
...
-				<!-- Child views if any -->
-				<NcAppNavigationItem v-for="child in childViews[view.id]"
-					:key="child.id"
-					:data-cy-files-navigation-item="child.id"
-					:exact-path="true"
-					:icon="child.iconClass"
-					:name="child.name"
-					:to="generateToNavigation(child)">
+	<Fragment>
+		<NcAppNavigation data-cy-files-navigation
```
ShGKme, line 24: The change with Fragment is not needed after the revert in `@nextcloud/vue`

ShGKme:
The bug on a small screen is not super critical, we can merge/backport the PR first and then updating on `@nextcloud/vue` should fix the layout for a small screen.

ShGKme (approved):

skjnldsv:
Broke cypress :grin:

PR #43855, opened 2024-02-27
Implement team provider api

### LINKED ISSUE #43683: API for teams
Part of https://github.com/nextcloud/server/issues/42433

In order to implement the different integration parts across apps we need a public API that allows

- List all providers that can share to a team (circle)
- List all entries of each providers that are shared with a given team (circle)

Example:
- Have a deck provider that list all deck boards shared with a circle
- Have a collectives provider that returns the collective of a given circle (there is only one as of right now)

## To be clarified if in scope
- Create a new share with a team through a generic API

juliusknorr:
Resolves https://github.com/nextcloud/server/issues/43683

Teams are considered evolution of circles where we now want to aggregate all resources that are related (shared or owned) by a given team. Therefore we need a generic API so that apps that offer sharing to teams can expose the list of resources and teams for a given resource.

Example implementation can be found for deck and collectives already.

In the future we may extend this namespace to become a stable OCP/OCS API for other methods related to circles which for now is still bound to the classes in `OCA\Circles`.

## ToDo

- [x] Verify team membership
- [ ] API documentation (openapi, phpdoc, dev docs)
  - [x] phpdoc: Define that resources should be returned in order of importance (e.g. folders before files)
  - [x] phpdoc: Clarify order of icons
  - [x] OpenAPI spec
  - [ ] Developer docs

## Provider implementations
- [ ] Collectives https://github.com/nextcloud/collectives/pull/1130
- [ ] Deck https://github.com/nextcloud/deck/pull/5625
- [ ] Files sharing
- [ ] Groupfolders
- [ ] Talk
- [ ] Others?


apps/files_sharing/lib/FilesTeamResourceProvider.php
```diff
@@ -0,0 +1 @@
+<?php
```
blizzz, line 1: :eyes:

core/Controller/TeamsApiController.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * @copyright Copyright (c) 2022 Julius Härtl <jus@bitgrid.net>
```
blizzz, line 5:
```suggestion
 * @copyright Copyright (c) 2024 Julius Härtl <jus@bitgrid.net>
```

lib/public/Teams/ITeamManager.php
```diff
@@ -0,0 +1,58 @@
...
+
+	/**
+	 * Get a specific team resource provider by its id
+	 *
+	 * @since 29.0.0
+	 */
+	public function getProvider($providerId): ITeamResourceProvider;
+
+	/**
+	 * Returns all team resources for a given team for the current user
```
blizzz, line 45: what if there is no current user (e.g. background jobs)?

lib/public/Teams/ITeamManager.php
```diff
@@ -0,0 +1,58 @@
...
+	 *
+	 * @return TeamResource[]
+	 * @since 29.0.0
+	 */
+	public function getSharedWith($teamId): array;
+
+	/**
+	 * Returns all teams for a given resource for the current user
```
blizzz, line 53: same question here

lib/private/Teams/TeamManager.php
```diff
@@ -0,0 +1,122 @@
...
+	public function getProviders(): array {
+		if ($this->providers !== null) {
+			return $this->providers;
+		}
+
+		$this->providers = [];
+		foreach ($this->bootContext->getRegistrationContext()->getTeamResourceProviders() as $providerRegistration) {
+			try {
+				$provider = \OCP\Server::get($providerRegistration->getService());
+				$this->providers[] = $provider;
```
blizzz, line 66: consider faster lookup in getProvider() by using the ID as key. Provided (free pun) there are no clashes… otoh currently the first registration would win.

lib/public/Teams/ITeamManager.php
```diff
@@ -0,0 +1,58 @@
...
+	 * @since 29.0.0
+	 */
+	public function getProviders(): array;
+
+	/**
+	 * Get a specific team resource provider by its id
+	 *
+	 * @since 29.0.0
+	 */
+	public function getProvider($providerId): ITeamResourceProvider;
```
blizzz, line 42: no type hint?

lib/public/Teams/ITeamManager.php
```diff
@@ -0,0 +1,58 @@
...
+
+	/**
+	 * Returns all team resources for a given team for the current user
+	 *
+	 * @return TeamResource[]
+	 * @since 29.0.0
+	 */
+	public function getSharedWith($teamId): array;
```
blizzz, line 50: no type hint? In a private implementation's private method (getTeam) it is set to string.

lib/private/Teams/TeamManager.php
```diff
@@ -0,0 +1,122 @@
...
+
+			return new Team(
+				$teamId,
+				$team->getDisplayName(),
+				$this->urlGenerator->linkToRouteAbsolute('contacts.contacts.directcircle', ['singleId' => $teamId]),
+			);
+		}, $provider->getTeamsForResource($resourceId))));
+	}
+
+	private function getTeam(string $teamId): ?Circle {
```
blizzz, line 113: here the team id requires being string, but it was never checked before

lib/private/Teams/TeamManager.php
```diff
@@ -0,0 +1,122 @@
...
+		try {
+			$federatedUser = $this->circlesManager->getFederatedUser($this->userId, Member::TYPE_USER);
```
blizzz, line 115: $this->userId might be null

core/Controller/TeamsApiController.php
```diff
@@ -0,0 +1,92 @@
...
+	#[ApiRoute(verb: 'GET', url: '/{teamId}/resources', root: '/teams')]
+	public function resolveOne(string $teamId): DataResponse {
+		/** @var CoreTeamResource[] $resolvedResources */
+		$resolvedResources = $this->teamManager->getSharedWith(trim($teamId));
+
+		return new DataResponse(['resources' => $resolvedResources]);
+	}
+
+	/**
+	 * Get all teams of a resources
```
blizzz, line 69:
```suggestion
	 * Get all teams of a resource
```

core/Controller/TeamsApiController.php
```diff
@@ -0,0 +1,92 @@
...
+	 * @return DataResponse<Http::STATUS_OK, array{teams: CoreTeam[]}, array{}>
+	 *
+	 * 200: Teams returned
+	 */
+	#[NoAdminRequired]
+	#[ApiRoute(verb: 'GET', url: '/resources/{providerId}/{resourceId}', root: '/teams')]
+	public function listTeams(string $providerId, string $resourceId): DataResponse {
+		$teams = $this->teamManager->getTeamsForResource($providerId, $resourceId);
+		/** @var CoreTeam[] $teams */
+		$teams = array_map(function (Team $team) {
```
blizzz, line 82: array_map is bad for RAM usage, would see a potential danger on big instances/many teams.

blizzz (commented):
just some detail findings (that partly I would have expected to be found by CI). In general, everything is layed out fine :+1:

core/ResponseDefinitions.php
```diff
@@ -161,6 +161,21 @@
...
+ *      icon: string,
+ * }
+ *
+ * @psalm-type CoreTeamResource = array{
+ *       id: int,
+ *       label: string,
+ *       url: string,
+ *       iconSvg: ?string,
+ *       iconURL: ?string,
+ *       iconEmoji: ?string,
```
provokateurin, line 177:
```suggestion
 *       iconSvg?: string,
 *       iconURL?: string,
 *       iconEmoji?: string,
```
IMO this makes the API simpler, but it is a bit more work to implement, so you might not want to go this route.

provokateurin (commented):
OpenAPI LGTM

core/Controller/TeamsApiController.php
juliushaertl: I'd say this should mostly be fine as we only list teams that a resource is shared with AND a the current user is member of here.

lib/public/Teams/ITeamManager.php
juliushaertl: Adjusted to require the user id to be passed along. Makes further checking as commented below also more straight forward then depending on the user session in here. :+1:

provokateurin (approved):
OpenAPI is good

core/Controller/TeamsApiController.php
blizzz: Mh, you did not set the 3rd – _userId_ – parameter in line 80.

juliushaertl: Sorry, one should push all local changes. Did a few more rounds to get psalm all happy. Should be good now.

juliushaertl:
Cypress failure is unrelated

core/Controller/TeamsApiController.php
blizzz: >  Sorry, one should push all local changes.

Came to the same conclusion recently :see_no_evil:

blizzz (approved):

joshtrichards:
Does this mean we need to add the Circles app to `alwaysEnabled` (as a shipped app) due to the tight coupling?

blizzz:
> Does this mean we need to add the Circles app to `alwaysEnabled` (as a shipped app) due to the tight coupling? (#45484 brought this up today).

One way to solve it, I'd rather say we need a `NullProvider` sort of things, provided it does not lead to odd effects/assumptions.

PR #43862, opened 2024-02-27
Add ability to cancel a scheduled transcription

julien-nc:
This makes it possible to cancel a scheduled transcription given the same information provided to schedule it.

We don't get the job ID when adding the job in the jobList. Even if we did, it is not possible to use job IDs to remove a job from the list.


marcelklehr (approved):

emoral435 (approved):
Nice changes :)

PR #43876, opened 2024-02-27
fix: changed text alignment and icon logic 

### LINKED ISSUE #43839: [Bug]: Padded breadcrumbs

### Bug description

image

### Steps to reproduce

1. Open Files
2. Go into a directory

### Expected behavior

The directory shows in the breadcrumbs with no additional padding on its left

### Installation method

None

### Nextcloud Server version

28

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

emoral435:
* Resolves #43839
## Summary
1. Previously, for every breadcrumb, we where loading an icon alongside its text. Because we where using NcButton, this meant that the button was trying to load an icon and its accompanying text, splitting the button width into two, which is why the padding looked funky!
2. The icon rendering logic now needs to change in order to fit the new NcBreadcrumb logic, so we serve the component an icon that we use now instead of having it on every section!

Unrelated, but while testing this, it seems that the uploading section of this component seems to block the view of the breadcrumbs even when you are not uploading something. May have to do with our upload stores queue not updating correctly? Will test later!

## Screenshots of the new change
SCR-20240227-jtcg

## Checklist


apps/files/src/components/BreadCrumbs.vue
```diff
@@ -97,6 +99,13 @@ export default defineComponent({
 		}
 	},
 
+	data() {
+		return {
+			// used to show the views icon for the first breadcrumb
+			icon: this.$navigation.active?.icon ?? null
```
susnux, line 105: This does not need to be a reactive variable, it is never set again so it stays at the initial value.
So I would move this to `computed` instead.

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -36,7 +36,9 @@
 			:aria-description="ariaForSection(section)"
 			@click.native="onClick(section.to)">
 			<template v-if="index === 0" #icon>
-				<NcIconSvgWrapper v-if="section.icon" :size="20" :svg="section.icon" />
+				<NcIconSvgWrapper v-if="icon"
```
susnux, line 39: Would this not just add an empty icon slot in case there is no icon? How about this:

```vue
			<template v-if="index === 0 && icon" #icon>
				<NcIconSvgWrapper :size="20" :svg="icon" />
```

skjnldsv: No because we want to fallback to the Home con if the section doesn't have one defined :)

susnux: Ah yes I see

apps/files/src/components/BreadCrumbs.vue
emoral435: Agreed! Will change this :)

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -132,6 +133,11 @@ export default defineComponent({
 		shouldShowBreadcrumbs(): boolean {
 			return this.filesListWidth < 768 && !this.isUploadInProgress
 		},
+
+		// used to show the views icon for the first breadcrumb
+		getViewIcon(): string | null {
+			return this.$navigation.active?.icon ?? null
+		}
```
susnux, line 140: I do not think computed are getters, so I would go with `viewIcon` directly.
But how about simplify the code a but and use:

```suggestion
		getViewIcon(): string {
			return this.$navigation.active?.icon ?? homeSvg
		}
```

Then you can remove the `v-if` `v-else` in the template. And replace the import of `Home` with:
`import homeSvg from '@mdi/svg/svg/home.svg?raw'`

skjnldsv: Clean!

susnux (approved):
Looks good to me now :)

skjnldsv (approved):

emoral435:
Fixed 🐘

skjnldsv:
https://github.com/nextcloud/server/pull/43773 is most likely getting in first :see_no_evil:

PR #43877, opened 2024-02-27
fix(provisioning_api): Fix AppInfo schema for OpenAPI

provokateurin:
## Summary

There are a few mistakes in the current schema (e.g. multiple language support for name, description and summary). Instead of trying to fix those problems I opted for generalizing the whole schema as the provided type information was mostly useless anyway (values typed as `?mixed`). This could have also led to problems in case the XSD was updated and this schema wasn't, so this avoids it. The real problem is that the AppStore data is simply passed through without any checking for any structure, but that can't be fixed easily.

## Checklist


come-nc (approved):
:shrug:

andrey18106 (approved):

PR #43884, opened 2024-02-28
fix(federation): Allow cloud federation providers to handle unsuccess…

nickvergessen:
…ful return codes

Otherwise they are put to retry and will immediately trigger bruteforce protection infinitely. This is currently easily producable when a federated user leaves a room during a "netsplit", maintenance mode or with a temporary error and afterwards people posting in the normal chat room.

* Required for https://github.com/nextcloud/spreed/issues/11278

## Checklist

### DIFF
--- a/lib/public/Federation/ICloudFederationProviderManager.php
+++ b/lib/public/Federation/ICloudFederationProviderManager.php
@@ -23,6 +23,9 @@
  */
 namespace OCP\Federation;
 
+use OCP\Http\Client\IResponse;
+use OCP\OCM\Exceptions\OCMProviderException;
+
 /**
  * Class ICloudFederationProviderManager
  *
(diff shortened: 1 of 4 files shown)


lib/public/Http/Client/IClient.php
```diff
@@ -207,6 +207,16 @@ public function delete(string $uri, array $options = []): IResponse;
...
 
+	/**
+	 * Get the response of a Throwable thrown by the request methods when possible
+	 *
+	 * @param \Throwable $e
+	 * @return IResponse
+	 * @throws \Throwable When $e did not have a response
+	 * @since 29.0.0
+	 */
+	public function getResponseFromThrowable(\Throwable $e): IResponse;
```
nickvergessen, line 218: I added and used this method. I also put it in OCP, because it's not the first time where I find myself wanting to get the response of a "Guzzle ClientException" as an IResponse, without `new \OC\…\Response()` violating private class access.

lib/public/Federation/ICloudFederationProviderManager.php
```diff
@@ -80,9 +83,18 @@ public function getCloudFederationProvider($resourceType);
...
 	 */
 	public function sendShare(ICloudFederationShare $share);
 
+	/**
+	 * @param ICloudFederationShare $share
+	 * @return IResponse
+	 * @throws OCMProviderException
+	 * @since 29.0.0
+	 */
+	public function sendCloudShare(ICloudFederationShare $share): IResponse;
```
miaulalala, line 96: why add two new methods when you could modify the exisitng one?

miaulalala (approved):

lib/public/Federation/ICloudFederationProviderManager.php
nickvergessen: because it would now throw instead of returning false, which would break the current consumers?

SystemKeeper (approved):

PR #43944, opened 2024-03-01
fix(settings): emails actions a11y and design

ShGKme:
## Summary

* Revert, a part of refactoring of `FederationActions.vue` from https://github.com/nextcloud/server/pull/43488/
  * `NcActions` doesn't support having `non-NcAction*` as direct children which causes a11y issues
* Change the design of additional emails and their actions, close to the old design
  * Use 2 buttons instead of 2-level menu
  * Show "This address is not confirmed" as input description instead of a caption in a menu
  * Add a gap between additional emails, otherwise they overlaps
  * Remove a margin on the right after actions
  * Renamed label "Additional email address N" with "Additional email N"

## Screenshots

All screenshots are taken on the layout with the smallest width of inputs

Old design | Before | After
---|---|---
emails-old | emails-before | emails-new

After design discussions:

image

## Checklist


ShGKme:
> Then we either should throw / show an error when non NcAction* components are used

Yep, though about it, should be added

> or fix the NcActions component to properly handle those. Otherwise this sounds like a hack.

I don't see any reliable way to fully fix this case, except quite complex solutions. Having several days before the final a11y testing and only a single place with this issue I'd prefer to go with a simple solution in place.

> Is this accessibility related?

Both.

From a design perspective, I'd try to avoid multi-level menus for simplicity and having a short way to the required action. It makes sense in a large menu, but here we have only 1-2 actions outside of the federation scope.

From an a11y perspective, I'm not sure. Neither submenu button nor the back button has anything except a text telling the user that it changes the menu, and I don't know if it is fine or not. 

For example, in Files "Set reminder" has a bit broken focus behavior and no labeling that "Set reminder" on submenu is actually a Back button.

In this place, for me, NVDA doesn't react on moving to the submenu and back until I press an arrow key.

In general, I'd say we should have an interface to provide submenus out of the box in `NcActions`.

> @nextcloud/designers whats your opinion here? Asking because for me it looks odd that the secondary email input fields now are a different width than the primary email.

We can also make the first input shorter 😀

ShGKme:
> Or use label outside for additional emails same as other labels on this page. Then scope federation menu can be on the label.

Draft:

image

susnux:
Or make the federation scope the trailing button? Then all input fields have the same width.

ShGKme:
> Or make the federation scope the trailing button? Then all input fields have the same width.

But we cannot put `NcActions` to the trailing button, `NcInputField` slot is only for an icon

jancborchardt (commented):
> Or make the federation scope the trailing button? Then all input fields have the same width.

The federation scope looks better next to the label, and also makes more sense next to it as it is closer to its description.

---

@ShGKme I would be fine with your draft too. If accessibility-wise it is fine to have the button floating inside the input (on the right) like in the old design, then we could do that?
It would accomodate the actions, and ensure that all fields have the same width. It is also commonly done with e.g. password input visibility buttons.

ShGKme:
> If accessibility-wise it is fine to have the button floating inside the input (on the right) like in the old design, then we could do that?

It is a problem implementation-wise with our input fields.

First, we can only specify an icon for the trailing button but not the button. So we cannot put an actions menu there.

Second, on error/success state there is always an icon.

image

All these problems are solvable, but with dirty hacks.

ShGKme:
@jancborchardt Is it fine?

image

JuliaKirschenheuter (approved):

jancborchardt (approved):
Yeah, that looks great in fact! :) The error icon was just duplicated so good on removing it.

susnux (approved):
Designers said looks good, code looks fine, so :+1:  :smile:

ShGKme:
- Rebased onto `master`
- Squashed `fixup` commits
- A small change - keep original input top padding to be consistent with all of the other inputs
- A small change - moved styles with margin between additional emails from `Email` component to `EmailSection` component

Pytal (approved):
Tested, and is much nicer!

Pytal:
Cypress is related

skjnldsv:
Thanks for this @ShGKme !! :bow: :hugs:

ShGKme:
Backport is not needed, at least, unless #43488 is backported

PR #43946, opened 2024-03-01
fix: changed login-page to reflect correct LDAP settings

### LINKED ISSUE #3278: Do not advertize the use of email as a valid username if explicitely disabled in configuration
By default, the username input box on the login page has a default value of "Username or email". While this is true with the default configuration, it will still be displayed even if ldap_loginfilter_email is set to false in the ldap configuration, thus misleading users.

The placeholder value for this input should be dependant of the dap_loginfilter_email and ldap_loginfilter_username values.

emoral435:
* Resolves nextcloud-gmbh/customer-feature-requests#315
* Resolves nextcloud/server#3278
## Summary
Special thanks to Côme for pointing me in the right direction, and giving me more insight into the problem.
* Users can have multiple servers configured for their LDAP configuration. On some, they can have email enabled, and on others, they can have emails not enabled.
* I do not believe users can choose which server to log in to on the `Log In` page, so if any of their servers have their emails disabled, then it is best to not convey that the user can login with their email.

[feat: added login's initial possible email-states](https://github.com/nextcloud/server/pull/43946/commits/5a8af329259714bb893add5e7b591d61c7df0ead) -> LoginController shares emailLDAP settings to the frontend

[fix: login page now correctly shows email LDAP settings](https://github.com/nextcloud/server/pull/43946/commits/383ccfedf9894bc0d5477a29b438444c7d6fffff) -> frontend conveys their setting correctly

🖼️ Screenshots
🏚️ Cant use email | 🏡 Can use email
---|---
firefox_fbrOsH8Qci| firefox_OEEF70tTen

## Checklist


emoral435:
Ignore the github security - all are outdated :)

core/Controller/LoginController.php
```diff
@@ -82,8 +85,12 @@ public function __construct(
 		private WebAuthnManager $webAuthnManager,
 		private IManager $manager,
 		private IL10N $l10n,
+		private Helper $helper,
+		private IDBConnection $connection,
+		private OC_App $app,
 	) {
 		parent::__construct($appName, $request);
+		$this->helper = new Helper($config, $connection);
```
come-nc, line 93: Is that needed?
If it is, remove the $helper parameter otherwise DI will build it and you re-build it.
If possible remove this line and rely on the DI.
```suggestion
```

core/Controller/LoginController.php
```diff
@@ -230,6 +239,30 @@ private function setPasswordResetInitialState(?string $username): void {
...
+	
+	/**
+	 * Sets the initial state of whether or not a user is allowed to login with their email
+	 * initial state is passed in the array of 1 for email allowed and 0 for not allowed
+	 */
+	private function setEmailStates(): void {
+		$emailStates = []; // true: can login with email, false otherwise - default to true
+
+		// check if user_ldap is enabled.
+		$enabledApps = $this->app->getEnabledApps();
```
come-nc, line 251: Please inject the app manager instead of legacy OC_App.
It’s `\OCP\App\IAppManager`.
Also no need to get the full list to know if user_ldap is enabled, use maybe `isAppLoaded`.

But actually I’m wondering, how can the constructor works with user_ldap disabled since you inject a class from user_ldap namespace?

I think you’ll have to do something like:
```php
if (class_exists(Helper::class)) {
$helper = \OCP\Server::get(Helper::class);
/* Rest of your code */
}
```

emoral435: Ahhh, I see what you mean. I agree, these changes would ensure that there is no error in logic when handling between user enabled apps 👍

core/Controller/LoginController.php
```diff
@@ -230,6 +236,32 @@ private function setPasswordResetInitialState(?string $username): void {
...
+	 * Sets the initial state of whether or not a user is allowed to login with their email
+	 * initial state is passed in the array of 1 for email allowed and 0 for not allowed
+	 */
+	private function setEmailStates(): void {
+		$emailStates = []; // true: can login with email, false otherwise - default to true
+
+		// check if user_ldap is enabled, and the required classes exist
+		if ($this->appManager->isAppLoaded('user_ldap')
+			&& class_exists(Helper::class)
+			&& class_exists(Configuration::class)) {
```
come-nc, line 250: I would say only one of those checks is enough.

core/src/components/login/LoginForm.vue
```diff
@@ -207,6 +213,15 @@ export default {
 		loginActionUrl() {
 			return generateUrl('login')
 		},
+		emailEnabled() {
+			return this.emailStates ? this.emailStates.every((state) => state === '1') : 1
```
come-nc, line 217: I would go the other way and show "or email" in **any** configuration allows it, no? Was there a design decision for this?

come-nc (commented):
Looks good, only a bit too precautionous on the class_exists tests, if the app is loaded the class are there, or if a class is there the app is loaded, no need to test both.

core/src/components/login/LoginForm.vue
emoral435: Hm. I can go either way on this, personally, however the ticket mentions:

> But on the login screen the users are asked for username or email address, which is wrong and so confuses users, which results in support effort.

IMO, I think having a user try to log in with their email and not be able to login in, and then cause additional support effort from the company is a bigger problem, which is why I chose to change the text if any server has email login's disabled. Thoughts? :)

core/Controller/LoginController.php
emoral435: Agreed :)

come-nc (approved):
I do not have a strong opinion about `every` vs `any`, so I let you decide.

skjnldsv (approved):

skjnldsv:
```
11) Tests\Core\Controller\LoginControllerTest::testLoginWithValidCredentials
ArgumentCountError: Too few arguments to function OC\Core\Controller\LoginController::__construct(), 13 passed in /home/runner/work/server/server/tests/Core/Controller/LoginControllerTest.php on line 135 and exactly 14 expected
```

Failure related

emoral435:
Fixed PHP test issues! It was due to me adding a new DI into LoginControllers constructor, and I did not change that for the tests, so it was hollering at me to change it 👌

susnux:
Error is related:

> There were 4 failures:
>
>1) Tests\Core\Controller\LoginControllerTest::testShowLoginFormForFlowAuth
OCP\IInitialStateService::provideInitialState('core', 'alternativeLogins', Array ()): void was not expected to be called more than 12 times.
>
>/home/runner/work/server/server/core/Controller/LoginController.php:203
>/home/runner/work/server/server/tests/Core/Controller/LoginControllerTest.php:331
>
>2) Tests\Core\Controller\LoginControllerTest::testShowLoginFormWithPasswordResetOption with data set #0 (true, true)
OCP\IInitialStateService::provideInitialState('core', 'alternativeLogins', Array ()): void was not expected to be called more than 11 times.
>
>/home/runner/work/server/server/core/Controller/LoginController.php:203
>/home/runner/work/server/server/tests/Core/Controller/LoginControllerTest.php:402
>
>3) Tests\Core\Controller\LoginControllerTest::testShowLoginFormWithPasswordResetOption with data set #1 (false, false)
OCP\IInitialStateService::provideInitialState('core', 'alternativeLogins', Array ()): void was not expected to be called more than 11 times.
>
>/home/runner/work/server/server/core/Controller/LoginController.php:203
>/home/runner/work/server/server/tests/Core/Controller/LoginControllerTest.php:402
>
>4) Tests\Core\Controller\LoginControllerTest::testShowLoginFormForUserNamed0
OCP\IInitialStateService::provideInitialState('core', 'alternativeLogins', Array ()): void was not expected to be called more than 11 times.
>
>/home/runner/work/server/server/core/Controller/LoginController.php:203
>/home/runner/work/server/server/tests/Core/Controller/LoginControllerTest.php:456

emoral435:
fixed tests 👍

PR #43975, opened 2024-03-04
Handle more cases in the MergeDistributive search query optimizer

icewind1991:
If an `AND`/`OR` operator contains more than just `AND`/`OR` arguments it was previously skipped. Now those parts are factored out properly and the rest is still optimized.


starypatyk (approved):
LGTM :+1:
Thanks @icewind1991 !

emoral435 (approved):
For context, I am no expert with the backend for server, but this LGTM :)

PR #43982, opened 2024-03-04
fix(filecache): Move from array_merge to avoid memory exhaustion on large scans

solracsf:
* Resolves part of https://github.com/nextcloud/server/issues/43964#issuecomment-1976976837

## Summary

When merging large arrays, PHP may require a significant amount of memory, especially if the arrays contain a lot of data.
This can be the case in file scans of large filesystems.

To avoid using `array_merge` and reduce memory consumption, concatenate the arrays using a loop instead.

## Checklist

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -585,8 +585,13 @@ class Cache implements ICache {
 				return $cacheEntry->getPath();
 			}, $children);
 
-			$deletedIds = array_merge($deletedIds, $childIds);
-			$deletedPaths = array_merge($deletedPaths, $childPaths);
+			foreach ($childIds as $childId) {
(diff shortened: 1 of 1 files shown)


icewind1991:
do you know if `$array += $other` has the same memory usage as `array_merge`?

solracsf:
> do you know if `$array += $other` has the same memory usage as `array_merge`?

Never tested such approach. In general, `foreach` gives good memory control and very little performance penalties so it's widelly adopted (for very large arrays).

icewind1991:
I'm not actually sure if `+=` works here since it overwrites items with the same (implicit) keys

icewind1991 (approved):

ChristophWurst (approved):

PR #43992, opened 2024-03-05
fix: Avoid clear cache with prefix

juliusknorr:
Calling `clear` on Redis is a heavy operation and can result in slow queries there. For the reference cache we can avoid in cases we know that we have never set entries with a prefix.

One common case is file references where we always called clear on [file deletion, share adding/removal](https://github.com/nextcloud/server/blob/a304ba5b5a0f05a7ba59c047bc8f44272ba1613e/lib/private/Collaboration/Reference/File/FileReferenceEventListener.php#L59) even though no references to those files were used.

### DIFF
--- a/lib/private/Collaboration/Reference/ReferenceManager.php
+++ b/lib/private/Collaboration/Reference/ReferenceManager.php
@@ -117,6 +117,11 @@ class ReferenceManager implements IReferenceManager {
 
 		$reference = $matchedProvider->resolveReference($referenceId);
 		if ($reference) {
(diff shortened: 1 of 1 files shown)


icewind1991 (approved):

nickvergessen (approved):

nickvergessen:
> There were 17 failures:
> 
> 1) OCA\UpdateNotification\Tests\Controller\AdminControllerTest::testCreateCredentials

Because @susnux PR was merged without passing CI, so either needs a rebase after https://github.com/nextcloud/server/pull/44074 is in or another blind red CI merge

PR #43997, opened 2024-03-05
fix: Make URL pattern regexes match localhost and URLs with ports

mejo-:
E.g. the file reference widget should work with links to Nextcloud on localhost.

This adjusts the URL pattern regexes in server to be in sync with the ones in `@nextcloud/vue`. See nextcloud-libraries/nextcloud-vue#5176 for further details.

## Checklist


juliusknorr (approved):

max-nextcloud (approved):

lib/public/IURLGenerator.php
```diff
@@ -51,7 +51,7 @@ interface IURLGenerator {
 	 *
 	 * @since 25.0.0
 	 */
-	public const URL_REGEX_NO_MODIFIERS = '(\s|\n|^)(https?:\/\/)((?:[-A-Z0-9+_]+\.)+[-A-Z]+(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(\s|\n|$)';
+	public const URL_REGEX_NO_MODIFIERS = '(\s|\n|^)(https?:\/\/)([-A-Z0-9+_.]+(?::[0-9]+)?(?:\/[-A-Z0-9+&@#%?=~_|!:,.;()]*)*)(\s|\n|$)';
```
nickvergessen, line 54: Please document an `@since` with the change and intention
Also on `URL_REGEX` which is changed indirectly

nickvergessen:
Also make sure to document the change, so apps using the regex know it changed

PR #43999, opened 2024-03-05
fix(user_ldap): Catch DB Exceptions when updating group memberships

come-nc:
* Resolves: # 

## Summary

When inserting/deleting group membership, it may simply fail because another job or request is doing the same thing concurrently. In this case we must not crash but simply ignore the error. When it fails for unexpected reason, we must log it but not crash.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/LoginListener.php
+++ b/apps/user_ldap/lib/LoginListener.php
@@ -27,6 +27,7 @@ namespace OCA\User_LDAP;
 
 use OCA\User_LDAP\Db\GroupMembership;
 use OCA\User_LDAP\Db\GroupMembershipMapper;
+use OCP\DB\Exception;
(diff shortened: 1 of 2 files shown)


icewind1991:
it probably makes sense to catch the unique errors in the `GroupMembershipMapper`

come-nc:
> it probably makes sense to catch the unique errors in the `GroupMembershipMapper`

But no other mapper is doing so, the insert/delete methods are standard from parent class?

icewind1991:
> > it probably makes sense to catch the unique errors in the `GroupMembershipMapper`
> 
> But no other mapper is doing so, the insert/delete methods are standard from parent class?

Fair enough

icewind1991 (approved):

ArtificialOwl (approved):

PR #44000, opened 2024-03-05
cleanup: remove long depricated share code

### LINKED ISSUE #4415: Cleanup \OCP\Share
Now that the access list is converted as well we should make work of cleaning up the old sharing code properly.

- [ ] Get rid of the registerbackend stuff (we have the factory for that, or we should add functions to the manager to check this)
- [ ] Get rid of all calls to \OCP\Share::
  - mainly in test code
  - maybe to this in small steps for easier review
- [ ] Kill \OC\Share\

icewind1991:
* Resolves https://github.com/nextcloud/server/issues/4415

11 versions seems like a long enough deprecation period :see_no_evil:


come-nc (approved):
:broom:

Altahrim (approved):

PR #44004, opened 2024-03-05
perf: improve performance of resolving group shares

icewind1991:
Noticed that the resolving of shares is creating a fairly significant load on the DB. Instead of sending a massive `IN` statement, filtering just by the recipient is enough and should be a lot better on the DB.

### DIFF
--- a/lib/private/Share20/DefaultShareProvider.php
+++ b/lib/private/Share20/DefaultShareProvider.php
@@ -829,7 +829,7 @@ class DefaultShareProvider implements IShareProvider {
 
 		// If the recipient is set for a group share resolve to that user
 		if ($recipientId !== null && $share->getShareType() === IShare::TYPE_GROUP) {
-			$share = $this->resolveGroupShares([$share], $recipientId)[0];
+			$share = $this->resolveGroupShares([(int) $share->getId() => $share], $recipientId)[0];
 		}
 
 		return $share;
@@ -1006,7 +1006,8 @@ class DefaultShareProvider implements IShareProvider {
 					}
 
 					if ($this->isAccessibleResult($data)) {
-						$shares2[] = $this->createShare($data);
+						$share = $this->createShare($data);
+						$shares2[$share->getId()] = $share;
 					}
 				}
 				$cursor->closeCursor();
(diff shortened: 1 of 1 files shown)


skjnldsv (approved):

lib/private/Share20/DefaultShareProvider.php
```diff
@@ -1120,61 +1121,41 @@ private function createShare($data) {
...
-			}
+		while ($data = $stmt->fetch()) {
+			$shareMap[$data['parent']]->setPermissions((int)$data['permissions']);
+			$shareMap[$data['parent']]->setStatus((int)$data['accepted']);
+			$shareMap[$data['parent']]->setTarget($data['file_target']);
+			$shareMap[$data['parent']]->setParent($data['parent']);
 		}
 
-		return $result;
+		return array_values($shareMap);
```
blizzz, line 1158: you're missing the `closeCursor` now. Or is it not a thing anymore?

icewind1991: cursors are closed automatically when they go out of scope

lib/private/Share20/DefaultShareProvider.php
```diff
@@ -1120,61 +1121,41 @@ private function createShare($data) {
...
+	private function resolveGroupShares($shareMap, $userId) {
+		$qb = $this->dbConn->getQueryBuilder();
+		$query = $qb->select('*')
+			->from('share')
+			->where($qb->expr()->eq('share_with', $qb->createNamedParameter($userId)))
+			->andWhere($qb->expr()->eq('share_type', $qb->createNamedParameter(IShare::TYPE_USERGROUP)))
+			->andWhere($qb->expr()->orX(
+				$qb->expr()->eq('item_type', $qb->createNamedParameter('file')),
+				$qb->expr()->eq('item_type', $qb->createNamedParameter('folder'))
+			));
```
Altahrim, line 1139: We could use a  `IN` here

Altahrim (approved):

lib/private/Share20/DefaultShareProvider.php
Altahrim: Modified

skjnldsv (approved):

PR #44013, opened 2024-03-05
feat: add pronouns to account profile

### LINKED ISSUE #40017: Allow setting third person pronouns in user's profile
When referencing a person in a conversation in third person you might want to use third person pronouns instead of their name. Assuming pronouns based on someone's appearance, name, … can be hurtful to the person being referred, because they might not like those pronouns to be used for them.

A field in the user's profile for entering the pronouns they'd like other people to use. Having an explicit entry for pronouns in a user's profile means you don't have to assume, ask them or look in their profile's bio for pronouns which makes it way easier and more convenient to use their preferred pronouns.

There should be default options like she/her/…, he/him/… and they/them/… but you should also be able to enter custom ones like [fae/faer/faer/faers/faerself](https://pronouns.within.lgbt/fae/faer/faer/faers/faerself) with every grammatical version specified.

None.

**Additional context**
This is probably an issue to implement because language have different rules for third person pronouns. Some might not have 5 versions (Subject, Object, Dependent Possessive, Independent Possessive, Reflexive) like English.

Supporting neopronouns with the custom option might be controversial. Neopronouns never really caught on outside of maybe queer communities but this would at least support specifying a generic gender-neutral they/them/…

skjnldsv:
Fix https://github.com/nextcloud/server/issues/40017

| Desktop | Mobile |
|:---------:|:------:|
|||

image
image

- [x] Add free text input  to enter pronouns
- [x] Always show the pronouns next to the displayName


apps/settings/src/components/PersonalInfo/PronounsSection.vue
```diff
@@ -0,0 +1,51 @@
+<!--
+	- @copyright 2022 Christopher Ng <chrng8@gmail.com>
+	-
+	- @author Christopher Ng <chrng8@gmail.com>
```
susnux, line 4: Thats you? ;)

lib/private/Accounts/AccountManager.php
```diff
@@ -103,6 +103,7 @@ class AccountManager implements IAccountManager {
 		self::PROPERTY_ROLE => self::SCOPE_LOCAL,
 		self::PROPERTY_HEADLINE => self::SCOPE_LOCAL,
 		self::PROPERTY_BIOGRAPHY => self::SCOPE_LOCAL,
+		self::PROPERTY_PRONOUNS => self::SCOPE_FEDERATED,
```
Pytal, line 106: I believe if this property should be federated more backend adjustments need to be made in `apps/dav/lib/CardDAV/Converter.php` maybe others as well but not familiar here

Pytal (commented):
Tests in apps/provisioning_api/tests/Controller/UsersControllerTest.php and tests/lib/Accounts/AccountManagerTest.php should be updated as well

botsarenthuman:
Please make this optional, or opt-in

jancborchardt (approved):
Nice, and looks good design-wise! They will also show in the visibility settings, right?

Can you show a screenshot how it will look like on the profile? I assume subline like the other things, possibly divided by the middle dot we often use? `·`

skjnldsv:
> Can you show a screenshot how it will look like on the profile? I assume subline like the other things, possibly divided by the middle dot we often use? `·`

It's still a draft, but that was the plan! I'll update with screenshots soon :+1:

botsarenthuman:
The comment is not off-topic, this is a controversial feature, it should be optional for those who don't want this feature displayed to users. This is not meant to be "troll comment" as we've used Nextcloud for near a decade and love the product. But you must realize that certain topics are not universally agreed upon and so it would be nice to allow some freedom here.

jancborchardt:
> Please make this optional, or opt-in

It's optional to put in pronouns, soo …

skjnldsv:
Let's go!! 🚀 
I'll see how we can improve the user hover menu as a follow-up!

core/src/views/Profile.vue
```diff
@@ -426,6 +436,11 @@ $content-max-width: 640px;
 
 		&__content {
 			display: block;
+
+			 .avatar {
+				// Overlap avatar to top header
+				margin-top: -110px !important;
```
Pytal, line 442:
```suggestion
				  // Overlap avatar to top header
				  margin-top: -110px !important;
```

Pytal (approved):
Code looks good!

No idea about LDAP part 🙈

skjnldsv:
> Code looks good!
> 
> No idea about LDAP part 🙈

I'll double check with Arthur tomorrow :)

come-nc:
@skjnldsv This broke the tests but the CI was wrongly green.
Please do a follow-up PR to fix the related tests.

PR #44014, opened 2024-03-05
fix: Remove unused endpoint from base.php

come-nc:
It’s not called anymore, and app loading cannot break boot anymore

It was added in https://github.com/nextcloud/server/commit/1dbe240b0e5b0a0b32d0e0eca596bf4e510f980e but the frontend code calling it was removed long ago. Also it’s pretty hard for an app to break the files app because Throwables at boot are catched.
It’s still possible to crash /app/files by listening to sidebar loading and crashing for instance, and there are some message about it in the UI but it does not seem to disable. This should be fixed in a follow-up and use the standard disable app route.

## Checklist

### DIFF
--- a/lib/base.php
+++ b/lib/base.php
@@ -1018,21 +1018,6 @@ class OC {
 			}
 		}
 
-		// emergency app disabling
-		if ($requestPath === '/disableapp'
-			&& $request->getMethod() === 'POST'
-		) {
-			\OC_JSON::callCheck();
-			\OC_JSON::checkAdminUser();
-			$appIds = (array)$request->getParam('appid');
-			foreach ($appIds as $appId) {
-				$appId = \OC_App::cleanAppId($appId);
-				Server::get(\OCP\App\IAppManager::class)->disableApp($appId);
-			}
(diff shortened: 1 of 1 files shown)


susnux (approved):
I can not find any usage of this anymore - moreover I tested to break server core but any bootstraping issue is catched and logged -> faulty apps do not harm the core anymore.

ChristophWurst (approved):

nickvergessen:
> but the frontend code calling it was removed long ago.

It was called automatically when you enabled an app via the UI that afterwards broke the index.php loading. But yeah might not be needed as much anymore with the register/bootstrap now

susnux:
> It was called automatically when you enabled an app via the UI that afterwards broke the index.php loading

That was missing for couple of years, see #44016

nickvergessen:
> That was missing for couple of years, see https://github.com/nextcloud/server/pull/44016

Well that might not work in most cases, exactly why we had this URL hardcoded to the beginning of the base.php handling so it's done before anything else.

PR #44022, opened 2024-03-06
feat(settings): add JavaScript Source Maps support `.js.map` setup check

### LINKED ISSUE #20549: Add setup check for map files serving
> we should have a config check
> for serving map files
> because I'm sure tons of admins did not updated the apache/nginx configs

cc @rullzer

skjnldsv:
Fix https://github.com/nextcloud/server/issues/20549

### DIFF
--- a/apps/settings/lib/SetupChecks/JavaScriptSourceMaps.php
+++ b/apps/settings/lib/SetupChecks/JavaScriptSourceMaps.php
@@ -53,7 +53,7 @@ class JavaScriptSourceMaps implements ISetupCheck {
 	}
 
 	public function getName(): string {
-		return $this->l10n->t('JavaScript Source Maps support');
+		return $this->l10n->t('JavaScript source map support');
 	}
 
 	public function run(): SetupResult {
@@ -63,6 +63,6 @@ class JavaScriptSourceMaps implements ISetupCheck {
 			return SetupResult::success();
 		}
 
-		return SetupResult::error($this->l10n->t('Your webserver is not set up to serve `.js.map` files. Without these files, JavaScript Source Maps won\'t function properly, making it more challenging to troubleshoot and debug any issues that may arise.'));
+		return SetupResult::warning($this->l10n->t('Your webserver is not set up to serve `.js.map` files. Without these files, JavaScript Source Maps won\'t function properly, making it more challenging to troubleshoot and debug any issues that may arise.'));
 	}
 }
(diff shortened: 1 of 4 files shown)


apps/settings/lib/SetupChecks/JavaScriptSourceMaps.php
```diff
@@ -0,0 +1,68 @@
...
+		protected LoggerInterface $logger,
+	) {
+	}
+
+	public function getCategory(): string {
+		return 'network';
+	}
+
+	public function getName(): string {
+		return $this->l10n->t('JavaScript Source Maps support');
```
susnux, line 56:
```suggestion
		return $this->l10n->t('JavaScript source map support');
```

from other wording?

apps/settings/lib/SetupChecks/JavaScriptSourceMaps.php
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	public function run(): SetupResult {
+		$testFile = $this->urlGenerator->linkTo('settings', 'js/map-test.js.map');
+
+		foreach ($this->runHEAD($testFile) as $response) {
+			return SetupResult::success();
```
susnux, line 63: Do we need some specific content type?

apps/settings/lib/SetupChecks/JavaScriptSourceMaps.php
```diff
@@ -0,0 +1,68 @@
...
+		}
+
+		return SetupResult::error($this->l10n->t('Your webserver is not set up to serve `.js.map` files. Without these files, JavaScript Source Maps won\'t function properly, making it more challenging to troubleshoot and debug any issues that may arise.'));
```
come-nc, line 66: Maybe warning if it’s only for debugging?

apps/settings/lib/SetupChecks/JavaScriptSourceMaps.php
skjnldsv: As long as it's serving it, I think it's fine

apps/settings/lib/SetupChecks/JavaScriptSourceMaps.php
skjnldsv, line 66:
```suggestion
		return SetupResult::warn($this->l10n->t('Your webserver is not set up to serve `.js.map` files. Without these files, JavaScript Source Maps won\'t function properly, making it more challenging to troubleshoot and debug any issues that may arise.'));
```

come-nc (approved):

susnux (approved):

PR #44023, opened 2024-03-06
chore(IURLGenerater): Document changes to `URL_REGEX_NO_MODIFIERS`

mejo-:
Follow-up for #43997

### DIFF
--- a/lib/public/IURLGenerator.php
+++ b/lib/public/IURLGenerator.php
@@ -41,6 +41,7 @@ interface IURLGenerator {
 	 * This is a copy of the frontend regex in core/src/OCP/comments.js, make sure to adjust both when changing
 	 *
 	 * @since 25.0.0
+	 * @since 29.0.0 changed to match localhost and hostnames with ports
 	 */
 	public const URL_REGEX = '/' . self::URL_REGEX_NO_MODIFIERS . '/mi';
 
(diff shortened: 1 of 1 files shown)


juliusknorr (approved):

lib/public/IURLGenerator.php
```diff
@@ -49,7 +49,7 @@ interface IURLGenerator {
 	 *
 	 * This is a copy of the frontend regex in core/src/OCP/comments.js, make sure to adjust both when changing
 	 *
-	 * @since 25.0.0
+	 * @since 25.0.0 changed to match localhost and hostnames with ports in Nextcloud 29
```
nickvergessen, line 52:
```suggestion
	 * @since 25.0.0
	 * @since 29.0.0 changed to match localhost and hostnames with ports
```

And also on `URL_REGEX` please :)

nickvergessen (changes requested):

lib/public/IURLGenerator.php
mejo-: Alright, added now :blush:

lib/public/IURLGenerator.php
```diff
@@ -40,7 +40,7 @@ interface IURLGenerator {
...
```
nickvergessen, line 43:
```suggestion
	 * @since 25.0.0
	 * @since 29.0.0 changed to match localhost and hostnames with ports
```

lib/public/IURLGenerator.php
nickvergessen, line 52:
```suggestion
	 * @since 25.0.0
	 * @since 29.0.0 changed to match localhost and hostnames with ports
```

lib/public/IURLGenerator.php
mejo-: Sorry, I didn't get what you meant before. In other examples from the past, the change was documented in the same `@since` annotation and I was not sure whether it's good style to add a second `@since` annotation. Changed now.

nickvergessen (approved):

Issue #8505, opened 2018-02-22
Consolidate app management code


MorrisJobke:
Currently the code for managing app (install, update, remove, upgradability, fetching info, ...) is distributed across 3 classes:

* `OC_App`
*  `OC\App\AppManager`
* `OC\Installer`

We should consolidate this into a single class to not spread code and functionality as well as reduce the risk to re-implement existing code, because it is no obvious where to find which code.

MorrisJobke:
cc @nickvergessen @rullzer @blizzz FYI ;)

nickvergessen:
Sounds good :+1:

MorrisJobke:
One little step is already done in #5833

MorrisJobke:
From https://github.com/nextcloud/server/pull/8506#issuecomment-374896672

> Learned some funny things about our terminology the hard way:
> 
> * isInstalled, getInstalledApps, isEnabledForUser
> * first and last one check if the app is enabled
> * the other one checks if the app code is present in any of the apps folders 🙈
> 
> I added a little PHPDoc comment and we should look into unify this stuff.

MorrisJobke:
See also #9324

szaimen:
I suppose this issue is still valid? If not, please close this issue!

susnux:
@come-nc already worked on OC_App vs AppManager. So OC_App can be removed soon.

PR #44025, opened 2024-03-06
Migrate away from OC_App and toward IAppManager.

come-nc:
## Summary

Migrate away from OC_App and toward IAppManager.

## Checklist

### DIFF
--- a/core/Command/App/GetPath.php
+++ b/core/Command/App/GetPath.php
@@ -80,7 +80,7 @@ class GetPath extends Base {
 	 */
 	public function completeArgumentValues($argumentName, CompletionContext $context): array {
 		if ($argumentName === 'app') {
-			return $this->appManager->getInstalledApps();
+			return \OC_App::getAllApps();
 		}
 		return [];
 	}
--- a/core/Command/App/GetPath.php
+++ b/core/Command/App/GetPath.php
@@ -1,4 +1,7 @@
 <?php
+
+declare(strict_types=1);
+
 /**
  * @copyright Copyright (c) 2016, ownCloud, Inc.
  *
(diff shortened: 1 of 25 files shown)


icewind1991 (approved):

ChristophWurst (approved):
To avoid some kaboom we might want to merge this after the 29 branch-off

skjnldsv (approved):

come-nc:
Rebased and fixed conflicts

core/Command/App/GetPath.php
```diff
@@ -69,7 +80,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 	 */
 	public function completeArgumentValues($argumentName, CompletionContext $context): array {
 		if ($argumentName === 'app') {
-			return \OC_App::getAllApps();
+			return $this->appManager->getInstalledApps();
```
nickvergessen, line 83: This is not an equal replacement.
The old one was iterating over the apps directories. This here is looping over `enabled` appconfig values.
So the command no longer autocompletes a path for an app that was cloned but not yet ever enabled.

core/Command/App/Update.php
```diff
@@ -88,7 +91,7 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 				return 1;
 			}
 		} elseif ($input->getOption('all') || $input->getOption('showonly')) {
-			$apps = \OC_App::getAllApps();
+			$apps = $this->manager->getInstalledApps();
```
nickvergessen, line 94: Same here:
This is not an equal replacement.
The old one was iterating over the apps directories. This here is looping over `enabled` appconfig values.
So the command no longer autocompletes a path for an app that was cloned but not yet ever enabled.

But I guess it's somewhat okay in this case.

nickvergessen: > Does the command work on a disabled app?

It's updating several apps on our servers we don't use at the moment.

come-nc:
@nickvergessen Reverted both calls. We’ll need later to add a replacement for getAllApps.

But the `maintenance:install` command segfaults since today rebase. I can reproduce locally but have no idea what happens.
I tried reverting the IURLGenerator injection in the appmanager but it does not change anything.

joshtrichards:
Related to open issue: #8505 :)

Issue #29901, opened 2021-11-25
Config setting `config_is_read_only` should prevent Nextcloud from writing to `config.php`


mejo-:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

### Description

The config setting `config_is_read_only` currently only prevents Nextcloud from checking for writable config in server health checks.

That's also how the setting is described at https://github.com/nextcloud/documentation/blob/6efc785734a23786bff4c6340a03fa2acbf12f42/admin_manual/configuration_server/config_sample_php_parameters.rst#nextcloud-verifications.

Still, users/admins expect the setting to prevent Nextcloud from writing to `config.php`. That would be helpful, e.g. to prevent admin users from accidently changing server configuration.

### Steps to reproduce
1. Set `'config_is_read_only' => true` in `config.php`
2. Make changes to Nextcloud config via admin interface or `occ`
3. Nextcloud writes to `config.php`

Dreamsorcerer:
The solution to this (but, not the problem description) directly contradicts the goals of #11075.
Before, it was possible to get through an upgrade by manually making changes to the config file, now it aborts immediately even when there are no changes to be made to the config file.

Old behaviour:
Nextcloud would call `set('theme', '')` etc. Because the existing config has the same values, it would not attempt to write anything to the config file. If the config file would get changed, then it gives an error if the read-only option is set.

New behaviour:
The same call is made and immediately errors despite the fact that the config file will not be changed.

PR #44039, opened 2024-03-06
fix: regression with updating read-only config

Dreamsorcerer:
Revert some changes for #29901.

Old behaviour before those changes (and again, after this PR):
Nextcloud calls set('theme', '') etc. Because the existing config has the same values, it does not attempt to write anything to the config file. If the config file would get changed, then it gives an error if the read-only option is set.

Regression behaviour:
The same call is made and immediately errors despite the fact that the config file will not be changed.

The set()/delete() methods are clearly written in a way that they skip writing changes if the value is already the same, but the previous changes make a check before those checks, thus erroring when there are no changes to write. The same PR also adds this check to writeData(), which is called by both of these methods, making them redundant.

### DIFF
--- a/lib/private/Config.php
+++ b/lib/private/Config.php
@@ -153,8 +153,6 @@ class Config {
 	 * @throws HintException
 	 */
 	protected function set($key, $value) {
-		$this->checkReadOnly();
-
 		if (!isset($this->cache[$key]) || $this->cache[$key] !== $value) {
 			// Add change
 			$this->cache[$key] = $value;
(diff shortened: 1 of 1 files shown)


mejo- (approved):
Looks sensible to me, given that `checkReadOnly()` is called again in `writeData()` anyway.

Dreamsorcerer:
Updated the commit message, I don't think the other failures are related to my change (maybe my fork was old though, so I've rebased).

Dreamsorcerer:
Any chance of getting this in for 29?

come-nc (approved):

Dreamsorcerer:
Looks like something got stuck on the CI, so this still hasn't merged yet..

mejo-:
I guess the cypress tests fail due to the parent branch living in a forked repository.

Dreamsorcerer:
> I guess the cypress tests fail due to the parent branch living in a forked repository.

Looks like my previous PRs were merged with these failing, so I'm guessing that making them required checks is probably not the right option while it doesn't work in forks.

PR #44044, opened 2024-03-06
Add interface to allow "more direct" access to the filecache without direct db

icewind1991:
Adds and interface for the filecache that is lower level than the "normal" filesystem apis to allow for some more optimized retrieval of filecache data without requiring direct database access.

This will be needed when we start restricting direct database access to the filecache for sharding.

Pr comes with a few cases of replacing direct-db with the new interface.

extracted from #43576


lib/private/Files/Cache/CacheAccess.php
```diff
@@ -0,0 +1,100 @@
...
+		);
+	}
+
+	public function getByFileIdInStorage(int $fileId, int $storageId): ?CacheEntry {
+		$query = $this->getQuery()->selectFileCache();
+		$query->andWhere($query->expr()->eq('filecache.fileid', $query->createNamedParameter($fileId, IQueryBuilder::PARAM_INT)));
+		$query->andWhere($query->expr()->eq('filecache.storage', $query->createNamedParameter($storageId, IQueryBuilder::PARAM_INT)));
+
+		$row = $query->executeQuery()->fetch();
+		return $row ? Cache::cacheEntryFromData($row, $this->mimeTypeLoader) : null;
```
artonge, line 41: Could this be simplified with the following:
```suggestion
    return $this->getByFileIdsInStorage([$fileId], $storageId);
```

Or is there an optimization in having a specific method for only one id?

lib/private/Files/Cache/CacheAccess.php
artonge, line 1: Is the naming appropriate?
Should we use `FileCacheAccess` to reflect the table name?
Or `FilesAccess` to reflect the purpose?

apps/federatedfilesharing/lib/FederatedShareProvider.php
```diff
@@ -605,12 +605,7 @@ public function getSharesInFolder($userId, Folder $node, $reshares, $shallow = t
 
 		$qb->innerJoin('s', 'filecache', 'f', $qb->expr()->eq('s.file_source', 'f.fileid'));
 
-		$qb->andWhere($qb->expr()->eq('f.storage', $qb->createNamedParameter($node->getMountPoint()->getNumericStorageId(), IQueryBuilder::PARAM_INT)));
-		if ($shallow) {
-			$qb->andWhere($qb->expr()->eq('f.parent', $qb->createNamedParameter($node->getId(), IQueryBuilder::PARAM_INT)));
-		} else {
-			$qb->andWhere($qb->expr()->like('f.path', $qb->createNamedParameter($this->dbConnection->escapeLikeParameter($node->getInternalPath()) . '/%')));
-		}
+		$qb->andWhere($qb->expr()->eq('f.parent', $qb->createNamedParameter($node->getId())));
```
come-nc, line 608: What is this change about?

apps/files_sharing/lib/Updater.php
```diff
@@ -58,20 +60,40 @@ private static function moveShareInOrOutOfShare($path): void {
...
 		if ($src instanceof Folder) {
-			$subShares = $shareManager->getSharesInFolder($userFolder->getOwner()->getUID(), $src, false, false);
+			$cacheAccess = Server::get(CacheAccess::class);
+
+			$sourceStorageId = $src->getStorage()->getCache()->getNumericStorageId();
+			$sourceInternalPath = $src->getInternalPath();
+			$subShares = array_merge(
+				$shareManager->getSharesBy($user->getUID(), IShare::TYPE_USER),
+				$shareManager->getSharesBy($user->getUID(), IShare::TYPE_GROUP),
+				$shareManager->getSharesBy($user->getUID(), IShare::TYPE_ROOM),
```
come-nc, line 85: Where did the `$src` parameter go?

lib/private/Files/Cache/CacheAccess.php
icewind1991: There is an optimization for the sharding-aware version of this implementation, but for this version it can indeed be simplified.

lib/private/Files/Cache/CacheAccess.php
```diff
@@ -0,0 +1,100 @@
+<?php
+
```
come-nc, line 2: strict typing missing

lib/private/Files/Config/UserMountCache.php
```diff
@@ -449,26 +428,26 @@ public function getUsedSpaceForUsers(array $users) {
...
-				))
 			->where($builder->expr()->eq('m.mount_point', $mountPoint))
 			->andWhere($builder->expr()->in('m.user_id', $builder->createNamedParameter($userIds, IQueryBuilder::PARAM_STR_ARRAY)));
 
 		$result = $query->execute();
 
 		$results = [];
 		while ($row = $result->fetch()) {
-			$results[$row['user_id']] = $row['size'];
+			$results[$row['user_id']] = $this->getSizeForHomeStorage($row['storage_id']);
```
come-nc, line 440: This means one query per row instead of getting all sizes in the first query, no?

lib/public/Files/Cache/ICacheAccess.php
```diff
@@ -0,0 +1,76 @@
+<?php
+
```
come-nc, line 2: strict typing

icewind1991:
Let me split out the interface and changes to remove filecache joins to simplify reviewing

PR #44049, opened 2024-03-06
feat: show the id of last author in versions metadata

emoral435:
* For #44109
## Summary
* The oc_files_version table is what stores the version information of each file. Each version can be categorized by its `file_id` and `timestamp` columns, but there is also a `metadata` column which stores JSON directly.
* Everytime the file node is overwritten, we should have an event listener that gets triggered and updates that meta data column with:
```js
{
  "author": "versionAuthorID",
}
```
* This PR also allows us to also potentially store messages that go along with a file version change, similar to a git commit in future releases! I tried to make this as flexible as possible, and also to make a PR as small as possible, but I believe these commits have to be bundled together to allow for a better sense while reading this commit 😋

## Screenshots

https://github.com/nextcloud/server/assets/110193237/88fd05b2-c7f9-4eab-ab46-4829fa10f2c7

image

## Checklist


apps/files_versions/lib/Listener/MetadetaFileEvents.php
```diff
@@ -0,0 +1,58 @@
...
+use OC\Files\Node\Folder;
+use OCA\Files_Versions\Versions\IStoreMetadataBackend;
+use OCA\Files_Versions\Versions\IVersionManager;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Files\Events\Node\NodeWrittenEvent;
+use OCP\Files\Node;
+use OCP\IUserSession;
+
+class MetadataFileEvents implements IEventListener {
```
susnux, line 34:
```suggestion
/** @template-implements IEventListener<NodeWrittenEvent> */
class MetadataFileEvents implements IEventListener {
```

For the psalm error

apps/files_versions/lib/Versions/IStoreMetadataBackend.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+interface IStoreMetadataBackend {
+	/**
+	 * For now, stores the version owner in the metadata column.
+	 *
+	 * @param File $file the file that triggered the Metadata event listener
+	 * @param IUserSession $user the user that will now have their data stored for this version
+	 * @since 29.0.0
+	 */
+	public function setMetadata(File $file, IUserSession $userSession): void;
```
susnux, line 42: Maybe make it more generic by using `Node` instead. Would also resolve the psalm warning in the event listener?

apps/files_versions/lib/Versions/IStoreMetadataBackend.php
```diff
@@ -0,0 +1,43 @@
...
+
+use OCP\Files\File;
+use OCP\IUserSession;
+
+/**
+ * This interface edits the metadata column of a node.
+ * Each column of the metadata has a key => value mapping.
+ * @since 29.0.0
+ */
+interface IStoreMetadataBackend {
```
susnux, line 34: From all other classes I would expect a different name.
E.g. the others are called ` INameableVersionBackend, IDeletableVersionBackend, INeedSyncVersionBackend ` so I would expect `IMetadataVersionBackend` or `IVersionsMetadataBackend` or similar

susnux (commented):
Nice, I like that you are using proper doc comments :)

apps/files_versions/lib/Listener/MetadetaFileEvents.php
emoral435: Thank you for the tip :)

apps/files_versions/lib/Versions/IStoreMetadataBackend.php
emoral435: Agreed. Anything that extends possible future additions to this feature (like eventually adding the `git commit body` feature) is preferred, so yes !

apps/files_versions/lib/Listener/MetadataFileEvents.php
```diff
@@ -0,0 +1,67 @@
...
+	 * @param Node $node the node that is currently being written
+	 */
+	public function post_write_hook(Node $node): void {
+		// Do not handle folders or users that we cannot get metadata from
+		if ($node instanceof Folder || is_null($this->userSession->getUser())) {
+			return;
+		}
+		// check if our version manager supports setting the metadata
+		if ($this->versionManager instanceof IMetadataVersionBackend) {
+			$author = $this->userSession->getUser()->getDisplayName();
```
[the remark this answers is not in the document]
emoral435: we check if the user is null beforehand.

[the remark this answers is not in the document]
susnux: Then add the cast manually (`/** @var IUser $author`)

apps/files_versions/lib/Sabre/VersionFile.php
```diff
@@ -109,6 +110,13 @@ public function setLabel($label): bool {
 		}
 	}
 
+	public function getMetadataAuthor(): string {
```
artonge, line 113: What do you think about making it generic here too?
```suggestion
	public function getMetadata($key string): string {
```

apps/files_versions/lib/Versions/IMetadataVersion.php
```diff
@@ -0,0 +1,38 @@
...
+ * @since 29.0.0
+ */
+interface IMetadataVersion {
+	/**
+	 * @abstract retrieves the metadata value from our $key param
+	 *
+	 * @param string $key the key for the json value of the metadata column
+	 * @since 29.0.0
+	 */
+	public function getMetadataValue(string $key): string;
```
artonge, line 37: Could you add deprecation notices to `INameableVersion`. We can port the naming logic in a follow up PR :).

apps/files_versions/lib/Versions/Version.php
```diff
@@ -121,4 +122,11 @@ public function getVersionPath(): string {
 	public function getUser(): IUser {
 		return $this->user;
 	}
+
+	public function getMetadataValue(string $key): string {
+		if ($this->backend instanceof IMetadataVersionBackend && $this->sourceFileInfo instanceof Node) {
+			return $this->backend->getMetadataValue($this->sourceFileInfo, "author") ?? '';
+		}
+		return '';
```
artonge, line 130: What about returning `null`?

apps/files_versions/lib/AppInfo/Application.php
```diff
@@ -113,6 +114,7 @@ public function register(IRegistrationContext $context): void {
 		$context->registerEventListener(NodeTouchedEvent::class, FileEventsListener::class);
 		$context->registerEventListener(BeforeNodeWrittenEvent::class, FileEventsListener::class);
 		$context->registerEventListener(NodeWrittenEvent::class, FileEventsListener::class);
+		$context->registerEventListener(NodeWrittenEvent::class, MetadataFileEvents::class);
```
artonge, line 117: I would move the line after all the `FileEventsListener` registrations. Just to ensure that `MetadataFileEvents` is correctly executed after `FileEventsListener`.

apps/files_versions/lib/Sabre/VersionFile.php
```diff
@@ -109,6 +110,13 @@ public function setLabel($label): bool {
...
+		if ($this->version instanceof IMetadataVersion) {
+			return $this->version->getMetadataValue("author");
+		}
+		return '';
```
artonge, line 117: What about returning `null`?

artonge (commented):
A few comments, but it looks really good ! :sparkles: :rocket:

artonge (commented):
Also, will actually displaying the author be part of this PR or planned for a follow-up?
Did you have mock-ups from @nextcloud/designers?

emoral435:
> Also, will actually displaying the author be part of this PR or planned for a follow-up?
Did you have mock-ups from @nextcloud/designers?

Yes, it is part of another PR within the chained PR's :) The reason I stored the userID instead of just their name was because Jan recommended I put the authors display-icon when I asked for a mock-up:

> Can we use the userbubble here? Then the avatar is also shown, and seeing faces is always good for recognizing

So this makes it easier to put their ID *and* their name together by just storing their identifier :)

apps/files_versions/lib/AppInfo/Application.php
emoral435: Nice catch! 👍

apps/files_versions/lib/Versions/Version.php
emoral435: agreed - this would make it more consistent with the implementation we already have for getting the label, and makes it more clear when we couldn't fetch the metadata value 👍

emoral435:
Deprecated INameableVersion, returned null in more instances to indicate unsuccessful retrieval from the database, and moved the event listener to the bottom so that it gets executed last :)

artonge:
> EDIT: Also, before merging, should I commit the changes from `./build/autoloaderchecker.sh`, similar to how the frontend does `npm run build` to compile their assets? It is needed in this case for everything to work 🤞

You can commit the changes in files named `autoload_classmap.php` and `autoload_static.php`.

apps/files_versions/lib/Db/VersionEntity.php
```diff
@@ -78,4 +78,23 @@ public function setLabel(string $label): void {
...
+	 * @param string $key key associated with the value
+	 */
+	public function getMetadataValue(string $key): ?string {
+		return $this->metadata[$key] ?? null;
+	}
+
+	/**
+	 * @abstract sets a key value pair in the metadata column
+	 * @param string $key key associated with the value
+	 * @param string $value value associated with the key
```
susnux, line 94: Here it is typed as `string` but in the function it is empty.
If it should only allow string, then also type the `$value` below?

apps/files_versions/lib/Listener/MetadataFileEvents.php
```diff
@@ -0,0 +1,68 @@
...
+	public function post_write_hook(Node $node): void {
+		$user = $this->userSession->getUser();
+		// Do not handle folders or users that we cannot get metadata from
+		if ($node instanceof Folder || is_null($user)) {
+			return;
+		}
+		// check if our version manager supports setting the metadata
+		if ($this->versionManager instanceof IMetadataVersionBackend) {
+			$author = $user->getUID();
+			$this->versionManager->setMetadataValue($node, "author", $author);
```
susnux, line 65: We use single quotes everywhere else if not a template string:

```suggestion
			$this->versionManager->setMetadataValue($node, 'author', $author);
```

apps/files_versions/lib/Sabre/Plugin.php
```diff
@@ -44,6 +44,8 @@ class Plugin extends ServerPlugin {
 
 	public const VERSION_LABEL = '{http://nextcloud.org/ns}version-label';
 
+	public const VERSION_AUTHOR = '{http://nextcloud.org/ns}version-author'; // dav property for author
```
susnux, line 47: Please do not forget to add this to the docs here: https://github.com/nextcloud/documentation/blob/master/developer_manual/client_apis/WebDAV/basic.rst

(It seems the version-label is also missing)

apps/files_versions/lib/Versions/IMetadataVersion.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+namespace OCA\Files_Versions\Versions;
+
+/**
+ * This interface allows for just direct accessing of the metadata column JSON
+ * @since 29.0.0
+ */
+interface IMetadataVersion {
+	/**
+	 * @abstract retrieves the metadata value from our $key param
```
susnux, line 32: You do not need the `@abstract` it will use the first text anyways, but it also does not hurt :)

susnux:
Really nice work :rocket: Just a few nitpicks and comments :)

apps/files_versions/lib/Sabre/Plugin.php
artonge: pending doc label added

apps/files_versions/lib/Db/VersionEntity.php
emoral435: nice catch, thank you :)

apps/files_versions/lib/Versions/IMetadataVersion.php
emoral435: AFAIK, `@abstract` is not really used anywhere else, so for consistency, I will take it off :)

emoral435:
I fixed the inconsistent typing of variables within functions, bad use of double-quotes, and licensing :) Just need to do documentation, but as I am at university currently, I will save that for when I get back home! 🚀 will edit this comment to include the updated documentations PR

artonge (approved):

susnux (approved):
thank you!

PR #44062, opened 2024-03-07
fix(files): properly reset preview component

skjnldsv:
Fix after https://github.com/nextcloud/server/pull/41416

The suggested fix does not match the implemented logic. It's more fitting to not reset the preview state if the source change matches the same node.


susnux (approved):

szaimen (approved):
🐘

emoral435 (approved):

PR #23792, opened 2020-10-30
Add a function to get the unread count for multiple objects in one go

nickvergessen:
Signed-off-by: Joas Schilling <coding@schilljs.com>

PR #44064, opened 2024-03-07
refactor: depricate getNumberOfUnreadCommentsForFolder and redo it's …

icewind1991:
…implementation based on getNumberOfUnreadCommentsForObjects

This seems to be unused since https://github.com/nextcloud/server/pull/23792

### DIFF
--- a/lib/private/Comments/Manager.php
+++ b/lib/private/Comments/Manager.php
@@ -29,6 +29,7 @@
 namespace OC\Comments;
 
 use Doctrine\DBAL\Exception\DriverException;
+use OCA\DAV\Connector\Sabre\File;
 use OCP\AppFramework\Utility\ITimeFactory;
 use OCP\Comments\CommentsEvent;
 use OCP\Comments\IComment;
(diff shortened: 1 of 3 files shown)


lib/private/Comments/Manager.php
```diff
@@ -820,54 +815,25 @@ public function getLastCommentDateByActor(
 	/**
 	 * Get the number of unread comments for all files in a folder
 	 *
+	 * This is unused since 8bd39fccf411195839f2dadee085fad18ec52c23
```
artonge, line 818: Why not delete the method if it is unused?

nickvergessen: because it's public API (marked deprecated in lib/public/ and therefore scheduled for deletion in 38/39)

icewind1991: Because it's exposed through OCP and thus subject to the deprecation period

artonge (approved):

lib/private/Comments/Manager.php
artonge: Right, I only checked the namespace of the file.

ArtificialOwl (approved):

PR #44066, opened 2024-03-07
files_versions: add missing null check

pulsejet:
The fact that this can be null could indicate a deeper problem.

See https://github.com/pulsejet/memories/issues/1053

### DIFF
--- a/apps/files_versions/lib/Listener/FileEventsListener.php
+++ b/apps/files_versions/lib/Listener/FileEventsListener.php
@@ -194,6 +194,11 @@ class FileEventsListener implements IEventListener {
 		}
 
 		$path = $this->getPathForNode($node);
+
+		if ($path === null) {
+			return;
+		}
+
 		$result = Storage::store($path);
 
 		// Store the result of the version creation so it can be used in post_write_hook.


icewind1991 (approved):

susnux (approved):

susnux:
please rephrase your commit to match conventional commits, like:
> fix(files_versions): Add missing null check

emoral435 (approved):
Nice change :) Currently looking at the files_version app right now, so if anything that can be the related to this pops up, I will come back here. but for now, this is a good change

pulsejet:
> please rephrase your commit to match conventional commits, like:
> 
> > fix(files_versions): Add missing null check

Done.

> The fact that this can be null could indicate a deeper problem.

I just want to mention again that this might just be a band aid

artonge:
Band-aid indeed, reverting.
@pulsejet Please report whether https://github.com/nextcloud/server/pull/42891 could have solved the issue: https://github.com/nextcloud/server/pull/42891
If it does not, then provide more information to improve the path resolution logic.

pulsejet:
> @pulsejet Please report whether #42891 could have solved the issue.


@pulsejet Please report whether #42891 could have solved the issue.

I've no idea since I can't reproduce the bug anyway. But that seems like a reasonable fix.

dinyar:
For the record: I have the same issue and could work around it by moving the image files from external storage to the "internal" one using the Nextcloud UI. I plan on moving them back to external storage again via the command line, but of course this isn't exactly practical.. (I'm on version 28.0.3 of Nexcloud and updated Memories yesterday.)

PR #44067, opened 2024-03-07
Migrate header check to setupcheck API

come-nc:
See #41364 

## Summary

Security related header check migrated to new API and done from backend now.
TODO:
- [x] Migrate the tests
- [x] Merge the HSTS check in there

## Checklist


apps/settings/lib/SetupChecks/OcxProviders.php
```diff
@@ -68,7 +68,7 @@ public function run(): SetupResult {
 		];
 
 		foreach ($providers as $provider) {
-			foreach ($this->runHEAD($this->urlGenerator->getWebroot() . $provider) as $response) {
```
susnux, line 71: If that is removed, then drop also `runHEAD` function?

come-nc: Still used in other checks I think but yeah we could move all of them.

apps/settings/lib/SetupChecks/SecurityHeaders.php
```diff
@@ -0,0 +1,160 @@
...
+						'link' => 'https://www.w3.org/TR/referrer-policy/',
+					];
+				}
+
+				$transportSecurityValidity = $response->getHeader('Strict-Transport-Security');
+				$minimumSeconds = 15552000;
+				if (preg_match('/^max-age=(\d+)(;.*)?$/', $transportSecurityValidity, $m)) {
+					$transportSecurityValidity = (int)$m[1];
+					if ($transportSecurityValidity < $minimumSeconds) {
+						$msg .= $this->l10n->t('- The `Strict-Transport-Security` HTTP header is not set to at least `%d` seconds (current value: `%d`). For enhanced security, it is recommended to enable HSTS.', [$minimumSeconds, $transportSecurityValidity])."\n";
```
Altahrim, line 122:
```suggestion
						$msg .= $this->l10n->t('- The `Strict-Transport-Security` HTTP header is not set to at least `%d` seconds (current value: `%d`). For enhanced security, it is recommended to use a long HSTS policy.', [$minimumSeconds, $transportSecurityValidity])."\n";
```

Altahrim (approved):
If domain contains only one dot (`xxxxx.yyy`), we could also advise to use HSTS `includeSubDomains` and `preload` (https://hstspreload.org/)

susnux (approved):

come-nc:
> If domain contains only one dot (`xxxxx.yyy`), we could also advise to use HSTS `includeSubDomains` and `preload` (https://hstspreload.org/)

I do not have any knowledge about this, I’d say that would be a followup, let’s migrate the feature first.

PR #44012, opened 2024-03-05
feat(files): add Viewer Files ressource handler

skjnldsv:
- Fix https://github.com/nextcloud/server/issues/42916
- Fix https://github.com/nextcloud/server/issues/42917
- [x] Needs https://github.com/nextcloud-libraries/nextcloud-dialogs/pull/1261

https://github.com/nextcloud/server/assets/14975046/1cf125e6-29cf-4ac0-af9e-318eadda3967

PR #44076, opened 2024-03-08
Use viewer to open files still on the simple widget

juliusknorr:
Follow up to https://github.com/nextcloud/server/pull/44012

This brings back the behaviour to open the viewer on non-interactive widget rendering.

Fixes recent failures on https://github.com/nextcloud/text/pull/5447 for the link bubble

## Checklist


mejo- (approved):

skjnldsv (changes requested):
It was done on purpose, you cannot open viewer on text as text us already using viewer. It will close the text app.

Can you explain the context? :)

skjnldsv:
> I don't quite remember how the behavior was if this is triggered trough text in viewer still

It wasn't. The Vue event was missing a stop and therefore the behaviour was unstable and still opening the file as if the link was clicked.

I suggest the following, until we support multiple viewer, on top of checking if Viewer supports the mime, please also check if a file is already opened in it. 
Something like `!OCA.Viewer.file` should work 😊

juliushaertl:
Thanks, adjusted

emoral435 (approved):

PR #44083, opened 2024-03-08
Revert "files_versions: add missing null check"

artonge:
Reverts nextcloud/server#44066


susnux (approved):
What is the problem? I mean with or without both is ok :sweat_smile:

Pytal:
> What is the problem? I mean with or without both is ok 😅

Probably better to log the error than silently suppress to catch potential bugs @artonge?

emoral435 (approved):
This is an interesting bug - we get `path` from either the currently logged in user, or the owner of the version. If not either of them, then we return null... but who else can edit a file? It's interesting, but after seeing the commits body message, I agree that having something in check without giving it a bandaid fix is good :)

PR #44096, opened 2024-03-08
fix app management scrolling for Safari/WebKit browsers

### LINKED ISSUE #42249: [Bug]: Apps UI not working in WebKit browsers (Safari, iOS/iPadOS, GNOME/Epiphany)

### Bug description

When selection any app in the "your apps" "active apps" "disabled apps" section always the same app will be displayed. E.g. selection "Activity" -> "Versions" is displayed. Scrolling doesn't work as well.
Screenshot 2023-12-13 at 21 53 33

### Steps to reproduce

1. Open Apps
2. Select an app to manipulate
3. in my case "Versions" is displayed

### Expected behavior

The selected app should be displayed to enable/disable

### Installation method

None

### Nextcloud Server version

28

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database engine version

MariaDB

Encryption is Disabled

- [x] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

linuzer:
fix #42249 scrolling for Safari-Browser.

## Checklist


apps/settings/src/components/AppList/AppItem.vue
```diff
@@ -210,7 +210,6 @@ export default {
 
 .app-name--link::after {
   content: '';
-  position: absolute;
   top: 0;
   left: 0;
   right: 0;
```
ShGKme, line 215: Removing `position: absolute` should be together with `inset` values - `top`, `left` etc. They doesn't work for static position. And without absolution position the whole `::after` doesn't work, it has no size.

But this pseudo-element `::after` was added intentionally. It allowed to click on the table row like it is a link. Without it, you need to click exactly on the text to open sidebar.

If it is removed, we need some alternative here. Or find the reason, why having it breaks Safari. I'll have a look in 2-3 hours on Safari.

image

ShGKme (changes requested):

apps/settings/src/components/AppList/AppItem.vue
ShGKme: This is a known Safari behavior. According to the standard, `position` behavior on `TR` is not defined. While it works fine in other browsers, Safari does not allow to set the position.

https://bugs.webkit.org/show_bug.cgi?id=240961

linuzer: I see what you say, but unfortunately I am not able to provide a better solution.

For the time being until someone finds a better solution or Safari provides a fix, I personally would prefer to have a working app management page with a "too-small" link on the app-name rather than a broken page. But that's just my 2 cents.

ShGKme:
Fixed in https://github.com/nextcloud/server/pull/44236

apps/settings/src/components/AppList/AppItem.vue
linuzer: No, thank you for this far better solution! 👍

PR #44119, opened 2024-03-11
fix(theming): fix incorrectly inverted favicons

szaimen:
Reason: invertTextColor has changed and is not now incorrectly inverting icons even with the default primary color. This now switches to isBrightColor which more correctly returns what we need here.

| Before | After with default primary color | After with white as primary color | After with black as primary color |
|---|---|---|---|
| image | image | image | image |

### DIFF
--- a/apps/theming/lib/IconBuilder.php
+++ b/apps/theming/lib/IconBuilder.php
@@ -181,7 +181,7 @@ class IconBuilder {
 			 * invert app icons for bright primary colors
 			 * the default nextcloud logo will not be inverted to black
 			 */
-			if ($this->util->invertTextColor($color)
+			if ($this->util->isBrightColor($color)
 				&& !$appIcon instanceof ISimpleFile
 				&& $app !== "core"
 			) {


susnux (approved):
Works but not sure if this is intended, because the background color is not the primary color.
The reason the default looks odd is that for our default color we use the accessible color instead (there is a `if color === ... then instead use other color`).

emoral435:
>The reason the default looks odd is that for our default color we use the accessible color instead

If that is the case, then should we change the colors? I like the change visually, but if it reverts accessibility, should it not be kept?

susnux:
> but if it reverts accessibility, should it not be kept?

No in this case it is only the icon. But until we (maybe) get #42977 in we use a hack in DefaultTheme and in Capabilities that changes the primary color if it is Nextcloud blue to some slighly darker blue -> results then in white text color.

emoral435:
> But until we (maybe) get https://github.com/nextcloud/server/pull/42977 in we use a hack in DefaultTheme and in Capabilities that changes the primary color if it is Nextcloud blue to some slighly darker blue -> results then in white text color.

I see! I like both solutions, but it seems that this would be more of a bandaid fix, so we might as well wait for the above PR to (hopefully?) be merged, seems much cleaner :) LGTM for now then!

emoral435 (approved):

jancborchardt (commented):
Thanks for fixing that! :)

PR #44126, opened 2024-03-11
fix(dav): Abort incomplete CalDAV changes sync

ChristophWurst:
* Ref https://github.com/nextcloud/server/issues/36644#issuecomment-1902087319

## Summary

CalDAV clients ask the server for changes since the last sync. If our records of modifications are too short, there is a gap of new/modified/deleted events that no longer sync.
This attempts to detect when it happens and forces the client to do a full resync as per https://datatracker.ietf.org/doc/html/rfc6578#section-3.2

> Servers might be unable to maintain all of the change data for
         a collection due to storage or performance reasons, e.g.,
         servers might only be able to maintain up to 3 weeks worth of
         changes to a collection, or only up to 10,000 total changes, or
         not wish to maintain changes for a deleted collection.

## TODO

- [ ] Find out why Thunderbird detects and error and doesn't handle the condition per RFC spec
- [ ] Test with other clients too

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/CalDavBackend.php
+++ b/apps/dav/lib/CalDAV/CalDavBackend.php
@@ -2376,7 +2376,7 @@ class CalDavBackend extends AbstractBackend implements SyncSupport, Subscription
 			if ($syncToken) {
 				$qb = $this->db->getQueryBuilder();
 
-				$qb->select('uri', 'operation')
+				$qb->select('uri', 'operation', 'synctoken')
 					->from('calendarchanges')
 					->where(
 						$qb->expr()->andX(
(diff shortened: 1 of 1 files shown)


floyd-fuh:
Just wanted to add +1 (connecting to a Nextcloud 28.0.5 that has this problem) and also there are more users than the ones on github experiencing this: https://help.nextcloud.com/t/calendar-sync-issues-on-ios-and-ipados/165952/3
Running into this with iPhone native webdav support and busycal on MacOS. When debugging with busycal we can see:

```
2024-05-08 14:05:06 -2- CalDAVSyncTask
(D6043455-BD72-4E6E-A53D-0B7A695740F2): calendar
'REDACTED-THIS-IS-BROKEN-CALENDAR' has changes, will fetch as delta
(sync property: sync-token, old token: http://sabre.io/ns/sync/2206, new
token: http://sabre.io/ns/sync/2206, pending sync: 0, subscribed: 0)
```

But then there are actually no changes sent. Forcing full-sync by unsubscribing and resubscribing is an annoying workaround.

PR #44129, opened 2024-03-11
feat: Apps discover section

### LINKED ISSUE #42547: 🆕 Apps "Discover" section: Showcasing app ecosystem inside the app store
As we have more and more apps, the app store makes it more and more difficult to find relevant content or featured apps. A mechanism to solve this could be a new "Discover" section as the main entry.

This would be to feature not just apps, but generally content, like the "Today" tab in the iOS app store: https://developer.apple.com/app-store/discoverability/
- We already have plenty of content on our blog like interviews, podcasts, developer stuff etc. that we can use.
- Similarly we can have "5 recommended productivity apps" etc, just groups of apps with a little text alongside it which can initially be our "app bundles" but presented in a much nicer way
- 5 top-rated apps (recently and/or all-time)
- Much more content

This does not need to always be updated but can be circled around every once in a while (this is done in the iOS App Store Today tab as well).

An example of that was in this mockup for the new online app store where we thought about inserting content like that inbetween the apps. But this would be a full-on way to discover apps.
Image

Also ref related "new in Nextcloud": https://github.com/nextcloud/server/issues/9165

susnux:
* Resolves: #42547 
* Requires:
  * [x] https://github.com/nextcloud/appstore/pull/1313

## Summary

https://github.com/nextcloud/server/assets/1855448/b36eb179-e062-4944-9099-00fd55e91153

## TODO

- [x] Drop the test commit before merge :exclamation: 
- [x] <del>(maybe follow up) Implement show case type</del> (follow up)

## Checklist


apps/settings/lib/Controller/AppSettingsController.php
```diff
@@ -106,6 +123,84 @@ public function viewApps(): TemplateResponse {
...
+	 * @PublicPage
+	 * @NoCSRFRequired
+	 *
+	 * Get a image for the app discover section - this is proxied for privacy and CSP reasons
+	 *
+	 * @param string $image
+	 * @throws \Exception
+	 */
+	public function getAppDiscoverMedia(string $fileName): Response {
+		$etag = $this->discoverFetcher->getETag() ?? date_format(new \DateTimeImmutable(), 'Y-m');
```
Altahrim, line 151: Or simply:
```suggestion
		$etag = $this->discoverFetcher->getETag() ?? date('Y-m');
```

apps/settings/lib/Controller/AppSettingsController.php
```diff
@@ -106,6 +123,84 @@ public function viewApps(): TemplateResponse {
...
+			$folder = $this->appData->getFolder('app-discover-cache');
+			// Cleanup old cache folders
+			$allFiles = $folder->getDirectoryListing();
+			foreach ($allFiles as $dir) {
+				if ($dir->getName() !== $etag) {
+					$dir->delete();
+				}
+			}
+		} catch (NotPermittedException $e) {
+			// can be ignored - this is from the cleanup
```
Altahrim, line 162: If a `$dir->delete()` fails, following deletes will not happen. Am I right?

apps/settings/lib/Controller/AppSettingsController.php
```diff
@@ -106,6 +123,84 @@ public function viewApps(): TemplateResponse {
...
+		}
+
+		$info = pathinfo($fileName);
+		$hashName = md5($fileName);
+		$allFiles = $folder->getDirectoryListing();
+		// Try to find the file
+		$file = array_filter($allFiles, function (ISimpleFile $file) use ($hashName) {
+			return str_starts_with($file->getName(), $hashName);
+		});
+		$file = $file[0] ?? null;
```
Altahrim, line 181: If key `0` if delete by `array_filter`, it will not work. Instead:
```suggestion
		$file = reset($file) ?? null;
```

Altahrim (commented):
Few comments on PHP part

apps/settings/lib/Controller/AppSettingsController.php
Altahrim: Yes, I think so :) Maybe with a function to avoid to many indentation levels
```php
function cleanOldCache(array $allFiles, string $etag): void {
	foreach ($allFiles as $dir) {
		if ($dir->getName() === $etag) { // Is it really getName?
			continue;
		}
		try {
			$dir->delete();
		} catch (NotPermittedException) {
			// ignore
		}	
	}
}

File: lib/composer/composer/autoload_psr4.php [hunk 4]
```diff
@@ -9,5 +9,6 @@
     'OC\\Core\\' => array($baseDir . '/core'),
     'OC\\' => array($baseDir . '/lib/private'),
     'OCP\\' => array($baseDir . '/lib/public'),
+    'Bamarni\\Composer\\Bin\\' => array($vendorDir . '/bamarni/composer-bin-plugin/src'),
```
Pytal, line 12:
```suggestion
```
Only for dev?

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
```diff
@@ -0,0 +1,114 @@
...
+		IRegistry $registry,
+		private CompareVersion $compareVersion,
+	) {
+		parent::__construct(
+			$appDataFactory,
+			$clientService,
+			$timeFactory,
+			$config,
+			$logger,
+			$registry
```
Pytal, line 54:
```suggestion
```
I believe DI already injects these

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
```diff
@@ -0,0 +1,114 @@
...
+		);
+
+		$this->fileName = 'discover.json';
+		$this->endpointName = 'discover.json';
```
Pytal, line 58: Could these be consts instead?

apps/settings/discover.json
```diff
@@ -0,0 +1,200 @@
+[
+	{
+		"type": "post",
+		"id": "1d41a08e-4aa6-49b3-ad1b-ff2e83bcdc85",
+		"headline": {
+			"en": "Amazing wallpapers",
+			"de": "Bezaubernde Hintergründe"
+		},
+		"text": {
+			"en": "This are the most amazing wallpaers that you can possible have. Get them now!",
```
skjnldsv, line 10:
```suggestion
			"en": "This are the most amazing wallpapers that you can possible have. Get them now!",
```

apps/settings/lib/Controller/AppSettingsController.php
susnux: Like that! Implemented!

apps/settings/lib/Controller/AppSettingsController.php
```diff
@@ -106,6 +124,86 @@
...
+			}
+		} catch (NotPermittedException $e) {
+			// can be ignored - this is from the cleanup
+		} catch (\Throwable $e) {
+			$folder = $this->appData->newFolder('app-discover-cache');
+		}
+
+		// Get the current cache folder
+		try {
+			$folder = $folder->getFolder($etag);
```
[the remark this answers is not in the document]
susnux: The folder can not be null, because either it is set in the previous try, or that throws.
Then it is set in the catch - or that throws and we never reach this part of the file.

apps/settings/lib/Controller/AppSettingsController.php
```diff
@@ -106,6 +125,98 @@ public function viewApps(): TemplateResponse {
...
+	 * Get all active entries for the app discover section
+	 *
+	 * @NoCSRFRequired
+	 */
+	public function getAppDiscoverJSON(): JSONResponse {
+		// $data = $this->discoverFetcher->get();
+
+		// TODO REMOVE JUST FOR TESTING
+		$file = file_get_contents(__DIR__ . '/../../discover.json');
+		$data = json_decode($file);
```
skjnldsv, line 138: ok, make sense! :+1:

apps/settings/discover.json
susnux: Yes, but this is only for testing and will (hopefully!) be removed before merge (see todo section) :)

apps/settings/src/components/AppStoreDiscover/CarouselType.vue
```diff
@@ -0,0 +1,204 @@
...
+
+			hasNext,
+			hasPrevious,
+			currentIndex,
+			shownElement,
+
+			transitionName,
+
+			translatedHeadline,
+			// translatedText,
```
Pytal, line 132:
```suggestion
```

apps/settings/src/components/AppStoreDiscover/CarouselType.vue
```diff
@@ -0,0 +1,204 @@
...
+		 */
+		content: {
+			type: Array as PropType<IAppDiscoverCarousel['content']>,
+			required: true,
+		},
+	},
+
+	setup(props) {
+		const translatedHeadline = useLocalizedValue(computed(() => props.headline))
+		// const translatedText = useLocalizedValue(computed(() => props.text))
```
Pytal, line 95:
```suggestion
```

Pytal (approved):
Other than minor remarks ✅, did not test

apps/settings/src/components/AppStoreDiscover/CarouselType.vue
susnux: :+1:

apps/settings/src/components/AppStoreDiscover/CarouselType.vue
susnux: :+1:

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
susnux: Not sure did it the same way it was done on the AppFetcher which also extends the Fetcher class

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
susnux: No we extend the parent so we need to pass it to the parent. DI only injects it in our constructor.

lib/composer/composer/autoload_psr4.php
susnux: Yes!

apps/settings/src/utils/appDiscoverTypeParser.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import type { IAppDiscoverCarousel, IAppDiscoverElements, IAppDiscoverPost, IAppDiscoverShowcase } from '../constants/AppDiscoverTypes.ts'
+
+/**
+ * Helper to transform the JSON API results to proper frontend objects (app discover section elements)
+ *
+ * @param element The JSON API element to transform
+ */
+export const apiTypeParser = (element: Record<string, unknown>): IAppDiscoverElements => {
```
Pytal, line 30: This file is not used?

susnux: Lost in rebase :see_no_evil: 
Its back :) We need this to validate the results and parse dates (while dates are not yet fully implemented)

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
```diff
@@ -0,0 +1,114 @@
...
+
+			// Remove expired or future entries
+			return array_filter($entries, function (array $entry) use ($now) {
+				try {
+					$date = new DateTimeImmutable($entry['date'] ?? '');
+					if ($date > $now) {
+						return false;
+					}
+				} catch (\Throwable $e) {
+					// ignore
```
emoral435, line 80: since the conversion throws an error, can't that indicate something is wrong with the formatting of the entry? so shouldnt we filter out the entry?

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
```diff
@@ -0,0 +1,114 @@
...
+				}
+
+				try {
+					$expiryDate = new DateTimeImmutable($entry['expiryDate'] ?? '');
+					if ($expiryDate < $now) {
+						return false;
+					}
+				} catch (\Throwable $e) {
+					// ignore
```
emoral435, line 89: same as above - unless you believe it best to keep :)

emoral435 (approved):
Tested on my local dev - looks really cool :) Small questions but overall - fantastic PR!

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
susnux: makes sense, then invalid dates are filtered out :+1:

lib/private/App/AppStore/Fetcher/AppDiscoverFetcher.php
susnux: makes sense, then invalid dates are filtered out :+1:

Altahrim:
Removed the test commit

PR #44132, opened 2024-03-11
fix: don't return null for SharedStorage::getWrapperStorage with share recursion

icewind1991:
While this might not fix errors from share recursion it should at least result in better errors.


come-nc (approved):

Altahrim (approved):

susnux:
@icewind1991 conflicts

come-nc:
> @icewind1991 conflicts

Fixed them

PR #44141, opened 2024-03-12
fix: Return only administered groups of a user for subadmins

Pytal:
## Summary

- Fixes a bug when disabling a user as a subadmin who does not have access to all a users groups shows an error
- Prevents returning extraneous user information to subadmins

## Checklist

### DIFF
--- a/apps/provisioning_api/lib/Controller/AUserData.php
+++ b/apps/provisioning_api/lib/Controller/AUserData.php
@@ -137,6 +137,9 @@ abstract class AUserData extends OCSController {
 		$groups = $this->groupManager->getUserGroups($targetUserObject);
 		$gids = [];
 		foreach ($groups as $group) {
+			if (!$this->groupManager->getSubAdmin()->isSubAdminOfGroup($currentLoggedInUser, $group)) {
+				continue;
+			};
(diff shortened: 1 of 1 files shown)


emoral435 (approved):

emoral435:
LGMT :)

emoral435:
Haha, just noticed that my review no longer works ^*^ but LGTM!

apps/provisioning_api/lib/Controller/AUserData.php
```diff
@@ -137,6 +137,9 @@ protected function getUserData(string $userId, bool $includeScopes = false): ?ar
 		$groups = $this->groupManager->getUserGroups($targetUserObject);
 		$gids = [];
 		foreach ($groups as $group) {
+			if (!$this->groupManager->getSubAdmin()->isSubAdminOfGroup($currentLoggedInUser, $group)) {
```
nickvergessen, line 140:
```suggestion
			if (!$isAdmin && !$this->groupManager->getSubAdmin()->isSubAdminOfGroup($currentLoggedInUser, $group)) {
```

But not sure we should do that.
It might be important to see that a user is in more/other groups?

PR #44162, opened 2024-03-12
fix(files): fixed breadcrumbs dissapearing on narrow screens

### LINKED ISSUE #44120: opening sidebar hides broadscrombs totally with some screen resolutions
Just tested on nc29.0.0b1,
opening the sidebar hides the broadscrumbs totally under some circumstances:
### Sidebar open:
image
### Sidebar closed:
image

emoral435:
* Resolves #44120

## Summary
NcBreadcrumbs already handled for when breadcrumbs are too long for the view-width by making the unmaintainable breadcrumbs composed into a drop-down list, so we only have to care about hiding the breadcrumbs on small view ports. I tried 420, as that is what we use everywhere for standard, but 400 worked much better with the styling of the navbar.

## Screenshot
https://github.com/nextcloud/server/assets/110193237/dca24ae3-9567-4e74-bf40-e3699e45ba9a

## Checklist


apps/files/src/components/BreadCrumbs.vue
```diff
@@ -126,9 +126,9 @@ export default defineComponent({
 			return this.uploaderStore.queue.length !== 0
 		},
 
-		// Hide breadcrumbs if an upload is ongoing on arrow screens
+		// Hide breadcrumbs if an upload is ongoing
 		shouldShowBreadcrumbs(): boolean {
-			return this.filesListWidth > 768 && !this.isUploadInProgress
+			return this.filesListWidth > 400 && !this.isUploadInProgress
```
skjnldsv, line 131: This is not really a number we use anywhere.
Can we use one of them? 512 or 768 ?

emoral435: Ahh, both suggested numbers are still not big enough to fix the initial problem of the breadcrumbs disappearing. 

There is also the route of simply removing the viewport width check, but then on big screens, the breadcrumbs leave for a second and then reappear, which looks weird in my personal opinion. 

420 works, but there is a small section of viewport where it disappears, but it does not seem to be that big of a problem. We use 420 in other libraries, should we use that instead?

susnux: Yes I think I have seen 420 as well for mobile devices like smartphones. So 420 would make sense.

>  512 or 768

Nice numbers, but they are not very useful for responsiveness here IMHO.

skjnldsv: > Nice numbers, but they are not very useful for responsiveness here IMHO.

Right
https://github.com/nextcloud-libraries/nextcloud-vue/blob/b3aee0c29b85618e591006fd7363224a57a3957b/src/assets/variables.scss#L57-L58

In any case, searching for `@media` on the nc org is a nightmare. We need to use appropriate numbers, everyone uses different ones :see_no_evil:

skjnldsv (approved):

apps/files/src/components/BreadCrumbs.vue
skjnldsv: We can keep 400 @emoral435

susnux: Yes I agree :) 
For me when talking about mobile devices I mostly think about smartphones used in portrait mode, so this would be something below 480px which mostly need special handling.

susnux (approved):

skjnldsv:
@emoral435 now, even on wide screens, we hide the breadcrumbs if any upload is ongoing, this seems like a regression?

emoral435:
@skjnldsv Agreed 👍 Will make a PR to fix this. However, I will test to see if the last return condition is even needed, as within somewhere in FilesList there already is a feature to truncate the breadcrumbs around that view port width anyways, so having two places hide the breadcrumbs just increases the complexity / amount of things we would have to find and then change in the future in case we decide to change this :)

PR #44178, opened 2024-03-13
fix(Navigation): ensure NavManager is inited when adding an entry

blizzz:
## Summary

otherwise custom order cannot be applied

Needed by https://github.com/nextcloud/tables/pull/904

## Checklist

### DIFF
--- a/lib/private/NavigationManager.php
+++ b/lib/private/NavigationManager.php
@@ -94,6 +94,7 @@ class NavigationManager implements INavigationManager {
 			$this->closureEntries[] = $entry;
 			return;
 		}
+		$this->init();
 
 		$id = $entry['id'];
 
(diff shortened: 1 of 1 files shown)


Pytal:
Would adding as a closure entry be better? ref: https://github.com/nextcloud/server/commit/970846624feb154800666c7c3293d9ec27861c71

blizzz:
> Would adding as a closure entry be better? ref: [9708466](https://github.com/nextcloud/server/commit/970846624feb154800666c7c3293d9ec27861c71)

Ah, probably, but it does not make this fix less useful.

Even better would be an Event one could listen to. Right now I react to `BeforeTemplateRenderedEvent`, but it does only feel so right.

Pytal:
> Ah, probably, but it does not make this fix less useful.

Agreed!

> Even better would be an Event one could listen to. Right now I react to `BeforeTemplateRenderedEvent`, but it does only feel so right.

👍

Pytal (approved):

blizzz:
:ballot_box_with_check:  Changed my code to supply a Closure – thanks!

nickvergessen (approved):

PR #44179, opened 2024-03-13
fix: Ensure correct stacking context for right click menu

Pytal:
## Summary

- Fix incorrect stacking context of `main.app-content` `position: static` when styles are loaded in a different order e.g. when the text app is enabled

image

## Checklist


apps/files/src/views/FilesList.vue
```diff
@@ -628,7 +628,7 @@ export default defineComponent({
 	overflow: hidden;
 	flex-direction: column;
 	max-height: 100%;
-	position: relative;
```
susnux, line 631: where is this applied? I do not see any "app-content" class in this file and the styles are scoped?

Pytal: It is in fact scoped yes, but it seems that it only adds the data-v-[hash] to the style declaration without requiring that the class itself is in the markup 🤔

And so it still applies:
image

skjnldsv (approved):

apps/files/src/views/FilesList.vue
susnux: weird

susnux (approved):

susnux:
@Pytal backport?

PR #44189, opened 2024-03-14
Do not throw an exception if the etag is not set in metadata

artonge:
Currently, if no `etag` is set for the metadata, the caller need to wrap `getEtag` in a `try/catch`, else, the provider will always fail at the same place. This returns an empty string instead of throwing.


susnux (approved):
Makes sense and is consistent with out other `getETag` functions

skjnldsv (approved):

PR #44217, opened 2024-03-15
fix: List specific share permissions as subline

nfebe:
## Summary

| Before | After |
| --------- | ------ |
| Screenshot from 2024-03-15 11-16-10 | Screenshot from 2024-03-15 11-26-54 |

Resolves : no-issue


apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -1056,9 +1056,15 @@ export default {
 						flex-direction: column;
 					}
 				}
-
-				.subline {
-					display: block;
+				
+				/* Target component based style in NcCheckboxRadioSwitch slot content*/
+				::v-deep span.checkbox-content__text.checkbox-radio-switch__text {
```
susnux, line 1061: v-deep is deprecated, use deep instead

(vue3 docs but also applies to vue 2.7): https://vuejs.org/api/sfc-css-features.html#deep-selectors

```suggestion
				:deep(span.checkbox-content__text.checkbox-radio-switch__text) {
```

emoral435 (approved):

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -1056,9 +1056,15 @@ export default {
 						flex-direction: column;
 					}
 				}
-
-				.subline {
-					display: block;
+				
+				/* Target component based style in NcCheckboxRadioSwitch slot content*/
+				:deep(span.checkbox-content__text.checkbox-radio-switch__text) {
```
susnux, line 1061: Is the `flex-wrap` really needed? 
Because using the internals of the checkbox is pretty dangerous as they are considered private API that can change at any time. E.g. `span` could be switched to `div`.
So if it also works without then you do not need any `deep` at all.

Fenn-CS: The component-scoped styles prevent the element from creating a new line. Initially, this feature worked on the master branch but failed to integrate into stable28 (https://github.com/nextcloud/server/pull/43367). Currently, it doesn't work on the master branch.

In the worst-case scenario, the component-scoped styles or element may change or update, causing it to break again for the master branch and versions thereafter. This is exactly what happened after version 27.

Previously, using 'display: block' was sufficient to move the text to the subline. However, the component updates in version 28 and beyond broke this functionality, even though it still works on version 27.

susnux: I see, thank you!

susnux (approved):

PR #44218, opened 2024-03-15
feat: Limit email input to 255 chars

nfebe:
# Screenshots

| A | B | C |
| -- | -- | -- |
| Screenshot from 2024-03-15 18-10-06 | Screenshot from 2024-03-16 01-06-51 | Screenshot from 2024-03-16 01-07-02 |


skjnldsv:
Labels pleaaase :)

core/Controller/LoginController.php
```diff
@@ -336,9 +336,15 @@ public function tryLogin(Chain $loginChain,
 			);
 		}
 
+		$user = trim($user);
+
+		if (strlen($user) > 255) {
+			return new JSONResponse($this->error($this->l10n->t('Unsupported email length (>255)')));
```
susnux, line 342: Instead: `return $this->createLoginFailedResponse(...)`

core/Controller/LoginController.php
```diff
@@ -336,9 +336,20 @@ public function tryLogin(Chain $loginChain,
 			);
 		}
 
+		$user = trim($user);
+
+		if (strlen($user) > 255) {
```
Altahrim, line 341: We limit to 255 bytes or chars?
If the later:
```suggestion
		if (mb_strlen($user) > 255) {
```

core/Controller/LoginController.php
```diff
@@ -336,9 +336,20 @@ public function tryLogin(Chain $loginChain,
 			);
 		}
 
+		$user = trim($user);
+
+		if (mb_strlen($user) > 255) {
```
susnux, line 341: I am not sure but is the field not set to length 255? For SQL as far as I know this is the length in bytes not chars so not mb_strlen?

Fenn-CS: SQL uses bytes to store character data, but the number of bytes required for each character depends on the database's collation configuration. In general, a single character can occupy between 1 and 4 bytes of storage space, with the exact size determined by the character set and collation used by the database.

I think it's best to return to byte based counter.

core/Controller/LoginController.php
Fenn-CS: @Altahrim see : https://github.com/nextcloud/server/pull/44218#discussion_r1532363550

core/Controller/LoginController.php
susnux: Yes but if the database field is created with `VARCHAR(255)` the n it only can take 255bytes not characters? So we need bytes here. (Even though emails can by RFC only hold up to 320 bytes)

susnux (approved):

core/Controller/LoginController.php
Altahrim: `VARCHAR(255)` can accept 255 characters. The number of bytes will vary according to charset.

I would be in favor to use a correct email validation and have a larger field in database (it costs nothing to have `VARCHAR(320)` instead of `VARCHAR(255)`).

For a first fix, `strlen` and `mb_strlen` are fine for me :)

core/Controller/LoginController.php
```diff
@@ -336,9 +336,20 @@
 			);
 		}
 
+		$user = trim($user);
+
+		if (strlen($user) > 255) {
+			return $this->createLoginFailedResponse(
+				$user,
+				$user,
+				$redirect_url,
```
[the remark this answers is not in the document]
Altahrim: Maybe the following or change default on line 317:
```suggestion
				(string) $redirect_url,
```

Altahrim (approved):

core/Controller/LoginController.php
susnux: Then this is highly database dependent, as msql states:

> varchar [ ( n | max ) ]
> Variable-size string data. Use n to define the string size in **bytes** and can be a value from 1 through 8,000

PR #44230, opened 2024-03-15
config: fix correctness issues in reading

pulsejet:
Multiple fixes here.

1. Checking readability is pointless, since fopen is guaranteed to not throw and return false if the file doesn't exist or is not readable ~~(it even logs a warning when the file isn't readable, which is very nice)~~ 
2. ~~Skip even for non-main fails when we can't open it (it'll crash later anyway otherwise)~~ Show the admin a better error message when a config file cannot be opened instead of throwing a TypeError (see https://github.com/nextcloud/server/pull/44230#issuecomment-2013126704)
3. Reduce the size of the critical block
4. Make sure to unlock the file even if something fails
5. Check opcache on every request. This is low overhead since it only revalidates the timestamp for `config.php` (even if timestamp validation is disabled in the opcache config) and actually fixes a _lot_ of correctness issues. Some actions such as upgrading an app make it obvious that the opcache may need to be refreshed, but config changes might happen more subtly, and opcache defeats the whole point of locking the file to begin with.

Side note: while `Config::writeData` does invalidate the opcache anyway, this is not sufficient since
1. It may be called from the CLI
2. It only affects a single PHP-FPM instance; other hosts wont be affected

### DIFF
--- a/lib/private/Config.php
+++ b/lib/private/Config.php
@@ -215,13 +215,24 @@ class Config {
 
 		// Include file and merge config
 		foreach ($configFiles as $file) {
-			$fileExistsAndIsReadable = file_exists($file) && is_readable($file);
-			$filePointer = $fileExistsAndIsReadable ? fopen($file, 'r') : false;
-			if ($file === $this->configFilePath &&
-				$filePointer === false) {
-				// Opening the main config might not be possible, e.g. if the wrong
-				// permissions are set (likely on a new installation)
-				continue;
+			unset($CONFIG);
+
+			// Invalidate opcache (only if the timestamp changed)
+			if (function_exists('opcache_invalidate')) {
+				opcache_invalidate($file, false);
+			}
+
+			$filePointer = @fopen($file, 'r');
+			if ($filePointer === false) {
+				// e.g. wrong permissions are set
+				if ($file === $this->configFilePath) {
+					// opening the main config file might not be possible
+					// (likely on a new installation)
+					continue;
+				}
+
+				http_response_code(500);
+				die(sprintf('FATAL: Could not open the config file %s', $file));
 			}
 
 			// Try to acquire a file lock
@@ -229,8 +240,14 @@ class Config {
 				throw new \Exception(sprintf('Could not acquire a shared lock on the config file %s', $file));
 			}
 
-			unset($CONFIG);
-			include $file;
+			try {
+				include $file;
+			} finally {
+				// Close the file pointer and release the lock
+				flock($filePointer, LOCK_UN);
+				fclose($filePointer);
+			}
+
 			if (!defined('PHPUNIT_RUN') && headers_sent()) {
 				// syntax issues in the config file like leading spaces causing PHP to send output
 				$errorMessage = sprintf('Config file has leading content, please remove everything before "<?php" in %s', basename($file));
@@ -242,10 +259,6 @@ class Config {
 			if (isset($CONFIG) && is_array($CONFIG)) {
 				$this->cache = array_merge($this->cache, $CONFIG);
 			}
-
-			// Close the file pointer and release the lock
-			flock($filePointer, LOCK_UN);
-			fclose($filePointer);
 		}
 
 		$this->envCache = getenv();


pulsejet:
I believe this is low risk and can be backported as a bugfix (there are technically two "bugs": the handle never being unlocked / closed in certain cases, and all the issues caused by the missing opcache invalidation)

lib/private/Config.php
```diff
@@ -215,12 +215,16 @@ private function readData() {
...
-				// permissions are set (likely on a new installation)
+			unset($CONFIG);
+
+			// Invalidate opcache (only if the timestamp changed)
+			if (function_exists('opcache_invalidate')) {
+				@opcache_invalidate($file, false);
+			}
+
+			$filePointer = @fopen($file, 'r');
+			if ($filePointer === false) {
```
emoral435, line 226: was it necessary to remove the check for `$file === $this->configFilePath`?

emoral435 (commented):
Overall, I like these changes a lot actually. No need to check if a file is readable or not if all we are doing is locking it :) just a small question regarding the sanitization that was removed for the `configFilePath`

EDIT: it does need to be removed after reading the explanation below!

lib/private/Config.php
emoral435: Ah, I see now! You are right, that is quite the bug. Agreed :)

emoral435 (approved):

come-nc:
What bugs are you trying to fix? Are those bugs that you encountered yourself?

I’d expect PHP to release all locks on crash, I’m surprised the finally block is needed. Otherwise we’d need this kind of thing in a lot of places.
Also you are silencing the fopen errors, which will make understanding the situation harder in cases where fopen fails.

pulsejet:
> What bugs are you trying to fix? Are those bugs that you encountered yourself?

Opcache. For example https://github.com/pulsejet/memories/issues/578#issuecomment-1998997728

> I’d expect PHP to release all locks on crash, I’m surprised the finally block is needed. Otherwise we’d need this kind of thing in a lot of places.

Correct, it's not strictly needed but this is cleaner / more resilient.

> Also you are silencing the fopen errors, which will make understanding the situation harder in cases where fopen fails.

Errors like? There were checks on existence and readability so this would've never logged anything either way (this one is a performance fix to save a couple of kernel calls)

come-nc:
> > What bugs are you trying to fix? Are those bugs that you encountered yourself?
> 
> Opcache. For example [pulsejet/memories#578 (comment)](https://github.com/pulsejet/memories/issues/578#issuecomment-1998997728) I've lost count how many times something similar has been reported
> 
> I'm aware the server tuning docs mention the pitfalls of disabling revalidation, but revalidating one file (`config.php`) and every file on every request aren't the same thing (not even close). By revalidating just the config file for every request (which carries almost no cost), disabling general revalidation becomes much more realistic.

We should not impact everyone to handle the case of a badly configured opcache. If opcache is used without revalidation it must be manually cleaned on app install/update, no? Or maybe it can be done automatically, don’t know. But disabling opcache for everyone on config file looks overkill.

> > I’d expect PHP to release all locks on crash, I’m surprised the finally block is needed. Otherwise we’d need this kind of thing in a lot of places.
> 
> Correct, it's not strictly needed but this is cleaner / more resilient. In case the exception is caught, we no longer rely on ~PHP GC behavior~ the request to end for the file to unlock.

Good point, did not think of the case it’s catched.

> > Also you are silencing the fopen errors, which will make understanding the situation harder in cases where fopen fails.
> 
> Errors like? There were checks on existence and readability so this would've never logged anything either way (this one is a performance fix to save a couple of kernel calls)

Any `fopen` failure other than non-existing file.
https://github.com/php/php-src/blob/master/main/fopen_wrappers.c#L700 too long path for instance.

pulsejet:
More thoughts?

lib/private/Config.php
```diff
@@ -215,12 +215,16 @@ private function readData() {
...
-			$filePointer = $fileExistsAndIsReadable ? fopen($file, 'r') : false;
-			if ($file === $this->configFilePath &&
-				$filePointer === false) {
-				// Opening the main config might not be possible, e.g. if the wrong
-				// permissions are set (likely on a new installation)
+			unset($CONFIG);
+
+			// Invalidate opcache (only if the timestamp changed)
+			if (function_exists('opcache_invalidate')) {
+				@opcache_invalidate($file, false);
```
come-nc, line 222:
```suggestion
				opcache_invalidate($file, false);
```
Any reason for the `@` operator here?
I do not see any warning mention on https://www.php.net/manual/en/function.opcache-invalidate.php

come-nc (commented):
Can you detail the rationale on "Skip even for non-main fails when we can't open it (it'll crash later anyway otherwise)" ?
With this PR the admin will have a hard time noticing the permissions on an additional config file are wrong, nothing is logged or shown.

pulsejet:
> Can you detail the rationale on "Skip even for non-main fails when we can't open it (it'll crash later anyway otherwise)" ? With this PR the admin will have a hard time noticing the permissions on an additional config file are wrong, nothing is logged or shown.

True ... the current behavior (before this patch) is odd; it just goes on with a `false` in the file and throws a TypeError. Looking at the error you can't get any idea what is wrong either.

I've updated the patch to show a better error. (note we can't actually `throw` here since the exception gets caught and the config not being loaded leads to more issues)

|image|image|

lib/private/Config.php
pulsejet: It was used in `writeData` below. Removed the `@`.

come-nc (approved):

Altahrim (approved):

lib/private/Config.php
```diff
@@ -215,22 +215,39 @@ private function readData() {
...
 				throw new \Exception(sprintf('Could not acquire a shared lock on the config file %s', $file));
 			}
 
-			unset($CONFIG);
-			include $file;
+			try {
+				include $file;
+			} finally {
+				// Close the file pointer and release the lock
+				flock($filePointer, LOCK_UN);
```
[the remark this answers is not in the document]
pulsejet: Agree with the first part but explicit seems better. For the second part though, the file is always locked here; it's a finally block and the lock happens before the try begins.

pulsejet:
I can't merge due to CI failure, seems unrelated

emoral435:
Rerunning CI

lib/private/Config.php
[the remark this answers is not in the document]
Altahrim: Yes you are right, sorry for the second part ^^ 
I was expecting to save a system call with the first part :)

pulsejet:
> /backport to stable29

@szaimen this is already merged to 29

szaimen:
Ah I see

PR #44258, opened 2024-03-17
fix: getLogLevel return type

skjnldsv:
Fixes https://github.com/nextcloud/server/pull/43534

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -281,7 +281,8 @@ class Log implements ILogger, IDataLogger {
 			}
 		}
 
-		return min($this->config->getValue('loglevel', ILogger::WARN) ?? ILogger::WARN, ILogger::FATAL);
+		$configLogLevel = $this->config->getValue('loglevel', ILogger::WARN);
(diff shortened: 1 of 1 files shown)


SystemKeeper (approved):

susnux:
Ah yes had the same issue as something set the default to `'0'`. But maybe we should add some logging for this case to warn the user that the log level config is invalid?

susnux (approved):
Works but I think the logging (see commit) would make sense

lib/private/Log.php
```diff
@@ -281,7 +281,8 @@ public function getLogLevel($context): int {
 			}
 		}
 
-		return min($this->config->getValue('loglevel', ILogger::WARN) ?? ILogger::WARN, ILogger::FATAL);
+		$configLogLevel = $this->config->getValue('loglevel', ILogger::WARN);
+		return min(is_int($configLogLevel) ? $configLogLevel : ILogger::WARN, ILogger::FATAL);
```
come-nc, line 285: As @susnux pointed out we should log an error or cast the value to int.

nickvergessen: Ref https://github.com/nextcloud/server/pull/44262

PR #44262, opened 2024-03-17
fix(Logger): Warn on invalid `loglevel` configuration option

susnux:
## Summary

When the configuration is not a valid integer do not silently fallback to warn, instead try to cast it (might be more what the admin wanted to configure, e.g. allows `'0'` when `0` was meant).
And if the value is invalid do handle it not silently but at least put a warning in the log (even we do not have access to the logger we can put it in the error log.

## Checklist


Altahrim (approved):

nickvergessen (approved):

come-nc (approved):

skjnldsv:
Thank you for the additional fix :)

PR #44274, opened 2024-03-18
Fixes incorrect "free" value in stat

hopleus:
## Summary
Fixed the problem with incorrect output of "free" when receiving stat

## Use case
If you connect an ObjectStore, for example S3, and do not set a quota for the user, then NextCloud returns the total free space, and since the ObjectStore is connected, "-2" (SPACE_UNKNOWN) is returned. When the frontend receives data, it relies on the "free" field to display a warning about the lack of disk space, which is not correct, because there is disk space, but it cannot be obtained due to the connected ObjectStore.

## Checklist

### DIFF
--- a/lib/private/legacy/OC_Helper.php
+++ b/lib/private/legacy/OC_Helper.php
@@ -522,6 +522,13 @@ class OC_Helper {
 			$total = $free + $used;
 		} else {
 			$total = $free; //either unknown or unlimited
+
+			// So that the front does not react to a negative free space value and does not show appropriate warnings
+			if ($free === \OCP\Files\FileInfo::SPACE_UNKNOWN
+				|| $free === \OCP\Files\FileInfo::SPACE_UNLIMITED
+				|| $free === \OCP\Files\FileInfo::SPACE_NOT_COMPUTED) {
+				$free = PHP_INT_MAX;
+			}
 		}
 		if ($total > 0) {
 			if ($quota > 0 && $total > $quota) {


skjnldsv:
@hopleus please remove the merge commits, if you want to keep your PR up to date, please use rebase to master instead :pray:

hopleus:
> @hopleus please remove the merge commits, if you want to keep your PR up to date, please use rebase to master instead 🙏
> 
> EDIT: I did it for you :)

I am very grateful to you. In the future, I will keep in mind

lib/private/legacy/OC_Helper.php
```diff
@@ -522,6 +522,13 @@ public static function getStorageInfo($path, $rootInfo = null, $includeMountPoin
 			$total = $free + $used;
 		} else {
 			$total = $free; //either unknown or unlimited
+
+			// So that the front does not react to a negative free space value and does not show appropriate warnings
+			if ($free === \OCP\Files\FileInfo::SPACE_UNKNOWN
+				|| $free === \OCP\Files\FileInfo::SPACE_UNLIMITED
+				|| $free === \OCP\Files\FileInfo::SPACE_NOT_COMPUTED) {
+				$free = PHP_INT_MAX;
+			}
```
come-nc, line 531: I’m not sure I like doing that here. It might break other code paths expecting to get the information when the free space is unknown.
Couldn’t the frontend be adapted instead?

icewind1991 (changes requested):
The frontend should be adjusted instead to handle these values instead

joshtrichards:
Related front-end PR #43666

PR #44282, opened 2024-03-18
fix(settings): Support `order` property on App Discover elements and hide future elements

susnux:
## Summary

This adds support to pinning elements by setting the `order` property on the element (e.g. `order: 0` will always be the first element to show). Also filter list of elements to remove upcoming and outdated elements (as the json might be cached).

## Checklist


apps/settings/src/components/AppStoreDiscover/AppStoreDiscoverSection.vue
```diff
@@ -50,7 +50,7 @@ const elements = ref<IAppDiscoverElements[]>([])
  * Shuffle using the Fisher-Yates algorithm
  * @param array The array to shuffle (in place)
  */
-const shuffleArray = (array) => {
+const shuffleArray = <T, >(array: T[]): T[] => {
```
Pytal, line 53:
```suggestion
const shuffleArray = <T>(array: T[]): T[] => {
```

Pytal (approved):

apps/settings/src/components/AppStoreDiscover/AppStoreDiscoverSection.vue
susnux: This does not work for lambda functions

artonge (approved):

apps/settings/src/utils/appDiscoverParser.ts
```diff
@@ -45,3 +45,25 @@ export const apiTypeParser = (element: Record<string, unknown>): IAppDiscoverEle
...
+	if (element.date) {
+		// Element not yet published
+		if (element.date - now > 0) {
+			return false
+		}
+	}
+
+	if (element.expiryDate) {
+		// Element expired
+		if (element.expiryDate - now < 0) {
```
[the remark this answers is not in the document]
susnux: yes!

PR #44294, opened 2024-03-19
fix: Pass the mountpoint target user to storages without owner

juliusknorr:
Storages that do not have a dedicated owner (e.g. groupfolders, external
storages) currently always assume the current session user as the owner.
This leads to several issues when there is no user session but a node is
obtained through a user folder.

In order to have the correct user available we need to pass the user
that is used to setup a mountpoint along to the storage layer as we
generally assume that an owner is available for those.

Steps to see the issue:

## Text

- Setup a local global external storage
- As a user with access to it
  - Have text enabled
  - Create a share link with edit permissions
  - Access the share link in a private browser window
  - Try to save the file which fails because file hooks like the version backend or files metadata extractors try to operate with the storage owner

### DIFF
--- a/lib/public/Files/Storage/IStorage.php
+++ b/lib/public/Files/Storage/IStorage.php
@@ -460,4 +460,16 @@ interface IStorage {
 	 * @since 9.0.0
 	 */
 	public function getWatcher();
+
+	/**
+	 * Allow setting the storage owner
+	 *
+	 * This can be used for storages that do not have a dedicated owner, where we want to
+	 * pass the user that we setup the mountpoint for along to the storage layer
+	 *
+	 * @param string|null $user Owner user id
+	 * @return void
+	 * @since 29.0.0
+	 */
+	public function setOwner(?string $user): void;
 }
--- a/apps/files_external/lib/Config/ConfigAdapter.php
+++ b/apps/files_external/lib/Config/ConfigAdapter.php
@@ -140,6 +140,7 @@ class ConfigAdapter implements IMountProvider {
 		}, $storages, $storageConfigs);
 
 		$mounts = array_map(function (StorageConfig $storageConfig, Storage\IStorage $storage) use ($user, $loader) {
+			$storage->setOwner($user->getUID());
 			if ($storageConfig->getType() === StorageConfig::MOUNT_TYPE_PERSONAL) {
 				return new PersonalMount(
 					$this->userStoragesService,
--- a/lib/private/Files/Storage/Common.php
+++ b/lib/private/Files/Storage/Common.php
@@ -868,6 +868,19 @@ abstract class Common implements Storage, ILockingStorage, IWriteStreamStorage {
 	}
 
 	/**
+	 * Allow setting the storage owner
+	 *
+	 * This can be used for storages that do not have a dedicated owner, where we want to
+	 * pass the user that we setup the mountpoint for along to the storage layer
+	 *
+	 * @param string|null $user
+	 * @return void
+	 */
+	public function setOwner(?string $user): void {
+		$this->owner = $user;
+	}
+
+	/**
 	 * @return bool
 	 */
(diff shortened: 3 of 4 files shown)


come-nc (commented):
Is it not possible to set the user in the session?
I’m not sure I understand how this works not having a session, apart from public share.

Other places like https://github.com/nextcloud/server/blob/master/apps/dav/lib/SystemTag/SystemTagsRelationsCollection.php#L56 (Just know this one because I had a conflict about this line in another PR) expects to have a user in session, won’t that cause troubles?

juliushaertl:
Yeah, when debugging further I noticed that setting the session user temporarily would be more sensible for the onlyoffice bug. I came up with https://github.com/nextcloud/server/pull/44295 for that. For public share links we though still need this as the node does not hold information about the share (or share owner) which is used to construct the user folder https://github.com/nextcloud/server/blob/e7a7b4a40184dc3da2c83e858c820625f660e48e/lib/private/Share20/Share.php#L188 but is lost afterwards before this PR.

come-nc:
Should something like https://github.com/nextcloud/server/blob/master/apps/dav/lib/Storage/PublicOwnerWrapper.php be used for public links?

juliushaertl:
That's a great hint, wasn't aware of that wrapper, let me see if that can be adopted

icewind1991 (approved):

juliushaertl:
> Is it not possible to set the user in the session?
> I’m not sure I understand how this works not having a session, apart from public share.

Collabora requests don't have a session itself. Using the current session user might lead to wrong assumptions further down, the mountpoint target user is the more correct one. The larger problem is that the storage API doesn't really cover the use case of having a storage that has "shared" ownership, but at least the mount owner is more correct then just always assuming the logged in user. The nodes API is flexible so that a user folder for a different user can be requested where the current session user would actually be incorrect.

> Should something like https://github.com/nextcloud/server/blob/master/apps/dav/lib/Storage/PublicOwnerWrapper.php be used for public links?

I checked this but think setting the owner on the storage itself seems cleaner to me as it is passed along when building the storage class instance.

lib/public/Files/Storage/IStorage.php
```diff
@@ -460,4 +460,16 @@ public function getUpdater();
...
 	 */
 	public function getWatcher();
+
+	/**
+	 * Allow setting the storage owner
+	 *
+	 * This can be used for storages that do not have a dedicated owner, where we want to
+	 * pass the user that we setup the mountpoint for along to the storage layer
+	 *
+	 * @param string|null $user
```
come-nc, line 470:
```suggestion
	 * @param string|null $user Owner user id
```

come-nc (approved):

lib/public/Files/Storage/IStorage.php
```diff
@@ -460,4 +460,16 @@ public function getUpdater();
...
+
+	/**
+	 * Allow setting the storage owner
+	 *
+	 * This can be used for storages that do not have a dedicated owner, where we want to
+	 * pass the user that we setup the mountpoint for along to the storage layer
+	 *
+	 * @param string|null $user Owner user id
+	 * @return void
+	 * @since 29.0.0
```
nickvergessen, line 472: 29 or 30?
Was merged into master after branch-off, so 30?
Also it's a BC break?

lib/public/Files/Storage/IStorage.php
```diff
@@ -460,4 +460,16 @@ public function getUpdater();
...
+	 */
+	public function setOwner(?string $user): void;
```
nickvergessen, line 474: @juliushaertl Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(all changes in lib/public/)

PR #44328, opened 2024-03-19
feat(settings): Implement `showcase` type for App Discover section

### LINKED ISSUE #42547: 🆕 Apps "Discover" section: Showcasing app ecosystem inside the app store

susnux:
* Resolves: #42547 

## Summary
This is the last missing piece the `showcase` type - basically a list of posts or apps.

Screenshot_20240319_164401

### TODO

In a follow up some design related changes might be needed as currently for apps the AppItem is reused, but that component should probably be split into two: `AppListTableRow` and `AppListGridItem` as otherwise the grid one (used here) look a bit weird - especially if the app is featured and have a rating.

## Checklist


Pytal:
🤔 
image
am I missing some commits to cherry-pick?

Pytal:
Seems to be from https://github.com/nextcloud/server/commit/73dec209037ef281afa2362ee9b17ff1afede661#diff-b905d152496173fc4f9cf6bb26f6f6547bb8b37e0dc1a3fd254c7e4cee427275R297 👀

susnux:
>  am I missing some commits to cherry-pick?

@Pytal no I needed to rebase, should work now :)

Pytal (approved):

AndyScherzinger (approved):

PR #44350, opened 2024-03-20
fix(LDAP): escape DN on check-user

blizzz:
## Summary

the DN has to be escaped differently when used as a base and we were missing it here in the search method call in the check-user command.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Access.php
+++ b/apps/user_ldap/lib/Access.php
@@ -279,6 +279,8 @@ class Access extends LDAPUtility {
 	 * Normalizes a result grom getAttributes(), i.e. handles DNs and binary
 	 * data if present.
 	 *
+	 * DN values are escaped as per RFC 2253
+	 *
 	 * @param array $result from ILDAPWrapper::getAttributes()
 	 * @param string $attribute the attribute name that was read
 	 * @return string[]
@@ -1259,6 +1261,8 @@ class Access extends LDAPUtility {
 
 	/**
 	 * Executes an LDAP search
+	 *
+	 * DN values in the result set are escaped as per RFC 2253
 	 *
 	 * @throws ServerNotAvailableException
 	 */
--- a/apps/user_ldap/lib/Command/CheckUser.php
+++ b/apps/user_ldap/lib/Command/CheckUser.php
@@ -138,7 +138,8 @@ class CheckUser extends Command {
 			$attrs = $access->userManager->getAttributes();
 			$user = $access->userManager->get($uid);
 			$avatarAttributes = $access->getConnection()->resolveRule('avatar');
-			$result = $access->search('objectclass=*', $user->getDN(), $attrs, 1, 0);
+			$baseDn = $this->helper->DNasBaseParameter($user->getDN());
+			$result = $access->search('objectclass=*', $baseDn, $attrs, 1, 0);
 			foreach ($result[0] as $attribute => $valueSet) {
 				$output->writeln('  ' . $attribute . ': ');
 				foreach ($valueSet as $value) {
--- a/apps/user_ldap/lib/Helper.php
+++ b/apps/user_ldap/lib/Helper.php
@@ -206,6 +206,21 @@ class Helper {
 	/**
 	 * sanitizes a DN received from the LDAP server
 	 *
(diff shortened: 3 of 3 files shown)


max-nextcloud:
backporting right away to have a patch the customer can test.

come-nc (commented):
So getDN does not return the raw dn but an escaped version?
Where is it escaped and how?

blizzz:
cf. https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/Access.php#L252 and https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/Access.php#L1600

come-nc:
@blizzz But I was sure we were using `ldap_escape` when using the dn (or anything really) in the search, with this system we may get double escaping.
It’s too late for changing this I suppose but it should be better documented.
https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/User/User.php#L365 does not specify that it’s escaped in any way.

blizzz:
> @blizzz But I was sure we were using `ldap_escape` when using the dn (or anything really) in the search, with this system we may get double escaping. It’s too late for changing this I suppose but it should be better documented. https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/User/User.php#L365 does not specify that it’s escaped in any way.

The DN is escaped in `Helper::sanitizeDN()` and this is used for queries/filters.

When the DN is used as base DN for operations though, the backlash must not be escaped as `\5c`. 

That is the difference.

come-nc (approved):
I still think this should be better documented.

blizzz:
@come-nc I added a commit  with the doc. Not in the user model though, because it returns only what it was being fed with. Is this acceptable?

come-nc:
> @come-nc I added a commit with the doc. Not in the user model though, because it returns only what it was being fed with. Is this acceptable?

Yeah it helps. It still bugs me that we do not escape upon building the filter instead. I think we avoid double-escaping only because we do not apply escape when searching for group members, and we got lucky that uids never need escaping.
To be safe https://github.com/nextcloud/server/blob/1fb5486059aa9846f555e1c0b949a1944695905e/apps/user_ldap/lib/Group_LDAP.php#L808 should escape for instance, because it is later used raw in the filter, and the dn version is already escaped. Our internal code is not clear enough on what is escaped or not and that’s dangerous.

blizzz:
> > @come-nc I added a commit with the doc. Not in the user model though, because it returns only what it was being fed with. Is this acceptable?
> 
> Yeah it helps. It still bugs me that we do not escape upon building the filter instead. I think we avoid double-escaping only because we do not apply escape when searching for group members, and we got lucky that uids never need escaping. To be safe
> 
> https://github.com/nextcloud/server/blob/1fb5486059aa9846f555e1c0b949a1944695905e/apps/user_ldap/lib/Group_LDAP.php#L808
> should escape for instance, because it is later used raw in the filter, and the dn version is already escaped. Our internal code is not clear enough on what is escaped or not and that’s dangerous.

There is no double escape per se. It is only about the backslash that is expected in different forms when used in search filter compared to when used as base. As usage in filters is the common usage, that format was chosen to be saved in the DB.

blizzz:
P.S.: When is the raw uid used in a search filter?

max-nextcloud (approved):

come-nc:
> P.S.: When is the raw uid used in a search filter? (Apart from override as uuid attribute)

https://github.com/nextcloud/server/blob/1fb5486059aa9846f555e1c0b949a1944695905e/apps/user_ldap/lib/Group_LDAP.php#L864 (following the line I posted earlier, `$dn` may be the uid there.)

blizzz:
> > P.S.: When is the raw uid used in a search filter? (Apart from override as uuid attribute)
> 
> https://github.com/nextcloud/server/blob/1fb5486059aa9846f555e1c0b949a1944695905e/apps/user_ldap/lib/Group_LDAP.php#L864
> (following the line I posted earlier, `$dn` may be the uid there.)

Indeed. In those cases when a DN is not being used to reference members :cold_sweat: But a valid case nonetheless. This is not in scope for this PR though, and also it is nothing that can be crafted by a shady user.

PR #44371, opened 2024-03-21
feat(config): implement config lexicon

ArtificialOwl:
A `Config Lexicon` is a list of config keys used by current app. Each entry also define the expected type for the config value, its lazyness/sensitivity.
The loading is done by registering a `IConfigLexicon`.

**Application.php**

```
	public function register(IRegistrationContext $context): void {
		$context->registerConfigLexicon(ConfigLexicon::class);
	}
```

**ConfigLexicon.php**

```
class Lexicon implements IConfigLexicon {
	public function isStrict(): bool {
		return false;
	}

	public function getAppConfigs(): array {
		return [
			new ConfigLexiconEntry('key1', IConfigLexiconEntry::TYPE_STRING, defintion: 'this is a test config key', lazy: true),
			new ConfigLexiconEntry('key2', IConfigLexiconEntry::TYPE_STRING, sensitive: true),
			new ConfigLexiconEntry('key3', IConfigLexiconEntry::TYPE_INT, 42),
			new ConfigLexiconEntry('key4', IConfigLexiconEntry::TYPE_STRING),
		];
	}

	public function getUserPreferences(): array {
		return [];
	}
}
```

**Note:** A description of the config key can be added as 3rd parameter of the `ConfigLexiconEntry`. This information is not stored when running from a web process.

# store and retrieve config value

Once this is done:

### any code trying to wrongly type config values will get an exception

```
$ ./occ config:app:set myapp key1 --value 1 --type integer

In AppConfig.php line 1544:                                                                                     
  The key is typed incorrectly in relation to the app config lexicon
```                                                                                     
                                                                
### set/get on config values set as lazy will work even if not specified in the process

```
	$this->appConfig->setValueInt('myapp', 'key1', 'abc');
```

```
$ ./occ config:app:get myapp key1 --details
  - app: myapp
  - key: key1
  - value: abc
  - type: string
  - lazy: true
  - sensitive: false
```

### default value can be overwritten

```
	$this->appConfig->getValueInt('myapp', 'key3', 0);
```

will returns `42`

### If configured as `strict`, setting an unlisted config values returns an exception

```
$ ./occ config:app:set myapp key5 --value 1

In AppConfig.php line 1530:                                                                                
  The key is not defined in the app config lexicon  
```                                                                                

Also, setting a config key as deprecated will generate a level 1 log entry when the config key is used


lib/private/ConfigLexicon/ConfigLexiconEntry.php
```diff
@@ -0,0 +1,249 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License, version 3,
+ * along with this program. If not, see <http://www.gnu.org/licenses/>
+ *
+ */
+
+namespace OC\ConfigLexicon;
```
nickvergessen, line 25: Has to be public, otherwise apps are not allowed to use it (as per your PR description)

lib/private/ConfigLexicon/ConfigLexiconEntry.php
```diff
@@ -0,0 +1,249 @@
...
+ * @see IConfigLexicon
+ * @since 30.0.0
+ */
+class ConfigLexiconEntry implements IConfigLexiconEntry {
+	private string $definition = '';
+	private ?string $default = null;
+	private bool $lazy = false;
+	private bool $sensitive = false;
+	private bool $deprecated = false;
+	private string $deprecateMessage = '';
```
nickvergessen, line 42: I'd make them all constructor args, to simplify the creation:
```php
		return [
			new ConfigLexiconEntry('key1', IConfigLexiconEntry::TYPE_STRING, 'this is a test config key', lazy: true),
			new ConfigLexiconEntry('key2', IConfigLexiconEntry::TYPE_STRING, sensitive: true),
			new ConfigLexiconEntry('key3', IConfigLexiconEntry::TYPE_INT, 42),
			new ConfigLexiconEntry('key4', IConfigLexiconEntry::TYPE_STRING),
		];
```

ArtificialOwl: done, while keeping the way to set a default value in a method. rest is in constructor

lib/public/ConfigLexicon/ConfigLexiconEntry.php
```diff
@@ -0,0 +1,197 @@
...
+	 * @param bool $deprecated set config key as deprecated
+	 * @since 30.0.0
+	 */
+	public function __construct(
+		private readonly string $key,
+		private readonly int $valueType,
+		string $definition = '',
+		private readonly bool $lazy = false,
+		private readonly bool $sensitive = false,
+		private readonly bool $deprecated = false
```
nickvergessen, line 52:
```suggestion
		private ?string $default = null,
		private readonly bool $lazy = false,
		private readonly bool $sensitive = false,
		private readonly bool $deprecated = false,
		string $definition = '',
```

Can add the default here too. It's mostlikely the most used option so would put it to the front of the list as well.

lib/public/ConfigLexicon/ConfigLexiconEntry.php
```diff
@@ -0,0 +1,197 @@
...
+ * @see IConfigLexicon
+ * @since 30.0.0
+ */
+class ConfigLexiconEntry implements IConfigLexiconEntry {
+	private string $definition = '';
+	private ?string $default = null;
+
+	/**
+	 * @param string $key config key
+	 * @param int $valueType type of config value ({@see self::TYPE_STRING} and others)
```
nickvergessen, line 39:
```suggestion
	 * @param self::TYPE_* $valueType type of config value ({@see self::TYPE_STRING} and others)
```
Should work fine: https://github.com/nextcloud/spreed/blob/52580808b097054143eca504d165adfbd3195212/lib/Service/ConsentService.php#L60-L66

lib/public/ConfigLexicon/ConfigLexiconEntry.php
ArtificialOwl: while the code convert the `$default` value to string (the same way it is stored in the database) forcing the string type here might generate conflicts. using method like `withDefault<type>()` ensure the dev set a correctly typed value

lib/public/ConfigLexicon/ConfigLexiconEntry.php
ArtificialOwl: Hmm. just realize we could use enums :]

lib/public/ConfigLexicon/ConfigLexiconEntry.php
nickvergessen: the constructor could call it as type is given as well?

lib/public/ConfigLexicon/IConfigLexiconEntry.php
```diff
@@ -26,23 +26,26 @@
 
 use OCP\IAppConfig;
 
+enum ConfigLexiconValueType {
```
nickvergessen, line 29: should be in it's own file to be autoloadable?

lib/private/AppConfig.php
```diff
@@ -1567,4 +1575,74 @@ private function getSensitiveKeys(string $app): array {
...
+		/** @var ConfigLexiconEntry $configValue */
+		$configValue = $configDetails['entries'][$key];
+		$type &= ~self::VALUE_SENSITIVE;
+
+		if ($configValue->getValueType() !== match($type) {
+			self::VALUE_STRING => ConfigLexiconValueType::STRING,
+			self::VALUE_INT => ConfigLexiconValueType::INT,
+			self::VALUE_FLOAT => ConfigLexiconValueType::FLOAT,
+			self::VALUE_BOOL => ConfigLexiconValueType::BOOL,
+			self::VALUE_ARRAY => ConfigLexiconValueType::ARRAY,
```
come-nc, line 1609: Why not back the enum with the int to avoid the match?
Is the backed string used anywhere currently?

ArtificialOwl: Enum is string-backed only to avoid conflict of remove/type.
The match is needed to separate the lexicon's type (generic) with `AppConfig`'s type (stored in database as numerical), as the lexicon will also manage `UserPreferences` (I dont feel confident on having a single type list for those 3 api)

skjnldsv:
replaced by https://github.com/nextcloud/server/pull/49399

PR #44392, opened 2024-03-21
feat(command): Print a success message with the add/remove use occ command

zak39:
## Summary

**Before**

When we add or remove an user from a group. The end-user doesn't know if the command was successful or not.

image
image

**After**

Now, the end-user is informed when he/she adds or removes an user from a group with the occ command.

image
image

## TODO

- [ ] Install your Nextcloud instance
- [ ] Create a few users
- [ ] Create a few groups
- [ ] From the terminal, run the following commands : 
  - [ ] `php occ group:adduser <groupname> <uid>`
  - [ ] `php occ group:remove <groupname> <uid>`

## Checklist

### DIFF
--- a/core/Command/Group/AddUser.php
+++ b/core/Command/Group/AddUser.php
@@ -49,7 +49,10 @@ class AddUser extends Base {
 			$output->writeln('<error>user not found</error>');
 			return 1;
 		}
+
 		$group->addUser($user);
+		$output->writeln('<info>success</info>');
+
 		return 0;
 	}
 
--- a/core/Command/Group/RemoveUser.php
+++ b/core/Command/Group/RemoveUser.php
@@ -50,6 +50,9 @@ class RemoveUser extends Base {
 			return 1;
 		}
 		$group->removeUser($user);
+
+		$output->writeln('<info>success</info>');
+
 		return 0;
 	}
 


susnux:
> When we add or remove an user from a group. The end-user doesn't know if the command was successful or not.

They do. The return code is either 0 for success (as usual) and non zero otherwise.

You can even configure your shell to show the exit code visually, or do
`php occ group:adduser <groupname> <uid> && echo 'success'`

zak39:
Hi @susnux :)

Yes, I know. But, I think it's not clear to some people.
Whether they are beginner or senior.

My suggestion is to clearly display a success message to confirm that the order is working.

Like the `php occ app:enable <appid>` command, which shows that the app is enabled.

tcitworld:
Instead of just `success`, it could have a proper sentence: `User $uid was added/removed to/from group $gid`

zak39:
> Instead of just `success`, it could have a proper sentence: `User $uid was added/removed to/from group $gid`

Yes, I think it's a good idea :+1: !

Should I wait for the developers Nextcloud or community to approve your suggestion ?

come-nc:
I agree with @tcitworld, make sure the message is informative.
Also double check that `--quiet` works as expected and silence your added output, for scripts which would want the old behavior.

PR #44412, opened 2024-03-22
fix(CSP): Add CSP nonce by default and convert `browserSupportsCspV3` to blacklist

### LINKED ISSUE #10207: CSP nonce by default

susnux:
* Resolves: https://github.com/nextcloud/server/issues/10207

## Summary

Every modern browser ("modern" like every version since 2013 (over 10 years now!)) support the nonce src.
So we should add it by default and only keep a blacklist instead.

This way even users with for us unknown user agents get the more secure version.

## Checklist

### DIFF
--- a/lib/private/Security/CSP/ContentSecurityPolicyNonceManager.php
+++ b/lib/private/Security/CSP/ContentSecurityPolicyNonceManager.php
@@ -65,17 +65,14 @@ class ContentSecurityPolicyNonceManager {
 	 * Check if the browser supports CSP v3
 	 */
 	public function browserSupportsCspV3(): bool {
-		$browserWhitelist = [
-			Request::USER_AGENT_CHROME,
-			Request::USER_AGENT_FIREFOX,
-			Request::USER_AGENT_SAFARI,
-			Request::USER_AGENT_MS_EDGE,
+		$browserBlocklist = [
+			Request::USER_AGENT_IE,
 		];
 
-		if ($this->request->isUserAgent($browserWhitelist)) {
-			return true;
+		if ($this->request->isUserAgent($browserBlocklist)) {
+			return false;
 		}
 
-		return false;
+		return true;
 	}
 }


SystemKeeper:
Tested with Talk iOS on iOS 15.5 and 17.0 and both worked for me.

MichaIng (approved):
Since IE is not (officially) supported by Nextcloud anymore, even the blacklist might not be needed. But good to have the facility, so we can add further entries in the unlikely but possible case that there are other browsers not supporting CSP nonce sources.

skjnldsv (approved):

lib/private/Security/CSP/ContentSecurityPolicyNonceManager.php
```diff
@@ -65,17 +65,14 @@ public function getNonce(): string {
 	 * Check if the browser supports CSP v3
 	 */
 	public function browserSupportsCspV3(): bool {
-		$browserWhitelist = [
-			Request::USER_AGENT_CHROME,
-			Request::USER_AGENT_FIREFOX,
-			Request::USER_AGENT_SAFARI,
-			Request::USER_AGENT_MS_EDGE,
+		$browserBlacklist = [
```
Altahrim, line 68: Nit:
```suggestion
		$browserBlocklist = [
```
But we can also get rid of the variable:
```php
public function browserSupportsCspV3(): bool {
	// Modern browsers all supports CSPv3
	return !$this->request->isUserAgent([Request::USER_AGENT_IE]]);
}
```

Altahrim (approved):

lib/private/Security/CSP/ContentSecurityPolicyNonceManager.php
susnux: The idea was to keep the variable for later usage.

Altahrim: We can easily add one later if needed :)

juliusknorr (approved):

PR #44450, opened 2024-03-25
fix(breadcrumbs): improved rendering of breadcrumb progress bar

emoral435:
## Summary
See https://github.com/nextcloud/server/pull/44162#issuecomment-2016436846 for context on this PR.
Essentially, users currently cannot browse and navigate to other files while an upload is in progress, so this change fixes the logic dedicated to rendering the breadcrumbs. 
We now only hide breadcrumbs when the width of the file list is 400px AND when there is an upload in progress.

https://github.com/nextcloud/server/assets/110193237/d1f0a74c-6208-4d11-8f8f-6067ab5c4cab

## Checklist


susnux:
Should we not always show the breadcrumbs? Otherwise mobile users can not navigate, no?
And for mobile let the uploader progress wrap (second line below breadcrumbs).

Maybe instead of breadcrumbs show a select component with the paths as option on small mobile?

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -128,7 +128,13 @@ export default defineComponent({
 
 		// Hide breadcrumbs if an upload is ongoing
 		shouldShowBreadcrumbs(): boolean {
-			return this.filesListWidth > 400 && !this.isUploadInProgress
+			// If we're uploading files, only show the breadcrumbs
+			// if the files list is greater than 400px wide
+			if (this.isUploadInProgress) {
+				return this.filesListWidth > 400
```
susnux, line 134: How about just wrap the uploader into second line so both are showed always?

emoral435: I think this actually makes more sense actually - so I agree! Always show the breadcrumbs, just switch the top breadcrumbs to be flex-column when uploading :)

susnux (approved):
Works!

But this makes a bug of the breadcrumbs more visible: They overflow their container sometimes. This also happens on master but is very visible now.
(But needs to be fixed on the components part)

emoral435:
Yup, I agree @susnux . Just testing it out, and the overflow can get quite clunky here and there. The solution you proposed on the design team is good to me!

https://github.com/nextcloud/server/assets/110193237/ce0aad41-8ad5-4df9-9786-a90cb81888be

Here is a video in case you need to showcase what you mean to designers.

susnux:
What is missing is making the uploader "new" button icon only on <370px width. But this needs to be fixed on the library I think.

skjnldsv:
Considering the changes and the discussion here, please loop the design team in before any approval or merge :)

susnux:
> Considering the changes and the discussion here, please loop the design team in before any approval or merge :)

Already done in design chat

susnux:
CC @nextcloud/designers what do you think? The very small breakpoint (icon only new-button) is not implemented as this needs fixes on the library.

Full bread crumbs with share button | full bread crumbs without share button | Icon only first breadcrumb | Smallest possible bread crumbs with full new-button
---|---|---|---
Screenshot_20240327_130623|Screenshot_20240327_130639|Screenshot_20240327_130708|Screenshot_20240327_130722

jancborchardt (commented):
> And this would be a mock of our smallest supported width of 300px, here I made the new-button icon only and reduced the min-width of the last breadcrumb to 72px so it does not overlap with the button

Looks good @susnux – I would even say that "All files" can get moved into the 3-dot menu before the "New" needs to get cut from the button label, as it saves a bunch more space.

emoral435:
@jancborchardt That can definitely be done! Just needs to be changed here within this logic, AFAIK :) :

https://github.com/nextcloud-libraries/nextcloud-vue/blob/d11028854b8cceba83725e417eb7b3563bcc2832/src/components/NcBreadcrumbs/NcBreadcrumbs.vue#L291-L299

However, just as a counterpoint, the last breadcrumb already just shows the user the directory they are in. Clicking on it just would take them to the same place - so isn't it more useful to have all files stay? Plus, we would be able to keep the cool icon :)

I do not have a huge opinion on this, so whatever you decide is best :)

emoral435:
So AFAIK, the changes that are needed for the breadcrumb are all library-dep changes.... so for the time being, anyone want to give this PR a green check 👀💚

jancborchardt:
> However, just as a counterpoint, the last breadcrumb already just shows the user the directory they are in. Clicking on it just would take them to the same place - so isn't it more useful to have all files stay?

It not only is a link though, it alsp is the only place where we show the name of the folder you are in. And that is more important to know than to quickly go back to "All files". :) (Good for asking though!)

susnux:
So I would say we start with this one and the either move the "all files" into the overflow menu (nextcloud-vue changes) or make the "new +" button icon only (nextcloud-upload change) later?

emoral435:
> So I would say we start with this one and the either move the "all files" into the overflow menu (nextcloud-vue changes) or make the "new +" button icon only (nextcloud-upload change) later?

I agree! This PR gets the mobile view to have correct reactivity, then we can improve it even further by what you mentioned. I will tackle the nextcloud-vue library, do you want to tackle the upload-library?

susnux:
> I will tackle the nextcloud-vue library, do you want to tackle the upload-library?

I think we only need one of them, hiding the icon in the overflow menu should already make enough space so we can keep the "+ new" (and also the other way around).

emoral435:
> I think we only need one of them, hiding the icon in the overflow menu should already make enough space so we can keep the "+ new" (and also the other way around).

Gotcha! Then I propose we keep the `+ New` for now, and I will submit a PR for collapsing the PR then :)

skjnldsv:
Needs a quick rebase, sorry for the conflicts, I forgot to remove some code, yours will replace it anyway

skjnldsv:
Rebasing... :hourglass:

skjnldsv (approved):

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -21,17 +21,17 @@
   -->
 
 <template>
-	<NcBreadcrumbs 
-		data-cy-files-content-breadcrumbs
-		:aria-label="t('files', 'Current directory path')">
+	<NcBreadcrumbs data-cy-files-content-breadcrumbs
+		:aria-label="t('files', 'Current directory path')"
+		class="files-list__breadcrumbs"
+		:class="{ 'files-list__breadcrumbs--with-progress': wrapUploadProgressBar }">
```
skjnldsv, line 27: Took the liberty of changing the class you used @emoral435.
Initially `breadcrumb__progress` was not BEM compliant to indicate a state difference :)

apps/files/src/components/BreadCrumbs.vue
```diff
@@ -280,15 +277,22 @@ export default defineComponent({
 </script>
 
 <style lang="scss" scoped>
-.breadcrumb {
+.files-list__breadcrumbs {
 	// Take as much space as possible
 	flex: 1 1 100% !important;
 	width: 100%;
+	height: 100%;
 	margin-inline: 0px 10px 0px 10px;
```
skjnldsv, line 285: Btw @emoral435 
image

susnux: should be
```css
margin-block: 0;
margin-inline: 10px;
```

susnux: (making it RTL compatible)

apps/files/src/components/BreadCrumbs.vue
emoral435: Looks amazing! 👌

apps/files/src/components/BreadCrumbs.vue
emoral435: Gotcha - will add this to the latest rebase :)

skjnldsv (approved):

skjnldsv:
Related cypress
```
  1) Files: Live photos
       'Show hidden files' is enabled
         Restores files when restoring the .jpg:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy-files-content-breadcrumbs]`, but never found it.
      at n (webpack://nextcloud/./cypress/e2e/files/FilesUtils.ts:119:7)
      at Context.eval (webpack://nextcloud/./cypress/e2e/files/live_photos.cy.ts:147:12)

  2) Files: Live photos
       'Show hidden files' is enabled
         Blocks restoration when restoring the .mov:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy-files-content-breadcrumbs]`, but never found it.
      at n (webpack://nextcloud/./cypress/e2e/files/FilesUtils.ts:119:7)
      at Context.eval (webpack://nextcloud/./cypress/e2e/files/live_photos.cy.ts:158:12)

```

emoral435:
Weird, we did not change anything that should be breaking the cypress 🙈 Will investigate and re-run to see if what works!

susnux:
@emoral435 

> ) Files: Live photos
       'Show hidden files' is enabled
         Restores files when restoring the .jpg:
     AssertionError: Timed out retrying after 4000ms: Expected to find element: `[data-cy-files-content-breadcrumbs]`, but never found it.
      at n (webpack://nextcloud/./cypress/e2e/files/FilesUtils.ts:119:7)
      at Context.eval (webpack://nextcloud/./cypress/e2e/files/live_photos.cy.ts:147:12)

Cypress seems to be related

PR #44455, opened 2024-03-25
fix(personal-files): correctly filters groupfolders now

### LINKED ISSUE #44359: `files` Personal View contains groupfolders
@emoral435 thanks for the work here ! this one has been expected for a long time and will make happy a lot of people :)
I was testing it, and on our cloud I can still see groupfolders under the Personal Files view, which I think should not be there 🙃
image

_Originally posted by @maximelehericy in https://github.com/nextcloud/server/issues/42919#issuecomment-2006968516_

emoral435:
* Resolves: #44359

## Summary
Previously, group folders where not being filtered out through the personal files view.
With this change, this cleaned up the code a bit and now correctly filters what it did previously, and now group folders :)

Check the screen recording below to see it in effect!

https://github.com/nextcloud/server/assets/110193237/2bec269a-ec01-42f7-a8a8-093a2ba3c686

## Checklist


susnux (approved):
makes sense!

Altahrim (approved):

apps/files/src/services/PersonalFiles.ts
```diff
@@ -36,19 +36,22 @@ const currUserID = getCurrentUser()?.uid
...
  * @return {Boolean}
  */
-export const personalFile = function(node: File): Boolean {
-	const isNotShared = currUserID ? node.owner === currUserID : true
-						&& node.attributes['mount-type'] !== 'group'
-						&& node.attributes['mount-type'] !== 'shared'
-	return isNotShared
+export const isPersonalFile = function(node: File): Boolean {
+	// the type of mounts that determine whether the file is shared
+	const sharedMountTypes = ["group", "shared"]
```
artonge, line 41: Shouldn't it be any kind of mount point? Are external storage excluded?

emoral435: Perhaps. Is it not possible that these external storage's containing files still belong solely to the user? These mount types are more defined to be shared, so I think its best to leave it to these. It can change easily in the future though.

artonge: True, but they are still not personal files as "files in your personal folder".
Maybe @nextcloud/designers can give us their opinion

PR #44457, opened 2024-03-25
fix(contactsinteraction): Allow vCard download

ChristophWurst:
## Summary

Sabre calculates a header for the size of a vcard, therefore we have to implement the size method.

## How to test

1. Share a file to an email address
2. Open Contacts and find a generated contact for the email address
3. Open the contact details
4. Click on the actions menu in the top right
5. Click *Download*

Master: nothing happens (URL is opened in a new tab and kind of silently fails).
Here: vcard is downloaded.

## Checklist

### DIFF
--- a/apps/contactsinteraction/lib/Card.php
+++ b/apps/contactsinteraction/lib/Card.php
@@ -96,7 +96,7 @@ class Card implements ICard, IACL {
 	 * @inheritDoc
 	 */
 	public function getSize(): int {
-		throw new NotImplemented();
+		return strlen($this->contact->getCard());
 	}
 
 	/**


miaulalala (approved):
maxence

apps/contactsinteraction/lib/Card.php
```diff
@@ -96,7 +96,7 @@ public function getETag(): ?string {
 	 * @inheritDoc
 	 */
 	public function getSize(): int {
-		throw new NotImplemented();
+		return mb_strlen($this->contact->getCard());
```
Altahrim, line 99: Why not simple `strlen`? It's supposed to be in bytes, no?

ChristophWurst: I looked at `\OCA\Deck\DAV\CalendarObject::getSize` and it uses mb too

ChristophWurst: https://stackoverflow.com/questions/7568949/measure-string-size-in-bytes-in-php

Altahrim: Since the overload have been removed since PHP 8.0, I think we can safely use `strlen` which should be more straightforward

Altahrim (approved):

Issue #44032, opened 2024-03-06
[Bug]: Sharing menu state not updated after "Create a new share link" with "Always ask for a password" admin setting


andrey18106:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

With Nextcloud 27.1.7 when the "Admin settings -> Sharing -> Always ask for a password" there is an incorrect sidebar sharing menu UI update after you "Create a new share link".

**Note**: After page reload it's displayed correctly.

### Steps to reproduce

1. Tick the Admin settings -> Sharing -> Always ask for a password checkbox
2. Try to create a new link share on any file
![1](https://github.com/nextcloud/server/assets/29692256/ca055631-4276-465f-be9d-a263c96cfdc9)

3. See the issue, after creating a link share sidebar menu UI does not refresh the state with created link share
![2](https://github.com/nextcloud/server/assets/29692256/d42b174e-eff3-4872-8c09-b9182a5ce866)

4. It's displayed correctly only after page reload

### Expected behavior

The sharing sidebar menu state is correctly reflected and refreshed after link share is created.

### Installation method

None

### Nextcloud Server version

27

### Operating system

None

### PHP engine version

PHP 8.2

### Web server

None

### Database engine version

None

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

DatNoHand:
I can confirm that we have this issue aswell with 28.0.2

andrey18106:
> I can confirm that we have this issue aswell with 28.0.2

With 28.0.3 it works fine

sorbaugh:
[...]

PR #44464, opened 2024-03-25
[stable27] 44032 fix show new shares without refresh

nfebe:
Resolves : https://github.com/nextcloud/server/issues/44032

#### Before

https://github.com/nextcloud/server/assets/14317775/565747a6-4e84-448f-9730-d08ffe606f72 

#### After

https://github.com/nextcloud/server/assets/14317775/aca51792-fad2-47cb-ac03-ccd460a02c0a


apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -57,8 +57,7 @@
 			class="sharing-entry__actions"
 			:aria-label="actionsTooltip"
 			menu-align="right"
-			:open.sync="open"
-			@close="onCancel">
```
artonge, line 61: Is the pending share removed if the user clicks away?

Fenn-CS: We explicitly call cancel, when the cancel button is pressed, if click-outside, it closes the menu without doing anything, new share processes starts with a new share object.

artonge: If you create a second public link, then clicking away do not remove the pending share. It does when clicking "Cancel". I think clicking away should have the same behavior as canceling.

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -407,6 +408,29 @@ export default {
 			return this.config.isDefaultExpireDateEnforced && this.share && !this.share.id
 		},
 
+		sharePolicyHasRequiredProperties() {
+			return this.config.enforcePasswordForPublicLink || this.config.isDefaultExpireDateEnforced
+		},
+
+		requiredPropertiesMissing() {
+			// Destructure share and config from this
+			const { share, config } = this
```
susnux, line 417: Why is this needed? Never saw this before in our code base, so maybe just directly access this.share

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -407,6 +408,29 @@ export default {
...
+			// Ensure share exist and the share policy has required properties
+			if (!this.sharePolicyHasRequiredProperties) return false
+
+			if (share) {
+				// If share has ID, then this is an incoming link share created from the existing link share
+				// Hence assume required properties
+				if (share.id) return true
```
susnux, line 424:
```suggestion
				if (share.id) {
					return true
				}
```

We never use short syntax.

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -407,6 +408,29 @@ export default {
...
+				// Check if either password or expiration date is missing and enforced
+				const isPasswordMissing = config.enforcePasswordForPublicLink && (share.password === null || share.password === undefined)
```
susnux, line 426: Can we not use
```suggestion
				const isPasswordMissing = config.enforcePasswordForPublicLink && !share.password
```

?

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -483,6 +507,7 @@ export default {
 		 * Create a new share link and append it to the list
 		 */
 		async onNewLinkShare() {
+			console.debug('onNewLinkShare called (this.share status)', this.share)
```
susnux, line 510: Please use the logger. The logger respects the configured logging level. 

```suggestion
			this.logger.debug('onNewLinkShare called (this.share status)', { share: this.share })
```

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -497,28 +522,13 @@ export default {
 				shareDefaults.expiration = this.formatDateToString(this.config.defaultExpirationDate)
 			}
 
+			console.debug('Missing required properties?', this.requiredPropertiesMissing)
```
susnux, line 525: Same

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -540,8 +550,33 @@ export default {
...
+					// if the share is valid, create it on the server
+					if (this.checkShare(this.share)) {
+						try {
+							this.logger.info('Sending existing share to server', this.share)
+							await this.pushNewLinkShare(this.share, true)
+							this.shareCreationComplete = true
+							this.logger.info('Share created on server', this.share)
+						} catch (e) {
+							this.pending = false
+							console.error(e)
```
susnux, line 565: Same (logger)

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -540,8 +550,33 @@ export default {
...
+							this.logger.error('Error creating share', e)
+							return false
+						}
+						return true
+					} else {
+						this.open = true
+						OC.Notification.showTemporary(t('files_sharing', 'Error, please enter proper password and/or expiration date'))
```
susnux, line 572: Better use `showError`

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -752,25 +791,23 @@ export default {
...
+				overflow: hidden;
+				white-space: nowrap;
+			}
 		}
-	  }
-
-	  &__copy {
 
-	  }
+		&__copy {}
```
susnux, line 810: Empty rule?

susnux (commented):
Logically makes sense, some code comments.

apps/files_sharing/src/components/SharingEntryLink.vue
Fenn-CS: The idea was to shorten these lines a bit

```js
const isPasswordMissing = config.enforcePasswordForPublicLink && (share.password === null || share.password === undefined)
const isExpireDateMissing = config.isDefaultExpireDateEnforced && (share.expireDate === null || share.expireDate === undefined)
```

apps/files_sharing/src/components/SharingEntryLink.vue
Fenn-CS: Lint fix formatting, was probably already emtpy

apps/files_sharing/src/components/SharingEntryLink.vue
Fenn-CS: The shortened version you proposed below, would suffice, so can take this out.

susnux (approved):
Code looks good now

apps/files_sharing/src/mixins/SharesMixin.js
```diff
@@ -80,6 +82,10 @@ export default {
 			 * ! do not remove it ot you'll lose all reactivity here
 			 */
 			reactiveState: this.share?.state,
+			logger: getLoggerBuilder()
```
artonge, line 85: I think this should better be an imported service rather than in this mixin.

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -407,6 +408,31 @@ export default {
...
+					return true
+				}
+				// Check if either password or expiration date is missing and enforced
+				const isPasswordMissing = this.config.enforcePasswordForPublicLink && !this.share.password
+				const isExpireDateMissing = this.config.isDefaultExpireDateEnforced && !this.share.expireDate
+
+				return isPasswordMissing || isExpireDateMissing
+			}
+			// if no share, we can't tell if properties are missing or not so we assume properties are missing
+			return true
```
artonge, line 434: Maybe you can return early for that one.

artonge (approved):
Looks good :)

Issue #5704, opened 2017-07-13
Improve group management


ogarcia:
At this moment the group management is a little hell. Groups are sorted by number of members instead alphabetically, cannot rename groups, etc. I think that this must be improved at least with:

- [x] Sort groups alphabetically
- [x] Rename groups
- [ ] Bulk add group members
- [ ] Bulk remove group members (https://github.com/nextcloud/server/issues/384)

detm54:
A small button/switch to sort groups by either number of members, creation date (age) or alphabetically would be pretty awesome. Especially a-z to match the order of grouplist when assigning groups for new users.

Have a nice day!

MorrisJobke:
See also #384

bpcurse:
The side-menu's sorting by user count shouldn't be the default view. Especially when you have created groups in a logical naming concept, it seems wiser to display them by default sorted by name and let the user choose another view if necessary.

DJCrashdummy:
IMHO these would be very nice features to have. :+1: but the **most important feature is renaming groups**, since all other can somehow be accomplished or worked around...

so is there any way to rename a group, until this gets implemented?

Dubidubiduu:
Yes, renaming groups is very important to me too, cuz I don't want to
.. create a group
.. give access to the new group
.. add all the members to the new group
.. and then delete the old group

DJCrashdummy:
@Dubidubiduu well, this would be a tedious but still practicable workaround... but i'm **concerned about the existing group-shares** because **with this workaround they will be lost**!

and for this case (keeping group-shares) there is no workaround i know; hence the question, if there is any chance to rename a group for now (via manipulating the database, etc.).

fredt34:
C'mon, how can it be soooo difficult to rename a group??

DjSni:
For the group administrator, it should also be possible to add existing users to his group(s).

fenvarien:
**Renaming groups** is still a much needed feature. Is it currently in development? I'm an admin for a nextcloud installation in a school. We need the ability to rename a class from eg "7a" to "8a" with the beginning of thr next school year. But we want to keep the students in that group/class.

SimJoSt:
@fenvarien there is a separate issue for this: https://github.com/nextcloud/server/issues/9209 :)

blackerking:
At least the sorting function would be an useful utility.

cl4x:
[...]

PR #44465, opened 2024-03-25
feat(settings): Allow to sort groups in the account management alphabetically

### LINKED ISSUE #39528: Feature Request : Sorting groups
In the user management page, sort groups alphabetically instead of sorting by number of members

susnux:
* For: https://github.com/nextcloud/server/issues/5704
* Resolves: https://github.com/nextcloud/server/issues/39528

## Summary

We can do this purely in the frontend - but when enforced from the backend using the existing system config, we need to follow the requirement. We then show a warning about the configuration.

## Screenshots

what | screenshot
---|---
Groups sorted alphabetically | Screenshot 2024-03-25 at 23-43-45 Settings - Users - Nextcloud
Sorting by name | Screenshot 2024-03-25 at 23-43-37 Settings - Users - Nextcloud
Sorting by member count | Screenshot 2024-03-25 at 23-43-25 Settings - Users - Nextcloud
When enforced by name without member count available | Screenshot 2024-03-25 at 23-43-03 Settings - Users - Nextcloud

## Checklist


apps/settings/src/store/users.js
```diff
@@ -254,6 +238,19 @@ const mutations = {
...
+		axios.post(
+			generateUrl('/settings/users/preferences/group.sortBy'),
+			{
+				value: String(sorting),
+			},
+		).catch((error) => {
+			showError(t('settings', 'Could not set group sorting'))
+			logger.error(error)
+		})
+		state.orderBy = sorting
```
emoral435, line 252: if there was an error caught due to the passed in sorting argument, do we still want to set `state.orderBy = sorting`?

emoral435 (commented):
Looks fantastic, just a small comment :)

apps/settings/src/store/users.js
susnux: makes sense!

emoral435 (approved):
Sweet!

apps/settings/lib/Controller/UsersController.php
```diff
@@ -124,8 +124,9 @@ public function usersList(): TemplateResponse {
 
 		/* SORT OPTION: SORT_USERCOUNT or SORT_GROUPNAME */
 		$sortGroupsBy = \OC\Group\MetaData::SORT_USERCOUNT;
+		$forceSortGroupByName = $this->config->getSystemValue('sort_groups_by_name', false);
```
come-nc, line 127:
```suggestion
		$forceSortGroupByName = $this->config->getSystemValueBool('sort_groups_by_name', false);
```

come-nc (commented):
Was this tested with LDAP backend? I think LDAP backend disabled user counts in some cases (always?).

susnux:
> Was this tested with LDAP backend?

I did not change the backend logic, I do not touch the group fetching, but simply add an option for front end sorting.
I just use the existing flag to overwrite the front end sorting to keep current behavior.

So yes works the same with LDAP as before.

apps/settings/lib/Controller/UsersController.php
susnux: done

apps/settings/lib/Controller/UsersController.php
```diff
@@ -212,13 +212,19 @@ public function usersList(): TemplateResponse {
 		/* LANGUAGES */
 		$languages = $this->l10nFactory->getLanguages();
 
+		/** Using LDAP or admins (system config) can enfore sorting by group name, in this case the fronend setting is overwritten */
```
Pytal, line 215:
```suggestion
		/** Using LDAP or admins (system config) can enfore sorting by group name, in this case the frontend setting is overwritten */
```

apps/settings/lib/Controller/UsersController.php
```diff
@@ -212,13 +212,19 @@ public function usersList(): TemplateResponse {
...
 		/* FINAL DATA */
 		$serverData = [];
 		// groups
 		$serverData['groups'] = array_merge_recursive($adminGroup, [$disabledUsersGroup], $groups);
 		// Various data
 		$serverData['isAdmin'] = $isAdmin;
-		$serverData['sortGroups'] = $sortGroupsBy;
+		$serverData['sortGroups'] = $forceSortGroupByName
+			? \OC\Group\MetaData::SORT_GROUPNAME
+			: (int)$this->config->getAppValue('core', 'group.sortBy', strval(\OC\Group\MetaData::SORT_USERCOUNT));
```
Pytal, line 226:
```suggestion
			: (int)$this->config->getAppValue('core', 'group.sortBy', (string)\OC\Group\MetaData::SORT_USERCOUNT);
```
**nit (non-blocking):** Since we cast with (int) on the same line

Pytal (approved):
Nice!

PR #44470, opened 2024-03-26
fix: Allow using replica config with split databases

juliusknorr:
Otherwise using [Replication](https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/replication.html) with [Splitting databases](https://docs.nextcloud.com/server/latest/admin_manual/configuration_database/splitting.html) would be incompatible as the split database would pick up the replica of the regular database.

With this change there is also the possibility to configure dedicated replicas for each config prefix of a database.

### DIFF
--- a/lib/private/DB/ConnectionFactory.php
+++ b/lib/private/DB/ConnectionFactory.php
@@ -233,7 +233,7 @@ class ConnectionFactory {
 			$connectionParams['persistent'] = true;
 		}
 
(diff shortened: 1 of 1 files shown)


nickvergessen (approved):

ChristophWurst (approved):

schiessle:
can we than also update the documentation with a example? Thanks a lot!

PR #44483, opened 2024-03-26
fix(settings): Also verify that `trusted_proxies` only contains IP addresses (with range)

susnux:
## Summary

`trusted_proxies` only handles IPv4 and IPv6 addresses, optionally with CIDR range notation, so we should validate this too. As it is a common mistake to put host names into this.

## Checklist

### DIFF
--- a/apps/settings/lib/SetupChecks/ForwardedForHeaders.php
+++ b/apps/settings/lib/SetupChecks/ForwardedForHeaders.php
@@ -59,6 +59,16 @@ class ForwardedForHeaders implements ISetupCheck {
 			return SetupResult::error($this->l10n->t('Your "trusted_proxies" setting is not correctly set, it should be an array.'));
 		}
 
+		foreach ($trustedProxies as $proxy) {
+			$addressParts = explode('/', $proxy, 2);
+			if (filter_var($addressParts[0], FILTER_VALIDATE_IP) === false || !ctype_digit($addressParts[1] ?? '24')) {
+				return SetupResult::error(
+					$this->l10n->t('Your "trusted_proxies" setting is not correctly set, it should be an array of IP addresses - optionally with range in CIDR notation.'),
+					$this->urlGenerator->linkToDocs('admin-reverse-proxy'),
+				);
+			}
+		}
+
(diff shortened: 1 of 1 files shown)


apps/settings/lib/SetupChecks/ForwardedForHeaders.php
```diff
@@ -59,6 +59,22 @@
...
 
+		foreach ($trustedProxies as $proxy) {
+			$addressParts = explode('/', $proxy, 2);
+			if (filter_var($addressParts[0], FILTER_VALIDATE_IP) === false) {
+				return SetupResult::error(
+					$this->l10n->t('Your "trusted_proxies" setting is not correctly set, it should be an array of IP addresses - optionally with range in CIDR notation.'),
+					$this->urlGenerator->linkToDocs('admin-reverse-proxy'),
+				);
+			}
+			if (count($addressParts) === 2 && !ctype_digit($addressParts[])) {
```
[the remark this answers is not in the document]
susnux: Yes! Wrong state staged and committed!

apps/settings/lib/SetupChecks/ForwardedForHeaders.php
```diff
@@ -59,6 +59,22 @@ public function run(): SetupResult {
 			return SetupResult::error($this->l10n->t('Your "trusted_proxies" setting is not correctly set, it should be an array.'));
 		}
 
+		foreach ($trustedProxies as $proxy) {
+			$addressParts = explode('/', $proxy, 2);
+			if (filter_var($addressParts[0], FILTER_VALIDATE_IP) === false) {
```
come-nc, line 64:
```suggestion
			if (filter_var($addressParts[0], FILTER_VALIDATE_IP) === false || !ctype_digit($addressParts[1] ?? 24)) {
```
This would avoid duplicating the if inner part.

Altahrim (approved):

apps/settings/lib/SetupChecks/ForwardedForHeaders.php
susnux: Looks cleaner! Thank you for the input :)

come-nc (approved):
Extra points if you add a unit test.

PR #44485, opened 2024-03-26
Respect empty `expiryDate` value in server

### LINKED ISSUE #44219: [Bug]: When creating email share expiry date is set by default without enforcement

### Bug description

When creating an email share with the option "Set expiration date" unchecked and "Enforce expiry date" unchecked, the default expiry date is set anyway.
You can see it by clicking "customize link".

This is how it looks like when creating the email share:
grafik

This is how it looks like when customizing the share afterwards:
grafik

After unchecking "Set expiration date" and updating share again, the values stays ok.

### Steps to reproduce

1. Configure the Sharing options like this.
    - [x] Allow users to share via link end emails
    - [x] Always ask for a password
    - [x] Enforce password protection
    - [x] Set default expiry date -> Expire after 30 day(s)
    - [ ] Enforce expiry date
2. Create an email share and uncheck the option "Set expiration date"
3. Click on "customize link" and see that it's wrong.
4. Uncheck "Set expiration date" again and update the link.
5. Click on "customize link" again and see that now it worked how it should work in first place.

### Expected behavior

If the option "Set expiration date" is unchecked when creating the share, it should be stored correctly in the database like this and not only at the second updating

### Nextcloud Server version

27.1.7 28.0.3

nfebe:
If `expireDate` is an empty string and not `null` then the server should not set a default.

Resolves : https://github.com/nextcloud/server/issues/44219

### DIFF
--- a/apps/files_sharing/lib/Controller/ShareAPIController.php
+++ b/apps/files_sharing/lib/Controller/ShareAPIController.php
@@ -549,7 +549,8 @@ class ShareAPIController extends OCSController {
 	 * @param string $publicUpload If public uploading is allowed
 	 * @param string $password Password for the share
 	 * @param string|null $sendPasswordByTalk Send the password for the share over Talk
-	 * @param string $expireDate Expiry date of the share using user timezone at 00:00. It means date in UTC timezone will be used.
+	 * @param ?string $expireDate The expiry date of the share in the user's timezone (UTC) at 00:00.
+	 *                If $expireDate is not supplied or set to `null`, the system default will be used.
 	 * @param string $note Note for the share
 	 * @param string $label Label for the share (only used in link and email)
 	 * @param string|null $attributes Additional attributes for the share
@@ -571,7 +572,7 @@ class ShareAPIController extends OCSController {
 		string $publicUpload = 'false',
 		string $password = '',
 		?string $sendPasswordByTalk = null,
-		string $expireDate = '',
+		?string $expireDate = null,
 		string $note = '',
 		string $label = '',
 		?string $attributes = null
@@ -646,12 +647,18 @@ class ShareAPIController extends OCSController {
 		}
 
 		//Expire date
-		if ($expireDate !== '') {
-			try {
-				$expireDateTime = $this->parseDate($expireDate);
-				$share->setExpirationDate($expireDateTime);
-			} catch (\Exception $e) {
-				throw new OCSNotFoundException($this->l->t('Invalid date, date format must be YYYY-MM-DD'));
+		if ($expireDate !== null) {
+			if ($expireDate !== '') {
+				try {
+					$expireDateTime = $this->parseDate($expireDate);
+					$share->setExpirationDate($expireDateTime);
+				} catch (\Exception $e) {
+					throw new OCSNotFoundException($this->l->t('Invalid date, date format must be YYYY-MM-DD'));
+				}
+			} else {
+				// Client sent empty string for expire date.
+				// Set noExpirationDate to true so overwrite is prevented.
+				$share->setNoExpirationDate(true);
 			}
 		}
 
--- a/build/integration/sharing_features/sharing-v1.feature
+++ b/build/integration/sharing_features/sharing-v1.feature
@@ -229,6 +229,24 @@ Feature: sharing
       | url | AN_URL |
       | mimetype | httpd/unix-directory |
 
+  Scenario: Creating a new share with expiration date removed, when default expiration is set
+    Given user "user0" exists
+    And user "user1" exists
+    And parameter "shareapi_default_expire_date" of app "core" is set to "yes"
+    And As an "user0"
+    When creating a share with
+      | path | welcome.txt |
+      | shareWith | user1 |
+      | shareType | 0 |
+      | expireDate | |
+    Then the OCS status code should be "100"
+    And the HTTP status code should be "200"
+    And Getting info of last share
+    Then the OCS status code should be "100"
+    And the HTTP status code should be "200"
+    And Share fields of last share match with
+      | expiration ||
+
   Scenario: Creating a new public share, updating its password and getting its info
     Given user "user0" exists
     And As an "user0"
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -348,26 +348,6 @@ class Manager implements IManager {
 
 		$expirationDate = $share->getExpirationDate();
 
-		if ($expirationDate !== null) {
-			$expirationDate->setTimezone($this->dateTimeZone->getTimeZone());
-			$expirationDate->setTime(0, 0, 0);
-
-			$date = new \DateTime('now', $this->dateTimeZone->getTimeZone());
-			$date->setTime(0, 0, 0);
-			if ($date >= $expirationDate) {
-				$message = $this->l->t('Expiration date is in the past');
-				throw new GenericShareException($message, $message, 404);
-			}
-		}
-
-		// If expiredate is empty set a default one if there is a default
-		$fullId = null;
-		try {
-			$fullId = $share->getFullId();
-		} catch (\UnexpectedValueException $e) {
-			// This is a new share
-		}
-
 		if ($isRemote) {
 			$defaultExpireDate = $this->shareApiRemoteDefaultExpireDate();
 			$defaultExpireDays = $this->shareApiRemoteDefaultExpireDays();
@@ -379,28 +359,53 @@ class Manager implements IManager {
 			$configProp = 'internal_defaultExpDays';
 			$isEnforced = $this->shareApiInternalDefaultExpireDateEnforced();
 		}
-		if ($fullId === null && $expirationDate === null && $defaultExpireDate) {
-			$expirationDate = new \DateTime('now', $this->dateTimeZone->getTimeZone());
-			$expirationDate->setTime(0, 0, 0);
-			$days = (int)$this->config->getAppValue('core', $configProp, (string)$defaultExpireDays);
-			if ($days > $defaultExpireDays) {
-				$days = $defaultExpireDays;
-			}
-			$expirationDate->add(new \DateInterval('P' . $days . 'D'));
-		}
 
-		// If we enforce the expiration date check that is does not exceed
-		if ($isEnforced) {
-			if ($expirationDate === null) {
-				throw new \InvalidArgumentException('Expiration date is enforced');
+		// If $expirationDate is falsy, noExpirationDate is true and expiration not enforced
+		// Then skip expiration date validation as null is accepted
+		if(!($share->getNoExpirationDate() && !$isEnforced)) {
+			if ($expirationDate != null) {
+				$expirationDate->setTimezone($this->dateTimeZone->getTimeZone());
+				$expirationDate->setTime(0, 0, 0);
+
+				$date = new \DateTime('now', $this->dateTimeZone->getTimeZone());
+				$date->setTime(0, 0, 0);
+				if ($date >= $expirationDate) {
+					$message = $this->l->t('Expiration date is in the past');
+					throw new GenericShareException($message, $message, 404);
+				}
 			}
 
-			$date = new \DateTime('now', $this->dateTimeZone->getTimeZone());
-			$date->setTime(0, 0, 0);
-			$date->add(new \DateInterval('P' . $defaultExpireDays . 'D'));
-			if ($date < $expirationDate) {
-				$message = $this->l->n('Cannot set expiration date more than %n day in the future', 'Cannot set expiration date more than %n days in the future', $defaultExpireDays);
-				throw new GenericShareException($message, $message, 404);
+			// If expiredate is empty set a default one if there is a default
+			$fullId = null;
+			try {
+				$fullId = $share->getFullId();
+			} catch (\UnexpectedValueException $e) {
+				// This is a new share
+			}
+
+			if ($fullId === null && $expirationDate === null && $defaultExpireDate) {
+				$expirationDate = new \DateTime('now', $this->dateTimeZone->getTimeZone());
+				$expirationDate->setTime(0, 0, 0);
+				$days = (int)$this->config->getAppValue('core', $configProp, (string)$defaultExpireDays);
+				if ($days > $defaultExpireDays) {
+					$days = $defaultExpireDays;
+				}
+				$expirationDate->add(new \DateInterval('P' . $days . 'D'));
+			}
+
+			// If we enforce the expiration date check that is does not exceed
+			if ($isEnforced) {
+				if (empty($expirationDate)) {
+					throw new \InvalidArgumentException('Expiration date is enforced');
+				}
+
+				$date = new \DateTime('now', $this->dateTimeZone->getTimeZone());
+				$date->setTime(0, 0, 0);
+				$date->add(new \DateInterval('P' . $defaultExpireDays . 'D'));
+				if ($date < $expirationDate) {
+					$message = $this->l->n('Cannot set expiration date more than %n day in the future', 'Cannot set expiration date more than %n days in the future', $defaultExpireDays);
+					throw new GenericShareException($message, $message, 404);
+				}
 			}
 		}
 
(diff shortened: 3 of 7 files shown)


Fenn-CS:
Related changes : https://github.com/nextcloud/server/pull/44912

lib/private/Share20/Manager.php
```diff
@@ -487,25 +417,6 @@ protected function validateExpirationDateLink(IShare $share) {
 			}
 		}
 
-		// If expiredate is empty set a default one if there is a default
```
nickvergessen, line 490: but you are removing the default functionality?

Fenn-CS: Yes :)... Since the users now have the opportunity to modify this settings before a share is created. `i.e` the user can see on the share creation menu, that an expiry date is required and can either allow the default or remove if not enforced.

Leaving this default functionality means, we always force shares to have expiry dates even when that's not the users intention.

nickvergessen: The problem is that currently existing clients and app might rely on the default and not populate it via the UI or send it with each request.
So this is a breaking change that is impossible to discover and will lead to a lot of shares not expiring although they should.

Fenn-CS: I see. This breaking change is inevitable if we would solve https://github.com/nextcloud/server/issues/44219 though.

I am not sure but perhaps I could introduce checks for clients/app calls.

Another though is that `Manager` class is only part of the private API and maybe all we need to do is make sure the apps we maintain that uses the non public share interface are updated?

nickvergessen: > Another though is that Manager class is only part of the private API and maybe all we need to do is make sure the apps we maintain that uses the non public share interface are updated?

I meant apps like Android, iOS and our Desktop Client:
- https://github.com/nextcloud/server/blob/490f545006024ad982c801b5c986ff8d03826821/apps/files_sharing/lib/Controller/ShareAPIController.php#L574
- https://github.com/nextcloud/server/blob/490f545006024ad982c801b5c986ff8d03826821/apps/files_sharing/lib/Controller/ShareAPIController.php#L648-L656

That being said, the update also supports it, but basically you need to set it to `''` (empty string) not null when updating a share and removing a expiration date.

Seems to work when updating, just not when creating, as the check is always executed then.
So instead of tampering with `validateExpirationDateInternal` like this, I would try to get the information into the share object that no expirationDate was sent and in https://github.com/nextcloud/server/blob/ec5133b739eabc76271789504b4dbb91a534f552/lib/private/Share20/Manager.php#L423 instead of only checking `$expirationDate === null` you also check whether it was sent with the empty string, and only set the default date when it was not sent as empty string.

nfebe (commented):
Backtracked on the initial radical and breaking approach where, we allow defaults to be set entire by the client.

Propose new approach: 

- If client sends empty string, then do not set any defaults
- Set defaults it no `expireDate` is sent
- Set default if `expireDate` sent is null

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -646,12 +646,16 @@ public function createShare(
...
+		if ($expireDate !== null) {
+			if ($expireDate !== '') {
+				try {
+					$expireDateTime = $this->parseDate($expireDate);
+					$share->setExpirationDate($expireDateTime);
+				} catch (\Exception $e) {
+					throw new OCSNotFoundException($this->l->t('Invalid date, date format must be YYYY-MM-DD'));
+				}
+			} else {
+				$share->setExpirationDate('');
```
Fenn-CS, line 658: Since, `null` is the basis for setting default expiration date in the share manager for new shares, we just can't set this to null here.

Fenn-CS, line 658:
```suggestion
				$share->setExpirationDate(null);
```

Since, `null` is the basis for setting default expiration date in the share manager for new shares, we just can't set this to null here.


But can be done much later in `validateExpirationDateInternal` and `validateExpirationDateLink`

lib/public/Share/IShare.php
```diff
@@ -385,11 +385,11 @@ public function getNote();
 	/**
 	 * Set the expiration date
 	 *
-	 * @param null|\DateTime $expireDate
+	 * @param \DateTime|string|null $expireDate
```
provokateurin, line 388:
```suggestion
	 * @param \DateTime|''|null $expireDate
```
To avoid any other string than the empty string. I'm questioning a bit if this couldn't be solved differently because two "null" values are not great

lib/private/Share20/Manager.php
```diff
@@ -389,6 +337,13 @@ protected function validateExpirationDateInternal(IShare $share) {
 
 		$expirationDate = $share->getExpirationDate();
 
+		// If $expirationDate is an empty string, then the client unset expiration  date
+		// Hence no validation required
```
nickvergessen, line 341: This is not correct. If the enforcement is enabled you still have to.

lib/private/Share20/Manager.php
```diff
@@ -475,6 +430,13 @@ protected function validateExpirationDateInternal(IShare $share) {
 	protected function validateExpirationDateLink(IShare $share) {
 		$expirationDate = $share->getExpirationDate();
 
+		// If $expirationDate is an empty string, then the client unset expiration  date
+		// Hence no validation required
```
nickvergessen, line 434: Same here

lib/public/Share/IShare.php
nickvergessen: Which is why i would have added another prop for this. Also helps to improve the enforcement handling above

nickvergessen (dismissed):

lib/private/Share20/Share.php
```diff
@@ -41,61 +41,34 @@
...
+	private ?IAttributes $attributes = null;
+	private ?int $status = null;
+	private ?string $note = null;
+	private ?\DateTime $expireDate = null;
+	private bool $overwriteFalsyExpirationDate = true;
+	private ?string $password = null;
 	private ?\DateTimeInterface $passwordExpirationTime = null;
-	/** @var bool */
-	private $sendPasswordByTalk = false;
-	/** @var string */
```
come-nc, line 81: You’re gonna get trouble with all this strong typing. I tried it in the psalm update PR and typing in this class is inconsistent between props and methods. Main problem is the class can be used to represent an incomplete share while some code path expect it to be complete.

I’d advice not touching the typing or do it in a separate PR which fixes the problems.

come-nc (commented):
I have to say I do not understand why the share object has to carry the information of whether a default date was used, it should simply contain the date. So this new method in IShare is not logical to me.

Code around this sounds quite convoluted.

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -646,12 +646,17 @@ public function createShare(
...
-				$share->setExpirationDate($expireDateTime);
-			} catch (\Exception $e) {
-				throw new OCSNotFoundException($this->l->t('Invalid date, date format must be YYYY-MM-DD'));
+		if ($expireDate !== null) {
+			if ($expireDate !== '') {
+				try {
+					$expireDateTime = $this->parseDate($expireDate);
+					$share->setExpirationDate($expireDateTime);
+				} catch (\Exception $e) {
+					throw new OCSNotFoundException($this->l->t('Invalid date, date format must be YYYY-MM-DD'));
```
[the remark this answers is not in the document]
come-nc: `	 * @throws OCSNotFoundException Creating the share failed
`
Apparently not :thinking: 
This API is really weird.

come-nc:
> I have to say I do not understand why the share object has to carry the information of whether a default date was used, it should simply contain the date. So this new method in IShare is not logical to me.
> 
> Code around this sounds quite convoluted.

Ok I’m starting to understand, the share object is used both to convey the parameter for creation and the final created object, this is quite bad.
So I do not see a better way than what you did for storing the info without a bigger refactor.
I do not understand the name `setOverwriteFalsyExpirationDate` though.
Why not `setNoExpirationDate(bool)` ?

lib/public/Share/IShare.php
```diff
@@ -385,20 +385,38 @@ public function getNote();
...
 	public function getExpirationDate();
 
+	/**
+	 * Set overwrite flag for falsy expiry date vavlues
+	 *
+	 * @param bool $overwriteFalsyExpirationDate
+	 * @return \OCP\Share\IShare The modified object
+	 * @since 27.0.0
+	 */
+	public function setOverwriteFalsyExpirationDate(bool $overwriteFalsyExpirationDate);
```
nickvergessen, line 409:
```suggestion
	public function setNoExpirationDate(bool $noExpirationDate);
```
etc

Fenn-CS:
> I do not understand the name setOverwriteFalsyExpirationDate though.
> Why not setNoExpirationDate(bool) ?

The method name `setOverwriteFalsyExpirationDate` was chosen for its explicitness in indicating that the method controls whether to overwrite falsy values, such as `null` or an empty string (`''`), in the `expireDate` field. 

Because, that is exactly what the method and its corresponding property is currently used for in the code, we use it to decide if we want to **overwrite** the value of `$expireDate` when it has *falsy* values (`null`, `''`, `0'`, etc) ...

However, following your suggestion, I have changed the name to `setNoExpirationDate(bool)` as it might be easier especially for non-native English speakers to grab.

lib/private/Share20/Manager.php
```diff
@@ -474,6 +429,13 @@ protected function validateExpirationDateInternal(IShare $share) {
 	 */
 	protected function validateExpirationDateLink(IShare $share) {
```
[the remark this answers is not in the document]
Fenn-CS: Yes the method is in use.

lib/private/Share20/Manager.php
```diff
@@ -474,6 +429,13 @@ protected function validateExpirationDateInternal(IShare $share) {
...
 		$expirationDate = $share->getExpirationDate();
+		$isEnforced = $this->shareApiLinkDefaultExpireDateEnforced();
+
+		// If $expirationDate is falsy, overwrite flag is false, no expiration is set
+		if(empty($expirationDate) && !$share->getOverwriteFalsyExpirationDate() && !$isEnforced) {
+			$share->setExpirationDate(null);
```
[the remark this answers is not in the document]
Fenn-CS: Yes because `''` (empty string) value would lead to `DBAL` exceptions when trying to convert a string into value that should be stored on a `nullable` `datetime` field.

lib/private/Share20/Share.php
Fenn-CS: Moved to : https://github.com/nextcloud/server/pull/45156

lib/private/Share20/Manager.php
```diff
@@ -409,17 +375,6 @@ protected function validateExpirationDateInternal(IShare $share) {
...
-			$isEnforced = $this->shareApiRemoteDefaultExpireDateEnforced();
-		} else {
-			$defaultExpireDate = $this->shareApiInternalDefaultExpireDate();
-			$defaultExpireDays = $this->shareApiInternalDefaultExpireDays();
-			$configProp = 'internal_defaultExpDays';
-			$isEnforced = $this->shareApiInternalDefaultExpireDateEnforced();
-		}
 		if ($fullId === null && $expirationDate === null && $defaultExpireDate) {
 			$expirationDate = new \DateTime('now', $this->dateTimeZone->getTimeZone());
 			$expirationDate->setTime(0, 0, 0);
```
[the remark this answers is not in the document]
Fenn-CS: @come-nc  not clear here please...

apps/files_sharing/lib/Controller/ShareAPIController.php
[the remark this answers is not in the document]
Fenn-CS: To avoid more back and forth for something unrelated moved to : https://github.com/nextcloud/server/pull/45157

lib/private/Share20/Manager.php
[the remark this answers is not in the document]
Fenn-CS: Figured.

nickvergessen:
Would really be good to add a new test into https://github.com/nextcloud/server/tree/058b69eef02be9bfa77e9b6a1d12d15f29ac6b0c/build/integration/sharing_features which proves the thing you fixed, so it does not get reverted at some point and no test fails :P

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -549,7 +549,8 @@ public function deleteShare(string $id): DataResponse {
 	 * @param string $publicUpload If public uploading is allowed
 	 * @param string $password Password for the share
 	 * @param string|null $sendPasswordByTalk Send the password for the share over Talk
-	 * @param string $expireDate Expiry date of the share using user timezone at 00:00. It means date in UTC timezone will be used.
+	 * @param ?string $expireDate The expiry date of the share in the user's timezone (UTC) at 00:00.
```
come-nc, line 552:
```suggestion
	 * @param ?string $expireDate The expiry date of the share in the user's timezone at 00:00.
```
The user’s timezone is not always UTC

lib/private/Share20/Manager.php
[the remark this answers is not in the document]
come-nc: But expirationDate is null or DateTime it cannot be `''` in the share object.

come-nc:
I agree that we need tests for this.

blizzz:
Is it possible to get some momentum into this PR? With regard to Red CI and the suggested test?

build/integration/sharing_features/sharing-v1.feature
```diff
@@ -229,6 +229,23 @@ Feature: sharing
       | url | AN_URL |
       | mimetype | httpd/unix-directory |
 
+  Scenario: Creating a new share with expiration date removed, when default expiration is set
```
Fenn-CS, line 232: Would fix up commit to main changes once tests are approved

nfebe (commented):
WIP commit present.

lib/private/Share20/Manager.php
```diff
@@ -379,28 +307,53 @@ protected function validateExpirationDateInternal(IShare $share) {
...
-		if ($fullId === null && $expirationDate === null && $defaultExpireDate) {
-			$expirationDate = new \DateTime('now', $this->dateTimeZone->getTimeZone());
-			$expirationDate->setTime(0, 0, 0);
-			$days = (int)$this->config->getAppValue('core', $configProp, (string)$defaultExpireDays);
-			if ($days > $defaultExpireDays) {
-				$days = $defaultExpireDays;
+
+		// If $expirationDate is falsy, noExpirationDate is true and expiration not enforced
+		// Then skip expiration date validation as null is accepted
+		if(!(empty($expirationDate) && $share->getNoExpirationDate() && !$isEnforced)) {
```
Fenn-CS, line 313: Does not make a difference, would update.

```suggestion
		if(!($share->getNoExpirationDate() && !$isEnforced)) {
```

nfebe (commented):
If `noExpirationDate` is set to true and no enforcement no validation is needed irresepective of the value of `expirationDate`

build/integration/sharing_features/sharing-v1.feature
```diff
@@ -229,6 +229,24 @@ Feature: sharing
...
+      | shareWith | user1 |
+      | shareType | 0 |
+      | expireDate | |
+    Then the OCS status code should be "100"
+    And the HTTP status code should be "200"
+    And Getting info of last share
+    Then the OCS status code should be "100"
+    And the HTTP status code should be "200"
+    And Share fields of last share match with
+      | expiration ||
```
Fenn-CS, line 248: @artonge In case you are wondering this is what the test suite is complaining about.

Context:

- We updated the backend to not set expiry date when an empty value is sent
- So when the expiry is created it returns expiration that should be nullable or an empty space.
- So next step is to try `| null |` currently I am passing `||` and not sure how Gherkin would parse that.

artonge (approved):

lib/public/Share/IShare.php
```diff
@@ -385,20 +385,38 @@ public function getNote();
...
 	 * @since 9.0.0
 	 */
 	public function getExpirationDate();
 
+	/**
+	 * Set overwrite flag for falsy expiry date vavlues
+	 *
+	 * @param bool $noExpirationDate
+	 * @return \OCP\Share\IShare The modified object
+	 * @since 27.0.0
```
come-nc, line 407:
```suggestion
	 * @since 29.0.0
```
And then adapted in backports

come-nc: (same for  `getNoExpirationDate`)

come-nc (approved):

ShGKme:
This breaks uploading in Talk which expects `$expireDate` to be `string`

blizzz:
> This breaks uploading in Talk which expects `$expireDate` to be `string`

Also on 27?

Antreesy:
> Also on 27?

Confirm for stable27
image

lib/public/Share/IShare.php
```diff
@@ -385,20 +385,38 @@ public function getNote();
...
+	public function setNoExpirationDate(bool $noExpirationDate);
+
+
+	/**
+	 * Get value of overwrite falsy expiry date flag
+	 *
+	 * @return bool
+	 * @since 30.0.0
+	 */
+	public function getNoExpirationDate();
```
nickvergessen, line 418: @Fenn-CS Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(all changes in lib/public/)

Fenn-CS: https://github.com/nextcloud/documentation/pull/12051

PR #44493, opened 2024-03-26
Add setting to force LDAP UUID to upper/lower case user- and group names

### LINKED ISSUE #44486: Add force lower/uppercase setting for LDAP UUIDs
Nextcloud usernames are case sensitive. Per default, the user_ldap app uses the ObjectGUID and similar attributes as the username. 

Per [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122) UUIDs are not case sensitive. This results in unexpected behavior which has already been reported, for example when combining LDAP with OIDC or SAML ( https://github.com/nextcloud/user_saml/issues/406 https://github.com/nextcloud/user_saml/issues/563 ).

Confusing the situation even more: the different handling of upper and lower case UUIDs by different LDAP servers. For example, LDAP via Samba returns uppercase ObjectGUIDs while OpenLDAP (for example via UCS) uses lowercase UUIDs. This makes migrating from one directory service to another, while keeping users in Nextcloud, very difficult. Dumping the database, replacing the UUIDs with their upper or lower case alternative, inserting the dump again and moving the user data directories in theory does work.

However, du to the fact that Nextcloud user_ldap uses UUIDs for mapping case sensitive usernames and UUIDs per definition are not case sensitive, the app should provide some kind of mechanism to deal with the issue.

The "Expert" tab of the user_ldap configuration wizard should have a new configuration option in the "Override UUID detection" section. The configuration option should let a admin choose one of 3 options:
- Accept upper/lower case as returned from directory service (default)
- Convert UUIDs to lower case
- Convert UUIDs to upper case

When migrating to a different LDAP directory, the database needs to be updated. When combining LDAP with OIDC or SAML, for example Keycloak allows to convert UUIDs to lower or uppercase with script mappers (https://github.com/netzbegruenung/keycloak-scriptmapper-ldap-id).

**Additional context**
None.

svenseeberg:
* Resolves: #44486

## Summary

Add a new expert setting for the user_ldap app to convert the LDAP user and group UUID attributes to lower or upper case.

Before I continue with writing tests, I would first like collect feedback for the proposed solution.

Screen before the change:

image

Screen after the change:

image

## TODO

- [ ] Translations
- [ ] Tests

## Checklist


svenseeberg:
As soon as I get some kind of heads up, I can open a PR for documentation. And for the tests I would be happy to get some kind of input what should be tested. A unit test that validates the different return values for the function should be easy enough.

come-nc:
Can you explain the usecase?
In the UI you put the option in the internal username but in the code you apply it to the UUID.

UUID and internal user name are two different things, even if by default UUID is used for internal user name.

svenseeberg:
> Can you explain the usecase?

Sure.

Nextcloud usernames are case sensitive. Per default, the user_ldap app uses the ObjectGUID and similar attributes as the username. 

Per [RFC4122](https://datatracker.ietf.org/doc/html/rfc4122) UUIDs are not case sensitive. This results in unexpected behavior which has already been reported, for example when combining LDAP with OIDC or SAML ( https://github.com/nextcloud/user_saml/issues/406 https://github.com/nextcloud/user_saml/issues/563 ).

Confusing the situation even more: the different handling of upper and lower case UUIDs by different LDAP servers. For example, LDAP via Samba returns uppercase ObjectGUIDs while OpenLDAP (for example via UCS) uses lowercase UUIDs. This makes migrating from one directory service to another, while keeping users in Nextcloud, very difficult. Dumping the database, replacing the UUIDs with their upper or lower case alternative, inserting the dump again and moving the user data directories in theory does work.

However, du to the fact that Nextcloud user_ldap uses UUIDs for mapping case sensitive usernames and UUIDs per definition are not case sensitive, the app should provide some kind of mechanism to deal with the issue.

>In the UI you put the option in the internal username but in the code you apply it to the UUID.

Yes, I'd expect that the LDAP server should not care if the UUIDs are upper or lower case if it is needed for other queries. But it will transform the UUID into upper or lower case and therefore will apply to the Nextcloud username.

> UUID and internal user name are two different things, even if by default UUID is used for internal user name.

Sure, but the username is generated from the UUID by default.

come-nc:
From what I see in the code even for UUID the casing from LDAP server is always used.
I’m not sure about this, adding an option for this scares me, it’s yet another option that people will not understand and misuse.

svenseeberg:
> From what I see in the code even for UUID the casing from LDAP server is always used.

Can you be a bit more specific why this is the case? This is the only place I can see which actually fetches the UUID from the LDAP server. If it is converted in this function, the converted case is used everywhere else. I actually tested this with a Nextcloud connected to a Samba LDAP. All users where created with upper case UUIDs as usernames. After switching this setting, new users were created with lower case UUIDs as usernames.

> I’m not sure about this, adding an option for this scares me, it’s yet another option that people will not understand and misuse.

This setting is definitely highly dangerous. As are all other settings in the Expert tab. With the default setting, nothing will change for existing Nextcloud installations.

Changing any of these settings on a running production server without knowing what will happen can definitely destroy a lot of data. However, the design choice was made that Nextcloud uses case sensitive usernames. And also the design choice was made that UUIDs from LDAP servers are not normalized to lower or upper case by default. If that would have happened in the past, this would not be necessary.

Also, it does generally not harm to normalize uppercase UUIDs to lowercase or vice versa as long as it is done from the beginning. The UUID can be used in both cases as LDAP servers ignore the case for queries. The following queries will return the same object on all LDAP implementation I'm aware of:

Samba:
```
ldbsearch -H private/sam.ldb "(objectGUID=46D6D630-44C6-4D97-A0DD-DE1E3AF97D35)"
ldbsearch -H private/sam.ldb "(objectGUID=46d6d630-44c6-4d97-a0dd-de1e3af97d35)"
```
OpenLDAP:
```
slapcat -a "(entryUUID=46D6D630-44C6-4D97-A0DD-DE1E3AF97D35)"
slapcat -a "(entryUUID=46D6d630-44c6-4d97-a0dd-de1e3af97d35)"
```

But this setting is the only sane way to get any of these setups to work in a good way (IMHO the OIDC and SAML apps should have a similar setting). And yes, this is a problem that Nextcloud as a software should solve, because it ignored RFC 4122.

## SSO with Keycloak via OIDC & Synchronization of Users and Groups with Samba
- Samba uses upper case UUIDs
- Keycloak imports the UUIDs in lower case

When you log in with OIDC, a second user object will be created in Nextcloud because the upper and lower case UUID based usernames do not match.

With this setting, it would be possible to force lower case usernames via LDAP and therefore the user object from LDAP and OIDC will be the same.

## Migration from Samba to UCS
- All user objects are created with upper case UUIDs as usernames because Samba returns the UUIDs in upper case
- When migrating to UCS, OpenLDAP will usually return lower case UUIDs

When a user logs in after the migration, a second account is created instead of the old one being reused.

In this case, switching the setting from "accept" to "uppercase" will convert the UUIDs from OpenLDAP to upper case and the existing user accounts continue to be used.

## Migrating from UCS to Samba
Basically the inverse procedure to "Migration from Samba to UCS".

svenseeberg:
I think I have to revise the use cases: migrating between different LDAP servers might be possible with `occ ldap:update-uuid`. However, the use case with OIDC still seems very valid. But even `update-uuid` looks a bit odd to me, as it seems to use the (not so permanent) DN to update the UUID (which is supposed to not change in most LDAP setups I know).

svenseeberg:
> From what I see in the code even for UUID the casing from LDAP server is always used.

Okay, now I got it. The case is not applied to https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/Access.php#L544-L550. I see 4 options for tackling the issue:
- Check if "Internal Username Attribute" is the same as the "UUID Attribute for Users". If yes, apply the case.
- Add a note that this does not work when explicitly choosing a username attribute.
- Check if `$username[0]` does actually contain a string that is a UUID.
- Add a checkbox to apply the case setting to the username attribute regardless of the attribute chosen.

svenseeberg:
> I’m not sure I understand the 4 options. 

I'll elaborate a bit:

>> * Rename the new setting to "Force case for username attribute" and apply always. (Currently my preferred option)

We can allow the admin to convert the string returned from the ldapExpertUsernameAttr in https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/Access.php#L544-L550. It would be up to the admin to decide for which attributes this setting is useful and for which it isn't.

* Advantages: very generic setting
* Disadvantages: none really (maybe it could accidentally applied to a cn based username)

>> * Check if "Internal Username Attribute" is the same as the "UUID Attribute for Users". If yes, apply the case.

Instead of letting the admin choose the case for the ldapExpertUsernameAttr, we let the code automatically decide. It will decide based on 2 factors: a) upper or lower case is enforced b) the attribute matches ldapExpertUUIDUserAttr (or the automatically detected value).

* Advantages: the setting would only apply to UUIDs and usernames derived from UUIDs but not to usernames based on the cn or other attributes
* Disadvantages: I think there is too much magic happening here which is hard to explain

>> * Check if `$username[0]` does actually contain a string that is a UUID.

If $username in https://github.com/nextcloud/server/blob/master/apps/user_ldap/lib/Access.php#L550-L553 actually contains a UUIDv4 (validate string with a new function), we apply the setting there.

* Advantages: the setting will only apply to UUIDv4 values
* Disadvantages: needs a function to parse a string to detect if it is a UUIDv4

>> * Add a checkbox to apply the case setting to the username attribute regardless of the attribute chosen.

Basically an extension to the first option to make it even more explicit. But I think we can skip that. Its not really useful.

> If you add a checkbox to force a case on the internal name it should always apply to internal name and leave UUID alone.

Not touching the UUID at all is definitely an option. As it should not matter whether it is upper or lower case, I thought it would be beneficial to have it as the same case as the username. I think there are good reasons for applying and for not applying the setting to the UUID stored in the database.

blizzz:
In most IdPs like keycloak you can modify the original value before submitting it to the SP. In Keycloak you can have a Javascript Mapper that would uppercase the UUID and you'd be done.

It would also not be an issue once we implement the user lookup from SAML to LDAP and get rid of the SAML-ID to Nextcloud User ID mapping.

I am not a fan of having the configuration there. Either this can be made an automatic lower-casing once when the value is identified as valid UUID – which still bears a risk. Or resort to the options just stated ^

svenseeberg:
> In most IdPs like keycloak you can modify the original value before submitting it to the SP. In Keycloak you can have a Javascript Mapper that would uppercase the UUID and you'd be done.

Yes, I implemented that. However the problem is that new users will receive an upper case UUID username after migrating to Samba while old users have a lower case UUID as username.

The only workaround here is to replace all user UUIDs in a database dump and insert the DB again AFAICT.

blizzz:
It is best dealt with by resolving the hard dependency between the SAML provided id and the Nextcloud user id by a look up mechanism. But, cannot say when it arrives, so for the short term i suppose the dump modification is better for you, but contains its own risks of course.

svenseeberg:
> It is best dealt with by resolving the hard dependency between the SAML provided id and the Nextcloud user id by a look up mechanism.

But wouldn't you have to implement the exact same (or an even more complicated) logic there? The is that you would have to implement some logic there as the usernames / lookup IDs are not normalized.

blizzz:
> > It is best dealt with by resolving the hard dependency between the SAML provided id and the Nextcloud user id by a look up mechanism.
> 
> But wouldn't you have to implement the exact same (or an even more complicated) logic there? The is that you would have to implement some logic there as the usernames / lookup IDs are not normalized.

Not the same, not really complicated, and the setting would not be potentially damaging.

svenseeberg:
I still think that user_ldap should provide a mechanism to deal with changing LDAP servers w/o having to manually change usernames in the database, but if the maintainers think that it is not worth the effort I can live with that.

Issue #43759, opened 2024-02-22
[Bug]: Restoring multiple big folders from trash bin fails


mgallien:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Restoring multiple big folders at once from trash bin fails with an error message: "Failed to restore some elements"

### Steps to reproduce

1. create 4 folders named `bigFolder1`, `bigFolder2`, `bigFolder3` and `bigFolder4`
2. create many empty files inside each folder `for i in {1..10000}; do touch test${i}.txt; done`
3. get the files scan to finish `sudo -u www-data php ./occ files:scan <uid>`
4. delete the 4 folders from web UI
5. try to restore them with trash bin web UI

### Expected behavior

all folders are restored without error

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

master

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
sudo -u www-data php occ config:list system 
{
    "system": {
        "debug": true,
        "profiler": true,
        "apps_paths": [
            {
                "path": "\/var\/www\/html\/apps",
                "url": "\/apps",
                "writable": false
            },
            {
                "path": "\/var\/www\/html\/apps-extra",
                "url": "\/apps-extra",
                "writable": false
            },
            {
                "path": "\/var\/www\/html\/apps-shared",
                "url": "\/apps-shared",
                "writable": false
            },
            {
[...]

PR #44504, opened 2024-03-27
Add retry logic to cover deadlock situations during move operations

artonge:
Should help with https://github.com/nextcloud/server/issues/43759

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -40,6 +40,7 @@
 
 namespace OC\Files\Cache;
 
+use Doctrine\DBAL\Exception\RetryableException;
 use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use OC\Files\Search\SearchComparison;
 use OC\Files\Search\SearchQuery;
@@ -692,7 +693,6 @@ class Cache implements ICache {
 				throw new \Exception('Invalid target storage id: ' . $targetStorageId);
 			}
 
-			$this->connection->beginTransaction();
 			if ($sourceData['mimetype'] === 'httpd/unix-directory') {
 				//update all child entries
 				$sourceLength = mb_strlen($sourcePath);
@@ -715,12 +715,31 @@ class Cache implements ICache {
 					$query->set('encrypted', $query->createNamedParameter(0, IQueryBuilder::PARAM_INT));
 				}
 
-				try {
-					$query->execute();
-				} catch (\OC\DatabaseException $e) {
-					$this->connection->rollBack();
-					throw $e;
+				// Retry transaction in case of RetryableException like deadlocks.
+				// Retry up to 4 times because we should receive up to 4 concurrent requests from the frontend
+				$retryLimit = 4;
+				for ($i = 1; $i <= $retryLimit; $i++) {
+					try {
+						$this->connection->beginTransaction();
+						$query->executeStatement();
+						break;
+					} catch (\OC\DatabaseException $e) {
+						$this->connection->rollBack();
+						throw $e;
+					} catch (RetryableException $e) {
+						// Simply throw if we already retried 4 times.
+						if ($i === $retryLimit) {
+							throw $e;
+						}
+
+						$this->connection->rollBack();
+
+						// Sleep a bit to give some time to the other transaction to finish.
+						usleep(100 * 1000 * $i);
+					}
 				}
+			} else {
+				$this->connection->beginTransaction();
 			}
 
 			$query = $this->getQueryBuilder();


lib/private/Files/Cache/Cache.php
```diff
@@ -715,12 +715,29 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
+						$this->connection->rollBack();
+						throw $e;
+					} catch (RetryableException $e) {
+						// Simply throw if we already retried 4 times.
+						if ($i === 4) {
+							throw $e;
+						}
+
+						// Sleep for 1 seconde to give some time to the other transaction to finish.
+						sleep(1);
```
icewind1991, line 735: I expect that there isn't a need to wait here, the deadlock error should clear both waiting locks I believe.

icewind1991: Or I would at least reduce the sleep to 10-100ms

Altahrim: I think it doesn't clear the original lock but a much shorter sleep should be enough.
We can also have an optimistic retry and wait longer after:
```suggestion
						usleep($i*$i*100);
```
I also think we should rollback before waiting

lib/private/Files/Cache/Cache.php
```diff
@@ -715,12 +715,31 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
+					} catch (RetryableException $e) {
+						// Simply throw if we already retried 4 times.
+						if ($i === $retryLimit) {
+							throw $e;
+						}
+
+						$this->connection->rollBack();
+
+						// Sleep a bit to give some time to the other transaction to finish.
+						sleep(100 * $i);
```
icewind1991, line 738: I think waiting for up to 400s isn't what you mean here :see_no_evil:

artonge: AH ! thanks :)

lib/private/Files/Cache/Cache.php
```diff
@@ -715,12 +715,31 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
+					} catch (RetryableException $e) {
+						// Simply throw if we already retried 4 times.
+						if ($i === $retryLimit) {
+							throw $e;
+						}
+
+						$this->connection->rollBack();
+
+						// Sleep a bit to give some time to the other transaction to finish.
+						usleep(100 * $i);
```
icewind1991, line 738: Now you're only sleeping for up to 0.4ms :)

`usleep(100 * 1000 * $i);` is what you're looking for

lib/private/Files/Cache/Cache.php
```diff
@@ -715,12 +715,31 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
-				} catch (\OC\DatabaseException $e) {
-					$this->connection->rollBack();
-					throw $e;
+				// Retry transaction in case of RetryableException like deadlocks.
+				// Retry up to 4 times because we should receive up to 4 concurrent requests from the frontend
+				$retryLimit = 4;
+				for ($i = 1; $i <= $retryLimit; $i++) {
+					try {
+						$this->connection->beginTransaction();
+						$query->execute();
```
come-nc, line 724:
```suggestion
						$query->executeStatement();
```

come-nc (commented):
I think I like it, especially if it works? Did not test yet.

Is there anyway to let the DB or our Query layer handle that? Have some kind of generic way of saying «run this query in a transaction, if it deadlocks retry up to X times» ?

lib/private/Files/Cache/Cache.php
artonge: Indeed, let see if it works as we expect it to, and then we can maybe add an argument to `executeStatement()`

artonge: I also apply your suggestion on the `execute` call

Altahrim (approved):

lib/private/Files/Cache/Cache.php
artonge: Well, thanks for the handholding :see_no_evil:

lib/private/Files/Cache/Cache.php
```diff
@@ -715,12 +715,31 @@ public function moveFromCache(ICache $sourceCache, $sourcePath, $targetPath) {
...
 				}
 
-				try {
-					$query->execute();
-				} catch (\OC\DatabaseException $e) {
-					$this->connection->rollBack();
-					throw $e;
+				// Retry transaction in case of RetryableException like deadlocks.
+				// Retry up to 4 times because we should receive up to 4 concurrent requests from the frontend
+				$retryLimit = 4;
```
juliushaertl, line 720: https://github.com/nextcloud/server/blob/050c6d53b38d6737a49c13d4e6a20a6062160503/lib/public/AppFramework/Db/TTransactional.php#L94 could be used for that though it does not have a wait in it

juliusknorr (approved):

lib/private/Files/Cache/Cache.php
artonge: Oh, seems better to use that, no? We could even add a support for an optional wait time.

artonge: But is a trait, I would rather have that in IDBConnection or IQueryBuilder

Issue #14917, opened 2019-03-31
trash bin: show user who deleted that file/folder


alexanderdd:
**Is your feature request related to a problem? Please describe.**
Sometimes when checking the trash, I find files from shared folders, I would like to know why they were deleted. I don't know which user to ask because I don't know who deleted it.

**Describe the solution you'd like**
In trash bin, show username of the user who deleted the file, for example via a sidebar like in the regular file view (where comments, versions etc are shown).

**Additional context**
Related to https://github.com/nextcloud/server/issues/14138

sehucke:
Probably related to #10017

alexanderdd:
Does a deleted file only appear in the trash bin of the person who deleted it? Or in everyone's trashbin?

And how is it for group folders?

being-peace:
Related to https://github.com/nextcloud/server/issues/24544.

PR #44643, opened 2024-04-03
feat(trashbin): Add deleted by properties

Pytal:
- Part of https://github.com/nextcloud/server/issues/14917

## Summary

- Store and provide a new `trashbin-deleted-by-id` and `trashbin-deleted-by-display-name` DAV properties

- Example PROPFIND request data
```xml
<?xml version="1.0"?>
<d:propfind xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns">
  <d:prop>
    <oc:fileid />
    <d:displayname />
    <d:getlastmodified />
    <d:getcontenttype />
    <oc:size />
    <oc:owner-id />
    <oc:share-types />
    <nc:sharees />
    <nc:share-download-limits />
    <nc:trashbin-filename />
    <nc:trashbin-deletion-time />
    <nc:trashbin-original-location />
    <nc:trashbin-title />
    <nc:trashbin-deleted-by />
  </d:prop>
</d:propfind>
```

## Checklist

### DIFF
--- a/apps/files_trashbin/lib/Migration/Version1020Date20240403003535.php
+++ b/apps/files_trashbin/lib/Migration/Version1020Date20240403003535.php
@@ -1,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * @copyright 2024 Christopher Ng <chrng8@gmail.com>
+ *
+ * @author Christopher Ng <chrng8@gmail.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\Files_Trashbin\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version1020Date20240403003535 extends SimpleMigrationStep {
+
+	/**
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('files_trash')) {
+			return null;
+		}
+
+		$table = $schema->getTable('files_trash');
+		$table->addColumn('deleted_by', Types::STRING, [
+			'notnull' => false,
+			'length' => 64,
+		]);
+
+		return $schema;
+	}
+}
--- a/apps/files_trashbin/lib/Trash/LegacyTrashBackend.php
+++ b/apps/files_trashbin/lib/Trash/LegacyTrashBackend.php
@@ -33,16 +33,16 @@ use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 use OCP\Files\Storage\IStorage;
 use OCP\IUser;
+use OCP\IUserManager;
 
 class LegacyTrashBackend implements ITrashBackend {
 	/** @var array */
 	private $deletedFiles = [];
 
-	/** @var IRootFolder */
-	private $rootFolder;
-
-	public function __construct(IRootFolder $rootFolder) {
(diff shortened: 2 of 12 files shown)


apps/files_trashbin/lib/Migration/Version1020Date20240403003535.php
```diff
@@ -0,0 +1,56 @@
...
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable('files_trash')) {
+			return null;
+		}
+
+		$table = $schema->getTable('files_trash');
+		$table->addColumn('deleted_by', Types::STRING, [
+			'notnull' => false,
```
Altahrim, line 50: It seems we don't really uses the `null`, we can avoid it: 
```suggestion
			'notnull' => true,
```
Empty string should be enough if value isn't filled.

tcitworld: Don't you need to enforce a default value in that case?

Pytal: Now using notnull + default empty string

apps/files_trashbin/lib/Trash/TrashManager.php
```diff
@@ -124,4 +133,30 @@ public function pauseTrash() {
...
+
+		/** @var false|string */
+		$deletedBy = $result->fetchOne();
+		$result->closeCursor();
+		if ($deletedBy === false) {
+			return null;
+		}
+
+		$user = $this->userManager->get($deletedBy);
+		return $user;
```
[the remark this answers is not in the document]
Pytal: Yes, good catch! Using `empty` check

icewind1991 (changes requested):
This would cause a query for every trash item in the trashbin.

icewind1991:
How about adding it to `ITrashItem` instead so the value can be retrieved when listing trash items

Pytal:
Switched the column back to nullable to fix potential migration errors with oracle which converts empty string to null

icewind1991 (approved):

sorbaugh (approved):

apps/files_trashbin/lib/Trash/LegacyTrashBackend.php
```diff
@@ -59,14 +59,17 @@ private function mapTrashItems(array $items, IUser $user, ?ITrashItem $parent =
 			if (!$originalLocation) {
 				$originalLocation = $file->getName();
 			}
+			/** @psalm-suppress UndefinedInterfaceMethod */
```
come-nc, line 62:
```suggestion
```
Why is that needed? You added it to the interface, it should find it.

apps/files_trashbin/lib/Trash/TrashItem.php
```diff
@@ -27,41 +27,31 @@
...
+		private ?IUser $deletedBy,
 	) {
 		$this->backend = $backend;
-		$this->orignalLocation = $originalLocation;
+		$this->originalLocation = $originalLocation;
 		$this->deletedTime = $deletedTime;
 		$this->trashPath = $trashPath;
 		$this->fileInfo = $fileInfo;
 		$this->user = $user;
+		$this->deletedBy = $deletedBy;
```
come-nc, line 46: You need to remove these lines, the point of constructor property promotion is that it’s done automagically.

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -126,23 +126,31 @@ public static function getUidAndFilename($filename) {
...
+	public static function getExtraData($user) {
 		$query = \OC::$server->getDatabaseConnection()->getQueryBuilder();
-		$query->select('id', 'timestamp', 'location')
+		$query->select('id', 'timestamp', 'location', 'deleted_by')
 			->from('files_trash')
 			->where($query->expr()->eq('user', $query->createNamedParameter($user)));
 		$result = $query->executeQuery();
 		$array = [];
 		while ($row = $result->fetch()) {
 			if (isset($array[$row['id']])) {
```
come-nc, line 142: I think this `if` is useless and autovivification can be used. You should be able to simply use the inside of the `if` branch and drop the rest.

come-nc (commented):
Did you test that with groupfolders?

apps/files_trashbin/lib/Trash/LegacyTrashBackend.php
Pytal: This is not for the method but the array key access as public FileInfo does not support it but private does https://github.com/nextcloud/server/blob/5b85562784be0c10cb885d2865f66cf6e959f90b/lib/private/Files/FileInfo.php#L124-L137

Pytal:
> Did you test that with groupfolders?

Seems it breaks groupfolders as it implements a custom trash backend, documented the change here https://github.com/nextcloud/documentation/pull/11757

apps/files_trashbin/lib/Migration/Version1020Date20240403003535.php
nickvergessen, line 1: When adding a migration file you need to bump the version of the app so updates are triggered

nickvergessen: Sent in https://github.com/nextcloud/server/pull/45000

Pytal: Oop missed that part, thanks!

PR #44644, opened 2024-04-04
fix(appconfig): format app values

ArtificialOwl:
returns formated values based on type when calling `getAllValues()` and `searchValues()`

note: deprecated method `getValues()` use `getAllValues()` but if the config value is defined as MIXED (default), the value is kept as a string.

original:
```
{
    "key1": "value1",
    "key2": "value0",
    "key6": "1",
    "key7": "{\"test1\":1,\"test2\":2}",
    "test8": "1",
    "key3": "value0",
    "key4": "3",
    "key5": "3.14"
}
```

new:

```
{
    "key1": "value1",
    "key2": "value0",
    "key6": true,
    "key7": {
        "test1": 1,
        "test2": 2
    },
    "test8": true,
    "key3": "value0",
    "key4": 3,
    "key5": 3.14
}
```

same as #43258 with tests

### DIFF
--- a/lib/private/AppConfig.php
+++ b/lib/private/AppConfig.php
@@ -219,8 +219,9 @@ class AppConfig implements IAppConfig {
 		// if we want to filter values, we need to get sensitivity
 		$this->loadConfigAll();
 		// array_merge() will remove numeric keys (here config keys), so addition arrays instead
+		$values = $this->formatAppValues($app, ($this->fastCache[$app] ?? []) + ($this->lazyCache[$app] ?? []));
 		$values = array_filter(
-			(($this->fastCache[$app] ?? []) + ($this->lazyCache[$app] ?? [])),
+			$values,
 			function (string $key) use ($prefix): bool {
 				return str_starts_with($key, $prefix); // filter values based on $prefix
 			}, ARRAY_FILTER_USE_KEY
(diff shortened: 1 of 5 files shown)


lib/private/AppConfig.php
```diff
@@ -271,7 +272,8 @@ public function searchValues(string $key, bool $lazy = false): array {
 
 		foreach (array_keys($cache) as $app) {
 			if (isset($cache[$app][$key])) {
-				$values[$app] = $cache[$app][$key];
+				$appCache = $this->formatAppValues((string)$app, $cache[$app], $lazy);
+				$values[$app] = $appCache[$key];
```
come-nc, line 276: This is formatting all keys to actually only use one?
Should use an other `formatAppValue` method with only the used one.

lib/private/AppConfig.php
```diff
@@ -1386,6 +1388,48 @@ public function getFilteredValues($app) {
...
+
+	/**
+	 * **Warning:** avoid default NULL value for $lazy as this will
+	 * load all lazy values from the database
+	 *
+	 * @param string $app
+	 * @param array $values
+	 * @param bool|null $lazy
+	 *
+	 * @return array
```
come-nc, line 1400:
```suggestion
	 * @param array<string,string> $values
	 * @param bool|null $lazy
	 *
	 * @return array<string,int|float|bool|array|string>
```

lib/private/AppConfig.php
ArtificialOwl: true, my bad ....
nice catch !

ArtificialOwl: should be good now

come-nc (approved):

susnux (approved):

PR #44658, opened 2024-04-04
fix: Remove obsolete resource typing

come-nc:
## Summary

In PHP>=8.1, LDAP and FTP resources are always typed objects

## TODO

- [x] Wait for https://github.com/nextcloud/server/pull/44607 and rebase
- [x] Update stubs for gd and ftp
- [x] Remove references to gd resource as well

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Wizard.php
+++ b/apps/user_ldap/lib/Wizard.php
@@ -48,8 +48,7 @@ use Psr\Log\LoggerInterface;
 class Wizard extends LDAPUtility {
 	protected static ?IL10N $l = null;
 	protected Access $access;
-	/** @var resource|\LDAP\Connection|null */
-	protected $cr;
+	protected ?\LDAP\Connection $cr = null;
 	protected Configuration $configuration;
 	protected WizardResult $result;
 	protected LoggerInterface $logger;
@@ -361,7 +360,7 @@ class Wizard extends LDAPUtility {
 		if (!$this->ldap->isResource($rr)) {
 			return false;
 		}
-		/** @var resource|\LDAP\Result $rr */
+		/** @var \LDAP\Result $rr */
 		$er = $this->ldap->firstEntry($cr, $rr);
 		$attributes = $this->ldap->getAttributes($cr, $er);
 		if ($attributes === false) {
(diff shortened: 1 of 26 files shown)


kesselb (approved):
Nice :pray:

apps/user_ldap/lib/Wizard.php
```diff
@@ -1214,9 +1209,7 @@ public function cumulativeSearchOnAttribute(array $filters, string $attr, int $d
 					$dnReadCount++;
 					$foundItems = array_merge($foundItems, $newItems);
 					$dnRead[] = $dn;
-				} while (($state === self::LRESULT_PROCESSED_SKIP
-						|| $this->ldap->isResource($entry))
-						&& ($dnReadLimit === 0 || $dnReadCount < $dnReadLimit));
+				} while ($dnReadLimit === 0 || $dnReadCount < $dnReadLimit);
```
blizzz, line 1212: You are not taking `$state` into account anymore. Is that intended?

build/stubs/ldap.php
```diff
@@ -25,4 +27,200 @@ final class Result
     final class ResultEntry
     {
     }
+
+}
+
+namespace {
+
+    #ifdef HAVE_ORALDAP
```
blizzz, line 35: what's that?

blizzz (commented):
One thing to doublecheck

nickvergessen (approved):
:elephant:

apps/user_ldap/lib/Wizard.php
come-nc: The call to `$this->ldap->isResource($entry)` cannot return `false`, by typing, so the `$state` was not taken into account before either. If this is/was a bug, we can change the condition.
What do you think?

build/stubs/ldap.php
come-nc: oracle ldap :-(
At some point I removed support for compiling php-ldap with that but a contribution added it back :upside_down_face:

apps/user_ldap/lib/Wizard.php
blizzz: Right. I was a bit puzzled also because the local `$state` var was not removed (NaaS). All good.

blizzz (approved):

PR #44661, opened 2024-04-04
 fix(files): Do not show files from hidden folders in "Recent"-view if hidden files are disabled by user

susnux:
## Summary
Have a hidden folder .foo and create a file inside (.foo/bar) then the file bar is shown in recent even if it is basically hidden.

## Checklist


apps/files/src/views/recent.ts
```diff
@@ -19,13 +19,24 @@
...
 import HistorySvg from '@mdi/svg/svg/history.svg?raw'
 
 import { getContents } from '../services/Recent'
-import { View, getNavigation } from '@nextcloud/files'
 
 export default () => {
+	// Current state of user config for hidden files
+	let { show_hidden: showHiddenFiles } = loadState<{ show_hidden: boolean }>('files', 'config')
+	// When the user config changes and the hidden files setting is changed we need to reload the directory content
+	subscribe('files:config:updated', ({ key, value }: { key: string, value: boolean}) => {
```
skjnldsv, line 34: Seeing this, I wonder if we should not update the initial state value on config change. That would also be nice in case something else is read in another location somewhere.

skjnldsv: Hmmm, couldn't you also just use the config store? Or would that be a bit too dirty?

susnux: > Hmmm, couldn't you also just use the config store? Or would that be a bit too dirty?

Using a store outside a component is possible, but that requires more work.

> I wonder if we should not update the initial state value on config change.

I am not sure how good this is, do we somewhere do this? I would expect it to be readonly

susnux: I adjusted the code to allow using the store already in the views

skjnldsv: > I am not sure how good this is, do we somewhere do this? I would expect it to be readonly

That's a good question.
To be fair, we could assume we have components loading some initial states on user interactions. Some which might be outdated.
Maybe it's worth updating the json-base64 value of the state input?
I'm not sure, might be an antipattern. 

I guess this all goes down to if we consider initial-states legitimacy to be on page load only and potentially outdated after that 🤔

skjnldsv: > I adjusted the code to allow using the store already in the views

I raised the idea but I'm far from being convinced that it's the way to go. What is your opinion on this? Which approach do you prefer? The previous one you wrote or the latest with the store?

susnux: I think the store (current) variant is the cleanest. We are in the same app so we only have one place to manage the state.

apps/files/src/views/FilesList.vue
```diff
@@ -476,7 +484,7 @@ export default defineComponent({
 			}
 
 			// Fetch the current dir contents
-			this.promise = currentView.getContents(dir) as Promise<ContentsWithRoot>
+			this.promise = currentView.getContents(dir, this.fetchContent) as Promise<ContentsWithRoot>
```
skjnldsv, line 487:
```suggestion
			this.promise = currentView.getContents(dir) as Promise<ContentsWithRoot>
```

apps/files/src/views/recent.ts
```diff
@@ -40,6 +42,15 @@ export default () => {
...
+		getContents: async (path = '/') => {
+			const store = useUserConfigStore(pinia)
+
+			const content = await getContents(path)
+			if (path === '/' && !store.userConfig.show_hidden) {
+				// We need to hide files from hidden directories in the root if not configured
+				content.contents = content.contents.filter((node) => node.dirname.split('/').some((dir) => dir.startsWith('.')))
+			}
+			return content
+		},
```
skjnldsv, line 54: Why not do that straight into the `services/Recent.ts` file?

susnux: Would be possible, yes. But I thought we might have a use case for the service to deliver unfiltered results. So only apply the filter in the UI / view.

But I have no strong opinion here.

skjnldsv: We do parsing, post-processing of values in the services already.
I think filtering is also good there.

I'd also add that it would simplify where to look for the `getContents` source

susnux: > I'd also add that it would simplify where to look for the getContents source

Agree, so `Recent.ts` from services it is.

apps/files/src/views/FilesList.vue
```diff
@@ -452,10 +454,16 @@ export default defineComponent({
...
+
+		// reload on settings change
+		this._unsubscribeStore = this.userConfigStore.$subscribe(() => this.fetchContent(), { deep: true })
 	},
 
 	unmounted() {
 		unsubscribe('files:node:updated', this.onUpdatedNode)
+		unsubscribe('nextcloud:unified-search.search', this.onSearch)
+		unsubscribe('nextcloud:unified-search.reset', this.onSearch)
+		this._unsubscribeStore()
```
emoral435, line 466: doesn't this just return {}, and nothing catches it? I am a bit unfamiliar with this syntax, so forgive me if this is an oblivious comment :)

apps/files/src/services/Recent.ts
```diff
@@ -20,17 +20,34 @@
...
+ * This takes the users preference about hidden files into account.
+ * If hidden files are not shown, then also recently changed files *in* hidden directories are filtered.
+ *
+ * @param path Path to search for recent changes
+ */
 export const getContents = async (path = '/'): Promise<ContentsWithRoot> => {
+	const store = useUserConfigStore(pinia)
+	const filterHidden = (node: Node) =>
+		// We need to hide files from hidden directories in the root if not configured to show
+		path === '/'
```
emoral435, line 47: Because this specifies the root folder with `path === '/'`, doesn't this mean that it only filters the root folder, and not any hidden filers that can possibly be nested?

For example, if we have path `/myFolder`, and that folder has the `.foo/bar.ts`, then it will not filter out the hidden folder, correct? I can be wrong, but perhaps the path check is not required?

emoral435 (commented):
Might be asking questions that are easily resolved, but everything LGTM!

apps/files/src/views/FilesList.vue
susnux: See the `mounted` function above, it is set to the unregister function of pinia::$subscribe (aka unsubscribe)

apps/files/src/services/Recent.ts
susnux: In every view the files list takes care about hidden files, but the root view.
In the root view we have the problem (at least for "recent") that there might be files from different folders, like `/foo/.bar/file` is shown in `/` for "recent".

If you are in `/foo` then `.bar` is automatically hidden.

emoral435: Ahhh, I see now! My question was indeed silly 🙈

emoral435 (approved):
Code looks good :)

Pytal (approved):

PR #39669, opened 2023-08-02
occ: Add `--all` support to `user:lastseen`

nooblag:
* Resolves: #39048

## Summary
Adds `--all` argument to `user:lastseen` for `occ`.

## Checklist

- [x] Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [X] [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [ ] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [ ] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #44666, opened 2024-04-04
occ: Improve `user:lastseen` timestamp

### LINKED ISSUE #44641: occ: Improve and make consistent `user:lastseen` and `user:info` timestamps
Regarding the improvements to `occ` in https://github.com/nextcloud/server/pull/39669  that were recently merged, the times returned always seem to be in UTC. This is a proposal to improve the date/time formatting to either:

A) leave all times in hardcoded UTC and make that explicit (by adding UTC to the end of the time), or
B) show the times from the PHP timezone set in `php.ini` (and display that timezone information after the date/time).

Also, at the moment, the output of `user:info` timestamp is: `2024-04-03T00:40:22+00:00` which is different to `user:lastseen`'s format of `2024-04-03 22:29`.

There is also a bug in `user:info` if login has never been seen, which I'll describe at the end of this post.
<br>

Personally, I'd like option B, to show the times from the PHP timezone set in `php.ini` and display that timezone information after the date/time, and unify for each output.

For example, something like:

```
$ sudo -u www-data php /var/www/nextcloud/occ user:lastseen test
test's last login: 2024-04-03 22:29:36 AEDT
```

```
$ sudo -u www-data php /var/www/nextcloud/occ user:info test
  - user_id: test
  - display_name: test
  - email: test@test.local
  - cloud_id: test@test.local
  - enabled: true
  - groups:
    - admin
  - quota: none
  - storage:
    - free: 2198323200
    - used: 38251355
    - total: 2236574555
    - relative: 1.71
    - quota: -3
  - last_seen: 2024-04-03 22:29:36 AEDT
  - user_directory: /var/www/nextcloud/data/test
  - backend: Database
```
<br>

The other alternative could be to leave the times displayed always in UTC, but to just make that explicit:

For example:

```
$ sudo -u www-data php /var/www/nextcloud/occ user:lastseen test
test's last login: 2024-04-03 11:29:36 UTC
```
or similar...

```
$ sudo -u www-data php /var/www/nextcloud/occ user:info test
  - user_id: test
  - display_name: test
  - email: test@test.local
  - cloud_id: test@test.local
  - enabled: true
  - groups:
    - admin
  - quota: none
  - storage:
    - free: 2198323200
    - used: 38251355
    - total: 2236574555
    - relative: 1.71
    - quota: -3
  - last_seen: 2024-04-03 11:29:36 UTC
  - user_directory: /var/www/nextcloud/data/test
  - backend: Database
```

---

There is also a bug in `user:info` if login has never been seen. It appears to show a Unix epoch time, instead of perhaps something like 'never', which we could also make consistent with `user:lastseen` output, for instance:

```
$ sudo -u www-data php /var/www/nextcloud/occ user:info bob
  - user_id: bob
  - display_name: Bob
  - email: bob@example.com
  - cloud_id: bob@test.local
  - enabled: true
  - groups:
    - Everyone
  - quota: 2 GB
  - storage:
    - free: 2146980495
    - used: 503153
    - total: 2147483648
    - relative: 0.02
    - quota: 2147483648
  - last_seen: 1970-01-01T00:00:00+00:00
  - user_directory: /var/www/nextcloud/data/bob
  - backend: Database
```

nooblag:
* Resolves: https://github.com/nextcloud/server/issues/44641

## TODO

- [x] Decide whether to display times always in UTC or pull Timezone from `php.ini` setting
- [x] Fix bug in `user:info` where last login is zero (don't show Unix epoch time)

## Checklist

- [X] Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [X] [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits

### DIFF
--- a/core/Command/User/Info.php
+++ b/core/Command/User/Info.php
@@ -49,9 +49,6 @@ class Info extends Base {
 
 		$groups = $this->groupManager->getUserGroupIds($user);
 
-		if (ini_get('date.timezone')) {
-			date_default_timezone_set(ini_get('date.timezone'));
-		}
 		if ($user->getLastLogin() == 0) {
 			$lastseen = "never";
 		} else {
--- a/core/Command/User/Info.php
+++ b/core/Command/User/Info.php
@@ -56,7 +56,7 @@ class Info extends Base {
 			'groups' => $groups,
 			'quota' => $user->getQuota(),
 			'storage' => $this->getStorageInfo($user),
-			'last_seen' => date(\DateTimeInterface::ATOM, $user->getLastLogin()), // ISO-8601
+			'last_seen' => date('Y-m-d H:i:s T', $user->getLastLogin()),
 			'user_directory' => $user->getHome(),
 			'backend' => $user->getBackendClassName()
 		];
--- a/core/Command/User/Info.php
+++ b/core/Command/User/Info.php
@@ -46,12 +46,18 @@ class Info extends Base {
 			$output->writeln('<error>user not found</error>');
 			return 1;
 		}
+
 		$groups = $this->groupManager->getUserGroupIds($user);
+
+		if (ini_get('date.timezone')) {
+			date_default_timezone_set(ini_get('date.timezone'));
+		}
 		if ($user->getLastLogin() == 0) {
 			$lastseen = "never";
 		} else {
 			$lastseen = date('Y-m-d H:i:s T', $user->getLastLogin());
 		}
+
 		$data = [
 			'user_id' => $user->getUID(),
 			'display_name' => $user->getDisplayName(),
@@ -65,6 +71,7 @@ class Info extends Base {
 			'user_directory' => $user->getHome(),
 			'backend' => $user->getBackendClassName()
 		];
+
 		$this->writeArrayInOutputFormat($input, $output, $data);
 		return 0;
 	}
--- a/core/Command/User/Info.php
+++ b/core/Command/User/Info.php
@@ -46,15 +46,7 @@ class Info extends Base {
 			$output->writeln('<error>user not found</error>');
 			return 1;
 		}
-
 		$groups = $this->groupManager->getUserGroupIds($user);
-
-		if ($user->getLastLogin() == 0) {
-			$lastseen = "never";
-		} else {
-			$lastseen = date('Y-m-d H:i:s T', $user->getLastLogin());
-		}
-
 		$data = [
 			'user_id' => $user->getUID(),
 			'display_name' => $user->getDisplayName(),
@@ -64,11 +56,10 @@ class Info extends Base {
 			'groups' => $groups,
 			'quota' => $user->getQuota(),
 			'storage' => $this->getStorageInfo($user),
-			'last_seen' => $lastseen,
+			'last_seen' => date(\DateTimeInterface::ATOM, $user->getLastLogin()), // ISO-8601
 			'user_directory' => $user->getHome(),
 			'backend' => $user->getBackendClassName()
 		];
-
 		$this->writeArrayInOutputFormat($input, $output, $data);
 		return 0;
 	}
--- a/core/Command/User/Info.php
+++ b/core/Command/User/Info.php
@@ -47,6 +47,11 @@ class Info extends Base {
 			return 1;
 		}
 		$groups = $this->groupManager->getUserGroupIds($user);
+		if ($user->getLastLogin() == 0) {
+			$lastseen = "never";
+		} else {
+			$lastseen = date('Y-m-d H:i:s T', $user->getLastLogin());
+		}
 		$data = [
 			'user_id' => $user->getUID(),
 			'display_name' => $user->getDisplayName(),
@@ -56,7 +61,7 @@ class Info extends Base {
 			'groups' => $groups,
 			'quota' => $user->getQuota(),
 			'storage' => $this->getStorageInfo($user),
-			'last_seen' => date('Y-m-d H:i:s T', $user->getLastLogin()),
+			'last_seen' => $lastseen,
 			'user_directory' => $user->getHome(),
 			'backend' => $user->getBackendClassName()
 		];
--- a/core/Command/User/LastSeen.php
+++ b/core/Command/User/LastSeen.php
@@ -43,6 +43,11 @@ class LastSeen extends Base {
 
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		$singleUserId = $input->getArgument('uid');
+
+		if (ini_get('date.timezone')) {
+			date_default_timezone_set(ini_get('date.timezone'));
+		}
+
 		if ($singleUserId) {
 			$user = $this->userManager->get($singleUserId);
 			if (is_null($user)) {
--- a/core/Command/User/LastSeen.php
+++ b/core/Command/User/LastSeen.php
@@ -56,7 +56,7 @@ class LastSeen extends Base {
 			} else {
 				$date = new \DateTime();
 				$date->setTimestamp($lastLogin);
-				$output->writeln($user->getUID() . "'s last login: " . $date->format('Y-m-d H:i'));
+				$output->writeln($user->getUID() . "'s last login: " . $date->format('Y-m-d H:i:s T'));
 			}
 
 			return 0;
@@ -74,7 +74,7 @@ class LastSeen extends Base {
 			} else {
 				$date = new \DateTime();
 				$date->setTimestamp($lastLogin);
-				$output->writeln($user->getUID() . "'s last login: " . $date->format('Y-m-d H:i'));
+				$output->writeln($user->getUID() . "'s last login: " . $date->format('Y-m-d H:i:s T'));
 			}
 		});
 		return 0;
--- a/core/Command/User/LastSeen.php
+++ b/core/Command/User/LastSeen.php
@@ -44,10 +44,6 @@ class LastSeen extends Base {
 	protected function execute(InputInterface $input, OutputInterface $output): int {
 		$singleUserId = $input->getArgument('uid');
 
-		if (ini_get('date.timezone')) {
-			date_default_timezone_set(ini_get('date.timezone'));
-		}
-
 		if ($singleUserId) {
 			$user = $this->userManager->get($singleUserId);
 			if (is_null($user)) {


nooblag:
Hello all @ nickvergessen @skjnldsv @ArtificialOwl @Fenn-CS @come-nc -- I understand you've got a lot going on, but just a friendly question to see if the timetable for this might make it to NC29 final at the end of this month?

Fenn-CS:
> but just a friendly question to see if the timetable for this might make it in before NC29 final at the end of this month, by any chance? If so, I'll get working on the documentation changes too, etc?

The next RC release for `Nextcloud 29` is on May 30th, not sure if that answers your question but I believe as soon as this is ready and is reviews it would definitely get into the next release if back-ported to `NC29` as master is now `NC30`.

nooblag:
Hi @Fenn-CS Thanks for all that, it's helpful information. This is ready for review now.

core/Command/User/Info.php
```diff
@@ -65,7 +65,18 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+
 		$groups = $this->groupManager->getUserGroupIds($user);
+
+		if (ini_get('date.timezone')) {
+			date_default_timezone_set(ini_get('date.timezone'));
+		}
+		if ($user->getLastLogin() == 0) {
+			$lastseen = "never";
+		} else {
+			$lastseen = date('Y-m-d H:i:s T', $user->getLastLogin());
```
nickvergessen, line 77: Both, the format change as well as the timezone change are breaking behaviour of the command that will result in breaking all tools/scripts that utilize those commands.

My recommendation would be to add an optional `--timezone` argument and use the given timezone when provided and leave any calls without it untouched.

nooblag: Thank you for the input. I understand it's a breaking change. The point of this PR is to unify the timestamps though, and since 29.0.0 by way of semantic versioning denotes breaking changes in the API, breaking changes are acceptable, no?

At the moment we have different formatting at `user:info` and `user:lastseen` anyway, so unifying them will be a breaking change in either place.

core/Command/User/Info.php
```diff
@@ -65,7 +65,18 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 			$output->writeln('<error>user not found</error>');
 			return 1;
 		}
+
 		$groups = $this->groupManager->getUserGroupIds($user);
+
+		if (ini_get('date.timezone')) {
+			date_default_timezone_set(ini_get('date.timezone'));
+		}
```
come-nc, line 73: Is that needed? This is a PHP ini setting, we should not read it and apply it manually.

nooblag: Well without it, times for me always displayed in UTC regardless of `php.ini` CLI setting, so I'd say, yes.

nickvergessen: Nextcloud always sets it to UTC:
https://github.com/nextcloud/server/blob/f2c36a55fe2a0ef2efd15a9ce6a4c8f8420d87dc/lib/base.php#L656-L658
And that should stay like this.
You can make the timestamp a `new \DateTime('@' . $timestamp)` and then use `IDateTimeFormatter` with the timezone given to the command as explained above.

core/Command/User/Info.php
nickvergessen: While we try to follow semantic versioning, breaking should be avoided as much as possible, that includes changing defaults and things like this here.
A newly added optional parameter can be used to fix your case without breaking existing usages.

nooblag: Thank you for the note.

I am saying there will be a breaking change regardless though. Putting the formatting aside, fixing the bug when last seen time = 0 by saying "never" instead of returning Unix epoch will be a breaking change. And so if we're fixing an issue that will result in a breaking change anyway, why not unify the timestamp formatting while we're at it? Unless there's a reason that unifying the formatting change here is specifically undesirable?

Is the existing `DateTimeInterface::ATOM` format consistent and the established default in `occ` in other areas?

For background, and FWIW, the changes merged in https://github.com/nextcloud/server/pull/39669 changed the timestamp formatting in `user:lastseen` from `d.m.Y H:i` to `Y-m-d H:i` without any controversy, and that's technically a breaking change for scripts.

I'm not trying to be difficult, just trying to understand the rationale.

core/Command/User/Info.php
nooblag: Okay, thanks. If it's hardcoded to always be UTC like that, I'm fine with leaving that as is, if that's what's desired for all `occ` times.

nooblag: A suggestion then, would be to make UTC explicit in all the times displayed, so the user knows that none of these are local times, as in OP https://github.com/nextcloud/server/issues/44641.

nooblag: Is there a reason for hardcoding UTC and ignoring PHP locale setting?

nickvergessen: > Is there a reason for hardcoding UTC and ignoring PHP locale setting?

this is done to ensure that all kind of apps and code places have a reliable base.

core/Command/User/Info.php
nickvergessen: > just trying to understand the rationale.

I missed it in the other PR.

core/Command/User/Info.php
nickvergessen: > A suggestion then, would be to make UTC explicit in all the times displayed

That's why we use `ATOM` normally. Not sure why it's not used on lastseen.

core/Command/User/Info.php
nooblag: Well, you'll also note there then that @come-nc wanted to do some other cool stuff like a complete overhaul of the timestamp line to show the date first to allow better sorting (which would also be good)... But anyways.

I'd obviously like to see some changes here to have consistency, which is the whole point of this discussion, but if that stalemates, I'll retract. I'd rather not introduce a `--timezone` argument that only applies to `user:info` and not `user:lastseen` -- my whole point here is looking for some consistency across `occ`, which at the moment, we don't have.

core/Command/User/Info.php
nooblag: Okay, thanks. Personally, I think `ATOM` is ugly and hard to read, which (I guess) is why it perhaps wasn't used on `lastseen` since I also guess that output is more meant to be human friendly anyways?

come-nc: > Okay, thanks. Personally, I think `ATOM` is ugly and hard to read, which (I guess) is why it perhaps wasn't used on `lastseen` since I also guess that output is more meant to be human friendly anyways?

Yes I think that is the reason

core/Command/User/Info.php
come-nc: But the timezone used is consistent accross `occ`, no? Only the date format is not always the same.

nooblag: Correct. :+1:

nooblag:
Just a friendly follow up on https://github.com/nextcloud/server/pull/44666#issuecomment-2080347746

come-nc:
> Hi all, so just wondering how to move forward with this, given it's been a few weeks since our discussion where we've landed on the decision around keeping everything in UTC. Just wondering about the other parts:
> 
> * Is the bug fix for `user:info` lastseen=0 okay?

Well, it’s complicated. I agree it would be better to show "never" rather than epoch, but changing that will break scripts relying on the output of the command.

> * Seems there's strong desire to keep the time and date format for `user:info` as `ATOM`. How about for `user:lastseen` though? Change that to ATOM too? Or leave it and just improve the "prettiness" (i.e. adding seconds and timezone)?

I have no strong opinion about adding seconds and timezone to `user:lastseen`.
Do you have before/after screenshots?

nooblag:
Hi all, just another friendly ping on https://github.com/nextcloud/server/pull/44666#issuecomment-2122395349.

nooblag:
Hi all, just another friendly ping on https://github.com/nextcloud/server/pull/44666#issuecomment-2122395349, please?

come-nc:
> Thanks for getting back.
> 
> > Well, it’s complicated. I agree it would be better to show "never" rather than epoch, but changing that will break scripts relying on the output of the command.
> 
> I appreciate and understand that. If we don't want to break things, then we can just leave it as it is. Or, as I've said [above](https://github.com/nextcloud/server/pull/44666#discussion_r1555529034), I'm open to suggestions on how to make changes in the least disruptive/most useful way.
> 
> PS what are some examples of all these scripts that are being referred to? And how is the importance of such 3rd party products considered against/above making changes to Nextcloud core?

I do not have a good answer for that. I do not have an example of a script, it’s just that there may be some out there and we should not break compatibility without a good reason.
Whether showing never instead of epoch is a good enough reason is hard to arbitrate. I’d say it should not be changed in the json output of the command.
Regarding the default output, yeah it’s nicer to see `never` but it’s not that bad to see epoch, no one has Nextcloud servers running since 1970 so it’s quite easy to understand that this means never.

> > I have no strong opinion about adding seconds and timezone to `user:lastseen`. Do you have before/after screenshots?
> 
> I would prefer to introduce seconds to `user:lastseen` along with the timezone to make it explicit to the user that the returned values are hardcoded UTC, and not local settings (I've been confused by this in the past, without the timezone information, thinking these times were local times). Sure, here are some screenshots.
> 
> Before PR:
> 
> Screenshot from 2024-05-21 21-12-52
> 
> Current PR:
> 
> Screenshot from 2024-05-21 21-12-23

That part looks nicer indeed and I’d be fine with merging that.

nooblag:
Thanks for responding.

Great, if we agree on the `--lastseen` changes, let's please merge them for NC 30?

And as I say, I won't keep pushing on the other, apart to still note in a friendly way, that it remains a bit confusing/dubious the reluctance over the tiniest breaking change for 3rd party scripts that are both uncited and still being considered above the level of core development...

Again, I'm no longer pushing, we can compromise here, and I'll remove it from the PR, but I do want to say that I think returning epoch date is not only incorrect data (even if no NC servers have been running since 1970), it remains my view that it's a bug, and it should say 'never', which is the correct data. But I've said all this already many times ;)

Anyway, will add a commit to reflect these discussions, and just accept the `--lastseen` change.

Thanks!

come-nc (approved):

come-nc:
CI failures unrelated, good for merging

AndyScherzinger (approved):
🐘

Issue #29678, opened 2021-11-12
22.2.1 is very slow


DmDS:
after upgrading to 22.2.1, nextcloud is very slow. after opening the page, everything is fine. after 1-2 minutes (possibly inactivity), the load on the web server increases greatly during rendering (I tried apache and nginx. php 7.4 and 8.0. and installed nextcloud from scratch). the site loads very slowly. if you log out of the account and log in again or restart the browser, it starts working fine, but after 1-2 minutes everything slows down again.
on a test machine with one user, this is still fast enough. in the working environment, it slows down much more.

<details>
 <summary>gif</summary>

![Запись_2021_11_12_14_24_27_953](https://user-images.githubusercontent.com/1803032/141460412-c436b2fc-0576-4e89-adb0-d64cd6b6402f.gif)
</details>

jacotec:
I copy this! Updated by "test cloud" and the "felt performance" dropped by around factor 10! Switching to a different app in the menu takes around 5 or more seconds, while it was "immediately" before.

I've rolled back the snapshot and I'm currently retesting. But for now I explicitely warn anyone from upgrading to 22.2.1 in production environments!

DmDS:
when disabling the Theming application, rendering is faster, but still much slower than in the first 1-2 minutes after opening the page. and the load on the web server is the same high.

jacotec:
Confirmed on a second instance!

Some times to compare on the instance before and after the upgrade (from clicking until all page elements have been loaded):

Changing from files view to the room overview of BigBlueButton:
22.2.0: 1,4 sec     /     22.2.1: 16,7 sec

Back to the file view:
22.2.0: 3.2 sec     /    22.2.1: 27,3 sec

To calendar view:
22.2.0: 3,8 sec   /    22.2.1: 30,2 sec

@szaimen Pulling back immediately the 22.2.1 release is a wise decision! ;-)

JB1985:
Are there any problems with 21.0.6?  Can anybody confirm this?

szaimen:
21.0.6 should work fine as https://github.com/nextcloud/server/pull/29380 wasn't merged yet

tdoerschel:
I can confirm this on apache 2.4.51 and PHP 8.0-FPM

nickvergessen:
Patch forward is at https://github.com/nextcloud/server/pull/29681

jacotec:
@nickvergessen Will there be a bugfix release (22.2.2) in short term, or what is the plan?

[...]

PR #44670, opened 2024-04-05
fix(session): Do not update authtoken last_check for passwordless

ChristophWurst:
## Summary

The column oc_authtoken.last_check is used to debounce the password checks of the login password encoded in app passwords. If instances have set `auth.storeCryptedPassword` to `false`, the password column stays NULL and there is no password check. In that case we keep the last_check column updated without any actual check performed. This is superfluous.

## How to test

1. Set `auth.storeCryptedPassword => false,` in config/config.php
2. Open the user security settings and generate an app password
3. Open the oc_authtoken table, find the last row and set last_check to 0
4. Run `curl -u "USER:APPPASSWORD" https://localhost/apps/files`
5. Look at the updated row in oc_authtoken

Master: last_check is set to the current timestamp. This indicates that there was an UPDATE query.
Here: last_check remains 0.

Note: in reality last_check will stay at the time of token creation. You can skip setting last_check to 0 but then you have to wait more than five minutes between creating and using the token.

## Checklist

### DIFF
--- a/lib/private/User/Session.php
+++ b/lib/private/User/Session.php
@@ -767,8 +767,6 @@ class Session implements IUserSession, Emitter {
 				return false;
 			}
 
-			$dbToken->setLastCheck($now);
-			$this->tokenProvider->updateToken($dbToken);
 			return true;
 		}
 
--- a/tests/lib/User/SessionTest.php
+++ b/tests/lib/User/SessionTest.php
@@ -11,9 +11,11 @@ namespace Test\User;
 use OC\AppFramework\Http\Request;
 use OC\Authentication\Events\LoginFailed;
 use OC\Authentication\Exceptions\InvalidTokenException;
+use OC\Authentication\Exceptions\PasswordlessTokenException;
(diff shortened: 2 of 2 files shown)


juliusknorr (approved):

nickvergessen:
Does this reintroduce https://github.com/nextcloud/server/issues/29678 ?

ChristophWurst:
> Does this reintroduce #29678 ?

Good catch but I don't think so. The old issue was that we missed to updated last_check when we actually did check the password. Here we skip the update intentionally because no checks are performed.

For the path of password checks the token update is still in place: https://github.com/nextcloud/server/blob/cc42b2eadb47f3424c59e3993251471ac92d5400/lib/private/User/Session.php#L787

The unit test added covers an assertion for the token update.

Altahrim (approved):

ChristophWurst:
> > Does this reintroduce #29678 ?
> 
> Good catch but I don't think so.

Tested by setting the last_activity to 0 and the timestamp *is* updated. :shipit:

PR #28077, opened 2021-07-21
Disable HEIC image preview provider for performance concerns

tobiasKaminsky:
Signed-off-by: tobiasKaminsky <tobias@kaminsky.me>

PR #44710, opened 2024-04-08
fix(preview): check mime type before processing with Imagick

pulsejet:


lib/private/Preview/Bitmap.php
```diff
@@ -57,7 +66,7 @@ public function getThumbnail(File $file, int $maxX, int $maxY): ?IImage {
 			\OC::$server->get(LoggerInterface::class)->info(
 				'File: ' . $file->getPath() . ' Imagick says:',
 				[
-					'exception' => $e,
+					'exception' => $e->getMessage(),
```
nickvergessen, line 69: This removes stakes from the logs. We should keep it the old way

pulsejet: Reverted. My logs were empty so I assumed it was a typo.

artonge (approved):

lib/private/Preview/Bitmap.php
```diff
@@ -37,6 +37,15 @@
  * @package OC\Preview
  */
 abstract class Bitmap extends ProviderV2 {
+	/**
+	 * List of MIME types that this preview provider is allowed to process.
+	 *
+	 * @return string Regular expression
+	 */
+	public function getAllowedMimeTypes(): string {
```
nickvergessen, line 45: The methods are not used "outside", so they could all be `protected` ?

pulsejet: Good catch, done

pulsejet:
I made one change: the default behavior of the `Bitmap` class is now to match **nothing** rather than **everything**. As a result, any derived classes will not match anything unless `getAllowedMimeTypes` is defined (the breaking change is intentional). Let me know if this isn't desirable for any strong reason.

lib/private/Preview/Bitmap.php
```diff
@@ -90,6 +99,13 @@ public function getThumbnail(File $file, int $maxX, int $maxY): ?IImage {
 	private function getResizedPreview($tmpPath, $maxX, $maxY) {
```
come-nc, line 99: Should be tagged `@throws \Exception`

lib/private/Preview/Bitmap.php
```diff
@@ -90,6 +99,13 @@ public function getThumbnail(File $file, int $maxX, int $maxY): ?IImage {
...
 		$bp = new Imagick();
 
+		// Validate mime type
+		$bp->pingImage($tmpPath . '[0]');
```
come-nc, line 103: Is it useful to do that instead of doing readImage first and checking mimeType after? We are reading part of the image twice only to avoid reading it when mimetype is invalid, which should be the rare case, no?

lib/private/Preview/IMagickSupport.php
```diff
@@ -37,4 +37,8 @@ public function supportsFormat(string $format): bool {
 		$this->cache->set($format, $cached);
 		return $formatSupported;
 	}
+
+	public static function isSafeMimeType(string $mimeType): bool {
+		return preg_match('/^image\/(png|jpeg|gif|bmp|tiff|webp|hei(f|c))$/', $mimeType) > 0;
```
come-nc, line 42: application/ or x- prefix are not needed here?

lib/private/Preview/Bitmap.php
```diff
@@ -37,6 +37,15 @@
...
 abstract class Bitmap extends ProviderV2 {
+	/**
+	 * List of MIME types that this preview provider is allowed to process.
+	 *
+	 * @return string Regular expression
+	 */
+	protected function getAllowedMimeTypes(): string {
+		return '/invalid\/invalid/';
+	}
+
```
come-nc, line 48: What is the difference with `getMimeType()` ?
https://nextcloud-server.netlify.app/classes/ocp-preview-iprovider#method_getMimeType

lib/private/Preview/Bitmap.php
pulsejet: We should not read the image until we know it's okay; it might have some RCE exploit. `pingImage` does exactly this.

lib/private/Preview/IMagickSupport.php
pulsejet: Ehh I thought I removed this. It's gone now.

lib/private/Preview/Bitmap.php
pulsejet: `getMimeType` specifies the types we want this class to handle. The new method specifies the types we allow passing to `Imagick` (which might be completely different since Imagick doesn't necessarily use standard types).

lib/private/Preview/Bitmap.php
pulsejet: Done

lib/private/Preview/Bitmap.php
pulsejet: One addition: I don't think `pingImage` reads the whole image anyway; it probably only looks at the header.

lib/private/Preview/Bitmap.php
nickvergessen: Specify in the docs that this mimetype is "identified" by Imagick and therefore can have "non-clean" versions **not** following our overwrite maps from `OCP\Files\IMimeTypeDetector`

come-nc: Could still default to getMimeType() rather than invalid, no?

lib/private/Preview/Bitmap.php
come-nc: Yeah but you are now reading the header twice? My question is whether invalid mimetype is supposed to be common or exceptional, if it’s the second one we should not pre-read the header on all calls for it.

lib/private/Preview/Bitmap.php
```diff
@@ -37,6 +37,20 @@
...
+	 * These should correspond to the MIME types *identified* by Imagemagick
+	 * for files to be processed by this provider. These do / will not
+	 * necessarily need to match the MIME types stored in the database
+	 * (which are identified by IMimeTypeDetector).
+	 *
+	 * @return string Regular expression
+	 */
+	protected function getAllowedMimeTypes(): string {
+		return '/invalid\/invalid/';
+	}
```
come-nc, line 52:
```suggestion
	abstract protected function getAllowedMimeTypes(): string;
```
This is the cleanest actually

lib/private/Preview/Bitmap.php
come-nc: See https://github.com/nextcloud/server/pull/44710#discussion_r1555581896

lib/private/Preview/Bitmap.php
pulsejet: Done. My original intention was to not fail with a TypeError if e.g. a third party app uses this but now that I think that's a better failure mode.

lib/private/Preview/Bitmap.php
pulsejet: Changed to abstract.

pulsejet:
Bump

pulsejet:
Bump (2)

nickvergessen:
yeah, there was a freeze recently for updates, so this could not proceed with all necessary energy

nickvergessen (approved):

lib/private/PreviewManager.php
```diff
@@ -376,9 +376,9 @@ protected function registerCoreProviders() {
 				'PSD' => ['mimetype' => '/application\/x-photoshop/', 'class' => Preview\Photoshop::class],
 				'EPS' => ['mimetype' => '/application\/postscript/', 'class' => Preview\Postscript::class],
 				'TTF' => ['mimetype' => '/application\/(?:font-sfnt|x-font$)/', 'class' => Preview\Font::class],
-				'HEIC' => ['mimetype' => '/image\/hei(f|c)/', 'class' => Preview\HEIC::class],
-				'TGA' => ['mimetype' => '/image\/t(ar)?ga/', 'class' => Preview\TGA::class],
-				'SGI' => ['mimetype' => '/image\/sgi/', 'class' => Preview\SGI::class],
+				'HEIC' => ['mimetype' => '/image\/(x-)?hei(f|c)/', 'class' => Preview\HEIC::class],
+				'TGA' => ['mimetype' => '/image\/(x-)?t(ar)?ga/', 'class' => Preview\TGA::class],
+				'SGI' => ['mimetype' => '/image\/(x-)?sgi/', 'class' => Preview\SGI::class],
```
Altahrim, line 381: Do you think it's possible to have `getMimeType` as `static` so we can avoid to duplicate the PCRE here?
```php
				'SGI' => ['mimetype' => Preview\SGI::getMimeType(), 'class' => Preview\SGI::class],
```

lib/private/Preview/Bitmap.php
```diff
@@ -37,6 +37,18 @@
...
+	 * List of MIME types that this preview provider is allowed to process.
+	 *
+	 * These should correspond to the MIME types *identified* by Imagemagick
+	 * for files to be processed by this provider. These do / will not
+	 * necessarily need to match the MIME types stored in the database
+	 * (which are identified by IMimeTypeDetector).
+	 *
+	 * @return string Regular expression
+	 */
+	abstract protected function getAllowedMimeTypes(): string;
```
susnux, line 50: Should this not be in a common interface?
So it would also be available in `SVG`?

lib/private/PreviewManager.php
pulsejet: I don't mind, but that will be a breaking change (will break third party preview apps) since this method is abstract, with minimal benefit.

lib/private/Preview/Bitmap.php
pulsejet: It's only relevant for providers using Imagick, so it can't be in `ProviderV2`. We could make another abstract class like `ImagickProvider`, but that seems pointless to me.

susnux: Sounds reasonable :)

lib/private/PreviewManager.php
Altahrim: It's on the private part so I think we can go with it. I'd like other inputs

pulsejet: With all due respect -- I just want to point out that this is a security patch, and it isn't the right place for (breaking) refactors. The fact that this has been sitting for 3 weeks is in itself alarming to me.
(note: I can't merge because of CI failure)

cc @nickvergessen

nickvergessen: Well then it should have been reported properly up front so we can coordinate it. Next releases are towards end of May, but yeah should make sure it's part of it.

pulsejet: It was reported 3 years ago, disclosed yet never fixed. Regardless, my point is not about that. I'm just saying nitpicking for refactors in a security patch is a waste of time.

Altahrim: It was not meant to block the PR.
I noticed something, I wanted to open the discussion about it. It can go in another PR, in `/dev/null` or in this one depending of different wills :)

Altahrim (approved):

vermeeren:
Hi @nickvergessen,

Recently I'm starting to use the HEIC format and saw the image previews are not working by default. I did some searching and found a nasty CVE from 2021 https://hackerone.com/reports/1261413 which was fixed by disabling the HEIC preview in https://github.com/nextcloud/server/pull/28077. With this change and checking the mime type it does seem safe to me to enable HEIC, but I do notice the default is still to keep it disabled.

Could you confirm whether it is considered secure to enable HEIC previews in the latest Nextcloud versions? I would love to have previews but of course not at the cost of a potentially critical security issue.

Many thanks!

PR #44714, opened 2024-04-08
Call davGetFavoritesReport after the registration of propfind properties

artonge:
Before | After
-- | --
image | Screenshot from 2024-04-08 10-58-00


susnux (approved):
Really important and nice catch!

susnux (approved):

Pytal (approved):

PR #44718, opened 2024-04-08
Move retry logic to the QueryBuilder

artonge:
Follow-up of https://github.com/nextcloud/server/pull/44504

### DIFF
--- a/lib/public/DB/QueryBuilder/IQueryBuilder.php
+++ b/lib/public/DB/QueryBuilder/IQueryBuilder.php
@@ -185,6 +185,13 @@ interface IQueryBuilder {
 	public function executeQuery(): IResult;
 
 	/**
+	 * @since 30.0.0
+	 *
+	 * @param int $retryLimit - Retry the query up to $retryLimit times in case of a RetryableException. Won't retry by default.
+	 */
+	public function setRetryLimit(int $retryLimit): void;
+
+	/**
 	 * Execute insert, update and delete statements
 	 *
 	 * @return int the number of affected rows
--- a/lib/private/DB/QueryBuilder/QueryBuilder.php
+++ b/lib/private/DB/QueryBuilder/QueryBuilder.php
@@ -30,6 +30,7 @@
  */
 namespace OC\DB\QueryBuilder;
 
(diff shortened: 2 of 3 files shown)


lib/public/DB/QueryBuilder/IQueryBuilder.php
```diff
@@ -184,6 +184,13 @@ public function execute();
 	 */
 	public function executeQuery(): IResult;
 
+	/**
+	 * @since 30.0.0
+	 *
+	 * @param int $retryLimit - Retry the query up to $retryLimit times in case of a RetryableException. Initially equal to 4.
```
nickvergessen, line 190: I don't think we should have retries by default :-X

artonge: Unsure about it either, but note that this is only for `RetryableException`, we don't try forever, and there is a log, so it won't go unnoticed.

Not sure what the argument against would be, that is why I set the default to 4. But I do not have any strong opinion, so I am fine either way.

come-nc (approved):

lib/public/DB/QueryBuilder/IQueryBuilder.php
juliushaertl: Retries only work if the whole transaction is retried, so I'm also more in favor of making this an opt-in to not cause unexpected behaviour there when developers don't think about this and just call execute.

artonge: > Retries only work if the whole transaction is retried

So it does not really make sense to have that logic here, no? I think I start to understand the value of https://github.com/nextcloud/server/blob/050c6d53b38d6737a49c13d4e6a20a6062160503/lib/public/AppFramework/Db/TTransactional.php#L94 :sweat_smile:

icewind1991 (changes requested):
I think retrying by default is a bad idea and can lead to unexpected behavior for developers

lib/public/DB/QueryBuilder/IQueryBuilder.php
artonge: Default is now set to 0

PR #44763, opened 2024-04-10
fix(groups): allows to save group names with more than 64 characters

Altahrim:
## Summary

Mimic behaviour from LDAP users and add a hard limit to 255 characters

## Checklist

### DIFF
--- a/lib/private/Group/Database.php
+++ b/lib/private/Group/Database.php
@@ -37,7 +37,7 @@ use OCP\Group\Backend\IAddToGroupBackend;
 use OCP\Group\Backend\IBatchMethodsBackend;
 use OCP\Group\Backend\ICountDisabledInGroup;
 use OCP\Group\Backend\ICountUsersBackend;
-use OCP\Group\Backend\ICreateGroupBackend;
+use OCP\Group\Backend\ICreateNamedGroupBackend;
 use OCP\Group\Backend\IDeleteGroupBackend;
 use OCP\Group\Backend\IGetDisplayNameBackend;
 use OCP\Group\Backend\IGroupDetailsBackend;
@@ -55,7 +55,7 @@ class Database extends ABackend implements
 	IAddToGroupBackend,
 	ICountDisabledInGroup,
 	ICountUsersBackend,
-	ICreateGroupBackend,
+	ICreateNamedGroupBackend,
 	IDeleteGroupBackend,
 	IGetDisplayNameBackend,
 	IGroupDetailsBackend,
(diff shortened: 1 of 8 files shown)


lib/private/Group/Database.php
```diff
@@ -88,21 +88,23 @@ private function fixDI() {
...
-	 * @param string $gid The name of the group to create
+	 * @param string $displayName The name of the group to create
 	 * @return bool
+	 * @psalm-suppress ParamNameMismatch
 	 *
 	 * Tries to create a new group. If the group name already exists, false will
 	 * be returned.
 	 */
-	public function createGroup(string $gid): bool {
+	public function createGroup(string $displayName): bool {
```
come-nc, line 98: Nope.

This method is in ICreateGroupBackend and it specifies gid there.
Parameter name is part of signature since named parameter are a thing in PHP, and also changing the meaning of the parameter is bad practice.
So, maybe add an ICreateNamedGroupBackend with a method to create a group by its display name rather than gid, or create first and rename after?

Altahrim: I added a new interface. 
I wanted to avoid to change behaviour for external backends

lib/private/Group/Manager.php
```diff
@@ -281,12 +284,22 @@ public function createGroup($gid) {
 			return null;
 		} elseif ($group = $this->get($gid)) {
 			return $group;
+		} elseif (mb_strlen($gid) > self::MAX_GROUP_LENGTH) {
+			throw new \Exception('Group name is limited to '. self::MAX_GROUP_LENGTH.' characters');
```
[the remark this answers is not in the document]
come-nc: According to my tests, `mb_strlen` will always count 1 for 4 bytes max it seems, combining characters are counted separately. So `$gid` here will not take more that 1024 bytes.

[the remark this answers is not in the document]
Altahrim: In UTF-8, characters are using up to 4 bytes yes :) 
Database allows up to 255 characters and not bytes, so I kept that

come-nc (approved):
I think I would prefer `string|false` instead of `?string`, or `string` and exception throwing. I did not check what is most consistent with the rest of OCP though.

artonge (approved):

lib/public/Group/Backend/ICreateGroupBackend.php
```diff
@@ -27,6 +27,7 @@
 
 /**
  * @since 14.0.0
+ * @deprecated 30.0.0 Use ICreateNamedGroupBackend instead
```
nickvergessen, line 30: @Altahrim Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(all changes in lib/public/)

PR #44767, opened 2024-04-10
fix(LDAP): do not count mapped users x-times active configs

blizzz:
## Summary

the method counts all rows in the mapping tables, which are not specific to single configurations, but contains users from all

So if you are operating on a one-click host with a user limit of 500, but have two LDAPs connected with altogether 300 users, without this fix they are interpreted as 600 users, and the 301st could not log in.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/User_Proxy.php
+++ b/apps/user_ldap/lib/User_Proxy.php
@@ -415,12 +415,7 @@ class User_Proxy extends Proxy implements IUserBackend, UserInterface, IUserLDAP
 	 */
 	public function countMappedUsers(): int {
 		$this->setup();
-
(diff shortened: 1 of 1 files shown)


come-nc (commented):
Then the method should be moved out of User_LDAP and directly in User_Proxy if it’s not specific to a connection.

blizzz:
> Then the method should be moved out of User_LDAP and directly in User_Proxy if it’s not specific to a connection.

Actually yes, and it crossed my mind yesterday as well. Revisiting this, I am not sure anymore why I decided against it – possibly to avoid injecting another dependency. But actually the abstract Proxy class pulls it in anyway (though on demand).

PR #44770, opened 2024-04-10
fix(notifications): Improved notification exceptions

nickvergessen:
## Summary

Currently all methods in the notification space throw `\InvalidArgumentException`s. The problem is that there are nested calls which make it impossible to handle errors correctly or to notify wrong/invalid notifiers properly.

E.g. an app calls `INotification::setLink()` with an invalid value in it's `INotifier::prepare()`. The `INotification` will throw and it can go unnoticed by the `INotifier` as it's not meant to throw. However the `IManager` only sees that `INotifier` threw the exception and that is currently meant to be the case when the `INotifier` does not want to handle the `INotification`, so the IManager will think no one wanted to handle the notification although the respective one just had "an" error from any level. Similar problems often appear when activities are put into notifications as both providers work based on `\InvalidArgumentException` so none of those knows whether activity or notifications was failing actually.

This PR introduces dedicated exceptions which allow us to log them if they happened while not being expected. For the moment `\InvalidArgumentException` is still allowed to be thrown, and only logged on debug level, but the `lib/private/Notification` will only throw the new exceptions (which all extend the generic `\InvalidArgumentException` to keep compatibility.

- `IncompleteNotificationException` - is thrown by `IApp::notify` when a required field was not set
- `IncompleteParsedNotificationException` - is thrown by `INotifier::prepare` when it's `IManager` and a required field was not set during the rendering
- `InvalidValueException` - is thrown by `IAction::set*` and `INotification::set*` when setting an invalid value (e.g. empty link, missing rich-object parameter, ...)
- `UnknownNotificationException` - should be thrown by `INotifier::prepare` when the notification is not handled by this notifier

## Companion / Sample PRs

- https://github.com/nextcloud/notifications/pull/1882
- https://github.com/nextcloud/documentation/pull/11729

## Checklist


lib/public/Notification/IAction.php
```diff
@@ -68,8 +69,9 @@ public function getLabel(): string;
 	/**
 	 * @param string $label
 	 * @return $this
-	 * @throws \InvalidArgumentException if the label is invalid
+	 * @throws InvalidValueException if the label is invalid
 	 * @since 9.0.0
+	 * @since 30.0.0 throws {@see \OCP\Notification\InvalidValueException} instead of \InvalidArgumentException
```
come-nc, line 74:
```suggestion
	 * @since 30.0.0 throws {@see InvalidValueException} instead of \InvalidArgumentException
```
Or does the `@see` need fqdn?

lib/public/Notification/IDismissableNotifier.php
```diff
@@ -30,16 +30,19 @@
...
 interface IDismissableNotifier extends INotifier {
 	/**
 	 * @param INotification $notification
-	 * @throws \InvalidArgumentException In case the handler can't handle the notification
+	 * @throws UnknownNotificationException when the notifier is not in charge of the notification
 	 *
 	 * @since 18.0.0
+	 * @since 30.0.0 Notifiers should throw {@see UnknownNotificationException} instead of \InvalidArgumentException
+	 *  when they did not handle the notification. Throwing \InvalidArgumentException directly is deprecated and will
+	 *  be logged as an error in a future version.
```
come-nc, line 45: Would it be possible to say which version will log an error?

lib/public/Notification/IncompleteNotificationException.php
```diff
@@ -0,0 +1,43 @@
...
+ * - app
+ * - user
+ * - dateTime
+ * - objectType
+ * - objectId
+ * - subject
+ *
+ * @since 30.0.0
+ */
+class IncompleteNotificationException extends \InvalidArgumentException {
```
come-nc, line 42: Are we sure we do not want a base class for all notification related exceptions? That would allow to group them all in a `catch(NotificationException)` when suitable for instance.

lib/public/Notification/InvalidValueException.php
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * @since 30.0.0
+ */
+class InvalidValueException extends \InvalidArgumentException {
+	/**
+	 * @since 30.0.0
+	 */
+	public function __construct(
+		protected string $field,
```
come-nc, line 37: What, no `Enum` :clown_face: ?

lib/public/Notification/InvalidValueException.php
```diff
@@ -0,0 +1,49 @@
...
+		?\Throwable $e = null,
```
come-nc, line 38:
```suggestion
		?\Throwable $previous = null,
```

lib/public/Notification/IncompleteNotificationException.php
nickvergessen: I find it kind of weird, as they are thrown on different levels and should be caught on different levels again.
Also the already existing exceptions in the Notification namespace have a different base Exception 🫣

lib/private/Notification/Manager.php
```diff
@@ -300,21 +303,24 @@ public function isFairUseOfFreePushService(): bool {
...
 	/**
-	 * @param INotification $notification
-	 * @throws \InvalidArgumentException When the notification is not valid
-	 * @since 8.2.0
+	 * {@inheritDoc}
 	 */
 	public function notify(INotification $notification): void {
 		if (!$notification->isValid()) {
-			throw new \InvalidArgumentException('The given notification is invalid');
+			throw new IncompleteNotificationException('The given notification is invalid');
```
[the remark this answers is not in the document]
nickvergessen: Should be invoked by the manager? Could do that, but it's an unnecessary detail I think. It's only helping on initial development. Never saw an app conditionally doing this wrong 🫣

lib/private/Notification/Manager.php
```diff
@@ -300,21 +303,24 @@ public function isFairUseOfFreePushService(): bool {
...
 		}
 
 		$apps = $this->getApps();
 
 		foreach ($apps as $app) {
 			try {
 				$app->notify($notification);
+			} catch (IncompleteNotificationException) {
 			} catch (\InvalidArgumentException $e) {
+				if (get_class($e) === \InvalidArgumentException::class) {
```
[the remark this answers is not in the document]
nickvergessen: My idea was to specifically only log actual thrown `InvalidArgumentException`, not anything extending it, but maybe that's not a clever idea 🤔

lib/public/Notification/IDismissableNotifier.php
nickvergessen: 33 or 39, depending on whether we think this is a BC break or not 🫣

lib/public/Notification/IAction.php
nickvergessen: No, just had it in a wrong folder before and using the refactor in phpstorm didn't remove the fqdn

bigcat88 (approved):
looks good

lib/private/Notification/Manager.php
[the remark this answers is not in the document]
come-nc: Hum nothing prevented throwing a subclass before, so no reason to log specifically on parent class.

come-nc (approved):

lib/private/Notification/Manager.php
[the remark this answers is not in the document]
nickvergessen: Yeah, but we would be fine with that. Already allows to distinguish the currently bad interference of notifications and activities if used as advertised.
But anyway I changed it

PR #44786, opened 2024-04-11
Reduce app container border radius

marcoambrosini:
Screenshot 2024-04-11 at 10 23 56


marcoambrosini:
/ compile amend/

marcoambrosini:
/ compile/

emoral435 (commented):
Will hard-coding introduce any new regressions for designs around the server? If there is a specific spot that we want to handle, should we not change the radius there instead? I am just curious for the reason, if its something that the designers have decided then I am fine with this :)

jancborchardt (commented):
@marcoambrosini what @emoral435 is saying though is that setting it to 12px only here is a bit strange. Do we never use 12px as radius anywhere else? Shouldn’t it be using a variable, maybe even just 3*default-grid-baseline?

emoral435 (approved):
Much nicer :) Thank you!

jancborchardt (approved):
Looks good! :)

jancborchardt:
@nimishavijay @susnux @artonge do you have any further feedback? Otherwise let’s get this in while we are still early in the cycle. :)

susnux (approved):
code looks good, if this was discussed by the design team then :+1:

PR #44791, opened 2024-04-11
fix: consider all paths the user has for a share source when considering max permissions

icewind1991:
To test

- Create an external storage "Storage" for all users with sharing enabled
- As `user1` create folder "Storage/folder" and share it read-only with `user2`
- As `user2` attempt to share "Storage/folder" with edit permissions to `user3`

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -45,7 +45,6 @@ use OC\Files\Mount\MoveableMount;
 use OC\KnownUser\KnownUserService;
 use OC\Share20\Exception\ProviderException;
 use OCA\Files_Sharing\AppInfo\Application;
-use OCA\Files_Sharing\ISharedStorage;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Files\File;
 use OCP\Files\Folder;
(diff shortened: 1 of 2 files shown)


icewind1991:
> when sharing Folder/Sub as user2, the api call has /Sub as the path.

I'm pretty sure this is because the share ui knows it's a reshare so this should be fine

Altahrim (approved):

come-nc (approved):
I cannot find it but you made a PR not long ago to use the node the user tried to share instead of the first one when checking reshare permission, is that a different fix for the same thing or two different cases?

come-nc:
> I cannot find it but you made a PR not long ago to use the node the user tried to share instead of the first one when checking reshare permission, is that a different fix for the same thing or two different cases?

Ok I’m not crazy I think you forced push over it in this very PR :see_no_evil:

icewind1991:
> > I cannot find it but you made a PR not long ago to use the node the user tried to share instead of the first one when checking reshare permission, is that a different fix for the same thing or two different cases?
> 
> Ok I’m not crazy I think you forced push over it in this very PR 🙈

Yeah that approach didn't end up working with in cases

PR #44794, opened 2024-04-11
fix(memcache): remove double $$ to fix error

Velwark:
Extra Dollar Sign caused errors in Nextcloud. Removing the Dollar Sign Solved the Problem.

* Resolves: # 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Memcache/LoggerWrapperCache.php
+++ b/lib/private/Memcache/LoggerWrapperCache.php
@@ -75,7 +75,7 @@ class LoggerWrapperCache extends Cache implements IMemcacheTTL {
 			FILE_APPEND
 		);
 
-		return $this->wrappedCache->set($key, $value, $$ttl);
+		return $this->wrappedCache->set($key, $value, $ttl);
 	}
 
 	/** @inheritDoc */


solracsf (approved):

tcitworld (approved):

susnux (approved):

come-nc (approved):

come-nc:
Nice catch.
This made search for $$ in our codebase, I think https://github.com/nextcloud/server/blob/fea0278947cafcd62877226dc5ddece04d596a98/lib/private/DB/QueryBuilder/QueryBuilder.php#L980 is also a typo, no?

Velwark:
> Nice catch. This made search for `$$` in our codebase, I think
> 
> https://github.com/nextcloud/server/blob/fea0278947cafcd62877226dc5ddece04d596a98/lib/private/DB/QueryBuilder/QueryBuilder.php#L980
> 
> is also a typo, no?

I‘m really no PHP expert myself. The Double Dollar Sign is a Variable Variable in PHP. As Long as ist doesn‘t throw any erros that could just be right.

The Double Dollar sign here in the Commit caused lots of errors on my Instance.

come-nc:
> > Nice catch. This made search for `$$` in our codebase, I think
> > https://github.com/nextcloud/server/blob/fea0278947cafcd62877226dc5ddece04d596a98/lib/private/DB/QueryBuilder/QueryBuilder.php#L980
> > 
> > is also a typo, no?
> 
> I‘m really no PHP expert myself. The Double Dollar Sign is a Variable Variable in PHP. As Long as ist doesn‘t throw any erros that could just be right.

I think `$groupBys` is an array there, so it makes no sense. I’ll try to find time to debug&fix.

> The Double Dollar sign here in the Commit caused lots of errors on my Instance.

Note that I think that means you have profiling enabled, you should disable it outside of dev instances. Check with `occ config:system:get profiler`.

> By the way, how do i merge this commit?

You do not, we need to wait for someone with the power to do so to force merge, because the CI cannot pass on PRs coming from forks.

AndyScherzinger:
Thanks for your contribution @Velwark 🎉

Highly appreciated 💯🙏🎉

come-nc:
> Looking at
> 
> https://github.com/nextcloud/server/blob/fea0278947cafcd62877226dc5ddece04d596a98/lib/private/DB/QueryBuilder/QueryBuilder.php#L952
> 
> it seems to be a bug in line 980

fixed it in https://github.com/nextcloud/server/pull/44533/commits/6953d1f9f1e570af775ee90ce4764d785e72bcb3 in the psalm bump PR which I intend to get merged at some point.

joshtrichards:
> The Double Dollar sign here in the Commit caused lots of errors on my Instance.

@Velwark What were the errors you were seeing/under what conditions? Asking since this PR wasn't connected to prior bug report and just to get it on record for searches/etc. Thanks!

nickvergessen:
Well it triggered a warning depending on PHP config because of the use of `Undefined variable $300` etc.

joshtrichards:
:+1: I just realized this is in the wrapper (so only when `redis_log_file` is configured).

Probably a less traveled code path. Explains why there weren't a bunch of related bug reports.

PR #44797, opened 2024-04-12
fix(Router): Stop loading routes of disabled apps

provokateurin:
## Summary

Routes of disabled apps should not be loaded. The classes are not loaded by the autoloader and the request will fail trying to use reflection on them.

## Checklist

### DIFF
--- a/lib/private/Route/Router.php
+++ b/lib/private/Route/Router.php
@@ -154,7 +154,7 @@ class Router implements IRouter {
 		}
 		$this->eventLogger->start('route:load:' . $requestedApp, 'Loading Routes for ' . $requestedApp);
 
-		if ($requestedApp !== null) {
+		if ($requestedApp !== null && in_array($requestedApp, \OC_App::getEnabledApps())) {
 			$routes = $this->getAttributeRoutes($requestedApp);
 			if (count($routes) > 0) {
 				$this->useCollection($requestedApp);


provokateurin:
/backport f523e29041bb9f1c1b96473b9dfd05bce32fd159 to stable28

lib/private/Route/Router.php
```diff
@@ -208,7 +210,7 @@ public function loadRoutes($app = null) {
 			$collection->addPrefix('/ocs');
 			$this->root->addCollection($collection);
 		}
-		$this->eventLogger->end('route:load:' . $requestedApp);
+		$this->eventLogger->end('route:load:' . $app);
```
nickvergessen, line 213: `$app` is overwritten by the for-loop inbetween

provokateurin: Ah didn't see that...

nickvergessen (approved):

provokateurin:
For some reason the comments integration test fails reliably because of 404 instead of 200. Seems like this patch is blocking too much? But if the dashboard app is not disabled the test should have never worked.

nickvergessen:
Let's not backport it to older versions in this case.
29 we can to after the final and patch it locally before hand to get insights

susnux (approved):

provokateurin:
I limited the check to the part that is responsible for parsing the attributes as blocking in any case was breaking the comments integration tests (which also seems to be a bug but not in scope of the fix).

nickvergessen (approved):

BJKle:
will this be in next release 29.0.1?
because currently my log file gets filled with errors "OCA\\UserStatus\\Controller\\HeartbeatController" whenever someone uses an app.
see: https://github.com/nextcloud/ios/issues/2927

provokateurin:
Yes, this will appear in 29.0.1

philippe-ritter:
Hi all,
even with version 29.0.1, and I checked in Router.php that I have the latest change, I got this error :
  "reqId": "Zl9qwRitzrnfd2ZKakQ-gAAAA1I",
  "level": 3,
  "time": "2024-06-04T19:28:01+00:00",
  "remoteAddr": "",
  "user": "phr",
  "app": "core",
  "method": "GET",
  "url": "/Nuage/index.php/apps/files/",
  "message": "Exception thrown: ReflectionException",
  "userAgent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/125.0.0.0 Safari/537.36",
  "version": "29.0.1.1",
  "exception": {
    "Exception": "ReflectionException",
    "Message": "Class \"OCA\\Files\\Controller\\AjaxController\" does not exist",
    "Code": -1,
    "Trace": [
      {
        "file": "/home/clients/659aacf688b6ffdff44c2c628e43d209/web/famille-ritter.ch/Nuage/lib/private/Route/Router.php",
        "line": 469,
        "function": "__construct",
        "class": "ReflectionClass",
        "type": "->",
        "args": [
          "OCA\\Files\\Controller\\AjaxController"
        ]
      },

What could cause this error ?

provokateurin:
Yeah I was about to say that this file doesn't even exist anymore in the codebase :see_no_evil:

kellergoech:
Thanks @philippe-ritter your comment (deleting **apps/files/lib/Controller/AjaxController.php**) and deleting 

- apps/files_sharing/appinfo/app.php
- apps/files_external/appinfo/app.php

Fixed the issue for me.

silopolis:
Thanks @philippe-ritter deleting `apps/files/lib/Controller/AjaxController.php` worked for me too, and I suspect we're (all ?) using the same hosting...

Stratosfear1:
> Thanks @philippe-ritter deleting `apps/files/lib/Controller/AjaxController.php` worked for me too, and I suspect we're (all ?) using the same hosting...



> Thanks @philippe-ritter deleting `apps/files/lib/Controller/AjaxController.php` worked for me too, and I suspect we're (all ?) using the same hosting...

Thanks a lot. Had the same issue. Solution of @philippe-ritter, deleting AjaxController.php, worked for me too. File was from 2022.

nickvergessen:
The integrity check should tell you that already.

Also it points to a wrong/problematic update strategy.
See steps 6-11 of https://docs.nextcloud.com/server/latest/admin_manual/maintenance/manual_upgrade.html

PR #44818, opened 2024-04-14
Enable more psalm rules

provokateurin:
## Summary

Enables a few more useful rules:
- MissingThrowsDocblock: Ensures every method has the appropriate `@throws` annotations
- InvalidOperand: Ensures values/variables of different types are not used in an operation without explicit casting (e.g. `'test' . 123`
- UnusedPsalmSuppress: Reports on unused supresses

I only fixed the problems of the last rule because there weren't a lot, but the first two ones I added to the baseline. Those errors will only show up for new or refactored code.

## Checklist

### DIFF
--- a/core/Controller/OCMController.php
+++ b/core/Controller/OCMController.php
@@ -58,6 +58,7 @@ class OCMController extends Controller {
 	 *
 	 * @PublicPage
 	 * @NoCSRFRequired
+	 * @psalm-suppress UnusedPsalmSuppress
 	 * @psalm-suppress MoreSpecificReturnType
 	 * @psalm-suppress LessSpecificReturnStatement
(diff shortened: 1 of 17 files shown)


nickvergessen (changes requested):
> MissingThrowsDocblock: Ensures every method has the appropriate @throws annotations

I do not agree with this.
Only expected throws should be documented.
Any method that does do anything with the DB, Redis, Ldap, ... would otherwise document to throw them and every consumer starts to wrap hundreds of catches or has their IDE crying about unhandled exceptions.

st3iny (changes requested):
I agree with Joas. Explicit error handling like in Go and Rust is nice but it's already to late to enforce that in our codebase.

Only relevant exceptions should be listed to make the caller aware of possible errors. The decision is up to the developer.

core/Controller/OCMController.php
```diff
@@ -58,6 +58,7 @@ public function __construct(
 	 *
 	 * @PublicPage
 	 * @NoCSRFRequired
+	 * @psalm-suppress UnusedPsalmSuppress
```
come-nc, line 61: :upside_down_face:

provokateurin: Yeah this piece of code is somehow very broken :grimacing:

PR #44819, opened 2024-04-14
fix(files): apply the renaming when leaving the input field

### LINKED ISSUE #42689: Validate renaming by clicking outside the name

I saw somebody trying to rename a file by typing the new name and then clicking outside the input.
They expected the rename to be taken in account. This would be consistent with most file managers I know, including Dolphin (KDE), and most probably Mac's Finder too.

The actual result is that the rename is canceled. One needs to type Enter for the file to be actually renamed. This person actually thought Nextcloud was having an issue. They actually tried 3 times in a row.

The rename should be taken in account when clicking outside the file input.

None

**Additional context**

https://github.com/nextcloud/server/assets/3817365/b1207ffb-b702-4589-890f-06983327423d

https://github.com/nextcloud/server/assets/3817365/15752da3-4a62-4574-81ba-72bf2d93ee35

See also https://github.com/nextcloud/server/issues/44832

raphj:
* Resolves: #42689

## Summary

With this change, clicking outside during rename behaves as in any other file manager and applies the rename instead of discarding it

## TODO

- [ ] The Nextcloud team to check that the change is indeed wanted

## Checklist


jancborchardt (commented):
Sounds ok to me. Additionally I think we need a simple "checkmark" icon-only button on the right inside the rename input field so people can decidedly confirm via a mouse click and not be confused about how to confirm.

emoral435:
> This should be done in a separate PR, right?
Yes, I think that that this should be a separate PR, with its own associated issue :)

As far as this PR goes, I havent tested it locally, but this makes sense to me, and if design team thinks this is a good change, then LGTM!

emoral435 (approved):

emoral435:
aha, seems like I was a bit too late to the draw for the issue xD i'll close mine!

jancborchardt:
This could use another designer/frontender review for correctness. :)
@marcoambrosini @szaimen

szaimen (approved):
I think this is fine. Also already got a +1 from a frontender :)

skjnldsv (approved):

PR #44839, opened 2024-04-15
fix(sharingDetails): Show correct share target

nfebe:
The share details page should show the target email or user/server to which
the user is creating the share.

| Before email | Before federation | After email | After federation |
| ----------------  | ---------------------- | --------------- | --------------------- |
| before-email | before-federation| after-email | after-federation |


kesselb:
Please use the right type for your commit message. Refactoring is changing code without changing behavior.

artonge (approved):

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -340,20 +340,25 @@ export default {
...
+					const [user, server] = this.share.shareWith.split('@')
+					return t('files_sharing', 'Share with {user} on remote server {server}', { user, server })
+				})()
 			case this.SHARE_TYPES.SHARE_TYPE_REMOTE_GROUP:
 				return t('files_sharing', 'Share with remote group')
 			case this.SHARE_TYPES.SHARE_TYPE_GUEST:
 				return t('files_sharing', 'Share with guest')
 			default:
-				return t('files_sharing', 'Share with')
+				return t('files_sharing', this.share.id ? 'Create share' : 'Update share')
```
susnux, line 361: This does not work as you might expect, this only works if you use exact the same strings in another place, as the l10n extraction does only handle plain strings.

```suggestion
				if (this.share.id) {
					// Share already exists
					return t('files_sharing', 'Update share')
				} else {
					return t('files_sharing', 'Create share')
				}
```

susnux (changes requested):

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -340,20 +340,32 @@ export default {
...
+			    return t('files_sharing', 'Share with email {email}', { email: this.share.shareWith })
 			case this.SHARE_TYPES.SHARE_TYPE_LINK:
 				return t('files_sharing', 'Share link')
 			case this.SHARE_TYPES.SHARE_TYPE_GROUP:
 				return t('files_sharing', 'Share with group')
 			case this.SHARE_TYPES.SHARE_TYPE_ROOM:
 				return t('files_sharing', 'Share in conversation')
 			case this.SHARE_TYPES.SHARE_TYPE_REMOTE:
-				return t('files_sharing', 'Share with remote')
+			    return (() => {
```
susnux, line 352: why is this a function?

Fenn-CS: Maintaining lexical structure of the switch case to avoid lint errors.

susnux (approved):
thank you!

PR #44848, opened 2024-04-16
fix(federation): re-add RequestSharedSecret job if necessary

blizzz:
fighting another race condition with federated server setup

## Summary

@Fenn-CS and I saw another race condition when trying to set up trusted servers. Even when running cron-jobs hand-pickedly, we could end up in state where the `requestSharedSecret` endpoint is called, claims to *We will initiate the exchange of the shared secret*, but does not have the appropriate Job in its list anymore.

The change queries the joblist to see whether this one is actually scheduled, and adds it back if it is not the case.

## Checklist

### DIFF
--- a/apps/federation/lib/Controller/OCSAuthAPIController.php
+++ b/apps/federation/lib/Controller/OCSAuthAPIController.php
@@ -7,6 +7,7 @@
  */
 namespace OCA\Federation\Controller;
 
+use OCA\Federation\BackgroundJob\RequestSharedSecret;
 use OCA\Federation\DbHandler;
 use OCA\Federation\TrustedServers;
 use OCP\AppFramework\Http;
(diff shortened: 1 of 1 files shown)


nfebe (approved):
Tested and it works, code looks good.

apps/federation/lib/Controller/OCSAuthAPIController.php
```diff
@@ -147,6 +148,21 @@ public function requestSharedSecret(string $url, string $token): DataResponse {
 				'remote server (' . $url . ') presented lower token. We will initiate the exchange of the shared secret.',
 				['app' => 'federation']
 			);
+
+			$hasJob = false;
+			foreach ($this->jobList->getJobsIterator(RequestSharedSecret::class, 1, 0) as $job) {
+				$hasJob = true;
```
nickvergessen, line 154: but this could be for another URL?

blizzz: Ah, good point. Shall compare that argument then.

nickvergessen: And then either loop over all or get chunks until you found it

blizzz: Yes. I would have looked for the hasJob, but the timestamp ruins it.

blizzz: Done. Went over all entries, I don't expect many, and would not be the first consumer to do so.

nickvergessen (approved):

nickvergessen:
Failing tests are related @blizzz

blizzz:
backlog :pensive:

PR #44858, opened 2024-04-16
Support multiple scopes in DAV search

artonge:
Instead of making the search on the targeted folders, the current approach does it on the user's root folder, with some extra conditions applied on the path.

This allows the following scope in WebDAV `SEARCH` requests:

```xml
<d:from>
	<d:scope>
		<d:href>/files/user1/a</d:href>
		<d:depth>infinity</d:depth>
	</d:scope>
	<d:scope>
		<d:href>/files/user1/b</d:href>
		<d:depth>infinity</d:depth>
	</d:scope>
</d:from>
```

This can be tested with the following PR in Photos: https://github.com/nextcloud/photos/pull/2426
With that PR, the SQL request will look like that:

```sql
SELECT
	[...]
FROM
	[...]
WHERE
	[...]
	AND (
		path COLLATE utf8mb4_general_ci LIKE "files/a/%"    -- <-- ADDED CONDITION
		OR path COLLATE utf8mb4_general_ci LIKE "files/b/%" -- <-- ADDED CONDITION
	)
	AND (
		storage = 1
		AND (
			path = "files"
			OR path LIKE "files/%"
		)
	)
ORDER BY
	[...]
LIMIT
	[...]
```

- Contributes to https://github.com/nextcloud/photos/issues/2343

### DIFF
--- a/apps/dav/lib/Files/FileSearchBackend.php
+++ b/apps/dav/lib/Files/FileSearchBackend.php
@@ -30,6 +30,7 @@ use OC\Files\Search\SearchBinaryOperator;
 use OC\Files\Search\SearchComparison;
 use OC\Files\Search\SearchOrder;
 use OC\Files\Search\SearchQuery;
+use OC\Files\Storage\Wrapper\Jail;
 use OC\Files\View;
 use OCA\DAV\Connector\Sabre\CachingTree;
 use OCA\DAV\Connector\Sabre\Directory;
@@ -39,6 +40,8 @@ use OCP\Files\Cache\ICacheEntry;
 use OCP\Files\Folder;
 use OCP\Files\IRootFolder;
 use OCP\Files\Node;
+use OCP\Files\Search\ISearchBinaryOperator;
+use OCP\Files\Search\ISearchComparison;
 use OCP\Files\Search\ISearchOperator;
 use OCP\Files\Search\ISearchOrder;
 use OCP\Files\Search\ISearchQuery;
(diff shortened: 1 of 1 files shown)


apps/dav/lib/Files/FileSearchBackend.php
```diff
@@ -157,23 +160,37 @@ public function preloadPropertyFor(array $nodes, array $requestProperties): void
...
+			/** @var Folder $folder */
+			$folder = $this->rootFolder->get($fileInfo->getPath());
+			$folderStorage = $folder->getStorage();
+			if ($folderStorage->instanceOfStorage(Jail::class)) {
+				/** @var Jail $folderStorage */
+				$internalPath = $folderStorage->getUnjailedPath($folder->getInternalPath());
+			} else {
+				$internalPath = $folder->getInternalPath();
+			}
+			$scopes[] = new SearchComparison(
```
icewind1991, line 182: It will also need to filter by the correct storage id

icewind1991 (changes requested):
If there is only a single folder, doing the search on the folder is more efficient as it can filter down the storages it searches in

apps/dav/lib/Files/FileSearchBackend.php
artonge: This is only needed for performance, not for the results, right?

Also, when making the search from the user's root folder, the filter for all sub storage is already present in the query, so the conditions would appear twice, no?

As an example, I added the result of `getQueryFilterForStorage` to the mix:

```sql
SELECT
	[...]
FROM
	[...]
WHERE
	[...]
	AND (
		(
			storage = 1
			AND path COLLATE utf8mb4_general_ci LIKE "files/Photos/%"
		)
		OR (
			storage = 2
			AND (
				path = "__groupfolders/1"
				OR path LIKE "\_\_groupfolders/1/%"
			)
			AND path COLLATE utf8mb4_general_ci LIKE "__groupfolders/1/%"
		)
	)
	AND (
		(
			storage = 1
			AND (
				(path = "files")
				OR (path LIKE "files/%")
			)
		)
		OR (
			storage = 2
			AND (
				(path = "__groupfolders/1")
				OR (path LIKE "\_\_groupfolders/1/%")
			)
		)
	)
ORDER BY
	[...]
LIMIT
	[...]
```

artonge: But maybe having this storage filter still help for performances

artonge: I also added a special handling for single folder searches.

apps/dav/lib/Files/FileSearchBackend.php
```diff
@@ -152,28 +155,74 @@ private function getPropertyDefinitionsForMetadata(): array {
...
+		return $this->rootFolder->get($fileInfo->getPath());
+	}
+
+	/**
+	 * @param Query $search
+	 * @return SearchResult[]
+	 */
+	public function search(Query $search): array {
+		switch (count($search->from)) {
+			case 0:
```
emoral435, line 181: nice switch cases! :)

emoral435 (approved):
This LGTM. In terms of optimization, I am generally unknowledgeable here, but the code logic makes sense :)

apps/dav/lib/Files/FileSearchBackend.php
```diff
@@ -152,28 +155,74 @@ private function getPropertyDefinitionsForMetadata(): array {
...
 			throw new \InvalidArgumentException('Search is only supported on directories');
 		}
 
 		$fileInfo = $node->getFileInfo();
-		$folder = $this->rootFolder->get($fileInfo->getPath());
-		/** @var Folder $folder $results */
-		$results = $folder->search($query);
+
+		/** @var Folder */
+		return $this->rootFolder->get($fileInfo->getPath());
```
come-nc, line 172: If I send a query with the path to a file as scope, what kind of error to I get back?

apps/dav/lib/Files/FileSearchBackend.php
icewind1991: Adding the storage id's there is required for correctness. Otherwise it can return results from a different storage but with the same internal path.

In your example, if the user also had an external storage that contained a `files/Photos` folder, it would return results from that folder even though it's not on of the folders searched

apps/dav/lib/Files/FileSearchBackend.php
icewind1991: Line 166 will throw a "Search is only supported on directories"

icewind1991 (approved):

apps/dav/lib/Files/FileSearchBackend.php
artonge:
```xml
  <s:exception>InvalidArgumentException</s:exception>
  <s:message>Search is only supported on directories</s:message>
</d:error>
```

PR #44861, opened 2024-04-16
chore: Fix update-psalm-baseline to use correct php version depending on branch

come-nc:
## Summary

~The workflow currently always use 8.1, it should use the same PHP version as the psalm run on the target branch.~
Adding the PHP version to psalm.xml so that the update-baseline workflow is aware of it. Backports will need to adapt the PHP version to the one used in their psalm workflow (the lowest PHP version supported usually).

## Checklist

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -8,6 +8,7 @@
     errorBaseline="build/psalm-baseline.xml"
 	findUnusedBaselineEntry="false"
 	findUnusedCode="false"
+	phpVersion="8.1"
 >
 	<plugins>
 		<plugin filename="build/psalm/AppFrameworkTainter.php" />


nickvergessen:
Should add it like in apps, e.g. https://github.com/nextcloud/spreed/blob/main/psalm.xml#L8
So it locally also always runs with the expected php version

come-nc:
> Should add it like in apps, e.g. https://github.com/nextcloud/spreed/blob/main/psalm.xml#L8 So it locally also always runs with the expected php version

Ok, trying this way then, putting 8.1 in master and will change to 8.0 in the backports.

provokateurin (approved):

nickvergessen (approved):

nickvergessen:
Backporting is not going to work, 29 and lower need 8.0 at least.
So let's save 2 CI cycles by backporting it after 29 was fixed...

PR #44868, opened 2024-04-16
fix(comments): Fix issues thrown by comments sidebar tab code

susnux:
* Resolves: # 

## Summary
When the comments tab is used instead of the merged activity+comments, then some issues are throws due to prop manipulation and duplicated names (resourceId as prop and data).

This is fixed as well as some other vue related errors in the sidebar

## Checklist


apps/comments/src/views/Comments.vue
```diff
@@ -124,7 +124,7 @@ export default {
 			loading: false,
 			done: false,
 
-			resourceId: null,
+			currentResourceId: this.resourceId,
```
artonge, line 127: Why can't it be used directly?

apps/comments/src/views/Comments.vue
```diff
@@ -113,7 +113,7 @@ export default {
 	},
 
 	directives: {
-		vElementVisibility,
+		elementVisibility,
```
artonge, line 116: Where is this used?

susnux: Line 25. Vue adds the `v-` always, so if we not rename it it needs to be called using `v-v-element-visibility`.

apps/comments/src/views/Comments.vue
susnux: Because the "open" function modifies the current resource id and you must not manipulate props (which include "resourceId" from the "CommentView" mixin.

apps/comments/src/views/Comments.vue
artonge: Ah ! it was already used :)

artonge (approved):

apps/comments/src/views/Comments.vue
susnux: Yes :sweat_smile:

Pytal (approved):
Makes sense 👍

PR #44871, opened 2024-04-17
feat: include file id in audit logs

yemkareems:
## Summary

* Few logs are changed after this implementation. Find attached is the file of logs before and after this implementation
[nc.logs.txt](https://github.com/nextcloud/server/files/15019598/nc.logs.txt)

## Checklist

### DIFF
--- a/apps/admin_audit/lib/Actions/Files.php
+++ b/apps/admin_audit/lib/Actions/Files.php
@@ -27,6 +27,18 @@ declare(strict_types=1);
  */
 namespace OCA\AdminAudit\Actions;
 
+use OCP\Files\Events\Node\BeforeNodeReadEvent;
+use OCP\Files\Events\Node\BeforeNodeWrittenEvent;
+use OCP\Files\Events\Node\NodeCopiedEvent;
+use OCP\Files\Events\Node\NodeCreatedEvent;
+use OCP\Files\Events\Node\NodeDeletedEvent;
+use OCP\Files\Events\Node\NodeRenamedEvent;
+use OCP\Files\Events\Node\NodeWrittenEvent;
+use OCP\Files\InvalidPathException;
+use OCP\Files\NotFoundException;
+use OCP\Preview\BeforePreviewFetchedEvent;
+use Psr\Log\LoggerInterface;
+
 /**
  * Class Files logs the actions to files
  *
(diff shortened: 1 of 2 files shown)


apps/admin_audit/lib/Actions/Files.php
```diff
@@ -90,11 +90,13 @@ public function create(array $params): void {
...
 	public function copy(array $params): void {
 		$this->log(
-			'File copied: "%s" to "%s"',
+			'File copied: "%s" to "%s", id from "%s" to "%s"',
 			$params,
 			[
 				'oldpath',
 				'newpath',
+				'oldid',
+				'newid'
```
artonge, line 99: It should stay the same, but I know that it is changing in some situation, so might makes sense to keep it.

apps/admin_audit/lib/Actions/Files.php
```diff
@@ -40,10 +40,10 @@ class Files extends Action {
 	 */
 	public function read(array $params): void {
 		$this->log(
-			'File accessed: "%s"',
+			'File accessed: "%s", id is: "%s"',
```
artonge, line 43: Maybe move the ID before the path so it is more readable? I am thinking of situations where the path is long, so having the ID at the beginning of the log line would help quickly find it.

```suggestion
			'File with id "%s" accessed: "%s"',
```

Same logic for other logs?

apps/admin_audit/lib/AppInfo/Application.php
```diff
@@ -206,47 +214,95 @@ function (BeforePreviewFetchedEvent $event) use ($fileActions) {
...
+
+		$eventDispatcher->addListener(
+			NodeCreatedEvent::class,
+			function (NodeCreatedEvent $event) use ($fileActions) {
+				$file = $event->getNode();
+				$fileActions->create([
+					'path' => mb_substr($file->getInternalPath(), 5),
+					'id' => $file->getId(),
+				]);
+			}
```
artonge, line 248: Please move that business logic to a new file.

artonge (commented):
Looks good overall

apps/admin_audit/lib/Actions/Files.php
yemkareems: Hi @artonge Here it is indeed changing when getting copied from one location to another. Hence leaving it as it is. However will change the text to `File id copied from: "%s" to "%s", path from "%s" to "%s"`

apps/admin_audit/lib/Actions/Files.php
yemkareems: @artonge Sure ll move the id in front and path to the end

apps/admin_audit/lib/AppInfo/Application.php
yemkareems: @artonge I have moved the logic to files class. However there are few exceptions getting thrown which i am currently debugging while file is getting created

apps/admin_audit/lib/AppInfo/Application.php
```diff
@@ -194,59 +201,81 @@ private function fileHooks(IAuditLogger $logger,
...
+			NodeRenamedEvent::class,
+			function (NodeRenamedEvent $event) use ($fileActions) {
+				$source = $event->getSource();
+				$target = $event->getTarget();
+				$fileActions->rename([
+					'oldpath' => mb_substr($source->getInternalPath(), 5),
+					'newpath' => mb_substr($target->getInternalPath(), 5),
+					'oldid' => $source->getId(),
+					'newid' => $target->getId()
 				]);
```
artonge, line 221: At that point, `$source->getInternalPath()` and `$source->getId()` will throw an error as the source is a `NonExistingFile`.

```suggestion
				$fileActions->rename([
					'oldpath' => mb_substr($source->getPath(), 5),
					'newpath' => mb_substr($target->getPath(), 5),
					'newid' => $target->getId(),
				]);
```

yemkareems: @artonge Thank you :). Refactored the rename event as suggested.

artonge (approved):

szaimen:
Sorry, don't feel confident enough to review this

apps/admin_audit/lib/AppInfo/Application.php
```diff
@@ -191,62 +197,68 @@ private function consoleHooks(IAuditLogger $logger,
...
-				$file = $event->getNode();
-				$fileActions->preview([
-					'path' => mb_substr($file->getInternalPath(), 5),
-					'width' => $event->getWidth(),
-					'height' => $event->getHeight(),
-					'crop' => $event->isCrop(),
-					'mode' => $event->getMode()
-				]);
+			function (BeforePreviewFetchedEvent $event) use ($logger) {
+				$fileActions = new Files($logger);
```
come-nc, line 203: Why did you change to create the `Files` instance inside the hook? Seems problematic performance-wise, at least.

apps/admin_audit/lib/Actions/Files.php
```diff
@@ -36,134 +47,186 @@ class Files extends Action {
...
 	 */
-	public function delete(array $params): void {
+	public function delete(NodeDeletedEvent $event): void {
+		try {
+			$params = [
+				'id' => $event->getNode()->getId(),
+				'path' => mb_substr($event->getNode()->getInternalPath(), 5),
+			];
+		} catch (InvalidPathException|NotFoundException $e) {
+			return;
```
come-nc, line 198: Should still log something, maybe even on error level, so that this is not missed.

apps/admin_audit/lib/AppInfo/Application.php
yemkareems: @come-nc No specific reason. Just had a look at appHooks/consoleHooks and did similarly. Shall i move it outside the listener

come-nc: Yes, build it once like before, seems better.

yemkareems: @come-nc fileActions restored to as it was earlier

apps/admin_audit/lib/Actions/Files.php
yemkareems: @come-nc logging the exception message as error in log file. I tried to delete a file that is already deleted from oc_filecache and also removed from data/admin/files/log.txt. I get a message in the UI saying "Delete file" action failed but i did not find it getting logged as a error message

apps/admin_audit/lib/Actions/Action.php
```diff
@@ -87,4 +87,12 @@ public function log(string $text,
...
 		);
 	}
+
+	/**
+	 * Log a error message with a log level of error
+	 * @param string $text
+	 */
+	public function error(string $text): void {
+		$this->logger->error($text, ['app' => 'admin_audit']);
+	}
```
come-nc, line 97: Please remove this method, it makes thing unnecessiraly convoluted.
Inject the loggen in the child class if possible, use `\OCP\Server::get` if not.

apps/admin_audit/lib/Actions/Files.php
```diff
@@ -36,134 +50,210 @@ class Files extends Action {
...
-	public function read(array $params): void {
+	public function read(BeforeNodeReadEvent $event): void {
+		try {
+			$params = [
+				'id' => $event->getNode()->getId(),
+				'path' => mb_substr($event->getNode()->getInternalPath(), 5),
+			];
+		} catch (InvalidPathException|NotFoundException $e) {
+			\OCP\Server::get(LoggerInterface::class)->error(
+				"Exception thrown in file read: ".$e->getMessage(), ['app' => 'admin_audit']
```
come-nc, line 63:
```suggestion
				"Exception thrown in file read: ".$e->getMessage(), ['app' => 'admin_audit', 'exception' => $e]
```
Always provide the exception to the logger when you have it, it helps debugging.

come-nc (commented):
Please fix the commit description:
Copie d'écran_20240429_132314

come-nc (approved):

PR #44884, opened 2024-04-17
feat: add request id as comment to all queries

icewind1991:
Having the request id in the sql can be useful for various database debugging

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -160,6 +160,13 @@ $CONFIG = [
 ],
 
 /**
+ * Add request id to the database query in a comment.
+ *
+ * This can be enabled to assist in mapping database logs to Nextcloud logs.
+ */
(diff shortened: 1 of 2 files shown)


icewind1991:
Silly me, expecting various SQL implementations to support the same comment format...

juliusknorr (approved):

juliushaertl:
After todays call with the db expert we could even consider this for selects (if the query cache is disabled), which would be recommended anyways these days for MySQL/MariaDB.

juliushaertl:
I think we should not have the request id there by default but could consider it as a config option if an admin is sure that they have the query cache enabled, otherwise there might be quite an overhead invalidating it with every query.

icewind1991:
Put the request id behind the `db.log_request_id` system flag

nickvergessen (approved):

PR #44897, opened 2024-04-17
fix(files): Close sidebar and update fileid when current node is deleted

### LINKED ISSUE #44637: [Bug]: UI bug when deleting a share

### Bug description

when deleting a share, the share interface remains open.
One has to reclick the delete share button so the share interface closes.
But then an error pops up, saying the share ID does not exist anymore, which makes sense, because it is the second time one tries to delete it...

### Steps to reproduce

1. on whatever folder, create a share (internal in my case)
2. then, go into the share advanced settings
3. click the delete share button
4. wait
5. wait
6. wait
7. the share menu should still be there
8. reclick the delete share button
9. the share menu disappear
10. an error pops up, see below
11. the share you've just tried to delete (or actuallly deleted 🤔) it still present in the share list
12. refresh
13. the share has disappeared as expected

image

### Expected behavior

When clicking (once) the delete share button:
- the share advanced setting should close
- the share should disappear from the list
- there should not be any error about missing share ID
- refreshing the page should be needed

### Installation method

None

### Nextcloud Server version

29

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

c.nc.c

susnux:
* Resolves: https://github.com/nextcloud/server/issues/44637

## Summary

In case we remove a node for which the sidebar is currently open, we need to do two things:
1. Close and reset the sidebar
2. Change the current fileid in the current route

This is needed because we do not change the route to the trashbin, but stay on the current route.
So we need to change the current fileid to the parent folder.

## Checklist


apps/files/src/actions/deleteAction.ts
```diff
@@ -140,14 +140,29 @@ export const action = new FileAction({
...
+				// Update the route silently (only change fileid - not triggering a reload)
+				window.OCP.Files.Router.goToRoute(
+					null,
+					{ view: view.id, fileid },
+					{ dir },
+					true,
+				)
+			}
+
 			return true
```
artonge, line 166: Maybe we could listen to the node deleted event in `Navigation.vue` and handle that there?
Feels out of topic here.

susnux: Yes and no, as I think we hack the behavior here (line 147-150) of not moving to the trash bin, so we should make sure to sanitize the route.

If we put this into the navigation (still not sure it fits there) and we would then want to move to trashbin here later, then we have two router calls conflicting.

Meaning I am fine with both, but both do not feel perfect to me :/

emoral435: If I had to choose, I would agree with the behavior being kept here. Just in case other features in the future use this, I think it makes sense if they all use the same functionality, but if there is consensus to bring this to `Navigation.vue` as well, that is fine with me!

emoral435 (approved):
I think this makes sense to me to put this here, as other features would most likely want to follow the same functionality. But if others decide to move this elsewhere, it is fine by me!

apps/files/src/views/Sidebar.vue
```diff
@@ -509,6 +512,16 @@ export default {
 			this.resetData()
 		},
 
+		/**
+		 * Handle if the current node was deleted
+		 * @param {import('@nextcloud/files').Node} node The deleted node
+		 */
+		onNodeDeleted(node) {
```
skjnldsv, line 519: that is the clean way :+1:

apps/files/src/actions/deleteAction.ts
```diff
@@ -140,14 +140,29 @@ export const action = new FileAction({
...
+
+			// If we remove the currently open file (sidebar),
+			// then we need to reset the current file id to the parent directory of that node
+			if (node.fileid && window.OCP.Files.Router.params?.fileid === node.fileid) {
+				// We can not access the paths store here because the action is loaded in a different chunk and webpack does not use the same pinia
+				const fileid = (await view.getContents(dir)).folder.fileid
+				// Update the route silently (only change fileid - not triggering a reload)
+				window.OCP.Files.Router.goToRoute(
+					null,
+					{ view: view.id, fileid },
```
skjnldsv, line 160: Do we actually need to provide the fileid?
I think nulling it and keeping the dir parameter should still work, without the need for an extra fetch.

But this is not a big deal, I'll let you be the judge of that :+1:

skjnldsv (approved):

skjnldsv (changes requested):
Wait :see_no_evil:

skjnldsv (changes requested):
Wait :see_no_evil: 

Actually, I think we're mixing two different approaches here.
I would agree with Louis, whichever service handle the routing should monitor itself.

I think having the same type of handling as you did on the sidebar would be cleaner.
The action is not really responsible for the routing, it doesn't care. It deletes the file and send the event.

The routing is managed by the FileList, if a file disappear, it should be the one adjusting the fileid param.

> If we put this into the navigation (still not sure it fits there) and we would then want to move to trashbin here later, then we have two router calls conflicting.

I fail to understand why we would have conflicting rule? If a file is removed from the list (moved or deleted), and the fileid is still in the URL, we should fallback to the folder fileid. What we're doing here on the deleteAction does also apply to the moveOrCopyAction :)

susnux:
@skjnldsv I refactored the deleted node handling of the current view.
So instead of doing it within the action, the action does not care about it now, it is moved to the view like handling updated nodes.

skjnldsv (approved):

artonge (approved):

PR #44900, opened 2024-04-17
[stable26] fix(sharing): store internal expire date when enabling

blizzz:
## Summary

When no expiration defaults are set for internal shares (e.g. to users, see screenshot),

Screenshot_20240417_195028

…but a sharer wants to set an expiration date, then visually the date is set to the next day, but it actually is not stored.

This naive patch fixes it. It is only necessary when enabling the expiration by ticking the check box. When disabling it, the change is already being propagated somewhere.

I would not be surprised if there is a smarter approach to this.

Only 26 is affected, no newer version.

## Checklist


apps/files_sharing/src/components/SharingEntry.vue
```diff
@@ -370,6 +370,9 @@ export default {
 					? this.formatDateToString(defaultExpirationDate)
 					: ''
 				console.debug('Expiration date status', enabled, this.share.expireDate)
+				if (enabled) {
+					this.queueUpdate('expireDate')
```
Fenn-CS, line 374: I know that this line does what is expected, but since I have not interacted with 26 and don't have visual picture, I want to point out that it is only useful if the share already exist. That is some prior action has led to the share being registered in the backend.

No manual testing done.

nfebe (approved):

apps/files_sharing/src/components/SharingEntry.vue
blizzz: Yes, on 26 the share already exists. That was one of the criticisms and is one of the improvements of 27.

artonge (approved):

PR #44904, opened 2024-04-17
fix(files): Also restore shares after ownership transfer for object storage

susnux:
* Resolves: 

## Summary

When a file is moved between different storages then the file id is not (always) preserved. This means the file id has to be adjusted for all shares.

So in case the file id does not exist anymore we try to find the new file id based on the target path of the transfer and the path suffix of the share.

## Checklist


susnux:
Background information:

Currently our file system does not support moving files between storages, so we need to hack this here.
Maybe we can adjust the storage code (especially `OCP\Files\Storage\Common::moveFromStorage`) so that at least the file ID in the file cache is preserved (idea: swap the rows in the database and then remove the "new" id with the old location).

Altahrim (approved):

artonge (approved):
Band-aid, but should works :+1:

PR #44945, opened 2024-04-20
fix(i18n): Changed grammar

rakekniven:
Reported at Transifex.


apps/files/js/mainfileinfodetailview.js
```diff
@@ -163,7 +163,7 @@
 					timestamp: this.model.get('mtime'),
 					date: OC.Util.relativeModifiedDate(this.model.get('mtime')),
 					hasFavoriteAction: hasFavoriteAction,
-					starAltText: isFavorite ? t('files', 'Favorited') : t('files', 'Favorite'),
+					starAltText: isFavorite ? t('files', 'Favoured') : t('files', 'Favorite'),
```
nickvergessen, line 166:
```suggestion
					starAltText: isFavorite ? t('files', 'Favoured') : t('files', 'Favour'),
```
But might need a translators hint then to make clear it's a "verb"?
https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#provide-context-hints-for-translators

apps/files/js/mainfileinfodetailview.js
```diff
@@ -159,7 +159,8 @@
 					timestamp: this.model.get('mtime'),
 					date: OC.Util.relativeModifiedDate(this.model.get('mtime')),
 					hasFavoriteAction: hasFavoriteAction,
-					starAltText: isFavorite ? t('files', 'Favorited') : t('files', 'Favorite'),
+					// TRANSLATORS "Favour" is a verb
+					starAltText: isFavorite ? t('files', 'Favoured') : t('files', 'Favour'),
```
nickvergessen, line 163: IIRC this does not work well. the translator hint is only used on the first translation, but maybe it was only in vue files

nickvergessen (approved):

Altahrim (approved):

miaulalala (changes requested):

nickvergessen:
> Conflicting files
> - apps/files/js/mainfileinfodetailview.js
> - apps/files/js/tagsplugin.js 

Both got deleted. So if we want to backport we should copy this commit to stable30 and redo the PR for master

Issue #44951, opened 2024-04-22
Make individual app boot methods more lazy


juliushaertl:
While looking at some flamegraphs I noticed that we still have plenty of apps that have logic in the boot method that could be done far more lazy. Those would trigger larger chunks of dependency injection pulling in class instances that might not be needed (yet). While some dependencies might still be pulled in later there is definitiely potential for saving some load in the app boot process:

![xdebug _ocs_v2_php_apps_notifications_api_v2_notifications 1713767602 182289 xt gz](https://github.com/nextcloud/server/assets/3404133/0c0774de-7b5e-41fc-b716-b9c630922f4d)

A few examples that could be beneficial to move:
- Mount provider registration (files_sharing, files_external, groupfolders, collectives, ...)
- Some apps like files_sharing still register events in a non-lazy way
- dav app hook manager pulls in lots of dependencies (https://github.com/nextcloud/server/blob/443b26761ddb80e8724b4eab1076f349cf62f6fd/apps/dav/lib/AppInfo/Application.php#L222)
- Legacy resource provider (deck/talk) is not done fully lazy
- Share provider registration requires to setup the ShareManager (currently only relevant for deck as talk is hard coded in server)
- Comments event handler registration requires to pull in ICommentsManager https://github.com/nextcloud/server/blob/aa5f037af71c915424c6dcfd5ad2dc82797dc0d6/apps/comments/lib/AppInfo/Application.php#L86
- Legacy ISearch provider registration (could be dropped)

PR #44953, opened 2024-04-22
perf(ILockManager): Allow registering a lock provider lazy

juliusknorr:
In order to register the lock provider without directly initializing required dependencies through DI we need to register it lazy.

As the LockManager is only handling registration/delegation of method calls it doesn't come with its own dependencies and therefore is enough to handle this. We could also move this to a dedicated registerLockProvider method but as of now there is just one use case with the files_lock app and this seems too little to justify for a explicit registration method.

Contributes to https://github.com/nextcloud/server/issues/44951

Required for https://github.com/nextcloud/files_lock/pull/297

### DIFF
--- a/lib/public/Files/Lock/ILockManager.php
+++ b/lib/public/Files/Lock/ILockManager.php
@@ -42,8 +42,16 @@ interface ILockManager extends ILockProvider {
 	/**
 	 * @throws PreConditionNotMetException if there is already a lock provider registered
 	 * @since 24.0.0
+	 * @deprecated 30.0.0 Use registerLazyLockProvider
 	 */
 	public function registerLockProvider(ILockProvider $lockProvider): void;
+
+	/**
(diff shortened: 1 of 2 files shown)


nfebe (approved):

hweihwang (approved):

lib/public/Files/Lock/ILockManager.php
```diff
@@ -42,9 +42,17 @@ interface ILockManager extends ILockProvider {
 	/**
 	 * @throws PreConditionNotMetException if there is already a lock provider registered
 	 * @since 24.0.0
+	 * @deprecated 30.0.0 Use registerLazyLockProvider
```
nickvergessen, line 45: @juliushaertl Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

Issue #44319, opened 2024-03-19
[Bug]: Broken avatar placeholders for federated users in new search


blizzz:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

In the search dialog there is no handling for not found avatars, resulting in broken images being displayed for example for federated users.

### Steps to reproduce

1. have a trusted server pairing
2. open the search dialog and there open the people one
3. check for federated users

### Expected behavior

a placeholder will be shown (as long as we do not build in support for actual federated avatars)

### Installation method

None

### Nextcloud Server version

28

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

blizzz:
It is similar to https://github.com/nextcloud/server/issues/44287

Fenn-CS:
https://github.com/nextcloud/server/pull/44972

PR #44954, opened 2024-04-22
fix(contactsMenu): Attach user cloud to each contact entry

nfebe:
The contact entry should carry cloud information, this can help to fix various issues including:

- Displaying correct avatar links for federated users

This can also lead to UI improvements in the frontend where, it's federated contacts are shown with a marker or indicator on the UI.

* Resolves: https://github.com/nextcloud/server/issues/44319

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Contacts/ContactsMenu/ContactsStore.php
+++ b/lib/private/Contacts/ContactsMenu/ContactsStore.php
@@ -31,6 +31,7 @@
 
 namespace OC\Contacts\ContactsMenu;
 
+use OC\Federation\CloudId;
 use OC\KnownUser\KnownUserService;
 use OC\Profile\ProfileManager;
 use OCA\UserStatus\Db\UserStatus;
@@ -44,6 +45,7 @@ use OCP\IURLGenerator;
 use OCP\IUser;
 use OCP\IUserManager;
 use OCP\L10N\IFactory as IL10NFactory;
+use Psr\Log\LoggerInterface;
 use function array_column;
 use function array_fill_keys;
 use function array_filter;
@@ -62,6 +64,7 @@ class ContactsStore implements IContactsStore {
 		private IGroupManager $groupManager,
 		private KnownUserService $knownUserService,
 		private IL10NFactory $l10nFactory,
+		private LoggerInterface $logger,
 	) {
 	}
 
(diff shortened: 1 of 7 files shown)


nfebe (commented):
To solve this issue we could simple use the `isUser` property of the contact along side `isNouser` boolean of [`NcAvatar`](https://nextcloud-vue-components.netlify.app/#/Components/NcAvatar) to show the text generated icons.

However, here:

- I instead try to link to the actual avatars on the remote server, since the links public
- I am adding the cloud data to be passed to the frontend

Problems:

- This code is able to load the avatars from a federated server but would like to "Content Security Policy" errors if the remote server is unsecured.

-----

There are other places in the code base such as [here](https://github.com/nextcloud/server/blob/097f04c8b78851eb556d9dd1de2786be5e5f9aa5/apps/comments/lib/Search/CommentsSearchProvider.php#L75) where the current approach is used but we could swap if this attempt to directly load the avatars from the federated server is successful/useful.


@nickvergessen since you are familiar with section of the code, does this approach make sense?

lib/private/Contacts/ContactsMenu/ContactsStore.php
```diff
@@ -351,19 +354,35 @@ public function findOne(IUser $user, int $shareType, string $shareWith): ?IEntry
...
-
 		if (!empty($contact['UID'])) {
 			$uid = $contact['UID'];
 			$entry->setId($uid);
 			$entry->setProperty('isUser', false);
+			$username = '';
+			$remoteServer = '';
+
+			if (isset($contact['CLOUD']) && is_array($contact['CLOUD']) && isset($contact['CLOUD'][0])) {
+				preg_match('/^(.*?)@(https?:\/\/.*?)$/', $contact['CLOUD'][0], $matches);
```
nickvergessen, line 365: Please use the `ICloudManager::resolve()` method instead?

lib/private/Contacts/ContactsMenu/Entry.php
```diff
@@ -163,8 +162,25 @@ public function getProperty(string $key): mixed {
...
+		$this->cloud = $cloudId;
+	}
+
+	public function getCloud(): CloudId {
+		return $this->cloud;
+	}
+
 	/**
-	 * @return array{id: int|string|null, fullName: string, avatar: string|null, topAction: mixed, actions: array, lastMessage: '', emailAddresses: string[], profileTitle: string|null, profileUrl: string|null, status: string|null, statusMessage: null|string, statusMessageTimestamp: null|int, statusIcon: null|string, isUser: bool, uid: mixed}
+	 * @return array{id: int|string|null, fullName: string, avatar: string|null, topAction: mixed, actions: array, lastMessage: '', emailAddresses: string[], profileTitle: string|null, profileUrl: string|null, status: string|null, statusMessage: null|string, statusMessageTimestamp: null|int, statusIcon: null|string, isUser: bool, uid: mixed, cloud: mixed}
```
nickvergessen, line 183: cloud should be called cloudId to be more clear and the type should be the array you define in the other file

lib/private/Federation/CloudId.php
```diff
@@ -88,4 +88,16 @@ public function getUser(): string {
 	public function getRemote(): string {
 		return $this->remote;
 	}
+
+	/**
+	 * @return array{id: string, user: string, remote: string, displayName: string|null}
+	 */
+	public function jsonSerialize(): array {
```
nickvergessen, line 95: Should implement the interface JsonSerializable?

lib/private/Contacts/ContactsMenu/ContactsStore.php
```diff
@@ -351,19 +354,35 @@ public function findOne(IUser $user, int $shareType, string $shareWith): ?IEntry
...
+				$this->logger->warning('Invalid remote server data');
+			}
 			// overloaded usage so leaving as-is for now
 			if (isset($contact['isLocalSystemBook'])) {
 				$avatar = $this->urlGenerator->linkToRouteAbsolute('core.avatar.getAvatar', ['userId' => $uid, 'size' => 64]);
 				$entry->setProperty('isUser', true);
-			} elseif (!empty($contact['FN'])) {
-				$avatar = $this->urlGenerator->linkToRouteAbsolute('core.GuestAvatar.getAvatar', ['guestName' => str_replace('/', ' ', $contact['FN']), 'size' => 64]);
+			} elseif ($username != '') {
+				$avatar = $this->urlGenerator->linkToRemoteRouteAbsolute($remoteServer, 'core.avatar.getAvatar', ['userId' => str_replace('/', ' ', $username), 'size' => 64]);
```
nickvergessen, line 383: This assumes every cloud ID is a nextcloud? 🫣

lib/private/URLGenerator.php
```diff
@@ -115,6 +115,19 @@ public function linkToRouteAbsolute(string $routeName, array $arguments = []): s
 		return $this->getAbsoluteURL($this->linkToRoute($routeName, $arguments));
 	}
 
+
+	public function linkToRemoteRouteAbsolute(string $remote, $routeName, array $arguments = []): string {
```
nickvergessen, line 119: This should not be here or on the public interface

lib/private/Federation/CloudId.php
nickvergessen: That being said, should be on the public interface?

nickvergessen (changes requested):

nickvergessen:
> does this approach make sense

From my POV it's not okay...

Fenn-CS:
Based on conversation with @nickvergessen, this needs more effort especially as the content security policy can't changed to allow all trusted servers.

So this would remain in draft in favor of : https://github.com/nextcloud/server/pull/44972

Eventually, I could copy the proxy strategy from https://github.com/nextcloud/spreed/blob/main/lib/Controller/AvatarController.php#L244 where the server downloads the image and sends in on the instance URL.

skjnldsv:
Closing as stale and EOL branch.
If still required, please resume your work on it by rebasing and requesting for reviews again :pray:

PR #44972, opened 2024-04-22
[stable29] fix(unifiedSearch): Prevent broken avatars for federated users in peo…

nfebe:
| Before | After |
| --------- | ------ |
|Screenshot from 2024-04-22 16-08-33 | Screenshot from 2024-04-22 20-40-10 |

Resolves : https://github.com/nextcloud/server/issues/44319


core/src/components/UnifiedSearch/SearchableList.vue
```diff
@@ -46,7 +46,8 @@
 						:wide="true"
 						@click="itemSelected(element)">
 						<template #icon>
-							<NcAvatar :user="element.user" :show-user-status="false" :hide-favorite="false" />
+							<NcAvatar v-if="element.isUser" :user="element.user" :show-user-status="false" :hide-favorite="false" />
```
Pytal, line 49:
```suggestion
							<NcAvatar v-if="element.isUser" :user="element.user" :show-user-status="false" />
```
Don't think this prop exists 🙈

core/src/components/UnifiedSearch/SearchableList.vue
```diff
@@ -46,7 +46,8 @@
...
+							<NcAvatar v-else :is-no-user="true" :display-name="element.displayName" :show-user-status="false" :hide-favorite="false" />
```
Pytal, line 50:
```suggestion
							<NcAvatar v-else :is-no-user="true" :display-name="element.displayName" :show-user-status="false" />
```
Same here

Pytal (approved):

susnux (approved):

PR #44996, opened 2024-04-23
fix(files): Use OCP\Util::getForbiddenFileNameChars instead of directaccess to system config

susnux:
## Summary

We (might) extend it in the future so we should only use one common place (OCP\Util) for getting forbidden characters.

## Checklist


apps/files/tests/Controller/ViewControllerTest.php
artonge, line 1: Why are those changes needed?

artonge (approved):

Altahrim (approved):

apps/files/tests/Controller/ViewControllerTest.php
susnux: I thought the test would (like usual) also check initial state and thus we need to ensure a defined value is set. But the state is not checked so no need here.
Removed that change.

come-nc (approved):

PR #44997, opened 2024-04-24
fix(lib):When performing a fresh installation, the version files under /core/Migrations/ cannot be loaded, resulting in the inability to install properly.

yrlchn:
fix(lib):When performing a fresh installation, the version files under /core/Migrations/ cannot be loaded, resulting in the inability to install properly.


kesselb:
Do you know which migration didn't run? Just wondering because we install nextcloud many times a day on our ci and I couldn't spot an issue.

PR #45013, opened 2024-04-24
fix(db): Prevent two connections for single node databases

### LINKED ISSUE #45003: [Bug]: Two database connections and missing transaction isolation level for read operations

### Bug description

I am capturing all queries of my MariaDB installation and noticed a strange query pattern for simple Nextcloud requests:

```sql
-- nc_admin@localhost on nextclouddev using Socket
SELECT `appid`, `configkey`, `configvalue`, `type` FROM `oc_appconfig` WHERE `lazy` = 0;
SELECT `uid`, `displayname`, `password` FROM `oc_users` WHERE `uid_lower` = 'admin'
SELECT `gu`.`gid`, `g`.`displayname` FROM `oc_group_user` `gu` LEFT JOIN `oc_groups` `g` ON `gu`.`gid` = `g`.`gid` WHERE `uid` = 'admin'
SELECT `provider_id`, `enabled` FROM `oc_twofactor_providers` WHERE `uid` = 'admin'
SELECT `appid`, `configkey`, `configvalue` FROM `oc_preferences` WHERE `userid` = 'admin'
-- nc_admin@localhost on nextclouddev using Socket
SET SESSION TRANSACTION ISOLATION LEVEL READ COMMITTED
SET SESSION AUTOCOMMIT=1
START TRANSACTION
SELECT * FROM `oc_authtoken` WHERE (`uid` = 'admin') AND (`version` = 2) LIMIT 1000
COMMIT
...
```

### Steps to reproduce

1. Set up Nextcloud with a single node database
2. Send requests

### Expected behavior

Single connection and read queries use READ COMMITTED too, not the global database default.

### Installation method

None

### Nextcloud Server version

29

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

Related to https://github.com/nextcloud/server/pull/41998.

ChristophWurst:
* Resolves https://github.com/nextcloud/server/issues/45003

### Two connections to one database node

Since https://github.com/nextcloud/server/pull/41998 our connection always extends Doctrine's PrimaryReadReplicaConnection. This baseclass requires a primary and at least one replica and always does the read-write split.
For simpler Nextcloud installations we only have one database node and no replica. We do not want two database connections.
This overwrites the ``\Doctrine\DBAL\Connections\PrimaryReadReplicaConnection::performConnect`` method so we always connect to the primary if there is no replica. If there is one, we do the Doctrine read-write split.

### ~~Missing transaction isolation level for replica~~

~~Nextcloud uses READ COMMITTED as transaction isolation level. Setting it for the session was missing for the replica connections.~~ Reverted because it implicitly causes a connection to the primary just to set the session variable.

## How to test - Single node

1. Set up nextcloud with a single database node (mysql, postgres, oracle)
2. Enable query logging
    I. `'log_query' => true',`
    II. `'query_log_file' => '/var/www/nextcloud/data/query.log',`
3. Send some requests

master: you see replica and primary connections prefixes for the logged queries. That indicates that Nextcloud opens two connections per request.
here: you only see primary connections. That indicates that Nextcloud only opens one connection per request.

## How to test - Cluster

 1. Set up nextcloud with a single database node (mysql, postgres, oracle)
 2. Add a fake replica by using a different hostname
3. Enable query logging
    I. `'log_query' => true',`
    II. `'query_log_file' => '/var/www/nextcloud/data/query.log',`
4. Send some requests

master and here: Nextcloud opens two connections.

## Checklist

### DIFF
--- a/tests/lib/DB/ConnectionTest.php
+++ b/tests/lib/DB/ConnectionTest.php
@@ -1,0 +1,101 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace Test\DB;
+
+use Doctrine\DBAL\Configuration;
+use Doctrine\DBAL\Driver;
+use Doctrine\DBAL\Driver\Connection as DriverConnection;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use OC\DB\Adapter;
+use OC\DB\Connection;
+use Test\TestCase;
+
+/**
+ * @group DB
+ */
+class ConnectionTest extends TestCase {
+
+	public function testSingleNodeConnectsToPrimaryOnly(): void {
+		$connectionParams = [
+			'user' => 'test',
+			'password' => 'topsecret',
+			'host' => 'test',
+		];
+		$adapter = $this->createMock(Adapter::class);
+		$driver = $this->createMock(Driver::class);
+		$configuration = $this->createMock(Configuration::class);
+		$connection = $this->getMockBuilder(Connection::class)
+			->onlyMethods(['connectTo'])
+			->setConstructorArgs([
+				[
+					'adapter' => $adapter,
+					'platform' => new MySQLPlatform(),
+					'tablePrefix' => 'nctest',
+					'primary' => $connectionParams,
+					'replica' => [
+						$connectionParams,
+					],
+				],
+				$driver,
+				$configuration,
+			])
+			->getMock();
+		$driverConnection = $this->createMock(DriverConnection::class);
+		$connection->expects(self::once())
+			->method('connectTo')
+			->with('primary')
+			->willReturn($driverConnection);
+
+		$connection->ensureConnectedToReplica();
+		$connection->ensureConnectedToPrimary();
+		$connection->ensureConnectedToReplica();
+	}
+
+	public function testClusterConnectsToPrimaryAndReplica(): void {
+		$connectionParamsPrimary = [
+			'user' => 'test',
+			'password' => 'topsecret',
+			'host' => 'testprimary',
+		];
+		$connectionParamsReplica = [
+			'user' => 'test',
+			'password' => 'topsecret',
+			'host' => 'testreplica',
+		];
+		$adapter = $this->createMock(Adapter::class);
+		$driver = $this->createMock(Driver::class);
+		$configuration = $this->createMock(Configuration::class);
+		$connection = $this->getMockBuilder(Connection::class)
+			->onlyMethods(['connectTo'])
+			->setConstructorArgs([
+				[
+					'adapter' => $adapter,
+					'platform' => new MySQLPlatform(),
+					'tablePrefix' => 'nctest',
+					'primary' => $connectionParamsPrimary,
+					'replica' => [
+						$connectionParamsReplica,
+					],
+				],
+				$driver,
+				$configuration,
+			])
+			->getMock();
+		$driverConnection = $this->createMock(DriverConnection::class);
+		$connection->expects(self::exactly(2))
+			->method('connectTo')
+			->willReturn($driverConnection);
+
+		$connection->ensureConnectedToReplica();
+		$connection->ensureConnectedToPrimary();
+		$connection->ensureConnectedToReplica();
+	}
+
+}
--- a/lib/private/DB/Connection.php
+++ b/lib/private/DB/Connection.php
@@ -31,6 +31,7 @@ use OCP\Profiler\IProfiler;
 use OCP\Server;
 use Psr\Clock\ClockInterface;
 use Psr\Log\LoggerInterface;
+use function count;
 use function in_array;
 
 class Connection extends PrimaryReadReplicaConnection {
@@ -73,7 +74,7 @@ class Connection extends PrimaryReadReplicaConnection {
 	 * @throws \Exception
 	 */
 	public function __construct(
-		array $params,
+		private array $params,
 		Driver $driver,
 		?Configuration $config = null,
 		?EventManager $eventManager = null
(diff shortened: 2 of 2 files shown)


ChristophWurst:
The downside of this approach is that the query logging will speak the truth for single database node setups and prefix queries with "primary" even if they would go to a replica on clustered setups.

ChristophWurst:
> The downside of this approach is that the query logging will speak the truth for single database node setups and prefix queries with "primary" even if they would go to a replica on clustered setups.

This feels tricky to achieve without breaking too much. The current approach works for single node and clustered databases. A unit test was added to ensure this doesn't happen again.

lib/private/DB/SetTransactionIsolationLevel.php
```diff
@@ -39,11 +39,9 @@ class SetTransactionIsolationLevel implements EventSubscriber {
...
 		$connection = $args->getConnection();
-		if ($connection instanceof PrimaryReadReplicaConnection && $connection->isConnectedToPrimary()) {
-			$connection->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
-			if ($connection->getDatabasePlatform() instanceof MySQLPlatform) {
-				$connection->executeStatement('SET SESSION AUTOCOMMIT=1');
-			}
+		$connection->setTransactionIsolation(TransactionIsolationLevel::READ_COMMITTED);
+		if ($connection->getDatabasePlatform() instanceof MySQLPlatform) {
+			$connection->executeStatement('SET SESSION AUTOCOMMIT=1');
 		}
```
ChristophWurst, line 45: this is giving me headache. the doctrine connection always switches to the primary when executing *statements*. if we want to set the isolation level of the replica connect, doctrine switchs to the primary. this totally defeats the idea of replicas.

I think this is the reason why https://github.com/nextcloud/server/pull/41998 came with the logic to only set the transaction isolation level for primary connections.

Altahrim: I am wondering if we should really take care of that.
Autocommit status and isolation level could be setup checks only no? If you know how to setup a replica, you should be able to set that too. Bonus, it saves two queries for MySQL ^^

ChristophWurst: I agree. We have already discussed the option of not setting the level for the session. If only every setup was configured correctly … :wink:

Altahrim: That's why the new setup check could be enough, at least for the replica thing.

I think the default is REPEATABLE_READ. For a replica, it should be good enough no?

kesselb: > I think this is the reason why https://github.com/nextcloud/server/pull/41998 came with the logic to only set the transaction isolation level for primary connections.

Yep, otherwise it would reconnect to primary right away. 

> I am wondering if we should really take care of that. Autocommit status and isolation level could be setup checks only no? If you know how to setup a replica, you should be able to set that too.

:+1: 

I guess we don't have many options here.

We need to update our documentation "warning, if you configure dbreplica the automatic configuration for the transaction isolation level is disabled" and add a setup check. For non-replica setups we can still set the transaction level. 

I took a brief look at "SET SESSION AUTOCOMMIT=1" when testing https://github.com/nextcloud/server/pull/41998. I think autocommit=1 is the default for a while now. Maybe it's not even necessary anymore to set it.

- https://stackoverflow.com/questions/2950676/difference-between-set-autocommit-1-and-start-transaction-in-mysql-have-i-misse
- https://mariadb.com/kb/en/server-system-variables/#autocommit
- https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_autocommit 
- https://dev.mysql.com/doc/refman/8.0/en/glossary.html#glos_autocommit
- https://github.com/nextcloud/server/issues/29911

The alternative solution would be to revert everything and say we don't support sql replication, use something like ProxySQL? Do we know what ProxySQL does with the set transaction level statement, is that also executed against the primary node?

ChristophWurst: > The alternative solution would be to revert everything and say we don't support sql replication, use something like ProxySQL?

AFAIR it's not the same as our replication. Read/write splits inside the application allows for finer control, and also prevent read after write scenarios. That's why we did the split in Nextcloud.

ChristophWurst:
I have reverted the second part of setting the transaction isolation level for read replicas.

This PR now only fixes having two connections for one request connecting to a single node database.

kesselb:
Let's try to rewrite this code sometime and make our connection class a decorator for the doctrine classes rather than extending it.

kesselb (approved):

ChristophWurst:
Squashed

tests/lib/DB/ConnectionTest.php
```diff
@@ -0,0 +1,98 @@
...
+class ConnectionTest extends TestCase {
+
+	public function testSingleNodeConnectsToPrimaryOnly(): void {
+		$connectionParams = [
+			'user' => 'test',
+			'password' => 'topsecret',
+			'host' => 'test',
+		];
+		$adapter = $this->createMock(Adapter::class);
+		$driver = $this->createMock(Driver::class);
```
artonge, line 31: You apparently need to mock `detectDatabasePlatform`

https://github.com/nextcloud/server/actions/runs/9265386150/job/25487375754?pr=45013

ChristophWurst: Fixed by passing an actual platform. Thanks!

Altahrim (approved):

szaimen:
Hi, is this fixing the problem https://github.com/nextcloud/server/issues/45257?

ChristophWurst:
I don't think so

PR #45026, opened 2024-04-25
Avoid updating the same oc_authtoken row twice

juliusknorr:
- **docs: Update token_auth_activity_update default value to match implementation**
- **fix: Always set last activity if we update the row of an authtoken anyways**

## Summary

Checking out query frequency on an instance I noticed that the following two queries roughly are happening quite frequently and checking the code I noticed that it might happen that we issue two update queries for the same row in a request.

    UPDATE `oc_authtoken` SET `last_check` = ? WHERE `id` = ?
    UPDATE `oc_authtoken` SET `last_activity` = ? WHERE ( `id` = ? ) AND ( `last_activity` < ? )

If we update the last_checked value of a row anyways I think we could directly also set the last activity value to avoid double updates of that row.

This is triggered from https://github.com/nextcloud/server/blob/ec5133b739eabc76271789504b4dbb91a534f552/lib/private/User/Session.php#L818-L829 where checkTokenCredentials might update the last_check value and updateTokenActivity could then perform a second update just for the activity.

I could verify this by manually `update oc_authtoken set last_check = 0, last_activity = 0;` to simulate older timestampts and stepping through the next request with a debugger

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -370,7 +370,7 @@ $CONFIG = [
  * Tokens are still checked every 5 minutes for validity
  * max value: 300
  *
- * Defaults to ``300``
+ * Defaults to ``60``
  */
 'token_auth_activity_update' => 60,
 
--- a/lib/private/Authentication/Token/PublicKeyTokenProvider.php
+++ b/lib/private/Authentication/Token/PublicKeyTokenProvider.php
@@ -308,6 +308,8 @@ class PublicKeyTokenProvider implements IProvider {
 		if (!($token instanceof PublicKeyToken)) {
 			throw new InvalidTokenException("Invalid token type");
 		}
+		$now = $this->time->getTime();
+		$token->setLastActivity($now);
 		$this->mapper->update($token);
 		$this->cacheToken($token);
 	}


Altahrim (approved):

joshtrichards (commented):
Do you foresee any downsides here since we'd be bypassing the `updateActivity()` tolerance optimizations?

https://github.com/nextcloud/server/blob/acf8ea176143fa7f6e2ca64eba6cfbfd49b49091/lib/private/Authentication/Token/PublicKeyTokenMapper.php#L203-L231

juliushaertl:
I don't see any additional risk there. The row is updated anways when this is being called and I also didn't see any hints on unnecessary further updates on query statistics of the high traffic instance where this finding originated from.

joshtrichards (approved):

juliushaertl:
Waiting with the merge since @ChristophWurst wanted to have another close look

ChristophWurst (approved):

ChristophWurst:
I've realized a logical issue where we now update the last_activity for tokens that are not active. E.g. ``\OC\Authentication\Token\PublicKeyTokenProvider::rotate``, ``\OC\Authentication\Token\PublicKeyTokenProvider::setPassword`` and ``\OCA\Settings\Controller\AuthSettingsController::update`` update tokens of other clients/sessions.

Especially with the rotation and the password update we will see tokens/clients as active even if they have not connected.

ChristophWurst:
> I've realized a logical issue where we now update the last_activity for tokens that are not active

Addressed with https://github.com/nextcloud/server/pull/45411

PR #45049, opened 2024-04-26
fix: dashboard performance problems

### LINKED ISSUE #44122: scrolling in dashboard is laggy
just tested on nc29.0.0b1
Scrolling in dashboard app is laggy in chromium based browser:

https://github.com/nextcloud/server/assets/42591237/cc20a0f0-4a9b-46b9-ab52-9da7a0445c20

szaimen:
- Close https://github.com/nextcloud/server/issues/44122
- linked to https://github.com/nextcloud-libraries/nextcloud-vue/pull/5607

### DIFF
--- a/apps/dashboard/src/DashboardApp.vue
+++ b/apps/dashboard/src/DashboardApp.vue
@@ -498,7 +498,8 @@ export default {
 	align-self: stretch;
 	background-color: var(--color-main-background-blur);
 	-webkit-backdrop-filter: var(--filter-background-blur);
-	backdrop-filter: var(--filter-background-blur);
+	// Commented because it leads to huge performance problems in Chromium based browsers on Windows
+	// backdrop-filter: var(--filter-background-blur);
 	border-radius: var(--border-radius-rounded);
 
 	#body-user.theme--highcontrast & {


susnux:
We use this also in other places, like login form. Same filter is also applied to the app navigation. 
So this should be discussed generally for our design, cc @nextcloud/designers .

Also note this only happens if you run a Blink engine browser **without** hardware acceleration. 

If we want to keep semi transparent backgrounds then we maybe should add this as a CSS variable an provide a variant of the default theme with transparency disabled.

apps/dashboard/src/DashboardApp.vue
```diff
@@ -498,7 +498,8 @@ export default {
 	align-self: stretch;
 	background-color: var(--color-main-background-blur);
 	-webkit-backdrop-filter: var(--filter-background-blur);
```
susnux, line 500: If we Termine remove it then also for Safari

szaimen: Are there also performance problems with this in safari?

susnux: No but you removed it for all browsers but left it for `-webkit` prefix (so safari). Meaning you also removed this for Android (there it is hardware accelerated on chrome / webview), for Firefox, for Chrome and Edge.
So currently the design would be different for all but Safari.

jancborchardt (commented):
> Also note this only happens if you run a Blink engine browser without hardware acceleration.
> If we want to keep semi transparent backgrounds then we maybe should add this as a CSS variable an provide a variant of the default theme with transparency disabled.

@susnux sounds reasonable to me! I would say we shouldn’t disable the frosted glass outright.

marcoambrosini (approved):

szaimen:
> > Also note this only happens if you run a Blink engine browser without hardware acceleration.
> > If we want to keep semi transparent backgrounds then we maybe should add this as a CSS variable an provide a variant of the default theme with transparency disabled.
> 
> @susnux sounds reasonable to me! I would say we shouldn’t disable the frosted glass outright.

So adding another theme? Sounds overengineered to me honestly... Is there maybe a way to add semi transparent backgrounds without crippling performance?

apps/dashboard/src/DashboardApp.vue
szaimen: Does not seem to be a performance problem on Safari. Just verified with Jos

PR #45057, opened 2024-04-26
feat(exception-template): allow to link to specific documentation for how to retreive server log

szaimen:
The idea is to allow to link to a certain documentation whenever the exception template comes up which is useful e.g. for AIO so that people can go to the correct directly without having to open a new thread for it.

image

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -2242,6 +2242,11 @@ $CONFIG = [
 'upgrade.cli-upgrade-link' => '',
 
 /**
+ * Allows to modify the exception server logs documentation link in order to link to a different documentation
+ */
+'documentation_url.server_logs' => '',
+
+/**
  * Set this Nextcloud instance to debugging mode
  *
  * Only enable this for local development and not in production environments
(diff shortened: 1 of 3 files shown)


config/config.sample.php
```diff
@@ -2235,6 +2235,11 @@
  */
 'upgrade.cli-upgrade-link' => '',
 
+/**
+ * Allows to modify the exception server logs documentation link in order to link to a different documentation
+ */
+'server_logs_documentation ' => '',
```
szaimen, line 2241: Not sure about this name. Should it be named differently?

juliushaertl: I could imagine this being useful also for other references, where we currently only have the old theme approach https://github.com/nextcloud/server/blob/9a89e9086f2f52cba45ba258757eb8e2ece7f667/lib/private/legacy/OC_Defaults.php#L175

Maybewe can namespace it somehow like `'documentation_url.server_logs`

szaimen: good idea, thanks!

juliusknorr (approved):

susnux (approved):
:crocodile:

PR #45058, opened 2024-04-26
perf(db): Sort data for IN before chunking

ChristophWurst:
## Summary

Nextcloud sends chunks of `DELETE FROM oc_filecache WHERE parentid IN (....long list...)` and *sometimes* MariaDB does not seem to pick up the index. Apparently this may happen when the query analyzer thinks it's cheaper to scan the table than to traverse the index. If we put the data of the chunks closer together, we can hopefully convince the db that the index is the faster choice.

@juliushaertl @nickvergessen as it was suggested to us.

I don't know how to trigger this code with lots of data so this is more or less untested.

## Checklist

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -620,6 +620,9 @@ class Cache implements ICache {
 		$query->delete('filecache')
 			->whereParentInParameter('parentIds');
 
+		// Sorting before chunking allows the db to find the entries close to each
+		// other in the index
+		sort($parentIds, SORT_NUMERIC);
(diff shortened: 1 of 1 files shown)


Altahrim (approved):
I didn't knew the trick to sort items. The other one is to use smaller chunks.

come-nc (approved):
@Altahrim Is that related to the big folder deletion&restore we looked into together? Would it help with the deadlock there?

nickvergessen (approved):

nickvergessen:
> Would it help with the deadlock there?

Could, as that is also the direction where we are coming from

PR #45064, opened 2024-04-26
Updating node.yml workflow from template

nextcloud-command:
Automated update of the node.yml workflow from https://github.com/nextcloud/.github

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -2,11 +2,13 @@
 #
 # https://github.com/nextcloud/.github
 # https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization
+#
+# SPDX-FileCopyrightText: 2021-2024 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
 
 name: Node
 
-on:
-  pull_request:
+on: pull_request
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


susnux (approved):

.github/workflows/node.yml
```diff
@@ -48,17 +50,17 @@ jobs:
...
+        uses: skjnldsv/read-package-engines-version-actions@06d6baf7d8f41934ab630e97d9e6c0bc9c9ac5e4 # v3
         id: versions
         with:
           fallbackNode: '^20'
-          fallbackNpm: '^9'
+          fallbackNpm: '^10'
 
       - name: Set up node ${{ steps.versions.outputs.nodeVersion }}
-        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
+        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v3
```
skjnldsv, line 63:
```suggestion
        uses: actions/setup-node@60edb5dd545a775178f52524783378180af0d1f8 # v4.0.2
```

susnux: I am not sure why dependabot sometimes updates that comment and sometimes ignores it.
My guess is if it matches it updates it, if it does not it thinks it is a comment and leaves it like it was (e.g. when someone manually updated it without touching the comment).

.github/workflows/node.yml
```diff
@@ -30,8 +32,8 @@ jobs:
           filters: |
             src:
               - '.github/workflows/**'
-              - '**/src/**'
-              - '**/appinfo/info.xml'
+              - 'src/**'
+              - 'appinfo/info.xml'
```
come-nc, line 36: This change should be cancelled I suppose for server?

susnux: done

come-nc (approved):

PR #45070, opened 2024-04-26
Fix Issue #44364 get rid of the additional margin 

### LINKED ISSUE #44364: [Bug]: Remove extra css margin for URL in task title

### Bug description

Originally reported in https://github.com/nextcloud/tasks/issues/2500

This issue is in the tasks app but the code causing it lives here.

### Steps to reproduce

1. Add a new task with a URL as the title https://wikipedia.org

### Expected behavior

The task title is in line with non-URL titles.

### Actual behaviour

It is off by 3px

image

### Towards solutions

This appears to be caused by the `a.external` CSS which adds a margin.

```css
a.external {
	margin: 0 3px;
	text-decoration: underline;
}
``` 

Looks like it actually comes from [here](https://github.com/nextcloud/server/blob/1d100bdcd989b1343d644f554a8990318abeed80/core/css/styles.scss#L81-L84) in `nextcloud/server/.../styles.scss`.

The external class is added by calling linkify on this line:
https://github.com/nextcloud/tasks/blob/8a1a4ab193e02a0c3e9c22817e39e46c99b6e180/src/components/TaskBody.vue#L53

The linkify function always adds the external class:
https://github.com/nextcloud-libraries/nextcloud-vue/blob/87322fabb7f12ed9a64862903a204969ee45c632/src/utils/Linkify.js#L30-L39

So the way I see it there are a few options:

* Add our a CSS to [taskbody.vue](https://github.com/nextcloud/tasks/blob/8a1a4ab193e02a0c3e9c22817e39e46c99b6e180/src/components/TaskBody.vue) that looks something like: `.task-body__info .title a { margin: unset; }`
  * Easy to do but maybe not the best for maintainability
* change styles.scss in nextcloud/server 
  * Probably not a good idea since it may be used many places but at the same time. But why would all external links in all of nextcloud be indented by 3px? Maybe this should be fixed?
  * The Margin was added way back in https://github.com/nextcloud/server/pull/7680 for a specific use case
* Don't use linkify
  * Honestly not totally sure of the purpose/history of linkify. It's probably there for a reason.
* Modify linkify to have an option to not add the external class.
  * Seems not ideal

### Installation method

None

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

RayanBekri:
* Resolves: #44364 

- Co-authored-by: JEEEEEEEEEEEEEEEEEEEEEED <jed.boulahya@medtech.tn> @jadjoud 

## Summary
fix the margin to make the link aligned with the other tasks by setting the margins for the links to 0px rather than 0 3px.

## before 
Before

## after
after

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -108,3 +108,7 @@
 
 AddDefaultCharset utf-8
 Options -Indexes
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
+
+ErrorDocument 403 /index.php/error/403
+ErrorDocument 404 /index.php/error/404
--- a/core/css/styles.scss
+++ b/core/css/styles.scss
@@ -79,7 +79,7 @@ a {
 }
 
 a.external {
-	margin: 0 3px;
+	margin: 0px;
 	text-decoration: underline;
 }
 


jadjoud:
@susnux

.htaccess
```diff
@@ -108,3 +108,7 @@
 
 AddDefaultCharset utf-8
 Options -Indexes
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
+
+ErrorDocument 403 /index.php/error/403
```
susnux, line 113: This should not be part of the pr

susnux:
If this is just about tasks then you should oberride the styles on the tasks app.
This will change all external links. So this might introduce unwanted sideffects. 

So cc the designers for styles review.

.htaccess
jadjoud: noted.

RayanBekri:
> If this is just about tasks then you should oberride the styles on the tasks app.
> 
> This will change all external links. So this might introduce unwanted sideffects. 
> 
> 
> 
> So cc the designers for styles review.


Understood just could please provide us with designer name
Thanks in advance.

susnux:
> Understood just could you please provide us with designer name

Already pinged @nextcloud/designers :)

jancborchardt (changes requested):
This is part of the external link styles so the icon to the right of it displays with a bit of whitespace.

As @susnux said, I would recommend doing this change directly in the Tasks app. :)

Thank you for contributing!

RayanBekri:
> This is part of the external link styles so the icon to the right of it displays with a bit of whitespace.
> 
> As @susnux said, I would recommend doing this change directly in the Tasks app. :)
> 
> Thank you for contributing!

Thanks for enlightening us exactly about the change that we did. 
We will try to continue working on other issues and we will see if we can do much with the task app.

Again thanks for your time.

Issue #44936, opened 2024-04-19
[Bug]: Search in the"disabled users" section does not work


bigcat88:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

In the Nextcloud 28.0.4 version, the search in the "disabled users" section does not work.
If a search string is entered in the search mask, nothing can be found and the list of all deactivated users is still displayed.

### Steps to reproduce

1. go to nc.url/settings/users/disabled
2. click on Unified Search
3. type anything and see that the list of users is not filtered

### Expected behavior

The same as for "Active Users", ability to use search like a filter

### Installation method

None

### Nextcloud Server version

28

### Operating system

None

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

None

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Additional info

Filter do not do it's job:

<img width="1040" alt="image" src="https://github.com/nextcloud/server/assets/13381981/7fc3066a-1924-4a24-ae50-82250d56683b">

sorbaugh:
Looping in @Altahrim and @Fenn-CS as this seems related to Unified Search

Pytal:
Had a look and filtering in the disabled users section doesn't work like it does in "Active users" because we use an optimized endpoint for querying disabled users which unlike the general users endpoint does not accept a search query

We would need to accept a search query here https://github.com/nextcloud/server/blob/880f33ec2dc0fa535872d394b386acb337f6b4e0/lib/public/User/Backend/IProvideEnabledStateBackend.php#L55 and update the LDAP implementation cc @come-nc

[...]

PR #45092, opened 2024-04-29
fix(search): show user search provider for deactivated and admin pages

Altahrim:
## Summary
Display user search provider in "Places" when browsing disabled and admin users

Backend part of #44936


susnux (approved):
makes sense

artonge (approved):

PR #45094, opened 2024-04-29
feat: TaskProcessing API

marcelklehr:
## TODO

- [x] Clean up old tasks
- [x] Clean up old app data files
- [x] Add integration tests
- [x] Check if user has access to files
- [x] OCS API

## Checklist


lib/public/TaskProcessing/Events/AbstractTextProcessingEvent.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+namespace OCP\TaskProcessing\Events;
+
+use OCP\EventDispatcher\Event;
+use OCP\TaskProcessing\Task;
+
+/**
+ * @since 30.0.0
+ */
+abstract class AbstractTextProcessingEvent extends Event {
```
julien-nc, line 34:
```suggestion
abstract class AbstractTaskProcessingEvent extends Event {
```

Or maybe it's intentional?

julien-nc (commented):
Yay, it looks gooood. Didn't make a full review yet, just using it for the work on the assistant.
It breaks NC though. I tried to rebase on master, no luck.
```
PHP Fatal error:  Class OC\\Files\\Node\\LazyR
oot contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (OCP\\Files\\IRootFolder::getAppDataDirectoryName) in /var/www/html/dev
/server/lib/private/Files/Node/LazyRoot.php on line 39
```

lib/public/TaskProcessing/Events/AbstractTextProcessingEvent.php
marcelklehr: oopsa

lib/public/TaskProcessing/IProvider.php
```diff
@@ -0,0 +1,77 @@
...
+	public function getName(): string;
+
+	/**
+	 * Returns the task type id of the task type, that this
+	 * provider handles
+	 *
+	 * @since 30.0.0
+	 * @return string
+	 */
+	public function getTaskType(): string;
```
julien-nc, line 54:
```suggestion
	public function getTaskTypeId(): string;
```
Maybe

marcelklehr: :+1:

lib/public/TaskProcessing/Task.php
```diff
@@ -0,0 +1,259 @@
...
+		return $this->userId;
+	}
+
+	/**
+	 * @psalm-return array{id: ?int, status: self::STATUS_*, userId: ?string, appId: string, input: ?array<array-key, mixed>, output: ?array<array-key, mixed>, identifier: ?string, completionExpectedAt: ?int, progress: ?float}
+	 * @since 30.0.0
+	 */
+	public function jsonSerialize(): array {
+		return [
+			'id' => $this->getId(),
```
julien-nc, line 201:
```suggestion
			'id' => $this->getId(),
			'type' => $this->getTaskTypeId(),
```
We need the type here otherwise it's not included when getting a task with `/ocs/v2.php/taskprocessing/task/TASK_ID` for example.

julien-nc (changes requested):
A bump of the server version is needed to trigger the migration step.

lib/public/TaskProcessing/Task.php
marcelklehr: Wow. Nice catch!

marcelklehr:
> A bump of the server version is needed to trigger the migration step.

Like this?

version.php
```diff
-$OC_Version = [30, 0, 0, 0];
+$OC_Version = [30, 0, 0, 1];
```

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,35 @@
...
+ *     type: int
+ * }
+ *
+ * @psalm-type CoreTaskProcessingTaskType = array{
+ *     name: string,
+ *     description: string,
+ *     inputShape: CoreTaskProcessingShape[],
+ *     optionalInputShape: CoreTaskProcessingShape[],
+ *     outputShape: CoreTaskProcessingShape[],
+ *     optionalOutputShape: CoreTaskProcessingShape[],
```
provokateurin, line 192: Having two separate arrays for optional/non-optional is a bit annoying to use as an API consumer. I'd prefer a simple boolean value in the CoreTaskProcessingShape that indicates if it is optional or not.

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,35 @@
...
+ * }
+ *
+ * @psalm-type CoreTaskProcessingTask = array{
+ *     id: ?int,
```
provokateurin, line 196: How can the id be null? I haven't looked at the implementation, but this looks counter intuitive to me.

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,35 @@
...
+ *     type: string,
+ *     status: 0|1|2|3|4|5,
+ *     userId: ?string,
+ *     appId: string,
+ *     input: ?array<string, mixed>,
+ *     output: ?array<string, mixed>,
+ *     identifier: ?string,
```
provokateurin, line 203: id is an abbreviation for identifier, so this field name is really confusing.

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,35 @@
...
+ *     completionExpectedAt: ?int,
+ *     progress: ?float
```
provokateurin, line 205: Is this 0-1 or 0-100? In the later case I'd prefer an int without any decimal places (as that would likely be imprecise anyway and not helpful for the end user).

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,35 @@
...
+ *     inputShape: CoreTaskProcessingShape[],
+ *     optionalInputShape: CoreTaskProcessingShape[],
+ *     outputShape: CoreTaskProcessingShape[],
+ *     optionalOutputShape: CoreTaskProcessingShape[],
+ * }
+ *
+ * @psalm-type CoreTaskProcessingTask = array{
+ *     id: ?int,
+ *     type: string,
+ *     status: 0|1|2|3|4|5,
```
provokateurin, line 198: Could the integer statuses be replaced by discrete strings? That is more human readable and also easier to work with as an API consumer because it is already clear from the value what the status means without looking into the source code or docs.

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,357 @@
...
+	#[ApiRoute(verb: 'GET', url: '/tasktypes', root: '/taskprocessing')]
+	public function taskTypes(): DataResponse {
+		$taskTypes = $this->taskProcessingManager->getAvailableTaskTypes();
+
+		$serializedTaskTypes = [];
+		foreach ($taskTypes as $key => $taskType) {
+			$serializedTaskTypes[$key] = [
+				'name' => $taskType['name'],
+				'description' => $taskType['description'],
+				'inputShape' => array_map(fn (ShapeDescriptor $descriptor) => $descriptor->jsonSerialize(), $taskType['inputShape']),
```
come-nc, line 86: Do you really need to do that by hand? 
I’m pretty sure json_encode will call jsonSerialize recursively.
Or is it because DataResponse does not always serialize to json?

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,357 @@
...
+		$ids = [];
+		$taskTypes = $this->taskProcessingManager->getAvailableTaskTypes();
+		if (!isset($taskTypes[$task->getTaskTypeId()])) {
+			throw new \OCP\TaskProcessing\Exception\NotFoundException('Could not find task type');
+		}
+		$taskType = $taskTypes[$task->getTaskTypeId()];
+		foreach ($taskType['inputShape'] + $taskType['optionalInputShape'] as $key => $descriptor) {
+			if (in_array(EShapeType::getScalarType($descriptor->getShapeType()), [EShapeType::File, EShapeType::Image, EShapeType::Audio, EShapeType::Video], true)) {
+				if (is_array($task->getInput()[$key])) {
+					$ids += $task->getInput()[$key];
```
come-nc, line 276:
```suggestion
				$value = $task->getInput()[$key];
				if (is_array($value)) {
					$ids += $value;
```
Avoid calling the method several times.

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,357 @@
...
+				} else {
+					$ids[] = $task->getInput()[$key];
+				}
+			}
+		}
+		if ($task->getOutput() !== null) {
+			foreach ($taskType['outputShape'] + $taskType['optionalOutputShape'] as $key => $descriptor) {
+				if (in_array(EShapeType::getScalarType($descriptor->getShapeType()), [EShapeType::File, EShapeType::Image, EShapeType::Audio, EShapeType::Video], true)) {
+					if (is_array($task->getInput()[$key])) {
+						$ids += $task->getOutput()[$key];
```
come-nc, line 286: This will also avoid inconsistencies like this `getInput` vs `getOutput` confusion.

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,357 @@
...
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTaskProcessingTask}, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND, array{message: string}, array{}>
+	 *
+	 *  200: File content returned
+	 *  404: Task not found
+	 */
+	#[NoAdminRequired]
+	#[ApiRoute(verb: 'POST', url: '/tasks/{taskId}/result', root: '/taskprocessing')]
+	public function setResult(int $taskId, ?array $output = null, ?string $errorMessage = null): DataResponse {
+		try {
+			$this->taskProcessingManager->getUserTask($taskId, $this->userId);
```
come-nc, line 341: Is this call necessary? If yes add a comment explaining why.

core/Controller/TaskProcessingApiController.php
provokateurin: It does, so this is indeed not necessary and it shouldn't be called explicitly to avoid returning json inside xml when the API is used in XML mode.

core/Migrations/Version30000Date20240429122720.php
```diff
@@ -0,0 +1,115 @@
...
+				'default' => 0,
+			]);
+			$table->addColumn('error_message', Types::STRING, [
+				'notnull' => false,
+				'length' => 255,
+			]);
+
+			$table->setPrimaryKey(['id'], 'tasks_id_index');
+			$table->addIndex(['status', 'type'], 'tasks_status_type');
+			$table->addIndex(['last_updated'], 'tasks_updated');
```
come-nc, line 107: Name seems dangerously generic, might clash no? Don’t we have a naming pattern or something?

lib/composer/composer/autoload_static.php
```diff
@@ -17,6 +17,10 @@ class ComposerStaticInit749170dad3f5e7f9ca158f5a9f04f6a2
             'OC\\' => 3,
             'OCP\\' => 4,
         ),
+        'B' => 
+        array (
+            'Bamarni\\Composer\\Bin\\' => 21,
+        ),
```
come-nc, line 23: Should not appear here I think, not sure.

lib/private/AppFramework/Bootstrap/RegistrationContext.php
```diff
@@ -163,6 +163,12 @@ class RegistrationContext {
 	/** @var ServiceRegistration<ITeamResourceProvider>[] */
 	private array $teamResourceProviders = [];
 
+	/** @var ServiceRegistration<\OCP\TaskProcessing\IProvider>[] */
+	private $taskProcessingProviders = [];
```
come-nc, line 167:
```suggestion
	private array $taskProcessingProviders = [];
```

lib/private/AppFramework/Bootstrap/RegistrationContext.php
```diff
@@ -163,6 +163,12 @@ class RegistrationContext {
...
+
+	/** @var ServiceRegistration<\OCP\TaskProcessing\ITaskType>[] */
+	private $taskProcessingTaskTypes = [];
```
come-nc, line 170:
```suggestion
	private array $taskProcessingTaskTypes = [];
```

core/Controller/TaskProcessingApiController.php
provokateurin: The jsonSerialize method is abused a little bit because it doesn't actually return json and only converts it to an array. If the API is used in json mode then it wouldn't be necessary, but for xml it is. It's also needed for psalm because it doesn't know the method would be called automatically and thus would only see the class and not the array.

lib/private/TaskProcessing/Db/Task.php
```diff
@@ -0,0 +1,134 @@
...
+
+	public function toRow(): array {
+		return array_combine(self::$columns, array_map(function ($field) {
+			return $this->{'get'.ucfirst($field)}();
+		}, self::$fields));
+	}
+
+	public static function fromPublicTask(OCPTask $task): Task {
+		/** @var Task $taskEntity */
+		$taskEntity = Task::fromParams([
```
come-nc, line 103:
```suggestion
	public static function fromPublicTask(OCPTask $task): self {
		/** @var Task $taskEntity */
		$taskEntity = self::fromParams([
```

lib/private/TaskProcessing/Manager.php
```diff
@@ -0,0 +1,855 @@
...
+		}
+
+		$providers = [];
+
+		foreach ($context->getTaskProcessingProviders() as $providerServiceRegistration) {
+			$class = $providerServiceRegistration->getService();
+			try {
+				/** @var IProvider $provider */
+				$provider = $this->serverContainer->get($class);
+				$providers[$provider->getId()] = $provider;
```
come-nc, line 388: No need to check for id clash?

lib/private/TaskProcessing/RemoveOldTasksBackgroundJob.php
```diff
@@ -0,0 +1,67 @@
...
+		try {
+			$this->taskMapper->deleteOlderThan(self::MAX_TASK_AGE_SECONDS);
+		} catch (\OCP\DB\Exception $e) {
+			$this->logger->warning('Failed to delete stale task processing tasks', ['exception' => $e]);
+		}
+		try {
+			$this->clearFilesOlderThan($this->appData->getFolder('text2image'), self::MAX_TASK_AGE_SECONDS);
+			$this->clearFilesOlderThan($this->appData->getFolder('audio2text'), self::MAX_TASK_AGE_SECONDS);
+			$this->clearFilesOlderThan($this->appData->getFolder('TaskProcessing'), self::MAX_TASK_AGE_SECONDS);
+		} catch (NotFoundException $e) {
```
come-nc, line 45: When does this happen?
If cleaning text2image throws, audio2text will never be cleaned.

core/Controller/TaskProcessingApiController.php
come-nc: That’s unfortunate. Most of the code in the API controller is serializing by hand.

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,357 @@
...
+	#[UserRateLimit(limit: 20, period: 120)]
+	#[AnonRateLimit(limit: 5, period: 120)]
+	#[ApiRoute(verb: 'POST', url: '/schedule', root: '/taskprocessing')]
+	public function schedule(array $input, string $type, string $appId, string $identifier = ''): DataResponse {
+		$task = new Task($type, $input, $appId, $this->userId, $identifier);
+		try {
+			$this->taskProcessingManager->scheduleTask($task);
+
+			/** @var CoreTaskProcessingTask $json */
+			$json = $task->jsonSerialize();
```
provokateurin, line 123: This should be absolutely avoided to prevent shadowing mismatches. Annotate the return type of the jsonSerialize instead to get it working.

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,357 @@
...
+	#[ApiRoute(verb: 'DELETE', url: '/task/{id}', root: '/taskprocessing')]
+	public function deleteTask(int $id): DataResponse {
+		try {
+			$task = $this->taskProcessingManager->getUserTask($id, $this->userId);
+
+			$this->taskProcessingManager->deleteTask($task);
+
+			return new DataResponse([]);
+		} catch (\OCP\TaskProcessing\Exception\NotFoundException $e) {
+			return new DataResponse([]);
```
provokateurin, line 188: Return null instead of an empty array whenever possible (unless it is actually an empty array).

core/ResponseDefinitions.php
marcelklehr: This naming is consistent with previous APIs, e.g. the text processing API. The purpose of the property is for apps to be able to reference other entities in their tasks without having to create a new table mapping task id to their internal id.

core/ResponseDefinitions.php
marcelklehr: It's 0-1

core/ResponseDefinitions.php
provokateurin: I see, I'd still prefer if a more expressive name would be used. We don't need to align this with existing APIs, we can fix it right here.

core/Migrations/Version30000Date20240429122720.php
marcelklehr: I've never heard of a naming pattern tbh, would be great

core/Controller/TaskProcessingApiController.php
marcelklehr: I know. I wasn't able to get it working without this.

lib/private/TaskProcessing/RemoveOldTasksBackgroundJob.php
marcelklehr: It happens when no tasks involving files or the mentioned task types have been run yet as the folders are created on demand.

core/Migrations/Version30000Date20240429122720.php
provokateurin: Yeah, the index names need to be completely unique for postgres

lib/composer/composer/autoload_static.php
provokateurin: Just don't commit these files (or run `composer i --no-dev` but that also causes a different diff)

core/Controller/TaskProcessingApiController.php
provokateurin: I'm sure this is possible, I'll give it a try

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,361 @@
...
+
+		return new DataResponse([
+			'types' => $serializedTaskTypes,
+		]);
+	}
+
+	/**
+	 * This endpoint allows scheduling a task
+	 *
+	 * @param array<string, mixed> $input Input text
```
provokateurin, line 101: Not only text, right?

core/Controller/TaskProcessingApiController.php
```diff
@@ -0,0 +1,361 @@
...
+		} catch (\OCP\TaskProcessing\Exception\NotFoundException $e) {
+			return new DataResponse(['message' => $this->l->t('Not found')], Http::STATUS_NOT_FOUND);
+		} catch (GenericFileException|NotPermittedException|LockedException|Exception $e) {
+			return new DataResponse(['message' => $this->l->t('Internal error')], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
+	/**
+	 * @param Task $task
+	 * @return list<mixed>
```
provokateurin, line 263: Should be int and not mixed

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,35 @@
...
+ * }
+ *
+ * @psalm-type CoreTaskProcessingTask = array{
+ *     id: int,
+ *     type: string,
+ *     status: 0|1|2|3|4|5,
+ *     userId: ?string,
+ *     appId: string,
+ *     input: ?array<string, mixed>,
+ *     output: ?array<string, mixed>,
```
provokateurin, line 202: I'm not happy about these mixed types, can we narrow them down in a sensible way?

provokateurin (changes requested):

core/ResponseDefinitions.php
marcelklehr: Yeah, I was just lazy.. I'll make it more specific :)

provokateurin: You could also define them as a separate type alias so you can easily reuse them everywhere.

lib/private/TaskProcessing/RemoveOldTasksBackgroundJob.php
come-nc: And you know for a fact that all 3 folders are created together, it’s not possble to have only one of them?

marcelklehr: No, I already fixed it :)

core/Controller/TaskProcessingApiController.php
marcelklehr: because it is nullable. When a consumer creates a task, they don't know the ID of the task. The id is only set after it's scheduled.

provokateurin: But isn't it referring to the ID in the database which must be know right after creation? Otherwise it should be fine to have it nullable if the API can return tasks that have it their ID set to null.

marcelklehr: the Task class represents a data object which exists before the data is entered into the db

core/ResponseDefinitions.php
julien-nc: identificator? :grin: 
custom_id?

lib/composer/composer/autoload_static.php
marcelklehr: The point of committing these files is to appease the autoloader checker

core/ResponseDefinitions.php
marcelklehr: i do think being consistent between APIs of the same software project is a value worth striving for.

lib/composer/composer/autoload_static.php
marcelklehr: cc @ChristophWurst as the person who introduced barmani's composer plugin

core/ResponseDefinitions.php
marcelklehr: I vote for keeping it like this, because I'm lazy

core/ResponseDefinitions.php
marcelklehr: Should be better now

core/ResponseDefinitions.php
provokateurin: In my opinion single arrays make it a lot easier to iterate and create UI elements for. I'm not vetoing, but I also don't see the point in not making it a bit nicer.

core/ResponseDefinitions.php
provokateurin: In principle yes, but I think it was already wrong to call it `id` and `identifier` in the previous APIs, so we can fix it here. I'd go with `custom_id` like Julien suggested.

lib/composer/composer/autoload_static.php
marcelklehr: And how should I know what to add? Doesn't seem like an ideal process

core/ResponseDefinitions.php
marcelklehr: Alright, I'm not willing to have a fight over this.

lib/composer/composer/autoload_static.php
ChristophWurst: It's far from ideal. The ideal solution would be to stop vendoring PHP dependencies.

core/Migrations/Version30000Date20240429122720.php
```diff
@@ -0,0 +1,115 @@
...
+			]);
+			$table->addColumn('error_message', Types::STRING, [
+				'notnull' => false,
+				'length' => 255,
+			]);
+
+			$table->setPrimaryKey(['id'], 'taskp_tasks_id_index');
+			$table->addIndex(['status', 'type'], 'taskp_tasks_status_type');
+			$table->addIndex(['last_updated'], 'taskp_tasks_updated');
+			$table->addIndex(['user_id', 'app_id', 'custom_id'], 'taskp_tasks_uid_appid_ident');
```
julien-nc, line 108:
```suggestion
			$table->addIndex(['user_id', 'app_id', 'custom_id'], 'taskp_tasks_uid_appid_cid');
```
Nitpicking

julien-nc (changes requested):

core/ResponseDefinitions.php
marcelklehr: :+1:

julien-nc (changes requested):
Another partial review:
* We could use (optional) default values for input shape fields. It might not make sense in many cases but for example: "number of images" for TextToImage could have 1 as default value. Wdyt?
* The assistant needs a manager method AND/OR an endpoint to get all user's tasks (with a task type ID optional filter). Wdyt?
* [minor] CSRF might not be needed for `taskprocessing/tasks/{taskId}/file/{fileId}`, not a big deal, I currently add the requesttoken GET param to display images in the UI.
* Task result files stored in data/appdata_iid/core/TaskProcessing/ are empty for the TextToImage task type (at least). I tried 2 providers (OpenAI and Replicate which are providers for the old TTI API).
* The manager is unable to read input files if they are in a user's storage. I tried an STT provider (for the old STT API), I get this exception raised: https://github.com/nextcloud/server/blob/enh/taskprocessing-api/lib/private/TaskProcessing/Manager.php#L763 which means this does not get the file: https://github.com/nextcloud/server/blob/enh/taskprocessing-api/lib/private/TaskProcessing/Manager.php#L759 . Happy to debug that together if you want.

marcelklehr:
* The assistant needs a manager method AND/OR an endpoint to get all user's tasks **fixed** :heavy_check_mark: 
 * [minor] CSRF might not be needed for taskprocessing/tasks/{taskId}/file/{fileId} **fixed** :heavy_check_mark: 
 * Task result files stored in data/appdata_iid/core/TaskProcessing/ are empty for the TextToImage task type. **fixed** :heavy_check_mark:

marcelklehr:
> The manager is unable to read input files if they are in a user's storage.

Can you check if it works now?

lib/private/TaskProcessing/Manager.php
```diff
@@ -0,0 +1,896 @@
...
+		} catch (MultipleObjectsReturnedException|\OCP\DB\Exception $e) {
+			throw new \OCP\TaskProcessing\Exception\Exception('There was a problem finding the task', 0, $e);
+		} catch (\JsonException $e) {
+			throw new \OCP\TaskProcessing\Exception\Exception('There was a problem parsing JSON after finding the task', 0, $e);
+		}
+	}
+
+	public function getUserTasks(?string $userId, ?string $taskTypeId = null, ?string $customId = null): array {
+		try {
+			$taskEntities = $this->taskMapper->findUserTasksByApp($userId, $taskTypeId, $customId);
```
julien-nc, line 815:
```suggestion
			$taskEntities = $this->taskMapper->findByUserAndTaskType($userId, $taskTypeId, $customId);
```

lib/public/TaskProcessing/IManager.php
```diff
@@ -0,0 +1,172 @@
...
+	 * @return Task
+	 * @throws Exception If the query failed
+	 * @throws NotFoundException If the task could not be found
+	 * @since 30.0.0
+	 */
+	public function getUserTask(int $id, ?string $userId): Task;
+
+	/**
+	 * @param string|null $userId The user id that scheduled the task
+	 * @param string|null $taskTypeId The task type id to filter by
```
julien-nc, line 140:
```suggestion
	 * @param string|null $taskTypeId The task type id to filter by
	 * @param string|null $customId
```

julien-nc (changes requested):
> The assistant needs a manager method AND/OR an endpoint to get all user's tasks fixed ✔️

Yay. Small mistake. It works with my suggested change.

> [minor] CSRF might not be needed for taskprocessing/tasks/{taskId}/file/{fileId} fixed ✔️

Thanks :ok_hand: 

> Task result files stored in data/appdata_iid/core/TaskProcessing/ are empty for the TextToImage task type. fixed ✔️

Yay, fixed.

> The manager is unable to read input files if they are in a user's storage

Still fails

[read.input.file.txt](https://github.com/nextcloud/server/files/15274330/read.input.file.txt)

julien-nc:
@marcelklehr 

* How do we deal with context write? I think it makes sense now to have a ContextWrite task type defined in server so we can have specific instruct strings in each provider. It would also remove one more "custom UI" and "custom logic" in the assistant. :soap: (this is a soap)
* We should talk about how/when `SynchronousBackgroundJob` is queued. As it is always in the job list, the worker constantly runs it which is weird when there is no task. Do you think it would be possible to only queue it when needed? I'm thinking:
    * Queue the job when scheduling a task
    * The job queues itself only if there is still at least one scheduled task

marcelklehr:
> I think it makes sense now to have a ContextWrite task type defined in server

Great idea!

> Do you think it would be possible to only queue it when needed?

Totally makes sense!

marcelklehr:
> The manager is unable to read input files if they are in a user's storage.

Can you check if it works now?

julien-nc:
@marcelklehr Now I get `OCA\\OpenAi\\SpeechToText\\STTProvider::transcribeFile(): Argument #1 ($file) must be of type OCP\\Files\\File, false given, called in /blabla/server/lib/private/TaskProcessing/Manager.php on line 364`.

You can reproduce by using the STT provider in integration_openai and the `enh/task-processing-api-support` branch of the assistant (don't forget to compile the JS). I can give you an OpenAI API key if you want.

marcelklehr:
That's a new error :tada: and a different bug. Now it should work.

julien-nc:
About queuing only when needed: So now the job is queued each time a task is scheduled and "re-queued" when it finishes if there still are some scheduled tasks, right?
If so, in the manager::schedule method, the bg job should probably better be scheduled only if it's not already scheduled, otherwise we can end up with the job being queued multiple times. Wdyt?

Thanks for adding the ContextWrite task type!

marcelklehr:
> If so, in the manager::schedule method, the bg job should probably better be scheduled only if it's not already scheduled, otherwise we can end up with the job being queued multiple times.

I don't think it  can be queued multiple times, actually... Let me check

julien-nc:
> I don't think it can be queued multiple times, actually... Let me check Yep, jobs are unique by class and argument.

Yay!

> I don't think it can be queued multiple times, actually... Let me check Yep, jobs are unique by class and argument.

It works now! You're faster than light.

julien-nc (approved):

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,34 @@
  *       iconURL: ?string,
  *       iconEmoji: ?string,
  *   }
+ *
+ * @psalm-type CoreTaskProcessingShape = array{
+ *     name: string,
+ *     description: string,
+ *     type: int,
```
provokateurin, line 183: Could this be a string enum instead? That would make it a lot easier to use IMO

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,34 @@
...
+ *     mandatory: bool,
```
provokateurin, line 184: I'd say `required` is a more common term for this, but it doesn't really make a difference.

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,34 @@
...
+ * }
+ *
+ * @psalm-type CoreTaskProcessingTaskType = array{
+ *     name: string,
+ *     description: string,
+ *     inputShape: CoreTaskProcessingShape[],
+ *     outputShape: CoreTaskProcessingShape[],
```
provokateurin, line 191:
```suggestion
 *     inputShapes: CoreTaskProcessingShape[],
 *     outputShapes: CoreTaskProcessingShape[],
```
Since they are an array the names should be plural as well.

core/ResponseDefinitions.php
```diff
@@ -176,6 +176,34 @@
...
+ * }
+ *
+ * @psalm-type CoreTaskProcessingTask = array{
+ *     id: int,
+ *     type: string,
+ *     status: 'STATUS_CANCELLED'|'STATUS_FAILED'|'STATUS_SUCCESSFUL'|'STATUS_RUNNING'|'STATUS_SCHEDULED'|'STATUS_UNKNOWN',
+ *     userId: ?string,
+ *     appId: string,
+ *     input: array<string, numeric|list<numeric>|string|list<string>>,
+ *     output: ?array<string, numeric|list<numeric>|string|list<string>>,
```
provokateurin, line 201: I think it would make sense to move this type to a separate type and reuse it

provokateurin, line 201:
```suggestion
 *     output: null|array<string, numeric|list<numeric>|string|list<string>>,
```
The question mark can be overlooked quickly and for me personally it's hard to parse if it means the first type can also be null or the whole type can be null, so this makes it clear what you want (though both should work).

provokateurin (changes requested):

core/ResponseDefinitions.php
marcelklehr: I disagree. It's the shape of the input. It's one input (because it's input once) so one shape.

provokateurin (approved):

core/ResponseDefinitions.php
provokateurin: Hm that makes sense, how about just `input` and `output`?

marcelklehr: Well it's not an input but just the shape of the input to be :)

marcelklehr:
:tada: :blue_heart:

provokateurin:
(Please note my last comment before merging)

lib/public/Files/IRootFolder.php
```diff
@@ -98,4 +98,10 @@ public function getNodeFromCacheEntryAndMount(ICacheEntry $cacheEntry, IMountPoi
 	 * @since 28.0.0
 	 */
 	public function getMount(string $mountPoint): IMountPoint;
+
+	/**
+	 * @return string
+	 * @since 30.0.0
+	 */
+	public function getAppDataDirectoryName(): string;
```
nickvergessen, line 106: @marcelklehr Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(all changes in lib/public/)

PR #45095, opened 2024-04-29
feat(files): Allow to upload folders

### LINKED ISSUE #14707: Upload folder menu option
It will be good and nice to have an "Upload folder" option in Nextcloud's "plus menu" between "Upload file" and "New folder" menu options.

Why? Because isn't intuitive that this option missing. Users can think that Nexcloud can not upload whole folders via web interface (yes, we have drag&drop option but sometime is better to have this option and dialog box to choose folder).  

BTW, drag&drop function sometimes do not work. I drag my folder on browser's screen, but after dropping nothing happened. So, I must repeat my drag&drop several times to success.

And maybe it will be better to rename "Upload file" option into "Upload file(s)" because users can think that only one file can be transferred what is not true!

susnux:
* Resolves https://github.com/nextcloud/server/issues/14707
* [x] Requires: https://github.com/nextcloud-libraries/nextcloud-upload/pull/1175

## Summary
Add support for uploading folders using the "new" menu.

## Checklist


apps/files/src/views/FilesList.vue
```diff
@@ -215,11 +221,16 @@ export default defineComponent({
...
-			_unsubscribeStore: () => {},
+			unsubscribeStore: () => {},
 		}
 	},
 
 	computed: {
+		getContent() {
+			const view = this.currentView
+			return async (path?: string) => (await view.getContents(normalize(`${this.currentFolder?.path ?? ''}/${path ?? ''}`))).contents
+		},
```
skjnldsv, line 232: Please document or expand this :see_no_evil: 
This is an important computed and we should be more verbose imho :)

susnux: Sure, this is currently only for testing the library PR

apps/files/src/views/FilesList.vue
```diff
@@ -37,10 +37,12 @@
...
 					<!-- Uploader -->
 					<UploadPicker v-else-if="currentFolder"
-						:content="dirContents"
-						:destination="currentFolder"
-						:multiple="true"
+						allow-folders
 						class="files-list__header-upload-button"
+						:content="getContent"
+						:destination="currentFolder"
+						:forbidden-characters="forbiddenCharacters"
```
artonge, line 44: Shouldn't this be built in the lib?
In Photos we are using the uploader, it would be nice to have this by default

susnux: Yes the `forbidden-characters` can be replaced by nextcloud-files version.

ShGKme (approved):
Tested

apps/files/src/views/FilesList.vue
susnux: We can do so in a new uploader version :)

artonge (approved):

PR #45126, opened 2024-04-30
fix(Authentication): correctly catch uniq constraint on token insert

ShGKme:
## TODO

- [x] Migrate from `UniqueConstraintViolationException` to `OCP\DB\Exception` in auth token manager

## Checklist

### DIFF
--- a/lib/private/Authentication/Token/Manager.php
+++ b/lib/private/Authentication/Token/Manager.php
@@ -55,6 +55,7 @@ class Manager implements IProvider, OCPIProvider {
 	 * @param int $type token type
 	 * @param int $remember whether the session token should be used for remember-me
 	 * @return OCPIToken
+	 * @throws Exception
 	 */
 	public function generateToken(string $token,
 		string $uid,
--- a/lib/private/Authentication/Token/Manager.php
+++ b/lib/private/Authentication/Token/Manager.php
@@ -27,7 +27,6 @@ declare(strict_types=1);
  */
 namespace OC\Authentication\Token;
 
-use Doctrine\DBAL\Exception\UniqueConstraintViolationException;
 use OC\Authentication\Exceptions\InvalidTokenException as OcInvalidTokenException;
 use OC\Authentication\Exceptions\PasswordlessTokenException;
 use OCP\Authentication\Exceptions\ExpiredTokenException;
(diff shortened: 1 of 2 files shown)


come-nc (approved):

lib/private/Authentication/Token/Manager.php
```diff
@@ -77,18 +77,22 @@ public function generateToken(string $token,
 				$type,
 				$remember
 			);
```
come-nc, line 79: Maybe add an `@throws OCP\DB\Exception` on `generateToken` because it’s not documented currently I think

come-nc:
I though `executeStatement` was a drop-in replacement for `execute` but it does change the exception thrown so I will try to be careful about that in the future.

nickvergessen:
> I though executeStatement was a drop-in replacement for execute but it does change the exception thrown so I will try to be careful about that in the future.

Is there something like a `@never-throws X` and then psalm complains when you `catch X` ?

lib/private/Authentication/Token/Manager.php
ShGKme:
```php
* @throws Exception
```

should also work here with use above, right?

nickvergessen (approved):

lib/private/Authentication/Token/Manager.php
```diff
@@ -55,6 +55,7 @@ public function __construct(PublicKeyTokenProvider $publicKeyTokenProvider) {
 	 * @param int $type token type
 	 * @param int $remember whether the session token should be used for remember-me
 	 * @return OCPIToken
+	 * @throws Exception
```
nickvergessen, line 58: This is not part of the expected API behaviour, so from my POV it should not be documented

lib/private/Authentication/Token/Manager.php
nickvergessen: It is not part of the expected API behaviour, so from my POV it should not be documented

come-nc: I was talking about the other generateToken, the one that is called, so https://github.com/nextcloud/server/blob/c31d6b1fdc502844ab907355bb553feb451cdda3/lib/private/Authentication/Token/PublicKeyTokenProvider.php#L104
But it inherits from IProvider, so not sure if easy to document.

PR #45129, opened 2024-04-30
fix(core): add fs_name_hash index to migration

pulsejet:
Automatically adds the `fs_name_hash` index for new installations. See https://github.com/nextcloud/server/pull/44586#issuecomment-2082281865


kesselb (approved):
Thank you

susnux (approved):

PR #45138, opened 2024-05-01
retry smb notify test a few times

icewind1991:
I'm out of ideas on how to make the test more reliable so instead just give it 5 attempts.


icewind1991:
re-ran the tests a few times and it seems to be reliable enough now

come-nc (approved):

skjnldsv (approved):

PR #45140, opened 2024-05-01
fix: fix FileAccess::getByFileId(InStorage)

icewind1991:


kesselb (approved):

lib/private/Files/Cache/FileAccess.php
```diff
@@ -56,7 +56,7 @@ private function getQuery(): CacheQueryBuilder {
 	}
 
 	public function getByFileIdInStorage(int $fileId, int $storageId): ?CacheEntry {
-		$items = $this->getByFileIdsInStorage([$fileId], $storageId);
+		$items = array_values($this->getByFileIdsInStorage([$fileId], $storageId));
 		return $items[0] ?? null;
```
come-nc, line 60:
```suggestion
		return reset($items) ?: null;
```
And then you do not need the `array_values` call.
But no idea if it’s more performant.

It just sucks that array_value_first did not make it in :crying_cat_face: https://wiki.php.net/rfc/array_key_first_last#vote

come-nc (approved):

lib/private/Files/Cache/FileAccess.php
susnux: Just a quick and direct benchmark with
* a and aa are arrays with 4096 elements internal pointer on element 1024
* b and bb are empty arrays

```
$benchmark = benchmark(function() use ($a, $b, &$c, &$d) {
    $f = array_values($a);
	$c = $f[0] ?? null;

	$e = array_values($b);
	$d = $e[0] ?? null;
});

$benchmark2 = benchmark(function() use ($aa, $bb, &$c, &$d) {
    $c = reset($aa) ?: null;
	$d = reset($bb) ?: null;
});
```

susnux: Performance seems to be quite the same, ~10% difference (faster `array_values`) on my system.

susnux (approved):

PR #45151, opened 2024-05-02
feat: Add getForbiddenFilenames to \OCP\Util and unify filename sanitizing

### LINKED ISSUE #44963: 🪟📁📄 Ability to enforce windows compatible file names on the server
### How to use GitHub

In some scenarios with a large chunk of the user base being Windows users, using the Desktop client for example sync conflict can get created easily via the web interface, other clients or directly via any client software supporting WebDAV by creating files or folders that cannot be synced down to Windows ever.

Add a setting to 'enforce windows compatible file names' on the server, so when the user tries to rename/create files/folders that would generate conflict when syncing, the server would prevent the action. This setting should be optional and needs to be turned on explicitly.
Also there is no logic expected to migrate existing data, just to prevent it from the moment on the setting is active.

- [ ] https://github.com/nextcloud/server/issues/40690 should be a stretch goal
- [x] define list invalid characters
- [ ] https://github.com/nextcloud/server/issues/45130

None, given the fact that the person creating a file/folder might not be the one being affected by it and the affected person might not have the permissions on the system to change the names, i.e. when shared read-only.

susnux:
Some preparations and refactoring for #44963 

## Summary
* Add a common access to the forbidden filename config (now `\OCP\Util::getForbiddenFilenames()`)
* Cache forbidden filenames and characters, as with #44963 the function for calculating those values might be more complex (also the forbidden filenames were already cached on the filesystem class).
* Drop `OC_Util::isValidFilename` as it was deprecated for a long time, not used anymore and not checking forbidden filenames
* Deprecate `\OCP\Files\FileInfo::BLACKLIST_FILES_REGEX` as this was not used anywhere (only in the util function but that is also not used anywhere + in the files app, but never enforced on webdav).
* Add the forbidden filenames and forbidden characters to the JS config for frontend validation use

## Todo :warning: - Needs discussion
- [x] It was never documented but the deny list of files was somewhat case in-sensitive. Document that or make it case-sensitive? (The first is currently done).
- [x] The file name blacklist regex was not enforced anywhere, meaning you can currently upload `.fileinfo` files. So I dropped it, but should be instead enfore it?

## Checklist


tests/lib/Files/FilesystemTest.php
```diff
@@ -284,7 +284,7 @@ public function isFileBlacklistedData() {
 			['/etc/foo\bar/.htaccess/', true],
 			['/etc/foo\bar/.htaccess/foo', false],
 			['//foo//bar/\.htaccess/', true],
-			['\foo\bar\.HTAccess', true],
```
artonge, line 287: Wasn't this on purpose to test for case insensitivity?

tests/lib/UtilTest.php
```diff
@@ -169,10 +266,8 @@ public function filenameValidationProvider() {
 			['. ', false],
 			[' .', false],
 
-			// part files not allowed
-			['.part', false],
-			['notallowed.part', false],
-			['neither.filepart', false],
```
artonge, line 175: Why are those removed?

susnux: Because they were never used, nothing enforced those rules as the only check that is currently used (IStorage::verifyPath) does not use the black list regex.

Meaning you can even right now on e.g. 28 upload a file ending with `.filepath`.

susnux: Fixed

lib/public/Util.php
```diff
@@ -489,36 +495,105 @@ public static function uploadLimit(): int|float {
 	}
 
 	/**
-	 * Get a list of characters forbidden in file names
+	 * Get a list of forbidden filename extensions that must not be used
+	 * This list should be checked case-insensitive, all names are returned lowercase.
+	 * @return string[]
```
come-nc, line 500:
```suggestion
	 * @return list<string>
```
Same type change on all private vars and public getters.

lib/public/Util.php
```diff
@@ -489,36 +495,105 @@ public static function uploadLimit(): int|float {
...
+			return false;
+		}
+
+		if (\OC\Files\Filesystem::hasFilenameInvalidExtension($file)) {
+			return false;
+		}
+
+		if (\OC\Files\Filesystem::hasFilenameInvalidCharacters($file)) {
+			return false;
+		}
```
come-nc, line 594: Not pretty to have `OC` references in `OCP`.
Could those be non-static and the filesystem get through OCP\Server::get?

susnux: The interface is public but only the implementation uses the private (OC) files.
I do not think it makes sense to have them non-static as you then need to instantiate the class just for functions that do not need a class context, no?

The problem: I want to have only one place for the logic, but we need two (three) version:
1. The public one in Util which should work without any specific Storage
2. The `verifyPath` of IStorage (might be more strict due to storage limitations)
3. `isFileBlacklisted` for internal private API which must not validate the extension as otherwise e.g. chunked uploads fail.

I do not want to introduce new public methods so I moved it to Filesystem.

lib/public/Util.php
susnux: makes sense, thank you!

lib/public/Util.php
come-nc: Using non-static method allows for keeping the implementation in OC while the interface is in OCP, this is what we do for a lot of stuff, not all of them requiring class context (most of them actually do not since we use singletons everywhere).

I can see that Filesystem is not instantiable and uses static methods and properties instead, that’s a bit sad.
For me we should have a long term goal of removing all `OC` references from `OCP` (those have to be added to stubs for applications using psalm and relying on OCP). So not happy about adding new ones without a plan how to remove them on the long term.
Maybe add a new class FileNameValidator with associated interface in OCP?

susnux: Would it make sense to remove then also everything from OCP\Util and move to that new class?

come-nc: Depend what you call everything, but everything related to file name validation, yes.

lib/private/Files/FilenameValidator.php
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * @since 30.0.0
+ */
+class FilenameValidator implements IFilenameValidator {
+
+	/**
+	 * @var list<string>
+	 */
+	private array $forbiddenNames;
```
come-nc, line 21:
```suggestion
	private array $forbiddenNames = [];
```
Either init to empty, or change for nullable and init with `null` and check for `null` instead of `empty()`.

PR #45157, opened 2024-05-02
fix(ShareApiController): Use correct exceptions

nfebe:
Fix incorrect exceptions used in the share api.


susnux:
This is stale and a breaking change for legacy client relying on parsing errors.
We fixed this now with the new unified sharing

PR #45163, opened 2024-05-03
fix(icons): Update files icon

### LINKED ISSUE #732: Enhancement: Icons are different than Nextcloud default icons
Hi,

it looks like the recommendations app is using different icons ( a in the screenshot ) than Nextcloud itself ( b in the screenshot )

diff

Would it be possible to use the standard Nextcloud iconset ( b ) as it would make it more consistent from a design perspective.

provokateurin:
* Resolves: https://github.com/nextcloud/recommendations/issues/732

## Summary

The files icon was not updated accordingly to the new app icon and code relying on the CSS icons were using the old version.

## Checklist


core/img/places/files.svg
```diff
@@ -1 +1 @@
-<svg width="16" height="16" version="1.1" viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg"><path d="m1.5 2c-0.25 0-0.5 0.25-0.5 0.5v11c0 0.26 0.24 0.5 0.5 0.5h13c0.26 0 0.5-0.241 0.5-0.5v-9c0-0.25-0.25-0.5-0.5-0.5h-6.5l-2-2z"/></svg>
+<svg xmlns="http://www.w3.org/2000/svg" id="mdi-folder" viewBox="0 0 24 24"><path d="M10,4H4C2.89,4 2,4.89 2,6V18A2,2 0 0,0 4,20H20A2,2 0 0,0 22,18V8C22,6.89 21.1,6 20,6H12L10,4Z" /></svg>
```
nickvergessen, line 1: Bildschirmfoto vom 2024-05-03 09-06-39

```suggestion
<svg xmlns="http://www.w3.org/2000/svg" id="mdi-folder" width="16" height="16" viewBox="0 0 24 24"><path d="M10,4H4C2.89,4 2,4.89 2,6V18A2,2 0 0,0 4,20H20A2,2 0 0,0 22,18V8C22,6.89 21.1,6 20,6H12L10,4Z" /></svg>
```

Need to add a width+heigh attribute otherwise there was an issue in some browser/client consuming the images.

provokateurin: But then the they should be 24 and not 16 because of the viewBox, right? At least in the dashboard app with the recommendations widget it was no problem with this icon :thinking:

nickvergessen: > But then the they should be 24 and not 16 because of the viewBox, right?

width/heigh is the size of the resulting file (when no size is given),
viewBox is how the coordinates map inside the drawing canvas.
https://webdesign.tutsplus.com/svg-viewport-and-viewbox-for-beginners--cms-30844t is quite a good sample to explain the differente

> At least in the dashboard app with the recommendations widget it was no problem with this icon

I think it was only safari or on iOS or something alike.

marcelklehr (approved):

nickvergessen (approved):

PR #45172, opened 2024-05-03
feat: add option to disable scanner transactions

icewind1991:
I'm not a 100% sure about the full impact of disabling this regarding performance/deadlocks/etc. But having a config flag for this makes it easier to figure out the impact.


artonge (approved):

nfebe (approved):
Add `filescanner_no_transactions` to sample config?

icewind1991:
> Add filescanner_no_transactions to sample config?

Given the uncertainty around the effects I think it's better not to advertise this at the moment. Instead I would rather change the behavior for everyone once we've done more testing around this.

icewind1991:
stable27: https://github.com/nextcloud/server/pull/45681

PR #45207, opened 2024-05-06
fix(dav): Try basic auth for ajax WebDAV requests

susnux:
## Summary

Allow using basic auth with ajax requests on WebDAV.
The idea why the code was added in the first place was to ensure that if basic auth fails the user gets not asked by the browser for the correct credentials.
We still keep this behavior by throwing a 401 error in case of failed basic auth.

## Checklist

### DIFF
--- a/apps/dav/tests/unit/Connector/Sabre/AuthTest.php
+++ b/apps/dav/tests/unit/Connector/Sabre/AuthTest.php
@@ -13,6 +13,7 @@ use OCP\IRequest;
 use OCP\ISession;
 use OCP\IUser;
 use OCP\Security\Bruteforce\IThrottler;
+use PHPUnit\Framework\MockObject\MockObject;
 use Sabre\DAV\Server;
 use Sabre\HTTP\RequestInterface;
 use Sabre\HTTP\ResponseInterface;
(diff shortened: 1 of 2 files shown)


susnux:
The problem gets visible because previous versions of `@nextcloud/axios` were not setting the header, but the latest version will set the `X-Requested-With` header by default.

Meaning the alternative to this is to drop the `X-Requested-With: XMLHttpRequest` from axios requests to DAV like we were doing with current versions (currently we do not set `X-Requested-With` for ajax requests to DAV to prevent this.
So I think this solution here is cleaner than faking the kind of request :sweat_smile:

apps/dav/tests/unit/Connector/Sabre/AuthTest.php
```diff
@@ -47,17 +48,17 @@
  * @group DB
  */
 class AuthTest extends TestCase {
-	/** @var ISession */
+	/** @var ISession|MockObject */
```
come-nc, line 51:
```suggestion
	/** @var ISession&MockObject */
```

come-nc (approved):

apps/dav/tests/unit/Connector/Sabre/AuthTest.php
come-nc: (and all the others)

nickvergessen (dismissed):
Blocking temporarily

icewind1991 (approved):

PR #45227, opened 2024-05-07
fix: Remove reminder action from trash

Pytal:
## Summary

- Fix invalid set reminder action displayed on trash files
- Deleted files should not have reminders and any pending reminders on files that get deleted are automatically removed

## Checklist


artonge (commented):
Should we also remove reminders from the DB when a file gets deleted?

Pytal:
> Should we also remove reminders from the DB when a file gets deleted?

Works this way already actually https://github.com/nextcloud/server/blob/master/apps/files_reminders/lib/Listener/NodeDeletedListener.php

susnux (approved):

skjnldsv (approved):

PR #45235, opened 2024-05-08
fix(caldav): automatically delete outdated scheduling objects

### LINKED ISSUE #43621: Remove old scheduling objects from INBOX and `oc_schedulingobjects` via cron
When a scheduling event is in the past, the scheduling event should be removed from oc_schedulingobjects and the associated ICS in the scheduling inbox should be deleted: https://github.com/sabre-io/dav/issues/1138

Any FREEBUSY request  checks the schedulingobjects / a users inbox). The search will be limited by a start and end date. Since there is no dedicated timestamp on each entry of the table, but instead the query is done on a text column, this is where I would expect the largest performance impact.

miaulalala:
* Resolves:  https://github.com/nextcloud/server/issues/43621

## Summary

oc_schedulingobjects currently grows without ever deleting outdated objects. This PR a repair step that is declared as expensive so admins can decide to run them at their convenience for the initial delete.

The delete is chunked to 50k rows on each transaction so the database isn't locked for a long time (especially in clustered setups this could cause issues). MySQL needs special treatment as it doesn't support LIMITs on DELETE queries, so it does a SELECT on the ids to delete, and then runs the delete on those.

After the repair step has run, a regular cron job is added to the Jobs List that runs every hour to get rid of scheduling objects that are older than an hour. We don't really need them and could theoretically delete them as soon as they're processed by the `ITip\Broker` but as rooms and resources are also run in a cron job, keeping them until the principal room and resources are added is probably a good idea as I can't exclude unwanted side effects. I also updated the runtime for rooms and resources to run every half hour for that reason.

## Checklist


miaulalala (commented):
Move the filtering logic to the CalDavBackend and do the timestamp check in a query. Introduce a new method for this.

miaulalala:
Do a slow repair job for this so we don't kill the instance.

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
```diff
@@ -0,0 +1,71 @@
...
+			$this->logger->info("$userId doesn't exist, removing job");
+			$this->jobList->remove(self::class, $argument);
+			return;
+		}
+
+		$principal = 'principals/users/'.$userId;
+		$children = $this->calDavBackend->getSchedulingObjects($principal);
+		$count = 0;
+		foreach($children as $object) {
+			if ($object['lastmodified'] < ($this->time->getTime() - 24 * 60 * 60)) {
```
tcitworld, line 63: An invite can be sent for an event far into the future…shouldn't we look at the actual datetime of the event occurence instead?

miaulalala: I don't think that's necessary - the inboxed scheduling events will have already been processed (`\Sabre\CalDAV\Schedule\Plugin::scheduleLocalDelivery` - the `iTip\Broker` processes the scheduling object both as  a calendar change and writes a copy of the iTip message to the inbox. So any changes from scheduling messages are written to oc_calendar_objects, regardless of the time the event happens at or how far into the future a recurrence happens, unless some preconditions aren't met.

I think this is only interesting i we support `COUNTER`, `DECLINECOUNTER` or `REFRESH` which isn't implemented AFAIK. Additionally, the row is a copy of the iTip message that is kept no matter if the `SEQUENCE` number is the right one to process a `REPLY`.

I checked for usages of the inbox in the 3rdparty code and ours, it's only used for the `VAVILABILITY` and to check the ACLs for scheduling, but feel free to look into this further, I might have missed something.

miaulalala:
https://github.com/nextcloud/server/blob/b0bfe3e2cd7b6ef393e74000ba083cd57145df2c/lib/private/Repair.php#L223-L228 for the expensive repair job

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -2738,6 +2739,71 @@ public function deleteSchedulingObject($principalUri, $objectUri) {
...
+			return;
+		}
+
+		$queryResults = $this->atomic(function () use ($modifiedBefore, $limit) {
+			$query = $this->db->getQueryBuilder();
+			$query->delete('schedulingobjects')
+				->where($query->expr()->lte('lastmodified', $query->createNamedParameter($modifiedBefore)))
+				->setMaxResults($limit);
+			return $query->executeStatement();
+		}, $this->db);
```
ChristophWurst, line 2765: explicit transaction is superfluous for a single query

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -2738,6 +2739,71 @@ public function deleteSchedulingObject($principalUri, $objectUri) {
...
+			$count = $result->rowCount();
+			if($count === 0) {
+				return;
+			}
+			$ids = array_map(static function (array $id) {
+				return (int)$id[0];
+			}, $result->fetchAll(\PDO::FETCH_NUM));
+			$result->closeCursor();
+
+			$queryResult = 0;
```
ChristophWurst, line 2790:
```suggestion
			$numDeleted = 0;
```

naming

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -2738,6 +2739,71 @@ public function deleteSchedulingObject($principalUri, $objectUri) {
...
+		}
+	}
+
+	private function deleteOutdatedSchedulingObjectsForMySQL(int $modifiedBefore, int $limit): void {
+		$queryResult = $this->atomic(function () use ($modifiedBefore, $limit) {
+			$query = $this->db->getQueryBuilder();
+			$query->select('id')
+				->from('schedulingobjects')
+				->where($query->expr()->lte('lastmodified', $query->createNamedParameter($modifiedBefore)))
+				->setMaxResults($limit);
```
ChristophWurst, line 2779: if you order by last modified, the values of the chunks will be closer together and make it more efficient for the database to traverse the index.

ChristophWurst: there is no index for this query. it will perform a full table scan:

```
> EXPLAIN SELECT id FROM oc_schedulingobjects WHERE lastmodified <= 1700000000;
+------+-------------+----------------------+------+---------------+------+---------+------+------+-------------+
| id   | select_type | table                | type | possible_keys | key  | key_len | ref  | rows | Extra       |
+------+-------------+----------------------+------+---------------+------+---------+------+------+-------------+
|    1 | SIMPLE      | oc_schedulingobjects | ALL  | NULL          | NULL | NULL    | NULL | 104  | Using where |
+------+-------------+----------------------+------+---------------+------+---------+------+------+-------------+
```

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
```diff
@@ -0,0 +1,52 @@
...
+
+class DeleteOutdatedSchedulingObjects extends TimedJob {
+	public function __construct(
+		private CalDavBackend $calDavBackend,
+		private LoggerInterface $logger,
+		ITimeFactory $timeFactory,
+	) {
+		parent::__construct($timeFactory);
+		$this->setInterval(60 * 60);
+		$this->setTimeSensitivity(self::TIME_SENSITIVE);
```
ChristophWurst, line 41: why is this time sensitive? wouldn't it be sufficient to clean up once a day in the maintenance window?

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	/**
+	 * @param array $argument
+	 */
+	protected function run($argument): void {
+		$time = $this->time->getTime() - (60 * 60);
```
ChristophWurst, line 48: caveat with timing this once per hour while the resources job runs twice: it assumes that background jobs run reliably. this might not be the case for ajax cron.

ChristophWurst (commented):
Let's get rid of the hourly full table scan

:+1: otherwise

apps/dav/lib/CalDAV/CalDavBackend.php
miaulalala: yeah I need to move this further up so each 1k delete doesn't do the full table scan. Will an index solve this? I don't think it will, seeing as there is a blob associated with this query. When I discussed it with Joas, this was the intention behind the transaction (even though I applied it at the wrong code point :wink: )

apps/dav/lib/CalDAV/CalDavBackend.php
miaulalala: Add an index on the lastmodified?

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
miaulalala: Well it entirely depends on the size of the instance. For 100 users, no problem. For 1k or 10k, a lot of data could be produced in a short amount of time, especially if there's a lot of usage on the calendar. That's why I'd rather clean up once an hour during business hours than have this run outside of it with lots of rows.

apps/dav/lib/CalDAV/CalDavBackend.php
ChristophWurst: Yes

apps/dav/lib/CalDAV/CalDavBackend.php
ChristophWurst: I don't see how an explicit transaction is different to our AUTOCOMMIT=1 sessions

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
miaulalala: The resources job running every 20 minutes is just a precaution, I don't remember my exact thought pattern as to why, but I can revert. It was something along the lines of not deleting scheduling objects that might not have been processed yet but the more I think about it the less it makes sense, as I also delete scheduling objects 1 hour in the past only.

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
ChristophWurst: Right now we have no cleanup at all and most instances are fine. So I think 99.99% of instances will also survive if data is only cleaned up daily :)

A large range DELETE query can cause table locks that block other operations. We have seen this with https://github.com/nextcloud/server/issues/27695.

ChristophWurst: Relevant too: https://github.com/nextcloud/server/pull/43605

apps/dav/lib/Migration/Version1031Date20240523183435.php
```diff
@@ -0,0 +1,51 @@
...
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$messagesTable = $schema->getTable('schedulingobjects');
+		$messagesTable->addIndex(['lastmodified'], 'scheduling_lastmodified_idx');
```
ChristophWurst, line 47: this will be slow on large installations and therefore block the upgrade.

register the index as missing in ``\OC\Core\Application`` (e.g. like calendarobject_calid_index). these will be applied async to the upgrade, at a point in time that the admins pick.

for new setups it would be best if the index exists right away (the table is empty so adding an index is cheap). find the migration that created the oc_schedulingobjects table and add the index there.

-> new installations will have the index out of the box
-> existing installations will receive the index after the upgrade

miaulalala: smort

kesselb (approved):

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -2738,6 +2738,44 @@ public function deleteSchedulingObject($principalUri, $objectUri) {
...
+		$result->closeCursor();
+
+		$numDeleted = 0;
+		$deleteQuery = $this->db->getQueryBuilder();
+		$deleteQuery->delete('schedulingobjects')
+			->where($deleteQuery->expr()->in('id', $deleteQuery->createParameter('ids'), IQueryBuilder::PARAM_INT_ARRAY));
+		foreach(array_chunk($ids, 1000) as $chunk) {
+			$deleteQuery->setParameter('ids', $chunk, IQueryBuilder::PARAM_INT_ARRAY);
+			$numDeleted += $deleteQuery->executeStatement();
+		}
```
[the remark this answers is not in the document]
ChristophWurst: We have to keep the transaction size reasonable and the table has a BLOB column. I would leave this outside of a longer running transaction.

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
```diff
@@ -0,0 +1,52 @@
...
+use Psr\Log\LoggerInterface;
+
+class DeleteOutdatedSchedulingObjects extends TimedJob {
+	public function __construct(
+		private CalDavBackend $calDavBackend,
+		private LoggerInterface $logger,
+		ITimeFactory $timeFactory,
+	) {
+		parent::__construct($timeFactory);
+		$this->setInterval(23 * 60 * 60);
```
ChristophWurst, line 40: Why not 24? :laughing:

ChristophWurst (approved):
Code looks good

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
miaulalala: @nickvergessen can explain :wink:

apps/dav/lib/CalDAV/CalDavBackend.php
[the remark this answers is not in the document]
miaulalala: See above comments from Christoph - having 50k rows in a transaction is not a good thing apparently.

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
kesselb, line 25:
```suggestion
/**
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```

apps/dav/lib/Migration/DeleteSchedulingObjects.php
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
kesselb, line 25:
```suggestion
/**
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```

apps/settings/lib/SetupChecks/SchedulingTableSize.php
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
```
kesselb, line 25:
```suggestion
/**
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```

apps/dav/lib/BackgroundJob/DeleteOutdatedSchedulingObjects.php
nickvergessen: it would move back all the time, as the run time is not 0 seconds and it can be started with up to 14 minutes delay.
That would mean it would fall further back and back in the "maintenance time window" and eventually skip some day as current time + 24h is outside of maintenance window.
With 23h it will just be saver to run every day.

PR #45237, opened 2024-05-08
perf(deleteAction): Queue delete requests

### LINKED ISSUE #45234: [Bug]: Performance degradation due to multiple simultaneous deletion requests from trashbin

### Bug description

An issue has been identified regarding the simultaneous deletion of multiple files from the trashbin in Nextcloud versions stable27, stable28, stable29, and master. When attempting to delete these files concurrently, a performance degradation occurs due to the simultaneous sending of multiple identical requests to the same endpoint.

### Steps to reproduce

1. Access Nextcloud and navigate to the trashbin.
2. Select multiple files for deletion.
3. Initiate simultaneous deletion of all selected files.

### Expected behavior

We should probably que deletion (and or "empty trash") request when there are too many to handle at once.

### Installation method

None

### Nextcloud Server version

27

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

nfebe:
When multiple files are deleted at once, all the requests bombard the server simultaneously, causing performance issues.

This commit adds queuing that limits the concurrency of these requests to 5 at a time.

--------------------

## Without Queuing

not-queued-requests

## With Queuing

queued-requests

Resolves : https://github.com/nextcloud/server/issues/45234


apps/files/src/actions/deleteAction.ts
```diff
@@ -183,7 +186,19 @@ export const action = new FileAction({
...
+		const promises = nodes.map(node => {
+		    // Create a promise that resolves with the result of exec(node)
+		    const promise = new Promise<boolean>(resolve => {
+				queue.add(async () => {
+					const result = await this.exec(node, view, dir)
+					resolve(result !== null ? result : false)
+				})
+			})
+			return promise
+		})
```
Pytal, line 199:
```suggestion
		const promises = nodes.map(node => queue.add(() => this.exec(node, view, dir)))
```
This should work the same, didn't test

Also null is valid as a file action return type

Fenn-CS: That was the first thing I tried. It didn't work.

The issue with that code snippet is related to the usage of `queue.add` within `Promise.all`

`queue.add` method returns a promise that resolves when the queued function is executed, but it doesn't return the actual result of the function. Therefore, using `queue.add` directly inside `Promise.all` will not work as intended because `Promise.all` expects an array of promises that resolve with the result of each executed function.

I didn't run it at all, but before I could have the chance to run it my code editor was complaining already.

Fenn-CS: Update : Even though my IDE type/linting config does not seem to like it, it works, so implemented. Thanks.

Pytal (approved):

joshtrichards (commented):
Certainly an improvement, but since this is all browser side aren't we creating a new problem: i.e. if the browser is closed before the entire delete session finishes?

Maybe we go farther and add the full item list to a new background job handled server-side? We already expire the trash contents in a background job so...

Perhaps more appropriate in a future follow-up (if the browser being closed aborting the delete queue is considered acceptable for now).

susnux:
> Maybe we go farther and add the full item list to a new background job handled server-side? We already expire the trash contents in a background job so...

I do not know any WebDAV standard allowing bulk delete so this would need a custom API so definitely a thing we need to think through.

susnux (approved):

solracsf:
> Certainly an improvement, but since this is all browser side aren't we creating a new problem: i.e. if the browser is closed before the entire delete session finishes?

What about adding a word on the already exitsting message like "Bulk deletinon in progress, please keep this window open in your browser..." ? Could this be a thing?

Fenn-CS:
@solracsf I tracked your concerns here : https://github.com/nextcloud/server/issues/45281 I think we can manage fix it from there.

artonge (approved):

apps/files/src/actions/deleteAction.ts
Fenn-CS: Hey @Pytal CI does not like this too : https://github.com/nextcloud/server/actions/runs/9076246755/job/24938632270?pr=45237

I guess I have to revert.

susnux: Maybe:
`const promises = nodes.map(node => queue.add(() => this.exec(node, view, dir)) ?? null)`

Fenn-CS: Still leading to type issues.

apps/files/src/actions/deleteAction.ts
```diff
@@ -119,6 +120,8 @@ const displayName = (nodes: Node[], view: View) => {
 	return t('files', 'Delete')
 }
 
+const queue = new PQueue({ concurrency: 1 })
```
SystemKeeper, line 123: @Fenn-CS According to your comment in the PR we do 5 requests at the same time. But here concurrency is set to 1? Am I missing something?

Fenn-CS: Hello @SystemKeeper you are right and thank you for catching that, see fix : https://github.com/nextcloud/server/pull/45902

PR #45241, opened 2024-05-09
fix(setupChecks): update db version checks

joshtrichards:
## Summary

- Update lowest supported version to match existing docs
- Adds check for highest supported version that matches existing docs
- Remove references to Nextcloud Server v21
- Clarifies the language a bit
- Adjust CI tests to test the PostgreSQL versions in the docs (drops 10, replaces 15 with 16)

## TODO

- If backported, each one will have to be slightly modified. This is spec'd for v30/master as it stands today.

Follow-up ideas (i.e. out of scope) and other related notes:

- Docs adjustment for MySQL 8.x: nextcloud/documentation#11801
- Add MySQL 8.4 to our test matrix (it's LTS!)
- Consider eliminating non-LTS releases from being referenced in our docs (but we can still pass them in this setup check I guess... as long as they're within range)

## Checklist

### DIFF
--- a/apps/settings/lib/SetupChecks/SupportedDatabase.php
+++ b/apps/settings/lib/SetupChecks/SupportedDatabase.php
@@ -62,7 +62,9 @@ class SupportedDatabase implements ISetupCheck {
 			$row = $result->fetch();
 			$version = $row['Value'];
 			$versionlc = strtolower($version);
-
+			// we only care about X.Y not X.Y.Z differences
+			[$major, $minor, ] = explode('.', $versionlc);
+			$versionlc = $major . '.' . $minor;
 			if (str_contains($versionlc, 'mariadb')) {
 				if (version_compare($versionlc, '10.3', '<') || version_compare($versionlc, '10.11', '>')) {
 					return SetupResult::warning($this->l10n->t('MariaDB version "%s" detected. MariaDB >=10.3 and <=10.11 is suggested for best performance, stability and functionality with this version of Nextcloud.', $version));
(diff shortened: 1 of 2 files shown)


susnux (commented):
Adjusting the minimum versions is very good! But I do not think we should add the upper version here.

joshtrichards:
> But I do not think we should add the upper version here.

Any particular reason?

nickvergessen (approved):

susnux (approved):

susnux:
Makes sense because of the wording ("suggestion").

susnux (approved):

apps/settings/lib/SetupChecks/SupportedDatabase.php
```diff
@@ -62,23 +62,29 @@ public function run(): SetupResult {
 			$row = $result->fetch();
 			$version = $row['Value'];
 			$versionlc = strtolower($version);
-
+			// we only care about X.Y not X.Y.Z differences
+			[$major, $minor, ] = explode('.', $versionlc);
+			$versionConcern = $major . '.' . $minor;
```
come-nc, line 67: `version_compare` is smart enough that you do not need to do this I think.

joshtrichards: It wasn't when I tried it. It compares the full version string provided. Also I already had to be more specific for the PostgreSQL check anyway (since we only care about major and not even minor for that one).

come-nc: https://3v4l.org/T1lkh It works also if you do not care about minor.

joshtrichards: It's mostly for the <= check because I wanted the code to literally state the min/max "versions" (release branches) supported just as they're written in the documentation.

Comparing against N+1 feels less self documenting. ;-)

PR #34178, opened 2022-09-21
Rename users to account or person

PVince81:
Replace translated texts from "user" to either "account" or "people" depending on context

Resolves #32720

PR #45245, opened 2024-05-09
feat: Split display name and username columns

Pytal:
## Summary

- Split the columns in preparation for making the table sortable and adding header sort buttons

Before | After
--- | ---
image | image

## Checklist


susnux (approved):
Makes sense for me, lets wait for designers.

joshtrichards:
Might be some intersection with #34178 (to avoid future code debt) though I didn't really follow where that stands since it's merged and the menu entry to get to *Accounts* is still labeled as *Users* so dunno.

In any case, seems header *Username* -> *Login* I think?

Pytal:
> Might be some intersection with #34178 (to avoid future code debt) though I didn't really follow where that stands since it's merged and the menu entry to get to _Accounts_ is still labeled as _Users_ so dunno.

I think this was intentional @skjnldsv ?

> In any case, seems header _Username_ -> _Login_ I think?

On the login page it is also "username" so I will keep it the same

jancborchardt (approved):
Seems good!

Agree with @joshtrichards that Username should be "Account name" (same on the login screen). But it’s something for a separate pull request. :)

Pytal:
> Agree with @joshtrichards that Username should be "Account name" (same on the login screen). But it’s something for a separate pull request. :)

Sure, can do in a followup :)

> Also question @Pytal – isn’t the sorting we add by "Last login" only? (And display name which is the default.) Or are we just going to make everything sortable?

Will just be displayname, userId, and last login for now as that is what has been implemented on the backend, not sure about others as it may be complicated

PR #45249, opened 2024-05-10
feat: add ability to sort by last login

yemkareems:
* Resolves: # 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -267,7 +267,6 @@ class UsersController extends AUserData {
 
 	/**
 	 * @NoAdminRequired
-	 * @NoCSRFRequired
 	 *
 	 * Get the list of last logged-in users and their details
 	 *
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -94,7 +94,7 @@ class UsersController extends AUserData {
 	 *
 	 * 200: Users returned
 	 */
-	public function getUsers(string $search = '', ?int $limit = null, int $offset = 0, string $sortMode = 'uid', string $sortOrder = 'asc'): DataResponse {
+	public function getUsers(string $search = '', ?int $limit = null, int $offset = 0): DataResponse {
 		$user = $this->userSession->getUser();
 		$users = [];
 
@@ -102,7 +102,7 @@ class UsersController extends AUserData {
 		$uid = $user->getUID();
 		$subAdminManager = $this->groupManager->getSubAdmin();
 		if ($this->groupManager->isAdmin($uid)) {
-			$users = $this->userManager->search($search, $limit, $offset, $sortMode, $sortOrder);
+			$users = $this->userManager->search($search, $limit, $offset);
 		} elseif ($subAdminManager->isSubAdmin($user)) {
 			$subAdminOfGroups = $subAdminManager->getSubAdminsGroups($user);
 			foreach ($subAdminOfGroups as $key => $group) {
@@ -131,13 +131,11 @@ class UsersController extends AUserData {
 	 * @param string $search Text to search for
 	 * @param int|null $limit Limit the amount of groups returned
 	 * @param int $offset Offset for searching for groups
-	 * @param string $sortMode Field to order the results with
-	 * @param string $sortOrder asc or desc
 	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, Provisioning_APIUserDetails|array{id: string}>}, array{}>
 	 *
 	 * 200: Users details returned
 	 */
-	public function getUsersDetails(string $search = '', ?int $limit = null, int $offset = 0, string $sortMode = 'uid', string $sortOrder = 'asc'): DataResponse {
+	public function getUsersDetails(string $search = '', ?int $limit = null, int $offset = 0): DataResponse {
 		$currentUser = $this->userSession->getUser();
 		$users = [];
 
@@ -145,7 +143,7 @@ class UsersController extends AUserData {
 		$uid = $currentUser->getUID();
 		$subAdminManager = $this->groupManager->getSubAdmin();
 		if ($this->groupManager->isAdmin($uid)) {
-			$users = $this->userManager->search($search, $limit, $offset, $sortMode, $sortOrder);
+			$users = $this->userManager->search($search, $limit, $offset);
 			$users = array_keys($users);
 		} elseif ($subAdminManager->isSubAdmin($currentUser)) {
 			$subAdminOfGroups = $subAdminManager->getSubAdminsGroups($currentUser);
@@ -155,7 +153,7 @@ class UsersController extends AUserData {
 
 			$users = [];
 			foreach ($subAdminOfGroups as $group) {
-				$users[] = array_keys($this->groupManager->displayNamesInGroup($group, $search, $limit, $offset, $sortMode, $sortOrder));
+				$users[] = array_keys($this->groupManager->displayNamesInGroup($group, $search, $limit, $offset));
 			}
 			$users = array_merge(...$users);
 		}
@@ -266,6 +264,96 @@ class UsersController extends AUserData {
 			'users' => $usersDetails
 		]);
 	}
+
+	/**
+	 * @NoAdminRequired
+	 * @NoCSRFRequired
+	 *
+	 * Get the list of disabled users and their details
+	 *
+	 * @param string $search Text to search for
+	 * @param ?int $limit Limit the amount of users returned
+	 * @param int $offset Offset
+	 * @param string $sortMode Field to order the results with
+	 * @param string $sortOrder asc or desc
+	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, Provisioning_APIUserDetails|array{id: string}>}, array{}>
+	 *
+	 * 200: Users details returned based on last logged in information
+	 */
+	public function getLastLoggedInUsers(string $search = '',
+										 ?int   $limit = null,
+										 int    $offset = 0,
+										 string $sortMode = 'lastLogin',
+										 string $sortOrder = 'desc'
+	): DataResponse {
+		$currentUser = $this->userSession->getUser();
+		if ($currentUser === null) {
+			return new DataResponse(['users' => []]);
+		}
+		if ($limit !== null && $limit < 0) {
+			throw new InvalidArgumentException("Invalid limit value: $limit");
+		}
+		if ($offset < 0) {
+			throw new InvalidArgumentException("Invalid offset value: $offset");
+		}
+
+		$users = [];
+
+		// Admin? Or SubAdmin?
+		$uid = $currentUser->getUID();
+		$subAdminManager = $this->groupManager->getSubAdmin();
+		if ($this->groupManager->isAdmin($uid)) {
+			$users = $this->userManager->getUsersSortedByLastLogin($limit, $offset, $search, $sortMode, $sortOrder);
+			$users = array_map(fn (IUser $user): string => $user->getUID(), $users);
+		} elseif ($subAdminManager->isSubAdmin($currentUser)) {
+			$subAdminOfGroups = $subAdminManager->getSubAdminsGroups($currentUser);
+
+			$users = [];
+			/* We have to handle offset ourselve for correctness */
+			$tempLimit = ($limit === null ? null : $limit + $offset);
+			foreach ($subAdminOfGroups as $group) {
+				$users = array_merge(
+					$users,
+					array_map(
+						fn (IUser $user): string => $user->getUID(),
+						array_filter(
+							$group->searchUsers($search, ($tempLimit === null ? null : $tempLimit - count($users))),
+							fn (IUser $user): bool => !$user->isEnabled()
+						)
+					)
+				);
+				if (($tempLimit !== null) && (count($users) >= $tempLimit)) {
+					break;
+				}
+			}
+			$users = array_slice($users, $offset);
+		}
+
+		$usersDetails = [];
+		foreach ($users as $userId) {
+			try {
+				$userData = $this->getUserData($userId);
+			} catch (OCSNotFoundException $e) {
+				// We still want to return all other accounts, but this one was removed from the backends
+				// yet they are still in our database. Might be a LDAP remnant.
+				$userData = null;
+				$this->logger->warning('Found one disabled account that was removed from its backend, but still exists in Nextcloud database', ['accountId' => $userId]);
+			}
+			// Do not insert empty entry
+			if ($userData !== null) {
+				$usersDetails[$userId] = $userData;
+			} else {
+				// Currently logged in user does not have permissions to see this user
+				// only showing its id
+				$usersDetails[$userId] = ['id' => $userId];
+			}
+		}
+
+		return new DataResponse([
+			'users' => $usersDetails
+		]);
+	}
+
 
 
 	/**
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -293,7 +293,7 @@ class UsersController extends AUserData {
 		$users = [];
 
 		// For Admin alone user sorting based on lastLogin. For sub admin and groups this is not supported
-		$users = $this->config->getLastLoggedInUsers($limit, $offset, $search);
+		$users = $this->userManager->getLastLoggedInUsers($limit, $offset, $search);
 
 		$usersDetails = [];
 		foreach ($users as $userId) {
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -267,14 +267,13 @@ class UsersController extends AUserData {
 
 	/**
 	 * @NoAdminRequired
+	 * @NoCSRFRequired
 	 *
-	 * Get the list of disabled users and their details
+	 * Get the list of last logged-in users and their details
 	 *
 	 * @param string $search Text to search for
 	 * @param ?int $limit Limit the amount of users returned
 	 * @param int $offset Offset
-	 * @param string $sortMode Field to order the results with
-	 * @param string $sortOrder asc or desc
 	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, Provisioning_APIUserDetails|array{id: string}>}, array{}>
 	 *
 	 * 200: Users details returned based on last logged in information
@@ -282,8 +281,6 @@ class UsersController extends AUserData {
 	public function getLastLoggedInUsers(string $search = '',
 		?int   $limit = null,
 		int    $offset = 0,
-		string $sortMode = 'lastLogin',
-		string $sortOrder = 'desc'
 	): DataResponse {
 		$currentUser = $this->userSession->getUser();
 		if ($currentUser === null) {
@@ -302,7 +299,7 @@ class UsersController extends AUserData {
 		$uid = $currentUser->getUID();
 		$subAdminManager = $this->groupManager->getSubAdmin();
 		if ($this->groupManager->isAdmin($uid)) {
-			$users = $this->userManager->getUsersSortedByLastLogin($limit, $offset, $search, $sortMode, $sortOrder);
+			$users = $this->userManager->getUsersSortedByLastLogin($limit, $offset, $search);
 			$users = array_map(fn (IUser $user): string => $user->getUID(), $users);
 		} elseif ($subAdminManager->isSubAdmin($currentUser)) {
 			$subAdminOfGroups = $subAdminManager->getSubAdminsGroups($currentUser);
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -94,7 +94,7 @@ class UsersController extends AUserData {
 	 *
 	 * 200: Users returned
 	 */
-	public function getUsers(string $search = '', ?int $limit = null, int $offset = 0, string $orderBy = 'lastLogin', string $sort = 'DESC'): DataResponse {
+	public function getUsers(string $search = '', ?int $limit = null, int $offset = 0, string $sortMode = 'uid', string $sortOrder = 'asc'): DataResponse {
 		$user = $this->userSession->getUser();
 		$users = [];
 
@@ -102,7 +102,7 @@ class UsersController extends AUserData {
 		$uid = $user->getUID();
 		$subAdminManager = $this->groupManager->getSubAdmin();
 		if ($this->groupManager->isAdmin($uid)) {
-			$users = $this->userManager->search($search, $limit, $offset, $orderBy, $sort);
+			$users = $this->userManager->search($search, $limit, $offset, $sortMode, $sortOrder);
 		} elseif ($subAdminManager->isSubAdmin($user)) {
 			$subAdminOfGroups = $subAdminManager->getSubAdminsGroups($user);
 			foreach ($subAdminOfGroups as $key => $group) {
@@ -131,13 +131,13 @@ class UsersController extends AUserData {
 	 * @param string $search Text to search for
 	 * @param int|null $limit Limit the amount of groups returned
 	 * @param int $offset Offset for searching for groups
-	 * @param string $orderBy Field to order the results with
-	 * @param string $sort ASC or DESC
+	 * @param string $sortMode Field to order the results with
+	 * @param string $sortOrder asc or desc
 	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, Provisioning_APIUserDetails|array{id: string}>}, array{}>
 	 *
 	 * 200: Users details returned
 	 */
-	public function getUsersDetails(string $search = '', ?int $limit = null, int $offset = 0, string $orderBy = 'lastLogin', string $sort = 'DESC'): DataResponse {
+	public function getUsersDetails(string $search = '', ?int $limit = null, int $offset = 0, string $sortMode = 'uid', string $sortOrder = 'asc'): DataResponse {
 		$currentUser = $this->userSession->getUser();
 		$users = [];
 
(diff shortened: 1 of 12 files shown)


apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -130,15 +130,15 @@ public function __construct(
 	 *
 	 * 200: Users returned
 	 */
-	public function getUsers(string $search = '', ?int $limit = null, int $offset = 0): DataResponse {
+	public function getUsers(string $search = '', ?int $limit = null, int $offset = 0, string $orderBy = 'lastLogin', string $sort = 'DESC'): DataResponse {
```
Pytal, line 133:
```suggestion
	public function getUsers(string $search = '', ?int $limit = null, int $offset = 0, string $orderBy = 'uid', string $sort = 'ASC'): DataResponse {
```
We should not break the existing sort order

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -167,19 +167,21 @@ public function getUsers(string $search = '', ?int $limit = null, int $offset =
 	 * @param string $search Text to search for
 	 * @param int|null $limit Limit the amount of groups returned
 	 * @param int $offset Offset for searching for groups
+	 * @param string $orderBy Field to order the results with
+	 * @param string $sort ASC or DESC
```
Pytal, line 171:
```suggestion
	 * @param string $sort asc or desc
```
Would use lowercase like our other API endpoints

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -167,19 +167,21 @@ public function getUsers(string $search = '', ?int $limit = null, int $offset =
...
 	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, Provisioning_APIUserDetails|array{id: string}>}, array{}>
 	 *
 	 * 200: Users details returned
 	 */
-	public function getUsersDetails(string $search = '', ?int $limit = null, int $offset = 0): DataResponse {
+	public function getUsersDetails(string $search = '', ?int $limit = null, int $offset = 0, string $orderBy = 'lastLogin', string $sort = 'DESC'): DataResponse {
```
Pytal, line 176:
```suggestion
	public function getUsersDetails(string $search = '', ?int $limit = null, int $offset = 0, string $orderBy = 'uid', string $sort = 'asc'): DataResponse {
```
Same to not break the sort order + lowercase sort param

lib/private/User/Database.php
```diff
@@ -264,27 +264,36 @@ public function getDisplayName($uid): string {
...
+			$configKey = 'lastLogin';
+		}
+
 		$query->select('uid', 'displayname')
 			->from($this->table, 'u')
 			->leftJoin('u', 'preferences', 'p', $query->expr()->andX(
 				$query->expr()->eq('userid', 'uid'),
-				$query->expr()->eq('appid', $query->expr()->literal('settings')),
-				$query->expr()->eq('configkey', $query->expr()->literal('email')))
+				$query->expr()->eq('appid', $query->expr()->literal($appId)),
```
kesselb, line 289: This change breaks searching by email.

1) Open user management
2) Click on the search icon on the upper right next to your profile picture
3) Enter an email address of a user

=> No result on this branch

The left join is there to look up a user by the email address. 
A subselect for lastLogin (similar to the preferences join) should work.

kesselb:
```
Index: lib/private/User/Database.php
IDEA additional info:
<+>UTF-8
diff --git a/lib/private/User/Database.php b/lib/private/User/Database.php
--- a/lib/private/User/Database.php	(revision 1b86b6c7f67d1fbc2936a76289e93df2a6e63445)
+++ b/lib/private/User/Database.php	(date 1715447859568)
@@ -275,25 +275,32 @@
 
 		$query = $this->dbConn->getQueryBuilder();
 
-		$appId = 'settings';
-		$configKey = 'email';
-		if($orderBy == 'lastLogin') {
-			$appId = 'login';
-			$configKey = 'lastLogin';
+		if ($orderBy === 'lastLogin') {
+			$lastLoginSubSelect = $this->dbConn->getQueryBuilder();
+			$lastLoginSubSelect->select('configvalue')
+				->from('preferences', 'p2')
+				->where($lastLoginSubSelect->expr()->andX(
+					$lastLoginSubSelect->expr()->eq('p2.userid', 'uid'),
+					$lastLoginSubSelect->expr()->eq('p2.appid', $lastLoginSubSelect->expr()->literal('login')),
+					$lastLoginSubSelect->expr()->eq('p2.configkey', $lastLoginSubSelect->expr()->literal('lastLogin')),
+				));
+			$orderByExpression = $query->createFunction('(' . $lastLoginSubSelect->getSQL() .')');
+		} else {
+			$orderByExpression = $query->func()->lower('displayname');
 		}
 
-		$query->select('uid', 'displayname')
+		$query->select('uid', 'displayname', $orderByExpression)
 			->from($this->table, 'u')
 			->leftJoin('u', 'preferences', 'p', $query->expr()->andX(
 				$query->expr()->eq('userid', 'uid'),
-				$query->expr()->eq('appid', $query->expr()->literal($appId)),
-				$query->expr()->eq('configkey', $query->expr()->literal($configKey)))
+				$query->expr()->eq('appid', $query->expr()->literal('settings')),
+				$query->expr()->eq('configkey', $query->expr()->literal('email')))
 			)
 			// sqlite doesn't like re-using a single named parameter here
 			->where($query->expr()->iLike('uid', $query->createPositionalParameter('%' . $this->dbConn->escapeLikeParameter($search) . '%')))
 			->orWhere($query->expr()->iLike('displayname', $query->createPositionalParameter('%' . $this->dbConn->escapeLikeParameter($search) . '%')))
 			->orWhere($query->expr()->iLike('configvalue', $query->createPositionalParameter('%' . $this->dbConn->escapeLikeParameter($search) . '%')))
-			->orderBy($query->func()->lower('configvalue'), $sort)
+			->orderBy($orderByExpression, $sort)
 			->addOrderBy('uid_lower', 'ASC')
 			->setMaxResults($limit)
 			->setFirstResult($offset);

```

kesselb: If getDisplayNames is called from getUsers (the route the user management xhr request is using) the list is sorted again in: https://github.com/nextcloud/server/blob/4560ddd706d2ced7fb47487cf57095513ae2de70/lib/private/User/Database.php#L428

lib/private/User/Manager.php
```diff
@@ -288,20 +288,49 @@ public function checkPasswordNoLogging($loginName, $password) {
...
 		foreach ($this->backends as $backend) {
-			$backendUsers = $backend->getUsers($pattern, $limit, $offset);
+			$backendUsers = $backend->getUsers($pattern, $limit, $offset, $orderBy, $sort);
 			if (is_array($backendUsers)) {
 				foreach ($backendUsers as $uid) {
 					$users[$uid] = new LazyUser($uid, $this, null, $backend);
 				}
 			}
 		}
+		switch ($orderBy.' '.$sort) {
```
kesselb, line 301: Sorting here only works with $limit and $offset null, for paginated requests the result will be wrong.

lib/private/User/Manager.php
```diff
@@ -324,9 +353,40 @@ public function searchDisplayName($pattern, $limit = null, $offset = null) {
 			}
 		}
 
-		usort($users, function (IUser $a, IUser $b) {
-			return strcasecmp($a->getDisplayName(), $b->getDisplayName());
-		});
+		switch ($orderBy.' '.$sort) {
```
kesselb, line 356: Sorting here only works with $limit and $offset null, for paginated requests the result will be wrong.

lib/private/User/Database.php
yemkareems: Changed the query to support  email based search and removed the sort in Database.php

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -300,6 +300,96 @@ public function getDisabledUsersDetails(?int $limit = null, int $offset = 0): Da
...
+	 * @param string $sortOrder asc or desc
+	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, Provisioning_APIUserDetails|array{id: string}>}, array{}>
+	 *
+	 * 200: Users details returned based on last logged in information
+	 */
+	public function getLastLoggedInUsers(string $search = '',
+										 ?int   $limit = null,
+										 int    $offset = 0,
+										 string $sortMode = 'lastLogin',
+										 string $sortOrder = 'desc'
```
Pytal, line 322:
```suggestion
	public function getLastLoggedInUsers(string $search = '',
										 ?int   $limit = null,
										 int    $offset = 0,
										 string $sortOrder = 'desc'
```
Let's keep it simple as we only need sorting by last login

lib/public/IUserManager.php
```diff
@@ -126,7 +126,7 @@ public function checkPassword($loginName, $password);
 	 * @return \OCP\IUser[]
 	 * @since 8.0.0
 	 */
-	public function search($pattern, $limit = null, $offset = null);
+	public function search($pattern, $limit = null, $offset = null, $orderBy = 'uid', $sort = 'ASC');
```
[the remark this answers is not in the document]
yemkareems: @kesselb reverted these changes and implemented a new endpoint that will not break the LDAP

lib/private/User/Database.php
yemkareems: @kesselb Reverted these changes and implemented a new users/recent endpoint to get the users sorted based on lastLogin asc or desc

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: users/recent will work as it is without any params (default sort is by lastLogin desc) and in case asc/desc is required frontend can pass users/recent?offset=0&limit=25&search=&sortOrder=asc&sortBy=lastLogin params accordingly.

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -265,6 +265,91 @@ public function getDisabledUsersDetails(string $search = '', ?int $limit = null,
...
+			/* We have to handle offset ourselve for correctness */
+			$tempLimit = ($limit === null ? null : $limit + $offset);
+			foreach ($subAdminOfGroups as $group) {
+				$users = array_merge(
+					$users,
+					array_map(
+						fn (IUser $user): string => $user->getUID(),
+						array_filter(
+							$group->searchUsers($search, ($tempLimit === null ? null : $tempLimit - count($users))),
+							fn (IUser $user): bool => !$user->isEnabled()
```
come-nc, line 316: Why is this filtering disabled users?

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -265,6 +265,91 @@ public function getDisabledUsersDetails(string $search = '', ?int $limit = null,
...
+						)
+					)
+				);
+				if (($tempLimit !== null) && (count($users) >= $tempLimit)) {
+					break;
+				}
+			}
+			$users = array_slice($users, $offset);
+		}
```
come-nc, line 325: This whole branch is returning the wrong thing.

come-nc: Maybe we should not support subadmins here, because perf of sorting all group members by last login may be terrible? @artonge ?

lib/private/AllConfig.php
```diff
@@ -491,6 +491,51 @@ public function getUsersForUserValue($appName, $key, $value) {
...
+
+		$lastLoginSubSelect = $this->connection->getQueryBuilder();
+		$lastLoginSubSelect->select('configvalue')
+			->from('preferences', 'p2')
+			->where($lastLoginSubSelect->expr()->andX(
+				$lastLoginSubSelect->expr()->eq('p2.userid', 'uid'),
+				$lastLoginSubSelect->expr()->eq('p2.appid', $lastLoginSubSelect->expr()->literal('login')),
+				$lastLoginSubSelect->expr()->eq('p2.configkey', $lastLoginSubSelect->expr()->literal('lastLogin')),
+			));
+		$orderByExpression = $query->createFunction('(' . $lastLoginSubSelect->getSQL() .')');
```
come-nc, line 514: This feels unefficient to me, but maybe I’m wrong, wouldn’t a second join be faster ?
ping @Altahrim

lib/private/User/Manager.php
```diff
@@ -342,6 +342,36 @@ public function getDisabledUsers(?int $limit = null, int $offset = 0, string $se
...
+			if (($tempLimit !== null) && (count($users) >= $tempLimit)) {
+				break;
+			}
+			if ($backend instanceof IProvideEnabledStateBackend) {
+				$backendUsers = $backend->getDisabledUserList(($tempLimit === null ? null : $tempLimit - count($users)));
+				foreach ($backendUsers as $uid) {
+					$users[$uid] = new LazyUser($uid, $this, null, $backend);
+				}
+			}
+		}
```
come-nc, line 369: This part has nothing to do here, it’s about disabled users.
@yemkareems please test your PR more in depth before asking for reviews.

lib/public/IConfig.php
```diff
@@ -249,4 +249,13 @@ public function deleteAppFromAllUsers($appName);
...
 	public function getUsersForUserValue($appName, $key, $value);
+
+	/**
+	 * Gets the list of users based on their lastLogin info asc or desc
+	 *
+	 * @param string $search search users based on search params
+	 * @return array of user IDs
+	 * @since 30.0.0
+	 */
+	public function getLastLoggedInUsers($search);
```
come-nc, line 260: It should also contain the limit, and maybe the offset, to be efficient.

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -265,6 +265,91 @@ public function getDisabledUsersDetails(string $search = '', ?int $limit = null,
...
+	 *
+	 * @param string $search Text to search for
+	 * @param ?int $limit Limit the amount of users returned
+	 * @param int $offset Offset
+	 * @return DataResponse<Http::STATUS_OK, array{users: array<string, Provisioning_APIUserDetails|array{id: string}>}, array{}>
+	 *
+	 * 200: Users details returned based on last logged in information
+	 */
+	public function getLastLoggedInUsers(string $search = '',
+		?int   $limit = null,
```
come-nc, line 281: limit should not default to null, and maybe not even be nullable, otherwise you just get a list of all users.

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: @come-nc I had a call with @artonge on this and it was decided we will support last login sorting for admin alone and not for subadmin and groups

lib/private/AllConfig.php
yemkareems: @come-nc, @Altahrim Intially this was written in getDisplayNames and that broke the search by email. I ll refactor this query only to sort by lastLogin configvalue desc

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: Will default the limit to 25

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: @come-nc removed the references to old disabled users code

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: Sub admin and groups are not supported. Hence removed this

lib/private/User/Manager.php
yemkareems: @come-nc Sorry about the old code leftover here. Removed the old disabled users code and added the limit and offset to the query and cleaned it up here

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: Added offset and limit to the query

lib/public/IConfig.php
yemkareems: Added offset and limit to the query

lib/private/AllConfig.php
yemkareems: Refactored this query to similar to getDisplayNames query to make it better

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: Limit is defaulted to null and if no limit is given all users are fetched ordered by last login desc and if limit, offset and search are given the result set is fetched accordingly @come-nc @Pytal

apps/provisioning_api/lib/Controller/UsersController.php
```diff
@@ -265,6 +265,68 @@ public function getDisabledUsersDetails(string $search = '', ?int $limit = null,
 		]);
 	}
 
+	/**
+	 * @NoAdminRequired
```
come-nc, line 269:
```suggestion
```
Since you removed subadmin handling, this should be for admins only now.

come-nc (changes requested):

apps/provisioning_api/lib/Controller/UsersController.php
yemkareems: Removed the doc block and if condition

lib/private/AllConfig.php
```diff
@@ -491,6 +491,46 @@ public function getUsersForUserValue($appName, $key, $value) {
 		return $userIDs;
 	}
 
+	/**
+	 * Gets the list of users based on their lastLogin info asc or desc
```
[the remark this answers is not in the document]
yemkareems: Changed comment

lib/private/AllConfig.php
```diff
@@ -534,4 +574,12 @@ public function getUsersForUserValueCaseInsensitive($appName, $key, $value) {
...
 		return $this->systemConfig;
 	}
+
+	private function fixLimit($limit) {
+		if (is_int($limit) && $limit >= 0) {
+			return $limit;
+		}
+
+		return null;
+	}
```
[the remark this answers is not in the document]
yemkareems: negative limit and offset throwing up is handled in controller. So removed the same here

lib/private/User/Manager.php
```diff
@@ -342,6 +342,21 @@ public function getDisabledUsers(?int $limit = null, int $offset = 0, string $se
...
+		return array_combine(
+			$users,
+			array_map(
+				fn (string $uid): IUser => new LazyUser($uid, $this),
+				$users
+			)
+		);
+	}
+
+
```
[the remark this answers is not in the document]
yemkareems: Removed this method and used the config directly in controller

lib/private/AllConfig.php
```diff
@@ -491,6 +491,45 @@ public function getUsersForUserValue($appName, $key, $value) {
...
+			->from('users', 'u')
+			->leftJoin('u', 'preferences', 'p', $query->expr()->andX(
+				$query->expr()->eq('userid', 'uid'),
+				$query->expr()->eq('appid', $query->expr()->literal('login')),
+				$query->expr()->eq('configkey', $query->expr()->literal('lastLogin')))
+			)
+			// sqlite doesn't like re-using a single named parameter here
+			->where($query->expr()->iLike('uid', $query->createPositionalParameter('%' . $this->connection->escapeLikeParameter($search) . '%')))
+			->orWhere($query->expr()->iLike('displayname', $query->createPositionalParameter('%' . $this->connection->escapeLikeParameter($search) . '%')))
+			->orWhere($query->expr()->iLike('configvalue', $query->createPositionalParameter('%' . $this->connection->escapeLikeParameter($search) . '%')))
```
come-nc, line 518: This only makes sense to search in email, but this will not work because you join on configkey lastLogin.
So either you have to join twice with `preferences` if you want search to work in email address, or you should remove this line and search will not search in email address.

lib/private/AllConfig.php
```diff
@@ -491,6 +491,45 @@ public function getUsersForUserValue($appName, $key, $value) {
...
+			->orderBy($query->func()->lower('configvalue'), 'DESC')
+			->addOrderBy('uid_lower', 'ASC')
+			->setFirstResult($offset)
+			->setMaxResults($limit);
+
+		$result = $query->executeQuery();
+		$displayNames = [];
```
come-nc, line 525: Why is this var called $displayNames? It contains uids.
```suggestion
		$uids = [];
```

lib/private/AllConfig.php
```diff
@@ -491,6 +491,45 @@ public function getUsersForUserValue($appName, $key, $value) {
...
+		while ($row = $result->fetch()) {
```
come-nc, line 526:
```suggestion
		while (($row = $result->fetch()) !== false) {
```
You have to check against false with a strong check to not stop on a uid being `'0'` which is valid.

lib/private/AllConfig.php
```diff
@@ -491,6 +491,45 @@ public function getUsersForUserValue($appName, $key, $value) {
...
+	 * @param string $search search users based on search params
+	 * @return array of user IDs
+	 */
+	public function getLastLoggedInUsers(?int $limit = null, int $offset = 0, string $search = ''): array {
+		// TODO - FIXME
+		$this->fixDIInit();
+
+		$query = $this->connection->getQueryBuilder();
+
+		$query->select('uid', 'displayname')
```
come-nc, line 508:
```suggestion
		$query->selectDistinct('uid')
```
You do not use displayname so no need to select it.
And if you use selectDistinct you can avoid using uid as key in the foreach loop below and can return a list which is better than an associative array.
You can even return $result->fetchAll in column mode.

lib/private/AllConfig.php
come-nc, line 525:
```suggestion
		$result = $query->executeQuery();
		$uids = $result->fetchAll(\PDO::FETCH_COLUMN);
		$result->closeCursor();
		return $uids;
```
Like that, if you only select uid with distinct you can use this simple code block instead of a while.

lib/public/IConfig.php
```diff
@@ -249,4 +249,15 @@ public function deleteAppFromAllUsers($appName);
...
 	 */
 	public function getUsersForUserValue($appName, $key, $value);
+
+	/**
+	 * Gets the list of users based on their lastLogin info asc or desc
+	 *
+	 * @param int|null $limit how many records to fetch
+	 * @param int $offset from which offset to fetch
+	 * @param string $search search users based on search params
+	 * @return array of user IDs
```
come-nc, line 259:
```suggestion
	 * @return list<string> list of user IDs
```

come-nc (changes requested):

lib/private/AllConfig.php
yemkareems: Will add a extra join to have the search with email intact

lib/private/AllConfig.php
yemkareems: changed it to uids

lib/private/AllConfig.php
yemkareems: used selectDistinct and fetchAll

lib/private/AllConfig.php
```diff
@@ -491,6 +491,50 @@ public function getUsersForUserValue($appName, $key, $value) {
 		return $userIDs;
 	}
 
+	/**
+	 * Gets the list of user ids sorted by lastLogin, from most recent to least recent
+	 *
+	 * @param int|null $limit how many users to fetch
+	 * @param int $offset from which offset to fetch
+	 * @param string $search search users based on search params
+	 * @return array of user IDs
```
come-nc, line 500:
```suggestion
	 * @return list<string> list of user IDs
```
To please psalm

come-nc (approved):

artonge (approved):

artonge:
Rebased

lib/public/IConfig.php
```diff
@@ -249,4 +249,15 @@ public function deleteAppFromAllUsers($appName);
 	 * @since 8.0.0
 	 */
 	public function getUsersForUserValue($appName, $key, $value);
+
+	/**
+	 * Gets the list of users based on their lastLogin info asc or desc
```
Pytal, line 254:
```suggestion
	 * Gets the list of users sorted by lastLogin, from most recent to least recent
```
Since we can't choose asc or desc, copied from the other method

Pytal (approved):

artonge:
You probably need to fix the openapi check

kesselb (dismissed):
Why IConfig/AllConfig? :confused:

yemkareems:
> Why IConfig/AllConfig? 😕

It is mostly in the lines of the getDisabledUsers which calls IUserManager/Manager which internally calls IConfig/AllConfig. The Manager was not required because we just wanted the list of user and the LazyUser in the Manager was not required. So i did IConfig/AllConfig

kesselb:
`IConfig` is the public api to create/read/update/delete configuration values, either from configuration file, app config, or user preferences. 

I assume `getDisabledUsers` uses `getUsersForUserValue` because that's the way to retrieve user preferences. 

`getLastLoggedInUsers` does not belong in IConfig and should go somewhere else. The UserManager looks like a good place, but please check with Louis, and we are already using the QueryBuilder there, regrettably not via DI.

come-nc:
> `IConfig` is the public api to create/read/update/delete configuration values, either from configuration file, app config, or user preferences.
> 
> I assume `getDisabledUsers` uses `getUsersForUserValue` because that's the way to retrieve user preferences.
> 
> `getLastLoggedInUsers` does not belong in IConfig and should go somewhere else. The UserManager looks like a good place, but please check with Louis, and we are already using the QueryBuilder there, regrettably not via DI.

I disagree, it’s reading in the configuration table so it’s better situated in IConfig than IUserManager, no?

artonge:
I agree with @kesselb, this seems way too specific to be located in `IConfig`.

yemkareems:
> I agree with @kesselb, this seems way too specific to be located in `IConfig`.

Sure @kesselb , @artonge i ll move this to UserManager/Manager as suggested

artonge (approved):

lib/public/IUserManager.php
```diff
@@ -210,4 +210,15 @@ public function getByEmail($email);
...
+	/**
+	 * Gets the list of users sorted by lastLogin, from most recent to least recent
+	 *
+	 * @param int|null $limit how many records to fetch
+	 * @param int $offset from which offset to fetch
+	 * @param string $search search users based on search params
+	 * @return list<string> list of user IDs
+	 * @since 30.0.0
+	 */
+	public function getLastLoggedInUsers(?int $limit = null, int $offset = 0, string $search = ''): array;
```
nickvergessen, line 223: Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

PR #45251, opened 2024-05-10
fix(files): do not rely on unique fileid

skjnldsv:
Fix the rendering of duplicates fileids in the Files app

## How to 
1. Create a local external storage
2. Create a second one wih the exact same path
3. Open the Files app
4. See an error in the log and scrolling starts to break

## Solution
- We do not store files by their ID anymore
- Once we receive a node event, we will use the node payload as a source of truth ONLY if we have one entry in the store. If we don't we fetch the node data in the dav backend to refresh the content. This is an edge case, so the perf loss is accepted here


AndyScherzinger (approved):

susnux (approved):
:elephant: :broom:

PR #45255, opened 2024-05-10
fix(ci): Align mariadb versions with list of supported versions

susnux:
## Summary

We currently support 10.3, 10.5, 10.6 (old LTS) and 10.11 (current LTS) so we can drop 10.4 from CI and save a bit of CI time.

## Checklist

### DIFF
--- a/.github/workflows/phpunit-mariadb.yml
+++ b/.github/workflows/phpunit-mariadb.yml
@@ -53,7 +53,7 @@ jobs:
     strategy:
       matrix:
         php-versions: ['8.1']
-        mariadb-versions: ['10.3', '10.4', '10.5', '10.6', '10.11']
+        mariadb-versions: ['10.3', '10.5', '10.6', '10.11']
         include:
           - php-versions: '8.3'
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

nickvergessen (commented):
Also in .github?

nickvergessen (approved):

susnux:
> Also in .github?

The templates only check for LTS version 10.6 and 10.11 (and in few weeks probably 11.4)

solracsf:
Why have dropped 10.4 isntead of 10.3?
https://endoflife.date/mariadb

nickvergessen:
Sounds like a mistake. I think someone mentioned .3 to be an LTS, but .4 actually is the LTS

susnux:
No 10.3 is supported because of Ubuntu 20.04 shipping it and we support Ubuntu 20.04

10.4 was dropped on purpose as it is only a short term stable and will be eol when Nextcloud 30 is released

PR #45262, opened 2024-05-10
fix(SetupChecks): Detect CLI mode in HTTPS / URL generator check

joshtrichards:
## Summary

Check for CLI mode in the second part of this check rather than return a result we can't be 100% certain about when running the check at the command line.

## Checklist


susnux:
Hm, I am not fully convinced here, as this check means that the `cli.overwrite.url` is not secure which might also be a problem?

joshtrichards:
@susnux Your right; I was looking at this wrong. Still needed adjustment, but a little differently. Updated. Give it another look.

susnux (approved):

come-nc (approved):

PR #45277, opened 2024-05-13
feat(log): Allow to combine log.conditions to only log (app&user)

nickvergessen:
## Summary

Improve log conditioning to allow only debug logging entries for user & app, dedicated log level for each condition and "log message contains" additionally.

## Checklist

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1057,6 +1057,9 @@ $CONFIG = [
  *                this condition is met
  *  - ``apps``:   if the log message is invoked by one of the specified apps,
  *                this condition is met
+ *  - ``matches``: if all the conditions inside a group match,
+ *                this condition is met. This allows to log only entries to an app
+ *                by a few users.
  *
  * Defaults to an empty array.
  */
@@ -1064,6 +1067,15 @@ $CONFIG = [
 	'shared_secret' => '57b58edb6637fe3059b3595cf9c41b9',
 	'users' => ['sample-user'],
 	'apps' => ['files'],
+	'matches' => [
+		[
+			'shared_secret' => '57b58edb6637fe3059b3595cf9c41b9',
+			'users' => ['sample-user'],
+			'apps' => ['files'],
+			'loglevel' => 1,
+			'message' => 'contains substring'
+		],
+	],
 ],
 
 /**
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -14,6 +14,7 @@ use OC\AppFramework\Bootstrap\Coordinator;
 use OC\Log\ExceptionSerializer;
 use OCP\EventDispatcher\IEventDispatcher;
 use OCP\ILogger;
+use OCP\IRequest;
 use OCP\IUserSession;
 use OCP\Log\BeforeMessageLoggedEvent;
 use OCP\Log\IDataLogger;
@@ -153,7 +154,7 @@ class Log implements ILogger, IDataLogger {
 	 * @param array $context
 	 */
 	public function log(int $level, string $message, array $context = []): void {
-		$minLevel = $this->getLogLevel($context);
+		$minLevel = $this->getLogLevel($context, $message);
 		if ($level < $minLevel
 			&& (($this->crashReporters?->hasReporters() ?? false) === false)
 			&& (($this->eventDispatcher?->hasListeners(BeforeMessageLoggedEvent::class) ?? false) === false)) {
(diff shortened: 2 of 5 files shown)


lib/private/Log.php
```diff
@@ -281,6 +293,22 @@ public function getLogLevel($context): int {
...
 
+		if (isset($logCondition['matches'])) {
+			foreach ($logCondition['matches'] as $option) {
+				if (
+					(!isset($option['shared_secret']) || $this->checkLogSecret($option['shared_secret']))
+					&& (!isset($option['users']) || in_array($userId, $option['users'], true))
+					&& (!isset($option['apps']) || (isset($context['app']) && in_array($context['app'], $option['apps'], true)))
+					&& (!isset($option['message']) || str_contains($message, $option['message']))
+				) {
+					if (!isset($option['apps']) && !isset($option['loglevel']) && !isset($option['message'])) {
```
come-nc, line 304: I do not understand this inside `if`

config/config.sample.php
```diff
@@ -1048,13 +1048,25 @@
...
+ *                this condition is met. This allows to log only entries to an app
+ *                by a few users.
  *
  * Defaults to an empty array.
  */
 'log.condition' => [
 	'shared_secret' => '57b58edb6637fe3059b3595cf9c41b9',
 	'users' => ['sample-user'],
 	'apps' => ['files'],
+	'matches' => [
```
come-nc, line 1061:
```suggestion
	'and' => [
```
Or is that too nerdy?

come-nc:
```suggestion
	'allof' => [
```
maybe?

nickvergessen: I had that at some point all ready, but my mind couldn't settle if it's allof or anyof and then I wondered if it's more understandable for people

lib/private/Log.php
```diff
@@ -271,6 +278,11 @@ public function getLogLevel($context): int {
 			return ILogger::DEBUG;
 		}
```
miaulalala, line 279: Extract this into it's own method:

```suggestion:

	/**
	 * @param array $logCondition
	 * @param string $userId
	 * @return bool
	 * @throws \Psr\Container\ContainerExceptionInterface
	 * @throws \Psr\Container\NotFoundExceptionInterface
	 */
	private function processLogConditions(array $logCondition): ?bool {
		/**
		 * check for a special log condition - this enables an increased log on
		 * a per request/user base
		 * If the log condition is null, we need to check the for user, otherwise we skip
		 */
		if($this->logConditionSatisfied !== null) {
			return $this->logConditionSatisfied;
		}
		// check for secret token in the request
		if (isset($logCondition['shared_secret']) && $this->checkLogSecret($logCondition['shared_secret'])) {
			return true;
		}

		// check for user
		if (isset($logCondition['users'])) {
			$user = \OCP\Server::get(IUserSession::class)->getUser();

			if ($user === null) {
				// User is not known for this request yet
				return null;
			}

			if (in_array($user->getUID(), $logCondition['users'], true)) {
				// if the user matches set the log condition to satisfied
				return true;
			}
		}
		return null;
	}
```
I hope I understood the point of `null` for the `$this->logConditionSatisfied` correctly

lib/private/Log.php
```diff
@@ -291,6 +319,21 @@ public function getLogLevel($context): int {
...
+			!str_contains($request->getHeader('Content-Type'), 'application/x-www-form-urlencoded') &&
+			!str_contains($request->getHeader('Content-Type'), 'application/json')) {
+			$logSecretRequest = '';
+		} else {
+			$logSecretRequest = $request->getParam('log_secret', '');
+		}
+
+		// if token is found in the request change set the log condition to satisfied
+		return hash_equals($conditionSecret, $logSecretRequest);
+	}
```
miaulalala, line 335: Early return

```suggestion
		if ($request->getMethod() === 'PUT' &&
			!str_contains($request->getHeader('Content-Type'), 'application/x-www-form-urlencoded') &&
			!str_contains($request->getHeader('Content-Type'), 'application/json')) {
			return hash_equals($conditionSecret, '');
		}

		// if token is found in the request change set the log condition to satisfied
		return hash_equals($conditionSecret, $request->getParam('log_secret', ''));
	}
```

lib/private/Log.php
```diff
@@ -281,6 +293,22 @@ public function getLogLevel($context): int {
...
+					&& (!isset($option['users']) || in_array($userId, $option['users'], true))
+					&& (!isset($option['apps']) || (isset($context['app']) && in_array($context['app'], $option['apps'], true)))
+					&& (!isset($option['message']) || str_contains($message, $option['message']))
+				) {
+					if (!isset($option['apps']) && !isset($option['loglevel']) && !isset($option['message'])) {
+						$this->logConditionSatisfied = true;
+					}
+					return $option['loglevel'] ?? ILogger::DEBUG;
+				}
+			}
```
miaulalala, line 309: Run through the function recursively maybe?
```suggestion
				$loglevel = $this->getLogLevel($option, $option['message'] ?? $message);
			}
```

miaulalala (changes requested):

lib/private/Log.php
nickvergessen: I didn't understand this one.
Fixed all the other comments

miaulalala (approved):

AndyScherzinger (approved):

come-nc (approved):

PR #45295, opened 2024-05-13
Add temporary variable for border radius

marcoambrosini:


come-nc (approved):

szaimen:
It is missing in the theming app...

skjnldsv (approved):
Please [fix it in the theming app](https://github.com/nextcloud/server/blob/ee8a97be4330d402d5bc926f64c6d8ab6812df23/apps/theming/lib/Themes/DefaultTheme.php#L123) yep :+1: 
Otherwise, good from me

skjnldsv:
Failure related

apps/theming/css/default.css
```diff
@@ -49,6 +49,7 @@
   --default-font-size: 15px;
   --animation-quick: 100ms;
   --animation-slow: 300ms;
+  --border-radius-new-design: 10px; /** Nextcloud 29+ | temp variable to use old design */
```
susnux, line 52: 1. Are we really want to go with this name (`-new-design`)? We will use it as at least one year (EOL of Nextcloud 29), so we will keep it in Nextcloud 30 and 31 before we can default to `--border-radius-large`.
  So instead we could also use some meaningful name like `--border-radius-input` or `--border-radius-element` or `--border-radius-button`. So something we could also reuse if we decide to round button again or unround even further...
2. This is Nextcloud 30+ (comment)

*(marking this as red to have discussed this first, but feel free to dismiss my review if this is agreed)*

susnux (dismissed):

apps/theming/css/default.css
ShGKme: We can call it `-nextcloud-30`

susnux: > We can call it -nextcloud-30

But then again you need to remove and adjust all usage code with Nextcloud 33 / 34.
As said if you all think it makes sense the way it is, then dismiss my review :)

marcoambrosini: I really don't care about the name so whatever you prefer, just let me know :)

susnux: It is not about the naming, but the question:
Is this something temporary (we remove it with Nextcloud 33 when 29 is EOL). Or is this a variable for the border radius of interactive elements.

Because for the first one I am fine with "new-design", but that means in roughly one year we need to remove it and change the library again.
But if we want to have this variable around for interactive elements (list item button etc), then we should call it with a meaningful name like `--border-radius-element`.

szaimen (approved):

apps/theming/css/default.css
```diff
@@ -49,6 +49,7 @@
   --default-font-size: 15px;
   --animation-quick: 100ms;
   --animation-slow: 300ms;
+  --border-radius-element: 10px; /** Nextcloud 29+ | temp variable to use old design */
   --border-radius: 3px;
   --border-radius-large: 10px;
   --border-radius-rounded: 28px;
```
ShGKme, line 55: - Changed order to be consistent with other variables in the file and with `DefaultTheme.php`
- Changed comment to reflect new actual meaning

```suggestion
  --border-radius: 3px;
  --border-radius-large: 10px;
  --border-radius-rounded: 28px;
  /* Border radius of interactive elements such as buttons, input, navigation and list items. Available since Nextcloud 30. */
  --border-radius-element: 10px; 
```

ShGKme (approved):

susnux (approved):

marcoambrosini:
Damn whitespace 🤦 😅 

Screenshot 2024-05-29 at 09 13 14

PR #45302, opened 2024-05-14
Revert: "Check datadirectory owner, not config owner."

szaimen:
Reverts https://github.com/nextcloud/server/pull/27613 as it breaks on many instances.

- We should rather add a config option that allows to define a user that will be used if the user cannot be determined which will be done in a https://github.com/nextcloud/server/pull/45307

### DIFF
--- a/console.php
+++ b/console.php
@@ -41,26 +41,15 @@ try {
 		exit(1);
 	}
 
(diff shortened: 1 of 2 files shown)


come-nc (approved):

susnux (approved):
:donkey:

PR #45307, opened 2024-05-14
feat: allow to configure php.user

szaimen:
- [x] Needs https://github.com/nextcloud/server/pull/45302

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -941,6 +941,15 @@ $CONFIG = [
 'config_is_read_only' => false,
 
 /**
+ * In certain environments it is desired to set the config.php owner to
+ * something else than the user that is running the php process.
+ * In that case in order to determine the user that the php process uses,
+ * you can overwrite the user with this config flag for console.php and cron.php
+ * Defaults to ``''`` (empty string)
+ */
+'php.user' => '',
+
+/**
  * Logging
  */
 
--- a/console.php
+++ b/console.php
@@ -42,7 +42,10 @@ try {
 	}
 
 	$user = posix_getuid();
-	$userName = posix_getpwuid($user) !== false ? posix_getpwuid($user)['name'] : '';
+	$userNameArray = posix_getpwuid($user);
+	if ($userNameArray !== false) {
+		$userName = $userNameArray['name'];
+	}
 	$configUser = fileowner(OC::$configDir . 'config.php');
 	$configuredUser = $config->getSystemValueString('php.user', '');
 	if ($user !== $configUser && $userName !== $configuredUser) {
--- a/console.php
+++ b/console.php
@@ -43,12 +43,13 @@ try {
 
 	$user = posix_getuid();
 	$userNameArray = posix_getpwuid($user);
+	$username = null;
 	if ($userNameArray !== false) {
 		$userName = $userNameArray['name'];
 	}
 	$configUser = fileowner(OC::$configDir . 'config.php');
 	$configuredUser = $config->getSystemValueString('php.user', '');
-	if ($user !== $configUser && $userName !== $configuredUser) {
+	if ($user !== $configUser && $username !== null && $userName !== $configuredUser) {
 		echo "Console has to be executed with the user that owns the file config/config.php" . PHP_EOL;
 		echo "Current user id: " . $user . PHP_EOL;
 		echo "Owner id of config.php: " . $configUser . PHP_EOL;
--- a/console.php
+++ b/console.php
@@ -42,13 +42,16 @@ try {
 	}
 
 	$user = posix_getuid();
+	$userName = posix_getpwuid($user)['name'];
 	$configUser = fileowner(OC::$configDir . 'config.php');
-	if ($user !== $configUser) {
+	$configuredUser = $config->getSystemValueString('php.user', '');
+	if ($user !== $configUser && $userName !== $configuredUser) {
 		echo "Console has to be executed with the user that owns the file config/config.php" . PHP_EOL;
 		echo "Current user id: " . $user . PHP_EOL;
 		echo "Owner id of config.php: " . $configUser . PHP_EOL;
 		echo "Try adding 'sudo -u #" . $configUser . "' to the beginning of the command (without the single quotes)" . PHP_EOL;
 		echo "If running with 'docker exec' try adding the option '-u " . $configUser . "' to the docker command (without the single quotes)" . PHP_EOL;
+		echo "Another option is to configure 'php.user' in config.php which will overwrite this check.";
 		exit(1);
 	}
 
--- a/console.php
+++ b/console.php
@@ -42,7 +42,7 @@ try {
 	}
 
 	$user = posix_getuid();
-	$userName = posix_getpwuid($user)['name'];
+	$userName = posix_getpwuid($user) !== false ? posix_getpwuid($user)['name'] : '';
 	$configUser = fileowner(OC::$configDir . 'config.php');
 	$configuredUser = $config->getSystemValueString('php.user', '');
 	if ($user !== $configUser && $userName !== $configuredUser) {
--- a/cron.php
+++ b/cron.php
@@ -130,7 +130,7 @@ Options:
 		}
 
 		$user = posix_getuid();
-		$userName = posix_getpwuid($user)['name'];
+		$userName = posix_getpwuid($user) !== false ? posix_getpwuid($user)['name'] : '';
 		$configUser = fileowner(OC::$configDir . 'config.php');
 		$configuredUser = $config->getSystemValueString('php.user', '');
 		if ($user !== $configUser && $userName !== $configuredUser) {
--- a/cron.php
+++ b/cron.php
@@ -130,7 +130,10 @@ Options:
 		}
 
 		$user = posix_getuid();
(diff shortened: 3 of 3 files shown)


config/config.sample.php
```diff
@@ -940,6 +940,15 @@
...
 'config_is_read_only' => false,
 
+/**
+ * In certain environments it is desired to set the config.php owner to
+ * something else than the user that is running the php process.
+ * In that case in order to determine the user that the php process uses,
+ * you can overwrite the user with this config flag for console.php and cron.php
+ * Defaults to ``''`` (empty string)
+ */
+'config.owner' => '',
```
come-nc, line 950: Should be `php.user` or something, no? Because the config owner we can get from the file system, what should be configurable is the target user to run php commands, defaulting to config owner.

console.php
```diff
@@ -43,12 +43,14 @@ function exceptionHandler($exception) {
 
 	$user = posix_getuid();
 	$configUser = fileowner(OC::$configDir . 'config.php');
-	if ($user !== $configUser) {
+	$configuredUser = $config->getSystemValueString('config.owner', '');
```
come-nc, line 46:
```suggestion
	$configUser = $config->getSystemValueString('php.user', $configUser);
```

szaimen: not sure about this

console.php
```diff
@@ -43,12 +43,14 @@ function exceptionHandler($exception) {
 
 	$user = posix_getuid();
 	$configUser = fileowner(OC::$configDir . 'config.php');
-	if ($user !== $configUser) {
+	$configuredUser = $config->getSystemValueString('php.user', '');
+	if ($user !== $configUser && $configuredUser === '') {
```
come-nc, line 47: You are not checking that the user is `$configuredUser`.

come-nc (changes requested):
I insist that it should use the new config value instead of the config.php owner if the value is filled.
Is if php.user is empty, it checks that running user is the owner of config.php.
If php.user is filled, it checks that running user is the one in php.user.

See my previous change suggestion.

szaimen:
> I insist that it should use the new config value instead of the config.php owner if the value is filled. Is if php.user is empty, it checks that running user is the owner of config.php. If php.user is filled, it checks that running user is the one in php.user.
> 
> See my previous change suggestion.

but we already get the user via posix_getuid or not?

come-nc:
> > I insist that it should use the new config value instead of the config.php owner if the value is filled. Is if php.user is empty, it checks that running user is the owner of config.php. If php.user is filled, it checks that running user is the one in php.user.
> > See my previous change suggestion.
> 
> but we already get the user via posix_getuid or not?

We do, and we want to check that it’s the correct one.

szaimen:
> > > I insist that it should use the new config value instead of the config.php owner if the value is filled. Is if php.user is empty, it checks that running user is the owner of config.php. If php.user is filled, it checks that running user is the one in php.user.
> > > See my previous change suggestion.
> > 
> > 
> > but we already get the user via posix_getuid or not?
> 
> We do, and we want to check that it’s the correct one.

all right, done!

cron.php
```diff
@@ -131,10 +131,12 @@
 
 		$user = posix_getuid();
 		$configUser = fileowner(OC::$configDir . 'config.php');
-		if ($user !== $configUser) {
+		$configuredUser = $config->getSystemValueString('php.user', '');
+		if ($user !== $configUser && $user !== $configuredUser) {
```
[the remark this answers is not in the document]
szaimen: apparently we need to rename php.user to php.userid? WDYT @come-nc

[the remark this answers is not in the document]
szaimen: tried to fix it with https://github.com/nextcloud/server/pull/45307/commits/b3d21be5abc71c7152747770fec8d385c73b5899

come-nc (approved):

console.php
```diff
@@ -42,13 +42,16 @@ function exceptionHandler($exception) {
 	}
 
 	$user = posix_getuid();
+	$userName = posix_getpwuid($user)['name'];
```
ChristophWurst, line 45: not handled: function might return false: https://www.php.net/manual/en/function.posix-getpwuid.php

ChristophWurst (commented):
:+1: otherwise

console.php
szaimen: https://github.com/nextcloud/server/pull/45307/commits/f49d3f07621f76c2f5ed210dd6776ad816683f45

console.php
```diff
@@ -42,13 +42,16 @@ function exceptionHandler($exception) {
 	}
 
 	$user = posix_getuid();
+	$userName = posix_getpwuid($user) !== false ? posix_getpwuid($user)['name'] : '';
```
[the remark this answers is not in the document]
szaimen: https://github.com/nextcloud/server/pull/45307/commits/137986f30d178cf15a4ded25e906df0e130560dd

console.php
```diff
@@ -42,13 +42,19 @@ function exceptionHandler($exception) {
...
 
 	$user = posix_getuid();
+	$userNameArray = posix_getpwuid($user);
+	if ($userNameArray !== false) {
+		$userName = $userNameArray['name'];
+	}
 	$configUser = fileowner(OC::$configDir . 'config.php');
-	if ($user !== $configUser) {
+	$configuredUser = $config->getSystemValueString('php.user', '');
+	if ($user !== $configUser && $userName !== $configuredUser) {
```
ChristophWurst, line 51: $userName will only be defined if the if block was executed. otherwise this code will trigger an error

https://3v4l.org/JRnhk

szaimen: Right 🙈

cron.php
```diff
@@ -130,11 +130,18 @@
...
 		$user = posix_getuid();
+		$userNameArray = posix_getpwuid($user);
+		$username = null;
+		if ($userNameArray !== false) {
+			$userName = $userNameArray['name'];
+		}
 		$configUser = fileowner(OC::$configDir . 'config.php');
-		if ($user !== $configUser) {
+		$configuredUser = $config->getSystemValueString('php.user', '');
+		if ($user !== $configUser && $username !== null && $userName !== $configuredUser) {
```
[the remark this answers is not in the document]
szaimen: Looks like this is not true? How can I fix this?

console.php
```diff
@@ -42,13 +42,20 @@ function exceptionHandler($exception) {
...
 	$configUser = fileowner(OC::$configDir . 'config.php');
-	if ($user !== $configUser) {
+	$configuredUser = $config->getSystemValueString('php.user', '');
+	if ($user !== $configUser && $username !== null && $userName !== $configuredUser) {
 		echo "Console has to be executed with the user that owns the file config/config.php" . PHP_EOL;
 		echo "Current user id: " . $user . PHP_EOL;
 		echo "Owner id of config.php: " . $configUser . PHP_EOL;
 		echo "Try adding 'sudo -u #" . $configUser . "' to the beginning of the command (without the single quotes)" . PHP_EOL;
 		echo "If running with 'docker exec' try adding the option '-u " . $configUser . "' to the docker command (without the single quotes)" . PHP_EOL;
+		echo "Another option is to configure 'php.user' in config.php which will overwrite this check.";
```
icewind1991, line 58: I think as worded this is misleading, the option doesn't overwrite the check, it changes that the expected value of the check is.

I would go with something like

>  If the config file is not owned by the user running the webserver you can set the correct user by setting the 'php.user' option in your config.php

console.php
```diff
@@ -42,13 +42,20 @@ function exceptionHandler($exception) {
...
 	$user = posix_getuid();
+	$userNameArray = posix_getpwuid($user);
+	$username = null;
+	if ($userNameArray !== false) {
+		$userName = $userNameArray['name'];
+	}
 	$configUser = fileowner(OC::$configDir . 'config.php');
-	if ($user !== $configUser) {
+	$configuredUser = $config->getSystemValueString('php.user', '');
+	if ($user !== $configUser && $username !== null && $userName !== $configuredUser) {
```
icewind1991, line 52: This check always needs to check for `$configuredUser` if set. Also accepting `$configUser` can lead to the very issue this is trying to prevent.

icewind1991 (changes requested):

cron.php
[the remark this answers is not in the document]
Dreamsorcerer: Are you sure? Documentation says it should return `false` on failure: https://www.php.net/manual/en/function.posix-getpwuid.php

Dreamsorcerer:
Looks like a reasonable solution, if there's not an easy way to just verify if we can write to the directory (as the user is not really relevant, just whether we have write access). I don't remember my PHP so well, but in Python I'd probably just try/except a file write as a test.

icewind1991:
> as the user is not really relevant, just whether we have write access

The user is actually relevant (there is a separate check for checking write access to the data dir iirc). If an `occ`/`cron` command ends up create a file it will be owned by whatever user the command is ran as. So if a user does `sudo occ ...`. Any file created by that will then be read-only to the webserver.

szaimen:
As I wrote in the internal chat: can someone please take over this PR? It looked like a simple change but got more and more complicated. I fear I currently do not have the capacity to finish this as I am busy with AIO and Enterprise-AIO.

PR #45330, opened 2024-05-15
fix: remove mention of disabling file locking from config sample

icewind1991:
Disabling transactional locking is almost always a bad idea and can easily lead to data corruption or another range of unexpected behavior.

Removing the mention of this footgun hopefully reduces the number of people disabling the locking without proper understanding.

### DIFF
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -2269,21 +2269,6 @@ $CONFIG = [
  */
 'max_filesize_animated_gifs_public_sharing' => 10,
 
-
-/**
- * Enables transactional file locking.
- * This is enabled by default.
- *
- * Prevents concurrent processes from accessing the same files
- * at the same time. Can help prevent side effects that would
- * be caused by concurrent operations. Mainly relevant for
- * very large installations with many users working with
- * shared files.
- *
- * Defaults to ``true``
- */
-'filelocking.enabled' => true,
-
 /**
  * Set the lock's time-to-live in seconds.
  *


config/config.sample.php
```diff
@@ -2185,21 +2185,6 @@
  */
 'max_filesize_animated_gifs_public_sharing' => 10,
 
-
```
joshtrichards, line 2188: GitHub won't let me make suggestions on deleted lines, but what if instead of removing it we adjusted the language to be a bit more scary like the [documentation](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/files_locking_transactional.html) already is + tried to emphasize it's not the same as `files_lock`:

```
/**
 * Transactional file locking
 * Locks files to avoid file corruption during normal operation. 
 * 
 * WARNING: Transactional file locking is not for preventing multiple users from editing 
 * the same document, or giving notice that other users are working on the same document. 
 * Multiple users can open and edit a file at the same time and Transactional File 
 * locking does not prevent this. Rather, it prevents simultaneous file saving. This
 * is enabled by default and should generally be left enabled. 
 * 
 * IMPORTANT: If you are experiencing problems with user facing locking, such 
 * as to avoid simultaneously edits, you may be looking for the `files_lock` app and 
 * its settings and/or settings within the client apps you're using. Do not disable this
 * parameter unless you are extremely sure you know what you're doing.

joshtrichards: Makes sense. :+1: 

The toggle was only added for turning it on from the looks of it (when it wasn't on by default since it was still experimental at the time): 2f4f468399d316157979f747d2418fb5cff8d3e0

That need no longer applies today.

Made sense to probably keep it around for a bit when it switched to being on by default in 4880d7729bf6442280be71ce91771ef3e077c54c, but those days are long over. :-)

joshtrichards (approved):

joshtrichards:
We can probably pull the unit test override too since it's extraneous:

https://github.com/nextcloud/server/blob/5a6e48e85034aa81862b1fcbd93cb4ce58165e72/lib/private/Server.php#L1088

skjnldsv (approved):

PR #45344, opened 2024-05-16
[stable27] fix: Keep download action for files and hide only for relevant files

juliusknorr:
Deleting the file actions if permissions do not apply causes them to generally disappear as the list of file actions to show is built once the user clicks the action menu, not when the file list row is rendered.

This PR will adapt the already existing shouldShow callback that is used for inline actions to regular ones so that we can check at time of displaying the menu with the context of the current file and hide it accordingly.


apps/files_sharing/src/share.js
```diff
@@ -86,19 +86,6 @@ import { getCapabilities } from '@nextcloud/capabilities'
 				var tr = oldCreateRow.apply(this, arguments)
 				var sharePermissions = OCA.Sharing.Util.getSharePermissions(fileData)
 
-				if (fileData.permissions === 0) {
-					// no permission, disabling sidebar
-					delete fileActions.actions.all.Comment
-					delete fileActions.actions.all.Details
-					delete fileActions.actions.all.Goto
```
juliushaertl, line 93: This is already done with the permissions checks.

artonge (approved):
The change make sense, but I am wondering whether the issue was always there or if it was introduced by a recent change.

nfebe (approved):
Changes look okay.

Two questions:

- How to reproduce this bug?
- Only nc27 affected?

juliushaertl:
> How to reproduce this bug?

Sorry should have mentioned that in the description:

1. User A: Create a share without download permissions to User B
2. User B: See the incoming share in the file list
3. User B: Try to download another file through the file actions menu

> Only nc27 affected?

Yes, seems fine with 28+ due to the files2vue migration.

nfebe (dismissed):
Sorry, this needs a little adjustment, so this works correctly on the main file listing page but then on the "Shares" tab the "Download" option still appears. Granted the user cannot download the file but it should not appear at all there.

juliushaertl:
> Sorry, this needs a little adjustment, so this works correctly on the main file listing page but then on the "Shares" tab the "Download" option still appears. Granted the user cannot download the file but it should not appear at all there.

I tried but wasn't able to come up with a proper fix for this before going on vacation. It seems we're lacking to expose the share attributes that are [used to check if the button should be hidden](https://github.com/nextcloud/server/blob/efa6a3138dc2d4d4f8f5bebb0034c68716286e3d/apps/files/js/fileactions.js#L659) when [constructing the objects for the shared file listing](https://github.com/nextcloud/server/blob/8f1bf13ae3046400ce6248fb13e5515e8e9ed5c4/apps/files_sharing/js/sharedfilelist.js#L390).

@artonge Maybe @Fenn-CS or someone else from the files team can take over this one to get into the second RC?

Pytal (approved):

PR #45347, opened 2024-05-16
fix: Only list remnants as disabled if option is enabled

come-nc:
## Summary

In 28 a new config option was added to consider remnants as disabled users.
But they were appearing in the list of disabled users even if the new option was not used, which can be confusing.

With this PR remnants are back at being ghosts by default and will only appear in disabled user list if the option is checked.
One problem is that the option is only checked on the first LDAP configuration and applied to all of them regarding to listing disabled users, because there is no efficient way to do otherwise (all remnants are in the same place). We already have this kind of trick for other things and expect that LDAP configurations are not too different from one another on the same server usually.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/User_Proxy.php
+++ b/apps/user_ldap/lib/User_Proxy.php
@@ -440,6 +440,9 @@ class User_Proxy extends Proxy implements IUserBackend, UserInterface, IUserLDAP
 	}
 
 	public function getDisabledUserList(?int $limit = null, int $offset = 0, string $search = ''): array {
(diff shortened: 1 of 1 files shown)


blizzz (approved):

artonge (approved):

Pytal (approved):

come-nc:
@Pytal We should backport to 29 and 28, right?

Pytal:
Not 100% sure here, will let you decide @come-nc

PR #45352, opened 2024-05-16
fix(login): Disable login button when already logging in

SystemKeeper:
## Summary

Fixes "State token does not match" in some cases (e.g. double click on the login button)

## How to reproduce:

* Start a login flow (V1 or V2), e.g. by logging in via the desktop client
* Enter username and password
* Click "Log in" and when the button changes to "Logging in" click the button again
* You're now prompted with a "temporary error". 
* Enter your password
* See the error "State token does not match"

## Before

image

## After

image

## Checklist


ChristophWurst (approved):

ChristophWurst:
To prevent double submit through enter we could adjust LoginForm.vue and ignore form submit while loading=true as well. Apparently this would be possible with a disabled field set: https://stackoverflow.com/questions/3507958/how-can-i-make-an-entire-html-form-readonly

SystemKeeper:
Right, I didn’t think of enter. Let me adjust the PR

SystemKeeper:
> Apparently this would be possible with a disabled field set

I tried that, but then I am unable to login in any way (it always ends up in a temporary error). Therefore I prevented another submit of the form, as we already have a submit handler.

ChristophWurst (approved):

provokateurin (approved):

juliusknorr (approved):

PR #45370, opened 2024-05-16
feat: Add back searching in disabled user list

### LINKED ISSUE #44936: [Bug]: Search in the"disabled users" section does not work

come-nc:
* Resolves: #44936 

## Summary

When disabled users where moved to their own endpoint we overlooked search, so adding it back.
The search is done case-insensitive in uid and display name.

## TODO

- [x] Decide if we should also search in email field, and implement it
- [x] Decide if the BC break is ok or should be avoided with a new interface

## Checklist


come-nc:
> ## TODO
> * [x]  Decide if we should also search in email field, and implement it

Implemented.

> * [ ]  Decide if the BC break is ok or should be avoided with a new interface

I think this is fine, because a class implementing the interface for 30 also implements the one for 28.0.0, as it’s only a new parameter.
So, as the breaking is minimal and it’s easy for concerned apps (and that’s only apps implementing a user backend with special disabled user treatment, so only user_ldap to my knowlegde) to support all Nextcloud versions.

provokateurin (commented):
I would prefer separate methods for searching quite a lot over changing the behavior of the existing methods. Would that be possible or is it too much work or impractical?

provokateurin:
Ok in light of the discussion in https://github.com/nextcloud/server/issues/44936 it makes sense to extend the existing endpoint instead of creating a new one

artonge (approved):
~~So BC is not a real BC?~~

Edit: seen your second comment

lib/public/User/Backend/IProvideEnabledStateBackend.php
```diff
@@ -49,8 +49,9 @@ public function setUserEnabled(string $uid, bool $enabled, callable $queryDataba
 	 * Get the list of disabled users, to merge with the ones disabled in database
 	 *
 	 * @since 28.0.0
+	 * @since 30.0.0 $search parameter added
```
Pytal, line 52:
```suggestion
```
Same here

Pytal (changes requested):

Altahrim (approved):

lib/public/User/Backend/IProvideEnabledStateBackend.php
come-nc: I planned on changing it in the backport but maybe already in this one we can use `@since 28.0.6 $search parameter added` ?

Pytal: Good idea, let's go with that then 🚀

provokateurin: IMO in the main branch it should be 30.0.0 and then for 29 and 28 it should be 28.x.x and 29.x.x respectively. It should not be 28.0.0/29.0.0 as those versions don't actually have the parameter.

Pytal (approved):

nickvergessen:
Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id2

PR #45375, opened 2024-05-16
Move EventSource to `OC` namespace

come-nc:
## Summary

There is an interface for it and a factory in OCP, so there’s no need to keep it in legacy folder.

## Checklist


kesselb (approved):

lib/private/EventSourceFactory.php
```diff
@@ -27,20 +30,12 @@
...
+	public function __construct(
+		private IRequest $request,
+	) {
 	}
 
-	/**
-	 * Create a new event source
-	 *
-	 * @return IEventSource
-	 * @since 28.0.0
```
susnux, line 41: why did you remove the comment?

susnux (approved):

lib/private/EventSourceFactory.php
come-nc: It’s duplicating info from the OCP interface.

PR #45391, opened 2024-05-17
feat(TaskProcessingApi): Add endpoint for getting the next task

provokateurin:
## Summary

The ExApps need a way to get the next task.
@marcelklehr I think there is a bug as the tasks are not marked as running when the getNextScheduledTask method of the service is called (or maybe it was also intentional and the endpoint should do it manually).

## Checklist


marcelklehr:
Ah, I missed this, yes.

> @marcelklehr I think there is a bug as the tasks are not marked as running when the getNextScheduledTask method of the service is called (or maybe it was also intentional and the endpoint should do it manually).

Yeah, let's do this manually directly in the endpoint

bigcat88:
We need endpoint for getting nextTask by specifying multiple task_types. 
One provider can process multiple task types and we do not want to spam NC with one nextTask request for each task_type.

Or just to change `string $taskTypeId` to array of strings.

julien-nc (changes requested):
* I agree with @marcelklehr that it makes more sense to change the task status to "running" in the controller method.
The `Manager::getNextScheduledTask` method might be used in some context where we don't want to change to task status.

* Also, how about having a `setTaskRunning` parameter (with `true` as default value) in `TaskProcessingApiController::getNextScheduledTask` ? Just in case one wants to get the next scheduled task without actually processing it.

* About requesting multiple task types in `getNextScheduledTask`, much needed indeed. @marcelklehr Would you go with another endpoint or change the current one?

provokateurin:
I don't see a reason why you would ever need to know the next task without also running it at the same time, but I can change it if you really want :woman_shrugging:

marcelklehr:
> I don't see a reason why you would ever need to know the next task without also running it at the same time

The synchronous background job will re-schedule itself based on the fact whether there is another job waiting to be executed.

marcelklehr:
> Would you go with another endpoint or change the current one?

We can add this to the current endpoint.

core/Controller/TaskProcessingApiController.php
```diff
@@ -402,4 +402,36 @@ public function cancelTask(int $taskId): DataResponse {
...
+			$task = $this->taskProcessingManager->getNextScheduledTask($taskTypeIds, true);
+
+			/** @var CoreTaskProcessingTask $json */
+			$json = $task->jsonSerialize();
+
+			return new DataResponse([
+				'task' => $json,
+				'provider' => [
+					// TODO: Fetch provider name from DB
+					'name' => 'llm2:tinyllama-1.1b-chat-v1.0.Q4_0:summary',
```
provokateurin, line 428: The only remaining part that needs to be implemented. Waiting on an answer from @marcelklehr

core/Controller/TaskProcessingApiController.php
```diff
@@ -402,4 +402,36 @@ public function cancelTask(int $taskId): DataResponse {
...
+	 *
+	 * @param list<string> $taskTypeIds The ids of the task types
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTaskProcessingTask, provider: array{name: string}}, array{}>|DataResponse<Http::STATUS_NO_CONTENT, null, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 *  200: Task returned
+	 *  204: No task found
+	 */
+	#[NoAdminRequired]
+	#[ApiRoute(verb: 'GET', url: '/tasks/next', root: '/taskprocessing')]
+	public function getNextScheduledTask(array $taskTypeIds): DataResponse {
```
nickvergessen, line 417: this method is callable for any user and it will freely give out any scheduled tasks of other users.

nickvergessen (changes requested):

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -191,6 +194,10 @@ public function beforeController($controller, $methodName) {
 			}
 		}
 
+		if ($this->hasAnnotationOrAttribute($reflectionMethod, 'ExAppRequired', ExAppRequired::class) && (!$this->userSession instanceof Session || $this->userSession->getSession()->get('app_api') !== true)) {
```
nickvergessen, line 197: Could also not introduce the annotation (and then simply directly only check for the attribute)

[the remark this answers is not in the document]
bigcat88: Suggestion:

1. Rename `#[ExAppRequired]` to `#[ExAppRoute]`
2. Add check for `ExAppRoute` in `if (!$this->isLoggedIn) {` at line 134

This will ensure that ExApp can call non public endpoints, and this still can be used with/without `PublicPage` and not change the logic of other attributes(like SubAdminRequired).

core/Controller/TaskProcessingApiController.php
```diff
@@ -348,15 +349,15 @@ public function setProgress(int $taskId, float $progress): DataResponse {
 	 *  200: Result updated successfully
 	 *  404: Task not found
 	 */
-	#[NoAdminRequired]
+	#[ExAppRequired]
 	#[ApiRoute(verb: 'POST', url: '/tasks/{taskId}/result', root: '/taskprocessing')]
 	public function setResult(int $taskId, ?array $output = null, ?string $errorMessage = null): DataResponse {
 		try {
 			// Check if the current user can access the task
```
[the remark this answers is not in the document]
provokateurin: Nope, this is how Marcel implemented the authorization check. I didn't change anything there.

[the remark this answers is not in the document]
provokateurin: Comment can be removed as only ExApps can call this endpoint, but it was also insecure because there was no actual check. I thought this was done automatically by getting the task, but apparently this was meant as a TODO and not as an explanation.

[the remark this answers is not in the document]
marcelklehr: It was meant as an explanation, when the line still read `getUserTask` ;)

bigcat88:
ExApp needs to call these APIs to process tasks:

`getNextScheduledTask`
`getFileContents`
`setProgress`
`setResult`

more precisely, there should be two versions of **getFileContents** - one for the UI with `userId` and other one for ExApp with `[ExAppRequired]` without checking for `userId`

cause only having those 4 endpoint ExApp can have no "system" flag to process tasks from what I see

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -127,7 +130,12 @@ public function beforeController($controller, $methodName) {
 
 		// security checks
 		$isPublicPage = $this->hasAnnotationOrAttribute($reflectionMethod, 'PublicPage', PublicPage::class);
-		if (!$isPublicPage) {
+
+		if ($this->hasAnnotationOrAttribute($reflectionMethod, 'ExAppRequired', ExAppRequired::class)) {
+			if ((!$this->userSession instanceof Session || $this->userSession->getSession()->get('app_api') !== true)) {
```
nickvergessen, line 135:
```suggestion
			if (!$this->userSession instanceof Session || $this->userSession->getSession()->get('app_api') !== true) {
```

lib/public/TaskProcessing/IManager.php
```diff
@@ -37,6 +37,14 @@ public function hasProviders(): bool;
 	 */
 	public function getProviders(): array;
 
+	/**
+	 * @param string $taskType
+	 * @return IProvider
+	 * @throws \OCP\TaskProcessing\Exception\Exception
+	 * @since 30.0.0
+	 */
+	public function getPreferredProvider(string $taskType);
```
nickvergessen, line 46: But you can't specify what you prefer?
It returns the first one matching the taskType?

```suggestion
	public function getFirstProviderForTaskType(string $taskType);
```

nickvergessen (approved):
The security middleware looks good now

lib/public/TaskProcessing/IManager.php
marcelklehr: It's a stand-in implementation until we implement the settings UI for the preferred providers

marcelklehr (approved):

bigcat88:
I didn't change the logic inside "getNextScheduledTask" - but it doesn't seem right to me for these reasons:

1. `$task->setStatus(Task::STATUS_RUNNING);` won't write anything to the database, right?
2. Reading and writing to the database must be atomic within “getNextScheduledTask” (so that two requests for processing a task in parallel do not receive the same task)
3. I didn’t understand the logic at all why `$providerIds` is there as input and what it should do, that’s why I didn’t change the entry to atomic inside “getNextScheduledTask”

core/Controller/TaskProcessingApiController.php
```diff
@@ -396,9 +437,48 @@ public function cancelTask(int $taskId): DataResponse {
...
+	public function getNextScheduledTask(array $providerIds, array $taskTypeIds): DataResponse {
+		try {
+			$task = $this->taskProcessingManager->getNextScheduledTask($taskTypeIds);
+
+			$provider = $this->taskProcessingManager->getPreferredProvider($task->getTaskTypeId());
+			if (!in_array($provider->getId(), $providerIds, true)) {
+				throw new NotFoundException();
+			}
+
+			$task->setStatus(Task::STATUS_RUNNING);
```
marcelklehr, line 468: Ah, yes, good point @bigcat88, this needs to be persisted

core/Controller/TaskProcessingApiController.php
```diff
@@ -396,9 +437,48 @@ public function cancelTask(int $taskId): DataResponse {
...
 			return new DataResponse(['message' => $this->l->t('Not found')], Http::STATUS_NOT_FOUND);
-		} catch (Exception $e) {
+		} catch (Exception) {
+			return new DataResponse(['message' => $this->l->t('Internal error')], Http::STATUS_INTERNAL_SERVER_ERROR);
+		}
+	}
+
+	/**
+	 * Returns the next scheduled task for the taskTypeId
+	 *
```
marcelklehr, line 449:
```suggestion
     * 
	 * It could be that a provider is looking for new tasks but the task type of the next scheduled task is actually configured to be run with a different provider
```

core/Controller/TaskProcessingApiController.php
```diff
@@ -396,9 +437,48 @@ public function cancelTask(int $taskId): DataResponse {
...
+	 *  204: No task found
+	 */
+	#[ExAppRequired]
+	#[ApiRoute(verb: 'GET', url: '/tasks_provider/next', root: '/taskprocessing')]
+	public function getNextScheduledTask(array $providerIds, array $taskTypeIds): DataResponse {
+		try {
+			$task = $this->taskProcessingManager->getNextScheduledTask($taskTypeIds);
+
+			$provider = $this->taskProcessingManager->getPreferredProvider($task->getTaskTypeId());
+			if (!in_array($provider->getId(), $providerIds, true)) {
```
marcelklehr, line 464: Imagine the task queue is as follows


````
1. free_prompt
2. summarize
````

and the configuration for the preferred providers says

```
free_prompt: context_chat
summarize: llm2-llama-3
```

Then llm2 calls this endpoint with `getNextScheduledTask([free_prompt, summarize], [llm2-llama-3])`

What will happen is:

 - getNextScheduledTask will return the free_prompt task
 - But that is configured to be run by context_chat
 - So, the endpoint returns NotFoundResponse, even though there is the summarize task waiting that fulfills the requirements of the request

Comment (bigcat88) [reply core/Controller/TaskProcessingApiController.php:464]: then we all this logic should do on the database level with query with automatic changing of `status` to `running` upon selecting task.

will try to do that...

**Review by bigcat88 — APPROVED** [review]

**Review by AndyScherzinger — APPROVED** [review]

File: core/Controller/TaskProcessingApiController.php [hunk 10]
```diff
@@ -396,9 +437,51 @@ public function cancelTask(int $taskId): DataResponse {
...
+	 * @param list<string> $taskTypeIds The ids of the task types
+	 * @return DataResponse<Http::STATUS_OK, array{task: CoreTaskProcessingTask, provider: array{name: string}}, array{}>|DataResponse<Http::STATUS_NO_CONTENT, null, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR, array{message: string}, array{}>
+	 *
+	 *  200: Task returned
+	 *  204: No task found
+	 */
+	#[ExAppRequired]
+	#[ApiRoute(verb: 'GET', url: '/tasks_provider/next', root: '/taskprocessing')]
+	public function getNextScheduledTask(array $providerIds, array $taskTypeIds): DataResponse {
+		try {
```
marcelklehr, line 460:
```suggestion
		try {
				// restrict $providerIds to providers that are configured as preferred for the passed task types
				$providerIds = array_values(array_intersect(array_unique(array_map(fn($taskTypeId) => $this->taskProcessingManager->getPreferredProvider($taskTypeId)->getId(), $taskTypeIds)), $providerIds));
				// restrict $taskTypeIds to task types that can actually be run by one of the now restricted providers
				$taskTypeIds = array_values(array_filter($taskTypeIds, fn($taskTypeId) => in_array($this->taskProcessingManager->getPreferredProvider($taskTypeId)->getId(), $providerIds, true));
				if (count($providerIds) === 0 || count($taskTypeIds) === 0) {
						return new DataResponse(null, Http::STATUS_NO_CONTENT);
				}
```

This will short circuit in cases where the passed providers are not configured as preferred for some of the passed task types.

## Example 1

```
configuration:
  summarize: llm2-llama3
  free_prompt: context_chat
```

Request:
```
GET /taskprocessing/tasks_provider/next

providerIds:
  - llm2-llama3
  - llm2-llama2
 taskTypeIds:
  - free_prompt
  - summarize
```

The endpoint will reduce the inputs to


```
$providerIds = [llm2-llama3]
$taskTypeIds = [summarize]
```

This will avoid unnecessary loops due to querying for free_prompt tasks.

## Example 2

```
configuration:
  summarize: llm2-llama3
  free_prompt:  llm2-llama3
```

Request:
```
GET /taskprocessing/tasks_provider/next

providerIds:
  - context_chat
 taskTypeIds:
  - free_prompt
```

The endpoint will reduce the inputs to

```
$providerIds = []
$taskTypeIds = []
```

This will avoid making a db request alltogether.

marcelklehr (approved):

core/Controller/TaskProcessingApiController.php
```diff
@@ -247,37 +255,72 @@ public function listTasks(?string $taskType, ?string $customId = null): DataResp
...
+	 * Returns the contents of a file referenced in a task(ExApp route version)
+	 *
+	 * @param int $taskId The id of the task
+	 * @param int $fileId The file id of the file to retrieve
+	 * @return DataDownloadResponse<Http::STATUS_OK, string, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND, array{message: string}, array{}>
+	 *
+	 *  200: File content returned
+	 *  404: Task or file not found
+	 */
+	#[ExAppRequired]
```
nickvergessen, line 276: This is now in "openapi-administration" while it's not accessible for administrators.
Should add `#[OpenAPI(scope: 'app_api')` or something a like?

marcelklehr: Mmh, @julien-nc are you using this endpoint in assistant?

provokateurin: I think fixing this in openapi-extractor is better, but a separate scope is also useful in addition. This can be done indepdently in openapi-extractor though.

provokateurin: @marcelklehr this endpoint can't not be used by assistant due to ExAppRequired, but the getFileContents endpoint can be used instead.

marcelklehr: Ah, yes, I was confused.

provokateurin: @nickvergessen I'll add a new scope to the OpenAPI attribute and then automatically move all endpoints to that scope that have the ExAppRequired attribute, just like we do with the administration scope.

nickvergessen: Makes sense, thanks

provokateurin:
Yeah I saw it, will fix all of it

julien-nc (approved):

lib/public/TaskProcessing/IManager.php
```diff
@@ -154,4 +163,13 @@ public function getUserTasksByApp(?string $userId, string $appId, ?string $custo
...
 	public function prepareInputData(Task $task): array;
+
+	/**
+	 * Changes the task status to STATUS_RUNNING and, if successful, returns True.
+	 *
+	 * @param Task $task
+	 * @return bool
+	 * @since 30.0.0
+	 */
+	public function lockTask(Task $task): bool;
```
nickvergessen, line 174: @provokateurin Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(all changes in lib/public/)

provokateurin: I think for this particular change it makes more sense to include it in https://github.com/nextcloud/server/pull/45094 because the whole documentation for that is missing in the first place.
I'll document the new attribute though.

PR #45394, opened 2024-05-17
refactor(files): Fix some ESLint and Typescript issues in FilesList

susnux:
## Summary

* Do not use names with `_` prefix on Vue instances (reserved by Vue)
* Make translation function calls more consistent by directly use the `t` function instead of calling it on the instance
* Fix typing of `this.promise`

## Checklist


apps/files/src/views/FilesList.vue
```diff
@@ -481,7 +499,7 @@ export default defineComponent({
 			}
 
 			// If we have a cancellable promise ongoing, cancel it
-			if (typeof this.promise?.cancel === 'function') {
+			if (this.promise && 'cancel' in this.promise && typeof this.promise.cancel === 'function') {
```
ShGKme, line 502: Why?

apps/files/src/views/FilesList.vue
```diff
@@ -515,8 +533,10 @@ export default defineComponent({
 
 				// Update paths store
 				const folders = contents.filter(node => node.type === 'folder')
-				folders.forEach(node => {
-					this.pathsStore.addPath({ service: currentView.id, fileid: node.fileid, path: join(dir, node.basename) })
+				folders.forEach((node) => {
+					if (node.fileid !== undefined) {
+						this.pathsStore.addPath({ service: currentView.id, fileid: node.fileid, path: join(dir, node.basename) })
+					}
 				})
```
ShGKme, line 540: Just to understand it better, in which case `fileid` is `undefined` here?

apps/files/src/views/FilesList.vue
susnux: `typeof this.promise?.cancel === 'function'` does not infer that `this.promise` is of type `CancelablePromise`.
You need to use `'cancel' in this.promise`, but this is invalid if `this.promise` is `undefined`.

Screenshot_20240522_161035

susnux: So I refactored this a bit (now) to make things more clear.

skjnldsv (approved):

ShGKme (approved):

PR #45396, opened 2024-05-17
fix(FileCache): mtime when not provided backend at zero

ze42:
* Resolves: #  (no related issue found)

## Summary

Some external backend always send 0 as modTime.

Creating a file without explicit timestamp coming from client (as in just creating it now, without having a modTime coming along the data), the file will be created in the backend, and nextcloud will read the modTime from the backend and store it as created in 1970...

[More details here](https://github.com/nextcloud/server/pull/45396#issuecomment-2118460687)

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Files/Cache/Cache.php
+++ b/lib/private/Files/Cache/Cache.php
@@ -278,6 +278,12 @@ class Cache implements ICache {
 	 * @throws \RuntimeException
 	 */
 	public function insert($file, array $data) {
+		// Some storage always return mtime at 0.
+		// Use now as time if nothing else is provided
+		if (array_key_exists('mtime', $data) && $data["mtime"] === 0) {
+			$data["mtime"] = time();
+		}
+
 		// normalize file
 		$file = $this->normalize($file);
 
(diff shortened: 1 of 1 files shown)


ze42:
> > Some external backend always send 0 as modTime.
> 
> Please elaborate.

I came accross an external storage backend not able to handle modtime, and always reporting `mtime = 0`.

When dumping a file that has a modtime, the modtime is stored in NextCloud, though the backend storage has zero.

When creating a file/directory without having a modtime coming along, the modtime is currently set to 0 (so, Jan 1st 1970), instead of having something like "now" used.

kesselb:
What storage backend / service is that, and how is it connected with Nextcloud?

I'm unsure if 0 = now is a sane default.
If there's known quirks with a storage backend, we usually try to fix it within the concrete implementation.

ze42:
> What storage backend / service is that, and how is it connected with Nextcloud?
> 
> I'm unsure if 0 = now is a sane default. If there's known quirks with a storage backend, we usually try to fix it within the concrete implementation.

We are trying to develop a backend using webdav and storing into IPFS.

IPFS is currently not able to store any date.

So, we have it as external storage on nextcloud, but without this patch, any files created there gets the date from 1970.


Would you have any alternative suggestion? Maybe a way to implement/configure per (external) storage a do-not-trust-dates?

kesselb:
> So, we have it as external storage on nextcloud, but without this patch, any files created there gets the date from 1970.

Because your backend sends 1970 as modified date ;)

> Would you have any alternative suggestion?

If your backend does not provide getlastmodified, why do you send it with 0 as value? 

Our code should already handle responses without getlastmodified[^1] properly.

[^1]: https://github.com/nextcloud/server/blob/d87c23242b96903d59b978a936179f5441c26773/lib/private/Files/Storage/DAV.php#L606

kesselb:
Hi @ze42, is this resolved?

PR #45407, opened 2024-05-20
fix(profile): Avoid updating the config at each call

solracsf:
## Summary

Updating the config at each call can be inefficient. Instead, check if the merged configuration is different from the existing one before performing the update. This optimization may reduce unnecessary database operations.

## Checklist

### DIFF
--- a/lib/private/Profile/ProfileManager.php
+++ b/lib/private/Profile/ProfileManager.php
@@ -341,12 +341,17 @@ class ProfileManager implements IProfileManager {
 				$config = $this->configMapper->get($targetUser->getUID());
 				$this->configCache[$targetUser->getUID()] = $config;
 			}
+
 			// Merge defaults with the existing config in case the defaults are missing
-			$config->setConfigArray(array_merge(
-				$defaultProfileConfig,
-				$this->filterNotStoredProfileConfig($config->getConfigArray()),
-			));
-			$this->configMapper->update($config);
+			$existingConfigArray = $this->filterNotStoredProfileConfig($config->getConfigArray());
+			$mergedConfigArray = array_merge($defaultProfileConfig, $existingConfigArray);
+
+			// Only update if the merged config is different from the existing config
+			if ($mergedConfigArray !== $existingConfigArray) {
+				$config->setConfigArray($mergedConfigArray);
+				$this->configMapper->update($config);
+			}
+
 			$configArray = $config->getConfigArray();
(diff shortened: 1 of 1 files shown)


Pytal (approved):

lib/private/Profile/ProfileManager.php
```diff
@@ -341,12 +341,17 @@ public function getProfileConfig(IUser $targetUser, ?IUser $visitingUser): array
...
-			$config->setConfigArray(array_merge(
-				$defaultProfileConfig,
-				$this->filterNotStoredProfileConfig($config->getConfigArray()),
-			));
-			$this->configMapper->update($config);
+			$existingConfigArray = $this->filterNotStoredProfileConfig($config->getConfigArray());
+			$mergedConfigArray = array_merge($defaultProfileConfig, $existingConfigArray);
+
+			// Only update if the merged config is different from the existing config
+			if ($mergedConfigArray !== $existingConfigArray) {
```
ChristophWurst, line 350: had to check if php considers two array identical but seems to be the case :+1: https://3v4l.org/oGKFE

lib/private/Profile/ProfileManager.php
```diff
@@ -341,12 +341,17 @@ public function getProfileConfig(IUser $targetUser, ?IUser $visitingUser): array
 				$config = $this->configMapper->get($targetUser->getUID());
 				$this->configCache[$targetUser->getUID()] = $config;
 			}
+
 			// Merge defaults with the existing config in case the defaults are missing
-			$config->setConfigArray(array_merge(
-				$defaultProfileConfig,
-				$this->filterNotStoredProfileConfig($config->getConfigArray()),
-			));
-			$this->configMapper->update($config);
```
ChristophWurst, line 349: The magic setter of an entity should already skip assignments when the value does not change: https://github.com/nextcloud/server/blob/474cbda59a4c5c4a225918724a871451a0f49a7f/lib/public/AppFramework/Db/Entity.php#L108-L110. The mapper skips the UPDATE query if there are no updated fields: https://github.com/nextcloud/server/blob/474cbda59a4c5c4a225918724a871451a0f49a7f/lib/public/AppFramework/Db/QBMapper.php#L184-L187.

Hence I wonder if the UPDATE is actually run before your changes. And if so, why. Possible causes
1. The arrays are not identical, or the json representation isnt.
2. Other fields of the config are assigned.

Could you check?

PR #45413, opened 2024-05-21
fix(files): Don't fail if dir exists

solracsf:
## Summary

In some cases, `mkdir` is returning `false` because directory already exists on disk (_race condition?_)
Nevertheless, we shouldn't need to `throw` if folder exists.

```
Exception":"OCP\\Files\\NotPermittedException","Message":"Could not create folder \"/appdata_ocwfs52oscxg/preview/5/e/8/b/b/c/e/3685938\
```

## Checklist

### DIFF
--- a/lib/private/Files/Node/Folder.php
+++ b/lib/private/Files/Node/Folder.php
@@ -133,7 +133,7 @@ class Folder extends Node implements \OCP\Files\Folder {
 			$fullPath = $this->getFullPath($path);
 			$nonExisting = new NonExistingFolder($this->root, $this->view, $fullPath);
 			$this->sendHooks(['preWrite', 'preCreate'], [$nonExisting]);
-			if (!$this->view->is_dir($fullPath) && !$this->view->mkdir($fullPath)) {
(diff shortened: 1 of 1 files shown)


lib/private/Files/Node/Folder.php
```diff
@@ -158,7 +158,7 @@ public function newFolder($path) {
 			$fullPath = $this->getFullPath($path);
 			$nonExisting = new NonExistingFolder($this->root, $this->view, $fullPath);
 			$this->sendHooks(['preWrite', 'preCreate'], [$nonExisting]);
-			if (!$this->view->mkdir($fullPath)) {
+			if (!$this->view->is_dir($fullPath) && !$this->view->mkdir($fullPath)) {
```
Altahrim, line 161: If it's a race condition, I think we should try to create the directory first and check if it exists after
```suggestion
			if (!$this->view->mkdir($fullPath) && !$this->view->is_dir($fullPath)) {
```

Altahrim (approved):

artonge (approved):

Issue #39551, opened 2023-07-24
[Bug]: Icons use dark mode color on light mode after upgrade NC27.01


grrdvnk:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

I've specified a pretty dark theme color (#002832), which had issues on dark mode on NC 26 and below (fixed in [pr 38159).](https://github.com/nextcloud/server/pull/38159).

Since I upgraded to NC 27.01, icons do not follow the specified theme color any longer in light mode. Instead it always uses the dark mode color (generated by pr 38159). I first noticed the Nextcloud iOS app was off, but is is also present in the web interface. Setting a different (not so dark) theme color does update the icon colors.

### Steps to reproduce

1. Set theme color to HEX 002832
2. Set interface to light mode
3. See icon colors are off:

<img width="605" alt="Screenshot 2023-07-24 at 11 23 53" src="https://github.com/nextcloud/server/assets/20874876/69959239-50ba-4497-bf7e-c478c46361ad">

### Expected behavior

Icon colors should follow theme color as specified when using light mode

### Installation method

Community Web installer on a VPS or web space

### Nextcloud Server version

27

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Upgraded to a MAJOR version (ex. 22 to 23)

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

szaimen:
[...]

Issue #4352, opened 2017-04-14
Password reset button broken


typorian:
### Steps to reproduce
1. Generate password reset link
2. click on link in the Mail
3. Type in new password
4. Press "reset" button

### Expected behaviour
Password should be reset

### Actual behaviour
Nothing

### Server configuration

**Operating system**:
debian 8.7 jessie
**Web server:**
apache2
**Database:**
mysql  Ver 14.14 Distrib 5.5.54, for debian-linux-gnu (x86_64) using readline 6.3
**PHP version:**
php 7.0
**Nextcloud version:** (see Nextcloud admin page)
11.02 (stable)
**Updated from an older Nextcloud/ownCloud or fresh install:**
updated from previous stable
**Where did you install Nextcloud from:**
debian image with nextcloud provided by hosting provider
**Signing status:**

<details>
<summary>Signing status</summary>

```
Login as admin user into your Nextcloud and access 
http://example.com/index.php/settings/integrity/failed 
paste the results here.
No errors have been found.
```
</details>

**List of activated apps:**
<details>
<summary>activity, gallery, contacts, calendar, gpxpod, gpxedit, tasks, collabora CODE</summary>

Enabled:
  - activity: 2.4.1
  - admin_audit: 1.1.0
  - admin_notifications: 1.0.0
  - calendar: 1.5.2
  - comments: 1.1.0
  - contacts: 1.5.3
  - dav: 1.1.1
  - federatedfilesharing: 1.1.1
  - federation: 1.1.1
  - files: 1.6.1
  - files_downloadactivity: 1.0.1
  - files_pdfviewer: 1.0.1
  - files_sharing: 1.1.1
  - files_texteditor: 2.2
  - files_trashbin: 1.1.0
  - files_versions: 1.4.0
  - files_videoplayer: 1.0.0
  - firstrunwizard: 2.0
  - gallery: 16.0.0
  - gpxedit: 0.0.5
  - gpxpod: 2.1.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.0.0
  - nextcloud_announcements: 1.0
  - notifications: 1.0.1
  - password_policy: 1.1.0
  - provisioning_api: 1.1.0
  - richdocuments: 1.1.25
  - serverinfo: 1.1.1
  - sharebymail: 1.0.1
  - survey_client: 0.1.5
  - systemtags: 1.1.3
  - tasks: 0.9.5
  - templateeditor: 0.2
  - theming: 1.1.1
[...]

PR #45422, opened 2024-05-21
[stable27] fix(theming): stable 27 disable accessible color config switch

skjnldsv:
Fix https://github.com/nextcloud/server/issues/39551

Only for 27! 28 and above have a different handling of themes and is fixed

## Apply this fix

1. add `'27-disable-accessible-theming-color' => true,` to your `config.php`
2. delete the `global` folder on your `data/appdata_xxxx/theming` directory (no need to delete the theming folder)
3. run an `occ files:scan-app-data theming`

## How to reproduce the issue

1. use a non accessible color like `#004352`
2. load the files app

| Before | After |
|:---------:|:------:|
|localhost_8081_index php_apps_files__dir=_ fileid=2 (1)|localhost_8081_index php_apps_files__dir=_ fileid=2|

### DIFF
--- a/apps/theming/lib/Util.php
+++ b/apps/theming/lib/Util.php
@@ -82,6 +82,11 @@ class Util {
 	 * @return string
 	 */
 	public function elementColor($color, ?bool $brightBackground = null) {
+		// Disable color luminance check if the feature is disabled. 27 ONLY!
+		if ($this->config->getSystemValueBool('27-disable-accessible-theming-color') === true) {
+			return $color;
+		}
+
 		$luminance = $this->calculateLuminance($color);
 
 		if ($brightBackground !== false && $luminance > 0.8) {


szaimen:
> 28 and above have a different handling of themes and is fixed

I fear it is not, seems to behave completely the same, just tested...

artonge (approved):

artonge:
> folder color (grey):

But that is expected, no? Black on black does not work for accessibility.

szaimen:
> > folder color (grey):
> 
> But that is expected, no? Black on black does not work for accessibility.

yes, this is also what I am understanding but seems like the customer complains about, no?

AndyScherzinger (approved):

szaimen:
> > yes, this is also what I am understanding but seems like the customer complains about exactly this, no?
> 
> No the complaint is the before image which shows the folder color used for dark _and_ light

I see. 👍

AndyScherzinger:
> We might have to check with the customer again that they're ok with it

Yes, we might. The main issue is with the light theme though. The light folder icons in your before screenshot are also used in light mode on 27...

So the main issue to be addressed with this fix is:
2024-05-21 16_43_47-Files - Cloud – Mozilla Firefox

AndyScherzinger:
@skjnldsv how woudl the light theme look like in 28 with that color? That is the main concern with the given color code

skjnldsv:
> @skjnldsv how would the light theme look like in 28 with that color? That is the main concern with the given color code

image
Because we fixed the computation on the client side, so we now which theme is active :+1:

AndyScherzinger:
Sweet, so v28 looks good to me 👍 
Can't tell if this works fine with they backgroudn image as already mentioned by @artonge to me.

PR #45430, opened 2024-05-21
[stable29] Fix header menu icon color

susnux:
Now that #42977 is merged we still need to fix stable29 and stable28 branches.
We can not backport but we can add a (maybe hacky) work around.

This overwrites the primary text color with a color based on the background not the primary color.

### Screenshot

before | after
---|---
Screenshot 2024-05-21 at 22-10-22 Appearance and accessibility - Nextcloud|Screenshot 2024-05-21 at 22-10-47 Appearance and accessibility - Nextcloud


szaimen (approved):
Tested and seems to work :)

Except for custom white background pictures:
image

Pytal (approved):

bigcat88 (approved):

PR #45431, opened 2024-05-21
fix(files): Add proper visual loading feedback for image preview

susnux:
## Summary

Currently only a white square is shown while loading the icon, this add a loading indicator.
(Especially the previous one had a short flash when both the background and icon are shown at the same time).

#### Before

https://github.com/nextcloud/server/assets/1855448/59344023-e738-4810-9a6f-7e43c105e5ec

#### After

https://github.com/nextcloud/server/assets/1855448/65329e3b-d342-4778-92a8-8fb7e129440e

## Checklist

### DIFF
--- a/apps/files/src/components/FileEntry/FileEntryPreview.vue
+++ b/apps/files/src/components/FileEntry/FileEntryPreview.vue
@@ -31,16 +31,13 @@
 			</template>
 		</template>
 
+		<NcLoadingIcon v-else-if="!error && !loaded" />
+
 		<!-- Decorative image, should not be aria documented -->
-		<img v-else-if="previewUrl && backgroundFailed !== true"
-			ref="previewImg"
+		<img v-else-if="previewUrl && loaded"
 			alt=""
 			class="files-list__row-icon-preview"
-			:class="{'files-list__row-icon-preview--loaded': backgroundFailed === false}"
-			loading="lazy"
-			:src="previewUrl"
-			@error="onBackgroundError"
-			@load="backgroundFailed = false">
+			:src="previewUrl">
 
 		<FileIcon v-else v-once />
 
@@ -65,6 +62,7 @@ import { translate as t } from '@nextcloud/l10n'
 import { Type as ShareType } from '@nextcloud/sharing'
 
 import Vue from 'vue'
+import NcLoadingIcon from '@nextcloud/vue/dist/Components/NcLoadingIcon.js'
 import AccountGroupIcon from 'vue-material-design-icons/AccountGroup.vue'
 import AccountPlusIcon from 'vue-material-design-icons/AccountPlus.vue'
 import FileIcon from 'vue-material-design-icons/File.vue'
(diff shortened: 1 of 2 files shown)


skjnldsv:
> Currently only a white square is shown while loading the icon, this add a loading indicator.
> (Especially the previous one had a short flash when both the background and icon are shown at the same time).

This is supposed to be a placeholder.
Somehow it's disabled https://github.com/nextcloud/server/blob/e210caf4cfb71195a53181cb53bab3fdcf893617/apps/files/src/components/FilesListVirtual.vue#L498

Imho, loading spinners are good for individual element loading.
For a lot of them (like here, in a list), it's distracting and a loading placeholder is what should be used (we do it in many other places)

skjnldsv:
Peek 22-05-2024 08-33

How t was supposed to look like

marcoambrosini (commented):
Out of the two, I think that the loading spinner looks nicer and is just as clear

susnux:
Makes sense, like a loading skeleton.

So I am fine with whatever the @nextcloud/designers prefer :)

skjnldsv:
> Out of the two, I think that the loading spinner looks nicer and is just as clear

For the record, my point is not about if it's clearer or not, but the fact that 20 spinners shown drag too much attention for very little value (it's just a preview) :)

Anyway, as Ferdinand said: I am fine with whatever the @nextcloud/designers prefer :wink:

marcoambrosini:
The gif you posted @skjnldsv  seem to draw more attention than the little spinners. But I'm fine either way tbh :)

nimishavijay:
> For a lot of them (like here, in a list), it's distracting and a loading placeholder is what should be used (we do it in many other places)

I would agree here, the square being distracting is rather a result of the color of the skeleton screen component being too light in dark mode IMO, so darkening the color of the squares would already make it look less busy. What do you think?

skjnldsv:
> darkening the color of the squares would already make it look less busy. What do you think?

Seconded :+1:

jancborchardt (changes requested):
Looks nice with the skeleton squares, and agree with the feedback of adjusting the color a bit (lightening/darkening).

An additional thing: When it’s a file which does not have a preview, like a PDF for example – why don’t we just instantly show the PDF filetype icon?

skjnldsv:
> An additional thing: When it’s a file which does not have a preview, like a PDF for example – why don’t we just instantly show the PDF filetype icon?

Because we don't always know if there is a preview. It can be available but not generated.
There is a ticket somewhere to have a property saying if the preview for thie mime is `enabled` or not, that's a different approach

jancborchardt:
> Because we don't always know if there is a preview. It can be available but not generated.

Right, so what I’m saying is that we can show the filetype icon instantly. Preview could be shown as soon as available. That makes things feel more instant instead of having loading bars/skeletons here.

jancborchardt (commented):
@ornanovitch you requested review – my review from the 2 comments at https://github.com/nextcloud/server/pull/45431#pullrequestreview-2116058855 still stands. :) 

@susnux let me know if that sounds good :)

ornanovitch:
> @ornanovitch you requested review – my review from the 2 comments at [#45431 (review)](https://github.com/nextcloud/server/pull/45431#pullrequestreview-2116058855) still stands. :)

Sorry I didn't see! :)

PR #45435, opened 2024-05-22
feat(dav): Add an API for upcoming events

ChristophWurst:
* For https://github.com/nextcloud/calendar/issues/5909

## Summary

Expose a convenience endpoint to query upcoming events. Events can be filtered by location to match Talk conversation URLs.

## Demo / how to test

```
$ curl -u admin:admin -X GET 'https://localhost/ocs/v1.php/apps/dav/api/v1/events/upcoming?location=https%3A%2F%2Flocalhost%2Fcall%2Fqkcxuuge' -H "OCS-APIRequest: true"
<?xml version="1.0"?>
<ocs>
 <meta>
  <status>ok</status>
  <statuscode>100</statuscode>
  <message>OK</message>
  <totalitems></totalitems>
  <itemsperpage></itemsperpage>
 </meta>
 <data>
  <events>
   <element>
    <uri>133A0EF2-15FA-46C3-9BB6-FA3629A94437.ics</uri>
    <recurrenceId/>
    <calendarUri>personal</calendarUri>
    <start>1723646700</start>
    <summary>Y</summary>
    <location>https://localhost/call/qkcxuuge</location>
    <calendarAppUrl>https://localhost/apps/calendar/</calendarAppUrl>
   </element>
  </events>
 </data>
</ocs>
```

## TODO

- [x] PoC
- [x] Possible timezone handling
- [x] Code clean-up
- [x] Tests
- [x] API docs

## Checklist


ChristophWurst:
@DorraJaouad could you please check the API and tell me if there is any more info we need to expose?

DorraJaouad:
In order to generate the url to calendar app with event details open in sidebar, we need to have start timestamp and calendar id. However, the calendar id in url doesn't seem to match the one provided in this API point when testing. so is it missing ?

ChristophWurst:
The API exposes the numeric (internal) calendar ID. The calendar app uses the calendar URI. I will change the API to expose the URI here as well :+1:

apps/dav/lib/CalDAV/UpcomingEventsService.php
```diff
@@ -0,0 +1,62 @@
...
+
+	public function getEvents(string $userId, string $location = null): array {
+		// TODO: move logic into a service class
+		$searchQuery = $this->calendarManager->newQuery('principals/users/' . $userId);
+		if ($location !== null) {
+			$searchQuery->addSearchProperty('LOCATION');
+			$searchQuery->setSearchPattern($location);
+		}
+		$searchQuery->addType('VEVENT');
+		$searchQuery->setLimit(5);
```
DorraJaouad, line 35: Maybe closest one is enough or limit could be defined in request? Talk UI will use only the first event coming up.

ChristophWurst: I've switched to 3. It's less overfetching and still a nice number should we consider showing more (IMO it makes sense because you might have more than one meeting a week in a room)

ChristophWurst:
@DorraJaouad is the API usable from your perspective?

DorraJaouad:
so this is the response so far
```
            "events": [
                {
                    "uri": "3F1F3474-0B83-4DA0-8E75-A10F5350A16D.ics",
                    "recurrenceId": null,
                    "calendarUri": "personal",
                    "start": 1723680000,
                    "summary": "Party",
                    "location": "https:\/\/nextcloud.local\/index.php\/call\/66kq5sfv",
                    "calendarAppUrl": "https:\/\/nextcloud.local\/index.php\/apps\/calendar\/"
                }
```

I believe recurrenceId is the `the calendar URI` (to be used in url) ?

apps/dav/lib/Controller/UpcomingEventsController.php
```diff
@@ -0,0 +1,62 @@
...
+	public function getEvents(?string $location = null): DataResponse {
+		if ($this->userId === null) {
+			return new DataResponse(null, Http::STATUS_UNAUTHORIZED);
+		}
+
+		return new DataResponse([
+			'events' => array_map(fn (UpcomingEvent $e) => $e->jsonSerialize(), $this->service->getEvents(
+				$this->userId,
+				$location,
+			)),
```
miaulalala, line 58: smoof :heart_eyes:

miaulalala (approved):

ChristophWurst:
Another force push for cs fixes

SebastianKrupinski (approved):

skjnldsv (approved):

SebastianKrupinski:
Tested.

Case 1: Blank Location
Result:  Returns events from personal and shared calendar

Case 2: Location matching a location in a event
Result: Returns events for specific location from personal and shared calendars

Case 2: Location partially matching the location of another event
Result: Returns any events that have similar location string (this is probably unwanted behavior)

image

PR #45455, opened 2024-05-22
fix(INavigationManager): Fix psalm alias not set in the right place

provokateurin:
## Summary

https://github.com/nextcloud/server/pull/41341 didn't set the type alias correctly as it needs to be on the same class to be "visible" (PHPStorm also displays that it can't resolve the type on the add method).

While looking at that I also found that the type for the endpoint is wrong and was missing some fields as well (guess who did that :sweat_smile:).

## Checklist

### DIFF
--- a/core/ResponseDefinitions.php
+++ b/core/ResponseDefinitions.php
@@ -26,11 +26,13 @@ namespace OCA\Core;
  *
  * @psalm-type CoreNavigationEntry = array{
  *     id: string,
- *     order: int,
+ *     order?: int,
  *     href: string,
  *     icon: string,
  *     type: string,
  *     name: string,
+ *     app?: string,
+ *     default?: bool,
  *     active: bool,
  *     classes: string,
  *     unread: int,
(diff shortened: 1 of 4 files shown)


core/ResponseDefinitions.php
```diff
@@ -43,11 +43,13 @@
  *
  * @psalm-type CoreNavigationEntry = array{
  *     id: string,
- *     order: int|string,
```
provokateurin, line 46: In reality this seems to be true where the order can be int or string, but this is terrible API design. I'd suggest we just cast to int everywhere as the strings are also just ints encoded as strings.

come-nc: Depends how many places in the code "everywhere" means :smile:

provokateurin: I think it would only be in the add method of NavigationManager

come-nc: Fine with me then.

susnux (approved):

come-nc (approved):

PR #45475, opened 2024-05-23
feat: Add support for webhook listeners

come-nc:
## Summary

Add support for registering webhooks as listeners of events.

## Checklist


ChristophWurst (commented):
Code looks good! How will this be used? Can we assume that apps know their webhook listeners when the app is just *registered* and the system isn't fully booted yet?

come-nc:
> Code looks good! How will this be used?

Plan is to use that for integration of an external workflow engine, but architecture is not final yet.
One problem is we’d prefer if the webhook gets called only on certains conditions, but then we end up with `apps/workflowengine` complexity.

> Can we assume that apps know their webhook listeners when the app is just _registered_ and the system isn't fully booted yet?

Hum, what is not accessible at that point?
If I want to have an app which search in its database and user settings to know which webhooks to register, will that work?

ChristophWurst:
> > Can we assume that apps know their webhook listeners when the app is just _registered_ and the system isn't fully booted yet?
> 
> Hum, what is not accessible at that point? If I want to have an app which search in its database and user settings to know which webhooks to register, will that work?

That works. Anything that is provided by other apps might be missing. Say information from a user backend.

If you always want to wire a specific event to  webhook listeners it's fine. But the lack of this information could be critical when you need to conditionally register webhooks. I haven't yet understood the full picture of the new automations so this might not be necessary.

https://docs.nextcloud.com/server/latest/developer_manual/app_development/bootstrap.html#nextcloud-20-and-later for my original ideas and thoughts about the two step register/boot process.

come-nc:
Webhooks are moved to an application with registrations in DB instead. They will also need to support filtering of events triggerring the webhook.

- [ ] Skeleton app
- [ ] API to register webhooks
- [ ] occ command to list them
- [ ] Actually call the webhooks, through background jobs
- [ ] Serialize the event to the webhook
- [ ] Add a filtering system

come-nc:
- [ ] Store app id from appapi header if available

apps/webhooks/lib/ResponseDefinitions.php
```diff
@@ -0,0 +1,26 @@
...
+namespace OCA\Webhooks;
+
+/**
+ * @psalm-type WebhooksListenerInfo = array{
+ *     id: string,
+ *     userId: string,
+ *     httpMethod: string,
+ *     uri: string,
+ *     event?: string,
+ *     eventFilter?: array<mixed>,
```
marcelklehr, line 19: Is this still failing?

come-nc:
Follow-up:

- [ ] Extract user filter from eventFilter so that the listener is not even registered depending on logged in user
- [ ] Same for group membership

apps/webhooks/lib/ResponseDefinitions.php
come-nc: Changed it to `array<string,mixed>`. It’s not exactly right because you can have an array of filters with integer keys I think, but it’s close enough and psalm type system does not allow us to define nested arrays correctly.

apps/webhooks/lib/ResponseDefinitions.php
```diff
@@ -16,7 +16,7 @@
  *     httpMethod: string,
  *     uri: string,
  *     event?: string,
- *     eventFilter?: array<mixed>,
+ *     eventFilter?: array<string,mixed>,
```
marcelklehr, line 19: It might be possible to define a recursive type, to avoid the "mixed" here, but I'm not sure

come-nc: https://github.com/vimeo/psalm/issues/5739 nope

apps/webhooks/lib/Controller/WebhooksController.php
```diff
@@ -0,0 +1,218 @@
...
+	#[ApiRoute(verb: 'POST', url: '/api/v1/webhooks')]
+	#[AuthorizedAdminSetting(settings:'OCA\Webhooks\Settings\Admin')]
+	public function create(
+		string $httpMethod,
+		string $uri,
+		string $event,
+		?array $eventFilter,
+		?array $headers,
+		?string $authMethod,
+		?array $authData,
```
julien-nc, line 101: Are we sure that authData will always be an array? Even with other future auth methods?

apps/webhooks/lib/Db/WebhookListener.php
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	public function getAuthDataClear(): array {
+		if ($this->authData === null) {
+			return [];
+		}
+		return json_decode($this->crypto->decrypt($this->getAuthData()), associative:true, flags:JSON_THROW_ON_ERROR);
+	}
+
+	public function setAuthDataClear(?array $data): void {
```
julien-nc, line 78: Nitpicking: I think it's enough to call those methods `getAuthData` and `setAuthData`. I feel like current names are ambiguous. Or maybe `getDecryptedAuthData` and `setEncryptedAuthData`. Not a big deal anyway.

apps/webhooks/lib/Db/WebhookListenerMapper.php
```diff
@@ -0,0 +1,159 @@
...
+	public function getAll(): array {
+		$qb = $this->db->getQueryBuilder();
+
+		$qb->select('*')
+			->from($this->getTableName());
+
+		return $this->findEntities($qb);
+	}
+
+	public function addWebhookListener(
```
julien-nc, line 59: Missing some method return types and PhpDoc blocks (useful for `@throws` IMO).

apps/webhooks/lib/Listener/WebhooksEventListener.php
```diff
@@ -0,0 +1,90 @@
...
+				);
+			}
+		}
+	}
+
+	private function serializeEvent(Event $event): array|\JsonSerializable {
+		if ($event instanceof \JsonSerializable) {
+			return $event;
+		} else {
+			/* Event is not serializable, we fallback to reflection to still send something */
```
julien-nc, line 63: Noice!

apps/webhooks/lib/Migration/Version1000Date20240527153425.php
```diff
@@ -0,0 +1,72 @@
...
+	/**
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		if (!$schema->hasTable(WebhookListenerMapper::TABLE_NAME)) {
+			$table = $schema->createTable(WebhookListenerMapper::TABLE_NAME);
+			$table->addColumn('id', Types::INTEGER, [
```
julien-nc, line 29:
```suggestion
			$table->addColumn('id', Types::BIGINT, [
```

apps/webhooks/lib/Service/PHPMongoQuery.php
```diff
@@ -0,0 +1,340 @@
...
+	 * @param array $query		A boolean value or an array defining a query
+	 * @param array $documents	The document to query
+	 * @param array $options	Any options:
+	 *	'debug' - boolean - debug mode, verbose logging
+	 *	'logger' - \Psr\LoggerInterface - A logger instance that implements {@link https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-3-logger-interface.md#3-psrlogloggerinterface PSR-3}
+	 *	'unknownOperatorCallback' - a callback to be called if an operator can't be found.  The function definition is function($operator, $operatorValue, $field, $document). return true or false.
+	 * @throws Exception
+	 */
+	public static function find(array $query, array $documents, array $options = []): array {
+		if(empty($documents) || empty($query)) {
```
julien-nc, line 31: php-cs will probably complain about the syntax. Should we avoid checking this file or fix the code style?

julien-nc (changes requested):
Works great! A few minor comments and questions.

apps/webhooks/lib/AppInfo/Application.php
```diff
@@ -0,0 +1,55 @@
...
+	): void {
+		/** @var WebhookListenerMapper */
+		$mapper = $container->get(WebhookListenerMapper::class);
+
+		/* Listen to all events with at least one webhook configured */
+		$configuredEvents = $mapper->getAllConfiguredEvents();
+		foreach ($configuredEvents as $eventName) {
+			// $logger->error($eventName.' '.\OCP\Files\Events\Node\NodeWrittenEvent::class, ['exception' => new \Exception('coucou')]);
+			$dispatcher->addServiceListener(
+				$eventName,
```
nickvergessen, line 49: Should we block certain event names and allow apps to defined blocked events?
E.g. login events should not send the users password to other servers.

apps/webhooks/lib/Migration/Version1000Date20240527153425.php
```diff
@@ -0,0 +1,72 @@
...
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->addColumn('http_method', Types::STRING, [
+				'notnull' => true,
+				'length' => 32,
+			]);
+			$table->addColumn('uri', Types::STRING, [
+				'notnull' => true,
+				'length' => 256,
```
nickvergessen, line 48: In talk we used 4k length for the url. Not sure 256 might be enough in all cases

apps/webhooks/appinfo/info.xml
```diff
@@ -0,0 +1,33 @@
...
+	<website>https://github.com/nextcloud/server</website>
+	<bugs>https://github.com/nextcloud/server/issues</bugs>
+	<repository>https://github.com/nextcloud/server.git</repository>
+
+	<dependencies>
+		<nextcloud min-version="30" max-version="30"/>
+	</dependencies>
+
+	<commands>
+		<command>OCA\Webhooks\Command\Index</command>
```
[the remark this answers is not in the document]
come-nc: `ListCommand`, like `user:list`?

apps/webhooks/lib/Db/WebhookListener.php
come-nc: It’s not possible, `getAuthData` and `setAuthData` are the magical setter/getter for the DB column. (And they are used by the mapper so we cannot make them encrypt/decrypt).
If we want to use those names I need to rename the DB column to `auth_data_encrypted` or something. Do you want me to do that? I’m fine with both approach.

apps/webhooks/appinfo/info.xml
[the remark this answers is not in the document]
julien-nc: Yes, `ListCommand` is more explicit, I think `ListWebhooks` matches more with `webhooks:list`.

apps/webhooks/lib/Service/PHPMongoQuery.php
come-nc: https://github.com/nextcloud/server/actions/runs/9367363495/job/25786877013?pr=45475
`php-cs` is happy, I ran `cs:fix` on the file already.

apps/webhooks/lib/BackgroundJobs/WebhookCall.php
```diff
@@ -0,0 +1,56 @@
...
+				case AuthMethod::None:
+					break;
+				case AuthMethod::Header:
+					$authHeaders = $webhookListener->getAuthDataClear();
+					$options['headers'] = array_merge($options['headers'], $authHeaders);
+					break;
+			}
+			$response = $client->request($webhookListener->getHttpMethod(), $webhookListener->getUri(), $options);
+			$statusCode = $response->getStatusCode();
+			if ($statusCode >= 200 && $statusCode < 300) {
```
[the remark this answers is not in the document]
come-nc: Yeah I messed it up, inverting the branch content, nice catch.

apps/webhooks/lib/Controller/WebhooksController.php
```diff
@@ -0,0 +1,218 @@
...
+	#[AuthorizedAdminSetting(settings:'OCA\Webhooks\Settings\Admin')]
+	public function destroy(int $id): DataResponse {
+		try {
+			$deleted = $this->mapper->deleteById($id);
+			return new DataResponse($deleted);
+		} catch (\UnexpectedValueException $e) {
+			throw new OCSBadRequestException($e->getMessage(), $e);
+		} catch (\DomainException $e) {
+			throw new OCSForbiddenException($e->getMessage(), $e);
+		} catch (Exception $e) {
```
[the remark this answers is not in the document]
come-nc: Was supposed to be `\Exception` like the other methods, fixing it.

apps/webhooks/lib/AppInfo/Application.php
julien-nc: Nice catch, yes, all authentication-related events should be "blacklisted". It's difficult to list all those dispatched by community apps but those should be enough:
* all auth events from core
* user_ldap
* user_saml
* user_oidc

@nickvergessen Do you think of any other kind of event that should be avoided?

apps/webhooks/lib/Db/WebhookListener.php
julien-nc: Oh right, no let's not rename the DB columns. Maybe we can find more explicit names. Maybe current names are fine... :grin:

provokateurin:
I'm unsubscribing from this PR for now, please ping me if you need a review of the (Open)API.

apps/webhooks/lib/AppInfo/Application.php
```diff
@@ -0,0 +1,55 @@
...
+	private function registerRuleListeners(
+		IEventDispatcher $dispatcher,
+		ContainerInterface $container,
+		LoggerInterface $logger,
+	): void {
+		/** @var WebhookListenerMapper */
+		$mapper = $container->get(WebhookListenerMapper::class);
+
+		/* Listen to all events with at least one webhook configured */
+		$configuredEvents = $mapper->getAllConfiguredEvents();
```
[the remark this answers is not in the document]
come-nc: Yes, see https://github.com/nextcloud/server/blob/9fc8959d9d236667c4761f11d7e935c22a374f0b/apps/webhooks/lib/Db/WebhookListenerMapper.php#L134
For the cache to make sense it needs to be in session, should I add that in the mapper or in the `Application` class?

apps/webhooks/lib/AppInfo/Application.php
come-nc: That’s a good point.
Then I think we should remove the fallback and only allow `\JsonSerializable` implementors to be listened to.
Then an application can decide to not implement it to disallow webhooks or to only serialize some properties (ie not the password for PostLoginEvent). @julien-nc @nickvergessen what do you think?

nickvergessen: > or to only serialize some properties (ie not the password for PostLoginEvent).

Sounds good.

apps/webhooks/lib/AppInfo/Application.php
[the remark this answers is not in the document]
nickvergessen: > For the cache to make sense it needs to be in session, should I add that in the mapper or in the Application class?

I prefer to keep the mapper plain, to know when a SQL query is fired.
If you use this in more places in the future you could move it to a service file.
So for now I would put it into Application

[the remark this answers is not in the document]
come-nc: See https://github.com/nextcloud/server/pull/45475/commits/57c5ddbd207fa8b7565f49a6e1ca2fa85814603d , but if this caching was done in the mapper instead, I could reset the cache whenever a webhook is inserted or updated, what do you think?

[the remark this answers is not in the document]
nickvergessen: so time for a service? :P 
Or the mapper is the service.
Could also have a private uncached function there which is called by the "public" cached function.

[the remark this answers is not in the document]
come-nc: I prefer if the mapper is the service, splitting this in two classes sounds overkill to me, + it weakens the idea that all db access should go through these functions. (what I mean is that if the caching is in the service it might miss calls directly on the mapper).

The Application class cannot use DI, so moving the cache to the mapper. Incomming.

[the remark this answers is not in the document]
come-nc: https://github.com/nextcloud/server/pull/45475/commits/0cbb36e232ddb91950f226cd2ea35ccf46b3c7c7

julien-nc (approved):

apps/webhooks/lib/BackgroundJobs/WebhookCall.php
```diff
@@ -0,0 +1,56 @@
...
+		ITimeFactory $timeFactory,
+	) {
+		parent::__construct($timeFactory);
+	}
+
+	protected function run($argument): void {
+		[$data, $webhookId] = $argument;
+		$webhookListener = $this->mapper->getById($webhookId);
+		$client = $this->clientService->newClient();
+		$options = [];
```
nickvergessen, line 33: might make sense to add verify with our certificate manager as well:
```
'verify' => $this->certificateManager->getAbsoluteBundlePath(),
```

Just learned about this today, but it then allows to easily also set up requests against self-signed servers by using the `occ security:certificate:import` command

apps/webhooks/lib/Db/WebhookListenerMapper.php
```diff
@@ -0,0 +1,201 @@
...
+		?string $appId,
+		string $userId,
+		string $httpMethod,
+		string $uri,
+		string $event,
+		?array $eventFilter,
+		?array $headers,
+		AuthMethod $authMethod,
+		?array $authData,
```
nickvergessen, line 82: Should we start adding:
```suggestion
		#[\SensitiveParameter]
		?array $authData,
```
for new things at least?

apps/webhooks/lib/Migration/Version1000Date20240527153425.php
```diff
@@ -0,0 +1,72 @@
...
+				'notnull' => true,
+				'length' => 64,
+			]);
+			$table->addColumn('http_method', Types::STRING, [
+				'notnull' => true,
+				'length' => 32,
+			]);
+			$table->addColumn('uri', Types::STRING, [
+				'notnull' => true,
+				'length' => 4096,
```
marcelklehr, line 48: $ php occ app:enable webhooks
Column "oc_webhook_listeners"."uri" is type String, but exceeding the 4.000 length limit.

apps/webhooks/lib/Controller/WebhooksController.php
```diff
@@ -0,0 +1,218 @@
...
+		try {
+			$webhookListener = $this->mapper->addWebhookListener(
+				$appId,
+				$this->userId,
+				$httpMethod,
+				$uri,
+				$event,
+				$eventFilter,
+				$headers,
+				AuthMethod::from($authMethod ?? AuthMethod::None->value),
```
marcelklehr, line 116: This causes a HTTP 500 if $authMethod is not recognized

marcelklehr:
It seems the node events that you added the interface to are not actually emitted at the moment. I can only see the old `\\OCP\\Files::preWrite` ones

come-nc:
> It seems the node events that you added the interface to are not actually emitted at the moment. I can only see the old `\\OCP\\Files::preWrite` ones

They are, I tested with a NodeWrittenEvent and the webhook was called. Please detail what scenario you tested and which event firing was missing and I’ll look into it.

come-nc:
- [ ] webhooks app id is already used on appstore, we should rename.

apps/webhooks/lib/BackgroundJobs/WebhookCall.php
```diff
@@ -0,0 +1,56 @@
...
+		$options['body'] = json_encode($data);
+		$options['headers'] = $webhookListener->getHeaders();
+		try {
+			switch ($webhookListener->getAuthMethodEnum()) {
+				case AuthMethod::None:
+					break;
+				case AuthMethod::Header:
+					$authHeaders = $webhookListener->getAuthDataClear();
+					$options['headers'] = array_merge($options['headers'], $authHeaders);
```
marcelklehr, line 42: "Exception": "TypeError",
    "Message": "array_merge(): Argument #1 must be of type array, null given",
    "Code": 0,
    "Trace": [
      {
        "file": "/home/marcel/code/server/apps/webhooks/lib/BackgroundJobs/WebhookCall.php",
        "line": 42,
        "function": "array_merge"
      },

marcelklehr (changes requested):

apps/webhook_listeners/lib/Controller/WebhooksController.php
```diff
@@ -0,0 +1,220 @@
...
+		try {
+			$webhookListener = $this->mapper->addWebhookListener(
+				$appId,
+				$this->userId,
+				$httpMethod,
+				$uri,
+				$event,
+				$eventFilter,
+				$headers,
+				AuthMethod::from($authMethod ?? AuthMethod::None->value),
```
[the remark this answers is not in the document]
come-nc: I want the valueerror, I just need to catch it and turn in into the proper OCS api response.

marcelklehr:
> They are, I tested with a NodeWrittenEvent and the webhook was called. Please detail what scenario you tested and which event firing was missing and I’ll look into it.

Nevermind, as discussed, it was the leading backslash that caused this

marcelklehr (approved):

lib/public/EventDispatcher/IWebhookCompatibleEvent.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+namespace OCP\EventDispatcher;
+
+/**
+ * Interface for events which can be listened to by webhooks
+ *
+ * @since 30.0.0
+ */
+interface IWebhookCompatibleEvent {
```
nickvergessen, line 17: @come-nc Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(all changes in lib/public/)

Issue #42517, opened 2023-12-29
[Bug]: Theme color not applied properly to icons


SMH17:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

After the recent upgrade to Nextcloud 28 the theme color for icons isn't applied correctly anymore.

In this picture you can see the folder icon color correctly matches the theme color, archive and document icons ignore the selected color.
![sample_bugx](https://github.com/nextcloud/server/assets/8108287/56551665-c035-4be0-9016-8a8821e49daa)

### Steps to reproduce

1. Select a custom theme color
2. Upload a compressed archive
3. Check the icon color in nextcloud web GUI

### Expected behavior

Theme color should be applied consistently to choice.

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

28

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.2

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Update

SMH17:
After the update to 28.0.2 the theme color is totally ignored even for folders and selection highlight. 

This bug can be reproduced when it is selected "Dark Theme" in "Appearance and accessibility settings".

susnux:
> In this picture you can see the folder icon color correctly matches the theme color, archive and document icons ignore the selected color.

I have to correct myself. This is by design.
Folders are color primary, while documents are by design always that blue.
Like pdf is always red, documents are always blue.

PR #45480, opened 2024-05-23
fix(theming): replace standard generic package icon by material design

skjnldsv:
From https://github.com/nextcloud/server/issues/42517

Considering we have replaced the office icons and other filetypes are the standard grey color, ket's replace that old `#0082c9` Nextcloud blue by the same `#969696` grey.
I also used that opportunity to update the icon from the material design, like it was done with others in the last months.

| Before | After |
|:---------:|:------:|
|||

cc @nextcloud/designers


marcoambrosini (approved):

susnux (approved):

susnux:
BTW if the icon has `#0082c9` as color then it will be automatically themed using the current primary color (see IconController).
But for this I think package in grey is better

skjnldsv:
yeah, I think we need to discuss what we really want to do with the theming mime colour api. I'm not sure this is really relevant anymore :thinking:

skjnldsv:
Discussion: https://github.com/nextcloud/server/issues/45579

Issue #14153, opened 2019-02-12
Searching users is not intuitive, can't do partial searches without special characters


C0rn3j:
### Steps to reproduce
1. Have a user with display name "Martin Rys"
2. Go to https://cloud.example.com/settings/users as an admin and try to search for "Rys"
3. 0 results

### Expected behaviour
"Martin Rys" should pop up because "Rys" is part of the string "Martin Rys"

### Actual behaviour
No results. Doing "*Rys" gets the result but that is far from intuitive.

### Server configuration

**Operating system**: Ubuntu 18.04

**Web server:** Nginx

**Database:** MariaDB

**PHP version:** 7.2.10

**Nextcloud version:** 15.0.2

**Updated from an older Nextcloud/ownCloud or fresh install:** Originally NC 12

**Where did you install Nextcloud from:**
Manually.

### Client configuration
**Browser:** Firefox 65

**Operating system:** Arch Linux

### Logs
N/A, will provide if they're actually needed for the search issue.

putt1ck:
NB I think I'm going to remark this a bug as we can show that the results are inconsistent regarding the wildcard/partial match user searches. Scenario:

installation has a user (from LDAP) username john.smith, display name John Smith, the user is a member of several groups;
a search for "john" returns the user regardless of the group searched in;
a search for "smith"returns the user in some groups but not ones with many users (including the "everyone" category, installation has >400 users from LDAP, maybe 10 local to NC database);
a search for "*smith" returns the user from the large groups but not the small ones.

Further:

a search for "a" returns all LDAP users with first letter a - and all NC database users whose username, display name or email address contain "a".

putt1ck:
NB that's NC16.04, updated in many steps from ~10. LDAP server is samba, backend DB is postgres.

skjnldsv:
Duplicate of #11540

PR #45555, opened 2024-05-28
Allow LDAP search users with *term*

### LINKED ISSUE #11540: User search not efficient, returns partial matches
### Steps to reproduce
1. Go to the user section
2. Search for part of the full name or username

### Expected behaviour
I should find users with parts of there full name or there username

### Actual behaviour
I only find user when i start with the full name, if i just enter the lastname or the username I don't find anything

### Server configuration

**Operating system**:CentOS

**Web server:** NginX

**Database:** MariaDB

**PHP version:** 7.1

**Nextcloud version:** 13.0.4

**Updated from an older Nextcloud/ownCloud or fresh install:** updated

**Where did you install Nextcloud from:** nextcloud.com

smarinier:
Fix #11540 and #14153

* Resolves: # 

Fix #11540 and #14153
This issues were close, but because the user "didn't use NextCloud anymore"

## Summary

In Database User Backend, search is based on "*term*". But in LDAP, it's only "term*". This is confusing for user experience.

## Checklist

- [ X] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [0] Screenshots before/after for front-end changes
       No front change
- [0] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
   This behaviour is not described in documentation

- [0] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

### DIFF
--- a/apps/user_ldap/lib/Access.php
+++ b/apps/user_ldap/lib/Access.php
@@ -1564,7 +1564,7 @@ class Access extends LDAPUtility {
 		if ($term === '') {
 			$result = '*';
 		} elseif ($allowEnum !== 'no') {
-			$result = $term . '*';
+			$result = '*' . $term . '*';
 		}
 		return $result;
 	}


smarinier:
My commit is not in the context of Arawa company. How can i have two different sign-offs ?

apps/user_ldap/lib/Access.php
```diff
@@ -1564,7 +1564,7 @@ private function prepareSearchTerm(string $term): string {
 		if ($term === '') {
 			$result = '*';
 		} elseif ($allowEnum !== 'no') {
-			$result = $term . '*';
+			$result = '*' . $term . '*';
```
artonge, line 1567: I think the limitation was on purpose. @come-nc and @blizzz?

blizzz: It is on purpose, yes.

artonge:
> My commit is not in the context of Arawa company. How can i have two different sign-offs ?

Regarding DCO, I think it has to do with your local git config

apps/user_ldap/lib/Access.php
smarinier: I read some supposed performance limitations on big LDAP organisations. But:
 - all i read on LDAP documentation never mention "don't do *term*" as it is even described in the main documentation.
 - this might be based on old benchmarks on bad LDAP system
 - why is the "*" allowed that will send "all users" in that case ?
 - the SQL part does the *term* and nobody asked to remove this for performance reason
 - this is very very confusing for the user experience (i thought this was a bug and i lost hours searching what was wrong with my LDAP until i debug in PHP) 
 
 In conclusion, at least, why presuming all ldap are made with millions of users ? Why not let users have the same behaviour between LDAP and NextCloud ? And maybe bring an LDAP option "don't allow *term* because my ldap is too slow" if someone complain for it ?

smarinier:
> > My commit is not in the context of Arawa company. How can i have two different sign-offs ?
> 
> Regarding DCO, I think it has to do with your local git config

Are you sure ? my git config does not have any information like my name. I'm registered only with my github account. On the other hand, this github account is configured with my two emails. I guess that there is something in your DCO telling smarinier is sebastien.marinier@arawa.fr and it refuses another email (as i was refused once because i use my real accentuated name "Sébastien" instead of "Sebastien", with the correct email).

artonge:
> Are you sure ?

No. Can you run the following from the nextcloud repo:

```shell
git config user.email
git config --global user.email
```

smarinier:
> > Are you sure ?
> 
> No. Can you run the following from the nextcloud repo:
> 
> ```shell
> git config user.email
> git config --global user.email
> ```

Dawn! It was in the global. I've changed it and re-push the sign off. Thanks

come-nc (commented):
If you make it a config option, maybe, but we cannot merge this as-is.
I do not have the details about when the decision was first taken, and I agree that with a good LDAP setup wildcard search is not an issue, but there is a lot of code in Nextcloud which is made to account for badly configured LDAP server, this is not the only place.

PR #45562, opened 2024-05-29
fix(ShareApiClients): Maintain meaning of `''` and null for share api cli…

nfebe:
Fix regression from : https://github.com/nextcloud/server/pull/44485


artonge (approved):

ShGKme (approved):

ShGKme:
Should also be backported to 28 and 29 after https://github.com/nextcloud/server/pull/44485 is backported

Fenn-CS:
> Should also be backported to 28 and 29 after https://github.com/nextcloud/server/pull/44485 is backported to 28 and 29

I would fix up? on 29 and 28? Since they are not merged yet? @ShGKme

PR #45571, opened 2024-05-29
fix(files): also trigger new tab on file name middle click

### LINKED ISSUE #44994: [Bug]: Files app doesn't open middle-clicked file/folder in new browser tab since Nextcloud 28

### Bug description

In Nextcloud 27.1.8 I could middle-click on files and folders in the Files app and the middle-clicked file/folder would then open in a new browser tab. This doesn't work anymore after upgrading to Nextcloud 28.0.4.

I found three places where this issue is mentioned: [PR #42457](https://github.com/nextcloud/server/pull/42457#issuecomment-1910393479), [Forum](https://help.nextcloud.com/t/after-update-to-nc28-mouse-wheel-doesnt-open-folders-and-files-in-new-tab-anymore/186972), [Reddit](https://www.reddit.com/r/sysadmin/comments/1b7bbmq/nextcloud_change_middleclick_behaviour_from/).

### Steps to reproduce
**Folder**
1. Open the Files app (https://<nextcloud_domain>/apps/files/files)
2. Middle-click a folder, e.g. the folder 'Documents'
3. Firefox: A new tab is opened with the URL https://<nextcloud_domain>/remote.php/dav/files/<user>/Documents showing a white page with only this text:
> This is the WebDAV interface. It can only be accessed by WebDAV clients such as the Nextcloud desktop sync client.
3. Chromium: A file names Documents.html is downloaded containing this text:
> This is the WebDAV interface. It can only be accessed by WebDAV clients such as the Nextcloud desktop sync client.

**File**
1. Open the Files app (https://<nextcloud_domain>/apps/files/files)
2. Middle-click a file
3. Either nothing happens (e.g, .pdf, .txt files)
4. Or the file is downloaded (e.g., .md, .gif, .odt, .ods files)

Note: after I click on the 'All files' button at the top left, nothing happens when I middle-click files/folders. The URL is then https://<nextcloud_domain>/apps/files/files?dir=/ instead of https://<nextcloud_domain>/apps/files/files.

### Expected behavior

I expect folders and files to open in a new browser tab, just like in Nextcloud 27.

### Installation method

Community Docker image

### Nextcloud Server version

28

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.2

### Web server

Nginx

### Database engine version

PostgreSQL

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

skjnldsv:
Fix https://github.com/nextcloud/server/issues/44994


ShGKme (commented):
`auxclick` is not supported on Safari, according to https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event
Why not use `click.middle`?

Anyway, in general, I'd prefer to have fixed links in the list rather than increase manual imitation. What is implemented in this PR should for by default in any browser with the correct link.

(Mentioned in https://github.com/nextcloud/server/pull/42457#issuecomment-1910393479)

skjnldsv:
> Anyway, in general, I'd prefer to have fixed links in the list rather than increase manual imitation. What is implemented in this PR should for by default in any browser with the correct link.

Oh, 100% with you on this!
I didn't want to change the manual imitation that someone else did, so I used the sape ways, but I will definitely try a different way since you're also on board with this! :)

Regarding the `click.middleè it was definitely not working initially, but I changed some stuff in between, so it mighth work again now :shrug:

skjnldsv:
Ah, no! Now I do remember!
We do indeed have different links, that was how it was before.
We can discuss the default link expectations like Julius said, but as a fix, I think we should still merge this PR (after the auxclick adjusting).

If we want to remove the default click-to-download behaviour is a different topic in itself @ShGKme :)

skjnldsv:
Alright, I tested on safari and it works!

- if I remove the `auxclick` it fails on chrome
- the click seems to only work on safari

So leaving them both is the way :+1:

szaimen (approved):
Just tested in a Chromium based browser on Windows and seems to work :)

ShGKme (approved):
Tested on Safari 16 and Safari 17, both testing cloud and real MacBook.

Although MDN and [caniuse](https://caniuse.com/auxclick) say this event is not supported on Safari completely, it works fine on both list and grid with the middle button click.

But `@click.middle` would avoid the need to check `event.button` 👀

artonge (approved):

artonge:
I think this also make the right-click trigger the default action

Issue #44933, opened 2024-04-18
database log filled with errors every second


warlordattack:
### Which version of recognize are you using?

latest

### Enabled Modes

Object recognition, Face recognition, Video recognition, Music recognition

### TensorFlow mode

WASM mode

### Downstream App

Memories App

### Which Nextcloud version do you have installed?

latest

### Which Operating system do you have installed?

ubuntu

### Which database are you running Nextcloud on?

postgres

### Which Docker container are you using to run Nextcloud? (if applicable)

recommended AIO install

### How much RAM does your server have?

16GiB

### What processor Architecture does your CPU have?

x86_64

### Describe the Bug

ERROR: duplicate key value violates unique constraint “oc_systemtag_object_mapping_pkey”
DETAIL: Key (objecttype, objectid, systemtagid)=(files, 112177, 34) already exists.
CEST [1789] STATEMENT: INSERT INTO “oc_systemtag_object_mapping” (“objectid”, “objecttype”, “systemtagid”) VALUES($1, $2, $3)

### Expected Behavior

shold not fill the database log with errors every second

### To Reproduce

installed fresh nextcloud via recommended AIO methos on a new Ubuntu proxmox VM, 
installed Memories and Recognize
activated auto tagging
dropped few thousands wallpapers there
opened the logs found database log filled with errors and 20Mo size

### Debug log

[nextcloud-aio-database-2024-04-18T15-05-49.log](https://github.com/nextcloud/recognize/files/15026932/nextcloud-aio-database-2024-04-18T15-05-49.log)

warlordattack:
If more information is needed i can provide.
Best regards :)

joshtrichards:
I know you mentioned new installation in the reproduction steps, but any chance your existing database was restored from backup at some point or migrated from another installation?

warlordattack:
It was a clean new install, i am testing, so i delete all containers and docker disks before new install, database was normally deleted when i deleted docker disks before install.

marcelklehr:
Looks like this is the code that causes this. The errors don't seem to be fatal, but sure are annoying I assume: https://github.com/nextcloud/server/blame/master/lib/private/SystemTag/SystemTagManager.php#L180

bigcat88:
We quite often receive tickets regarding "Duplicate Key Errors" from the oc_filecache/oc_file_metadata/or oc_systemtag_object_mapping tables.

[...]

PR #45580, opened 2024-05-30
fix: avoid duplicate tag inserts by checking if the mapping exists already in db

yemkareems:
fix: do a select in systemtag_object_mapping to see if tag exists already in db. if it does not exist alone insert the same or else do nothing

* Resolves: # 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/SystemTag/SystemTagObjectMapper.php
+++ b/lib/private/SystemTag/SystemTagObjectMapper.php
@@ -126,11 +126,16 @@ class SystemTagObjectMapper implements ISystemTagObjectMapper {
 		$result = $query->executeQuery();
 		$rows = $result->fetchAll();
 		$existingTags = [];
-		foreach ($rows as $k => $row) {
+		foreach ($rows as $row) {
 			$existingTags[] = $row['systemtagid'];
 		}
 		//filter only tags that do not exist in db
 		$tagIds = array_diff($tagIds, $existingTags);
+		if (empty($tagIds)) {
+			// no tags to insert so return here
+			$this->connection->commit();
+			return;
+		}
 
 		$query = $this->connection->getQueryBuilder();
 		$query->insert(self::RELATION_TABLE)
--- a/lib/private/SystemTag/SystemTagObjectMapper.php
+++ b/lib/private/SystemTag/SystemTagObjectMapper.php
@@ -117,6 +117,15 @@ class SystemTagObjectMapper implements ISystemTagObjectMapper {
 		$this->assertTagsExist($tagIds);
 
 		$query = $this->connection->getQueryBuilder();
+		$query->select('systemtagid')
+			->from(self::RELATION_TABLE)
+			->where($query->expr()->in('systemtagid', $query->createNamedParameter($tagIds, IQueryBuilder::PARAM_INT_ARRAY)))
+			->andWhere($query->expr()->eq('objecttype', $query->createNamedParameter($objectType)))
+			->andWhere($query->expr()->eq('objectid', $query->createNamedParameter($objId)));
+		$result = $query->executeQuery();
+		$rows = $result->fetchAll();
+
+		$query = $this->connection->getQueryBuilder();
 		$query->insert(self::RELATION_TABLE)
 			->values([
 				'objectid' => $query->createNamedParameter($objId),
@@ -126,12 +135,17 @@ class SystemTagObjectMapper implements ISystemTagObjectMapper {
 
 		$tagsAssigned = [];
 		foreach ($tagIds as $tagId) {
-			try {
-				$query->setParameter('tagid', $tagId);
-				$query->execute();
-				$tagsAssigned[] = $tagId;
-			} catch (UniqueConstraintViolationException $e) {
-				// ignore existing relations
+			if(!in_array($tagId, array_column($rows, 'systemtagid'))) {
+				// tag not in db so create new one
+				try {
+					$query->setParameter('tagid', $tagId);
+					$query->execute();
+					$tagsAssigned[] = $tagId;
+				} catch (UniqueConstraintViolationException $e) {
+					// ignore existing relations
+				}
+			} else {
+				//tag exists already don't insert
 			}
 		}
 
--- a/lib/private/SystemTag/SystemTagObjectMapper.php
+++ b/lib/private/SystemTag/SystemTagObjectMapper.php
@@ -115,6 +115,7 @@ class SystemTagObjectMapper implements ISystemTagObjectMapper {
 		}
 
 		$this->assertTagsExist($tagIds);
+		$this->connection->beginTransaction();
 
 		$query = $this->connection->getQueryBuilder();
 		$query->select('systemtagid')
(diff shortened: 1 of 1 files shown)


lib/private/SystemTag/SystemTagObjectMapper.php
```diff
@@ -126,12 +135,17 @@ public function assignTags(string $objId, string $objectType, $tagIds): void {
 
 		$tagsAssigned = [];
 		foreach ($tagIds as $tagId) {
-			try {
-				$query->setParameter('tagid', $tagId);
-				$query->execute();
-				$tagsAssigned[] = $tagId;
-			} catch (UniqueConstraintViolationException $e) {
-				// ignore existing relations
+			if(!in_array($tagId, array_column($rows, 'systemtagid'))) {
```
artonge, line 138: I guess you could use `$tagIds = array_diff($tagIds, $rows)` to explicitly filter out existing tags.

lib/private/SystemTag/SystemTagObjectMapper.php
```diff
@@ -116,6 +116,15 @@ public function assignTags(string $objId, string $objectType, $tagIds): void {
 
 		$this->assertTagsExist($tagIds);
 
+		$query = $this->connection->getQueryBuilder();
```
artonge, line 119: Maybe wrap the select and the inserts into a transaction?
```php
$this->connection->beginTransaction();
...
select
...
inserts
...
$this->connection->commit();
```

szaimen:
Sorry, cannot judge the code

lib/private/SystemTag/SystemTagObjectMapper.php
yemkareems: @artonge Yes done. filtered tags that do not exist and using them alone for the insert

lib/private/SystemTag/SystemTagObjectMapper.php
yemkareems: @artonge wrapped the select/insert around a transaction

artonge (approved):

kesselb:
SystemTagObjectMapper / oc_systemtag_object_mapping state if a tag belongs to a given object (e.g. a file).  The code does a select to see if a given mapping already exists, not a tag. I suggest updating the commit message accordingly.

Is there a reason for this change and the additional select? The situation that a mapping already exists is handled properly by the catching the UniqueConstraintViolationException.

artonge:
> Is there a reason for this change and the additional select? The situation that a mapping already exists is handled properly by the catching the UniqueConstraintViolationException.

The catch is enough to prevent Nextcloud logs, but not enough to prevent logs from the DB which apparently can grow very large in short time spans.

kesselb:
> The catch is enough to prevent Nextcloud logs, but not enough to prevent logs from the DB which apparently can grow very large in short time spans.

Thanks. 

I'm aware about this problem with filecache and authtokens. 
This is indeed an annoying situation, especially with PostgreSQL. 

It surprised me a bit to see this problem with the systemtag_object_mapping table as well. 
Is that in combination with files_automatedtagging or recognize?

kesselb:
For some cases, it's possible to use our "insertIgnoreConflict" method: https://github.com/nextcloud/server/blob/d0454827321b6f589819cf8664f827d1caa38538/lib/public/IDBConnection.php#L147-L158

For example:
https://github.com/nextcloud/server/commit/4361019f2f3cdb48da90af6345b21bd579c725d2

kesselb:
> The implementation only covers PostgreSQL, is there any equivalent for MySQL?

[INSERT IGNORE](https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#ignore-effect-on-execution) (also non-standard unfortunately)

Altahrim:
`INSERT IGNORE` is a bit dangerous because it will also ignore other errors 

My proposal: https://github.com/nextcloud/server/pull/45655

bigcat88:
can I ask if there a backport planned for this(or a linked PR) for NC28 or NC29 or it is not a option for this PR and these really cool changes will come only with NC30?

lib/private/SystemTag/SystemTagObjectMapper.php
```diff
@@ -115,6 +115,22 @@ public function assignTags(string $objId, string $objectType, $tagIds): void {
...
+		$query->select('systemtagid')
+			->from(self::RELATION_TABLE)
+			->where($query->expr()->in('systemtagid', $query->createNamedParameter($tagIds, IQueryBuilder::PARAM_INT_ARRAY)))
+			->andWhere($query->expr()->eq('objecttype', $query->createNamedParameter($objectType)))
+			->andWhere($query->expr()->eq('objectid', $query->createNamedParameter($objId)));
+		$result = $query->executeQuery();
+		$rows = $result->fetchAll();
+		$existingTags = [];
+		foreach ($rows as $k => $row) {
```
kesselb, line 129:
```suggestion
		foreach ($rows as $row) {
```

$k is unused.

lib/private/SystemTag/SystemTagObjectMapper.php
```diff
@@ -115,6 +115,22 @@ public function assignTags(string $objId, string $objectType, $tagIds): void {
...
+			$existingTags[] = $row['systemtagid'];
+		}
+		//filter only tags that do not exist in db
+		$tagIds = array_diff($tagIds, $existingTags);
```
kesselb, line 133: I think we could check if $tagIds are empty and early exit, but what happens to the transaction, do we need to commit it in this case too?

kesselb:
I read Alexander's support ticket and https://github.com/nextcloud/server/issues/44933 and now the change makes much more sense to me. 

To my understanding, the "duplicate key value violates unique constraint" errors for filecache and auththokes are caused by operations running in parallel. 

The error for systemtag_object_mapping because our implementation is incomplete or at least strange ;)

Take the following flow: 

image

1. Upload an image
2. Rename the image
3. Move the image somewhere else

The flow is triggered for all operations and will trigger the warning for 2. and 3. 

So it's not an edge case or a parallel operation. The api consumer like files_automatedtagging or recognize expect that assignTags takes care if a tag is already assigned. 

It would be still nice to improve the commit message (it's checking if the mapping does exist and not the tag).
I think we could backport it as well, as it's more a bug fix than a feature.

kesselb (approved):

lib/private/SystemTag/SystemTagObjectMapper.php
yemkareems: Added a extra check for the empty tag ids @kesselb

yemkareems:
> can I ask if there a backport planned for this(or a linked PR) for NC28 or NC29 or it is not a option for this PR and these really cool changes will come only with NC30?

@bigcat88 we will backport this to NC28 and NC29

kesselb:
cc @marcelklehr because I recall some reports for recognize about it. Not sure though if about that exact problem, but related to the tagging.

marcelklehr (approved):
LGTM

PR #45585, opened 2024-05-30
refactor(files): Adjust some Typescript code in virtual files list

susnux:
## Summary
- Correctly cast table ref to fix TypeScript errors (I think this also makes it a bit better readable :) )
- Refactor the `filesListWidth` mixin
  Use `defineComponent` to properly inherit typings, also add the `$resizeOberserver` through the setup to the component instance.
  `filesListWidth` is now a number which defaults to 0, making compares like `this.fileListWidth < 768` valid in Typescript (there are a lot of Type error in other places because of this).
- Properly add type information to `FilesListTableFooter`
  - Add correct type for `nodes` prop.
  - Use `defineComponent` to properly infer Typescript information
  - Correct usage of nullish coalescing operator
- Make `currentView` a prop of `FilesListTableFooter`
  We already have this as a required prop in `FilesListVirtual`, so passing it to the footer is straight forward and removed need of computed value.

## Checklist


susnux (commented):
Added some hints about the problems I try to solve

apps/files/src/mixins/filesListWidth.ts
```diff
@@ -3,26 +3,31 @@
...
+			filesListWidth: 0,
 		}
 	},
+
 	mounted() {
 		const fileListEl = document.querySelector('#app-content-vue')
-		this.filesListWidth = fileListEl?.clientWidth ?? null
+		this.filesListWidth = fileListEl?.clientWidth ?? 0
 
+		// @ts-expect-error The resize observer is just now attached to the object
```
susnux, line 19: We can not use `setup` in mixins, so we can not attach it properly with type information.
Using `data` also feels wrong, it should not be made reactive.

ShGKme: This mixin is quite small and only defines one property. Maybe we can rewrite it to a composable then?

ShGKme (approved):

ShGKme:
Thanks for detailed description about all the changes!

apps/files/src/mixins/filesListWidth.ts
susnux: Yes I also was thinking about that.
Also a composable for `dir` and `fileid` makes sense as we use it in many places and just duplicate logic there.

skjnldsv (approved):

Issue #6914, opened 2017-10-23
Config Param 'overwritecondaddr' not working


ghmax:
### Steps to reproduce
1. clean install nextcloud
2. add conditional overwrite settings to config.php 
   'overwritehost' => 'my.domain.com',
   'overwritecondaddr' => '^192\.168\.0\.1$',

### Expected behaviour
Hostname should be overwritten when request comes in via 192.168.0.1, otherwise NOT.

### Actual behaviour
Hostname is always overwritten

### Server configuration

**Operating system**:
official Docker Image on different Docker Hosts

**Nextcloud version:** (see Nextcloud admin page)
12.0.3.3

**Updated from an older Nextcloud/ownCloud or fresh install:**
fresh

**Where did you install Nextcloud from:**
docker

**Nextcloud configuration:**
<details>
<summary>Config report</summary>

```
<?php
$CONFIG = array (
  'htaccess.RewriteBase' => '/',
  'memcache.local' => '\\OC\\Memcache\\APCu',
  'apps_paths' => 
  array (
    0 => 
    array (
      'path' => '/var/www/html/apps',
      'url' => '/apps',
      'writable' => false,
    ),
    1 => 
    array (
      'path' => '/var/www/html/custom_apps',
      'url' => '/custom_apps',
      'writable' => true,
    ),
  ),
  'instanceid' => 'xxx',
  'passwordsalt' => 'xxx',
  'secret' => 'xxx',
  'trusted_domains' => 
  array (
    0 => '192.168.99.100:9000',
  ),
  'datadirectory' => '/var/www/html/data',
  'overwrite.cli.url' => 'http://192.168.99.100:9000',
  'dbtype' => 'sqlite3',
  'version' => '12.0.3.3',
  'installed' => true,
  
   'overwritecondaddr' => '^192\.168\.99\.11$',
   'overwritehost' => 'my.domain.com',
  
);
```
</details>

### Code causing Troubles...
Problem is in File .../lib/private/AppFramework/Http/Request.php in Method 'private function isOverwriteCondition($type = '')' ... type is empty when checking for e.g. Host-Override in Method 'getOverwriteHost' ... so with empty type check override method (isOverwriteCondition) always returns true cause of:

...
return $regex === '//' || preg_match($regex, $remoteAddr) === 1 || $type !== 'protocol';
...

-> type === '' ... so $type !== 'protocol' is always true ... so method always returns true in this case.

blizzz:
[...]

PR #45651, opened 2024-06-03
feat(lib): Improve remote address handling and overwritecondaddr checks

aschieweck:
## Summary

This PR has two related improvements:
1. Include support for the [Forward Header format](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded) as alternative to the [X-Foraward-For Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For) when checking for the remote address.
2. Check the `overwritecondaddr` regex from the config against the remote address AND the trusted proxy chain.

I ran into a similar issue as [danielwbn](https://github.com/nextcloud/server/issues/6914#issuecomment-416461735) in #6914. My remote address is always the IP of the client, which is great and shows that my reverse proxy is working, but I need to set specific overrides depending on the proxy route the request has taken. 

The `isOverwriteCondition()` method was using the direct call to the `server['REMOTE_ADDR'])` instead of the `getRemoteAddress()` method. As both methods would need to interact with the trusted proxies and the headers, I refactored this part out into a common method. This can now also handle both styles of forward headers used by reverse proxies.

## TODO

- [x] Tests for the override
- [ ] Check contributor fluff (first PR here, I'll have to check what I'm missing)

## Checklist

- [ ] ~Screenshots before/after for front-end changes~


aschieweck:
Hi @kesselb, any thoughts on the idea or the current state of this PR?

kesselb:
Thanks for your pull request :+1: 

I'm unsure if we should have two default `forwarded_for_headers`. 

What is your use case for `overwritecondaddr`?

aschieweck:
The `X-Forwared-For` is still the most common style, but as the _X_ indicates, it was experimental and is now replaced in the standard by the `Forwared` header. I think it's a good idea to support both in Nextcloud. I don't have a strong opinion which one, or maybe both, should be the default. This PR has both as default, has both header types are part of the standard. But this can be overwritten in the server config, should it be required. Happy to change it to what makes the most sense to you!

My use case for allowing proxies in the `overwritecondaddr` is the use of Cloudflare tunnels. I have the local tunnel endpoint in a docker container and I would like to change some things if the traffic is coming from this source. At the moment, `overwritecondaddr` is only checking the remote address reported to Nextcloud, which in the case of Cloudflare tunnels is the real IP address of the user. I _might_ be able to write a regex if the remote address is not from the local range, but this feels more like a hack. Checking against the proxy is just cleaner.

kesselb:
> Just merged the current state of the master branch into this one. I hope this should fix most of the checks, as they failed on some setup procedures, i.e. running composer. I haven't touched composer things in this PR, so I'm a bit lost while this wouldn't work.

Yep, some of our CI have issues with forks. 

It looks like a reasonable change. We need the ip addresse for the brute force detection, and thus it makes sense to also check the reverse proxy for the overwritecondaddr. We only take trusted proxies into account. 

> I would like to change some things if the traffic is coming from this source. 

What would you like to change? I'm just wondering because most of the use cases for overwritecondaddr are usually look like hack. Do you set a different hostname if the request is coming through cf?

aschieweck:
> Yep, some of our CI have issues with forks.

OK, I had a look at the failing checks, and I'd say the failures are due to the CI & fork thing. Nothing related to this PR. But please let me know if there is a check, I should have a closer look at

> We only take trusted proxies into account.

This PR is only looking at the remote address report and the chain of trusted proxies. Let's say we have a request with the following header information: *Client - Proxy 1 - Trusted Proxy 1 - Proxy 2 - Trusted Proxy 2 - Trusted Proxy 3 - Nextcloud*. Then `getRemoteAddressAndProxyChain` would return the *Proxy 2* as remote address. As everything before *Trusted Proxy 2* could be spoofed.  This is similar to how it was before. But we know also get the chain of *Trusted Proxy 2, Trusted Proxy 3* as chain for other checks.

> Do you set a different hostname if the request is coming through cf?

That's exactly it. The cleanest way would be to have separate domain and path settings based on the upstream proxy. But that sound like a lot of work for a very niche problem. I'm happy with some jankiness and understand that this isn't a common (or even supported) use case.

aschieweck:
Hello @nickvergessen & @Altahrim! Do you guys have any feedback on this?

aschieweck:
As there was no update in over a year and now conflicts are showing up, I'm going to withdraw this PR.

PR #45655, opened 2024-06-04
feat(dbal): insert ignore conflict method for MySQL and SQLite

Altahrim:
## Summary

Add ignore conflicts method for MySQL
Similar to PgSQL: https://github.com/nextcloud/server/blob/master/lib/private/DB/AdapterPgSql.php#L26-L37

## Checklist

### DIFF
--- a/lib/private/DB/AdapterMySQL.php
+++ b/lib/private/DB/AdapterMySQL.php
@@ -35,4 +35,32 @@ class AdapterMySQL extends Adapter {
 
 		return $this->collation;
 	}
+
+	public function insertIgnoreConflict(string $table, array $values): int {
+		$builder = $this->conn->getQueryBuilder();
+		$builder->insert($table);
+		$updates = [];
+		foreach ($values as $key => $value) {
+			$builder->setValue($key, $builder->createNamedParameter($value));
+		}
+
+		/*
+		 * We can't use ON DUPLICATE KEY UPDATE here because Nextcloud use the CLIENT_FOUND_ROWS flag
+		 * With this flag the MySQL returns the number of selected rows
+		 * instead of the number of affected/modified rows
+		 * It's impossible to change this behaviour at runtime or for a single query
+		 * Then, the result is 1 if a row is inserted and also 1 if a row is updated with same or different values
+		 *
+		 * With INSERT IGNORE, the result is 1 when a row is inserted, 0 otherwise
+		 *
+		 * Risk: it can also ignore other errors like type mismatch or truncated data…
+		 */
+		$res = $this->conn->executeStatement(
+			preg_replace('/^INSERT/i', 'INSERT IGNORE', $builder->getSQL()),
+			$builder->getParameters(),
+			$builder->getParameterTypes()
+		);
+
+		return $res;
+	}
 }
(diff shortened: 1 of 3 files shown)


artonge (approved):

come-nc (commented):
On conflict this ones updates the row while pgsql implementation will not touch it, no?

Altahrim:
> On conflict this ones updates the row while pgsql implementation will not touch it, no?

Yep, see above comment :)

solracsf:
> On conflict this ones updates the row while pgsql implementation will not touch it, no?

Can this be an option?
https://mariadb.com/kb/en/insert-ignore/

Altahrim:
> Can this be an option?
> [mariadb.com/kb/en/insert-ignore](https://mariadb.com/kb/en/insert-ignore/)

For me it ignores too many errors. I would prefer to avoid this.

lib/private/DB/AdapterMySQL.php
```diff
@@ -35,4 +35,20 @@ protected function getCollation(): string {
...
+	public function insertIgnoreConflict(string $table, array $values): int {
+		$builder = $this->conn->getQueryBuilder();
+		$builder->insert($table);
+		$updates = [];
+		foreach ($values as $key => $value) {
+			$builder->setValue($key, $builder->createNamedParameter($value));
+		}
+
+		$field = key($values);
+		return $this->conn->executeStatement(
```
Altahrim, line 48: This method doesn't return the `affected_rows` as MySQL do in CLI
That's why the tests are failing. If anyone have an idea, I am interested :) 

Example from the tests:
```sql
INSERT INTO oc_twofactor_providers (provider_id, uid, enabled) VALUES ('test', 'user123', 0) ON DUPLICATE KEY UPDATE provider_id = provider_id;
-- MySQL CLI: Query OK, 1 row affected (0.018 sec)
-- PHP returns: 1
INSERT INTO oc_twofactor_providers (provider_id, uid, enabled) VALUES ('test', 'user123', 1) ON DUPLICATE KEY UPDATE provider_id = provider_id;
-- MySQL CLI: Query OK, 0 rows affected (0.001 sec)
-- PHP returns: 1 => Should be 0, one row selected, but no row affected
```

Altahrim:
Finally, it seems `INSERT IGNORE` is the only way to achieve this with our current setup of MySQL.

From [MySQL documentation](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html)
> With ON DUPLICATE KEY UPDATE, the affected-rows value per row is 1 if the row is inserted as a new row, 2 if an existing row is updated, and 0 if an existing row is set to its current values. If you specify the CLIENT_FOUND_ROWS flag to the [mysql_real_connect()](https://dev.mysql.com/doc/c-api/8.0/en/mysql-real-connect.html) C API function when connecting to [mysqld](https://dev.mysql.com/doc/refman/8.0/en/mysqld.html), the affected-rows value is 1 (not 0) if an existing row is set to its current values.

The other way is to update all fields instead of ignore them but I failed to achieve this with Oracle. It's sad because it could make sense when we launch an update after the `insertIgnoreConflict`

Altahrim:
@nickvergessen Do you think it is a correct solution given the risks of ignoring other errors?

come-nc (commented):
A unit test would be nice

yemkareems (approved):

Issue #45682, opened 2024-06-05
[Bug]: Change Warn on excessive memory consumption in background jobs to info


AndyXheli:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description
Log file it getting spammed with these warning. i would recommended this to  info and not warning, 
Used memory grew by more than 10 MB when executing job OC\FilesMetadata\Job\UpdateSingleMetadata (id: 175902, arguments: ["usernae",1042806]): 26.1 MB (before: 10.7 MB)

### Steps to reproduce

N/A

### Expected behavior

Should not see this as a warning more info then a warning 

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

29

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.3

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Upgraded to a MAJOR version (ex. 22 to 23)

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Additional info

![image](https://github.com/nextcloud/server/assets/59488153/31ed58af-fe9d-4efa-b109-c1e9ebde33bf)

solracsf:
Already on its way at https://github.com/nextcloud/server/pull/45530

AndyXheli:
Hey @joshtrichards sorry to bother . 

Thanks for the update so under the Nc29 backport till shows 10mb but under the original pull the 10mb was changed to 50 ? 

![image](https://github.com/nextcloud/server/assets/59488153/d0ad3e7d-2ecc-41dd-8aa3-6678cbc25d22)

joshtrichards:
[...]

PR #45530, opened 2024-05-27
fix(cron): Log excess memory usage at debug and info levels

joshtrichards:
* Resolves: #45189
* Resolves: https://github.com/nextcloud/server/issues/45682

## Summary

Admins/operators don't need to see this in day-to-day scenarios >= the default log level. Moved the 10MiB (common) to DEBUG level and the 300 MiB to INFO level (rarer but still common enough).

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #45683, opened 2024-06-05
Update Level to Info for  Warn on excessive memory consumption in background jobs

AndyXheli:
Signed-off-by: Andy Xheli <andyxheli@gmail.com>

Fix 

[Bug]: Change Warn on excessive memory consumption in background jobs to info #45682


joshtrichards:
Hi @AndyXheli - see #45530 :)

PR #45702, opened 2024-06-07
feat(SecurityMiddleware): Allow app_api to bypass user check

provokateurin:
## Summary

Necessary for https://github.com/nextcloud/server/pull/45391. ExApps need to be able to call endpoints without specifying a user.

## Checklist


nickvergessen (changes requested):
this is not the way to go to solve the issue in #45391

bigcat88:
PR is good, we just make sure that the new AI APIs that we are developing for NC30 can be called without a specified user in the session, and that they can be used by ExApps that do not have “system” privileges (they **cannot** impersonate the default user)

This condition does not change any operating logic; they still cannot call admin endpoints.

bigcat88 (approved):

PR #45706, opened 2024-06-07
Updated images for Material icons and added SPDX info

AndyScherzinger:
* Resolves: # further SPDX infos and M3 icon replacements aligned with Jan

## Summary

* files with copyright information: 10749 / 11030
* files with license information: 10732 / 11030

## TODO

- [ ] review and merge

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #45712, opened 2024-06-07
fix(dist): Make sure changed icons are compiled

nickvergessen:
* Regression from #45706 which only changed icons
* it did not recompile the icons.css
* but since no matching file was changed, the node.yml job didn't run there
* Since then all PRs that do change a checked file fail with red CI without having changed the file.


blizzz (approved):

provokateurin (approved):

susnux (approved):

susnux:
Missing: `core/css`

Noticed on this PR, which was merged without compiled styles https://github.com/nextcloud/server/pull/46270

PR #45716, opened 2024-06-07
feat: Add Recently active accounts section

Pytal:
## Summary

- Requires https://github.com/nextcloud/server/pull/45249
- Supersede https://github.com/nextcloud/server/pull/45500

| Screenshot |
| --- |
| image |

## Checklist


apps/settings/lib/Controller/UsersController.php
```diff
@@ -164,9 +164,15 @@ public function usersList(): TemplateResponse {
 			$userCount -= $disabledUsers;
 		}
 
+		$recentUsersGroup = [
+			'id' => 'recent',
+			'name' => $this->l10n->t('Recent accounts'),
```
artonge, line 169: Not sure that it reflect the recently logged in logic. It sounds like recently created. Maybe @nextcloud/designers can give some input here.

artonge (approved):

susnux (approved):

susnux:
> BREAKING CHANGE: The "recent" group id is now reserved for the Recent accounts section and will not be shown under the Groups heading

BTW could we maybe rename that to be less likly to conflict? Like `__nc_internal_recent`?

apps/settings/lib/Controller/UsersController.php
susnux: I think the problem is that we call it "active" when we mean "enabled".
Because then we could name it:
* Enabled -> all enabled ones
* Disabled -> all disabled (like it is)
* Active -> recently logged in as they are "active"

(not 100% sure that naming makes that much sense in English like it makes in German :sweat_smile: )

jancborchardt: How about:
- All accounts
- Admins
- Recently active
- Disabled accounts

This sounds most like understandable language to me?

PR #45761, opened 2024-06-10
fix restore of multiple folders with many files from trash_bin

mgallien:
## Summary

fixes restore from trash bin of multiple folders with many files inside them
switch the DB statement execution method

related to https://github.com/nextcloud/server/pull/44504

Some tests were carried, and this simple change seems to fix the issue.


mgallien:
would backport if approved

kesselb:
I assume the retry logic does not work because the wrong exception is caught.

artonge:
> I assume the retry logic does not work because the wrong exception is caught.

:brain: so maybe unwrapping the exception would be enough?

artonge:
@mgallien can you test  https://github.com/nextcloud/server/pull/45768 ?

PR #45771, opened 2024-06-11
test(logger): Fix `loglevel is not a valid integer` warnings

joshtrichards:
## Summary

#44262 introduced a warning which is adding noise (x42 each run) to the logger unit tests: 

> `Nextcloud configuration: "loglevel" is not a valid integer`

This is happening because of the mocking for `SystemConfig::class` returning null for all `getValue()` calls for all but one of the Logger tests.

One lone test already had a stub in place for `getValue()` + `loglevel`.  I moved this existing stub into `setUp()` so all the tests could benefit from it. This cut out all 42 warnings.

I don't see any downsides, in this case, to this particular stub being shared across the Logger tests (i.e. the two specified values don't change the spirit of any of the existing tests).

If there's a better spot to stick stubs that are shared across a group of tests, let me know. 

Apologies if the above was super obvious to anyone reading this. I'm still getting familiar with the testing. :)

## Checklist

### DIFF
--- a/tests/lib/LoggerTest.php
+++ b/tests/lib/LoggerTest.php
@@ -34,6 +34,13 @@ class LoggerTest extends TestCase implements IWriter {
 		$this->config = $this->createMock(SystemConfig::class);
 		$this->registry = $this->createMock(IRegistry::class);
 		$this->logger = new Log($this, $this->config, null, $this->registry);
+
+		$this->config->expects($this->any())
+			->method('getValue')
+			->will(($this->returnValueMap([
+				['loglevel', ILogger::WARN, ILogger::WARN],
+				['log.condition', [], ['apps' => ['files']]]
+			])));
 	}
 
 	public function testInterpolation() {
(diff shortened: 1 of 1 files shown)


tests/lib/LoggerTest.php
```diff
@@ -34,6 +34,13 @@ protected function setUp(): void {
 		$this->config = $this->createMock(SystemConfig::class);
 		$this->registry = $this->createMock(IRegistry::class);
 		$this->logger = new Log($this, $this->config, null, $this->registry);
+
+		$this->config->expects($this->any())
+			->method('getValue')
+			->will(($this->returnValueMap([
+				['loglevel', ILogger::WARN, ILogger::WARN],
+				['log.condition', [], ['apps' => ['files']]]
```
come-nc, line 42: The `log.condition` line was specific to `testAppCondition` it should not be made general like that.

joshtrichards: I didn't want the `log.condition` one there either, but phpunit doesn't make it easy to override stubs once they're defined. I was trying to avoid creating a bunch of duplicate boilerplate code (as in adding the `loglevel` stub individually to *all* 42 other tests that rely on it).

I guess we either use boilerplate stub code in each test or create a `generalStubs()` (or whatever) helper function that each test calls other than the ones that will be providing their own stubs?

come-nc: Can you use a function callback with a `$this->configReturnMap` property which tests can fill up for instance?

kesselb: I usually go with a global mock in setUp and just create a local mock if a test needs a much different state.

kesselb:
Resolved via https://github.com/nextcloud/server/pull/46284

PR #45782, opened 2024-06-11
feat(core): add selection for recommended apps

### LINKED ISSUE #45084: Check/Uncheck button to install specific recommended app from "/apps/recommended" page
I am always frustrated when I just want to install Nextcloud Office from recommended apps but I need to install all of them. I know you can install app from "Apps".  But if we have option to check/uncheck particular apps it would be great 
image

There should be option to check/uncheck for individual recommended apps

sanskar-soni-9:
* Resolves: #45084

## Summary
Added functionality to selectively install recommended apps

## Before
240611_14h57m50s_screenshot

## After

https://github.com/nextcloud/server/assets/88997516/9202f245-877c-4b0e-bf50-23b6b7dee143

## Checklist


sanskar-soni-9:
I need opinion for this arrow should I remove it?

240611_15h56m45s_screenshot

susnux:
> I need opinion for this arrow should I remove it?

I would say yes.

sanskar-soni-9:
> I need opinion for this arrow should I remove it?

arrow?!, I meant icon 😅, anyways removed!

sanskar-soni-9:
Now looking at the diff I think it can be there can be more solutions for it, like adding another state variable for selected app ids instead of introducing new property in all apps or something similar, in any case pls let me know if any approach is more preferred here

jancborchardt (changes requested):
Very nice! Just some feedback:
- [ ] All apps should be pre-checked to be selected, at least the ones which are compatible (which on a release should be all of them)
- [ ] If none is selected, the buttons below can be disabled rather than hidden to prevent jumping of the interface
- [ ] Checking and unchecking of the box shouldn’t need a spinner, since the installing/downloading is only in the next step? Checking/unchecking should be instant

Nice work @sanskar-soni-9! :)

sanskar-soni-9:
Thanks @jancborchardt, and checking/unchecking is instant I am using checkbox loading state as an indicator that which app is downloading/installing which I think looks better and cleaner, so I removed previous spinner and installed check icon and overloaded checkbox for every state/step:
1) when user will select app to install checkbox will be blue and enabled
2) then when he will click on install button then loading checkbox will indicate which app is currently getting installed
3) and then when app will be installed the checkbox will be disabled and will indicate that the app is installed

I can always bring back previous spinner and check icon if that is needed?

sanskar-soni-9:
Here is the updated UI flow:

https://github.com/nextcloud/server/assets/88997516/e2edccfa-df82-4b94-92f8-a3f06feba557

Current UI shift is due to installing status line which get added when apps are installing
240615_17h17m10s_screenshot

and I still kept loading checkbox because of the previous reason so please let me know your review on this

jancborchardt (commented):
Awesome, really nice @sanskar-soni-9, and thanks for the explanation! :)

> Current UI shift is due to installing status line which get added when apps are installing

I would recommend to put this line next to the "Install recommended apps" button, or better yet change the button to "Installing apps …", cause up there the line is not really visible, and then also the layout does not shift. :)

sanskar-soni-9:
@jancborchardt done! 😃

jancborchardt (approved):
Looks good design-wise now! :) But has to be checked regarding code.

susnux (approved):
:rocket:

susnux:
Can you rebase onto current master?

sanskar-soni-9:
@susnux done :)

sanskar-soni-9:
Oops I forgot about this and deleted the repo, new PR in progress

PR #45804, opened 2024-06-12
fix(cron): Log long running jobs

ChristophWurst:
* Resolves: production instance with database transactions spanning for days. One theory is that it's a stuck cron job.

## Summary

If cron jobs take a very long time to complete they will start to run in parallel. That's because jobs are only reserved for 12h. Afterwards we just assume that the jobs failed and start the job again. In faulty situations that can lead to more and more server load.

Here is an example:

Bildschirmfoto vom 2024-06-12 09-40-13
Bildschirmfoto vom 2024-06-12 09-40-30
Bildschirmfoto vom 2024-06-12 09-40-35
Bildschirmfoto vom 2024-06-12 09-40-38
Bildschirmfoto vom 2024-06-12 09-40-43
Bildschirmfoto vom 2024-06-12 09-40-48

It looks like a job executes an expensive query over and over. After 12h the query time doubles, after 24h it triplicates, after 36h it quadruples, etc. It's not clear if that is what is really happening.

I've tried to be reasonable with the log level so we don't spam the logs too much:

* Job executes longer than 5m -> debug
* Job executes longer than 20m -> info
* Job executes longer than 1h20m -> warning
* Job executes longer than 5h20m -> error
* Job executes longer than 10h40m -> fatal

## TODO

- [x] Add logging

## Checklist

### DIFF
--- a/cron.php
+++ b/cron.php
@@ -154,14 +154,33 @@ Options:
 			$jobDetails = get_class($job) . ' (id: ' . $job->getId() . ', arguments: ' . json_encode($job->getArgument()) . ')';
 			$logger->debug('CLI cron call has selected job ' . $jobDetails, ['app' => 'cron']);
 
+			$timeBefore = time();
 			$memoryBefore = memory_get_usage();
 			$memoryPeakBefore = memory_get_peak_usage();
 
 			/** @psalm-suppress DeprecatedMethod Calling execute until it is removed, then will switch to start */
 			$job->execute($jobList);
 
+			$timeAfter = time();
 			$memoryAfter = memory_get_usage();
 			$memoryPeakAfter = memory_get_peak_usage();
+
+			$cronInterval = 5 * 60;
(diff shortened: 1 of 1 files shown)


AndyScherzinger (approved):

kesselb (approved):

AndyScherzinger:
@marinofaggiana I don't know, let's try and see if the bot can create them 🤞

AndyScherzinger:
So PR could be created but is incomplete according to the bot https://github.com/nextcloud/server/pull/46706 @marinofaggiana - best you align with @ChristophWurst to have them wrapped up for 25 and 26

ChristophWurst:
Let's try porting from 27, where I have already had to resolve conflicts: https://github.com/nextcloud/server/pull/45855#issuecomment-2247725115.

ChristophWurst:
Worked. @AndyScherzinger @marinofaggiana if you need more backports use stable27 as base

Issue #45657, opened 2024-06-04
Make the interface more compact


marcoambrosini:
Part of  #45751 

There's a lot of whitespace between elements, some elements are too big and the usability of Nextcloud would improve if the UI was overall more compact. 

* We've used so far a margin of 8 pixels between list items, buttons and boundaries, we could cut that down to 4 (--default-grid-baseline).
* We are using a minimum target size of 44 pixels for all clickable elements, while the actual accessibility requirement is much lower than that (24px). Ref: https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html
I know that touchscreens are a concern for these elements, but we can have fallbacks in place for those, or preserve a bigger size for main action targets. 

As an example, this is the range of target sizes in dropbox and google drive. They range from 17px (a link) to 74px (main actions) and a hierarchy is established. Conversely almost all our elements have a 44px minimum target size, even the table header.

<img width="1648" alt="Screenshot 2024-06-10 at 12 46 34" src="https://github.com/nextcloud/server/assets/26852655/4795d513-9e5a-4ba2-8fbf-470feade3af2">
<img width="1606" alt="Screenshot 2024-06-10 at 13 43 58" src="https://github.com/nextcloud/server/assets/26852655/90d68ee4-cde6-4242-9a9b-6383ab888103">

### TODO
**📐 Main work**
- [x] Add new variables https://github.com/nextcloud/server/pull/45818
- [x] Reduce `--border-radius-element` to 8px now that clickable elements are smaller https://github.com/nextcloud/server/pull/45819
- [x] Update component library https://github.com/nextcloud-libraries/nextcloud-vue/pull/5694
- [x] Developer documentation with step-by-step upgrade instructions
- [x] Update the documentation about these changes
- [x] Add it to the App upgrade guide: https://github.com/nextcloud/documentation/blob/master/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.rst
- [x] Post on the forum in the [Development category](https://help.nextcloud.com/c/dev/11) about this update. Good examples are these posts by Daphne: https://help.nextcloud.com/t/on-the-roadmap-deprecation-of-php-7-4/149848 and https://help.nextcloud.com/t/update-on-32-bit-support-decision-to-block-the-update-on-nextcloud-26-reverted/153345

susnux:
Which parts were mentioned? Because I often read exactly the opposite, that the UI is too compact and you end up with a lot of dead space on desktop screens.

---

[...]

PR #45818, opened 2024-06-12
Feat: add new clickable area variables

marcoambrosini:
Add new variables for clickable areas and reduces size of `--default-clickable-area` 

Part of [#45657](https://github.com/nextcloud/server/issues/45657)

### DIFF
--- a/apps/theming/lib/Themes/DefaultTheme.php
+++ b/apps/theming/lib/Themes/DefaultTheme.php
@@ -172,7 +172,9 @@ class DefaultTheme implements ITheme {
 			// pill-style button, value is large so big buttons also have correct roundness
 			'--border-radius-pill' => '100px',
 
-			'--default-clickable-area' => '44px',
+			'--default-clickable-area' => '32px',
+			'--clickable-area-large' => '48px',
+			'--clickable-area-small' => '24px',
 			'--default-line-height' => '24px',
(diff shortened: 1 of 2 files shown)


artonge (approved):

susnux (approved):
Maybe keep the large clickable area at 44px, but either way ok for me :)

apps/theming/lib/Themes/DefaultTheme.php
```diff
@@ -172,7 +172,9 @@ public function getCSSVariables(): array {
 			// pill-style button, value is large so big buttons also have correct roundness
 			'--border-radius-pill' => '100px',
 
-			'--default-clickable-area' => '44px',
+			'--default-clickable-area' => '32px',
```
susnux, line 175:
```suggestion
			'--default-clickable-area' => '34px',
```

AndyScherzinger (approved):

jancborchardt (commented):
Should clickable-area-large not be 44px, so we don't make elements even bigger?

marcoambrosini:
@jancborchardt sorry I missed your comment, I thought it'd be good to boost it a bit since it would be only one button per app basically. This would be also in line with what google and dropbox do. But we can see how it looks and dial it back before release in case it's too much

PR #45819, opened 2024-06-12
Feat: reduce border radius element to 8 px

marcoambrosini:
part of [#45657](https://github.com/nextcloud/server/issues/45657)

With smaller size clickable elements, we also need to dial down the border radius a bit.


artonge (approved):

susnux (approved):
Makes sense, for large buttons we then should use `--border-radius-large`.

jancborchardt (approved):

PR #45832, opened 2024-06-12
chore: Adjust post-build script to add license files for sourcemaps on build

susnux:
## Summary

sourcemap and js share the same license, so we can simply create a copy from the license files,
or like here to save some space use symlinks.

## Checklist


AndyScherzinger (approved):
Nice work @susnux 🎉

build/npm-post-build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Build CSS files from SCSS
+npm run sass
+# Build icons
+npm run sass:icons
+
+# Add licenses for source maps
+if [ -d "dist" ]; then
+	for f in dist/*.js; do
+		# If license file and source map exists copy license for the sourcemap
```
nickvergessen, line 16:
```suggestion
		# If license file and source map exists copy license for the source map
```

nickvergessen (approved):

PR #45833, opened 2024-06-12
chore: Properly extract copyright information from bundled packages

susnux:
## Summary

Currently we only extract comments, this does not work as most dependencies are either minified or do not contain license headers.
So this instead adds a webpack plugin to iterate all bundled sources, extract the license information and create a proper `.license` file.

I really hope to drop this when migrating to vite :sweat_smile: 

## Checklist


AndyScherzinger:
Hi @carmenbianca - just wanted to get some feedback (given you have some time for it of course). 😃 

During the quest of making the Nextcloud server REUSE compliant we faces Javascript compiled assets - which are combined from various sources - welcome to modern web development... - so they come with various licenses in some scenarios, like 

https://github.com/nextcloud/server/pull/45833/files#diff-efb7ecbf3b0f9ef7721e44bea59a37e4a447e4f64e190d21dcc61befead2b719

as one example. Given the licenses are compatible would the `AND` operator be the right choice? I guess so, but just wanted to be sure, since I personally got it wrong on the first try of the Android repo and than had to fix it afterwards, thanks to your feedback.

carmenbianca:
@AndyScherzinger If I understand correctly, the source code from multiple files is mashed into a single file?

`AND` is a safe bet in that case.

Alternatively, instead of doing `AND`, you can define the `SPDX-License-Identifier:` tag several times, once for each source file. I'd have to re-read the SPDX spec, but these tags should be implicitly `AND`-ed.

AndyScherzinger:
@carmenbianca yes, that is correct.

So @susnux than we can keep the current logic for the moment.

Of course it would still be interesting to know if a list of license ID lines would implicitly be `AND`ed.

AndyScherzinger (approved):

skjnldsv (approved):

package.json
```diff
@@ -188,6 +189,7 @@
     "webpack": "^5.91.0",
     "webpack-cli": "^5.0.2",
     "webpack-merge": "^5.8.0",
+    "webpack-stats-plugin": "^1.1.3",
```
susnux, line 192:
```suggestion
```

AndyScherzinger:
Woohoo, all green @susnux 🎉

Now I will have to find the wrong license identifier sources and how to get the license comment into some dist JS file where there is none.

Again thanks a lot for your contributions in getting this topic done for the server repo! Makes my days 💯🎉🙏😊👍💪

PR #33729, opened 2022-08-28
Add regex to match Gnome and KDE calendar user-agents, for propagating calendar subscriptions

msrn:
Refers: #17754 
This pull requests adds Gnome Calendar/Evolution and KDE PIM/Akonadi calendars to the list of user-agents in WebCalCaching -plugin, and exposes subscribed calendars to these calendar clients.

User-agent of Gnome Calendar/Evolution
`Evolution/3.40.3`

KDE PIM/Akonadi user-agent
`Mozilla/5.0 (X11; Linux x86_64) KIO/5.86 akonadi_davgroupware_resource_1/5.18.2 (21.08.2)`

PR #45841, opened 2024-06-12
feat(dav): add regex to match Gnome and KDE calendar user-agents

### LINKED ISSUE #17754: Propagate subscribed calendars to CalDAV clients
I'm using the Nextcloud Calendar mostly in desktop/mobile clients. I'm subscribed to several remote calendars via ical. I can see them in the web interface, but I can't in the clients.

It would be great if the calendars I'm subscribed to got propagated to CalDAV clients just like calendars that are located in my Nextcloud.

The alternative is to subscribe to the calendars in each client, but that's additional work in the initial setup and maintenance. I'd rather set them up in one location for all clients.

**Additional context**
Proprietary competition (Google Calendar, iCloud) supports it.

Vivida1:
* Resolves: #17754

## Summary
Original Pull Request by @msrn: https://github.com/nextcloud/server/pull/33729 which became stale.

Quote: "This pull requests adds Gnome Calendar/Evolution and KDE PIM/Akonadi calendars to the list of user-agents in WebCalCaching -plugin, and exposes subscribed calendars to these calendar clients.

User-agent of Gnome Calendar/Evolution
Evolution/3.40.3

KDE PIM/Akonadi user-agent
Mozilla/5.0 (X11; Linux x86_64) KIO/5.86 akonadi_davgroupware_resource_1/5.18.2 (21.08.2)"

### DIFF
--- a/apps/dav/lib/CalDAV/WebcalCaching/Plugin.php
+++ b/apps/dav/lib/CalDAV/WebcalCaching/Plugin.php
@@ -23,7 +23,7 @@ class Plugin extends ServerPlugin {
 	 * that do not support subscriptions on their own
 	 *
 	 * /^MSFT-WIN-3/ - Windows 10 Calendar
-     * /Evolution/ - Gnome Calendar/Evolution
+	 * /Evolution/ - Gnome Calendar/Evolution
 	 * /KIO/ - KDE PIM/Akonadi
 	 * @var string[]
 	 */
--- a/apps/dav/lib/CalDAV/WebcalCaching/Plugin.php
+++ b/apps/dav/lib/CalDAV/WebcalCaching/Plugin.php
@@ -23,10 +23,14 @@ class Plugin extends ServerPlugin {
 	 * that do not support subscriptions on their own
 	 *
 	 * /^MSFT-WIN-3/ - Windows 10 Calendar
+     * /Evolution/ - Gnome Calendar/Evolution
+	 * /KIO/ - KDE PIM/Akonadi
 	 * @var string[]
 	 */
 	public const ENABLE_FOR_CLIENTS = [
-		"/^MSFT-WIN-3/"
+		"/^MSFT-WIN-3/",
+		"/Evolution/",
+		"/KIO/"
 	];
 
 	/**


miaulalala (approved):
I agree with Barth https://github.com/nextcloud/server/pull/33729#issuecomment-2121922691

msrn:
Thanks for opening the PR again! I have been busy so wasn't able to work on this. I originally envisioned that I would test the user agents again if they still work, because there might have been user agent changes on those calendar client during these two years.

skjnldsv (approved):

tcitworld:
Still conflicted with what I said on the topic in https://github.com/nextcloud/server/pull/33729#pullrequestreview-1088174080, but I don't mind much.

However, we might as well include all clients but DAVx⁵, since it's the only one supporting subscription collection discovery (and others did not make progress on this topic).

A drawback that might get observed is that the refresh rate (long by default on NC side, probably very short on client side) might be an issue for most people.

skjnldsv:
I'll let @Vivida1 decide :+1:

Vivida1:
I think reversing the logic makes more sense from a maintainability standpoint.

@tcitworld, can you explain the refresh rate drawback?

I can prepare a commit but for that I need to find the DAVx⁵ user-agent first.

msrn:
> I think reversing the logic makes more sense from a maintainability standpoint.
> 
> @tcitworld, can you explain the refresh rate drawback?
> 
> I can prepare a commit but for that I need to find the DAVx⁵ user-agent first.

Not 100% sure, but it might be just `DAVx5` as per
https://github.com/bitfireAT/davx5-ose/blob/1d7084b55533b7000fd049f3d97f92d6ee748a59/app/build.gradle.kts#L31

tcitworld:
> @tcitworld, can you explain the refresh rate drawback?

The Nextcloud server caches the subscriptions (stores data in the database) and exposes them as regular calendars. The default refresh rate for this [is of one week](https://docs.nextcloud.com/server/latest/admin_manual/groupware/calendar.html#refresh-rate), which is appropriate for things like public holiday calendars, but might not be for other things, for instance subscribing to a public calendar from another Nextcloud user, and expecting real-time sync.

This high refresh rate makes sense because on servers with a lot of users and subscriptions, the cron would keep wasting time refreshing calendars for nothing, and there could be a risk of the subscription source blocking the server. For instance, as subscriptions can't be shared (or provided by the admin), if all 10 000 users from a company need to subscribe to the holiday calendar from their local government, that's 10 000 requests you need to make every X minutes/hours/days when the original calendar data only changes once or twice a year.

Nextcloud supports properties available for the subscription source to tell what's the appropriate refresh rate ([`X-PUBLISHED-TTL`, `REFRESH-INTERVAL`](https://github.com/nextcloud/server/blob/b5275f703537f01fa27b042e1b619f56a223ab49/apps/dav/lib/CalDAV/WebcalCaching/RefreshWebcalService.php#L255-L285)), but in practice no-one provides them.

I think there's an issue about providing the setting to select an appropriate refreshment rate for each subscription to end-users in calendar, which would kinda solve most of the issue, though there's a risk users could abuse it, so a minimal rate setting for the admin would be nice as well. In any case, that's not the current state.

So on the other hand, clients have no issue with refreshing the local subscription every 15 minutes/hour/day or even each time the app is opened, as it's only doing it for a single user and from a dedicated IP address.

Therefore, people who would have been forced to manually add the subscription on their device before this change and will now just used the exposed subscriptions from the NC server may have the impression that subscriptions don't get updated when they expect it.

miaulalala:
> > @tcitworld, can you explain the refresh rate drawback?
> 
> The Nextcloud server caches the subscriptions (stores data in the database) and exposes them as regular calendars. The default refresh rate for this [is of one week](https://docs.nextcloud.com/server/latest/admin_manual/groupware/calendar.html#refresh-rate), which is appropriate for things like public holiday calendars, but might not be for other things, for instance subscribing to a public calendar from another Nextcloud user, and expecting real-time sync.
> 
> This high refresh rate makes sense because on servers with a lot of users and subscriptions, the cron would keep wasting time refreshing calendars for nothing, and there could be a risk of the subscription source blocking the server's IP. For instance, as subscriptions can't be shared (or provided by the admin), if all 10 000 users from a company need to subscribe to the holiday calendar from their local government, that's 10 000 requests you need to make every X minutes/hours/days when the original calendar data only changes once or twice a year.
> 
> Nextcloud supports properties available for the subscription source to tell what's the appropriate refresh rate ([`X-PUBLISHED-TTL`, `REFRESH-INTERVAL`](https://github.com/nextcloud/server/blob/b5275f703537f01fa27b042e1b619f56a223ab49/apps/dav/lib/CalDAV/WebcalCaching/RefreshWebcalService.php#L255-L285)), but in practice no-one provides them.
> 
> I think there's an issue about providing the setting to select an appropriate refreshment rate for each subscription to end-users in calendar, which would kinda solve most of the issue, though there's a risk users could abuse it, so a minimal rate setting for the admin would be nice as well. In any case, that's not the current state.
> 
> So on the other hand, clients have no issue with refreshing the local subscription every 15 minutes/hour/day or even each time the app is opened, as it's only doing it for a single user and from a dedicated IP address.
> 
> Therefore, people who would have been forced to manually add the subscription on their device before this change and will now just used the exposed subscriptions from the NC server may have the impression that subscriptions don't get updated when they expect it.

If we were to improve how we refresh subscriptions, we might be able to not care as much about the refresh rate - IIRC we're not using a delta but completely deleting and re- importing all events. It's not going to be easy to implement but I definitely want to do that at some point.

What we could also do for subscriptions is to compare the URLs for subscription calendars and do an update on all calendars that share the same subscription URLs in a single run.

PR #45848, opened 2024-06-13
Fix/comment deleting with activities installed

GVodyanov:
## Summary
Fixes issue where you:
- Delete a comment
- Add a new comment before the old comment gets deleted in the back-end
- See deleted comment again

## TODO
- [x] Add store for synchronizing deleted comments between the mixin and activities 

## Checklist


ShGKme (commented):
I don't think it makes sense to introduce a global storage in the app just to store ids or deleted comments in 2-3 components.

If storing `deleted` state in the comment it self is a problem, it's enough to store it in the parent.

Besides, it seems that the initial issue is that adding a comment triggers an entire tab reload through loading state. What's when `Comment` instances are being destroyed and re-mounted without old `deleted` state.

GVodyanov:
@ShGKme Would you have any suggestions on how to not make the tab reload then, I'm having trouble figuring out the activities API

ShGKme:
So, I made a deeper look at the app.

With activity app case it is indeed not possible to store data in the parent, because on the Activity tab there is no parent between comments and the form. Every comment is a new Vue app. So, as a simple workaround from the server side, it is a good workaround.

A better solution would be to adjust Activity API to not reload everything on change. But with the current approach, it is not so simple.

So, let's fix Ferdinand's suggestion and go with it for now

cc @GretaD

susnux:
> A better solution would be to adjust Activity API to not reload everything on change. But with the current approach, it is not so simple.

:100: agree

susnux (approved):

ChristophWurst (approved):

ChristophWurst:
The branch is outdated. Other PRs touched the frontend bundles. It will need a rebase and a recompile.

susnux:
@GVodyanov 

> /home/runner/actions-runner/_work/server/server/apps/comments/src/store/deletedCommentLimbo.js
> Error:    6:1  error    Expected space or tab after '//' in comment  spaced-comment

ChristophWurst:
conflicts -> needs rebase + rebuild

susnux:
> The backport to `stable28` failed. Please do this backport manually.

When doing the backports @GVodyanov please directly include https://github.com/nextcloud/server/pull/46643

PR #45788, opened 2024-06-11
refactor(files): Use `@nextcloud/files` functions where possible and make files views cancelable

susnux:
* (by using nextcloud-files) Resolves: https://github.com/nextcloud/server/issues/45723

## Summary
1. Use the functions from `@nextcloud/files` where possible, replaces a lot of duplicated code in the files views.
2. Make the `getContents` functions cancelable to possibly save some network requests
3. Fix the favorites :star: view by adding a fake root like for recent etc, because you shall not be able to upload to the "favorites" root.

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #45860, opened 2024-06-13
fix(files): Bring back handling of failed files

susnux:
* Follow up (broken by) #45788 

## Summary
The fileid might be set to -1 in case of an error, this brings back proper handling of this case.

## Checklist


susnux:
Best with, but no blocker: https://github.com/nextcloud-libraries/nextcloud-files/pull/985

apps/files/src/components/FileEntry/FileEntryActions.vue
```diff
@@ -152,7 +153,7 @@ export default defineComponent({
 
 		// Sorted actions that are enabled for this node
 		enabledActions() {
-			if (this.source.attributes.failed) {
+			if (this.source.status === NodeStatus.FAILED) {
```
Fenn-CS, line 155: We can't use `isFailedSource()` from the mixin here?

nfebe (approved):

apps/files/src/components/FileEntry/FileEntryActions.vue
susnux: Not without passing it from the file entry as a prop to the actions.
But I do not think it makes much sense to do so as we can easily (this one line) get it from the source.

artonge (approved):

PR #45861, opened 2024-06-13
fix(files): Ignore right click for default action

artonge:
Follow up of https://github.com/nextcloud/server/pull/45571


skjnldsv (approved):

skjnldsv (approved):

.htaccess
```diff
@@ -108,3 +108,7 @@
 
 AddDefaultCharset utf-8
 Options -Indexes
+#### DO NOT CHANGE ANYTHING ABOVE THIS LINE ####
```
susnux, line 111: This should not be committed

apps/files/src/components/FileEntryMixin.ts
```diff
@@ -208,6 +211,11 @@ export default defineComponent({
 		},
 
 		execDefaultAction(event) {
+			// Ignore right click.
+			if (event.button === 2) {
```
susnux, line 215: Should we ignore all non-main buttons? I mean if I press my tertiary button I would not expect the action to be performed.
The only button I am not sure about is middle-mouse button for opening in new tab?

```suggestion
			if (event.button != 2) {
```

.htaccess
artonge: Arg, thanks

apps/files/src/components/FileEntryMixin.ts
artonge: From what I understand,
0 is left
1 is middle
2 is right
3 and above is whatever

`if (event.button >= 1) {` ?

apps/files/src/components/FileEntryMixin.ts
```diff
@@ -208,6 +208,11 @@ export default defineComponent({
 		},
 
 		execDefaultAction(event) {
+			// Ignore right click.
+			if (event.button > 1) {
+				retrurn
```
nickvergessen, line 213:
```suggestion
				return
```

nickvergessen (dismissed):

apps/files/src/components/FileEntryMixin.ts
artonge: :face_exhaling:

blizzz (approved):

apps/files/src/components/FileEntryMixin.ts
susnux: Makes sense

skjnldsv (approved):

PR #45869, opened 2024-06-14
feat: Update user list on guest user events

Pytal:
## Summary

- Requires https://github.com/nextcloud/guests/pull/1174

Needed to update the user list without a page reload

## Checklist

### DIFF
--- a/apps/settings/src/components/UserList.vue
+++ b/apps/settings/src/components/UserList.vue
@@ -277,6 +277,9 @@ export default {
 		 * If disabled group but empty, redirect
 		 */
 		await this.redirectIfDisabled()
+
+		subscribe('guests:user:deleted', (userId) => this.$store.commit('deleteUser', userId))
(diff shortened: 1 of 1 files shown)


artonge (approved):

apps/settings/src/components/UserList.vue
```diff
@@ -277,6 +277,9 @@ export default {
 		 * If disabled group but empty, redirect
 		 */
 		await this.redirectIfDisabled()
+
+		subscribe('guests:user:deleted', (userId) => this.$store.commit('deleteUser', userId))
+		subscribe('guests:user:created', (userId) => this.$store.dispatch('addUserData', userId))
```
artonge, line 282: Should we really listen to external app even in server? Can't we emit a generic event from the guest app?

artonge (changes requested):

apps/settings/src/components/UserList.vue
Pytal: Agreed, something like `nextcloud:user:created`?

PR #45875, opened 2024-06-14
fix(theming): also apply enforced theme for guests

### LINKED ISSUE #41048: Theming: Force Light or Dark Theme Regardless of System
Uniform design of our Nextcloud web interface is important to us. Unfortunately, Nextcloud consistently looks entirely different depending on the device settings, especially for unregistered users with whom we want to share data.

We desire a switch in the administration settings of Nextcloud that enforces either the light or dark theme of Nextcloud for both unauthenticated and authenticated users, regardless of device settings.

No alternative ideas. The default, current behavior of Nextcloud should not be influenced. For an alternative solution, such as using custom CSS rules, I would also be grateful.

**Additional context**
Nothing to add

skjnldsv:
Fix https://github.com/nextcloud/server/issues/41048

### DIFF
--- a/apps/theming/lib/Service/ThemesService.php
+++ b/apps/theming/lib/Service/ThemesService.php
@@ -16,24 +16,22 @@ use OCA\Theming\Themes\LightTheme;
 use OCP\IConfig;
 use OCP\IUser;
 use OCP\IUserSession;
+use Psr\Log\LoggerInterface;
 
 class ThemesService {
-	private IUserSession $userSession;
-	private IConfig $config;
-
 	/** @var ITheme[] */
 	private array $themesProviders;
 
-	public function __construct(IUserSession $userSession,
-		IConfig $config,
-		DefaultTheme $defaultTheme,
+	public function __construct(
+		private IUserSession $userSession,
+		private IConfig $config,
+		private LoggerInterface $logger,
+		private DefaultTheme $defaultTheme,
 		LightTheme $lightTheme,
 		DarkTheme $darkTheme,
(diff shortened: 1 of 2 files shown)


Oriann:
Any chance this will be backported to NC version 28 and 29 ?

apps/theming/lib/Service/ThemesService.php
```diff
@@ -26,7 +26,7 @@ class ThemesService {
 
 	public function __construct(IUserSession $userSession,
 		IConfig $config,
```
susnux, line 28: Maybe then also refactor them if you do it for the default theme :)

```suggestion
	public function __construct(
		private IUserSession $userSession,
		private IConfig $config,
```

apps/theming/lib/Service/ThemesService.php
```diff
@@ -52,6 +52,17 @@ public function __construct(IUserSession $userSession,
 	 * @return ITheme[]
 	 */
 	public function getThemes(): array {
+		// Enforced theme if configured
+		$enforcedTheme = $this->config->getSystemValueString('enforce_theme', '');
+		if ($enforcedTheme !== '' && isset($this->themesProviders[$enforcedTheme])) {
```
susnux, line 57: If the `isset` fails it should at least log a warning that the config is invalid for admins to figure out whats wrong.

AndyScherzinger (approved):

susnux (approved):

Oriann:
PR broken in NC version 29.0.7

https://github.com/nextcloud/server/issues/41048#issuecomment-2385236674

PR #45884, opened 2024-06-14
fix(files): Properly handle denied ownership transfers

susnux:
* Resolves https://github.com/nextcloud/server/issues/45753

## Summary
When the receiver denies the transfer the notification handler was missing, so no notification was created for the transfer owner.

But also the internal notification was created two times:
1. When rejecting the transfer
2. By the reject function when dismissing the notification

This is fixed by only relying on the dismiss function.

---

image

## Checklist


SystemKeeper (approved):
Tested and works: 
image

come-nc (approved):

botsarenthuman:
Is there a test to cover this actually works?

PR #45922, opened 2024-06-17
fix(files_sharing): global search in sharing tab

ShGKme:
* Resolves: clicking **Global Search** breaks the sharing tab

## Summary

- Probably a regression from new flow: https://github.com/nextcloud/server/pull/39472
- In the past, on select it (in `addShare`):
  1. Checked if it was not the `lookup` = `global search` option
  2. Checked if it was not the `external` with custom handler option
  3. Created a share with default params
- Then it was moved to a new create share details view
  - `onSelect` now simply opens the details view where the share can be created
  - p.2 is also moved to the `SharingDetails`
  - p.1 was missed

## TODO

- [x] Migrate to new built-in `NcSelect` label
  image
- [x] Remove old now unused `addShare` method
- [x] Restore `lookup` select processing

## Screenshots 

🏚️ Before | 🏡 After
---|---
global-search-sharing-before | global-search-sharing-after 

## Checklist


susnux (approved):

apps/files_sharing/src/components/SharingInput.vue
```diff
@@ -138,8 +136,20 @@ export default {
...
+
+			// Not a real option, but a lookup = global search button
+			if (option.lookup) {
+				await this.getSuggestions(this.query, true)
+				// Wait until select is re-rendered with new options
+				await this.$nextTick()
+				// Open the select again
+				this.$refs.select.$children[0].open = true
+
+				return true
```
Pytal, line 150:
```suggestion
				return
```
nit: Unused return value

Pytal (approved):

apps/files_sharing/src/components/SharingInput.vue
ShGKme: Done

ShGKme:
Rebased onto `master` to resolve conflict and removed unused return value `true`

apps/files_sharing/src/components/SharingInput.vue
```diff
@@ -5,10 +5,9 @@
 
 <template>
 	<div class="sharing-search">
-		<label for="sharing-search-input">{{ t('files_sharing', 'Search for share recipients') }}</label>
 		<NcSelect ref="select"
 			v-model="value"
-			input-id="sharing-search-input"
+			:input-label="t('files_sharing', 'Search for share recipients')"
```
susnux, line 10: This breaks cypress tests.
I agree we do not need the ID but then you need to change the selector in the Cypress tests.

PR #45930, opened 2024-06-17
fix: move repair mimetype repair step to the expensive steps

icewind1991:
Looking through the entire filecache takes a while


juliusknorr (approved):

ChristophWurst (approved):

kesselb:
Not ideal. 

Adding mime types to existing files can be important to make certain functionality work. 

Could you add a hint to `lib/private/Repair/RepairMimeTypes.php` that any update to the file must be mentioned in our release notes so that admin's know they have to run occ maintenance:repair --include-expensive?

kesselb:
> * Added a separate "version" to track the ran mimetype migrations
> * Added a setup check for when mimetype repair needs to be ran

Thank you :+1:

juliusknorr (commented):
Ideally we would split the operations that are performed for the new mimetype:

The insert into oc_mimetypes could be still in regular migration steps as it is just one insert for a new mimetype, that way it will already work for new files: https://github.com/nextcloud/server/blob/repair-mimetype-expensive/lib/private/Repair/RepairMimeTypes.php#L42-L44

Otherwise 👍

juliusknorr (approved):

icewind1991:
redoing the other backports based on the 29 one

PR #45947, opened 2024-06-18
feat: Add yieldEntities wrapper for entity mapping in QBMapper

juliusknorr:
This could be a nice small addition to the QBMapper API to allow apps iterating over results in a more memory efficient way that does not need to keep the full result set in memory.

## Checklist

### DIFF
--- a/lib/public/AppFramework/Db/QBMapper.php
+++ b/lib/public/AppFramework/Db/QBMapper.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
  */
 namespace OCP\AppFramework\Db;
 
+use Generator;
 use OCP\DB\Exception;
 use OCP\DB\QueryBuilder\IQueryBuilder;
 use OCP\IDBConnection;
@@ -316,6 +317,26 @@ abstract class QBMapper {
 				$entities[] = $this->mapRowToEntity($row);
 			}
 			return $entities;
+		} finally {
+			$result->closeCursor();
+		}
+	}
+
+	/**
+	 * Runs a sql query and yields each resulting entity to obtain database entries in a memory-efficient way
+	 *
+	 * @param IQueryBuilder $query
+	 * @return Generator Generator of fetched entities
+	 * @psalm-return Generator<T> Generator of fetched entities
+	 * @throws Exception
+	 * @since 30.0.0
+	 */
+	protected function yieldEntities(IQueryBuilder $query): Generator {
+		$result = $query->executeQuery();
+		try {
+			while ($row = $result->fetch()) {
+				yield $this->mapRowToEntity($row);
+			}
 		} finally {
 			$result->closeCursor();
 		}


lib/public/AppFramework/Db/QBMapper.php
```diff
@@ -321,6 +322,26 @@ protected function findEntities(IQueryBuilder $query): array {
 		}
 	}
 
+	/**
+	 * Runs a sql query and returns an array of entities
```
nickvergessen, line 326: Does it? :P

ChristophWurst (approved):

lib/public/AppFramework/Db/QBMapper.php
```diff
@@ -321,6 +322,26 @@ protected function findEntities(IQueryBuilder $query): array {
...
+	 * @param IQueryBuilder $query
+	 * @return Generator Generator of fetched entities
+	 * @psalm-return Generator<T> Generator of fetched entities
+	 * @throws Exception
+	 * @since 30.0.0
+	 */
+	protected function yieldEntities(IQueryBuilder $query): Generator {
+		$result = $query->executeQuery();
+		try {
+			while ($row = $result->fetch()) {
```
come-nc, line 337:
```suggestion
			while (($row = $result->fetch()) !== false) {
```
0, "" or '0' are valid values.

lib/public/AppFramework/Db/QBMapper.php
```diff
@@ -321,6 +322,26 @@ protected function findEntities(IQueryBuilder $query): array {
...
+				yield $this->mapRowToEntity($row);
+			}
+		} finally {
+			$result->closeCursor();
+		}
```
come-nc, line 342: Does that work? :exploding_head: 
I always wanted to implement that but thought it was not a good idea because there was no way to call closeCursor if the generator is not entirely consumed.

come-nc: https://3v4l.org/8rS4U
Seems to work nicely, nice trick!

lib/public/AppFramework/Db/QBMapper.php
nickvergessen: but for entities it has to be arrays? :P

juliushaertl: Yep, we use `PDO::FETCH_ASSOC` as the default mode which returns `array<string,mixed>|false`

come-nc: Hum okay I missed that it was specific to entities.
Should we directly add a `$result->yieldRows()`?

How about an empty array, is that a valid row value?

I would still prefer the strict comparison to false, because this kind of code gets copy/pasted around and then we get bugs.

nickvergessen: > How about an empty array, is that a valid row value?

An entity without data?

nickvergessen (approved):

come-nc (approved):

Issue #45918, opened 2024-06-17
login field placeholders look weird


szaimen:
Hi, just tested on daily.ltd2.nextcloud.com: the login field look weird:
![image](https://github.com/nextcloud/server/assets/42591237/2d2aa76b-109c-46b7-bb14-cdd8468fbc31)
cc @nextcloud/server-frontend @nextcloud/designers

susnux:
Caused by the clickable area size and needs this to be fixed:
https://github.com/nextcloud-libraries/nextcloud-vue/pull/5694

PR #45948, opened 2024-06-18
Fix the placeholder look

rjayswal:
Fix the issue #45918

**Before fixed**
image

**After fixed**
image


jancborchardt:
This was already fixed via the components https://github.com/nextcloud/server/issues/45918#issuecomment-2173214874

PR #45951, opened 2024-06-18
chore: Move comments event handler to use proper event dispatcher

juliusknorr:
Registering comments event handlers through the manager requires the comment manager to be initialized during app boot, which is something we can avoid by using the existing event dispatcher to make the registration more lazy.

## Checklist


apps/comments/lib/Listener/CommentsEventListener.php
```diff
@@ -0,0 +1,74 @@
+<?php
```
nickvergessen, line 1: declare strict + SPDX header?

blizzz (approved):

nickvergessen (approved):

PR #45953, opened 2024-06-18
chore: Add event to register mount providers more lazy

juliusknorr:
Mount provider registration currently requires to obtain the MountProviderCollection and create instances of apps mount providers, which might pull in more heavy dependencies. While this is not a problem in files_sharing, other apps like circles or collectives have a larger tree of dependencies that the mount provider class is pulling in.

We don't need the mount providers always available so we can save those DI setup times for all requests that don't access the mount provider list through accessing the storage.

## Checklist

### DIFF
--- a/lib/private/Files/Config/MountProviderCollection.php
+++ b/lib/private/Files/Config/MountProviderCollection.php
@@ -10,11 +10,13 @@ namespace OC\Files\Config;
 use OC\Hooks\Emitter;
 use OC\Hooks\EmitterTrait;
 use OCP\Diagnostics\IEventLogger;
+use OCP\EventDispatcher\IEventDispatcher;
 use OCP\Files\Config\IHomeMountProvider;
 use OCP\Files\Config\IMountProvider;
 use OCP\Files\Config\IMountProviderCollection;
 use OCP\Files\Config\IRootMountProvider;
 use OCP\Files\Config\IUserMountCache;
+use OCP\Files\Events\RegisterMountProviderEvent;
 use OCP\Files\Mount\IMountManager;
 use OCP\Files\Mount\IMountPoint;
 use OCP\Files\Storage\IStorageFactory;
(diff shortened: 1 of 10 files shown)


lib/public/Files/Events/RegisterMountProviderEvent.php
```diff
@@ -0,0 +1,45 @@
...
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
```
kesselb, line 21:
```suggestion
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
 */
```

juliushaertl: Seems I haven't updated that IDE config yet 👍

lib/private/Files/Config/MountProviderCollection.php
```diff
@@ -230,12 +216,17 @@ public function getRootMounts(): array {
 	}
 
 	public function clearProviders() {
+		$this->$this->registerEventEmitted = false;
```
come-nc, line 219:
```suggestion
		$this->registerEventEmitted = false;
```
I suppose?

come-nc (commented):
An other solution would be to register provider classes and let the mountprovidercollection get them from DI when needed?

Only problem is if an app want to register instances with parameters in the constructor, or even several instances of the same class with different config. In this situation your solution works better.

icewind1991 (approved):

PR #45963, opened 2024-06-19
Rename openFileInfo to fileInfo

artonge:
And check if openFile is set in the URL before opening a file.


skjnldsv (approved):

skjnldsv:
@ShGKme because we talked about this iirc

apps/files/src/components/FilesListVirtual.vue
```diff
@@ -207,7 +207,7 @@ export default defineComponent({
 		mainContent.addEventListener('dragover', this.onDragOver)
 
 		// handle initially opening a given file
-		const { id } = loadState<{ id?: number }>('files', 'openFileInfo', {})
+		const { id } = loadState<{ id?: number }>('files', 'fileInfo', {})
```
ShGKme, line 210: Comment is not actual anymore 👀

ShGKme (approved):
Thanks 💙

In which case do we have no `fileid` from route and initial state is required? Except the public sharing.

artonge:
> In which case do we have no `fileid` from route and initial state is required? Except the public sharing.

I don't remember, but I think this use case does not exists anymore with F2V.

PR #45967, opened 2024-06-19
Fix rmdir() to release possible directory lock prior to removing directory

dennisverspuij:
As described in this [Nextcloud community issue](https://help.nextcloud.com/t/unable-to-delete-directories-text-file-busy/86468) some storage systems (such as VirtualBox shared folders) may lock a directory when traversing it using `opendir()` or a `DirectoryIterator`. This patch fixes `Local::rmdir()` to release such locks prior to removing a directory recursively, avoiding failure to do so with 'Text file busy" warnings.

### DIFF
--- a/lib/private/Files/Storage/Local.php
+++ b/lib/private/Files/Storage/Local.php
@@ -117,6 +117,7 @@ class Local extends \OC\Files\Storage\Common {
 				}
 				$it->next();
 			}
+			unset ($it);
 			clearstatcache(true, $this->getSourcePath($path));
 			return rmdir($this->getSourcePath($path));
 		} catch (\UnexpectedValueException $e) {
--- a/lib/private/Files/Storage/Local.php
+++ b/lib/private/Files/Storage/Local.php
@@ -117,7 +117,7 @@ class Local extends \OC\Files\Storage\Common {
 				}
 				$it->next();
 			}
-			unset ($it);
+			unset($it);
 			clearstatcache(true, $this->getSourcePath($path));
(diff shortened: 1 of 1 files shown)


lib/private/Files/Storage/Local.php
```diff
@@ -117,6 +117,7 @@ public function rmdir($path) {
 				}
 				$it->next();
 			}
+			unset ($it);
```
solracsf, line 120:
```suggestion
			unset($it);
```

lib/private/Files/Storage/Local.php
```diff
@@ -117,6 +117,7 @@ public function rmdir($path) {
 				}
 				$it->next();
 			}
+			unset($it);
```
come-nc, line 120: That needs a comment to explain what it does.

dennisverspuij: Please see first commit message, I do not understand what else I need to explain about unsetting the iterator to free the directory lock.

come-nc: You need to explain it in the code with a comment.
Otherwise this will get removed later or confuse future code readers.

dennisverspuij: Sorry I seem to have closed and  messed up this PR, don;t know how I did that.
However, I reset the branch with a single proper commit with the desired comment, can you reopen this PR or should I create a new one with it?

dennisverspuij: Never mind, I created a new PR.

PR #45973, opened 2024-06-19
chore: remove chunking-v1

icewind1991:
This was originally removed in https://github.com/nextcloud/server/pull/36589 but that got reverted after other changes from that PR were causing issues.

I've split this off to make any future attempt at removing the "cache gc"/"setup filesystem on every basic auth" easier to consider in the future.


come-nc (approved):
:+1:

juliusknorr (approved):

JulienFS:
thanks a lot, it really makes the project easier to dive in :raised_hands:

PR #45974, opened 2024-06-19
fix(caldav): lower scheduling table size warning

miaulalala:
* Resolves: # 

## Summary
 Reduce the size warning - the table is quite big even with just a 100k rows.

## Checklist

### DIFF
--- a/apps/settings/lib/SetupChecks/SchedulingTableSize.php
+++ b/apps/settings/lib/SetupChecks/SchedulingTableSize.php
@@ -14,6 +14,8 @@ use OCP\SetupCheck\ISetupCheck;
 use OCP\SetupCheck\SetupResult;
 
 class SchedulingTableSize implements ISetupCheck {
+	public const MAX_SCHEDULING_ENTRIES = 50000;
+
 	public function __construct(
 		private IL10N $l10n,
 		private IDBConnection $connection,
@@ -36,9 +38,11 @@ class SchedulingTableSize implements ISetupCheck {
 		$count = $query->fetchOne();
 		$query->closeCursor();
 
-		if ($count > 500000) {
+		if ($count > self::MAX_SCHEDULING_ENTRIES) {
 			return SetupResult::warning(
-				$this->l10n->t('You have more than 500 000 rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive')
+				$this->l10n->t('You have more than %s rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive.', [
+					self::MAX_SCHEDULING_ENTRIES,
+				])
 			);
 		}
 		return SetupResult::success(


nickvergessen (approved):

apps/settings/lib/SetupChecks/SchedulingTableSize.php
```diff
@@ -36,9 +36,9 @@ public function run(): SetupResult {
 		$count = $query->fetchOne();
 		$query->closeCursor();
 
-		if ($count > 500000) {
+		if ($count > 50000) {
 			return SetupResult::warning(
-				$this->l10n->t('You have more than 500 000 rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive')
+				$this->l10n->t('You have more than 50 000 rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive')
```
ChristophWurst, line 41: nit: make the number a variable so it can be changed without the need for a new translation string. it will also allow you to move the magic number into a constant

ChristophWurst (approved):

apps/settings/lib/SetupChecks/SchedulingTableSize.php
nickvergessen: Was thinking the same :P

apps/settings/lib/SetupChecks/SchedulingTableSize.php
```diff
@@ -36,9 +38,9 @@ public function run(): SetupResult {
 		$count = $query->fetchOne();
 		$query->closeCursor();
 
-		if ($count > 500000) {
+		if ($count > self::MAX_SCHEDULING_ENTRIES) {
 			return SetupResult::warning(
-				$this->l10n->t('You have more than 500 000 rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive')
+				$this->l10n->t('You have more than ' . self::MAX_SCHEDULING_ENTRIES . ' rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive')
```
ChristophWurst, line 43: you have to use a placeholder.
concatenated strings can not be translated

ChristophWurst (changes requested):

apps/settings/lib/SetupChecks/SchedulingTableSize.php
```diff
@@ -36,9 +38,11 @@ public function run(): SetupResult {
 		$count = $query->fetchOne();
 		$query->closeCursor();
 
-		if ($count > 500000) {
+		if ($count > self::MAX_SCHEDULING_ENTRIES) {
 			return SetupResult::warning(
-				$this->l10n->t('You have more than 500 000 rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive')
+				$this->l10n->t('You have more than %s rows in the scheduling objects table. Please run the expensive repair jobs via occ maintenance:repair --include-expensive.', [
```
nickvergessen, line 43: Is it worth a plural now? :speak_no_evil:

miaulalala: wdym? `row(s)`?

nickvergessen: https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#php
```
    public function getAuthors($count, $city) {
        // Translation with plural
        return $this->l->n(
            '%n author is currently in the city %1$s',  // singular string
            '%n authors are currently in the city %1$s',  // plural string
            $count, // number to decide which plural to use
            [$city] // further parameters are possible
        );
    }
```

So maybe hardcoding is better :-X

ChristophWurst: We'll always use this in the thousands so singular/plural distinction will never be used. Let's translate a single string with a placeholder

nickvergessen: So, refresh your review and we are good to go :P

ChristophWurst (approved):

PR #45977, opened 2024-06-19
Update favorites navigation list on folder renames

artonge:
- Waiting for https://github.com/nextcloud/richdocuments/pull/3797 to be resolved


apps/files/src/views/favorites.ts
```diff
@@ -106,6 +106,21 @@ export default () => {
...
 
+	/**
+	 * Update favourites navigation when a folder is renamed
+	 */
+	subscribe('files:node:renamed', (node: Node) => {
+		if (node.type !== FileType.Folder) {
+			return
+		}
+
+		if (node.attributes.favorite !== 1) {
```
susnux, line 117: Maybe to make sure if we ever fix the type to be a real boolean

```suggestion
		if (!!node.attributes.favorite) {
```

susnux (approved):

apps/files/src/views/favorites.ts
artonge: I prefer explicit comparisons, I have seen too many issues without, especially with lightly typed languages like typescript. :)

nfebe (approved):

skjnldsv:
Nice! Pretty cool

PR #45979, opened 2024-06-19
Signed requests

ArtificialOwl:
### Signed Request

Signing request allows to confirm the identity of the sender.
Signing request does not encrypt nor affect its payload.
Signing request only adds metadata to the headers of the request.

### Signature

The concept is to add unique metadata and sign them using a private/public key pair.
The location of the public key used to verify the signature will confirm the origin of the request.

Signature does not affect the data of the request, it only adds headers to it:
```
 {
     "(request-target)": "post /path",
     "content-length": 380,
     "date": "Mon, 08 Jul 2024 14:16:20 GMT",
     "digest": "SHA-256=U7gNVUQiixe5BRbp4Tg0xCZMTcSWXXUZI2\\/xtHM40S0=",
     "host": "hostname.of.the.recipient",
     "Signature": "keyId=\"https://author.hostname/ocm#signature\",algorithm=\"sha256\",headers=\"request-target) content-length date digest host\",signature=\"DzN12OCS1rsA[...]o0VmxjQooRo6HHabg==\""
 }
```

- 'content-length' is the total length of the data/content
- 'date' is the datetime the request have been initiated
- 'digest' is a checksum of the data/content
- 'host' is the hostname of the recipient of the request (remote when signing outgoing request, local on incoming request)
- 'Signature' contains the signature generated using the private key, and metadata:
     * 'keyId' is a unique id, formatted as an url. hostname is used to retrieve the public key via custom discovery
     * 'algorithm' define the algorithm used to generate signature
     * 'headers' contains a list of element used during the generation of the signature
     * 'signature' is the encrypted string, using local private key, of an array containing elements
       listed in 'headers' and their value. Some elements (content-length date digest host) are mandatory
       to ensure authenticity override protection.

_(Those are the minimum required headers, some can be added via options during the process)_

### ISignatoryManager

Because each protocol have different ways to obtain the public key of a remote instance or entity, some part of the signing/verifying process is managed by a custom provider, one for each protocol.

- ``getProviderId`` should returns a unique string
- ``getOptions`` can returns an array that can contains those entries:
	 *  'ttl' (300) is the lifetime (in secondes) of the signature
	 *  'ttlSignatory' (86400*3) the cache lifetime on a remote signatory
	 *  'extraSignatureHeaders' ([]) (list of extra headers to include to the signature)
	 *  'algorithm' ('sha256') is the algorithm used to generate the signature
	 *  'dateHeader' ("D, d M Y H:i:s T") the format of the 'date' header
	 
- ``getLocalSignatory`` should return the local signatory, including the full (public+private) key pair.
- ``getRemoteSignatory`` should returns a remote signatory based on the requested data, must at least contains key id and public key

### ISignatureManager

`ISignatureManager` is a service integrated to core that provide tools to set/get authenticity of/from outgoing/incoming requests.

- ``getIncomingSignedRequest`` extract data from the incoming request and compare headers to confirm authenticity of remote instance
- ``getOutgoingSignedRequest`` prep signature to sign an outgoing request.
- ``signOutgoingRequestIClientPayload`` is the one method to call to fully process of signing and fulfilling the payload for an outgoing request using IClient
- ``searchSignatory`` get a remote signatory from the database


lib/public/Migration/Attributes/TableCreation.php
nickvergessen, line 1: Unrelated

lib/private/Security/Signature/Model/SignedRequest.php
```diff
@@ -0,0 +1,149 @@
...
+	private string $digest;
+	private string $signedSignature = '';
+	private array $signatureHeader = [];
+	private ?ISignatory $signatory = null;
+
+	public function __construct(
+		private readonly string $body
+	) {
+		// digest is created on the fly using $body
+		$this->digest = 'SHA-256=' . base64_encode(hash("sha256", utf8_encode($body), true));
```
nickvergessen, line 30:
```suggestion
		$this->digest = 'SHA-256=' . base64_encode(hash('sha256', utf8_encode($body), true));
```

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,721 @@
...
+	 * @throws SignatureNotFoundException if incoming request is not signed
+	 * @throws SignatureException if signature could not be confirmed
+	 * @since 30.0.0
+	 */
+	public function getIncomingSignedRequest(
+		ISignatoryManager $signatoryManager,
+		?string $body = null
+	): IIncomingSignedRequest {
+		$body = $body ?? file_get_contents('php://input');
+		$this->logger->debug('[<<] incoming signed request', ['body length' => strlen($body)]);
```
nickvergessen, line 68: :broom:

core/Migrations/Version30000Date20240101084401.php
```diff
@@ -0,0 +1,114 @@
...
+			$table->addColumn('key_id', Types::STRING, [
+				'notnull' => true,
+				'length' => 512
+			]);
+			// host/provider_id/account will help generate a unique entry, not based on key_id
+			// this way, a spoofed instance cannot publish a new key_id for same host+provider_id
+			// account will be used only to stored multiple keys for the same provider_id/host
+			$table->addColumn('host', Types::STRING, [
+				'notnull' => true,
+				'length' => 127
```
nickvergessen, line 62: Please use 512 like we do for "remote" in external shares:
```suggestion
				'length' => 512,
```

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,721 @@
...
+
+			try {
+				$this->updateKnownSignatory($signatory);
+			} catch (SignatoryNotFoundException $e) {
+				$this->logger->warning('strange behavior, signatory not found ?', ['exception' => $e]);
+			}
+		}
+	}
+
+	private function insertSignatory(ISignatory $signatory): void {
```
nickvergessen, line 602: By now our Entity+QBMapper pattern is widely adapted. Any reason why you didn't go for it and instead went back to doing all this manually again?

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,721 @@
...
+	}
+
+
+	private function updateSignatoryPublicKey(ISignatory $signatory): void {
+		$qb = $this->connection->getQueryBuilder();
+		$qb->update(self::TABLE_SIGNATORIES)
+		   ->set('signatory', $qb->createNamedParameter($signatory->getPublicKey()))
+		   ->set('last_updated', $qb->createNamedParameter(time()));
+
+		$qb->where($qb->expr()->eq('`key_id_sum`', $qb->createNamedParameter($this->hashKeyId($signatory->getKeyId()))));
```
nickvergessen, line 681:
```suggestion
		$qb->where($qb->expr()->eq('key_id_sum', $qb->createNamedParameter($this->hashKeyId($signatory->getKeyId()))));
```

nickvergessen (dismissed):

apps/cloud_federation_api/lib/Controller/RequestHandlerController.php
```diff
@@ -201,6 +226,17 @@ public function addShare($shareWith, $name, $description, $providerId, $owner, $
 	 * 501: The resource type is not supported
 	 */
 	public function receiveNotification($notificationType, $resourceType, $providerId, ?array $notification) {
+		try {
+			// if request is signed and well signed, no exception are thrown
+			// if request is not signed and host is known for not supporting signed request, no exception are thrown
+			$signedRequest = $this->getSignedRequest();
+			$this->confirmShareOrigin($signedRequest, $notification['sharedSecret'] ?? '');
```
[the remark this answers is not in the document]
ArtificialOwl: this would not break it as I send the exception for the exact same reason of a missing 'sharedSecret' entry in the notifications request. The only thing is that I do this check earlier than others but I can add a prefix if you want (while I dont feel like necessary myself)

lib/private/Security/Signature/SignatureManager.php
ArtificialOwl: no real reason other that personal preference

apps/cloud_federation_api/lib/Controller/RequestHandlerController.php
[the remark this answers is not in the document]
nickvergessen: > this would not break it as I send the exception for the exact same reason of a missing 'sharedSecret' entry in the notifications request.

Nextcloud Talk is sending a data field `'sharedSecret'` and you either overwrite that and it breaks Talk Federation with older servers or you need to use a different key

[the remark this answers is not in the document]
ArtificialOwl: As said before, I only use this entry because it exists in the OCM protocol and doing so to compare that the origin of the reshare request, based on the token (and the linked recipient stored in the database), confirm the identity used to sign the request.

If Talk is using this endpoint to initiate anything, signature are to be required

apps/cloud_federation_api/lib/Capabilities.php
```diff
@@ -6,20 +6,27 @@
  * SPDX-FileCopyrightText: 2017 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
-
 namespace OCA\CloudFederationAPI;
 
+use NCU\Security\PublicPrivateKeyPairs\Exceptions\KeyPairException;
```
come-nc, line 11:
```suggestion
use NCU\Security\KeyPairs\Exceptions\KeyPairException;
```
Might be bikeshedding, but PublicPrivateKeyPair sounds redundant?

apps/cloud_federation_api/lib/Capabilities.php
```diff
@@ -60,6 +72,17 @@ public function getCapabilities() {
...
+			if (!$this->appConfig->getValueBool('core', OCMSignatoryManager::APPCONFIG_SIGN_DISABLED, lazy: true)) {
+				$this->provider->setSignatory($this->ocmSignatoryManager->getLocalSignatory());
+			} else {
+				$this->logger->debug('ocm public key feature disabled');
+			}
+		} catch (SignatoryException|KeyPairException $e) {
+			$this->logger->warning('cannot generate local signatory', ['exception' => $e]);
+		}
+
+		return ['ocm' => json_decode(json_encode($this->provider->jsonSerialize()), true)];
```
come-nc, line 86: Is the `json_decode(json_encode(` really needed?

apps/cloud_federation_api/lib/Controller/RequestHandlerController.php
```diff
@@ -286,4 +319,124 @@ private function mapUid($uid) {
...
+		$port = parse_url($rightPart, PHP_URL_PORT);
+		if ($port !== null && $port !== false) {
+			$host .= ':' . $port;
+		}
+
+		if (is_string($host) && $host !== '') {
+			return $host;
+		}
+
+		throw new IncomingRequestException('host is empty');
```
come-nc, line 440: This should be checked before the port to avoid returning ':239'

build/integration/federation_features/cleanup-remote-storage.feature
```diff
@@ -35,6 +35,9 @@ Feature: cleanup-remote-storage
     # server may have its own /textfile0.txt" file)
     And User "user1" copies file "/textfile0.txt" to "/remote-share.txt"
     And User "user1" from server "REMOTE" shares "/remote-share.txt" with user "user0" from server "LOCAL"
+    And As an "user1"
+    And sending "GET" to "/apps/files_sharing/api/v1/shares"
+    And the list of returned shares has 1 shares
```
come-nc, line 40: What change results in this being needed?

lib/private/Federation/CloudFederationProviderManager.php
```diff
@@ -193,9 +226,17 @@ public function sendCloudNotification(string $url, ICloudFederationNotification
...
+			$uri = $ocmProvider->getEndPoint() . '/notifications';
+			$payload = array_merge($this->getDefaultRequestOptions(), ['body' => json_encode($notification->getMessage())]);
+			if (!$this->appConfig->getValueBool('core', OCMSignatoryManager::APPCONFIG_SIGN_DISABLED, lazy: true)) {
+				$signedPayload = $this->signatureManager->signOutgoingRequestIClientPayload(
+					$this->signatoryManager,
+					$payload,
+					'post', $uri
+				);
+			}
+			return $client->post($uri, $signedPayload ?? $payload);
```
come-nc, line 239: code is duplicated 4 times, please wrap it in a method.

artonge (approved):

lib/private/OCM/OCMSignatoryManager.php
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * @inheritDoc
+ *
+ * returns local signatory using IKeyPairManager
+ * extract optional signatory (keyId+public key) from ocm discovery service on remote instance
+ *
+ * @since 31.0.0
+ */
+class OCMSignatoryManager implements ISignatoryManager {
```
come-nc, line 33: Why is it injected using the class name in applications if it has an interface?

lib/private/Security/PublicPrivateKeyPairs/KeyPairManager.php
```diff
@@ -0,0 +1,182 @@
...
+use NCU\Security\PublicPrivateKeyPairs\IKeyPairManager;
+use NCU\Security\PublicPrivateKeyPairs\Model\IKeyPair;
+use OC\Security\PublicPrivateKeyPairs\Model\KeyPair;
+use OCP\IAppConfig;
+
+/**
+ * @inheritDoc
+ *
+ * KeyPairManager store internal public/private key pair using AppConfig, taking advantage of the encryption
+ * and lazy loading.
```
come-nc, line 22: It should use its own table, it’s a bad habit to put everything into appconfig.

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,828 @@
...
+	 * @param string|null $body if NULL, body will be extracted from php://input
+	 *
+	 * @return IIncomingSignedRequest
+	 * @throws IncomingRequestException if anything looks wrong with the incoming request
+	 * @throws SignatureNotFoundException if incoming request is not signed
+	 * @throws SignatureException if signature could not be confirmed
+	 * @since 31.0.0
+	 */
+	public function getIncomingSignedRequest(
+		ISignatoryManager $signatoryManager,
```
come-nc, line 97: Why is that not injected? It creates a dependency loop?

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,828 @@
...
+	 * @param ISignatoryManager $signatoryManager
+	 * @param array $payload original payload, will be used to sign and completed with new headers with
+	 *                       signature elements
+	 * @param string $method needed in the signature
+	 * @param string $uri needed in the signature
+	 *
+	 * @return array new payload to be sent, including original payload and signature elements in headers
+	 * @since 31.0.0
+	 */
+	public function signOutgoingRequestIClientPayload(
```
come-nc, line 180:
```suggestion
	public function signOutgoingRequestPayload(
```

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,828 @@
...
+		$result->closeCursor();
+
+		if (!$row) {
+			throw new SignatoryNotFoundException('no signatory found');
+		}
+
+		$signature = new Signatory($row['key_id'], $row['public_key']);
+
+		return $signature->importFromDatabase($row);
+	}
```
come-nc, line 224: Is there no mapper and entities for signatories?

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,828 @@
...
+		if ($identity === null || $identity === false) {
+			throw new IdentityNotFoundException('cannot extract identity from ' . $uri);
+		}
+
+		if ($port !== null && $port !== false) {
+			$identity .= ':' . $port;
+		}
+
+		return $identity;
+	}
```
come-nc, line 269: I have Déjà-vu about this code

apps/cloud_federation_api/lib/Capabilities.php
ArtificialOwl: it is only requested for psalm as jsonSerialize returns serializable objects

apps/cloud_federation_api/lib/Capabilities.php
ArtificialOwl: done

lib/private/OCM/OCMSignatoryManager.php
ArtificialOwl: `ISignatoryManager` is not limited to OCM, it is an independant API that could be used to sign request for any other protocol. the idea is to generate its own way to obtain (ie. discovery) the needed information for `SignatureManager` to sign/verify payload

lib/private/Security/PublicPrivateKeyPairs/KeyPairManager.php
ArtificialOwl: I think it is the main idea of appconfig to avoid creating a new table for only few entries, and the lazy loading helps keeping it light

lib/private/Security/Signature/SignatureManager.php
ArtificialOwl: yes, fixed

ArtificialOwl:
Since last review, new commit:

- switching to `IdentityProof` 
- removing features not fully related to this PR
- fixing some edge use case on federationId,
- fixing an issue to extract the origin of a notification on a reshare,
- fixing test by adding threads to httpd,
- fixing all comments/requests from previous reviews
- upgrading version to trigger database migration

apps/cloud_federation_api/lib/Capabilities.php
come-nc: @provokateurin Any idea how to avoid it?

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,823 @@
...
+
+	private function setOutgoingSignedSignature(IOutgoingSignedRequest $signedRequest): void {
+		$clear = $signedRequest->getClearSignature();
+		$signed = $this->signString(
+			$clear, $signedRequest->getSignatory()->getPrivateKey(), $signedRequest->getAlgorithm()
+		);
+		$signedRequest->setSignedSignature($signed);
+	}
+
+	private function signingOutgoingRequest(IOutgoingSignedRequest $signedRequest): void {
```
come-nc, line 541:
```suggestion
	private function signOutgoingRequest(IOutgoingSignedRequest $signedRequest): void {
```

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,823 @@
...
+		$signatureHeader = $signedRequest->getSignatureHeader();
+		$headers = array_diff(array_keys($signatureHeader), ['(request-target)']);
+		$signatory = $signedRequest->getSignatory();
+		$signatureElements = [
+			'keyId="' . $signatory->getKeyId() . '"',
+			'algorithm="' . $this->getChosenEncryption($signedRequest->getAlgorithm()) . '"',
```
come-nc, line 547:
```suggestion
			'algorithm="' . $signedRequest->getAlgorithm() . '"',
```
Why not this directly?

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,823 @@
...
+			default => 'ras-sha256',
+		};
+	}
+
+	public function getOpenSSLAlgo(string $algorithm): int {
+		return match ($algorithm) {
+			'sha512' => OPENSSL_ALGO_SHA512,
+			default => OPENSSL_ALGO_SHA256,
+		};
+	}
```
come-nc, line 604: If you use an enum backed by a string you should not then write your own methods to translate it to and from string, use the ones from PHP.

Also the enum can host the method to translate to opensslalgo int.

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,823 @@
...
+
+	/**
+	 * This is called when a remote signatory does not exist anymore
+	 *
+	 * @param ISignatory|null $knownSignatory NULL is not known
+	 *
+	 * @throws SignatoryConflictException
+	 * @throws SignatoryNotFoundException
+	 */
+	private function manageDeprecatedSignatory(?ISignatory $knownSignatory): void {
```
come-nc, line 750:
```suggestion
	private function checkDeprecatedSignatory(?ISignatory $knownSignatory): void {
```

apps/cloud_federation_api/lib/Capabilities.php
nickvergessen: https://github.com/nextcloud/server/pull/45979/files#diff-16c946a83c0045d121ec926b69acadd4ccad303eb46144b3dfd087664522b8b2R212
```
publicKey: ISignatory|null,
```

I think this is the problem, instead you need to jsonSerialize that already and make it an array

lib/private/Security/Signature/SignatureManager.php
ArtificialOwl: the purpose is not to check but to set an action in case the remote instance does not provide signatory/public key anymore. Base on the type of link between the 2 instances, some action is performed

come-nc (changes requested):
There should be a mapper for signatory
There should not be a digest sent to openssl_* which is creating a digest by itself.
clearSignature should be renamed to something more appropriate.

lib/private/Security/Signature/Model/OutgoingSignedRequest.php
```diff
@@ -0,0 +1,173 @@
...
+			'(request-target)' => strtolower($method) . ' ' . $path,
+			'content-length' => strlen($this->getBody()),
+			'date' => gmdate($options['dateHeader'] ?? SignatureManager::DATE_HEADER),
+			'digest' => $this->getDigest(),
+			'host' => $this->getHost()
+		], $options['extraSignatureHeaders'] ?? []);
+
+		$signing = $headerList = [];
+		foreach ($headers as $element => $value) {
+			$value = $headers[$element];
```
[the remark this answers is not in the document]
ArtificialOwl: wtf.

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,637 @@
...
+				break;
+
+			case SignatoryType::TRUSTED:
+				// TODO: send notice to admin
+				throw new SignatoryConflictException();
+
+			case SignatoryType::STATIC:
+				// TODO: send warning to admin
+				throw new SignatoryConflictException();
+		}
```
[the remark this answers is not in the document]
ArtificialOwl: it is set by the 3rd party `ISignatoryManager` that will decide how must be handle a change of the remote public key

lib/private/Security/Signature/Model/IncomingSignedRequest.php
```diff
@@ -0,0 +1,209 @@
...
+	 */
+	private function extractSignatureHeaderFromRequest(): void {
+		$sign = [];
+		foreach (explode(',', $this->getRequest()->getHeader('Signature')) as $entry) {
+			if ($entry === '' || !strpos($entry, '=')) {
+				continue;
+			}
+
+			[$k, $v] = explode('=', $entry, 2);
+			preg_match('/"([^"]+)"/', $v, $var);
```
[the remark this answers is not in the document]
ArtificialOwl: I know mine is working, what should I expected different from yours ?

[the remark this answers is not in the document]
come-nc: https://regex101.com/r/BPNH0j/1
Yours will match quotes even if there is unexpected stuff around.
Mine makes sure that quotes are first and last characters.

sorbaugh:
Hello, due to the importance of having this issue merged, please let's postpone all approach-related optimization to later. If this PR checks this boxes, please let's approve and merge:

- [ ] PR fixes the issue / adds the feature as intended
- [ ] PR is backport compatible (needs a config switch that's OFF by default, needs to work with the OCM ecosystem)

After that we should update the documentation about the config switch.

Thanks! 🙏

lib/private/Security/Signature/Model/SignedRequest.php
```diff
@@ -0,0 +1,187 @@
...
+	private array $signatureElements = [];
+	private string $clearSignature = '';
+	private string $signedSignature = '';
+	private ?ISignatory $signatory = null;
+
+	public function __construct(
+		private readonly string $body,
+	) {
+		// digest is created on the fly using $body
+		$this->digest = 'SHA-256=' . base64_encode(hash('sha256', utf8_encode($body), true));
```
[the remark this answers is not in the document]
ArtificialOwl: I am not a fan of signing a document including part of content that could be generated from the outside. Using a checksum of the content (with a base64 encoding on top) makes everything more random. It also limit the size of the data sent to be signed which (I think) should be better (theoretically) to ensure the private key private

lib/private/Security/Signature/Model/OutgoingSignedRequest.php
```diff
@@ -0,0 +1,173 @@
...
+
+		$this->setHeaderList($headerList)
+			->setClearSignature(implode("\n", $signing));
+	}
+
+	/**
+	 * @inheritDoc
+	 *
+	 * @param string $host
+	 * @return IOutgoingSignedRequest
```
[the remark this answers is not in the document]
ArtificialOwl: i'll have a look

lib/private/Security/Signature/SignatureManager.php
```diff
@@ -0,0 +1,637 @@
...
+			}
+			$value = (strtolower($key) === 'host') ? $request->getServerHost() : $request->getHeader($key);
+			if ($value === '') {
+				throw new SignatureException('missing header ' . $key . ' in request');
+			}
+
+			$estimated[] = $key . ': ' . $value;
+		}
+
+		$signedRequest->setClearSignature(implode("\n", $estimated));
```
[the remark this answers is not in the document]
ArtificialOwl: i'll get a new term

lib/private/Security/Signature/Model/SignedRequest.php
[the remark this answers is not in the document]
come-nc: That is already what openssl_sign is doing. You do not need to prepare a hash before hand.
openssl_sign takes a hash method as a parameter, which it applies to the string before encryption to create the signature.
Why hash twice? It’s both a performance loss and a security issue because you rely on 2 hash methods and one of them being broken is enough.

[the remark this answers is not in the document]
ArtificialOwl: This is the process described in [RFC 9421: HTTP Message Signatures](https://datatracker.ietf.org/doc/rfc9421/)

Our implementation does not use the same syntax that the one defined in that RFC but the process stay the same.
The syntax used here is closer to ActivityPub/Mastodon.

There is this article describing the verification of Mastodon's HTTP message signatures in PHP: https://shkspr.mobi/blog/2024/02/a-simpleish-guide-to-verifying-http-message-signatures-in-php/ and the code from this PR is based on my previous work on `social`

lib/public/OCM/IOCMProvider.php
```diff
@@ -120,6 +120,22 @@ public function getResourceTypes(): array;
...
+	 */
+	public function setSignatory(Signatory $signatory): void;
+
+	/**
+	 * signatory (public/private key pair) used to sign outgoing/incoming request
+	 *
+	 * @return Signatory|null returns null if no Signatory available
+	 * @since 31.0.0
+	 */
+	public function getSignatory(): ?Signatory;
```
come-nc, line 137: Should remove those from OCP, at least for now.

lib/unstable/Security/Signature/IIncomingSignedRequest.php
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * set the hostname at the source of the request,
+	 * based on the keyId defined in the signature header.
+	 *
+	 * @param string $origin
+	 * @return IIncomingSignedRequest
+	 * @experimental 31.0.0
+	 */
+	public function setOrigin(string $origin): IIncomingSignedRequest;
```
come-nc, line 40:
```suggestion
```

lib/unstable/Security/Signature/ISignedRequest.php
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * set algorithm used to generate digest
+	 *
+	 * @param DigestAlgorithm $algorithm
+	 *
+	 * @return self
+	 * @experimental 31.0.0
+	 */
+	public function setDigestAlgorithm(DigestAlgorithm $algorithm): self;
```
come-nc, line 41:
```suggestion
```

Should only be on outgoing requests

lib/unstable/Security/Signature/ISignedRequest.php
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * set the list of headers related to the signature of the request
+	 *
+	 * @param array $elements
+	 *
+	 * @return self
+	 * @experimental 31.0.0
+	 */
+	public function setSigningElements(array $elements): self;
```
come-nc, line 67: Should be on outgoing requests

lib/unstable/Security/Signature/ISignedRequest.php
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * store data used to generate signature
+	 *
+	 * @param array $data
+	 *
+	 * @return self
+	 * @experimental 31.0.0
+	 */
+	public function setSignatureData(array $data): self;
```
come-nc, line 94: Should not be set from outside?

lib/unstable/Security/Signature/ISignedRequest.php
```diff
@@ -0,0 +1,147 @@
...
+
+	/**
+	 * set the signed version of the signature
+	 *
+	 * @param string $signature
+	 *
+	 * @return self
+	 * @experimental 31.0.0
+	 */
+	public function setSignature(string $signature): self;
```
come-nc, line 112: Same thing.

lib/unstable/Security/Signature/Model/Signatory.php
```diff
@@ -0,0 +1,193 @@
...
+	/**
+	 * update an entry in metadata
+	 *
+	 * @param string $key
+	 * @param string|int|float|bool|array $value
+	 * @experimental 31.0.0
+	 */
+	public function setMetaValue(string $key, string|int|float|bool|array $value): void {
+		$this->metadata[$key] = $value;
+	}
```
come-nc, line 157: Needs to add metadata to updated fields

lib/unstable/Security/Signature/ISignedRequest.php
ArtificialOwl: this is also used on incoming request to reconstruct the signature

lib/unstable/Security/Signature/ISignedRequest.php
ArtificialOwl: used in both incoming/outgoing

lib/unstable/Security/Signature/ISignedRequest.php
ArtificialOwl: set from outgoing/incoming

lib/unstable/Security/Signature/ISignedRequest.php
come-nc: Where?
```
> grep -R setSigningE lib apps/**.php core
lib/unstable/Security/Signature/ISignedRequest.php:     public function setSigningElements(array $elements): self;
lib/private/Security/Signature/Model/OutgoingSignedRequest.php:         $this->setSigningElements(
lib/private/Security/Signature/Model/SignedRequest.php: public function setSigningElements(array $elements): self {
lib/private/Security/Signature/Model/IncomingSignedRequest.php:         $this->setSigningElements($details);
```

I only see it being called internally, should not be part of the interface for incoming request, we do not get to chose the signing elements for an incoming request, it’s written in the headers.

lib/unstable/Security/Signature/IIncomingSignedRequest.php
ArtificialOwl: done

lib/unstable/Security/Signature/ISignedRequest.php
ArtificialOwl: done

lib/private/Security/Signature/SignatureManager.php
ArtificialOwl: switched to QBMapper

lib/unstable/Security/Signature/ISignedRequest.php
ArtificialOwl: misinterpretation of your review, this is now as expected

come-nc (approved):

core/Migrations/Version31000Date20240101084401.php
```diff
@@ -0,0 +1,135 @@
...
+				'length' => 127,
+				'default' => ''
+			]);
+			$table->addColumn('public_key', Types::TEXT, [
+				'notnull' => true,
+				'default' => ''
+			]);
+			$table->addColumn('metadata', Types::TEXT, [
+				'notnull' => true,
+				'default' => '[]'
```
nickvergessen, line 100:
```sql
mysql> DESCRIBE oc_sec_signatory;
+--------------+-----------------+------+-----+---------+----------------+
| Field        | Type            | Null | Key | Default | Extra          |
+--------------+-----------------+------+-----+---------+----------------+
| id           | bigint unsigned | NO   | PRI | NULL    | auto_increment |
| key_id_sum   | varchar(127)    | NO   | MUL | NULL    |                |
| key_id       | varchar(512)    | NO   |     | NULL    |                |
| host         | varchar(512)    | NO   |     | NULL    |                |
| provider_id  | varchar(31)     | NO   | MUL | NULL    |                |
| account      | varchar(127)    | YES  |     |         |                |
| public_key   | longtext        | NO   |     | NULL    |                |
| metadata     | longtext        | NO   |     | NULL    |                |
| type         | smallint        | NO   |     | 9       |                |
| status       | smallint        | NO   |     | 0       |                |
| creation     | int unsigned    | YES  |     | 0       |                |
| last_updated | int unsigned    | YES  |     | 0       |                |
+--------------+-----------------+------+-----+---------+----------------+
12 rows in set (0,00 sec)
```

Seems this does not work. TEXT fields can not have a default.

nickvergessen: This breaks Talk integration tests on:
MySQL
```
{"Exception":"OC\\DB\\Exceptions\\DbalException","Message":"An exception occurred while executing a query: SQLSTATE[HY000]: General error: 1364 Field 'metadata' doesn't have a default value"
```

juliusknorr: I think there is another failure with oracle catched by richdocuments integration tests:

```
{"Exception":"Exception","Message":"Cannot assign null to property NCU\\Security\\Signature\\Model\\Signatory::$account of type string in file '\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/public\/AppFramework\/Db\/Entity.php' line 149","Code":0,"Trace":[{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/AppFramework\/App.php","line":161,"function":"dispatch","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/Route\/Router.php","line":306,"function":"main","class":"OC\\AppFramework\\App","type":"::"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/base.php","line":1019,"function":"match","class":"OC\\Route\\Router","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/index.php","line":24,"function":"handleRequest","class":"OC","type":"::"}],"File":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/AppFramework\/Http\/Dispatcher.php","Line":146,"Previous":{"Exception":"TypeError","Message":"Cannot assign null to property NCU\\Security\\Signature\\Model\\Signatory::$account of type string","Code":0,"Trace":[{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/public\/AppFramework\/Db\/Entity.php","line":59,"function":"setter","class":"OCP\\AppFramework\\Db\\Entity","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/public\/AppFramework\/Db\/QBMapper.php","line":315,"function":"fromRow","class":"OCP\\AppFramework\\Db\\Entity","type":"::"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/public\/AppFramework\/Db\/QBMapper.php","line":375,"function":"mapRowToEntity","class":"OCP\\AppFramework\\Db\\QBMapper","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/Security\/Signature\/Db\/SignatoryMapper.php","line":56,"function":"findEntity","class":"OCP\\AppFramework\\Db\\QBMapper","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/Security\/Signature\/SignatureManager.php","line":326,"function":"getByKeyId","class":"OC\\Security\\Signature\\Db\\SignatoryMapper","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/Security\/Signature\/SignatureManager.php","line":140,"function":"getStoredSignatory","class":"OC\\Security\\Signature\\SignatureManager","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/Security\/Signature\/SignatureManager.php","line":108,"function":"confirmIncomingRequestSignature","class":"OC\\Security\\Signature\\SignatureManager","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/apps\/cloud_federation_api\/lib\/Controller\/RequestHandlerController.php","line":339,"function":"getIncomingSignedRequest","class":"OC\\Security\\Signature\\SignatureManager","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/apps\/cloud_federation_api\/lib\/Controller\/RequestHandlerController.php","line":104,"function":"getSignedRequest","class":"OCA\\CloudFederationAPI\\Controller\\RequestHandlerController","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/AppFramework\/Http\/Dispatcher.php","line":200,"function":"addShare","class":"OCA\\CloudFederationAPI\\Controller\\RequestHandlerController","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/AppFramework\/Http\/Dispatcher.php","line":114,"function":"executeController","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/AppFramework\/App.php","line":161,"function":"dispatch","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/private\/Route\/Router.php","line":306,"function":"main","class":"OC\\AppFramework\\App","type":"::"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/base.php","line":1019,"function":"match","class":"OC\\Route\\Router","type":"->"},{"file":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/index.php","line":24,"function":"handleRequest","class":"OC","type":"::"}],"File":"\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/public\/AppFramework\/Db\/Entity.php","Line":149},"message":"Cannot assign null to property NCU\\Security\\Signature\\Model\\Signatory::$account of type string in file '\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/public\/AppFramework\/Db\/Entity.php' line 149","exception":{},"CustomMessage":"Cannot assign null to property NCU\\Security\\Signature\\Model\\Signatory::$account of type string in file '\/home\/runner\/actions-runner\/_work\/richdocuments\/richdocuments\/lib\/public\/AppFramework\/Db\/Entity.php' line 149"}
```

PR #46004, opened 2024-06-20
perf(capabilities): Log capabilities providers that are slow

nickvergessen:
## Summary

Added a `sleep(3)` to test the log:
```
Capabilities of OCA\Talk\Capabilities took 3 seconds to generate.
```

## Checklist

### DIFF
--- a/lib/private/CapabilitiesManager.php
+++ b/lib/private/CapabilitiesManager.php
@@ -15,6 +15,12 @@ use OCP\Capabilities\IPublicCapability;
 use Psr\Log\LoggerInterface;
 
 class CapabilitiesManager {
+	/**
+	 * Anything above 0.1s to load the capabilities of an app qualifies for bad code
+	 * and should be cached within the app.
+	 */
+	public const ACCEPTABLE_LOADING_TIME = 0.1;
+
 	/** @var \Closure[] */
 	private $capabilities = [];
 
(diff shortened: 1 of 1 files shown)


lib/private/CapabilitiesManager.php
```diff
@@ -51,7 +51,16 @@ public function getCapabilities(bool $public = false, bool $initialState = false
 						// that we would otherwise inject to every page load
 						continue;
 					}
+					$startTime = microtime(true);
 					$capabilities = array_replace_recursive($capabilities, $c->getCapabilities());
+					$endTime = microtime(true);
+					$duration = $endTime - $startTime;
+					if ($duration > 0.3) {
+						$this->logger->warning('Capabilities of {className} took {duration} seconds to generate.', [
```
szaimen, line 59:
```suggestion
						$this->logger->info('Capabilities of {className} took {duration} seconds to generate.', [
```

Altahrim (approved):

lib/private/CapabilitiesManager.php
nickvergessen: Then no one will see it from a real instance. But maybe we simply start with a warning at 1s or something and info below it

ChristophWurst: You could also make it adaptive like https://github.com/nextcloud/server/blob/08cf60923c1ed8b41f4f0b9e34bea9a0ad00fd31/cron.php#L171-L177

ChristophWurst (approved):

szaimen (changes requested):
See my comment

lib/private/CapabilitiesManager.php
```diff
@@ -51,7 +57,27 @@ public function getCapabilities(bool $public = false, bool $initialState = false
 						// that we would otherwise inject to every page load
 						continue;
 					}
+					$startTime = microtime(true);
 					$capabilities = array_replace_recursive($capabilities, $c->getCapabilities());
+					$endTime = microtime(true);
+					$timeSpent = $endTime - $startTime;
+					if ($timeSpent > self::ACCEPTABLE_LOADING_TIME) {
```
ChristophWurst, line 64: microtime gives you µs so we'll log anything that is slower than 100ns, right?

nickvergessen: > microtime gives you µs

its float with seconds? https://3v4l.org/YF2v8

ChristophWurst: Whooops. I read *returns the current Unix timestamp with microseconds* as *in microseconds*

ChristophWurst (approved):

szaimen (approved):
Fine by me but didnt test

PR #46007, opened 2024-06-20
feat(files_sharing): add `new file request` option

### LINKED ISSUE #45446: File request
This is a modal dialog that helps a user create a file drop link & send it to people. A specific dialog makes it easier to find and use, and more visible to users. But the end result is 'just' a file drop link.

Similar to https://help.dropbox.com/share/create-file-request
Sub task: Enter name when you upload some files, sort in folder

Use case:
* university teacher requests homework from the students. The student has to enter their name. Their uploads are stored in subfolders by their name. Expiration date = deadline for homework turn in time!

Flow:
* Somewhere (maybe in the + menu) choose 'request a file'
* Give a title (will be the name of the file-drop link)
* Description (optional) (put in note to recipient)
* choose a location (folder) the files should go into
* choose if the files should go into a subfolder based on name of uploader (optional)
* choose deadline (optional but might be set by default) (goes into expiration date) - should have a warning that once the expiration date is hit, the request disappears and a new one has to be created.
* choose password (optional but might be enforced) 
* choose recipients (optional) - should be possible to be users/groups/email addresses (from address book?)
* click 'create' & get a link.
* would be good to go to the folder and how the file drop in the side bar, so the user also 'learns' where the link now is, where they can modify it etc.
* we should add a 'file requests' to the share sidebar, showing all file drop links
* We also need to make the share note much more visible to the recipient - it is now quite hidden. It should ideally also allow a bit more 'richness' in its markup, but that's not so critical.

skjnldsv:
closes https://github.com/nextcloud/server/issues/45446


lib/private/DB/Connection.php
```diff
@@ -283,15 +283,15 @@ public function executeQuery(string $sql, array $params = [], $types = [], ?Quer
...
-			$this->logger->log(
-				(int) ($this->systemConfig->getValue('loglevel_dirty_database_queries', null) ?? 0),
-				'dirty table reads: ' . $sql,
-				[
-					'tables' => array_keys($this->tableDirtyWrites),
-					'reads' => $tables,
-					'exception' => new \Exception('dirty table reads: ' . $sql),
-				],
-			);
+			// $this->logger->log(
```
skjnldsv, line 286: REMOVE

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -12,14 +12,15 @@
...
 use OC\Files\FileInfo;
 use OC\Files\Storage\Wrapper\Wrapper;
-use OCA\Files\Helper;
 use OCA\Files_Sharing\Exceptions\SharingRightsException;
 use OCA\Files_Sharing\External\Storage;
-use OCA\Files_Sharing\ResponseDefinitions;
 use OCA\Files_Sharing\SharedStorage;
+use OCA\Files\Helper;
 use OCP\App\IAppManager;
-use OCP\AppFramework\Http;
```
susnux, line 21: This causes the psalm error below because now `Http::STATUS_OK` is unknown.

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -538,7 +544,8 @@ public function createShare(
 		?string $expireDate = null,
 		string $note = '',
 		string $label = '',
-		?string $attributes = null
+		?string $attributes = null,
+		?string $mailSend = null
```
susnux, line 549: why is this a string and not a boolean?

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -1134,7 +1153,8 @@ public function updateShare(
 		?string $note = null,
 		?string $label = null,
 		?string $hideDownload = null,
-		?string $attributes = null
+		?string $attributes = null,
+		?string $mailSend = null,
```
susnux, line 1158: same here

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -2025,6 +2058,64 @@ private function checkInheritedAttributes(IShare $share): void {
 				}
 			}
 		}
+	}
+
+	/**
+	 * @param string $id
+	 * @param string[] $emails a list of emails to send the notification to
```
susnux: Param does not exist (but `password`)

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestDatePassword.vue
```diff
@@ -0,0 +1,236 @@
+<!--
```
susnux, line 1: Vue naming schema would be to always keep the "parent" name.
`NewFileRequestDialogIntro`
`NewFileRequestDialogDatePasswort`
`NewFileRequestDialogFinish`

See https://vuejs.org/style-guide/rules-strongly-recommended.html#tightly-coupled-component-names

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestDatePassword.vue
```diff
@@ -0,0 +1,236 @@
...
+		NcPasswordField,
+	},
+
+	props: {
+		disabled: {
+			type: Boolean,
+			required: false,
+			default: false,
+		},
+		deadline: {
```
susnux, line 116: I think it would make thing clearer to call it `expirationDate`

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestDatePassword.vue
```diff
@@ -0,0 +1,236 @@
...
+			this.showPassword()
+		},
+
+		async generatePassword() {
+			await GeneratePassword().then(password => {
+				this.$emit('update:password', password)
+			})
+		},
+
+		showPassword() {
```
susnux, line 222: good point we should expose a function from the component so you can programmatically show the password without access to private internals :)

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
```diff
@@ -0,0 +1,219 @@
...
+			:value="shareLink"
+			:label="t('files_sharing', 'Share link')"
+			:readonly="true"
+			:show-trailing-button="true"
+			:trailing-button-label="t('files_sharing', 'Copy to clipboard')"
+			@click="copyShareLink"
+			@click-trailing-button="copyShareLink">
+			<template #trailing-button-icon>
+				<IconCheck v-if="isCopied" :size="20" @click="isCopied = false" />
+				<IconClipboard v-else :size="20" @click="copyShareLink" />
```
susnux, line 24: Having click listeners on non-interactive elements (icons here) is bad and will cause accessibility warning / maybe even errors.

Instead just use `click-trailing-button` which you already do for the clipboard one, so just adjust the function (see below).

```suggestion
				<IconCheck v-if="isCopied" :size="20" />
				<IconClipboard v-else :size="20" />
```

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
```diff
@@ -0,0 +1,219 @@
...
+		shareLink() {
+			return window.location.protocol + '//' + window.location.host + generateUrl('/s/') + this.share.token
+		},
+	},
+
+	methods: {
+		async copyShareLink(event: MouseEvent) {
+			if (!navigator.clipboard) {
+				// Clipboard API not available
+				showError(this.t('files_sharing', 'Clipboard is not available'))
```
susnux, line 129: We could add a fallback for local networks, had a lot of complains about that.
Like
```suggestion
				window.prompt(t('files_sharing', 'Automatically copying failed, please copy the sharelink manually:') this.shareLink)
```

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
```diff
@@ -0,0 +1,219 @@
...
+
+	computed: {
+		shareLink() {
+			return window.location.protocol + '//' + window.location.host + generateUrl('/s/') + this.share.token
+		},
+	},
+
+	methods: {
+		async copyShareLink(event: MouseEvent) {
+			if (!navigator.clipboard) {
```
susnux, line 127:
```suggestion
			if (this.isCopied) {
				this.isCopied = false
				return
			}
			if (!navigator.clipboard) {
```

apps/files_sharing/lib/Controller/ShareAPIController.php
susnux: Was discussed somehow this will cause an issue

skjnldsv: Because it doesn't work, see the other booleans types like `publicUpload` or `sendPasswordByTalk`

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestDatePassword.vue
skjnldsv: Good call :+1:

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,418 @@
...
+		<!-- Header -->
+		<NcNoteCard v-show="currentStep === STEP.FIRST" type="info" class="file-request-dialog__header">
+			<p id="file-request-dialog-description" class="file-request-dialog__description">
+				{{ t('files_sharing', 'Collect files from others even if they don\'t have an account.') }}
+				{{ t('files_sharing', 'To ensure you can receive files, verify you have enough storage available.') }}
+			</p>
+		</NcNoteCard>
+
+		<!-- Main form -->
+		<form ref="form"
```
susnux, line 23: I think because of changing content you need `aria-live="polite"`

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
skjnldsv: Yeah, I did this before and the click did not work either. That's why I tried to move it to the icon element. That is temporary :+1: 
I need to double check again

apps/files_sharing/src/components/NewFileRequestDialog.vue
skjnldsv: Thanks! I wanted to ask the accessibility team, I wasn't sure about how to handle that

apps/files_sharing/src/models/Share.js
```diff
@@ -532,25 +532,36 @@ export default class Share {
...
+	/**
+	 * Is this mail share a file request ?
+	 *
+	 * @return {boolean}
+	 * @readonly
+	 * @memberof Share
+	 */
+	get isFileRequest() {
+		const isFileRequest = (attribute) => {
+			return attribute.scope === 'fileRequest' && attribute.key === 'enabled' && attribute.value === true
```
artonge, line 550: Shouldn't it be based on the permissions and not an extra attribute?

apps/files_sharing/src/new/newFileRequest.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const NewFileRequestDialogVue = defineAsyncComponent(() => import('../components/NewFileRequestDialog.vue'))
+
+export const entry = {
+	id: 'file-request',
+	displayName: t('files', 'Create new file request'),
+	iconSvgInline: FileUploadSvg,
+	order: 30,
+	enabled(): boolean {
+		// determine requirements
```
artonge, line 19:
```suggestion
		// TODO: determine requirements
```

come-nc (approved):
Reviewed PHP only, all good except from psalm errors to fix. (and openapi)

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
```diff
@@ -0,0 +1,219 @@
...
+
+			if (validEmails.length > 0) {
+				showSuccess(this.n('files_sharing', '1 email address added', '{count} email addresses added', validEmails.length, { count: validEmails.length }))
+			}
+
+			this.email = ''
+		},
+
+		isValidEmail(email) {
+			const regExpEmail = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
susnux: Not sure how much sense such a regex makes, it often does not sattiesfies all valid cases, I would recommend to use the following which only has 2 limitiations on valid emails:
1. @ is only allowed once (in theory you can have it quoted in valid emails like `"my@home"@domain.com`
2. Require a TLD, in theory you can have just the hostname for local networks

`/^[^@]+@[^@]+\.[^.]{2,}$/`

And then the backend should verify it properly as we have the tool for it:
https://nextcloud-server.netlify.app/classes/ocp-mail-imailer#method_validateMailAddress

apps/sharebymail/lib/ShareByMailProvider.php
```diff
@@ -230,30 +225,51 @@ protected function createMailShare(IShare $share): int {
...
-		if (!$this->mailer->validateMailAddress($share->getSharedWith())) {
+		$emails = $this->getSharedWithEmails($share);
+		$validEmails = array_filter($emails, function (string $email) {
+			return $this->mailer->validateMailAddress($email);
+		});
+
+		if (count($validEmails) === 0) {
 			$this->removeShareFromTable($shareId);
-			$e = new HintException('Failed to send share by mail. Got an invalid email address: ' . $share->getSharedWith(),
+			$e = new HintException('Failed to send share by mail. Could not find a valid email address: ' . join(', ', $emails),
```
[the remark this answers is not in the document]
skjnldsv: It is logged as error right below :)

apps/sharebymail/lib/ShareByMailProvider.php
```diff
@@ -230,30 +225,51 @@ protected function createMailShare(IShare $share): int {
...
-			);
+			$this->sendEmail($share, $validEmails);
+
+			// If we have a password set, we send it to the recipient
+			if ($share->getPassword() !== null) {
+				// Sends share password to receiver when it's a permanent one (otherwise she will have to request it via the showShare UI)
+				// or to owner when the password shall be given during a Talk session
+				$passwordExpire = $this->config->getSystemValue('sharing.enable_mail_link_password_expiration', false);
+				$passwordEnforced = $this->shareManager->shareApiLinkEnforcePassword();
+				if ($passwordExpire === false || $share->getSendPasswordByTalk()) {
```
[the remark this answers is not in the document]
skjnldsv: Didn't touch that part, that's how it was before.
It's compatible with Mail shares in general, but we don't use it from the File Request feature :)

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestIntro.vue
```diff
@@ -0,0 +1,153 @@
...
+				<IconFolder :size="18" />
+			</NcTextField>
+		</fieldset>
+
+		<!-- Request note -->
+		<fieldset class="file-request-dialog__note" data-cy-file-request-dialog-fieldset="note">
+			<legend>
+				{{ t('files_sharing', 'Add a note') }}
+			</legend>
+			<NcTextArea :value="note"
```
susnux: There is no label outside, the legend labels only the fieldset but not the text area.
So the textarea is missing a label for accessibility.

I would remove the fieldset and instead make it a `div` and move the legend text to the built-in `label` prop of the textarea, as there is only one field inside the fieldset.

Otherwiese add an ID to the legend and `aria-labelledby` to the text area

apps/files_sharing/src/components/SharingEntryLink.vue
```diff
@@ -554,9 +564,13 @@ export default {
 		},
 
 		canChangeHideDownload() {
-			const hasDisabledDownload = (shareAttribute) => shareAttribute.key === 'download' && shareAttribute.scope === 'permissions' && shareAttribute.enabled === false
+			const hasDisabledDownload = (shareAttribute) => shareAttribute.scope === 'permissions' && shareAttribute.key === 'download'&& shareAttribute.value === false
```
susnux, line 567: Maybe keep it readable :sweat_smile:  (or at least add a space between `'download' and '&&') ;)

```suggestion
			const hasDisabledDownload = (shareAttribute) => 
				shareAttribute.scope === 'permissions'
				&& shareAttribute.key === 'download'
				&& shareAttribute.value === false
```

apps/files_sharing/src/new/newFileRequest.ts
```diff
@@ -0,0 +1,50 @@
...
+		return true
+	},
+	async handler(context: Folder, content: Node[]) {
+		// Create document root
+		const mountingPoint = document.createElement('div')
+		mountingPoint.id = 'file-request-dialog'
+		document.body.appendChild(mountingPoint)
+
+		// Init vue app
+		const NewFileRequestDialog = new Vue({
```
susnux, line 31: You can just use `spawnDialog` helper from `@nextcloud/dialogs` if you like :)

lib/public/Share/IAttributes.php
```diff
@@ -17,19 +17,19 @@ interface IAttributes {
 	 *
 	 * @param string $scope scope
 	 * @param string $key key
-	 * @param bool $enabled enabled
+	 * @param bool|string|array|null $value value
```
artonge, line 20: Isn't it a breaking change?

apps/sharebymail/lib/ShareByMailProvider.php
```diff
@@ -342,13 +368,26 @@ protected function sendMailNotification(
...
+	 *  2. the setting to send the password by mail is disabled
+	 *  3. the share is set to send the password by talk
+	 *
+	 * @param IShare $share
+	 * @param string $password
+	 * @param array $emails
+	 * @return bool
 	 */
-	protected function sendPassword(IShare $share, string $password): bool {
+	protected function sendPassword(IShare $share, string $password, array $emails) {
```
susnux: why removing the return type? I think we should in general do the opposite.
So probably should be `: void` and in line 357 you forgot to remove the `return false`

skjnldsv: Refactor issue :)
I had some pending changes to push which  included the boolean being back

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
skjnldsv: This is the regex mail uses

apps/files_sharing/src/models/Share.js
skjnldsv: is it really a permission though?

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,418 @@
...
+				:disabled="loading"
+				:title="t('files_sharing', 'Cancel the file request creation')"
+				data-cy-file-request-dialog-controls="cancel"
+				type="tertiary"
+				@click="onCancel">
+				{{ t('files_sharing', 'Cancel') }}
+			</NcButton>
+
+			<!-- Align right -->
+			<span class="dialog__actions-separator" />
```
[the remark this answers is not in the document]
skjnldsv: I'll let them speak for themselves when I ask for a design review

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
skjnldsv: https://github.com/nextcloud/mail/blob/1b2f3a012fb8be2ac2ec6dc5fb4294bff9a4e618/src/components/AccountForm.vue#L649

jancborchardt:
@skjnldsv could you add some screenshots? :) And cc @nimishavijay for the design

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -538,7 +546,8 @@ public function createShare(
 		?string $expireDate = null,
 		string $note = '',
 		string $label = '',
-		?string $attributes = null
+		?string $attributes = null,
+		?string $sendMail = null
```
provokateurin, line 550: This should just be a boolean

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -1134,7 +1155,8 @@ public function updateShare(
 		?string $note = null,
 		?string $label = null,
 		?string $hideDownload = null,
-		?string $attributes = null
+		?string $attributes = null,
+		?string $sendMail = null,
```
provokateurin, line 1159: Boolean as well

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -2025,6 +2060,71 @@ private function checkInheritedAttributes(IShare $share): void {
 				}
 			}
 		}
+	}
+
+	/**
+	 * Send a mail notification again for a share.
+	 * The mail_send option must be enabled for the given share.
+	 * @param string $id the share ID
+	 * @param string $password optional, the password to check against. Necessary for password protected shares.
```
provokateurin, line 2069:
```suggestion
	 * @param string $password the password to check against. Necessary for password protected shares.
```

No need to say it's optional, since it has a default value

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestDatePassword.vue
```diff
@@ -0,0 +1,227 @@
...
+			defaultExpireDate: window.OC.appConfig.core.defaultExpireDate as number,
+			// Default expiration date is enabled for public links (can be disabled)
+			defaultExpireDateEnabled: window.OC.appConfig.core.defaultExpireDateEnabled === true,
+			// Default expiration date is enforced for public links (can't be disabled)
+			defaultExpireDateEnforced: window.OC.appConfig.core.defaultExpireDateEnforced === true,
+
+			// Default password protection is enabled for public links (can be disabled)
+			enableLinkPasswordByDefault: window.OC.appConfig.core.enableLinkPasswordByDefault === true,
+			// Password protection is enforced for public links (can't be disabled)
+			enforcePasswordForPublicLink: window.OC.appConfig.core.enforcePasswordForPublicLink === true,
```
Fenn-CS: I don't think it's this is needed we have https://github.com/nextcloud/server/blob/master/apps/files_sharing/src/services/ConfigService.js

```js
import Config from '../services/ConfigService.js

export default {
	data() {
		return {
			config: new Config(),
			}
		}
}
```

All of those then become available in template via `this.config.*` using the `ShareMixin` would provide this out of the box (but could bloat this too)

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -2025,6 +2060,71 @@ private function checkInheritedAttributes(IShare $share): void {
...
+	public function sendShareEmail(string $id, $password = ''): DataResponse {
+		try {
+			$share = $this->getShareById($id);
 
+			if (!$this->canAccessShare($share, false)) {
+				throw new OCSNotFoundException($this->l->t('Wrong share ID, share does not exist'));
+			}
+	
+			if (!$this->canEditShare($share)) {
+				throw new OCSForbiddenException('You are not allowed to send mail notifications');
```
susnux, line 2088: Why is this not translated? (the one above is)

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -2025,6 +2060,71 @@ private function checkInheritedAttributes(IShare $share): void {
...
+			}
+
+			// For mail and link shares, the user must be
+			// the owner of the share, not only the file owner.
+			if ($share->getShareType() === IShare::TYPE_EMAIL
+				|| $share->getShareType() === IShare::TYPE_LINK) {
+				if ($share->getSharedBy() !== $this->currentUser) {
+					throw new OCSForbiddenException('You are not allowed to send mail notifications');
```
susnux, line 2096: Same here, why not translated but others are?

apps/files_sharing/src/components/NewFileRequestDialog/FileRequestFinish.vue
```diff
@@ -0,0 +1,217 @@
...
+		},
+	},
+
+	emits: ['add-email', 'remove-email'],
+
+	setup() {
+		return {
+			n: translatePlural,
+			t: translate,
+			isShareByMailEnabled: getCapabilities()?.files_sharing?.sharebymail?.enabled === true,
```
Fenn-CS: Present in `this.config.isMailShareAllowed`

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,421 @@
...
+			<p id="file-request-dialog-description" class="file-request-dialog__description">
+				{{ t('files_sharing', 'Collect files from others even if they don\'t have an account.') }}
+				{{ t('files_sharing', 'To ensure you can receive files, verify you have enough storage available.') }}
+			</p>
+		</NcNoteCard>
+
+		<!-- Main form -->
+		<form ref="form"
+			class="file-request-dialog__form"
+			aria-labelledby="file-request-dialog-description"
```
susnux, line 25: This is not a label but a description, you can use `aria-describedby` here but please use a short descriptive label (the aria label is the accessible name but you use two sentences here as a name).

nfebe (commented):
For commit => feat(files_sharing): add new file request option.

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,421 @@
...
+		class="file-request-dialog"
+		data-cy-file-request-dialog
+		:close-on-click-outside="false"
+		:name="currentStep !== STEP.LAST ? t('files_sharing', 'Create a file request') : t('files_sharing', 'File request created')"
+		size="normal"
+		@closing="onCancel">
+		<!-- Header -->
+		<NcNoteCard v-show="currentStep === STEP.FIRST" type="info" class="file-request-dialog__header">
+			<p id="file-request-dialog-description" class="file-request-dialog__description">
+				{{ t('files_sharing', 'Collect files from others even if they don\'t have an account.') }}
```
susnux, line 17: I am not 100% sure about our guidelines here, but I would think we do not use abbreviations

```suggestion
				{{ t('files_sharing', 'Collect files from others even if they do not have an account.') }}
```

apps/files_sharing/src/components/NewFileRequestDialog.vue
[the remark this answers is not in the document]
susnux: I think it would be good to have this clarified as IIRC the Jan wanted all buttons to be on the same side of dialogs separated by color (primary actions vs secondary action) (from when I created the filepicker).
Would be good to have things like this all consistent by design.

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,421 @@
...
+				</template>
+				{{ finishButtonLabel }}
+			</NcButton>
+		</template>
+	</NcDialog>
+</template>
+
+<script lang="ts">
+// eslint-disable-next-line n/no-extraneous-import
+import type { AxiosError } from 'axios'
```
susnux, line 107: You can import from `@nextcloud/axios`, no need to mess with eslint disable comments here

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,418 @@
...
+
+			share: null as Share | null,
+			emails: [] as string[],
+		}
+	},
+
+	computed: {
+		finishButtonLabel() {
+			if (this.emails.length === 0) {
+				return this.t('files_sharing', 'Close')
```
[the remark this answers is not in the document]
susnux: see not above

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
```diff
@@ -0,0 +1,153 @@
...
+				:placeholder="t('files_sharing', 'Birthday party photos, History assignment…')"
+				:required="false"
+				name="label"
+				@update:value="$emit('update:label', $event)" />
+		</fieldset>
+
+		<!-- Request destination -->
+		<fieldset class="file-request-dialog__destination" data-cy-file-request-dialog-fieldset="destination">
+			<legend>
+				{{ t('files_sharing', 'Where should these files go ?') }}
```
susnux, line 25:
```suggestion
				{{ t('files_sharing', 'Where should these files go?') }}
```

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,421 @@
...
+		onPageNext() {
+			const form = this.$refs.form as HTMLFormElement
+			if (!form.checkValidity()) {
+				form.reportValidity()
+			}
+
+			// custom destination validation
+			// cannot share root
+			if (this.destination === '/' || this.destination === '') {
+				const destinationInput = form.querySelector('input[name="destination"]') as HTMLInputElement
```
susnux, line 215: Always very confusing to mess with elements not part of the component, this makes it hard to find out where this is coming from and prone to break because now this component needs to know about other components templates.
Instead it is recommended to only access the component, e.g. you could expose a `validate` method on the component an call that in this step to let the component handle the validation.

This makes the relationships clear.

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,421 @@
...
+
+		async setShareEmails() {
+			this.loading = true
+
+			// This should never happen™
+			if (!this.share || !this.share?.id) {
+				throw new Error('Share ID is missing')
+			}
+
+			const shareUrl = generateOcsUrl('apps/files_sharing/api/v1/shares/' + this.share.id)
```
susnux, line 313:
```suggestion
			const shareUrl = generateOcsUrl('apps/files_sharing/api/v1/shares/{id}', { id: this.share.id })
```

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,421 @@
...
+
+		async sendEmails() {
+			this.loading = true
+
+			// This should never happen™
+			if (!this.share || !this.share?.id) {
+				throw new Error('Share ID is missing')
+			}
+
+			const shareUrl = generateOcsUrl('apps/files_sharing/api/v1/shares/' + this.share.id + '/send-email')
```
susnux, line 344:
```suggestion
			const shareUrl = generateOcsUrl('apps/files_sharing/api/v1/shares/{id}/send-email', { id: this.share.id })
```

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,421 @@
...
+
+	&__form {
+		position: relative;
+		overflow: auto;
+		padding: var(--secondary-margin) var(--margin);
+		// overlap header bottom padding
+		margin-top: calc(-1 * var(--secondary-margin));
+	}
+
+	:deep(fieldset) {
```
susnux, line 393: Is this really needed? You have full control over the components so a `deep` should not be needed here, instead you could just have either shared styles or make it a component for consistent layout (also simplifies by dividing into smaller).

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,418 @@
...
+		margin-top: var(--secondary-margin);
+
+		:deep(legend) {
+			display: flex;
+			align-items: center;
+			width: 100%;
+		}
+	}
+
+	:deep(.dialog__actions) {
```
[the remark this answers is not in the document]
susnux: see above

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,418 @@
...
+		// align left and remove margin
+		margin-left: 0;
+		span.dialog__actions-separator {
+			margin-left: auto;
+		}
+	}
+
+	:deep(.input-field__helper-text-message) {
+		// reduce helper text standing out
+		color: var(--color-text-maxcontrast);
```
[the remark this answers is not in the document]
susnux: not resolved

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogDatePassword.vue
```diff
@@ -0,0 +1,235 @@
...
+				:disabled="disabled || enforcePasswordForPublicLink"
+				@update:checked="onTogglePassword">
+				{{ t('files_sharing', 'Set a password') }}
+			</NcCheckboxRadioSwitch>
+
+			<div v-if="password !== null" class="file-request-dialog__password-field">
+				<NcPasswordField ref="passwordField"
+					:check-password-strength="true"
+					:disabled="disabled"
+					:label-outside="true"
```
susnux, line 62: I do not see any linked the label?

I think it would be better to just add the "inside" label here, could be simple as `Password`.

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogDatePassword.vue
```diff
@@ -0,0 +1,235 @@
...
+		</fieldset>
+
+		<!-- Password -->
+		<fieldset class="file-request-dialog__password" data-cy-file-request-dialog-fieldset="password">
+			<NcNoteCard v-if="enforcePasswordForPublicLink" type="info">
+				{{ t('files_sharing', 'Your administrator has enforced a password protection.') }}
+			</NcNoteCard>
+
+			<!-- Enable password -->
+			<legend>{{ t('files_sharing', 'What password should be used for the request ?') }}</legend>
```
susnux, line 50:
```suggestion
			<legend>{{ t('files_sharing', 'What password should be used for the request?') }}</legend>
```

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogDatePassword.vue
```diff
@@ -0,0 +1,235 @@
...
+			<NcCheckboxRadioSwitch v-show="!defaultExpireDateEnforced"
+				:checked="defaultExpireDateEnforced || expirationDate !== null"
+				:disabled="disabled || defaultExpireDateEnforced"
+				@update:checked="onToggleDeadline">
+				{{ t('files_sharing', 'Set a submission expirationDate') }}
+			</NcCheckboxRadioSwitch>
+
+			<!-- Date picker -->
+			<NcDateTimePickerNative v-if="expirationDate !== null"
+				id="file-request-dialog-expirationDate"
```
susnux, line 30: unused id?

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogDatePassword.vue
```diff
@@ -0,0 +1,235 @@
...
+				:disabled="disabled"
+				:hide-label="true"
```
susnux, line 32: I do not see any label? This needs to have either a linked label or just use the inline one

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogFinish.vue
```diff
@@ -0,0 +1,224 @@
...
+			:value="shareLink"
+			:label="t('files_sharing', 'Share link')"
+			:readonly="true"
+			:show-trailing-button="true"
+			:trailing-button-label="t('files_sharing', 'Copy to clipboard')"
+			@click="copyShareLink"
+			@click-trailing-button="copyShareLink">
+			<template #trailing-button-icon>
+				<IconCheck v-if="isCopied" :size="20" @click="isCopied = false" />
+				<IconClipboard v-else :size="20" @click="copyShareLink" />
```
susnux, line 24:
```suggestion
				<IconCheck v-if="isCopied" :size="20" />
				<IconClipboard v-else :size="20" />
```

No interaction listeners on non interactive elements.
This should be done in `copyShareLink`

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogFinish.vue
```diff
@@ -0,0 +1,224 @@
...
+			</template>
+		</NcInputField>
+
+		<template v-if="isShareByMailEnabled">
+			<!-- Email share-->
+			<NcTextField :value.sync="email"
+				:label="t('files_sharing', 'Send link via email')"
```
susnux, line 31:
```suggestion
				:label="t('files_sharing', 'Recipients email address')"
```

or just `Email address` or similar, as this label currently sounds like a checkbox instead of an address input

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogFinish.vue
```diff
@@ -0,0 +1,224 @@
...
+	data() {
+		return {
+			isCopied: false,
+			email: '',
+		}
+	},
+
+	computed: {
+		shareLink() {
+			return window.location.protocol + '//' + window.location.host + generateUrl('/s/') + this.share.token
```
susnux, line 121:
```suggestion
			return generateUrl('/s/{token}', { token: this.share.token }, { baseURL: getBaseUrl() })
```

(`getBaseUrl` from `@nextcloud/router`)

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogFinish.vue
```diff
@@ -0,0 +1,224 @@
...
+				showSuccess(this.n('files_sharing', '1 email address added', '{count} email addresses added', validEmails.length, { count: validEmails.length }))
+			}
+
+			this.email = ''
+		},
+
+		isValidEmail(email) {
+			const regExpEmail = /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
+			return regExpEmail.test(email)
+		},
```
susnux, line 209: Really going into that rabbit hole? I am pretty sure this does not cover all valid cases.
We validate on server properly so I think we can just use something that validates that it is roughly an email like `/[^@]+@[^@]+\.[a-z]{2,}/` (this covers all valid emails except for local (without TLD) and for quoted `@` in name part.

But feel free to ignor this :)

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
```diff
@@ -0,0 +1,153 @@
...
+<template>
+	<div>
+		<!-- Request label -->
+		<fieldset class="file-request-dialog__label" data-cy-file-request-dialog-fieldset="label">
+			<legend>
+				{{ t('files_sharing', 'What are you requesting ?') }}
+			</legend>
+			<NcTextField :value="label"
+				:disabled="disabled"
+				:label-outside="true"
```
susnux, line 15: Where is the label? The `legend` is not automatically associated with the input but with the fieldset.
So the fieldset has a name, but the input does not. Both need an accessible name.

Maybe add at least an `aria-label` like `Request subject`.

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
```diff
@@ -0,0 +1,153 @@
...
+  - SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+
+<template>
+	<div>
+		<!-- Request label -->
+		<fieldset class="file-request-dialog__label" data-cy-file-request-dialog-fieldset="label">
+			<legend>
+				{{ t('files_sharing', 'What are you requesting ?') }}
```
susnux, line 11:
```suggestion
				{{ t('files_sharing', 'What are you requesting?') }}
```

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
```diff
@@ -0,0 +1,153 @@
...
+
+		<!-- Request destination -->
+		<fieldset class="file-request-dialog__destination" data-cy-file-request-dialog-fieldset="destination">
+			<legend>
+				{{ t('files_sharing', 'Where should these files go ?') }}
+			</legend>
+			<NcTextField :value="destination"
+				:disabled="disabled"
+				:helper-text="t('files_sharing', 'The uploaded files are visible only to you unless you choose to share them.')"
+				:label-outside="true"
```
susnux, line 30: Same here no accessible name.

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
```diff
@@ -0,0 +1,153 @@
...
+		</fieldset>
+
+		<!-- Request note -->
+		<fieldset class="file-request-dialog__note" data-cy-file-request-dialog-fieldset="note">
+			<legend>
+				{{ t('files_sharing', 'Add a note') }}
+			</legend>
+			<NcTextArea :value="note"
+				:disabled="disabled"
+				:label-outside="true"
```
susnux, line 54: Same here no accessible name

susnux (dismissed):
Some comments, but especially a lot of interactive elements / input elements are missing an accessible name, which is a real issue.

apps/files_sharing/lib/Controller/ShareAPIController.php
skjnldsv: No, it doesn't work. See other legacy values like `$hideDownload` or `$publicUpload`, they need to be string for god knows why.
I don't want to change those areas as well. I'll be consistent with the legacy implementation for now :)

apps/files_sharing/src/components/NewFileRequestDialog.vue
skjnldsv: I'd say, for an input component with a name attribute, `input[name="destination"] is a pretty clean approach.
I ain't gonna rewrite a `validate` implementation on the nextcloud vue components. That would mean we have to do it for like 5+ different components which are far out of scope here

apps/files_sharing/src/components/NewFileRequestDialog.vue
skjnldsv: That sentence came right from Jan. I'll adjust anyway, no big deal

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,423 @@
...
+				:disabled="loading"
+				data-cy-file-request-dialog-controls="back"
+				type="tertiary"
+				@click="currentStep = STEP.FIRST">
+				{{ t('files_sharing', 'Previous') }}
+			</NcButton>
+
+			<!-- Next -->
+			<NcButton v-if="currentStep !== STEP.LAST"
+				:aria-label="t('files_sharing', 'Continue')"
```
susnux, line 77: Just to note: You do not need aria labels if the button already has text (it does not hurt, but if the button has text the aria label should never be different from the text so it makes it easier to only maintain on location).

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -0,0 +1,423 @@
...
+			<span class="dialog__actions-separator" />
+
+			<!-- Back -->
+			<NcButton v-show="currentStep === STEP.SECOND"
+				:aria-label="t('files_sharing', 'Previous step')"
+				:disabled="loading"
+				data-cy-file-request-dialog-controls="back"
+				type="tertiary"
+				@click="currentStep = STEP.FIRST">
+				{{ t('files_sharing', 'Previous') }}
```
susnux, line 72: Visual and accessible name differ, this is an accessibility issue.

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogDatePassword.vue
susnux: oh.

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogDatePassword.vue
susnux:
```suggestion
				:hide-label="true"
				:label="t('files_sharing', 'Expiration date')"
```

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogDatePassword.vue
susnux:
```suggestion
					:label="t('files_sharing', 'Password')"
```

or 

```suggestion
					:aria-label="t('files_sharing', 'Password')"
```

apps/files_sharing/src/components/NewFileRequestDialog.vue
skjnldsv: Noted for followup! :+1: 
Thanks

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
susnux:
```suggestion
				:label="t('files_sharing', 'Request subject')"
```

or

```suggestion
				:aria-label="t('files_sharing', 'Request subject')"
				:label-outside="true"
```

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
susnux:
```suggestion
				:label="t('files_sharing', 'Upload destination')"
```

or

```suggestion
				:aria-label="t('files_sharing', 'Upload destination')"
				:label-outside="true"
```

apps/files_sharing/src/components/NewFileRequestDialog/NewFileRequestDialogIntro.vue
susnux:
```suggestion
				:label="t('files_sharing', 'Note for recipient')"
```

or

```suggestion
				:aria-label="t('files_sharing', 'Note for recipient')"
				:label-outside="true"
```

susnux:
List of accessibility bugs I think that needs to be resolved before  (unrelated to design or code style or other):
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675905122
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675928843
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675932604
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675935792
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675716753
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675939663
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675942342
- [ ] https://github.com/nextcloud/server/pull/46007/files#r1675944287

apps/files_sharing/src/components/NewFileRequestDialog.vue
susnux: Simple:

```suggestion
				{{ t('files_sharing', 'Previous step') }}
```

nfebe (approved):

susnux (approved):
thank you so much!

lib/public/Share/IAttributes.php
```diff
@@ -13,40 +13,41 @@
...
 	 * 			"key" => <string>,
-	 * 			"enabled" => <bool>
+	 * 			"value" => <bool|string|array|null>,
 	 * 		],
 	 * 	...
 	 * ]
 	 *
 	 * @return array formatted IAttributes
 	 * @since 25.0.0
+	 * @since 30.0.0, `enabled` was renamed to `value`
```
nickvergessen, line 50: @skjnldsv Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(escpecially this "breakage", but also all changes in lib/public/ (even if you type only))

skjnldsv: image

juliushaertl: Also FYI @tobiasKaminsky since this is actually used when you want to set the attribute through the OCS API.

tobiasKaminsky: which attribute? :see_no_evil:

juliushaertl: Share attributes for the "Allow download" option on internal shares, not sure if clients make use of that

https://docs.nextcloud.com/server/latest/developer_manual/client_apis/OCS/ocs-share-api.html#share-attributes

Screenshot 2024-07-19 at 11 07 33

skjnldsv: https://github.com/nextcloud/documentation/pull/12043

PR #32781, opened 2022-06-09
fix: add moveFromStorage for objectstorage to fix transfer ownership on s3

unteem:
This PR fixes https://github.com/nextcloud/server/issues/25693 and https://github.com/nextcloud/server/issues/31870

When transferring ownership on s3 primary storage it was doing a copy and delete leading to shares being linked to non-existing files and failing to transfer

Implementing `moveFromStorage` in `lib/private/Files/ObjectStore/ObjectStoreStorage.php` fixes it. 

It doesn't fix the `isSameStorage` (https://github.com/nextcloud/server/blob/e56a072fa73ec6dc717569a169c532e07443d782/lib/private/Files/Storage/Common.php#L678) that should return true as mentioned in https://github.com/nextcloud/server/issues/31870 which anyway leads to another issue. If this condition passes it then goes to `rename` which only works for renaming a folder/files in the same user storage.

Issue #25693, opened 2021-02-17
Transfer Ownership: "Share with id XX points at deleted file, skipping" and shares lost during transfer


mseiwald:
Hi,

I want to transfer files/folders and the existing shares on those folders from one user to the other using the command:

```su -p www-data -s /bin/sh -c "php /var/www/html/occ files:transfer-ownership --path oldUser newUser";```

This is an example output:

```Validating quota
Analysing files of oldUser ...
    0 [->--------------------------]
Collecting all share information for files and folders of oldUser ...
    4 [============================]
Transferring files to newUser/files/transferred from oldUser on 2021-02-17 12-33-27 ...
Restoring shares ...
Share with id 27 points at deleted file, skipping
 1/4 [=======>--------------------]  25%Share with id 28 points at deleted file, skipping
Share with id 29 points at deleted file, skipping
Share with id 30 points at deleted file, skipping
 4/4 [============================] 100%
```

First of all I don't know what the error `Share with id XX points at deleted file, skipping` means. I found previous reports related to quota etc. In my case both users have a large enough quota and there are no deleted files. I'm getting this error everytime even with an empty test folder.

Furthermore, shares on the transferred folders are lost.

I would appreciate any hint on how to troubleshoot this issue and fix it.

Thanks!

Nextcloud version: 20.0.5

acsfer:
> Furthermore, shares on the transferred folders are lost.

Check the doc: https://docs.nextcloud.com/server/20/admin_manual/configuration_server/occ_command.html#file-operations

> Incoming shares are not moved because the sharing user holds the ownership of the respective files and therefore owns the decision which actual users/groups to be shared with.

mseiwald:
Sorry if I wasn't clear enough, I was only referring to outgoing shares (i.e. shared folders **owned** by the old user) which are also lost but shouldn't be according to the documentation:

> Outgoing shares are moved as well.

szaimen:
Is this Issue still valid in NC21.0.2? If not, please close this issue. Thanks! :)

pierreozoux:
We experience the same with a Nc22. We'll share details in a moment.

unteem:
I was also thinking about an alternative to transfer ownership.

Instead of copying a new file and deleting the original one (at least in the database)

in the filecache table changing the storageID to the new one could be enough
then for shares uid_owner should be changed and we are done 

[...]

PR #46013, opened 2024-06-20
fix: write object to the correct urn when moving from another storage to object store

icewind1991:
The default `copyFromStorage` + `delete` approach loads to the object being written to the wrong urn.

See also https://github.com/nextcloud/server/pull/43660 for a description of the problem and alternate approach.

### DIFF
--- a/lib/private/Files/ObjectStore/ObjectStoreStorage.php
+++ b/lib/private/Files/ObjectStore/ObjectStoreStorage.php
@@ -594,6 +594,31 @@ class ObjectStoreStorage extends \OC\Files\Storage\Common implements IChunkedFil
 		return parent::copyFromStorage($sourceStorage, $sourceInternalPath, $targetInternalPath);
 	}
 
+	public function moveFromStorage(IStorage $sourceStorage, $sourceInternalPath, $targetInternalPath, ?ICacheEntry $sourceCacheEntry = null): bool {
+		$sourceCache = $sourceStorage->getCache();
+		if (!$sourceCacheEntry) {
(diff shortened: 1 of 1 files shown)


lib/private/Files/ObjectStore/ObjectStoreStorage.php
```diff
@@ -594,6 +594,30 @@ public function copyFromStorage(
...
+				$this->moveFromStorage($sourceStorage, $child->getPath(), $targetInternalPath . '/' . $child->getName());
+			}
+			$sourceStorage->rmdir($sourceInternalPath);
+		} else {
+			// move the cache entry before the contents so that we have the correct fileid/urn for the target
+			$this->getCache()->moveFromCache($sourceCache, $sourceInternalPath, $targetInternalPath);
+			try {
+				$this->writeStream($targetInternalPath, $sourceStorage->fopen($sourceInternalPath, 'r'), $sourceCacheEntry->getSize());
+			} catch (\Exception $e) {
+				$this->getCache()->remove($targetInternalPath);
```
come-nc, line 613: Should we also put back the file in the source cache on exceptions?
Here it’s lost.

icewind1991: done

artonge (approved):

kesselb (approved):
I've run into the problem, that moving a file from S3 external to S3 primary uses the wrong urn, on my development setup. The patch resolves the problem.

joshtrichards:
On the surface, this appears like it *could* (untested) also fix: #31870 / #25693 (which was attempted in #32781).

PR #46021, opened 2024-06-20
perf: delay getting (sub)admin status for user in the security middleware untill we need it

icewind1991:
No need to check if we're admin for non-admin routes.

Should save 1 or 2 queries for most routes


nickvergessen (approved):

yemkareems (approved):

susnux:
> Conflicting files
> tests/lib/AppFramework/Middleware/Security/SecurityMiddlewareTest.php

susnux (approved):

PR #46030, opened 2024-06-21
fix(FilesView): Update files view upon share creation/delete

### LINKED ISSUE #40263: Share node change between sidebar and Files app
> I tested this and the icon does not seem to directly update after I shared the folder for example. I guess this should be the case though? Apart from that seems to work well so good work!

Needs a separate fix. We'll need to do a pass on global events. But right now the Sidebar doesn't update the files store, you are correct :)

_Originally posted by @skjnldsv in https://github.com/nextcloud/server/issues/40192#issuecomment-1703257591_

### LINKED ISSUE #44961: [Bug]: Reflect shared state in file list after sharing without reload

### Bug description

After sharing a folder the folder icon does not reflect the fact that the folder was shared.
Neither does the share column which:

[Bildschirmaufzeichnung vom 22.04.2024, 13:14:03.webm](https://github.com/nextcloud/server/assets/97337118/6c0d7e7c-25fe-42f9-9b60-1a410e77b3d7)

### Steps to reproduce

1. Create a new folder in the files app
2. Open the sharing tab for that folder
3. Share the folder
4. See that folder icon and icon in the share column remain the same

### Expected behavior

1) Folder icon changes to a shared folder icon
2) share column shows `shared` just like for other shared files.

### Server Info

I've been able to reproduce this with
* on cloud.nextcloud.com
* with stable29
* with stable28
* on firefox and chromium

Therefore I don't think it depends on the browser used nor the server configuration.

nfebe:
## Summary

* Resolves : https://github.com/nextcloud/server/issues/44961
* Resolves : #40263 

## Before 

[Bildschirmaufzeichnung vom 22.04.2024, 13:14:03.webm](https://github.com/nextcloud/server/assets/97337118/6c0d7e7c-25fe-42f9-9b60-1a410e77b3d7)

## After

[refresh-files-view.webm](https://github.com/nextcloud/server/assets/14317775/78033873-2ae8-4943-946b-76fe41485a91)


apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -334,12 +337,17 @@ export default {
 			isFirstComponentLoad: true,
 			test: false,
 			creating: false,
+			logger,
```
susnux, line 340: The logger should not be made reactive. So just import it but not add to the component.

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -896,6 +913,9 @@ export default {
 				this.queueUpdate(...permissionsAndAttributes)
 			}
 
+			await this.getNode()
+			emit('files:node:updated', this.node)
```
susnux, line 917: I am not sure, as this might cause additional fetch in the files list due to store handling.
Maybe here it would be enough to just add the share type to `node.attribute['share-types']` ?

But just thinking loud, so this is probably fine :)

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -727,15 +735,24 @@ export default {
...
 	},
 
 	methods: {
+		async getNode() {
+			const node = { path: this.path }
+			try {
+				this.node = await fetchNode(node)
+				this.logger.info('Fetched Node:', this.node)
+			} catch (error) {
+				this.logger.error('Error:', error)
```
susnux, line 753: This is even custom formatting for errors in the logger if you just pass it like:
```suggestion
				this.logger.error(error)
```

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -727,15 +735,24 @@ export default {
 	beforeMount() {
 		this.initializePermissions()
 		this.initializeAttributes()
-		console.debug('shareSentIn', this.share)
-		console.debug('config', this.config)
+		this.logger.debug('shareSentIn', this.share)
+		this.logger.debug('config', this.config)
```
susnux, line 739: Note that the logger is not like `console.debug`, the first argument is a message, but the second is the whole context object which will be merged with internal properties like `user` or `app`.

So in most cases you should just use it like this:
```js
		this.logger.debug('Your message', { config: this.config })
```

With one exception: the `error` property of the context will be specially formatted for exceptions
(like `logger.debug('message', { error: someErrorInstance })`)

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -276,6 +276,9 @@ import Share from '../models/Share.js'
 import ShareRequests from '../mixins/ShareRequests.js'
 import ShareTypes from '../mixins/ShareTypes.js'
 import SharesMixin from '../mixins/SharesMixin.js'
+import { fetchNode } from '../services/WebdavClient'
+import logger from '../services/logger'
+import { emit } from '@nextcloud/event-bus'
```
susnux, line 281: nitpick: maybe sort this up to the other named imports from external modules (line 248)

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -727,15 +734,24 @@ export default {
...
 	mounted() {
 		this.$refs.quickPermissions?.querySelector('input:checked')?.focus()
 	},
 
 	methods: {
+		async getNode() {
+			const node = { path: this.path }
+			try {
+				this.node = await fetchNode(node)
+				logger.info('Fetched Node:', this.node)
```
susnux, line 750: nitpick: Prevent the context from being merged with the node:

```suggestion
				logger.debug('Fetched Node:', { node: this.node })
```

susnux (approved):

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -727,15 +734,24 @@ export default {
 	beforeMount() {
 		this.initializePermissions()
 		this.initializeAttributes()
-		console.debug('shareSentIn', this.share)
-		console.debug('config', this.config)
+		logger.debug('Share object received', { share: this.share })
+		logger.debug('Configuration object received', { config: this.config })
```
max-nextcloud, line 738: Somehow this does not work for me when I try it locally.
Having run `npm run build` I get:
grafik

apps/files_sharing/src/views/SharingDetailsTab.vue
```diff
@@ -245,7 +245,10 @@
 </template>
 
 <script>
+import { emit } from '@nextcloud/event-bus'
+import { fetchNode } from '../services/WebdavClient'
 import { getLanguage } from '@nextcloud/l10n'
+import { logger } from '../services/logger'
```
max-nextcloud, line 251:
```suggestion
import logger from '../services/logger'
```
Fixes the issue i reported below for me.

max-nextcloud:
After adjusting the logger import sharing with alice works nicely.
However if I just create a link share the left side is still not updated.

[Bildschirmaufzeichnung vom 25.06.2024, 06:06:20.webm](https://github.com/nextcloud/server/assets/97337118/e06abc2a-60d5-4995-a479-9b4aea5c74e1)

Would be nice to tackle that as well.

szaimen:
This also resolves https://github.com/nextcloud/server/issues/40263, I suppose?

max-nextcloud (approved):
Tried again... works like a charm now :magic_wand: 

Thanks! :heart:

PR #46058, opened 2024-06-23
Fix task processing input validation

julien-nc:
It was dropping some optional inputs.

### DIFF
--- a/lib/private/TaskProcessing/Manager.php
+++ b/lib/private/TaskProcessing/Manager.php
@@ -489,7 +489,7 @@ class Manager implements IManager {
 	 * @psalm-template T
 	 */
 	private function removeSuperfluousArrayKeys(array $array, ...$specs): array {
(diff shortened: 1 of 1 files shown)


marcelklehr (approved):
:see_no_evil: I recently discovered that I've been using the + operator wrong

bigcat88 (approved):

julien-nc:
@marcelklehr I had to adjust again to make the tests pass. `removeSuperfluousArrayKeys` now only tries to keep the keys that are defined in the task input/output array. Tests were failing because optional values might not be set.

As the purpose of this method is to remove keys, this fix seems fine to me. But do we actually check that all the input shape fields are defined in the task? If not, do you think it should be done?

marcelklehr:
> But do we actually check that all the input shape fields are defined in the task? If not, do you think it should be done?

Yes, the validateInput and validateOutput methods check this

marcelklehr (approved):

PR #46068, opened 2024-06-24
fix(ExternalSharing): Handle template shares from external sources

nfebe:
## Summary

Update UI correctly upon guest invite. 

_The new sharing flow requires or implies that users should edit share before creating.
External sources should not created the share IF we would upon sharing details tab on first request._

-----

The new sharing flow since NC27 requires that users open the sharing details tab and
customize their share before creating it.

In https://github.com/nextcloud/server/pull/39472 the work of `addShare` was handed down to
`openSharingDetails` that opens the sharing details tab for the user to customize and manually
creat their share.

Related : https://github.com/nextcloud/guests/pull/1189 

## TODO

- [x] Test with https://github.com/nextcloud/guests/pull/1189


artonge (approved):

Fenn-CS:
@nextcloud/server-frontend

susnux (approved):

Issue #42540, opened 2024-01-01
[Bug]: SQL errors generated by background job OC\\Preview\\BackgroundCleanupJob


timwalls:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description
I recently upgraded my NextCloud instance to 27.1.5; since then I’ve started seeing SQL errors in my PostgreSQL logs like this:

```
2023-12-29 06:00:01.624 UTC [153025] STATEMENT:  SELECT "a"."name" FROM "oc_filecache" "a" LEFT JOIN "oc_filecache" "b" ON CAST("a"."name" AS INT) = "b"."fileid" WHERE ("a"."storage" = $1) AND ("b"."fileid" IS NULL) AND ("a"."path" LIKE $2) AND ("a"."mimetype" = $3)
2023-12-29 07:00:06.333 UTC [153593] ERROR:  invalid input syntax for type integer: "2048-1152-max.png"
```

These happen roughly once an hour, which suggests it’s something being called in a cronjob.

It doesn’t look like a data error per-se; the query is just wrong - the oc_filecache.name field is a varchar, and this query seems to try to cast that to an integer. I’m not a PostgreSQL expert though, maybe that’s a thing that’s supposed to be possible :-).

I checked the nextcloud server log, and the errors appear to originate in OC\\Preview\\BackgroundCleanupJob; I've snipped (`[...]`) this down to the interesting bits to meet the ticket length limit, but a full row is pasted in the 'logs' section:

```
{
  "reqId": "yqyioTldcIjXCEeOdUGe",
  "level": 3,
   "2024-01-01T10:20:01+00:00",
  "remoteAddr": "",
  "user": "--",
  "app": "core",
  "method": "",
  "url": "--",
  "message": "Error while running background job (class: OC\\Preview\\BackgroundCleanupJob, arguments: )",
[...]

PR #46072, opened 2024-06-24
Refined BackgroundCleanupJob queries for robustness, allowing CACHEDIR.TAG in preview root

dennisverspuij:
I personally like NextCloud AIO to exclude previews from backup, by adding a appdata_*/preview/CACHEDIR.TAG file.
However the queries of the \OC\Preview\BackgroundCleanupJob (may) fail because of this,
This PR fixes this by making the queries more precise.


kesselb:
Related: https://github.com/nextcloud/server/issues/42540

Haven't tested it but looks reasonable. 

I am not sure if I like the idea of having unrelated files there, but I see the use case.

artonge (approved):

come-nc:
Hello,

https://github.com/nextcloud/server/pull/48581 has been merged to fix the same thing, and it showed better performances. So, closing this one.
Thank you for your contribution!

PR #46076, opened 2024-06-24
feat(webhooks): Add support for a userid filter

come-nc:
## Summary

This allows to register a userId to filter on along with the webhooks. This webhook will then only be triggered if the given userId is the one in session. This is more efficient than filtering by user in the event filter because the listener is not even registered if the user id does not match.

## Checklist

### DIFF
--- a/apps/webhook_listeners/lib/AppInfo/Application.php
+++ b/apps/webhook_listeners/lib/AppInfo/Application.php
@@ -16,6 +16,7 @@ use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
 use OCP\EventDispatcher\IEventDispatcher;
+use OCP\IUserSession;
 use Psr\Container\ContainerInterface;
 use Psr\Log\LoggerInterface;
 
@@ -40,9 +41,10 @@ class Application extends App implements IBootstrap {
 	): void {
 		/** @var WebhookListenerMapper */
 		$mapper = $container->get(WebhookListenerMapper::class);
+		$userSession = $container->get(IUserSession::class);
 
 		/* Listen to all events with at least one webhook configured */
-		$configuredEvents = $mapper->getAllConfiguredEvents();
+		$configuredEvents = $mapper->getAllConfiguredEvents($userSession->getUser()?->getUID());
 		foreach ($configuredEvents as $eventName) {
 			$logger->debug("Listening to {$eventName}");
 			$dispatcher->addServiceListener(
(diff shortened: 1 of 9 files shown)


come-nc:
I did not test it in real conditions yet, but code is ready for review.

apps/webhook_listeners/lib/Db/WebhookListenerMapper.php
```diff
@@ -181,27 +189,38 @@ private function getAllConfiguredEventsFromDatabase(): array {
...
+					'user_id_filter',
+					$qb->createNamedParameter(array_unique(['',$userId ?? '']), IQueryBuilder::PARAM_STR_ARRAY),
+				)
+			);
 
 		return $this->findEntities($qb);
 	}
+
+	private function buildCacheKey(?string $userIdFilter = ''): string {
+		return self::EVENTS_CACHE_KEY_PREFIX.'_'.($userIdFilter ?? '');
```
bigcat88, line 224: The null coalescing operator (??) is used to provide a default value if the variable is null. 

Since the default value of $userIdFilter is an empty string (''), and an empty string is not null, the null coalescing operator here is unnecessary.

apps/webhook_listeners/lib/Migration/Version1000Date20240527153425.php
```diff
@@ -53,6 +53,10 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 			$table->addColumn('event_filter', Types::TEXT, [
 				'notnull' => false,
 			]);
+			$table->addColumn('user_id_filter', Types::STRING, [
+				'notnull' => true,
```
bigcat88, line 57: When using Oracle database the empty value is `null`, afaik..

apps/webhook_listeners/lib/AppInfo/Application.php
```diff
@@ -40,9 +41,10 @@ private function registerRuleListeners(
 	): void {
 		/** @var WebhookListenerMapper */
 		$mapper = $container->get(WebhookListenerMapper::class);
+		$userSession = $container->get(IUserSession::class);
 
 		/* Listen to all events with at least one webhook configured */
-		$configuredEvents = $mapper->getAllConfiguredEvents();
+		$configuredEvents = $mapper->getAllConfiguredEvents($userSession->getUser()?->getUID());
```
marcelklehr, line 47: This potentially calls the method with `null` but the method doesn't accept null

come-nc: The method does accept null

marcelklehr (approved):

bigcat88 (approved):

PR #46083, opened 2024-06-24
refactor(files): Migrate logger to Typescript

susnux:
1. This is just a tiny rename to make our files code base more consistent by using Typescript where possible.
2. Plus another small refactor to fix invalid doc block and resort imports


Pytal (commented):
Maybe make it a named export?

artonge (approved):

susnux:
> Maybe make it a named export?

In general I would agree, here it is so small and single use-case, so both work. I have not strong opinion on either of them.
Meaning I can do so if that is preferred :sweat_smile:

Pytal (approved):

skjnldsv (approved):

PR #43428, opened 2024-02-07
feat(share): save date and time for expiration

Altahrim:
## Summary

When we store an expiration date for a share, we store only the date, without time.
If an user is in UTC + 9 and share a file before 9am, the server will consider the expiration date to be one day earlier (ie. D+6 instead of D+7). Then, the display is wrong.

To fix that, we have to store time.

Fix #43457 
Fix https://github.com/nextcloud/server/issues/43243

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #46097, opened 2024-06-25
fix(files_sharing): Also set the expiration date timezone during validation

artonge:
This is needed as we want to store the difference between the server's and the user's timezones.

Follow-up of https://github.com/nextcloud/server/pull/43428


nfebe (approved):
🐘

Altahrim (approved):

PR #16828, opened 2019-08-21
🔗☑️ Accept all incoming shares

nickvergessen:
- [x] Respect the accepted flag from the share table (not used at all at the moment)
- [x] Generate a notification for a user when an individual share is done
- [x] Generate a notification for all members when a group share is done
- [x] Generate notifications for all group shares if a user is added to a group
- [x] Remove the notifications when the share is deleted
- [x] Remove the notifications when a member is removed from a group
- [x] Update the accepted flag to 1 (accepted) for all existing shares on update
- [ ] Adjust the email that is sent for user shares or drop it

PR #46099, opened 2024-06-25
fix: Add `acceptShare` as an interface

susnux:
## Summary

It was commented out because at that time no other share provider supported the `acceptShare` method. Today it is the same no other provider supports it, but we should make it discoverable by adding it to the public API as an additional interface.

5 years after no other provider supports it, so maybe an additional interface would be better like done here?
Ref: https://github.com/nextcloud/server/pull/16828#discussion_r321740211

## Checklist

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -44,6 +44,7 @@ use OCP\Share\IManager;
 use OCP\Share\IProviderFactory;
 use OCP\Share\IShare;
 use OCP\Share\IShareProvider;
+use OCP\Share\IShareProviderSupportsAccept;
 use OCP\Share\IShareProviderWithNotification;
 use Psr\Log\LoggerInterface;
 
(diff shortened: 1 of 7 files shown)


lib/public/Share/IShareProviderSupportsAccept.php
```diff
@@ -0,0 +1,27 @@
...
+ * @since 30.0.0
+ */
+interface IShareProviderSupportsAccept {
+	/**
+	 * Accept a share.
+	 *
+	 * @param IShare $share
+	 * @param string $recipient
+	 * @return IShare The share object
+	 * @since 17.0.0
```
nickvergessen, line 24:
```suggestion
	 * @since 30.0.0
```

lib/public/Share/IShareProviderSupportsAccept.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Interface IShareProviderSupportsAccept
+ *
+ * This interface allows to define IShareProvider that can handle the `acceptShare` method,
+ * which is available since Nextcloud 17.
+ *
+ * @since 30.0.0
+ */
+interface IShareProviderSupportsAccept {
```
nickvergessen, line 17: Could also:
```suggestion
interface IShareProviderSupportsAccept extends IShareProvider {
```

but that might only work atm while it's still linear?

lib/private/Share20/Manager.php
```diff
@@ -1018,6 +1019,7 @@ public function acceptShare(IShare $share, string $recipientId): IShare {
 			// TODO FIX ME
```
nickvergessen, line 1019: check for the new interface instead?

lib/public/Share/IShareProviderSupportsAccept.php
susnux: We often just add new interfaces so you can extend both or just one.
I am fine with both, what do you prefer?

come-nc (approved):

skjnldsv (approved):

lib/public/Share/IShareProviderSupportsAccept.php
susnux: > I'd say as you can not be one without the other it makes sense to extend?

Makes sense and the `IShareProviderWithNotification` was implemented similar, so :+1:

nickvergessen (approved):

susnux:
Forgot to add the autoload stuff...

PR #46106, opened 2024-06-25
fix(session): Log when session_* calls are slow

ChristophWurst:
* Resolves: # 

## Summary

session_start and friends can take a long time when the session is locked. This slows down Nextcloud. It might be a good idea to catch when this happens.

* session_* needs more than 25s -> error
* session_* needs more than 10s -> warning
* session_* needs more than 0.5s -> info
* session_* needs more than 0.1s -> debug

## TODO

- [x] ...

## Checklist

### DIFF
--- a/lib/private/Session/Internal.php
+++ b/lib/private/Session/Internal.php
@@ -11,7 +11,11 @@ namespace OC\Session;
 
 use OC\Authentication\Token\IProvider;
 use OCP\Authentication\Exceptions\InvalidTokenException;
+use OCP\ILogger;
 use OCP\Session\Exceptions\SessionNotAvailableException;
+use Psr\Log\LoggerInterface;
+use function call_user_func_array;
+use function microtime;
 
 /**
  * Class Internal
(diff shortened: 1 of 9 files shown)


juliusknorr (approved):

Altahrim (approved):

lib/private/Session/Internal.php
```diff
@@ -183,11 +188,31 @@ public function trapError(int $errorNumber, string $errorString) {
...
-				return @call_user_func_array($functionName, $parameters);
+				$result = @call_user_func_array($functionName, $parameters);
 			} else {
-				return call_user_func_array($functionName, $parameters);
+				$result = call_user_func_array($functionName, $parameters);
 			}
+			$timeAfter = microtime(true);
+			$timeSpent = $timeAfter - $timeBefore;
+			if ($timeSpent > 0.1) {
+				$logLevel = match (true) {
```
nickvergessen, line 200: should we add a function which takes the time list as an argument? :-X 
Doing this in so many places already

nickvergessen (approved):

lib/private/Session/Internal.php
ChristophWurst: I'm not sure how this can be abstracted the best way. The thresholds do not always follow the same pattern. Sometimes it's linear, other times it's exponential.

ChristophWurst:
Cypress is probably failing for unrelated reasons

juliushaertl:
Odd that I cannot merge, but let me rebase again then

Screenshot 2024-08-07 at 09 00 45

PR #46114, opened 2024-06-25
fix(user_ldap): Avoid extra LDAP request when mapping a user for the first time

come-nc:
## Summary

Avoids using several LDAP searches to get UUID, display name and
 internal name, now gets all attributes at the same time.
Also avoids extra request to build an unused user object in userExists.

## TODO

- [x] ~~Avoid extra request to fetch the display name by directly caching it. (The display name is fetched by dav to update the calendar entry as soon as the user is mapped).~~ This is too complicated because of internal quirks, keeping an extra request for the display name for now.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Access.php
+++ b/apps/user_ldap/lib/Access.php
@@ -122,10 +122,10 @@ class Access extends LDAPUtility {
 	 *
 	 * @param string $dn the record in question
 	 * @param array $attrs the attributes that shall be retrieved
-	 *        if empty, just check the record's existence
+	 *                     if empty, just check the record's existence
 	 * @param string $filter
 	 * @return array|false an array of values on success or an empty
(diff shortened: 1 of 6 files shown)


come-nc:
Tests are failing indicating problems with attributes case, but I have a hard time understanding how it’s supposed to be.

There is this worrying comment:
https://github.com/nextcloud/server/blob/3a97dbf248b3e581b5782a638743958eb6f2a640/apps/user_ldap/lib/Configuration.php#L301-L303

It’s not the only one a few of the attributes are not lower cased by the configuration it seems, but as far as I can see all methods getting record from the LDAP do lower case the keys so I’m not sure how it works.

artonge (approved):

blizzz (approved):
did not test, but looks good and reasonable

Issue #30989, opened 2022-02-03
occ flag to create a debug report


come-nc:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
Often when asking someone to run commands to diagnose a problem on their installation, debug log would help understanding the command result. This is especially true for LDAP related issues.

**Describe the solution you'd like**
Add a general flag to the occ command like `--report` that:
- Turn on debug log level for the command (same as setting env var NC_loglevel=0)
- Create a report file (zip or txt or json, not sure which is best)
- Put in this report:
  - The command and its parameters
  - The output of the command and return code
  - The log lines from the command execution (all new log lines since the command started is fine)
  - If possible the trace of the eventual Exception? Most occ commands have a surrounding try/catch that hides the trace. Maybe the elegant solution here is to log as debug the exception so that it will be in the log.

**Describe alternatives you've considered**

**Additional context**

PR #46115, opened 2024-06-25
feat(occ): Add a --debug option to output all log levels to the output

come-nc:
Related to #30989 

## Summary

Adds a universal option to pass to occ commands so that all levels of logs are printed in the output, even if the loglevel on the instance is higher.
This allows to inspect all the logging related to a command without having it mixed up with logs from other stuff going on in the server. Also allows to have logs and command output in the same place to see in which order things happen.
This is especially useful with user_ldap as all LDAP calls are logged in debug level.

## TODO

- [x] Decide on the correct naming for the option, currently `--debug` bug maybe this will cause conflicts too easily. **Changed to `--debug-log`**
- [ ] Decide whether to document the option in the usage output for classes based on `OC\Core\Command\Base`.
- [x] Determine whether the output format is the right one. (Do we want raw json instead?) **Abandoned this idea, there is not much data to put in json anyway**
- [x] Do we need to be able to do the same with an arbitrary loglevel, like --debug=1 to only show info and above? **Added support for `--debug-log-level` when `--debug-log` is used**
- [x] Should we activate the profiler when this option is used and output the profile? **Not automatically**
- [x] How can we make it output DB requests to the output as well? The profiler idea above would work. **Left as an exercise for later**

## Checklist

### DIFF
--- a/core/Listener/BeforeMessageLoggedEventListener.php
+++ b/core/Listener/BeforeMessageLoggedEventListener.php
@@ -1,0 +1,38 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC\Core\Listener;
+
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Log\BeforeMessageLoggedEvent;
+
+/**
+ * Listen to log calls and output them to STDOUT for debug purposes
+ * @template-implements IEventListener<BeforeMessageLoggedEvent>
+ */
+class BeforeMessageLoggedEventListener implements IEventListener {
+	public function handle(Event $event): void {
+		if (!$event instanceof BeforeMessageLoggedEvent) {
+			return;
+		}
+		echo
+			match($event->getLevel()) {
+				0 => '[debug]',
+				1 => '[info]',
+				2 => '[warning]',
+				3 => '[error]',
+				4 => '[fatal]',
+				default => '['.$event->getLevel().']',
+			}
+		.' ['.$event->getApp().'] '
+		.$event->getMessage()['message']
+		."\n";
+	}
+}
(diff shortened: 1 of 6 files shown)


come-nc:
Hum, I think this is a duplicate of https://github.com/nextcloud/logreader/pull/873 .
I’d prefer having this in core and as an option though, as setting an env var is not always the same syntax for customers, and `logreader` may be disabled. The point was to have a command we can pass to any user to get a detailed idea of what happens when they run a command.

core/Listener/BeforeMessageLoggedEventListener.php
```diff
@@ -0,0 +1,68 @@
...
+
+	/**
+	 * Register listener to log messages and remove debug options from $_SERVER['argv']
+	 */
+	public static function setup(): void {
+		$eventDispatcher = Server::get(IEventDispatcher::class);
+		$argv = $_SERVER['argv'];
+		$level = 0;
+		foreach ($argv as $key => $arg) {
+			if ($arg === '--debug-log') {
```
nickvergessen, line 57: Should we put them into OCP somewhere so app devs have a list/track of "reserved arguments"?

nickvergessen (approved):

lib/base.php
```diff
@@ -578,6 +579,10 @@ public static function init(): void {
 		self::$server = new \OC\Server(\OC::$WEBROOT, self::$config);
 		self::$server->boot();
 
+		if (self::$CLI && in_array('--debug-log', $_SERVER['argv'])) {
```
artonge, line 582: Isn't it possible?

```suggestion
		if (self::$CLI && in_array(ReservedOptions::DEBUG_LOG, $_SERVER['argv'])) {
```

artonge (approved):

lib/public/Console/ReservedOptions.php
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCP\Console;
+
+/**
+ * @since 30.0.0
+ */
+final class ReservedOptions {
```
nickvergessen, line 15: @come-nc Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

PR #46123, opened 2024-06-25
feat: Allow getting/setting the password hash of a user

Pytal:
## Summary

- Allows getting/setting the password hash of a given user

## Checklist

### DIFF
--- a/lib/private/User/User.php
+++ b/lib/private/User/User.php
@@ -25,6 +25,7 @@ use OCP\IUser;
 use OCP\IUserBackend;
 use OCP\Notification\IManager as INotificationManager;
 use OCP\User\Backend\IGetHomeBackend;
+use OCP\User\Backend\IPasswordHashBackend;
 use OCP\User\Backend\IProvideAvatarBackend;
 use OCP\User\Backend\IProvideEnabledStateBackend;
 use OCP\User\Backend\ISetDisplayNameBackend;
@@ -317,6 +318,20 @@ class User implements IUser {
 		} else {
 			return false;
 		}
+	}
+
+	public function getPasswordHash(): ?string {
+		if (!($this->backend instanceof IPasswordHashBackend)) {
+			return null;
+		}
+		return $this->backend->getPasswordHash($this->uid);
+	}
+
+	public function setPasswordHash(string $passwordHash): bool {
+		if (!($this->backend instanceof IPasswordHashBackend)) {
+			return false;
+		}
+		return $this->backend->setPasswordHash($this->uid, $passwordHash);
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


icewind1991:
This should probably include some info about the hashing scheme used in the getter and setter to prevent future issues.

come-nc (commented):
As Robin said, document the hash format in the phpdoc, otherwise looks good.

Also maybe in setPasswordHash check that the hash format is valid?

artonge (approved):

Pytal:
> Also maybe in setPasswordHash check that the hash format is valid?

Clean way needs https://github.com/nextcloud/server/pull/46186

lib/private/User/User.php
```diff
@@ -319,6 +320,25 @@ public function setPassword($password, $recoveryPassword = null) {
 		}
 	}
 
+	public function getPasswordHash(): ?string {
+		if (!($this->backend instanceof IPasswordHashBackend)) {
+			return null;
+		}
+		/** @var IPasswordHashBackend $backend */
+		$backend = $this->backend;
+		return $backend->getPasswordHash($this->uid);
```
come-nc, line 329:
```suggestion
		return $this->backend->getPasswordHash($this->uid);
```
With the test above psalm should already know the interface is implemented, no?

lib/private/User/User.php
```diff
@@ -319,6 +320,25 @@ public function setPassword($password, $recoveryPassword = null) {
...
+	}
+
+	public function setPasswordHash(string $passwordHash): bool {
+		if (!($this->backend instanceof IPasswordHashBackend)) {
+			return false;
+		}
+		/** @var IPasswordHashBackend $backend */
+		$backend = $this->backend;
+		$result = $backend->setPasswordHash($this->uid, $passwordHash);
+		return $result;
```
come-nc, line 339: Same thing

lib/public/IUser.php
```diff
@@ -76,6 +76,23 @@ public function delete();
 	 */
 	public function setPassword($password, $recoveryPassword = null);
 
+	/**
+	 * Get the password hash of the user
+	 *
+	 * @return ?string the password hash hashed by `\OCP\Security\IHasher::hash()`
+	 * @since 30.0.0
+	 */
+	public function getPasswordHash(): ?string;
```
nickvergessen, line 85: this will also break all community user backends...
So must be added to https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html

Pytal: https://github.com/nextcloud/documentation/pull/11988

come-nc (approved):

PR #46174, opened 2024-06-27
fix(IntegrityCheck): Ensure the check is run if no results are available

susnux:
## Summary

If there are no cached results the current implementation was also returning an empty array, but this was the same as when there was a successful run. So to distinguish this we return `null` if there are *no* results. In this case we need to rerun the integrity checker.

## Checklist

### DIFF
--- a/lib/private/IntegrityCheck/Checker.php
+++ b/lib/private/IntegrityCheck/Checker.php
@@ -373,7 +373,7 @@ class Checker {
 	 */
 	public function hasPassedCheck(): bool {
 		$results = $this->getResults();
-		if (empty($results)) {
+		if ($results !== null && empty($results)) {
 			return true;
 		}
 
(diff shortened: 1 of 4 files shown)


lib/private/IntegrityCheck/Checker.php
```diff
@@ -399,6 +404,7 @@ public function getResults(): array {
 	 * @param array $result
 	 */
 	private function storeResults(string $scope, array $result) {
+		/** @var array */
```
come-nc, line 407: How do you know it’s not null here?
Can it be solved with psalm-assert instead of doc typing?

susnux: Usage, but probably right I make it more obvious and declarative :)

icewind1991 (approved):

yemkareems (approved):

Issue #34012, opened 2022-09-10
[Bug]: NC 24.0.5 hash_hkdf(): Argument #2 ($key) cannot be empty in .../lib/private/Security/Crypto.php:97


j-ed:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on Github _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] Nextcloud Server **is** running on 64bit capable CPU, PHP and OS.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

After upgrading my server from Nextcloud v24.0.4 to v24.0.5 and also upgrading to PHP8 (I don't know if this is of importance), the Nextcloud cron job generates the following error output:

```
ValueError: hash_hkdf(): Argument #2 ($key) cannot be empty in .../lib/private/Security/Crypto.php:97
Stack trace:
#0 .../nextcloud/lib/private/Security/Crypto.php(97): hash_hkdf('sha512', '')
#1 .../nextcloud/lib/private/Security/IdentityProof/Manager.php(110): OC\Security\Crypto->encrypt('-----BEGIN PRIV...')
#2 .../nextcloud/lib/private/Security/IdentityProof/Manager.php(133): OC\Security\IdentityProof\Manager->generateKey('user-juergen')
#3 .../nextcloud/lib/private/Security/IdentityProof/Manager.php(146): OC\Security\IdentityProof\Manager->retrieveKey('user-juergen')
#4 .../nextcloud/lib/private/Security/IdentityProof/Signer.php(64): OC\Security\IdentityProof\Manager->getKey(Object(OC\User\User))
#5 .../nextcloud/apps/lookup_server_connector/lib/BackgroundJobs/RetryJob.php(150): OC\Security\IdentityProof\Signer->sign('lookupserver', Array, Object(OC\User\User))
#6 .../nextcloud/lib/public/BackgroundJob/Job.php(79): OCA\LookupServerConnector\BackgroundJobs\RetryJob->run(Array)
[...]

PR #31499, opened 2022-03-09
Add fallback routines for empty secret cases

juliusknorr:
Make sure to keep authentication working when an instance has been setup without a secret after adding the secret manually afterwards.

- Password auth to webdav still works
- App password auth to webdav still works
- Browser session keeps being active
- PublicKeyTokens get rotated if the fallback is hit so their private key is reencrypted with the secret now

Provides a possible migration path for https://github.com/nextcloud/server/pull/31492

## ToDo
- [ ] Add migration step to update the config if possible and set a secret before the upgrade
  - [ ] secret
  - [ ] paswordsalt

PR #31492, opened 2022-03-08
Require the secret config to be configured

CarlSchwan:
If it's not configured the instance will look like it is working but
various features will silently break (end to end encryption, setting
alternate email and probably more).

One issue is that changing the secret from empty to something will
break various other stuff (app token). I don't think there is a good way
to solve this issue other than breaking early instead of having to
handle a painful migration later on.

image

PR #23496, opened 2020-10-16
Fixes related key issue.

lynn-stephenson:
This should fix it, I implemented tests, too. Everything seems to work fine when I ran the tests. I decided to leave additional changes for a later pull request. I don't really see any sense in overhauling the encryption scheme to where it breaks a bunch of unit tests if I'm not going to just pull in [Halite](https://github.com/paragonie/halite), which would be far better than my implementation anyways.

Here you go. :smile:

PR #46176, opened 2024-06-27
fix(Security): Don't attempt empty secret fallback on >=v3 ciphertext

joshtrichards:
## Summary

The v3 ciphertext format (added in #23496) has never worked with an empty secret/password. However, the v3 implementation was added prior to the existence of the empty secret fallback code added in #31499. 

While looking into #34012 I noted that the only reason the `hash_hkdf()` PHP ValueError is occurring is because the fallback is being attempted not only anytime the initial decrypt attempt fails, but also against newer ciphertext versions (where the fallback isn't relevant).

This PR adjusts the fallback logic to only run against <=2 versioned or non-versioned (which is even older) ciphertext.

Note: This does *not* fix the underlying cause of #34012 (which is probably in at least most cases a configuration problem), but it does:

- clean-up the logic to avoid the PHP error
- avoids doing the fallback on newer ciphertext where it's pointless (and most ciphertext is probably v3 at this point)

References:

- #31499
- #31492 
- #23496 
- #20915 

## TODO

- [x] Adjust the tests (since I also change one of the Exception strings)
- [ ] Add a test (for the fallback; I don't believe there is one but need to double-check again)

## Checklist

### DIFF
--- a/lib/private/Security/Crypto.php
+++ b/lib/private/Security/Crypto.php
@@ -97,9 +97,25 @@ class Crypto implements ICrypto {
 			}
 			return $this->decryptWithoutSecret($authenticatedCiphertext, $password);
 		} catch (Exception $e) {
+			// If password is empty and the current secret didn't work, attempt with an empty secret as a fallback 
+			// for instances where the secret might not have been set for a time *IF* the ciphertext version 
+			// indicates an empty secret is a possibility. For example, it's pointless (and will fail hard if tried)
+			// to try the fallback with v3 versioned ciphertext since the keying material can't be empty with that version.
+			// SO: Only attempt the fallback on older versions (including non-versioned) ciphertexts
 			if ($password === '') {
-				// Retry with empty secret as a fallback for instances where the secret might not have been set by accident
-				return $this->decryptWithoutSecret($authenticatedCiphertext, '');
+				// Determine the crypto version, if any
+				$parts = explode('|', $authenticatedCiphertext);
+				$partCount = \count($parts);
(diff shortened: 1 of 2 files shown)


lib/private/Security/Crypto.php
```diff
@@ -97,9 +97,25 @@ public function decrypt(string $authenticatedCiphertext, string $password = ''):
...
+				$parts = explode('|', $authenticatedCiphertext);
+				$partCount = \count($parts);
+				if ($partCount < 3 || $partCount > 4) {
+					throw new Exception('Authenticated ciphertext could not be decoded (invalid format).');
+				}
+				if ($partCount === 4) { // only newer ciphertext has a version field
+					$version = $parts[3];
+				}
+				
+				if ((!empty($version) && $version <= '2') || empty($version)) { // only <3 versioned or old non-versioned ciphertext ever supported empty secrets
```
come-nc, line 116:
```suggestion
				if (empty($version) || $version <= '2') { // only <3 versioned or old non-versioned ciphertext ever supported empty secrets
```

lib/private/Security/Crypto.php
```diff
@@ -110,9 +126,6 @@ private function decryptWithoutSecret(string $authenticatedCiphertext, string $p
 
 		$parts = explode('|', $authenticatedCiphertext);
 		$partCount = \count($parts);
-		if ($partCount < 3 || $partCount > 4) {
-			throw new Exception('Authenticated ciphertext could not be decoded.');
-		}
```
come-nc, line 115: Why is this removed?
Should be added back.

PR #46184, opened 2024-06-27
fix: Authorization header can be an empty string

juliusknorr:
On our instance we've noticed that macOS dav clients are actually sending tons of requests. When looking at the network dump of what is happening it turned out that every request is sent twice, one time with a cookie which fails with a 401 and once with basic auth then.

I've been stepping through such a request now and it turned out that our check to see if a `Authorization` header is sent along was not working as it is always filled with an empty string from `$_SERVER['HTTP_AUTHORIZATION']`

Turns out that once fixing this check, macOS is able to properly reuse the cookies it sends along.

<details>
<summary>Some screenshots from debugging to easier follow the code that leads to this </summary>

Screenshot 2024-06-27 at 20 25 14
Screenshot 2024-06-27 at 20 25 41
Screenshot 2024-06-27 at 20 27 02

</details>


nickvergessen (approved):
🙈

bigcat88 (approved):

PR #46186, opened 2024-06-27
feat: Add utility method to validate an IHasher hash

Pytal:
## Summary

Adds a method for validating that a hashed password has been hashed by IHasher::hash()

This can be used when the original plain text password is not known, otherwise IHasher::verify() should be used

## Checklist

### DIFF
--- a/lib/public/Security/IHasher.php
+++ b/lib/public/Security/IHasher.php
@@ -47,4 +47,11 @@ interface IHasher {
 	 * @since 8.0.0
 	 */
 	public function verify(string $message, string $hash, &$newHash = null): bool ;
+
+	/**
+	 * Check if the prefixed hash is valid
+	 *
+	 * @since 30.0.0
+	 */
+	public function validate(string $prefixedHash): bool;
 }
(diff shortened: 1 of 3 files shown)


come-nc (approved):
A unit test would be nice

artonge (approved):

lib/public/Security/IHasher.php
```diff
@@ -47,4 +47,11 @@ public function hash(string $message): string;
 	 * @since 8.0.0
 	 */
 	public function verify(string $message, string $hash, &$newHash = null): bool ;
+
+	/**
+	 * Check if the prefixed hash is valid
+	 *
+	 * @since 30.0.0
+	 */
+	public function validate(string $prefixedHash): bool;
```
nickvergessen, line 56: @Pytal Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(all changes in lib/public/)

Pytal: https://github.com/nextcloud/documentation/pull/12129

PR #46192, opened 2024-06-28
 🏗️ chore(deps): Bump doctrine/dbal from 3.8.3 to 4.0.4

nickvergessen:
- Ref https://github.com/nextcloud/3rdparty/pull/1877

## 🔴 Platform requirements

- Removed support for Oracle 12c (12.2.0.1) and older
- Removed support for MariaDB 10.4.2 and older
- Removed support for MySQL 5.6 and older
- Removed active support for Postgres 9

## :rescue_worker_helmet: Need Help
- [ ] Oracle: lastInserId is no longer supported: https://github.com/nextcloud/server/pull/46304
- [ ] PostgreSQL: Last insert ID things need a migration on update: https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/how-to/postgresql-identity-migration.html

## ⚠️ Breaking API changes
- `Doctrine\DBAL\Platforms\SqlitePlatform` was renamed, use `\OCP\IDBConnection::getDatabaseProvider()` instead
- `Doctrine\DBAL\Platforms\PostgreSQL94Platform` was removed, use `\OCP\IDBConnection::getDatabaseProvider()` instead
- `\OCP\DB\IQueryBuilder::getQueryParts()` and `\OCP\DB\IQueryBuilder::getQueryPart()` were removed
- `\OCP\DB\IQueryBuilder::resetQueryParts()` and `\OCP\DB\IQueryBuilder::resetQueryPart()` can only reset `'where'|'having'|'orderBy'|'groupBy'`. For all other cases a new querybuilder has to be created.
- `Doctrine\DBAL\Types\Type::getName()` aka. `$column->getType()->getName()` was removed is removed, use newly added `\OCP\DB\Types::getType($column->getType())`
- `Doctrine\DBAL\Schema\Table::changeColumn()` was removed, use `Table::modifyColumn()` instead
- `Doctrine\DBAL\Schema\Table::hasPrimaryKey()` was removed, use `Table::getPrimaryKey()` instead
- `Doctrine\DBAL\Schema\Schema::getTableNames()` was removed, use `OCP\DB\ISchemaWrapper::getTableNames()` instead
- `Doctrine\DBAL\Connection::getSchemaManager()` was removed, use `Connection::createSchemaManager()` instead
- `Doctrine\DBAL\FetchMode` was removed, please use the dedicated fetch methods instead
- `\OCP\DB\IQueryBuilder::delete()` and `\OCP\DB\IQueryBuilder::update()` no longer support table alias
- `Doctrine\DBAL\Platforms\MySQL80Platform` requires the length of a VARCHAR column to be specified - 255 was added
- `OCP\DB\IPreparedStatement::bindParam()` is deprecated and calls `bindValue()` internally
- `OCP\Diagnostics\IQueryLogger` no longer extends `\Doctrine\DBAL\Logging\SQLLogger`
- `OCP\Diagnostics\IQuery` data structure is now typed and the structure of the data can have changed depending on the type of the query due to upstream changes
- `OCP\DB\QueryBuilder\IExpressionBuilder::orX()` and `OCP\DB\QueryBuilder\IExpressionBuilder::andX()` can no longer be called without arguments. Create an array outside and only create the composition once you know your entries.
- Affected row counts are limited to PHP_INT_MAX

## Checklist


susnux:
> Doctrine\DBAL\Platforms\SqlitePlatform was renamed to Doctrine\DBAL\Platforms\SQLitePlatform

Does is make sense to abstract this for Nextcloud? It always feels bad to directly use Doctrine in apps if I just need the information which DB I am using (corner case but sometimes needed).
Maye within `IDBConnection` so to check `getDatabasePlatform`?

nickvergessen:
> Does is make sense to abstract this for Nextcloud? It always feels bad to directly use Doctrine in apps if I just need the information which DB I am using (corner case but sometimes needed).

I was thinking the same, works for the high-level platform. Some apps also check for `PostgreSQL94Platform` (also removed, just didn't reach it yet), but seems a base `PostgreSQLPlatform` was added in 3.3.8 which we have since a longer time. I assume no one cares whether its 9.4 or now, but only whether it's postgres.

susnux:
> I assume no one cares whether its 9.4 or newer, but only whether it's postgres.

Yes exactly!

nickvergessen:
- Cool stuff was already added for 28 by Anna: https://github.com/nextcloud/server/blob/dae7c159f728a90ffa53247d6e033abdae5d2bd6/lib/public/IDBConnection.php#L340-L346

lib/private/DB/ConnectionAdapter.php
```diff
@@ -172,12 +171,21 @@ public function close(): void {
...
 
+	/**
+	 * @param mixed $input
+	 * @param int $type
+	 * @deprecated 30.0.0 Only strings are supported as database type in the end and the $type parameter is ignored going forward
+	 */
 	public function quote($input, $type = IQueryBuilder::PARAM_STR) {
-		return $this->inner->quote($input, $type);
+		if ($type !== IQueryBuilder::PARAM_STR) {
+			\OC::$server->getLogger()->debug('Parameter $type is no longer supported and the function only handles resulting database type string', ['exception' => new \InvalidArgumentException('$type parameter is no longer supported')]);
```
come-nc, line 181: Please avoid using `getLogger` it’s deprecated.
```suggestion
			\OCP\Server::get(LoggerInterface::class)->debug('Parameter $type is no longer supported and the function only handles resulting database type string', ['exception' => new \InvalidArgumentException('$type parameter is no longer supported')]);
```

come-nc: Also you use $type in single quotes so it will not be interpolated.

lib/private/DB/ConnectionAdapter.php
```diff
@@ -238,7 +246,7 @@ public function getDatabaseProvider(): string {
 			return IDBConnection::PLATFORM_ORACLE;
 		} elseif ($platform instanceof PostgreSQLPlatform) {
 			return IDBConnection::PLATFORM_POSTGRES;
-		} elseif ($platform instanceof SqlitePlatform) {
+		} elseif ($platform instanceof SQLitePlatform) {
```
come-nc, line 249: This method can now return `$this->inner->getDatabaseProvider` ?

lib/private/DB/PreparedStatement.php
```diff
@@ -57,16 +58,26 @@ public function fetchOne() {
...
+	public function bindValue($param, $value, $type = IQueryBuilder::PARAM_STR): bool {
+		$this->statement->bindValue($param, $value, $this->convertParameterTypeToDoctrine($type));
+		return true;
 	}
 
-	public function bindParam($param, &$variable, $type = ParameterType::STRING, $length = null): bool {
-		return $this->statement->bindParam($param, $variable, $type, $length);
+	public function bindParam($param, &$variable, $type = IQueryBuilder::PARAM_STR, $length = null): bool {
+		if ($type !== IQueryBuilder::PARAM_STR) {
+			\OC::$server->getLogger()->warning('PreparedStatement::bindParam() is no longer supported. Use bindValue() instead.', ['exception' => new \BadMethodCallException('bindParam() is no longer supported')]);
```
come-nc, line 68: :warning:

lib/private/DB/QueryBuilder/ExpressionBuilder/ExpressionBuilder.php
```diff
@@ -400,7 +400,10 @@ public function bitwiseOr($x, int $y): IQueryFunction {
 	 * @return ILiteral
 	 */
 	public function literal($input, $type = IQueryBuilder::PARAM_STR): ILiteral {
-		return new Literal($this->expressionBuilder->literal($input, $type));
+		if ($type !== IQueryBuilder::PARAM_STR) {
+			\OC::$server->getLogger()->debug('Parameter $type is no longer supported and the function only handles resulting database type string', ['exception' => new \InvalidArgumentException('$type parameter is no longer supported')]);
```
come-nc, line 404: :warning:

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -95,16 +109,16 @@ public function automaticTablePrefix($enabled) {
 	 * @return \OCP\DB\QueryBuilder\IExpressionBuilder
 	 */
 	public function expr() {
-		if ($this->connection->getDatabasePlatform() instanceof OraclePlatform) {
+		if ($this->connection->getDatabaseProvider() === IDBConnection::PLATFORM_ORACLE) {
```
come-nc, line 112: Use a match :bulb:

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -575,11 +598,16 @@ public function addSelect(...$selects) {
 	 * @param string $alias The table alias used in the constructed query.
 	 *
 	 * @return $this This QueryBuilder instance.
+	 * @since 30.0.0 Alias is no longer supported
 	 */
 	public function delete($delete = null, $alias = null) {
+		if ($alias !== null) {
+			$this->logger->debug('DELETE queries with alias are no longer supported and the provided alias is ignored', ['exception' => new \InvalidArgumentException('Table alias provided for DELETE query')]);
```
come-nc, line 605: warning rather than debug maybe?

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -1096,18 +1135,20 @@ public function addOrderBy($sort, $order = null) {
 	 * @param string $queryPartName
 	 *
 	 * @return mixed
```
come-nc, line 1137:
```suggestion
	 * @return never
```

lib/private/DB/ResultAdapter.php
```diff
@@ -30,14 +30,25 @@ public function closeCursor(): bool {
 	}
 
 	public function fetch(int $fetchMode = PDO::FETCH_ASSOC) {
-		return $this->inner->fetch($fetchMode);
+		if ($fetchMode === PDO::FETCH_ASSOC) {
```
come-nc, line 33: Use a match :firecracker:

come-nc: (seriously, there is no match emoji?)

lib/private/DB/SchemaWrapper.php
```diff
@@ -43,26 +44,27 @@ public function performDropTableCalls() {
...
 	/**
-	 * @return array
+	 * @return list<string>
 	 */
-	public function getTableNames() {
-		return $this->schema->getTableNames();
+	public function getTableNames(): array {
+		return array_map(function (Table $table) {
+			return $table->getName();
+		}, $this->schema->getTables());
```
come-nc, line 67:
```suggestion
		return array_map(fn (Table $table) => $table->getName(), $this->schema->getTables());
```

lib/private/DB/TDoctrineParameterTypeMap.php
```diff
@@ -0,0 +1,45 @@
...
+
+use Doctrine\DBAL\ArrayParameterType;
+use Doctrine\DBAL\ParameterType;
+use OCP\DB\QueryBuilder\IQueryBuilder;
+
+trait TDoctrineParameterTypeMap {
+	protected function convertParameterTypeToDoctrine(string|int|null $type): ArrayParameterType|ParameterType|string {
+		return match($type) {
+			IQueryBuilder::PARAM_DATE,
+			IQueryBuilder::PARAM_JSON => $type,
```
come-nc, line 19: How come these are not translated?

lib/private/Setup/PostgreSQL.php
```diff
@@ -35,7 +35,7 @@ public function setupDatabase($username) {
 					->andWhere($builder->expr()->eq('rolname', $builder->createNamedParameter($this->dbUser)));
 
 				try {
-					$result = $query->execute();
+					$result = $query->executeQuery();
 					$canCreateRoles = $result->rowCount() > 0;
 				} catch (DatabaseException $e) {
```
come-nc, line 40: iirc execute and executeQuery do not throw the same class and this needs to be adapted, same for other similar changes.

come-nc (commented):
Sorry for the bit of bikeshedding :slightly_smiling_face:

apps/federatedfilesharing/lib/Migration/Version1011Date20201120125158.php
```diff
@@ -32,7 +32,7 @@ public function changeSchema(IOutput $output, Closure $schemaClosure, array $opt
 		if ($schema->hasTable('federated_reshares')) {
 			$table = $schema->getTable('federated_reshares');
 			$remoteIdColumn = $table->getColumn('remote_id');
-			if ($remoteIdColumn && $remoteIdColumn->getType()->getName() !== Types::STRING) {
+			if (Types::getType($remoteIdColumn->getType()) !== Types::STRING) {
```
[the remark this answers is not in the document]
nickvergessen: Psalm complained that it's always truthy...

lib/private/DB/Connection.php
```diff
@@ -100,14 +103,11 @@ public function __construct(
...
-		/** @var \OCP\Profiler\IProfiler */
+		/** @var IProfiler */
 		$profiler = Server::get(IProfiler::class);
 		if ($profiler->isEnabled()) {
-			$this->dbDataCollector = new DbDataCollector($this);
+			$this->dbDataCollector = new DbDataCollector(Server::get(IQueryLogger::class));
 			$profiler->add($this->dbDataCollector);
-			$debugStack = new BacktraceDebugStack();
-			$this->dbDataCollector->setDebugStack($debugStack);
-			$this->_config->setSQLLogger($debugStack);
```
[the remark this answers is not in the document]
nickvergessen: Upstream removed it, it's now a new "Middleware" (see lib/private/DB/Logging)

lib/private/DB/ConnectionAdapter.php
nickvergessen: > Also you use $type in single quotes so it will not be interpolated.

The stacktrace should tell you what the value was. It was never meant to "interpolate"

lib/private/DB/QueryBuilder/QueryBuilder.php
nickvergessen: Not sure for now. I' wait until people read the changelogs and modified their apps. Also it's not a problem I think?

lib/private/DB/TDoctrineParameterTypeMap.php
nickvergessen: I have no clue, but they were also strings before while everything else was integer constants from doctrine

core/Migrations/Version13000Date20170718121200.php
```diff
@@ -238,7 +237,7 @@ public function changeSchema(IOutput $output, \Closure $schemaClosure, array $op
 			$table->addIndex(['name'], 'fs_name_hash');
 			$table->addIndex(['mtime'], 'fs_mtime');
 			$table->addIndex(['size'], 'fs_size');
-			if (!$schema->getDatabasePlatform() instanceof PostgreSQL94Platform) {
+			if ($this->connection->getDatabaseProvider() !== IDBConnection::PLATFORM_POSTGRES) {
```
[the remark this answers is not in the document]
nickvergessen: The only other option is `PostgreSQL100Platform` which does: `PostgreSQL100Platform extends PostgreSQL94Platform`
So this is an equal replacement.
In fact `PostgreSQLPlatform` only got added in 3.2 before "94" was the "root" class: https://github.com/doctrine/dbal/blob/3.1.0/src/Platforms/PostgreSQL94Platform.php#L40

lib/private/DB/Connection.php
```diff
@@ -174,26 +174,13 @@ public function getQueryBuilder(): IQueryBuilder {
 	 * @return \Doctrine\DBAL\Query\QueryBuilder
 	 * @deprecated please use $this->getQueryBuilder() instead
 	 */
-	public function createQueryBuilder() {
+	public function createQueryBuilder(): \Doctrine\DBAL\Query\QueryBuilder {
```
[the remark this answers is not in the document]
nickvergessen: There are quite some deprecated methods we can remove in a follow up. For now I'd like this to be as little breaking as possible to move it forward and get it merged quickly.

lib/private/DB/ConnectionAdapter.php
nickvergessen: Right, now that I added it there

lib/private/Setup/PostgreSQL.php
nickvergessen: https://github.com/nextcloud/server/blob/5fdc5d50114e130d157a142bd72bce3f8d0ac373/lib/public/DB/QueryBuilder/IQueryBuilder.php#L149-L154
🆒 how the docs are lying then...

come-nc: They are not, you do need to use executeQuery, but you also need to adapt catch to the wrapped Exception.
I checked and `DatabaseException` is not part of the wrapped ones, only the dbal exception get wrapped so you should be fine.

PR #46193, opened 2024-06-28
fix(carddav): limit vcard size

SebastianKrupinski:
* Resolves: Card size limit

### DIFF
--- a/apps/dav/appinfo/v1/carddav.php
+++ b/apps/dav/appinfo/v1/carddav.php
@@ -11,6 +11,7 @@ use OCA\DAV\AppInfo\PluginManager;
 use OCA\DAV\CardDAV\AddressBookRoot;
 use OCA\DAV\CardDAV\CardDavBackend;
 use OCA\DAV\CardDAV\Security\CardDavRateLimitingPlugin;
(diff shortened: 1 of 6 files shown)


tcitworld:
Ah, this reminds me I never added the repair steps I was talking about here: https://github.com/nextcloud/contacts/issues/2743

SebastianKrupinski:
> Ah, this reminds me I never added the repair steps I was talking about here: [nextcloud/contacts#2743](https://github.com/nextcloud/contacts/issues/2743)

@tcitworld Are you going to write the repair step? If you are I modified the PR to use a configurable app setting, please use the same setting for the repair step.

apps/dav/lib/CardDAV/CardDavBackend.php
```diff
@@ -59,6 +61,7 @@ public function __construct(
 		private IUserManager $userManager,
 		private IEventDispatcher $dispatcher,
 		private Sharing\Backend $sharingBackend,
+		private IConfig $config,
```
kesselb, line 64: IAppConfig should also work

hamza221 (approved):

st3iny (approved):
Looks good to me.

PR #46197, opened 2024-06-28
Use the same `setup-php` version on all tests

solracsf:
## Summary

- Use same version on all tests (no need to run some old versions on some tests)
- This versions ships a fallback mirror for ondrej/php PPA when launchpad is down wich can improve test reliability. See https://github.com/shivammathur/setup-php/releases/tag/2.31.0

## TODO

- [x] CI

## Checklist

### DIFF
--- a/.github/workflows/autocheckers.yml
+++ b/.github/workflows/autocheckers.yml
@@ -56,7 +56,7 @@ jobs:
           submodules: true
 
       - name: Set up php ${{ matrix.php-versions }}
-        uses: shivammathur/setup-php@7fdd3ece872ec7ec4c098ae5ab7637d5e0a96067 # v2
+        uses: shivammathur/setup-php@2e947f1f6932d141d076ca441d0e1e881775e95b #v2.31.0
         with:
           php-version: ${{ matrix.php-versions }}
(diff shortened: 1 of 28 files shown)


joshtrichards:
I wonder if it's worth either:

- using the `v2` tag (I get we may not want to let it float to even minor patch version changes to avoid test diffs; though I guess we already have been using inconsistent hashes since we're doing this all manually at the moment)
- go ahead and use a specific `2.x.y` version tag and integrate with dependabot

?

Refs:
- https://github.com/shivammathur/setup-php/tree/2.31.0?tab=readme-ov-file#bookmark-versioning
- https://docs.github.com/en/github/administering-a-repository/keeping-your-actions-up-to-date-with-github-dependabot

solracsf:
To read also: https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

joshtrichards (approved):
Makes sense

susnux (approved):

susnux:
Could you please use conventional commits, in this case probably:
`ci: Use the same setup-php version on all tests`

PR #46201, opened 2024-06-29
Update dashboard panels border radius

marcoambrosini:


ShGKme:
Changes dashboard widgets to be the same as main container in apps:

Before | After
---|---
image | image

apps/dashboard/src/DashboardApp.vue
```diff
@@ -506,7 +506,7 @@ export default {
 	background-color: var(--color-main-background-blur);
 	-webkit-backdrop-filter: var(--filter-background-blur);
 	backdrop-filter: var(--filter-background-blur);
-	border-radius: var(--border-radius-rounded);
+	border-radius: var(--body-container-radius);
```
ShGKme, line 509: Probably, we should also add this variable to the theming.

ShGKme (approved):
Maybe also change the padding to be in line with border radius?

apps/dashboard/src/DashboardApp.vue
marcoambrosini: Isn't this one already there?

susnux:
I wonder why this is not styled in the [NcDashboardWidget](https://nextcloud-vue-components.netlify.app/#/Components/NcDashboard?id=ncdashboardwidget) :thinking: 

But you then also need to adjust all "guest-box" components (login etc) which we unified with the dashboard widgets not long ago

marcoambrosini:
> I wonder why this is not styled in the [NcDashboardWidget](https://nextcloud-vue-components.netlify.app/#/Components/NcDashboard?id=ncdashboardwidget) 🤔

I also thought I would find it there 

> But you then also need to adjust all "guest-box" components (login etc)

Could you please point me to the components @susnux?

marcoambrosini:
@susnux should we tackle this in separate prs so that we get the important stuff in?

apps/dashboard/src/DashboardApp.vue
susnux: I think @ShGKme means adding it to theming / the theme from the theming app

susnux (approved):

PR #46231, opened 2024-07-01
 fix(Wipe): Better handle missing tokens 

joshtrichards:
## Summary

Check for non-empty token string to avoid 500/internal server errors in the Remote wipe API endpoints due to `Argument \#1 ($token) must be of type string, null given`.

Changed:

- Refactor slightly to streamline code
- Make `$token` nullable in both `checkWipe()` and `wipeDone()` then check for an empty value to avoid 500 error
- Expand test coverage

(arose after reproducing an issue initially reported in the community forum triggered by what seems to the iOS client sending us a null `token` sometimes, which is an issue for the client to address but better error handling in Server seemed a good idea too)

## Checklist

### DIFF
--- a/tests/Core/Controller/WipeControllerTest.php
+++ b/tests/Core/Controller/WipeControllerTest.php
@@ -55,6 +55,12 @@ class WipeControllerTest extends TestCase {
 			$this->remoteWipe->method('start')
 				->with('mytoken')
 				->willThrowException(new InvalidTokenException());
+			$this->remoteWipe->method('start')
+				->with('')
+				->willThrowException(new InvalidTokenException());
+			$this->remoteWipe->method('start')
+				->with(NULL)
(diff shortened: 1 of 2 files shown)


tests/Core/Controller/WipeControllerTest.php
```diff
@@ -55,6 +55,12 @@ public function testCheckWipe(bool $valid, bool $couldPerform, bool $result) {
 			$this->remoteWipe->method('start')
 				->with('mytoken')
 				->willThrowException(new InvalidTokenException());
+			$this->remoteWipe->method('start')
+				->with('')
+				->willThrowException(new InvalidTokenException());
+			$this->remoteWipe->method('start')
+				->with(NULL)
```
solracsf, line 62:
```suggestion
				->with(null)
```

Issue #46102, opened 2024-06-25
Hide featured apps from the list


the-djmaze:
I would like to hide featured apps from the apps list.
Because most of the time i want to ignore them, as someone else decided they should be "featured" and not the community as a whole.

So inside Nextcloud Server `/nextcloud/settings/apps/` it would be nice to hide the featured opinion of someone.

And also awesome if it can be done here: https://apps.nextcloud.com/featured

Or just drop the whole "featured" everywhere and prevent community annoyances due to one person his/her/them featured opinion.

jancborchardt:
Also for reference as said in https://github.com/nextcloud/server/pull/46232#pullrequestreview-2153303926

> "Featured" is a way to highlight not only apps developed by Nextcloud GmbH, but outstanding apps from the community like Forms, Maps, News and Tasks (and formerly Notes).
>
> Just like the new "Discover" page and the rating system it’s a nice way to explore good apps. Rather than dropping this entirely, we can update it. Do you have any suggestions @the-djmaze @solracsf? :)
>
> I also now added the very well rated Memories and Cookbook, as well as the previously missing Assistant.

We could have a more open curation process for featuring, or show more transparently why things were picked, this is just the first time it came up. Normally the reason is because they have great functionality, are well maintained, have very nice design, are nicely integrated and so on.

the-djmaze:
We could monitor "usage" but this defeats the whole privacy point.

So maybe use the community rating system better?
I came up with these ideas:
- [ ] "Featured apps" have 4+ star rating with at least 10 entries (or more to prevent 1x 5-star review boosting and such)
- [ ] Apps can be filtered based on 1-5 star rating
- [ ] Track active development (last release must be in previous N months)

jancborchardt:
> Review stars could also have an issue

Yeah, I’m not even quite sure how the weighing of the votes is calculated in the store. And again, it’s only one data point. Other important things are if it’s possibly useful for lots of people, and if it’s well-designed and well-maintained. And that has to be curated by an actual person or people.

the-djmaze:
> Other important things are if it’s possibly useful for lots of people, and if it’s well-designed and well-maintained. And that has to be curated by an actual person or people.

well-designed is an objective personal decision.
well-maintained is less objective.

Maybe have more detailed reviews where people can give stars for:
- usefullness
- design
- stabillity/code
- support

Kehino:
[...]

PR #46232, opened 2024-07-01
Drop "featured" apps

the-djmaze:
As i reported in #46102

> I would like to hide featured apps from the apps list.
Because most of the time i want to ignore them, as someone else decided they should be "featured" and not the community as a whole.

The l10n strings say:
> Featured apps are developed by and within the community. They offer central functionality and are ready for production use.

Yet there are way more production use apps developed by the community.
So this PR removes the conflicting "Featured".

### DIFF
--- a/apps/settings/lib/Controller/AppSettingsController.php
+++ b/apps/settings/lib/Controller/AppSettingsController.php
@@ -434,7 +434,7 @@ class AppSettingsController extends Controller {
 					$nextCloudVersionDependencies,
 					$phpDependencies
 				),
-				'level' => ($app['isFeatured'] === true) ? 200 : 100,
+				'level' => 100,
 				'missingMaxOwnCloudVersion' => false,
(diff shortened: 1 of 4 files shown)


solracsf:
I agree that Featured is not a good wording. But  I think there should be a way to identify which apps are developed and maintained by Nextcloud GmbH.

jancborchardt (changes requested):
"Featured" is a way to highlight not only apps developed by Nextcloud GmbH, but outstanding apps from the community like Forms, Maps, News and Tasks (and formerly Notes).

Just like the new "Discover" page and the rating system it’s a nice way to explore good apps. Rather than dropping this entirely, we can update it. Do you have any suggestions @the-djmaze @solracsf? :)

I also now added the very well rated Memories and Cookbook, as well as the previously missing Assistant.

the-djmaze:
Ok, we will keep the discussion and i close this.

PR #31216, opened 2022-02-16
only security updates for 21

blizzz:
disable patch versions update for stable21, letting through security ones only.

PR #46235, opened 2024-07-02
fix(dependabot): add back npm patch upgrades for stable branches

skjnldsv:
Seems like copy/pasting went wrong at some point :'(

oringaly from https://github.com/nextcloud/server/pull/31216 by @blizzz

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -116,7 +116,7 @@ updates:
   ignore:
     # ignore all GitHub linguist patch updates
     - dependency-name: "*"
-      update-types: ["version-update:semver-major", "version-update:semver-minor", "version-update:semver-patch"]
+      update-types: ["version-update:semver-major", "version-update:semver-minor"]
   # Disable automatic rebasing because without a build CI will likely fail anyway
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -136,7 +136,7 @@ updates:
   ignore:
     # ignore all GitHub linguist patch updates
```
susnux, line 137: This comment makes not really sense

artonge (approved):

.github/dependabot.yml
```diff
@@ -116,7 +116,7 @@ updates:
...
```
skjnldsv, line 117:
```suggestion
```

.github/dependabot.yml
skjnldsv:
```suggestion
```

susnux (approved):

blizzz:
> oringaly from #31216 by @blizzz

There the changes applied only to an EOL branch. Not against stable ones, as they are now. Blind copy and paste, or blind overwriting of branch names?

PR #46270, opened 2024-07-03
Match input height with clickable area

marcoambrosini:
Brings it from hardcoded 36 px to current default (34px)


skjnldsv (approved):

Antreesy (approved):
Placeholder is now too close to the bottom, don't forget to adjsut it

image

PR #46271, opened 2024-07-03
Update dashboard icon

AndyScherzinger:
* Resolves: none

## Summary

replace icon with the exact sizes of other icons, like activity, so the original viewport size matches height/width, to be rendered more crisp

## Checklist


susnux:
@AndyScherzinger where did you take the icon from? It seems to be not the Material Symbol with `grade: 0`, `optical size: 20` `weight: 400`.

Especially optical size of 20px is very important as the icon will in 99% of all places be rendered 20px.

So for consistent style we need to use the 20px optical size version everywhere.

susnux:
Moreover this again breaks theming as you have removed the `fill="#fff"`

AndyScherzinger:
@susnux I took the icon from Material Symbols. Can't say for sure which size though.
I can just do another PR on Monday with the optical size 20 yet when I checked other app icons their svg size was 32,no matter the optical size.
Can't say if that makes a difference here.

Also with the other apps I might even say the files app icon is the one smaller than the others?

susnux:
> Also with the other apps I might even say the files app icon is the one smaller than the others?

Thats a bug. In the menu all icons should be 20px but a styling update broke the selector so that now the original size is shown.

AndyScherzinger:
Ah, okay. Now I got it. Thanks for explaining it again. So what is the way forward? Style gets fixed or also the icon needs to be changed in size within the svg?

susnux:
Well I would say if you selected this style of the icon for all other apps now, then lets stick with it.
I will see if I find some time fixing styles, but lets keep it consistent :)

AndyScherzinger:
Well, the actual size might have been slightly to big. So I adjusted it to match the others, i.e. the photos app icon  #46360

PR #46286, opened 2024-07-03
fix(TaskProcessing): Set task status to running when processing via ISynchronousProvider

marcelklehr:
## Summary
When processing an ISynchronousProvider in a background job the task would not be set to running status. This PR fixes that.

## Checklist

### DIFF
--- a/lib/private/TaskProcessing/SynchronousBackgroundJob.php
+++ b/lib/private/TaskProcessing/SynchronousBackgroundJob.php
@@ -18,6 +18,7 @@ use OCP\TaskProcessing\Exception\ProcessingException;
 use OCP\TaskProcessing\Exception\ValidationException;
 use OCP\TaskProcessing\IManager;
 use OCP\TaskProcessing\ISynchronousProvider;
+use OCP\TaskProcessing\Task;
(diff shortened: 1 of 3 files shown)


julien-nc (approved):
Yay

bigcat88 (approved):

skjnldsv:
Labels please :)

lib/private/TaskProcessing/SynchronousBackgroundJob.php
```diff
@@ -61,6 +62,7 @@ protected function run($argument) {
 					return;
 				}
 				try {
+					$this->taskProcessingManager->setTaskStatus($task, Task::STATUS_RUNNING);
```
kyteinsky, line 65: errors and successes should be reported too, no? Or is that auto-inferred?

marcelklehr: for that we call setTaskResult further below

kyteinsky: cool

kyteinsky (approved):

PR #46295, opened 2024-07-04
chore(caldav): remove obsolete function

SebastianKrupinski:
Found a function that is not used anywhere. Full texted searched the server and calendar repo's code, there is only one reference to it which is the function declaration


ChristophWurst (approved):
:saluting_face: :coffin:

st3iny (approved):
:bomb:

PR #46299, opened 2024-07-04
[stable23] fix(CI): Copy smb-kerberos workflow from 24

nickvergessen:
Now I'm facing a problem with icewind1991/samba-krb-test-apache It comes with PHP 8.1 which is not supported on those old versions:
https://github.com/nextcloud/server/actions/runs/9789645621/job/27029727477?pr=46266
but there is no "tag" with an older PHP version.

Would it be possible to revert https://github.com/icewind1991/samba-krb-test/commit/4e8750b56b6737c5474bedec27b9b147aead0e7d create a new image and tag it with 7.4 and publish that image?


SystemKeeper (approved):

blizzz:
@nickvergessen dev or close

nickvergessen:
It would be helpful to have green CI on backports, but if the expert is not available there is not a lot I can do.

PR #46320, opened 2024-07-05
perf: Set session.cache_limiter at runtime to avoid clients caching static assets served by PHP

juliusknorr:
By default there is a Pragma: no-cache header set due to the default
value `no-cache` of session.cache-limiter, which will cause Chrome and
iOS to not cache even with a different Cache-Control header set on the
response.

Signed-off-by: Julius Härtl <jus@bitgrid.net>

* Resolves: # 

## TODO

- [ ] ...

## Checklist


nickvergessen (approved):

SystemKeeper (approved):
Tested and works 👍

PR #46324, opened 2024-07-05
Fix/update share entry quick select

nfebe:
`SharingEntryQuickSelect` should respond to changes and reflect the new state of shares when updates are done in the `SharingDetailsTab`


susnux:
replaced with new unified sharing #62593

PR #46352, opened 2024-07-08
feat: add occ command files:list

yemkareems:
Duplicate pr of https://github.com/nextcloud/server/pull/43342 since CI CD issues are there in the original PR
Possible  usage  example with params
occ files:list admin --path abc --sort name --order ASC --minSize 100 --maxSize 1008400 --type application

Sample usage

image
occ files:list admin

if path is not mentioned list all files in the root directory of the user

image

occ files:list admin --path abc

if path is  mentioned list all files in the path mentioned alone

image

occ files:list admin --sort name --order ASC

image

occ files:list alice --path Media --sort size --order DESC

image

occ files:list admin --type image

image

occ files:list admin --type application --minSize 2750 --maxSize 148415

* Resolves: # 

## TODO

- [ ] ...

## Checklist


apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,297 @@
...
+			->addOption(
+				"sort",
+				"name",
+				InputArgument::OPTIONAL,
+				"Sort by name, path, size, owner, type, perm, created-at"
+			)
+			->addOption("order", "ASC", InputArgument::OPTIONAL, "Order is either ASC or DESC");
+	}
+
+	private function getNodeInfo(File|Folder $node): array {
```
artonge, line 71:
```suggestion
	private function getNodeInfo(Node $node): array {
```

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,297 @@
...
+	): void {
+		try {
+			/** @var Folder $userFolder **/
+			$userFolder = $this->rootFolder->get($path);
+
+			$files = $userFolder->getDirectoryListing();
+			foreach ($files as $file) {
+				$includeType = $includeMin = $includeMax = true;
+				$nodeInfo = $this->getNodeInfo($file);
+				if ($type != "" && $type != $nodeInfo['type']) {
```
artonge, line 103:
```suggestion
				if ($type !== "" && $type !== $nodeInfo['type']) {
```

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,297 @@
+<?php
+/**
+ * SPDX-FileCopyrightText: 2017-2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-FileCopyrightText: 2016 ownCloud, Inc.
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+namespace OCA\Files\Command;
```
artonge, line 7:
```suggestion
declare(strict_types=1);

namespace OCA\Files\Command;
```

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,297 @@
...
+	): int {
+		$inputPath = $input->getArgument("path");
+		if ($inputPath) {
+			$inputPath = ltrim($inputPath, "path=");
+			[, $user] = explode("/", rtrim($inputPath, "/").'/', 4);
+		}
+
+		$this->initTools($output);
+
+		if (is_object($user)) {
```
artonge, line 157: I don't think `$user` could ever be an object at this point.

artonge (approved):

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,300 @@
...
+		parent::configure();
+
+		$this->setName("files:list")
+			->setDescription("List filesystem in the path mentioned in path argument")
+			->addArgument(
+				"path",
+				InputArgument::REQUIRED,
+				'List all the files and folder mentioned in this path, eg occ files:list path="/alice/files/Music", the path being a required argument to determine the user'
+			)
+			->addOption("type", "", InputArgument::OPTIONAL, "Filter by type like application, image, video etc")
```
come-nc, line 52: Is that mimetype?

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,300 @@
...
+				InputArgument::OPTIONAL,
+				"Sort by name, path, size, owner, type, perm, created-at"
+			)
+			->addOption("order", "ASC", InputArgument::OPTIONAL, "Order is either ASC or DESC");
+	}
+
+	private function getNodeInfo(Node $node): array {
+		$nodeInfo = [
+			"name" => $node->getName(),
+			"size" => $node->getSize() . " bytes",
```
come-nc, line 77:
```suggestion
			"size" => \OCP\Util::humanFileSize($node->getSize()),
```
Should look better

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,300 @@
...
+		}
+	}
+
+	protected function execute(
+		InputInterface $input,
+		OutputInterface $output
+	): int {
+		$inputPath = $input->getArgument("path");
+		$user = '';
+		if ($inputPath) {
```
come-nc, line 156:
```suggestion
```
$inputPath is required, no?

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,300 @@
...
+				$item["owner"],
+				$item["created-at"],
+				$item["name"],
+				$item["type"],
+			];
+		}
+
+		$table = new Table($output);
+		$table->setHeaders($headers)->setRows($rows);
+		$table->render();
```
come-nc, line 298: `writeTableInOutputFormat` should be used.

come-nc (commented):
I’m not sure I understand the reasoning behind the path handling.

Would it not be better to have a user argument and a path argument inside this user folder?
Here it forces a path starting with a username, that feels unexpected. Did you try it with groupfolders and such?

apps/files/lib/Command/ListFiles.php
yemkareems: getMimePart of the node, for example it could be text, image, application and for folder it is directory

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,300 @@
...
+		string $user,
+		string $path,
+		OutputInterface $output,
+		?string $type = "",
+		?int $minSize = 0,
+		?int $maxSize = 0
+	): void {
+		try {
+			/** @var Folder $userFolder **/
+			$userFolder = $this->rootFolder->get($path);
```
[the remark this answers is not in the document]
yemkareems: getUserFolder works with a user and hence for path i used get

come-nc:
> > I’m not sure I understand the reasoning behind the path handling.
> > Would it not be better to have a user argument and a path argument inside this user folder? Here it forces a path starting with a username, that feels unexpected. Did you try it with groupfolders and such?
> 
> @come-nc this was implemented similar to files:scan which had both user_id and path. Since user was extracted from the path it was omitted and path was made a required argument

This is not the same thing, `files:scan` scans the disk folders so it’s limited to paths on the disk, but I expect `files:list` is expected to list files as in the UI, no? Otherwise people can use `ls`.

After discussing with the team:
I think we should go with having two arguments, userid and path.
Then you call getUserFolder($userId) on the root folder, and you can use get($path) on the userfolder instance.

apps/files/lib/Command/ListFiles.php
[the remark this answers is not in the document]
yemkareems: Replaced get with getUserFolder and path is got using get from the getUserFolder instance.

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,293 @@
...
+			$output->writeln("<error>" . $e->getTraceAsString() . "</error>");
+		}
+	}
+
+	protected function execute(
+		InputInterface $input,
+		OutputInterface $output
+	): int {
+		$inputPath = $input->getArgument("path");
+		$inputPath = ltrim($inputPath, "path=");
```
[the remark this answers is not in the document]
yemkareems: user is made a mandatory argument and path is made a optional argument. user is used to getUserFolder and path is used to get the path from the user folder

yemkareems:
> path



> > > I’m not sure I understand the reasoning behind the path handling.
> > > Would it not be better to have a user argument and a path argument inside this user folder? Here it forces a path starting with a username, that feels unexpected. Did you try it with groupfolders and such?
> > 
> > 
> > @come-nc this was implemented similar to files:scan which had both user_id and path. Since user was extracted from the path it was omitted and path was made a required argument
> 
> This is not the same thing, `files:scan` scans the disk folders so it’s limited to paths on the disk, but I expect `files:list` is expected to list files as in the UI, no? Otherwise people can use `ls`.
> 
> After discussing with the team: I think we should go with having two arguments, userid and path. Then you call getUserFolder($userId) on the root folder, and you can use get($path) on the userfolder instance.

Did it as suggested. Now user is a mandatory argument and path may or may not be present. Used getUserFolder in place of get and then get is used on top of getUserFolder instance

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,289 @@
...
+					}
+				} elseif ($file instanceof Folder) {
+					if ($includeType && $includeMin && $includeMax) {
+						$this->dirInfo[] = $nodeInfo;
+					}
+				}
+			}
+		} catch (ForbiddenException $e) {
+			$output->writeln(
+				"<error>Home storage for user $user not writable or 'files' subdirectory missing</error>"
```
come-nc, line 131:
```suggestion
				"<error>Home storage for user $user not readable or 'files' subdirectory missing</error>"
```

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,289 @@
...
+			$nodeInfo['type'] = 'directory';
+		}
+
+		return $nodeInfo;
+	}
+
+	protected function listFiles(
+		string $user,
+		OutputInterface $output,
+		?string $path = "",
```
come-nc, line 94:
```suggestion
		string $path = "",
```

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,289 @@
...
+		} catch (NotFoundException $e) {
+			$output->writeln(
+				"<error>Path not found: " . $e->getMessage() . "</error>"
+			);
+		} catch (\Exception $e) {
+			$output->writeln(
+				"<error>Exception during list: " . $e->getMessage() . "</error>"
+			);
+			$output->writeln("<error>" . $e->getTraceAsString() . "</error>");
+		}
```
come-nc, line 149:
```suggestion
		}
```
You can simply let unexpected exceptions bubble up, parent class takes care of showing the trace depending on verbosity if I recall correctly.

apps/files/lib/Command/ListFiles.php
```diff
@@ -0,0 +1,289 @@
...
+		$e = new \ErrorException($message, 0, $severity, $file, $line);
+		$output->writeln(
+			"<error>Error during list: " . $e->getMessage() . "</error>"
+		);
+		$output->writeln(
+			"<error>" . $e->getTraceAsString() . "</error>",
+			OutputInterface::VERBOSITY_VERY_VERBOSE
+		);
+		return true;
+	}
```
come-nc, line 235: Why all this?
This is the only command that do that to my knowledge.
Nextcloud already has an exception handler which logs them I think.

I suggest to remove this.

PR #46354, opened 2024-07-08
stop file actions when conflict dialog skipped

### LINKED ISSUE #45548: [Bug]: Duplicate File/folder names delete file when moving

### Bug description

NC Versions: 28,29,30
In 27 the skip Files dialog is not present

Files/Folders get deleted if decided to be skipped because the name already exists.
This bug existed in 28.0.3 also for drag and drop move.

### Steps to reproduce

1. create a folder that has a subfolder called ```test```
2. create folder also called ```test``` somewhere else
3. move the second ```test``` folder into the first folder that contains the first ```test``` folder
4. ⚠️ don't use drag and drop, use the three dots menu and select move or copy!
5. When asked how to resolve the name issue: select skip this File

The moved Folder disappears from the files list and is also not in the target directory. It can be restored from deleted files

### Expected behavior

The File should remain untouched and only appear where it appeared before

grnd-alt:
* Resolves: #45548 
## Summary
relies on: https://github.com/nextcloud-libraries/nextcloud-upload/pull/1270

## Checklist


susnux (dismissed):
Single node action, so if skipped return. Makes sense

skjnldsv:
I see the point, but looking a the PR that introduced that issue: #43667, 

> If the conflict dialog returns two empty arrays, selected and renamed, that means the user chose to keep the file already in the directory path, thereby choosing to delete the current selected file they want to move


I actually don't understand this anymore.
Is this really the outcome we want, if someone didn't pick the old NOR the new, that means we want to delete the source file? I'm far from certain :thinking:

grnd-alt:
> I see the point, but looking a the PR that introduced that issue: #43667,
> 
> > If the conflict dialog returns two empty arrays, selected and renamed, that means the user chose to keep the file already in the directory path, thereby choosing to delete the current selected file they want to move
> 
> I actually don't understand this anymore. Is this really the outcome we want, if someone didn't pick the old NOR the new, that means we want to delete the source file? I'm far from certain 🤔

I've wondered about that behavior as well. Having both arrays (selected and renamed) empty actually does not mean no file has been selected but that only the existing file has been selected (the one on the right). Without this PR, skipping files caused the same result as selecting only existing files, now we have different cases for each. But i don't know if it's expected behavior to delete files when it's selected to keep only the existing, or if that should mean the same as skipping(keeping both)

susnux:
From my understanding

```
N = Selected new ones
O = Selected old ones

Renamed = (n e N) in O
Selected = (n e N) not in O
```

So if you have `upload` (files to upload) and `existing` (all files that are already existing) and resolve using the conflict picker, then you have:
* `upload`
* `existing`
* `renamed` (from the submit event)
* `selected` (from the submit event)

So now we have two simple tasks
1. Handle `selected`
    1. delete from `existing` if it is in `selected`
    2. upload (or move) all `selected`
2. upload all `renamed` with their new name (or move with new name)

But now we also have all files not in `renamed` and not in `selected`, meaning those files where we selected to keep the old one.
* For upload this is basically a "skip uploading file"
* For moving **I** think it should also just cancel (skip) that file, this is how the file managers on Linux / Windows work.

So what do two empty arrays mean?
All old files where selected -> I would say we should simply skip the moving in those cases.

susnux:
TLDR:

Files not in `renamed` and not in`selected` are marked for "use the old one" -> File managers on Linux / Windows will just cancel the move and (my opinion) we should also just cancel the move operation for those files.

susnux (approved):
Was also the [suggested solution](https://cloud.nextcloud.com/call/gqff69i8#message_2982024) by designers.

skjnldsv (approved):

PR #46359, opened 2024-07-08
[TaskProcessing] Add start, stop and schedule time to tasks

julien-nc:
* [ ] Check the values are set everywhere it makes sense during the task lifecycle
* [x] Decide how to use those values. Should there be an occ command to generate a report? An API endpoint?
    * new `taskprocessing:task:list` occ command
    * new `taskprocessing:task:stats` occ command

Both commands accept those filters:
* userId
* task type
* custom ID
* status
* scheduled after
* ended before


marcelklehr:
>  Should there be an occ command to generate a report?

Can we hook into the system report command? If not, I'd vote for a separate command

julien-nc:
Let's make occ commands. Updates are in the PR initial comment.

core/Command/TaskProcessing/Statistics.php
```diff
@@ -0,0 +1,145 @@
...
+			$averageQueuingTime = (int)($totalQueuingTime / $queuingTimeCount);
+			$stats['Average queuing time'] = $averageQueuingTime;
+			$stats['Queuing time count'] = $queuingTimeCount;
+		}
+		if ($userWaitingTimeCount > 0) {
+			$stats['Max user waiting time'] = $maxUserWaitingTime;
+			$averageUserWaitingTime = (int)($totalUserWaitingTime / $userWaitingTimeCount);
+			$stats['Average user waiting time'] = $averageUserWaitingTime;
+			$stats['User waiting time count'] = $userWaitingTimeCount;
+		}
```
marcelklehr, line 140: Mmh, using SQL directly may be faster and a bit more idomatic, perhaps. Not sure if it's possible, though.

marcelklehr:
In our brainstorm we also had the idea to maybe record the I/O size. E.g. we could record the byte size of the JSON encoded input, as well as the output. What do you think?

julien-nc:
> In our brainstorm we also had the idea to maybe record the I/O size. E.g. we could record the byte size of the JSON encoded input, as well as the output. What do you think?

Fine by me. So we can take the size of the JSON but we also need the size of the input/output files, right?
It can then become very expensive to compute.

marcelklehr:
Perhaps let's leave the files out for now?

marcelklehr:
> Both commands accept those filters:

Perhaps also an app filter?

marcelklehr (approved):

lib/private/TaskProcessing/Db/TaskMapper.php
```diff
@@ -136,6 +136,51 @@ public function findUserTasksByApp(?string $userId, string $appId, ?string $cust
...
+		}
+		if ($scheduleAfter !== null) {
+			$qb->andWhere($qb->expr()->isNotNull('scheduled_at'));
+			$qb->andWhere($qb->expr()->gt('scheduled_at', $qb->createPositionalParameter($scheduleAfter, IQueryBuilder::PARAM_INT)));
+		}
+		if ($endedBefore !== null) {
+			$qb->andWhere($qb->expr()->isNotNull('ended_at'));
+			$qb->andWhere($qb->expr()->lt('ended_at', $qb->createPositionalParameter($endedBefore, IQueryBuilder::PARAM_INT)));
+		}
+		return array_values($this->findEntities($qb));
```
marcelklehr, line 181: Should we have db indeces for these?

julien-nc: I think since the amount of task is limited by the cleanup bg job, it's fine. Feel free to add indices in this branch if you think it's necessary.

bigcat88 (approved):
great changes, you will be able to see statistics on average how long each task was executed. nice!

PR #46368, opened 2024-07-09
TaskProcessing follow-up

marcelklehr:
## Summary
Pick up and solve lose ends in TaskProcessing

## Checklist

### DIFF
--- a/core/Controller/TaskProcessingApiController.php
+++ b/core/Controller/TaskProcessingApiController.php
@@ -309,11 +309,11 @@ class TaskProcessingApiController extends \OCP\AppFramework\OCSController {
 			if (!isset($file['tmp_name'])) {
 				return new DataResponse(['message' => $this->l->t('Bad request')], Http::STATUS_BAD_REQUEST);
 			}
-			$data = file_get_contents($file['tmp_name']);
-			if (!$data) {
+			$handle = fopen($file['tmp_name'], 'r');
+			if (!$handle) {
 				return new DataResponse(['message' => $this->l->t('Internal error')], Http::STATUS_INTERNAL_SERVER_ERROR);
 			}
-			$fileId = $this->setFileContentsInternal($data);
+			$fileId = $this->setFileContentsInternal($handle);
 			return new DataResponse(['fileId' => $fileId], Http::STATUS_CREATED);
 		} catch (NotFoundException) {
 			return new DataResponse(['message' => $this->l->t('Not found')], Http::STATUS_NOT_FOUND);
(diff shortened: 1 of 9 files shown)


core/Controller/TaskProcessingApiController.php
```diff
@@ -286,6 +289,39 @@ public function getFileContentsExApp(int $taskId, int $fileId): Http\DataDownloa
...
+	 * @return DataDownloadResponse<Http::STATUS_CREATED, array{fileId: int}, array{}>|DataResponse<Http::STATUS_INTERNAL_SERVER_ERROR|Http::STATUS_NOT_FOUND, array{message: string}, array{}>
+	 *
+	 *  201: File created
+	 *  404: Task not found
+	 */
+	#[ExAppRequired]
+	#[ApiRoute(verb: 'POST', url: '/tasks_provider/{taskId}/file', root: '/taskprocessing')]
+	public function setFileContentsExApp(int $taskId): DataResponse {
+		try {
+			$task = $this->taskProcessingManager->getTask($taskId);
```
bigcat88, line 307: why we need this line?

marcelklehr: This ensures that the task exists. I thought it would make sense to force the use of an existing task ID. We can also remove the taskId param altogether.

bigcat88: yes, you are right, then this check is needed

julien-nc (commented):
LGTM

core/Controller/TaskProcessingApiController.php
```diff
@@ -493,4 +529,15 @@ public function getNextScheduledTask(array $providerIds, array $taskTypeIds): Da
...
+
+	private function setFileContentsInternal(Task $task, string $data) {
+		try {
+			$folder = $this->appData->getFolder('TaskProcessing');
+		} catch (\OCP\Files\NotFoundException) {
+			$folder = $this->appData->newFolder('TaskProcessing');
+		}
+		/** @var SimpleFile $file */
+		$file = $folder->newFile((string) rand(0, 10000000), $data);
+		return $file->getId();
```
provokateurin, line 541: What happens if the file already exists? Will the file be overwritten or will it fail?

bigcat88: from the provided code it is obvious: new file will be created

marcelklehr:
```suggestion
		/** @var SimpleFile $file */
		$file = $folder->newFile(time() . '-' . rand(0, 10000000), $data);
		return $file->getId();
```

May be better to use current time for the file name as well.

bigcat88 (approved):
LGTM

julien-nc (approved):

PR #46371, opened 2024-07-09
feat: Add `IFilenameValidator` to have one consistent place for filename validation

susnux:
* Part of #44963 

## Summary

This is the first part, split https://github.com/nextcloud/server/pull/45151 to make review easier.

This adds a new interface `IFilenameValidator` that allows validating filenames to have one consistent location for filename handling.
We need two¹ functions:
1. Validation with meaningful messages (as errors)
2. Validation for `if` with boolean return

¹ We need in some rare private filesystem locations a third function: We need to check if a file is forbidden but can not validate all other restrictions, because we use our filesystem API also for setup where we use possibly forbidden characters as we are only setup Nextcloud itself not files later synced.

## Checklist


lib/private/Files/FilenameValidator.php
```diff
@@ -0,0 +1,249 @@
...
+	public function getForbiddenCharacters(): array {
+		if (empty($this->forbiddenCharacters)) {
+			// Get always forbidden characters
+			$forbiddenCharacters = str_split(\OCP\Constants::FILENAME_INVALID_CHARS);
+			if ($forbiddenCharacters === false) {
+				$forbiddenCharacters = [];
+			}
+
+			// Get admin defined invalid characters
+			$additionalChars = $this->config->getSystemValue('forbidden_chars', []);
```
come-nc, line 126:
```suggestion
			$additionalChars = $this->config->getSystemValue('forbidden_filename_characters', []);
```
You used the legacy key twice I think?

lib/private/Files/FilenameValidator.php
```diff
@@ -0,0 +1,249 @@
...
+			throw new FileNameTooLongException();
+		}
+
+		if ($this->isForbidden($filename)) {
+			throw new ReservedWordException();
+		}
+
+		$this->checkForbiddenExtension($filename);
+
+		$this->checkForbiddenCharacters($filename);
```
come-nc, line 188: Shouldn’t all these tests be made on the trimmed var instead?

lib/private/Files/FilenameValidator.php
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	/**
+	 * Check if the filename is forbidden
+	 * @param string $filename
```
come-nc, line 193: Can be a path if I understand correctly, as basename is called before anything else.

susnux:
Follow up with usage here: https://github.com/nextcloud/server/pull/46375

lib/private/Files/FilenameValidator.php
susnux: No having whitespace in the name is pretty valid, so rules might apply to those whitespace names.
The trim is only required for the empty check

lib/private/Files/FilenameValidator.php
susnux: Yes sadly required for some function calls in very internal filesystem logic :pensive: 
I will rename it

susnux: done

lib/private/Files/FilenameValidator.php
```diff
@@ -0,0 +1,249 @@
...
+			// Get always forbidden characters
+			$forbiddenCharacters = str_split(\OCP\Constants::FILENAME_INVALID_CHARS);
+			if ($forbiddenCharacters === false) {
+				$forbiddenCharacters = [];
+			}
+
+			// Get admin defined invalid characters
+			$additionalChars = $this->config->getSystemValue('forbidden_filename_characters', []);
+			if (!is_array($additionalChars)) {
+				$this->logger->error('Invalid system config value for "forbidden_chars" is ignored.');
```
come-nc, line 128:
```suggestion
				$this->logger->error('Invalid system config value for "forbidden_filename_characters" is ignored.');
```

come-nc (commented):
Almost there

lib/private/Files/FilenameValidator.php
susnux: :see_no_evil:

come-nc (approved):

sorbaugh (approved):

PR #46307, opened 2024-07-04
feat(files): Refresh grid view layout

artonge:
- Fix https://github.com/nextcloud/server/issues/45438

## Mockups

341374888-2f82d769-475b-4717-9580-d93a9a29cac2

## Current state

Before | Screenshots
-- | --
Overview | Screenshot from 2024-07-04 16-43-38
Focus states | Screenshot from 2024-07-04 16-43-59 Screenshot from 2024-07-04 16-44-01 Screenshot from 2024-07-04 16-44-03

PR #46373, opened 2024-07-09
fix(files): Add rounded corners back in grid view

artonge:
Was removed intentionally due to mockups.

- Follow-up of https://github.com/nextcloud/server/pull/46307

Before | After
-- | --
image | image


szaimen (approved):
LGTM

jancborchardt (changes requested):
`var(--border-radius)` for the image preview itself looks good, but the outside element (the hover/focus background) should have a bigger radius to match, so best use `var(--border-radius-large)` for that.

This is the same how it is in the action menus for example.

artonge:
Updated

jancborchardt (approved):
:sparkles: :blue_square:

susnux (approved):

PR #46375, opened 2024-07-09
refactor: Use `IFilenameValidator` for all filename validation and remove other validation code

susnux:
* [x] Chained on #46371 
* For #44963 

## Summary

This will migrate all filename validation from custom code to `IFilenameValidator`

## Checklist

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Node.php
+++ b/apps/dav/lib/Connector/Sabre/Node.php
@@ -122,12 +122,11 @@ abstract class Node implements \Sabre\DAV\INode {
 
 		[$parentPath,] = \Sabre\Uri\split($this->path);
 		[, $newName] = \Sabre\Uri\split($name);
-
-		// verify path of the target
-		$this->verifyPath();
-
 		$newPath = $parentPath . '/' . $newName;
 
+		// verify path of the target
+		$this->verifyPath($newPath);
+		
 		if (!$this->fileView->rename($this->path, $newPath)) {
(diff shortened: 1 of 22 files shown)


apps/dav/lib/Connector/Sabre/Node.php
```diff
@@ -122,12 +122,11 @@ public function setName($name) {
...
 		[$parentPath,] = \Sabre\Uri\split($this->path);
 		[, $newName] = \Sabre\Uri\split($name);
-
-		// verify path of the target
-		$this->verifyPath();
-
 		$newPath = $parentPath . '/' . $newName;
 
+		// verify path of the target
+		$this->verifyPath($newPath);
```
susnux, line 128: We wanted to verify the target but we were validating the source previously.
This breaks things if you have fileA but that name is now forbidden -> you need to be able to rename it.

(also it was simply wrong)

apps/dav/tests/unit/Connector/Sabre/DirectoryTest.php
```diff
@@ -397,7 +397,7 @@ public function testMoveFailedInvalidChars($source, $destination, $updatables, $
 
 	public function moveFailedInvalidCharsProvider() {
 		return [
-			['a/b', 'a/*', ['a' => true, 'a/b' => true, 'a/c*' => false], []],
+			['a/b', 'a/	', ['a' => true, 'a/b' => true, 'a/c	' => false], []],
```
susnux, line 400: legacy leftover from when we supported Windows Servers and `*` was forbidden char, nowadays it is not but so we use a really forbidden char (ascii < 31)

Issue #45978, opened 2024-06-19
Resolving references from public share


mejo-:
**Is your feature request related to a problem? Please describe.**

Currently references don't get resolved in public shares. The [reference resolve endpoint](https://github.com/nextcloud/server/blob/42fa0072046b3258062b50dbbd29110adefa1d71/core/Controller/ReferenceApiController.php#L65-L83) allows only authenticated requests. This results in a broken user experience when browsing public shares, e.g. when text documents or collectives pages contain link previews.

**Describe the solution you'd like**

The idea is to add new `PublicPage` endpoints to the `ReferenceApiController` that lookup ReferenceProviders which explicitly support public lookups. It's necessary to separate public from authenticated lookups as many providers look up data that is only accessible to authenticated users.

I'm not exactly sure yet how the implementation details could look like. One idea:
1. Add a new interface `IReferenceProviderPublic` with functions `matchReferencePublic()` and `resolveReferencePublic()`
2. In [`ReferenceApiController`](https://github.com/nextcloud/server/blob/42fa0072046b3258062b50dbbd29110adefa1d71/core/Controller/ReferenceApiController.php) define new endpoints like `GET /references/resolvePublic` that call `$this-referenceManager->resolveReference()` with `$public = true`.
3. In [`ReferenceManager`](https://github.com/nextcloud/server/blob/dae7c159f728a90ffa53247d6e033abdae5d2bd6/lib/private/Collaboration/Reference/ReferenceManager.php) only regard providers that implement `IReferenceProviderPublic` for public lookups (with `$public = true`).
4. The default [`LinkReferenceProvider`](https://github.com/nextcloud/server/blob/dae7c159f728a90ffa53247d6e033abdae5d2bd6/lib/public/Collaboration/Reference/LinkReferenceProvider.php) should just do the same for public and authenticated lookups as it only does opengraph anyway.
5. Custom reference providers (e.g. the Collectives one) can implement custom `resolveReferencePublic()` functions for a public share of a collective.

In the frontend, [`NcReferenceList`](https://github.com/nextcloud-libraries/nextcloud-vue/blob/0365599b008fb1807a5a4578af468a856a64f97f/src/components/NcRichText/NcReferenceList.vue) would call either the public or the authenticated reference API endpoints depending on whether it's an authenticated user session or not.

**Open questions**

[...]

PR #46378, opened 2024-07-09
feat(Reference): Add public API endpoints to get references

mejo-:
Calling the public API endpoints will check for matching registered reference providers that implement `IPublicReferenceProvider` and call their respective functions. If no matching provider is found, the default `LinkReferenceProvider` will be used to provide open graph data.

The frontend reference widget components will call these endpoints from unauthorized sessions, e.g. in public shares.

If present, the sharing token of the origin URL is passed to `resolveReferencePublic()` as additional information for the reference provider to determine the access scope. This allows the respective reference providers to determine whether the origin share has access to the linked resource.
    
`getCacheKeyPublic` also gets the sharing token so it can scope the cached entry to it.

Contributes to #45978

## Checklist

### DIFF
--- a/core/Controller/ReferenceApiController.php
+++ b/core/Controller/ReferenceApiController.php
@@ -10,6 +10,7 @@ namespace OC\Core\Controller;
 
 use OC\Core\ResponseDefinitions;
 use OCP\AppFramework\Http;
+use OCP\AppFramework\Http\Attribute\AnonRateLimit;
 use OCP\AppFramework\Http\Attribute\ApiRoute;
 use OCP\AppFramework\Http\DataResponse;
 use OCP\Collaboration\Reference\IDiscoverableReferenceProvider;
@@ -22,6 +23,8 @@ use OCP\IRequest;
  * @psalm-import-type CoreReferenceProvider from ResponseDefinitions
  */
 class ReferenceApiController extends \OCP\AppFramework\OCSController {
+	private const LIMIT_MAX = 15;
+
 	public function __construct(
 		string $appName,
 		IRequest $request,
@@ -63,6 +66,39 @@ class ReferenceApiController extends \OCP\AppFramework\OCSController {
 	}
 
 	/**
+	 * @PublicPage
+	 *
+	 * Extract references from a text
+	 *
+	 * @param string $text Text to extract from
+	 * @param string $sharingToken Token of the public share
+	 * @param bool $resolve Resolve the references
+	 * @param int $limit Maximum amount of references to extract, limited to 15
+	 * @return DataResponse<Http::STATUS_OK, array{references: array<string, CoreReference|null>}, array{}>
+	 *
+	 * 200: References returned
+	 */
+	#[ApiRoute(verb: 'POST', url: '/extractPublic', root: '/references')]
+	#[AnonRateLimit(limit: 10, period: 120)]
+	public function extractPublic(string $text, string $sharingToken, bool $resolve = false, int $limit = 1): DataResponse {
+		$references = $this->referenceManager->extractReferences($text);
+
+		$result = [];
+		$index = 0;
+		foreach ($references as $reference) {
+			if ($index++ >= min($limit, self::LIMIT_MAX)) {
+				break;
+			}
+
+			$result[$reference] = $resolve ? $this->referenceManager->resolveReference($reference, true, $sharingToken)?->jsonSerialize() : null;
+		}
+
+		return new DataResponse([
+			'references' => $result
+		]);
+	}
+
+	/**
 	 * @NoAdminRequired
 	 *
 	 * Resolve a reference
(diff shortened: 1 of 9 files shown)


core/Controller/ReferenceApiController.php
```diff
@@ -62,6 +63,38 @@ public function extract(string $text, bool $resolve = false, int $limit = 1): Da
...
+	 *
+	 * @param string $text Text to extract from
+	 * @param string $origin URL of the origin
+	 * @param bool $resolve Resolve the references
+	 * @param int $limit Maximum amount of references to extract
+	 * @return DataResponse<Http::STATUS_OK, array{references: array<string, CoreReference|null>}, array{}>
+	 *
+	 * 200: References returned
+	 */
+	#[ApiRoute(verb: 'POST', url: '/extractPublic', root: '/references')]
```
juliushaertl, line 79: We should have some rate limit on that as well

juliusknorr (commented):
Haven't tested yet but looks good from a first read

core/Controller/ReferenceApiController.php
```diff
@@ -82,6 +115,28 @@ public function resolveOne(string $reference): DataResponse {
...
+	 *
+	 * @param string $reference Reference to resolve
+	 * @param string $origin URL of the origin
+	 * @return DataResponse<Http::STATUS_OK, array{references: array<string, ?CoreReference>}, array{}>
+	 *
+	 * 200: Reference returned
+	 */
+	#[ApiRoute(verb: 'GET', url: '/resolvePublic', root: '/references')]
+	#[AnonRateLimit(limit: 5, period: 120)]
+	public function resolveOnePublic(string $reference, string $origin): DataResponse {
```
mejo-, line 131: I'm a bit unsure about passing the origin URL as additional information. While it should usually contain the share token, every reference provider has to implement their own logic to extract the share token from it then, and every reference provider would have to support different origin URL formats anyway, to support resolving target URLs within the same share but in another app.

E.g. when referencing a Collectives share link within the same share from Files, the Collectives reference provider would need to parse the origin URL for file sharing and collectives sharing link format.

Maybe a better approach is to expect the share token as parameter to the public API calls instead of `origin` and let the frontend extract the sharing token from `document`. In other words, `NcReferenceList` would do something like `const shareToken = document.getElementById('sharingToken').value` and all apps implementing public share pages would need to make sure this element exists.

mejo-: As discussed with @juliushaertl, implemented this now.

core/Controller/ReferenceApiController.php
mejo-: Indeed, fixed now.

juliusknorr (approved):

lib/private/Collaboration/Reference/ReferenceManager.php
```diff
@@ -86,20 +87,24 @@
...
-		$cacheKey = $this->getFullCacheKey($matchedProvider, $referenceId);
+		$cacheKey = $this->getFullCacheKey($matchedProvider, $referenceId, $public, $sharingToken);
 		$cached = $this->cache->get($cacheKey);
 		if ($cached) {
 			return Reference::fromCache($cached);
 		}
 
-		$reference = $matchedProvider->resolveReference($referenceId);
+		if ($public) {
+			$reference = $matchedProvider->resolveReferencePublic($referenceId, $sharingToken);
```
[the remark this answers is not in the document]
max-nextcloud: grafik
Is fixed.

core/Controller/ReferenceApiController.php
```diff
@@ -62,6 +63,39 @@ public function extract(string $text, bool $resolve = false, int $limit = 1): Da
...
+	 * @param string $sharingToken Token of the public share
+	 * @param bool $resolve Resolve the references
+	 * @param int $limit Maximum amount of references to extract
+	 * @return DataResponse<Http::STATUS_OK, array{references: array<string, CoreReference|null>}, array{}>
+	 *
+	 * 200: References returned
+	 */
+	#[ApiRoute(verb: 'POST', url: '/extractPublic', root: '/references')]
+	#[AnonRateLimit(limit: 10, period: 120)]
+	public function extractPublic(string $text, string $sharingToken, bool $resolve = false, int $limit = 1): DataResponse {
```
max-nextcloud, line 81: You are rate limiting here - which is good. But at the same time you allow resolving aka requesting a large number of resources in a single request, right?
There seems to be no limit to `limit`. Are we using more than 1 anyway? What's the highest limit we actually use?

core/Controller/ReferenceApiController.php
```diff
@@ -110,6 +167,36 @@ public function resolve(array $references, int $limit = 1): DataResponse {
...
+	 * @param string[] $references References to resolve
+	 * @param string $sharingToken Token of the public share
+	 * @param int $limit Maximum amount of references to resolve
+	 * @return DataResponse<Http::STATUS_OK, array{references: array<string, CoreReference|null>}, array{}>
+	 *
+	 * 200: References returned
+	 */
+	#[ApiRoute(verb: 'POST', url: '/resolvePublic', root: '/references')]
+	#[AnonRateLimit(limit: 10, period: 120)]
+	public function resolvePublic(array $references, string $sharingToken, int $limit = 1): DataResponse {
```
max-nextcloud, line 184: Same as above. Would be good to have a limit for limit.

max-nextcloud (changes requested):
One comment that will need to be addressed I think.

core/Controller/ReferenceApiController.php
mejo-: Good catch. Limited `$limit` to maximum 15 now in public endpoints.

max-nextcloud (approved):

lib/public/Collaboration/Reference/IPublicReferenceProvider.php
```diff
@@ -0,0 +1,33 @@
...
+ * SPDX-FileCopyrightText: 2022 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCP\Collaboration\Reference;
+
+/**
+ * @since 30.0.0
+ */
+interface IPublicReferenceProvider extends IReferenceProvider {
```
nickvergessen, line 14: @mejo- Please document in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#id1

(also all other API additions/changes in lib/public/)

mejo-: Done so now in https://github.com/nextcloud/documentation/pull/12037. The HTTP API additions got already documented with https://github.com/nextcloud/documentation/pull/12030

lib/public/Collaboration/Reference/LinkReferenceProvider.php
```diff
@@ -26,7 +26,7 @@
 /**
  * @since 29.0.0
  */
-class LinkReferenceProvider implements IReferenceProvider {
+class LinkReferenceProvider implements IReferenceProvider, IPublicReferenceProvider {
```
julien-nc, line 29: `IPublicReferenceProvider` extends `IReferenceProvider` so `LinkReferenceProvider` could just extend `IPublicReferenceProvider`, right?

julien-nc (commented):
A bit late to the party... LGTM

PR #46379, opened 2024-07-09
fix: `OCP\Files\Node\Folder::search` was not setting the owner

susnux:
* Fixed Ticket#9663948

## Summary

The owner was not set on the file info causing e.g. webdav searches to never return the known owner.

## Checklist

### DIFF
--- a/lib/private/Files/Node/Folder.php
+++ b/lib/private/Files/Node/Folder.php
@@ -12,6 +12,7 @@ use OC\Files\Search\SearchComparison;
 use OC\Files\Search\SearchOrder;
 use OC\Files\Search\SearchQuery;
 use OC\Files\Utils\PathHelper;
+use OC\User\LazyUser;
 use OCP\Files\Cache\ICacheEntry;
 use OCP\Files\FileInfo;
 use OCP\Files\Mount\IMountPoint;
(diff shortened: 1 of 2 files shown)


susnux:
cc @mgallien

lib/private/Files/Node/Folder.php
```diff
@@ -245,7 +245,23 @@ private function cacheEntryToFileInfo(IMountPoint $mount, string $appendRoot, IC
 		$cacheEntry['internalPath'] = $cacheEntry['path'];
 		$cacheEntry['path'] = rtrim($appendRoot . $cacheEntry->getPath(), '/');
 		$subPath = $cacheEntry['path'] !== '' ? '/' . $cacheEntry['path'] : '';
-		return new \OC\Files\FileInfo($this->path . $subPath, $mount->getStorage(), $cacheEntry['internalPath'], $cacheEntry, $mount);
+		$storage = $mount->getStorage();
+
+		$owner = null;
+		$ownerId = $storage->getOwner($cacheEntry['internalPath']);
+		if (!empty($ownerId)) {
+			$userManager = \OCP\Server::get(IUserManager::class);
```
icewind1991, line 253: It would be better to get the user manager once, either in the constructor or saving it in a member.

susnux: Makes sense, fixed.

come-nc (commented):
It would make a lot of sense to use a LazyUser instance for this, no?

susnux:
> It would make a lot of sense to use a LazyUser instance for this, no?

Good idea! Implemented!

come-nc (approved):

mgallien:
@susnux  I tested it successfully
I also backported it to stable29
should I approve ?
can you keep in mind that this needs backport to stable29

yemkareems (approved):

PR #46383, opened 2024-07-09
feat: Template field workflow

elzody:
* Resolves: [nextcloud-gmbh/office-team#51](https://github.com/nextcloud-gmbh/office-team/issues/51)

## Summary

This PR introduces the server API and front end changes for filling out template form controls before creating a document based on a template which has fields. If a template doesn't have fields, it will continue like the normal process before.

## TODO

- [ ] Testing -- test coverage will likely be implemented in a follow-up PR
- [ ] Documentation -- if not updated, will also be updated in a follow-up PR

## Checklist


lib/public/Files/Template/Template.php
```diff
@@ -48,6 +50,13 @@ public function setHasPreview(bool $hasPreview): void {
 		$this->hasPreview = $hasPreview;
 	}
 
+	/**
+	 * @since 30.0.0
+	 */
+	public function setFields(array $fields): void {
```
juliushaertl, line 56: Maybe we could actually move the Field class type to server and make it a public API.

lib/public/Files/Template/Field.php
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Richdocuments\Template;
+
+class Field {
+	private FieldType $type;
+	private int $index;
```
juliushaertl, line 12: Maybe we make this a string as identifier to be more generic to also be able to identify b a unique id rather then an index

juliusknorr (approved):

skjnldsv:
> * test coverage will ~~likely~~ be implemented in a follow-up PR


Would be really preferable :pray:

apps/files/src/components/TemplateFiller.vue
```diff
@@ -0,0 +1,103 @@
...
+			this.localFields[index] = {
+				content: value,
+			}
+		},
+		async submit() {
+			this.loading = true
+
+			await this.onSubmit(this.localFields)
+
+			this.show = false
```
skjnldsv, line 79: Wouldn't that leave the modal instance alive?
Emitting a close event should be enough

juliushaertl: Fixed

lib/public/Files/Template/Field.php
```diff
@@ -0,0 +1,53 @@
...
+	 */
+	public function __construct(string $index, string $content, string|FieldType $type, ?string $alias = null, ?int $id = null, ?string $tag = null) {
+		$this->index = $index;
+		$this->alias = $alias;
+		$this->id = $id;
+		$this->tag = $tag;
+		$this->content = $content;
+
+		if ($type instanceof FieldType) {
+			$this->type = $type;
```
blizzz, line 34: why is `$type` allowed to be a string then?

blizzz (commented):
one question, skipped looking at vue files.

lib/public/Files/Template/Field.php
juliushaertl: Adjusted

juliushaertl: And moved to https://github.com/nextcloud/richdocuments/pull/3798/commits/5f730415c096a4201011e74e51490ec934808a10

skjnldsv (approved):

blizzz (approved):

PR #46391, opened 2024-07-10
chore: Make app icons consistent Material Symbols with same settings

susnux:
## Summary

Make all app icons the same style Material Symbols (e.g. files was still MDI, and some had different optical size).

## Checklist


nickvergessen (approved):

jancborchardt (approved):
Looks very nice!

apps/files_external/img/app-dark.svg
AndyScherzinger, line 1: not sure about changing this icon this way. I'd rather stick to the former one which is also Material Symbol https://fonts.google.com/icons?icon.query=open&selected=Material+Symbols+Outlined:open_in_new:FILL@1;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23e8eaed

What do you think @jancborchardt

apps/files_external/img/app.svg
AndyScherzinger, line 1: same as here https://fonts.google.com/icons?icon.query=open&selected=Material+Symbols+Outlined:open_in_new:FILL@1;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23e8eaed

apps/files_versions/img/app-dark.svg
AndyScherzinger, line 1: https://fonts.google.com/icons?icon.query=restore&selected=Material+Symbols+Outlined:settings_backup_restore:FILL@1;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23e8eaed the former icon has also been used as the icon on the tab. So for consistency I suggest to keep that one or replace the tab icon as well in the right sidebar

apps/files_versions/img/app.svg
AndyScherzinger, line 1: same as here https://github.com/nextcloud/server/pull/46391/files#r1672231666

AndyScherzinger (approved):
some minor changes suggested

apps/files_external/img/app-dark.svg
susnux: "Open in new" is well known as a link icon, so not sure.
Maybe also an option:
https://fonts.google.com/icons?icon.query=server&selected=Material+Symbols+Outlined:dns:FILL@1;wght@400;GRAD@0;opsz@20&icon.size=20&icon.color=%23fff

apps/files_versions/img/app.svg
susnux: Makes sense, fixed

apps/files_versions/img/app-dark.svg
susnux: Makes sense, fixed

susnux:
Potential follow up:
There seem to be quite some images we do not use anymore and possibly can remove.
Also the weather status has some other icons that we could migrate to Material Symbols (they provide all kind of weather icons).

apps/files_external/img/app-dark.svg
AndyScherzinger: Well, not too important, so I am also fine to just merge this one as is.
If we get feedback that it confuses folks, we can still change it then.
DNS icon also looks fine from my pov.

AndyScherzinger:
Let's merge it than as-is right now, to have it deployed.

Replacing weather icons with material would be nice in case there actually are variants for all cases.

AndyScherzinger (approved):

PR #46395, opened 2024-07-10
fix: set default TTL for APCu cache as per docs

icewind1991:
Make the method behavior match the description

### DIFF
--- a/lib/public/ICache.php
+++ b/lib/public/ICache.php
@@ -16,6 +16,11 @@ namespace OCP;
  */
 interface ICache {
 	/**
+	 * @since 30.0.0
+	 */
+	public const DEFAULT_TTL = 24 * 60 * 60;
+
+	/**
 	 * Get a value from the user cache
 	 * @param string $key
 	 * @return mixed
--- a/lib/private/Memcache/APCu.php
+++ b/lib/private/Memcache/APCu.php
@@ -25,6 +25,9 @@ class APCu extends Cache implements IMemcache {
 	}
 
 	public function set($key, $value, $ttl = 0) {
+		if ($ttl === 0) {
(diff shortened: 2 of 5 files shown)


sorbaugh (approved):

lib/public/ICache.php
```diff
@@ -15,6 +15,8 @@
  * @since 6.0.0
  */
 interface ICache {
+	const DEFAULT_TTL = 24 * 60 * 60;
```
nickvergessen, line 18:
```suggestion
	/**
	 * @since 30.0.0
	 */
	const DEFAULT_TTL = 24 * 60 * 60;
```

nickvergessen (approved):
Memcached seems to have the same problem?

joshtrichards (commented):
Couple thoughts:

- Should we do this for `inc` and `dec` too? e.g. https://www.php.net/manual/en/function.apcu-inc.php
- Should we adjust the logging wrapper to include the ttl at the same time?
- Probably out of scope, but may also be an opportunity clean-up the no longer needed code for an early PHP 7 bug in `inc` & `dec` a bit farther down

icewind1991:
- clarified the documentation for `inc` and `dec` and cleanup up the implementation now that all supported versions behave the same

ChristophWurst (approved):

Altahrim (approved):

SystemKeeper:
> * clarified the documentation for `inc` and `dec` and cleanup up the implementation now

Is it correct that you removed the comment from `dec` but did not change the implementation? (Only for `inc`)?

juliusknorr (approved):

solracsf:
This is causing troubles on v30: https://github.com/nextcloud/server/issues/48007

PR #46414, opened 2024-07-10
feat: Add new forbidden filename options to Capabilities

susnux:
* For #44963 

## Summary

Allow clients to access the new filename validation options and make frontend name validation possible.

## Checklist

### DIFF
--- a/apps/files/lib/Capabilities.php
+++ b/apps/files/lib/Capabilities.php
@@ -7,28 +7,31 @@
  */
 namespace OCA\Files;
 
+use OC\Files\FilenameValidator;
 use OCP\Capabilities\ICapability;
-use OCP\IConfig;
 
 class Capabilities implements ICapability {
 
-	protected IConfig $config;
-
-	public function __construct(IConfig $config) {
-		$this->config = $config;
+	public function __construct(
+		protected FilenameValidator $filenameValidator,
+	) {
 	}
 
 	/**
 	 * Return this classes capabilities
 	 *
-	 * @return array{files: array{bigfilechunking: bool, blacklisted_files: array<mixed>, forbidden_filename_characters: array<string>}}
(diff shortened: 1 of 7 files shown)


apps/files/lib/Capabilities.php
```diff
@@ -7,28 +7,31 @@
...
 	 * Return this classes capabilities
 	 *
-	 * @return array{files: array{bigfilechunking: bool, blacklisted_files: array<mixed>, forbidden_filename_characters: array<string>}}
+	 * @return array{files: array{'$comment': ?string, bigfilechunking: bool, blacklisted_files: array<mixed>, forbidden_filenames: list<string>, forbidden_filename_characters: list<string>, forbidden_filename_extensions: list<string>}}
 	 */
-	public function getCapabilities() {
+	public function getCapabilities(): array {
 		return [
 			'files' => [
+				'$comment' => '"blacklisted_files" is deprecacted as of Nextcloud 30, use "forbidden_filenames" instead',
```
provokateurin, line 28:
```suggestion
				'$comment' => '"blacklisted_files" is deprecated as of Nextcloud 30, use "forbidden_filenames" instead',
```

provokateurin, line 28: TBH I don't think any developer will read this deprecation notice in the responses. I'd suggest you just put this in the app upgrade docs where it will actually be read.

come-nc (approved):

apps/files/lib/Capabilities.php
come-nc: Adding it to both the documentation and the code is good I think

apps/files/lib/Capabilities.php
```diff
@@ -7,28 +7,31 @@
  */
 namespace OCA\Files;
 
+use OC\Files\FilenameValidator;
```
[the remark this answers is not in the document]
susnux: This is the only place we need this functions (the getters), thats why I asked if it should be public or private interface.

provokateurin (approved):

apps/files/lib/Capabilities.php
susnux: Documentation: https://github.com/nextcloud/documentation/pull/11995

PR #46418, opened 2024-07-10
feat(users): Add users and group management to admin delegation

artonge:
## Done

- [x] Fix some error handling in the front end
- [x] Create a `IDelegatedSettings` for users management
- [x] Show the 'Accounts' menu to delegated admins
- [x] Add the `AuthorizedAdminSetting` annotation to endpoints that are admin restricted
- [x] Tweak the inner permissions conditions in endpoints that are not admin restricted 
- [x] Hide the 'Admins' section to delegated admins

I suspect the most critical part is to not let delegated admins escalate privileges to full admins. I tried to ensure that this is not possible. So a delegated admin cannot:
- Create a user with admin rights
- Delete, edit, wipe devices, disable an admin
- Add a user to the admin group
- Remove an admin from the admin group

But a delegated admin can:
- Escalate admin settings delegation by adding himself to a group

But I might have missed a scenario.

- Doc https://github.com/nextcloud/documentation/pull/12056


nickvergessen:
Should adjust the docs then :) 

https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/admin_delegation_configuration.html

The Info box
> [!NOTE]
> Delegation of user management isn’t handled here, but through the use of [Group Administrators](https://docs.nextcloud.com/server/latest/admin_manual/configuration_user/user_configuration.html).

apps/settings/lib/Controller/UsersController.php
```diff
@@ -200,7 +205,8 @@ public function usersList(): TemplateResponse {
 		// groups
 		$serverData['groups'] = array_merge_recursive($adminGroup, [$recentUsersGroup, $disabledUsersGroup], $groups);
 		// Various data
-		$serverData['isAdmin'] = $isAdmin;
+		$serverData['isAdmin'] = $isAdmin || $isDelegatedAdmin;
```
Pytal, line 208: Not defensive enough and could lead to issues

Better to check explicitly if admin or delegatedAdmin on the frontend code like the backend no?

Pytal (commented):
Tested and code makes sense, one question though

lib/private/Group/Manager.php
```diff
@@ -333,6 +328,14 @@ public function isAdmin($userId) {
 		return $this->isInGroup($userId, 'admin');
 	}
 
+	public function isDelegatedAdmin(string $userId) {
```
blizzz, line 331:
```suggestion
	public function isDelegatedAdmin(string $userId): bool {
```

lib/private/SubAdmin.php
```diff
@@ -233,6 +233,11 @@ public function isSubAdmin(IUser $user): bool {
 			return true;
 		}
 
+		// Check if the user is already an admin
+		if ($this->groupManager->isDelegatedAdmin($user->getUID())) {
+			return true;
+		}
```
blizzz, line 239: Is it good here? Sometimes subadmins cannot do as much as full admins, and delegated ones are rather full ones in their scope.

lib/public/IGroupManager.php
```diff
@@ -114,6 +114,14 @@ public function displayNamesInGroup($gid, $search = '', $limit = -1, $offset = 0
 	 */
 	public function isAdmin($userId);
 
+	/**
+	 * Checks if a userId is eligible to users administration delegation
+	 * @param string $userId
+	 * @return bool if delegated admin
+	 * @since 30.0.0
+	 */
+	public function isDelegatedAdmin(string $userId);
```
blizzz, line 123:
```suggestion
	public function isDelegatedAdmin(string $userId): bool;
```

apps/settings/lib/Settings/Admin/Users.php
```diff
@@ -0,0 +1,61 @@
...
+		private IL10N $l10n,
+	) {
+	}
+
+	/**
+	 * Empty template response
+	 */
+	public function getForm(): TemplateResponse {
+
+		return new /** @template-extends TemplateResponse<\OCP\AppFramework\Http::STATUS_OK, array{}> */ class($this->appName, '') extends TemplateResponse {
```
blizzz, line 32: weirdly formatted. But the comment above?

blizzz (commented):
A question and nitpicks, did not test so far.

lib/private/SubAdmin.php
artonge: We are also checking whether the user is an Admin right above, so I think it makes sense to also check whether it is a delegated admin.

apps/settings/lib/Controller/UsersController.php
artonge: You're right. I made it more fine-grained :)

apps/settings/lib/Settings/Admin/Users.php
artonge: Copy-pasted from another one. But I agree, let's see if psalm is happy.

apps/dav/lib/CalDAV/Schedule/IMipService.php
```diff
@@ -209,6 +209,13 @@ public function generateWhenStringSingular(EventReader $er): string {
 			$endTime = $this->l10n->l('time', $er->endDateTime(), ['width' => 'short']) . ' (' . $er->endTimeZone()->getName() . ')';
 		}
 		// generate localized when string
+		// TRANSLATORS
+		// Indicates when a calendar event will happen, shown on invitation emails
+		// Output produced in order:
+		// In a day/week/month/year on July 1, 2024 for the entire day
+		// In a day/week/month/year on July 1, 2024 between 8:00 AM - 9:00 AM (America/Toronto)
+		// In 2 days/weeks/monthss/years on July 1, 2024 for the entire day
+		// In 2 days/weeks/monthss/years on July 1, 2024 between 8:00 AM - 9:00 AM (America/Toronto)
```
ChristophWurst, line 218: rebase gone bad?

artonge: Yes, sorry for the ping, I think master got messed with.

apps/settings/lib/Settings/Admin/Users.php
artonge: Pslam is not happy, keeping as is :)

Pytal (approved):

yemkareems (approved):

PR #46449, opened 2024-07-11
fix(settings/SetupChecks): Don't follow redirects

marcelklehr:
## Summary
In some cases the reverse proxy config redirects to login when accessing unknown URLs like .ocdata. Login, however, returns a HTTP 200, which is taken to be a failure by the datadirectory protection setup check.

- [ ] Is this a good idea?

## Checklist


come-nc:
I agree with Ferdinand it makes sense to follow redirects, at least for some tests.
But this test is too fragile and should be improved.

Maybe we should add content to `.ocdata` and check that the content is part of the response? (maybe it’s time to rename that file also)

susnux:
See https://github.com/nextcloud/server/pull/46456

PR #46452, opened 2024-07-11
feat(editLocallyAction): Handle possible no local client scenario

### LINKED ISSUE #46438: Workaround when selecting "edit locally" as an option
When selecting "edit locally" as an option, we cannot know if the client is installed on the OS (that's just not possible).

richdocument has a nice way around that, we need to do the same here: https://github.com/nextcloud/server/blob/master/apps/files/src/actions/editLocallyAction.ts

Richdocs solution: https://github.com/nextcloud/richdocuments/blob/main/src/mixins/openLocal.js#L57-L73

nfebe:
## Summary

Resolves: https://github.com/nextcloud/server/issues/46438

## Screencast

[Screencast from 2024-07-12 01-43-50.webm](https://github.com/user-attachments/assets/9e77ae73-7f82-4781-9ad0-117934e4dd0e)

## TODO

- [x] Screenshots ?


apps/files/src/components/modals/ConfirmLocalEditModal.vue
```diff
@@ -0,0 +1,82 @@
+<!--
+  - SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
+  - SPDX-License-Identifier: AGPL-3.0-or-later
+-->
+<template>
+	<NcModal>
```
susnux, line 6: Do not use NcModal for this, if it is a dialog always use NcDialog.
Only for special things like the viewer use a NcModal.

apps/files/src/actions/editLocallyAction.ts
```diff
@@ -27,6 +27,28 @@ const openLocalClient = async function(path: string) {
...
+		},
+		(decision) => {
+			if (!decision) {
+				window.OCA.Viewer.open({ path: fileName })
+				return
+			}
+			this.openingLocally = true
+			this.openLocally()
+		},
+	)
```
susnux, line 49: I would instead of that custom dialog just use the dialog built by the dialogs builder:
https://nextcloud-libraries.github.io/nextcloud-dialogs/classes/index.DialogBuilder.html

Fenn-CS: Yes! Makes more sense. I had checked the NcModals if there's anything configurable there as I was reluctant to reference a component from outside the action.

apps/files/src/actions/editLocallyAction.ts
```diff
@@ -6,11 +6,74 @@ import { encodePath } from '@nextcloud/paths'
...
+				callback: () => {
+					callbackCalled = true
+					localEditCallback(true)
+				},
+			},
+		])
+		.build()
+		.show()
+		.then(() => {
+			// Ensure the callback is called even if the dialog is dismissed in other ways
```
Pytal, line 50: .catch should handle these cases I believe

apps/files/src/actions/editLocallyAction.ts
```diff
@@ -6,11 +6,74 @@ import { encodePath } from '@nextcloud/paths'
...
+	attemptOpenLocalClient(path)
+}
+
+const attemptOpenLocalClient = async (path: string) => {
+	openLocalClient(path)
+	confirmLocalEditDialog(
+		t('files', 'Edit file locally'),
+		t('files', 'The file should now open locally. If you don\'t see this happening, make sure that the desktop client is installed on your system.'),
+		t('files', 'Retry local edit'),
+		t('files', 'Edit online'),
```
Pytal, line 67: Instead of passing these as arguments why not pass them directly into `.setName(t(...)), etc.` directly?

apps/files/src/actions/editLocallyAction.ts
Fenn-CS: .catch is for errors, this is not an error.

Pytal: Ahh didn't work how I thought then, from the source code and it seems docs are wrong https://nextcloud-libraries.github.io/nextcloud-dialogs/classes/index.Dialog.html#show cc @susnux

susnux: I will have a look at the dialogs library, because docs should be right here :see_no_evil:

apps/files/src/actions/editLocallyAction.ts
```diff
@@ -6,11 +6,66 @@ import { encodePath } from '@nextcloud/paths'
...
+					localEditCallback(true)
+				},
+			},
+		])
+		.build()
+		.show()
+		.then(() => {
+			// Ensure the callback is called even if the dialog is dismissed in other ways
+			if (!callbackCalled) {
+				localEditCallback(true)
```
Pytal, line 48: IMO closing the dialog should have the same effect as cancelling

apps/files/src/actions/editLocallyAction.ts
```diff
@@ -43,7 +98,7 @@ export const action = new FileAction({
 	},
 
 	async exec(node: Node) {
-		openLocalClient(node.path)
+		startOpenLocalProcess(node.path)
```
Pytal, line 101:
```suggestion
		attemptOpenLocalClient(node.path)
```
The `startOpenLocalProcess` wrapper seems unnecessary, would call attempt directly

apps/files/src/actions/editLocallyAction.ts
Fenn-CS: I thought so too, but I was like maybe it's better to just proceed with action the user was attempting to do in the first place if they are clicking outside the modal?

Fenn-CS: I don't have any strong feeling about this, any is fine for me.

Pytal (approved):

skjnldsv (approved):

skjnldsv:
@Fenn-CS can you fix the test please?

PR #46456, opened 2024-07-11
fix(SetupCheck): Properly check public access to data directory

susnux:
* Replaces https://github.com/nextcloud/server/pull/46449
* Based on Cômes comment: https://github.com/nextcloud/server/pull/46449#issuecomment-2222988922

## Summary

When checking for public (web) access to the data directory the status is not enough as you might have a webserver that forwards to e.g. a login page. So instead check that the content of the file matches.

For this the `.ncdata` file (renamed from `.ocdata`¹) has minimal text content to allow checking.

¹The file was renamed from the legacy `.ocdata`, there is a repair step to remove the old one.

## Checklist

### DIFF
--- a/lib/private/legacy/OC_Util.php
+++ b/lib/private/legacy/OC_Util.php
@@ -687,7 +687,7 @@ class OC_Util {
 
 	/**
 	 * Check that the data directory exists and is valid by
-	 * checking the existence of the ".ocdata" file.
+	 * checking the existence of the ".ncdata" file.
 	 *
 	 * @param string $dataDirectory data directory path
 	 * @return array errors found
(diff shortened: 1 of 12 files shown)


lib/private/legacy/OC_Util.php
```diff
@@ -701,11 +701,11 @@ public static function checkDataDirectoryValidity($dataDirectory) {
...
 			];
 		}
-		if (!file_exists($dataDirectory . '/.ocdata')) {
+
+		if (!file_exists($dataDirectory . '/.ncdata')) {
 			$errors[] = [
 				'error' => $l->t('Your data directory is invalid.'),
-				'hint' => $l->t('Ensure there is a file called ".ocdata"' .
-					' in the root of the data directory.')
+				'hint' => $l->t('Ensure there is a file called ".ncdata" in the root of the data directory.')
```
nickvergessen, line 708: since the content now matters add that?

come-nc (approved):
Nice!

lib/private/legacy/OC_Util.php
susnux: Should we include the expected content?

AndyScherzinger: I'd say so 👍

susnux: done

AndyScherzinger (approved):

PR #46504, opened 2024-07-13
fix(theming): enforce theme should still provide dark as fallback

skjnldsv:
Discovered from
- https://github.com/nextcloud-libraries/nextcloud-vue/pull/5656
- https://github.com/nextcloud/viewer/pull/2358

Otherwise apps cannot enforce some areas of their UI to be dark, like the Modal component which is always dark.
Here is the result otherwise: image


susnux (approved):

apps/theming/lib/Service/ThemesService.php
```diff
@@ -28,7 +28,7 @@ public function __construct(
 		private LoggerInterface $logger,
 		private DefaultTheme $defaultTheme,
 		LightTheme $lightTheme,
-		DarkTheme $darkTheme,
+		private DarkTheme $darkTheme,
```
ShGKme, line 31: Why make it private?

ShGKme (approved):

ShGKme:
Backport?

apps/theming/lib/Service/ThemesService.php
susnux: Because it is now used in `getThemes`method, before this was not a class attribute but just a constructor parameter (it is now promoted as a private class attribute).

PR #46508, opened 2024-07-13
feat(preview): introducing new class ImaginaryPDF

### LINKED ISSUE #42175: Add Option to Disable Imaginary PDF Previews
### How to use GitHub

I would like to have an option to use Imaginary to generate only image previews not pdf's previews. 

N/A .

**Additional context**
N/A.

ernolf:
* Resolves: #42175

* Replaces: #46447

## Summary
Document previews often appear as minimal, almost blank "miniatures" that can inadvertently reveal content, which may not always be desired.
Many users prefer having document icons instead of these previews.
The classic preview provider settings have considered this preference.

feat: Add separate provider for Imaginary PDF handling

- Refactor Imaginary to handle only image mimetypes.
- Introduce ImaginaryPDF class to handle PDF mimetypes.
- Register ImaginaryPDF in PreviewManager.

This allows for more granular control over which mimetypes are processed by Imaginary.

The approach from #46447 was too cumbersome and deviated too much from the uniform handling of preview providers. This approach has now fully implemented the suggestions and ideas from there.

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/private/Preview/Imaginary.php
+++ b/lib/private/Preview/Imaginary.php
@@ -40,7 +40,7 @@ class Imaginary extends ProviderV2 {
 	}
 
 	public static function supportedMimeTypes(): string {
-		return '/(image\/(bmp|x-bitmap|png|jpeg|gif|heic|heif|svg\+xml|tiff|webp)|application\/(pdf|illustrator))/';
+		return '/(image\/(bmp|x-bitmap|png|jpeg|gif|heic|heif|svg\+xml|tiff|webp)|application\/illustrator)/';
 	}
 
 	public function getCroppedThumbnail(File $file, int $maxX, int $maxY, bool $crop): ?IImage {
--- a/lib/private/Preview/ImaginaryPDF.php
+++ b/lib/private/Preview/ImaginaryPDF.php
@@ -1,0 +1,14 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC\Preview;
+
+class ImaginaryPDF extends Imaginary {
+	public static function supportedMimeTypes(): string {
+		return '/application\/pdf/';
+	}
+}
--- a/lib/private/PreviewManager.php
+++ b/lib/private/PreviewManager.php
@@ -346,6 +346,7 @@ class PreviewManager implements IPreview {
 		$this->registerCoreProvider(Preview\MP3::class, '/audio\/mpeg/');
 		$this->registerCoreProvider(Preview\OpenDocument::class, '/application\/vnd.oasis.opendocument.*/');
 		$this->registerCoreProvider(Preview\Imaginary::class, Preview\Imaginary::supportedMimeTypes());
+		$this->registerCoreProvider(Preview\ImaginaryPDF::class, Preview\ImaginaryPDF::supportedMimeTypes());
 
 		// SVG and Bitmap require imagick
 		if ($this->imagickSupport->hasExtension()) {
--- a/config/config.sample.php
+++ b/config/config.sample.php
@@ -1296,7 +1296,13 @@ $CONFIG = [
 
 /**
  * Set the URL of the Imaginary service to send image previews to.
- * Also requires the ``OC\Preview\Imaginary`` provider to be enabled.
+ * Also requires the
+ *  - ``OC\Preview\Imaginary``
+ * provider to be enabled in the 'enabledPreviewProviders' array, to create previews for these mimetypes:
+ * bmp, x-bitmap, png, jpeg, gif, heic, heif, svg+xml, tiff, webp and illustrator.
+ * If you want Imaginary to also create preview images from PDF Documents, you have to add
+ *  - ``OC\Preview\ImaginaryPDF``
+ * provider as well.
  *
  * See https://github.com/h2non/imaginary
  */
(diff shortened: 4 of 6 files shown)


lib/private/PreviewManager.php
```diff
@@ -346,6 +346,7 @@ protected function registerCoreProviders() {
 		$this->registerCoreProvider(Preview\MP3::class, '/audio\/mpeg/');
 		$this->registerCoreProvider(Preview\OpenDocument::class, '/application\/vnd.oasis.opendocument.*/');
 		$this->registerCoreProvider(Preview\Imaginary::class, Preview\Imaginary::supportedMimeTypes());
+		$this->registerCoreProvider(Preview\ImaginaryPDF::class, '/application\/pdf/');
```
kesselb, line 349:
```suggestion
		$this->registerCoreProvider(Preview\ImaginaryPDF::class, Preview\ImaginaryPDF::supportedMimeTypes());
```

ernolf:
> The rebase went bad

Yes. Fixed. Sory for the noise.

lib/private/PreviewManager.php
ernolf: [c2b2045](https://github.com/nextcloud/server/pull/46508/commits/c2b20453e47bab0e4d01c6829cf06d30e2efadb2)

lib/private/Preview/ImaginaryPDF.php
```diff
@@ -0,0 +1,14 @@
...
+
+/**
+ * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC\Preview;
+
+class ImaginaryPDF extends Imaginary {
+	public function getMimeType(): string {
```
kesselb, line 11:
```suggestion
	public static function supportedMimeTypes(): string
```

It won't work otherwise.

kesselb (changes requested):

kesselb:
Please revert all changes to the autoloaders in `apps/`

lib/private/Preview/ImaginaryPDF.php
ernolf: 9f48093b7bf23b9b3f80858528ef9b9fb62b0e7d

> It won't work otherwise.

Yes. Because of a41ed809f9c10c9532e974af95911933ded047c3 😉

lib/private/Preview/Imaginary.php
```diff
@@ -40,7 +40,7 @@ public function getMimeType(): string {
 	}
 
 	public static function supportedMimeTypes(): string {
-		return '/(image\/(bmp|x-bitmap|png|jpeg|gif|heic|heif|svg\+xml|tiff|webp)|application\/(pdf|illustrator))/';
+		return '/(image\/(bmp|x-bitmap|png|jpeg|gif|heic|heif|svg\+xml|tiff|webp)|application\/illustrator)/';
```
szaimen, line 43:
```suggestion
		return '/(image\/(bmp|x-bitmap|png|jpeg|gif|heic|heif|svg\+xml|tiff|webp)/';
```

lib/private/Preview/ImaginaryPDF.php
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * SPDX-FileCopyrightText: 2020 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC\Preview;
+
+class ImaginaryPDF extends Imaginary {
+	public static function supportedMimeTypes(): string {
+		return '/application\/pdf/';
```
szaimen, line 12:
```suggestion
		return '/application\/(illustrator|pdf)/';
```

szaimen (commented):
an illustrator file is basically also only a pdf file, so should probably be included the imaginarypdf as well I guess

ernolf:
> an illustrator file is basically also only a pdf file, so should probably be included the imaginarypdf as well I guess

Did you read https://github.com/nextcloud/server/pull/46447#issuecomment-2226867819

I wouldn't classify Illustrator files as PDF files, they are more like (non text-)postscript files, which are of course all very closely related. In the narrowest sense, they are vector-based graphics. What convinced me not to include them in PDF is that they have the file extension .ai and not .pdf and therefore are not assigned a unique icon in Nextcloud like is the case with PDF.
I'm happy to be convinced otherwise, but I personally know .ai files as a graphic element that is used (among other) in PDF files.
It would be possible to add an additional class ImaginaryIllustrator instead, but then the same question would arise as with all other MIME types.

Mi aim was to clearly devide pure PDF documents - they have a good Icon which is way better than any preview thumb - from other, not (pure PDF) documents.

szaimen:
> Did you read [#46447 (comment)](https://github.com/nextcloud/server/pull/46447#issuecomment-2226867819)

No I havent.

> I wouldn't classify Illustrator files as PDF files, they are more like (non text-)postscript files, which are of course all very closely related. In the narrowest sense, they are vector-based graphics. What convinced me not to include them in PDF is that they have the file extension .ai and not .pdf and therefore are not assigned a unique icon in Nextcloud like is the case with PDF. I'm happy to be convinced otherwise, but I personally know .ai files as a graphic element that is used (among other) in PDF files. It would be possible to add an additional class ImaginaryIllustrator instead, but then the same question would arise as with all other MIME types.
> 
> My aim was to clearly devide pure PDF documents - they have a good Icon which is way better than any preview thumb - from other, not (pure PDF) documents.

Fine by me then 👍

szaimen (approved):
LGTM but didnt test

kesselb:
Thank you! 

I rebased your pr, updated the autoloaders again and squashed the commits.

kesselb (approved):

kesselb:
Docs: https://github.com/nextcloud/documentation/pull/12061

PR #46527, opened 2024-07-15
feat(search): allows to search by parent ID

Altahrim:
## Summary

Allows to filter on parent ID in a `SEARCH` request

```xml
<?xml version="1.0" encoding="UTF-8"?>
<d:searchrequest xmlns:d="DAV:" xmlns:oc="http://owncloud.org/ns" xmlns:nc="http://nextcloud.org/ns" xmlns:ns="https://github.com/icewind1991/SearchDAV/ns">
  <d:basicsearch>
    <d:select>
      <d:prop>
        <d:displayname/>
        <d:getcontentlength/>
        <d:getcontenttype/>
      </d:prop>
    </d:select>
    <d:from>
      <d:scope>
        <d:href>/files/admin/Media</d:href> <!-- Filter from path -->
        <d:depth>1</d:depth>
      </d:scope>
    </d:from>
    <d:where>
      <d:eq>
        <d:prop>
          <d:parentid/>  <!-- Filter on parentid -->
        </d:prop>
        <d:literal>3</d:literal> <!-- Filter value (oc:fileid of parent directory) -->
      </d:eq>
    </d:where>
    <d:limit>
      <d:nresults>2</d:nresults>
      <ns:firstresult>0</ns:firstresult>
    </d:limit>
  </d:basicsearch>
</d:searchrequest>
```

## Checklist

### DIFF
--- a/apps/dav/lib/Files/FileSearchBackend.php
+++ b/apps/dav/lib/Files/FileSearchBackend.php
@@ -82,6 +82,7 @@ class FileSearchBackend implements ISearchBackend {
 			new SearchPropertyDefinition('{DAV:}displayname', true, true, true),
 			new SearchPropertyDefinition('{DAV:}getcontenttype', true, true, true),
 			new SearchPropertyDefinition('{DAV:}getlastmodified', true, true, true, SearchPropertyDefinition::DATATYPE_DATETIME),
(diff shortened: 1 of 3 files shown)


icewind1991:
The change looks fine, but is there any reason for not just using the parent folder as scope and searching with depth 1?

Altahrim:
Good question! It doesn't work because we need to have a search but then I forgot the reason…
It could be a better choice to allow an empty search

Altahrim:
Another reason: according to my tests `depth` isn't respected…

PR #46531, opened 2024-07-15
fix(CI): Skip some CI work from forks as they don't work anyway

nickvergessen:
## Checklist

### DIFF
--- a/.github/workflows/command-compile.yml
+++ b/.github/workflows/command-compile.yml
@@ -26,6 +26,12 @@ jobs:
       base_ref: ${{ steps.comment-branch.outputs.base_ref }}
 
     steps:
+      - name: Disabled on forks
+        if: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
+        run: |
+          echo 'Can not execute /compile on forks'
+          exit 1
+
       - name: Check actor permission
         uses: skjnldsv/check-actor-permission@69e92a3c4711150929bca9fcf34448c5bf5526e7 # v2
         with:
@@ -56,6 +62,15 @@ jobs:
       - name: Init branch
         uses: xt0rted/pull-request-comment-branch@d97294d304604fa98a2600a6e2f916a84b596dc7 # v1
         id: comment-branch
+
+      - name: Add reaction on failure
+        uses: peter-evans/create-or-update-comment@71345be0265236311c031f5c7866368bd1eff043 # v4.0.0
+        if: failure()
+        with:
+          token: ${{ secrets.COMMAND_BOT_PAT }}
+          repository: ${{ github.event.repository.full_name }}
+          comment-id: ${{ github.event.comment.id }}
+          reactions: "-1"
 
   process:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


provokateurin (approved):

bigcat88 (approved):

.github/workflows/command-compile.yml
```diff
@@ -26,6 +26,12 @@ jobs:
       base_ref: ${{ steps.comment-branch.outputs.base_ref }}
 
     steps:
+      - name: Disabled on forks
+        if: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
```
st3iny, line 30:
```suggestion
        if: ${{ github.event.pull_request.head.repo.fork }}
```

:nerd_face: 

(There is no functional difference though.)

AndyScherzinger (approved):

.github/workflows/cypress.yml
```diff
@@ -32,6 +32,12 @@ jobs:
       PUPPETEER_SKIP_DOWNLOAD: true
 
     steps:
+      - name: Disabled on forks
+        if: ${{ github.event.pull_request.head.repo.full_name != github.repository }}
+        run: |
+          echo 'Can not run cypress on forks'
+          exit 1
```
st3iny, line 39: I think we would then also need to make cypress not required which is not good.

You should also add a similar check to the final `summary` step to make it not fail in case of a fork.

st3iny: Otherwise, the CI will always be red on forks I think.

provokateurin: But that can be dangerous as well, if cypress displays green just because it is from a fork and didn't run any tests.

provokateurin: This PR only makes the workflow fail much earlier to save CI time

st3iny: Right, thanks. That makes sense.

nickvergessen: No, it basically means you need to send the PR from local or there will be no cypress results... Not a lot we can do either way.
It must be required as it could be breaking. But a PR from 3rdparty can never validate if it breaks it.

skjnldsv:
image

skjnldsv:
https://github.com/nextcloud/server/pull/46522

skjnldsv:
```
##[debug]Evaluating condition for step: 'Disabled on forks'
##[debug]Evaluating: (success() && (github.event.pull_request.head.repo.full_name != github.repository))
##[debug]Evaluating And:
##[debug]..Evaluating success:
##[debug]..=> true
##[debug]..Evaluating NotEqual:
##[debug]....Evaluating Index:
##[debug]......Evaluating Index:
##[debug]........Evaluating Index:
##[debug]..........Evaluating Index:
##[debug]............Evaluating Index:
##[debug]..............Evaluating github:
##[debug]..............=> Object
##[debug]..............Evaluating String:
##[debug]..............=> 'event'
##[debug]............=> Object
##[debug]............Evaluating String:
##[debug]............=> 'pull_request'
##[debug]..........=> null
##[debug]........=> null
##[debug]......=> null
##[debug]....=> null
##[debug]....Evaluating Index:
##[debug]......Evaluating github:
##[debug]......=> Object
##[debug]......Evaluating String:
##[debug]......=> 'repository'
##[debug]....=> 'nextcloud/server'
##[debug]..=> true
##[debug]=> true
##[debug]Expanded: (true && (null != 'nextcloud/server'))
##[debug]Result: true
##[debug]Starting: Disabled on forks
```

nickvergessen:
Ah, issue_comment runs on master.
Checking

PR #46539, opened 2024-07-15
feat: allow for ExApps to call Admin endpoints marked with specific attribute

bigcat88:
## Summary

We need this for the Workflow Engine Project, and not only.

The option of duplicating ednpoints where they do not require the admin flag to be set seems to us not quite the right solution; sometimes for ExApps you still need to check whether the user is an administrator or not.

1. If you need to check the user for ExApp and it works in the user’s context, it sets the userId and everything works as it did.
2. Only if the request comes from ExApp and the user is not set in the session, then the check for the attributes "AuthorizedAdminSetting" and "SubAdminRequired" is skipped.

## Checklist


julien-nc:
@nickvergessen What do you think about that?

nickvergessen:
Will have a look and answer tomorrow.

lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
```diff
@@ -136,11 +137,19 @@ public function beforeController($controller, $methodName) {
 				throw new ExAppRequiredException();
 			}
 		} elseif (!$isPublicPage) {
-			if (!$this->isLoggedIn) {
+			$authorized = false;
+			if ($this->hasAnnotationOrAttribute($reflectionMethod, 'AppApiAdminAccessWithoutUser', AppApiAdminAccessWithoutUser::class)) {
```
nickvergessen, line 141: Let's not introduce legacy structure for new things:
```suggestion
			if ($this->hasAnnotationOrAttribute($reflectionMethod, null, AppApiAdminAccessWithoutUser::class)) {
```

And apply this:
```diff
diff --git a/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php b/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
index bc2014da246..5418a083ad2 100644
--- a/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/SecurityMiddleware.php
@@ -233,16 +233,16 @@ class SecurityMiddleware extends Middleware {
         * @template T
         *
         * @param ReflectionMethod $reflectionMethod
-        * @param string $annotationName
+        * @param ?string $annotationName
         * @param class-string<T> $attributeClass
         * @return boolean
         */
-       protected function hasAnnotationOrAttribute(ReflectionMethod $reflectionMethod, string $annotationName, string $attributeClass): bool {
+       protected function hasAnnotationOrAttribute(ReflectionMethod $reflectionMethod, ?string $annotationName, string $attributeClass): bool {
                if (!empty($reflectionMethod->getAttributes($attributeClass))) {
                        return true;
                }
 
-               if ($this->reflector->hasAnnotation($annotationName)) {
+               if ($annotationName && $this->reflector->hasAnnotation($annotationName)) {
                        return true;
                }
 
```

nickvergessen (commented):
Since this opens a wide angle on our APIs, we should make sure that logging (and admin audit logging) is also aware of `$this->userSession->getSession()->get('app_api')` and properly logs that an action was performed "on behalf of a user" and not "by a user".

But be very careful in the logging so it never fails (catch throwable, etc.) as otherwise logging breaks completely.

nickvergessen:
Additionally this would allow a malicious admin to escape `AuthorizedAdminSetting` ?

bigcat88:
> Additionally this would allow a malicious admin to escape `AuthorizedAdminSetting` ?

If I correctly understood the situation you describe, then the answer is probably “no”:
to do this, the administrator must have access to the infrastructure to obtain the ExApp key.



> Since this opens a wide angle on our APIs, we should make sure that logging (and admin audit logging) is also aware of `$this->userSession->getSession()->get('app_api')` and properly logs that an action was performed "on behalf of a user" and not "by a user".
> 
> But be very careful in the logging so it never fails (catch throwable, etc.) as otherwise logging breaks completely.

should this be done in a separate PR for cleanliness?

nickvergessen:
> If I correctly understood the situation you describe, then the answer is probably “no”:
to do this, the administrator must have access to the infrastructure to obtain the ExApp key.

Might be a "no brainer", but I meant if they can install exapps they can now also perform admin actions on the server (on any level), by writing their own exapp and enabling it.
But mostlikely they could also do a normal php app then

nickvergessen:
> should this be done in a separate PR for cleanliness?

If you can do it in a separate PR before beta1 next week?

bigcat88:
I will re-create this PR as I messed it with merging commits from master branch.

PR #46543, opened 2024-07-15
fix(htaccess): Serve images as static assets

susnux:
## Checklist

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -40,7 +40,7 @@
   </IfModule>
 
   # Add cache control for static resources
-  <FilesMatch "\.(css|js|mjs|svg|gif|png|jpg|ico|wasm|tflite)$">
+  <FilesMatch "\.(css|js|mjs|svg|gif|png|jpg|webp|ico|wasm|tflite)$">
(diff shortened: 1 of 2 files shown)


susnux:
I hope this fixes the daily instance not serving images correctly (especially `webp`).

SystemKeeper (approved):
Makes sense, but guess needs an htaccess update on daily then.

susnux:
@SystemKeeper yes probably but maybe even some Caddy config update.

ChristophWurst (approved):
Makes sense

ChristophWurst:
https://docs.nextcloud.com/server/latest/admin_manual/installation/nginx.html#nextcloud-in-the-webroot-of-nginx will need an update too and should be noted in https://docs.nextcloud.com/server/latest/admin_manual/release_notes/upgrade_to_30.html

susnux:
But I do not think this is the root cause as locally deployed with Apache it works, only the daily instance with Caddy seems to have problem :(

susnux:
(I mean it is an improvement anyways)

PR #46545, opened 2024-07-15
feat: Add `forbidden_filename_basenames` config option

susnux:
* For #44963 

## Summary
This allows to configure forbidden filenames (the full filename like `.htaccess`) and also forbidden basenames like `com0` where `com0`, `com0.txt` and `com0.tar.gz` will match. We need this as only using basenames was too restrictive and will cause problems on some systems when updating.

## Checklist

### DIFF
--- a/apps/files/lib/Capabilities.php
+++ b/apps/files/lib/Capabilities.php
@@ -20,7 +20,7 @@ class Capabilities implements ICapability {
 	/**
 	 * Return this classes capabilities
 	 *
-	 * @return array{files: array{'$comment': ?string, bigfilechunking: bool, blacklisted_files: array<mixed>, forbidden_filenames: list<string>, forbidden_filename_characters: list<string>, forbidden_filename_extensions: list<string>}}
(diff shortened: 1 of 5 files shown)


solracsf:
`'forbidden_filename_basenames' => ['archive'],`

This will also block `Archive.2024.tar.gz`

If the function remains as-is, where first dot defines the basename, this needs perhaps a better documentation (a warning ⚠️ ) about this behaviour so admin fully understand the risk. _Or a refactor 😄_

susnux:
> This will also block `Archive.2024.tar.gz`

Yes this is exactly what it should do, it is only for filtering basenames. Needed for Windows support as on Windows `com0` and `com0.tar.gz` are both the same invalid filename (see [windows namespaces](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces)).

skjnldsv (approved):

yemkareems (approved):

PR #46547, opened 2024-07-15
feat: allow running QueryBuilder queries on different connections

icewind1991:
Allows building a query from one connection but executing it on a different one.

Needed for, and extracted from the sharding code.

### DIFF
--- a/lib/private/DB/QueryBuilder/QueryBuilder.php
+++ b/lib/private/DB/QueryBuilder/QueryBuilder.php
@@ -13,6 +13,7 @@ use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
 use Doctrine\DBAL\Query\QueryException;
 use OC\DB\ConnectionAdapter;
+use OC\DB\Exceptions\DbalException;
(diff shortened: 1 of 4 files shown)


artonge (approved):

yemkareems (approved):

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -253,48 +246,63 @@ public function execute() {
...
+	 * @return IResult|int
+	 */
+	public function execute(?IDBConnection $connection = null) {
+		try {
+			if ($this->getType() === \Doctrine\DBAL\Query\QueryBuilder::SELECT) {
+				return $this->executeQuery($connection);
+			} else {
+				return $this->executeStatement($connection);
+			}
+		} catch (DBALException $e) {
```
come-nc, line 266: `DBALException` does not exists here.
What are you trying to catch?

come-nc:
```suggestion
		} catch (DbalException $e) {
```

Actually it’s only a case issue, there is a `use` statement for `DbalException`

PR #46557, opened 2024-07-16
refactor: Improve design of declarative settings

susnux:
## Summary

* Remove the hint from DOM if empty (otherwise they still take some place in layout)
* Individual fields of the same section should not be spaced 60px, 20px (2x 10px) should be enough
* For checkboxes there is no need for an external label if there is the checkbox label, so as only `title` is required but not `label` we set the checkbox label if only `title` is set and set both if both are provided.
* We have the type `multi-checkbox` for real checkboxes where you can select on of multiple, and we have `checkbox` which is only used for enable / disable, so we shall use `switch` type of the checkbox component as we do for all other (non declarative) settings

### Screenshots

before | after
---|---
Screenshot 2024-07-16 at 12-34-44 Basic settings - Administration settings - Nextcloud|Screenshot 2024-07-16 at 12-34-10 Basic settings - Administration settings - Nextcloud

## Checklist


marcoambrosini (approved):
✅ for the design side, nice improvement :)

andrey18106 (approved):
👍

artonge (approved):

Pytal (approved):

bigcat88 (approved):

PR #46572, opened 2024-07-17
feat!: Adjust default php timeout for controller requests and add attribute to adjust

juliusknorr:
## Summary

Right now we always have a max_execution_time of 1h which is really long and can block php workers quite a bit. In most cases requests should finish a lot faster and if they don't there is something wrong.

Long timeouts are as far as I see only needed for file uploads.

This PR introduces a RequestTimeout attribute for controllers that can be used to control expected maximum execution time and in addition uses a default of 25 seconds (to be below any default webserver timeouts). As this only applies to Controllers dav is not affected by this change.

## TODO

- [ ] Check controllers for where we can apply even lower timeouts

## Checklist


juliushaertl:
@nickvergessen @Altahrim Would be curious about your thoughts on this

lib/private/AppFramework/Middleware/RequestTimeMiddleware.php
```diff
@@ -0,0 +1,29 @@
...
+		$timeout = 30;
+		$reflectionMethod = new ReflectionMethod($controller, $methodName);
+		$attributes = $reflectionMethod->getAttributes(RequestTimeout::class);
+		foreach ($attributes as $attribute) {
+			/** @var RequestTimeout $timeout */
+			$timeoutAttribute = $attribute->newInstance();
+			$timeout = $timeoutAttribute->getTimeout();
+		}
+
+		@ini_set('max_execution_time', strval($timeout));
```
nickvergessen, line 27: Do we need to be more passive?
https://github.com/nextcloud/server/blob/9100b8757e3cf9b5588ae58728e17fee6a38ced6/lib/base.php#L612-L627

lib/private/AppFramework/Middleware/RequestTimeMiddleware.php
```diff
@@ -0,0 +1,29 @@
...
+
+use OCP\AppFramework\Controller;
+use OCP\AppFramework\Http\Attribute\RequestTimeout;
+use OCP\AppFramework\Middleware;
+use ReflectionMethod;
+
+class RequestTimeMiddleware extends Middleware {
+	public function beforeController(Controller $controller, string $methodName) {
+		// Default timeout
+		$timeout = 30;
```
nickvergessen, line 18: Not a fan of this.
E.g. Talk recording upload and all other things that are on normal controllers have a changing behaviour and something like this is hard to notice in dev life :see_no_evil: 
Also it breaks debuggers? :see_no_evil:

juliushaertl: Yeah, though i think it would be a great move towards having more reasonable request behaviour on larger instances, especially to prevent overload due to very long running requests, but would also close it for now as I don't see capacity to actually move that way.

PR #46573, opened 2024-07-17
feat: Add global rate limiting config

juliusknorr:
## Summary

Request bursts are a regular problem and can cause a Nextcloud server to become unavailable. We already have rate limiting on a per controller annotation but for general bursts we don't have any way to handle them. 

This PR proposes as general rate limit config that admins can set and we should probably have some sane defaults.

## TODO

- [ ] See if the defaults make sense
- [ ] Check if we also not only want to apply based on the controller method identifier but also globally per user/annon user

## Checklist

### DIFF
--- a/lib/private/AppFramework/Middleware/Security/RateLimitingMiddleware.php
+++ b/lib/private/AppFramework/Middleware/Security/RateLimitingMiddleware.php
@@ -19,6 +19,7 @@ use OCP\AppFramework\Http\DataResponse;
 use OCP\AppFramework\Http\Response;
 use OCP\AppFramework\Http\TemplateResponse;
 use OCP\AppFramework\Middleware;
+use OCP\IConfig;
 use OCP\IRequest;
 use OCP\ISession;
 use OCP\IUserSession;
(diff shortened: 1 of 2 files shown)


lib/private/AppFramework/Middleware/Security/RateLimitingMiddleware.php
```diff
@@ -93,7 +95,26 @@ public function beforeController(Controller $controller, string $methodName): vo
 				$rateLimit->getPeriod(),
 				$this->request->getRemoteAddress()
 			);
+			return;
 		}
+
+		$globalLimit = $this->config->getSystemValueInt('global-rate-limit.limit', 10);
```
nickvergessen, line 101: this will break heavily :/

juliushaertl: In terms of the number 10 or the general concept?

nickvergessen: loading a talk conversation during a (starting call) loads quite some avatars, chat messages, participant list, previews, ...
reached rate limit will break the request with 429 …

nickvergessen: And also to have it said. a global rate limit means each request will do 2 queries without redis and 2 redis requests otherwise:
1. check current count of requests
2. add new API call

juliushaertl:
Let me close this for now, we can reopen whenever it becomes relevant to discuss again

PR #46579, opened 2024-07-17
[TaskProcessing] Add support for webhooks

julien-nc:
* 2 new columns/attributes for tasks: webhookUri and webhookMethod
* When a task finishes (is canceled, fails or succeeds), we try to "run" the webhook
* If the method is "HTTP:GET", "HTTP:POST"... we make a simple HTTP request
* If the method is "AppAPI:APP_ID:HTTP_METHOD" we directly send a request to the exApp
* The request body contains the task ID and status

Questions:

* [ ] Do we want to combine so much information in the method column/attr?
* [x] What do we want to include in the request body/headers?

### DIFF
--- a/lib/private/TaskProcessing/Manager.php
+++ b/lib/private/TaskProcessing/Manager.php
@@ -9,9 +9,12 @@ declare(strict_types=1);
 
 namespace OC\TaskProcessing;
 
+use GuzzleHttp\Exception\ClientException;
+use GuzzleHttp\Exception\ServerException;
 use OC\AppFramework\Bootstrap\Coordinator;
 use OC\Files\SimpleFS\SimpleFile;
 use OC\TaskProcessing\Db\TaskMapper;
+use OCP\App\IAppManager;
 use OCP\AppFramework\Db\DoesNotExistException;
 use OCP\AppFramework\Db\MultipleObjectsReturnedException;
 use OCP\BackgroundJob\IJobList;
@@ -27,6 +30,7 @@ use OCP\Files\IRootFolder;
 use OCP\Files\Node;
 use OCP\Files\NotPermittedException;
 use OCP\Files\SimpleFS\ISimpleFile;
+use OCP\Http\Client\IClientService;
 use OCP\IConfig;
 use OCP\IL10N;
 use OCP\IServerContainer;
@@ -53,6 +57,8 @@ use OCP\TaskProcessing\TaskTypes\TextToText;
 use OCP\TaskProcessing\TaskTypes\TextToTextHeadline;
 use OCP\TaskProcessing\TaskTypes\TextToTextSummary;
 use OCP\TaskProcessing\TaskTypes\TextToTextTopics;
+use Psr\Container\ContainerExceptionInterface;
+use Psr\Container\NotFoundExceptionInterface;
 use Psr\Log\LoggerInterface;
 
 class Manager implements IManager {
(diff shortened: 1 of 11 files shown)


lib/private/TaskProcessing/Manager.php
```diff
@@ -884,4 +897,69 @@ public function setTaskStatus(Task $task, int $status): void {
...
+		if (!$uri || !$method) {
+			return;
+		}
+
+		if (in_array($method, ['HTTP:GET', 'HTTP:POST', 'HTTP:PUT', 'HTTP:DELETE'], true)) {
+			$client = $this->clientService->newClient();
+			$httpMethod = preg_replace('/^HTTP:/', '', $method);
+			$options = [
+				'body' => json_encode([
+					'taskId' => $task->getId(),
```
marcelklehr, line 919: I would perhaps send the whole json representation of the task, because the hook endpoint may not have access.

lib/private/TaskProcessing/Manager.php
```diff
@@ -884,4 +897,69 @@ public function setTaskStatus(Task $task, int $status): void {
...
+					'status' => Task::statusToString($task->getStatus()),
```
marcelklehr, line 920: Can we set a timeout here for the request?

lib/private/TaskProcessing/Manager.php
```diff
@@ -668,17 +678,20 @@ public function setTaskResult(int $id, ?string $error, ?array $result): void {
...
 				$error = 'The task was processed successfully but the provider\'s output doesn\'t pass validation against the task type\'s outputShape spec and/or the provider\'s own optionalOutputShape spec';
 				$task->setErrorMessage($error);
+				$this->runWebhook($task);
 				$this->logger->error($error, ['exception' => $e]);
 			} catch (NotPermittedException $e) {
 				$task->setProgress(1);
 				$task->setStatus(Task::STATUS_FAILED);
 				$error = 'The task was processed successfully but storing the output in a file failed';
 				$task->setErrorMessage($error);
+				$this->runWebhook($task);
```
marcelklehr, line 694: Can we move this after the `update` call, so we don't delay the delivery to the user if any.

lib/private/TaskProcessing/Manager.php
julien-nc: Agreed. Done

lib/private/TaskProcessing/Manager.php
julien-nc: Agreed. Done

lib/private/TaskProcessing/Manager.php
julien-nc: Agreed. Done. It is cleaner as well.

marcelklehr (approved):

bigcat88 (approved):
LGTM

provokateurin:
I'm not sure I understand the purpose of these changes. Is this part of the new webhook stuff or specific to task processing api? If it is the former then it seems wrong to me to implement it in the task processing api? Maybe I'm missing something though.

julien-nc:
It seems like the 3rd party repo is not up to date in master. The `Block merging with outdated 3rdparty` action will be fixed when someone pushes the new subproject commit hash.

I don't really get why the OpenAPI action is failing though.

provokateurin:
The OpenAPI failure is random, it seems to happen once in a while without an apparent reason.

julien-nc:
@provokateurin If we think about the workflow engine, it would indeed be possible in a flow to:
* schedule a task, get its ID
* run an custom approval script to block the flow execution
* in the approval script, register a webhook for the "task finished" event with a filter on the task ID, setting the target URI to the resume endpoint of the approval script
* after the approval step is resumed, run a script that unregisters the webhook because we don't need it anymore

But it is way more simple to:
* run an approval script
* schedule a task in the approval script and tell the task processing API to call the resume endpoint when the task has finished

Having the webhook_listeners app does not prevent us from implementing webhook-like features in other places where it's convenient. That's the reasoning behind this PR.

PR #46583, opened 2024-07-17
Make filecache queries compatible with sharding

icewind1991:
Various tweaks to the filecache queries to make them work with the upcoming sharding work

- Use named parameters instead of positional ones
- Explicitly filter by storage id in more places
- Use parameters instead of literals
- Remove redundant join condition

### DIFF
--- a/lib/private/Files/Cache/QuerySearchHelper.php
+++ b/lib/private/Files/Cache/QuerySearchHelper.php
@@ -110,7 +110,6 @@ class QuerySearchHelper {
 		$query
 			->leftJoin('file', 'vcategory_to_object', 'tagmap', $query->expr()->eq('file.fileid', 'tagmap.objid'))
 			->leftJoin('tagmap', 'vcategory', 'tag', $query->expr()->andX(
-				$query->expr()->eq('tagmap.type', 'tag.type'),
 				$query->expr()->eq('tagmap.categoryid', 'tag.id'),
 				$query->expr()->eq('tag.type', $query->createNamedParameter('files')),
 				$query->expr()->eq('tag.uid', $query->createNamedParameter($user->getUID()))
--- a/tests/lib/Files/Config/UserMountCacheTest.php
+++ b/tests/lib/Files/Config/UserMountCacheTest.php
@@ -7,11 +7,13 @@
 
 namespace Test\Files\Config;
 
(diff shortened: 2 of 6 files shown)


tests/lib/Files/Config/UserMountCacheTest.php
```diff
@@ -335,29 +337,38 @@ private function sortMounts(&$mounts) {
...
-			return (int)$query->fetchOne();
+		} catch (DbalException $e) {
+			if ($e->getReason() === DbalException::REASON_UNIQUE_CONSTRAINT_VIOLATION) {
+				$query = $this->connection->getQueryBuilder();
+				$query->select('fileid')
+					->from('filecache')
+					->where($query->expr()->eq('storage', $query->createNamedParameter($storageId)))
+					->andWhere($query->expr()->eq('path_hash', $query->createNamedParameter(md5($internalPath))));
+				$id = (int)$query->execute()->fetchColumn();
+			} else {
```
ChristophWurst, line 369: could two requests race for the same INSERT? then you could have a dirty read here when the query goes to a read replica that is slightly behind …

icewind1991: it's a test case so no concurrency

ChristophWurst: Right :see_no_evil:

artonge (approved):
Actually, good anyway, it is just a question :)

lib/private/Files/Cache/QuerySearchHelper.php
```diff
@@ -110,7 +110,6 @@ protected function equipQueryForDavTags(CacheQueryBuilder $query, IUser $user):
 		$query
 			->leftJoin('file', 'vcategory_to_object', 'tagmap', $query->expr()->eq('file.fileid', 'tagmap.objid'))
 			->leftJoin('tagmap', 'vcategory', 'tag', $query->expr()->andX(
-				$query->expr()->eq('tagmap.type', 'tag.type'),
```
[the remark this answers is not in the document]
icewind1991: I guess, since `tag.id` is unique there shouldn't be a need to store it in the `tagmap`. Maybe it's done for easier filtering.

sorbaugh (approved):

ChristophWurst:
Yes? :)

PR #46596, opened 2024-07-18
feat: Navigate via folder tree

### LINKED ISSUE #15253: Implement a folder tree sidebar
Nextcloud displays files either in a grid or a list. However this is not convenient / suitable for quickly navigating complex folder trees and long lists of folders. A classic Windows Explorer style sidebar that displays folders as a tree would be a very nice improvement.

I've seen it was proposed in #4994 but the aspect of a tree view seems to be forgotten as the "moving files" feature seems like the main point of that request. I'd like to bring attention to accessibility / usability aspect of tree view.

Pytal:
## Summary

- Resolves https://github.com/nextcloud/server/issues/15253

image

- [x] Allow toggling folder tree

Split commits for easier reviewing :)

## Checklist


apps/files/src/services/FolderTree.ts
```diff
@@ -0,0 +1,77 @@
...
+	return `${folderTreeId}-${folder.fileid}`
+}
+
+export const getFolderTreeParentId = (folder: Folder): string => {
+	if (folder.dirname === '/') {
+		return folderTreeId
+	}
+	const mountType = folder.attributes['mount-type']
+	if (mountType !== '') { // If not local mount
+		return `${folderTreeId}-${mountType}-${folder.attributes.parentid}` // Include mount type as fileids may conflict across storage mounts
```
artonge, line 74: You could also have two files with the same id from the same kind of mount, no?

apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -62,6 +62,7 @@ class FilesPlugin extends ServerPlugin {
 	public const SUBFILE_COUNT_PROPERTYNAME = '{http://nextcloud.org/ns}contained-file-count';
 	public const FILE_METADATA_PREFIX = '{http://nextcloud.org/ns}metadata-';
 	public const HIDDEN_PROPERTYNAME = '{http://nextcloud.org/ns}hidden';
+	public const PARENTID_PROPERTYNAME = '{http://nextcloud.org/ns}parentid';
```
artonge, line 65: I don't understand why this is needed

apps/files/lib/Controller/ViewController.php
```diff
@@ -190,6 +223,7 @@ public function index($dir = '', $view = '', $fileid = null, $fileNotFound = fal
 		$this->initialState->provideInitialState('config', $this->userConfig->getConfigs());
 		$this->initialState->provideInitialState('viewConfigs', $this->viewConfig->getConfigs());
 		$this->initialState->provideInitialState('favoriteFolders', $favElements);
+		$this->initialState->provideInitialState('folderTreeFolders', $folderTreeFolders);
```
artonge, line 226: Could contain a very large list of folders, so would increase the response size. And most of the content won't be needed. Have you considered fetching the need folder list from the frontend?

apps/files/src/services/FolderTree.ts
Pytal: Good point, would mount id work?

apps/dav/lib/Connector/Sabre/FilesPlugin.php
Pytal: We need to provide the parentid and register it as a default dav property so when a node is created i.e. https://github.com/nextcloud/server/blob/d5bb37ab0d872fb4bdb1ec61809c575a46341a25/apps/files/src/services/DropServiceUtils.ts#L139 then the new child view shows up in the tree correctly nested under its parent

apps/files/lib/Controller/ViewController.php
Pytal: From my measurements on the impact on load time this is negligible. With a few folders the total resource transfer size (compressed) over the network is 4355kB bytes and with 1000 folders it is 4370kB

The time for retrieving the $folderTreeFolders data is 10ms and 50ms for 1000 folders in my testing

Pytal: Also better UX to have the tree available immediately instead of it popping up some time after the rest of the navigation is shown

skjnldsv: What about the backend computation time?
We have customers with 1M+ folders, this makes it synchronous. I'm afraid of the performances impact

apps/dav/lib/Connector/Sabre/FilesPlugin.php
skjnldsv: Can't we just use the path to create the tree?
Do we actually need any ID at all?

susnux: Yes especially as files no longer "rely" on fileid but the source, we could simply build the tree using the node source property

skjnldsv (changes requested):
Blocking, some of the comments are too important imho

susnux (commented):
I am not sure computing the whole list on the backend is a good idea, especially on every request.
Especially as we can have the views collapsed we could instead just load the folders needed for the currently expanded view list

apps/files/src/components/FileEntry/FileEntryName.vue
```diff
@@ -305,8 +305,10 @@ export default defineComponent({
 				})
 
 				// Success 🎉
-				emit('files:node:updated', this.source)
-				emit('files:node:renamed', this.source)
+				const node = this.source
+				node.attributes['old-name'] = oldName
+				emit('files:node:updated', node)
+				emit('files:node:renamed', node)
```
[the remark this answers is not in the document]
Pytal: Will change to `{ newName: string, oldName: string }` 👍

apps/files/lib/Controller/ViewController.php
```diff
@@ -160,6 +164,35 @@ public function index($dir = '', $view = '', $fileid = null, $fileNotFound = fal
...
+			$searchQuery = new SearchQuery(
+				new SearchComparison(ISearchComparison::COMPARE_EQUAL, 'mimetype', ICacheEntry::DIRECTORY_MIMETYPE),
+				0,
+				0,
+				[],
+				$this->userSession->getUser(),
+				false,
+			);
+			/** @var Folder[] $folders */
+			$folders = $userFolder->search($searchQuery);
```
Pytal, line 178: @icewind1991 do you know of a trick to generate a list of deepest paths from a user folder?

apps/files/src/services/FolderTree.ts
Pytal: Changing to use source as discussed :)

apps/dav/lib/Connector/Sabre/FilesPlugin.php
Pytal: Parent id removed

apps/files/lib/Controller/ViewController.php
skjnldsv: What about just implementing this method?

skjnldsv: `apps/files/lib/Controller/ApiController.php`
```php
	/**
	 * Returns a folder tree of the user's files.
	 *
	 * @NoAdminRequired
	 *
	 * @return JSONResponse
	 */
	public function getFolderTree($maxDepth = 10): JSONResponse {
		$depth = 0;
		$nodes = $this->rootFolder->getUserFolder($this->userSession->getUser()->getUID())->getDirectoryListing();

		/**
		 * @param \OCP\Files\Node[] $nodes
		 * @param int $depth
		 * @param int $maxDepth
		 */
		function getChildren(Array $nodes, $depth = 0, $maxDepth = 10): array {
			if ($depth >= $maxDepth) {
				return [];
			}

			$children = [];
			foreach ($nodes as $node) {
				if (!($node instanceof Folder)) {
					continue;
				}

				$children[] = [
					'basename' => basename($node->getPath()),
					'name' => $node->getName(),
					'children' => getChildren($node->getDirectoryListing(), $depth + 1, $maxDepth),
				];
			}
			return $children;
		}

		return new JSONResponse(getChildren($nodes, $depth, $maxDepth));
	}
```

With 2500 folders, a max depth of 10 takes 650ms, which is a lot!
A depth of 5 takes 350ms

skjnldsv: I tried the Search method and it was over 4 times slower (>2.5 seconds)

skjnldsv: Ok, quick brain thrust:
1. Load the page as master currently does
2. Async fetch the file tree and render it on the sidebar
3. Add a user config to disable the file tree

icewind1991: Having a limited-depth endpoint with possible requesting deeper items when navigating the tree seems like the only option that isn't guaranteed to explode on scalability.

apps/files/lib/Controller/ViewController.php
Pytal: > What about the backend computation time?

Posted above

> 10ms for a few folders and 50ms for 1000 folders

Need to run some more benchmarks

apps/files/lib/Controller/ViewController.php
Pytal: Will need to run some benchmarks

How did you measure the times @skjnldsv? I will run the same on my system

For the 50ms 1000 folders measurement search I used microtime, and that includes calling $folder getters as well not just the search

skjnldsv: > How did you measure the times @skjnldsv? I will run the same on my system

Called on ocs api endpoint with fetch. It's not php compute, but real request times

Pytal: > I tried the Search method to compute the same type of array and it was over 4 times slower (>2.5 seconds)

Tested full request times and getting the opposite with search being 4 times faster, weird... 🤔

icewind1991: which one is faster will be highly dependent on the instance.

Given the performance tradeoffs, my approach would be to have the client only request a single level of the folder structure at a time.

While this does mean that there is a bit of loading every time a folder is expanded, it's the only option that wouldn't cause performance issues imo.

icewind1991 (dismissed):
(Leaving this as a review to note that I find the performance questions/discussion to be blocking)

apps/files/lib/Controller/ViewController.php
skjnldsv: @icewind1991 would have been nice sure, but the specifications seems to not have been thoroughly thought with performance in mind. The entire tree is always displayed, there is currently no expectations on when to stop from the initial feature planning (Maybe it's time to think this through now before it bites us later)

Pytal: By default the expanded state of each folder is persisted so if 20 folders are expanded then on first load this would mean making 20 separate requests

How about not storing the expanded state of folder tree views specifically then a request is made only when a folder is expanded on click?

Pytal: Any of these edge cases can be addressed in separate PRs as we should get a first version in that works for the average instance

sorbaugh: I think we should really tackle the performance issues but would also do so during the rest of NC30 beta in separate PRs as Pytal suggested to have a working version. 

That means proceeding to merge as soon as we have a stable version.

This way we can already get some field testing and even confirm/disconfirm the concerns. A win-win in my opinion.

apps/files/lib/Controller/ViewController.php
Pytal: Fetching from the frontend now 👍

Pytal:
@provokateurin the openapi script crashes, maybe you know how to fix it?

apps/files/src/components/FileEntry/FileEntryName.vue
```diff
@@ -275,6 +276,10 @@ export default defineComponent({
 				// Success 🎉
 				emit('files:node:updated', this.source)
 				emit('files:node:renamed', this.source)
+				emit('files:node:moved', {
+					node: this.source,
+					oldSource: `${dirname(this.source.source)}/${oldName}`,
```
skjnldsv, line 281: what about just `new`/`old` ?
Or `newSource`/`oldSource` to stay consistent?

apps/files/src/store/viewConfig.ts
```diff
@@ -44,9 +44,10 @@ export const useViewConfigStore = function(...args) {
 			 * @param value
 			 */
 			async update(view: ViewId, key: string, value: string | number | boolean) {
-				axios.put(generateUrl(`/apps/files/api/v1/views/${view}/${key}`), {
+				const params = new URLSearchParams({ view, key })
+				axios.put(generateUrl('/apps/files/api/v1/views'), {
 					value,
-				})
+				}, { params })
```
skjnldsv, line 50: why?

skjnldsv (approved):
Comments to be addresses later :+1:

sorbaugh (approved):

apps/files/src/store/viewConfig.ts
Pytal: Needed as the source is used as the view id but was thinking in GET, adjusted to send view and key as request data now

apps/files/src/components/FileEntry/FileEntryName.vue
Pytal: Will do in a followup :)

skjnldsv:
Congrats :tada: :tada: :tada:

nickvergessen:
The `#[ApiRoute(verb: 'GET', url: '/api/v1/folder-tree')]` takes down our server and makes browsers freeze.
Had to shortcut it with:
```
		return new JSONResponse([], Http::STATUS_INTERNAL_SERVER_ERROR, [], JSON_FORCE_OBJECT);
```

frake65:
> The `#[ApiRoute(verb: 'GET', url: '/api/v1/folder-tree')]` takes down our server and makes browsers freeze. Had to shortcut it with:
> 
> ```
> 		return new JSONResponse([], Http::STATUS_INTERNAL_SERVER_ERROR, [], JSON_FORCE_OBJECT);
> ```

So this works for me. - What exactly should the code in function "getFolderTree()" be doing?

Pytal:
We are working on optimizing this feature @frake65, though the page might be too unresponsive with very large trees there is an "Enable folder tree" checkbox located in the Files settings dialog which can be toggled per-user and will prevent the endpoint from being called completely

frake65:
Have removed the shortcut and am now using the toggle.
Sorry to say, but this implementation seems to be still buggy to me - not only slow:
From a users view point at this time I get a file tree, where folder names do not match the folders that they link to at all, but have random names (from sowhere within the file tree).
What should be the order of the folder names by the way? - Alphabetically would be my first guess.

In my opinion probably this feature should be postponed

PR #46606, opened 2024-07-18
feat(Security): Warn about using annotations instead of attributes

provokateurin:
## Summary

Using annotation is a lot less safe since there is no checking that they are correctly set. For example a typo can make the annotation in-effective and thus lead to security problems. Same goes for giving arguments to the annotations, since they can not be type checked or any other type of check other than at runtime.
In contrast to that the attributes provide more safety and have been available since PHP 8.0 and thus can be used by almost every app developer out there. Adding these debug logs hopefully encourage developers to migrate to the attributes.

## Checklist

### DIFF
--- a/lib/private/AppFramework/DependencyInjection/DIContainer.php
+++ b/lib/private/AppFramework/DependencyInjection/DIContainer.php
@@ -207,7 +207,8 @@ class DIContainer extends SimpleContainer implements IAppContainer {
 					$c->get(IRequest::class),
 					$c->get(IControllerMethodReflector::class),
 					$c->get(IUserSession::class),
-					$c->get(IThrottler::class)
+					$c->get(IThrottler::class),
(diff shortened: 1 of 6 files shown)


nickvergessen (approved):
This will spam so hard :-X

provokateurin:
I already have a branch where I migrate all annotations in server (only partially done so far), does it makes sense to send a PR for that or let the app maintainers do it gradually? I fear such a big PR is not easy to review and any mistakes can be easily overlooked.

blizzz:
> This will spam so hard :-X

Good medicine has to be bitter.

blizzz (approved):

nickvergessen:
I would do server in one PR yeah.
Psalm should cover missing imports etc. and the rest should be reviewed by the reviewers and the author of the PR

PR #46615, opened 2024-07-18
feat: allow for ExApps to use webhook_listeners without user context

bigcat88:
## Summary

For the Workflow Engine project, we need to register webhooks from the background job, where we do not have a userId.

This will also be useful in other ExApps in the future.

This special attribute that allows you to do this was added today:

https://github.com/nextcloud/server/pull/46607

## Checklist

### DIFF
--- a/apps/webhook_listeners/appinfo/info.xml
+++ b/apps/webhook_listeners/appinfo/info.xml
@@ -9,7 +9,7 @@
 	<name>Nextcloud webhook support</name>
 	<summary>Nextcloud webhook support</summary>
 	<description>Nextcloud webhook support</description>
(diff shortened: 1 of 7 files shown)


AndyScherzinger (approved):

julien-nc (changes requested):
You need to change the app version to make sure the migration step runs. Apart from that: LGTM :+1:

bigcat88:
> You need to change the app version to make sure the migration step runs. Apart from that: LGTM 👍

version changed from `1.0.0-dev` to `1.1.0-dev`

julien-nc (approved):

PR #46618, opened 2024-07-18
feat(core): Expose app versions, API versions and features supported by the instance

provokateurin:
## Summary

The idea behind this is to have a standardized way for apps to declare their API versions and supported features. Many apps with an externally accessible API already do this to some degree using capabilities or dedicated endpoints, but all of them do it differently.

This API makes it easier for clients to figure out if they support the software running on the server, either by looking at the app version, the API versions or the list of features. The features list is not a concept that is widely used, but Talk has this as well and in my opinion this is the best way to check if a certain operation is supported on the server, since the app doesn't need to compare the app or API versions which prevents mistakes and possibly extends the range of supported versions if not all features are required.

To show it works I added the respective entries for the dashboard app, but ideally we'd add it for all apps with an API.

## Checklist

### DIFF
--- a/core/openapi-full.json
+++ b/core/openapi-full.json
@@ -2724,7 +2724,9 @@
                                                     "type": "object",
                                                     "required": [
                                                         "version",
-                                                        "capabilities"
+                                                        "capabilities",
+                                                        "features",
+                                                        "apps"
                                                     ],
                                                     "properties": {
                                                         "version": {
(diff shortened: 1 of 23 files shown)


nickvergessen:
From my understanding and POV that is exactly why we have the capabilities. The server's capabilities show what the server and apps provide and make available? 😅

provokateurin:
Sure, we can also call this capabilities v2 or something like that. The benefit of this is that it is standardized and easier to implement, while "limiting" the developers by not allowing any free form data.

nickvergessen:
And then clients have to fetch 2 endpoints to get the list of features and configs?

We can standardize the format a little bit in addition, but we need what currently works.

provokateurin:
My perspective for this is that this new endpoint is going to be used only for versioning stuff and the capabilities no longer need to expose this information. Then at some point it can be removed and the clients have one endpoint for getting the versions they want to know about (new endpoint) and another one for just getting config stuff (old capabilities endpoint).
For some time this leads to duplication, but ideally after enough time we have a much cleaner and easier to use API than before. Capabilities are still useful since many apps expose config values through them and it is impossible to enforce a specific format for that.

provokateurin:
The alternative would be to just add this data to the existing capabilities response, then clients don't need to call separate endpoints and use the new data more easily. That would probably be preferable, would that be fine for you?

nickvergessen:
The other thing I would prefer to change is providing the data with PHP instead of the info.xml to allow exposing features only sometimes. Does happen often, but in Talk we have 2 features that are only available with some specific server configuration
https://github.com/nextcloud/spreed/blob/main/lib/Capabilities.php#L223

provokateurin:
That was also something I had in mind, it could be a new optional interface that can be implemented by the capabilities classes.

core/openapi-full.json
```diff
@@ -2724,7 +2724,9 @@
                                                     "type": "object",
                                                     "required": [
                                                         "version",
-                                                        "capabilities"
+                                                        "capabilities",
+                                                        "features",
+                                                        "apps"
```
come-nc, line 2729: What’s the difference between features and capabilities?

provokateurin: The idea is to only have a list of strings each app can declare. This avoids the free-form data structures that are possible with capabilities and allows clients to easily check which features are supported by an app and which are not. It's similar to the way it is done in Talk: https://github.com/nextcloud/spreed/blob/main/lib/Capabilities.php

PR #46661, opened 2024-07-20
fix(caldav): added localization hints for translators

SebastianKrupinski:
* Resolves: #46632

### DIFF
--- a/apps/dav/lib/CalDAV/Schedule/IMipService.php
+++ b/apps/dav/lib/CalDAV/Schedule/IMipService.php
@@ -209,6 +209,13 @@ class IMipService {
 			$endTime = $this->l10n->l('time', $er->endDateTime(), ['width' => 'short']) . ' (' . $er->endTimeZone()->getName() . ')';
 		}
 		// generate localized when string
+		// TRANSLATORS
+		// Indicates when a calendar event will happen, shown on invitation emails
+		// Output produced in order:
+		// In a day/week/month/year on July 1, 2024 for the entire day
+		// In a day/week/month/year on July 1, 2024 between 8:00 AM - 9:00 AM (America/Toronto)
+		// In 2 days/weeks/monthss/years on July 1, 2024 for the entire day
+		// In 2 days/weeks/monthss/years on July 1, 2024 between 8:00 AM - 9:00 AM (America/Toronto)
 		return match ([($occuring[0] > 1), !empty($endTime)]) {
 			[false, false] => $this->l10n->t('In a %1$s on %2$s for the entire day', [$occuring[1], $startDate]),
 			[false, true] => $this->l10n->t('In a %1$s on %2$s between %3$s - %4$s', [$occuring[1], $startDate, $startTime, $endTime]),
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):

rakekniven (commented):
Great.
Actually I am not sure if multi line comments will show up correctly at Transifex.
But let us give it a try.

rakekniven (approved):

SebastianKrupinski:
@rakekniven

Hi,

I am not on Transifex, can you DM on your instance or should I setup an account?

keunes:
Hi @SebastianKrupinski,

I'm currently translating the DAV component to Dutch and have a couple of questions and thoughts.

**First:**

https://github.com/nextcloud/server/blob/53028d845a26a71e56adcff20635aab5560eb856/apps/dav/lib/CalDAV/Schedule/IMipService.php#L655

https://app.transifex.com/nextcloud/nextcloud/translate/#nl/dav/568171276

In one / n minute(s), or in one or more hour would always be today  - or tomorrow, if you're close to midnight. Either way, the starting time is so close that printing the actual date here feels a bit odd. I would propose:
```diff
- // In a minute/hour on July 1, 2024
+ // In a minute/hour
- // In a minute/hour on July 1, 2024 then on July 3, 2024 and July 5, 2024
+ // In a minute/hour, then on July 3, 2024 and July 5, 2024
```

**Second:**

https://github.com/nextcloud/server/blob/53028d845a26a71e56adcff20635aab5560eb856/apps/dav/lib/CalDAV/Schedule/IMipService.php#L1284

https://app.transifex.com/nextcloud/nextcloud/translate/#nl/dav/507691943

'Last' is clear, and 'Second last' I read as the time before the last; e.g. not the last Wednesday of the year but the Wednesday before that. However, what does 'third/fourth/fifth last' mean?

Many thanks for your help!

CC: @rakekniven

SebastianKrupinski:
Good Morning,

> In one / n minute(s), or in one or more hour would always be today - or tomorrow, if you're close to midnight. Either way, the starting time is so close that printing the actual date here feels a bit odd. I would propose:

```
Occurring:
In 10 hours on July 10, 2025 then on July 17, 2025 and July 24, 2025
```

You are correct it is a bit odd, but when you factory in time zone(s) off sender and receiver, I thought it might be more clear to have a date, this way is shows the senders date, as we don't know what day this might be for the receiver.

> 'Last' is clear, and 'Second last' I read as the time before the last; e.g. not the last Wednesday of the year but the Wednesday before that. However, what does 'third/fourth/fifth last' mean?

This refers to the relative offset of a event, in a month. For instance the "Third Friday of July" this would be relative to the start of the month, but "Third last Friday of July", would be relative to the end of the month, which technically is the 2nd or 3rd week depending on the month.  "Third last Friday of July" is the 2nd week but "Third last Friday of August" is the 3rd week.

Hope that helps

keunes:
Thanks @SebastianKrupinski!

> You are correct it is a bit odd, but when you factory in time zone(s) off sender and receiver, I thought it might be more clear to have a date

Right, that makes sense.

Do I understand correctly then that these messages will end up in emails that are sent to invitees? In that case I don't think it makes sense to have the relative time at all, to be honest. You'd never know when the recipient will read the message, and relative time is very likely to be useless (it'll only be useful if they read it the second it's received).

Having the start time would be more helpful I think:
```
Occurring:
At 19:00 CEST on 10 July 2025, then on 17 July 2025 and 24 July 2025
```

> For instance the "Third Friday of July" this would be relative to the start of the month, but "Third last Friday of July", would be relative to the end of the month, which technically is the 2nd or 3rd week depending on the month.

Right. In daily (work)life, who does that? :P Apparently it's not even possible in Nextcloud Calendar (which you undoubtedly know already :-) ):
image

A translation of 'fifth to last' would also feel odd. ('Last' is 'laatste' and 'Second to last' is 'voorlaatste'; 'fifth to last' would be 'de vijfde voor de laatste'). But I guess you'll want to support it anyway as this might come from other servers - I'll get to it then.

PR #46691, opened 2024-07-22
feat(sidebar): Show node owner in metadata subline 

### LINKED ISSUE #46178: [App Files] Can't see owner's name if a file/folder is reshared twice
# Reproduction
**Reproducible ?**
- [x] Yes.
- [ ] No.

**Regularity ?**  
- [x] Always.
- [ ] Sometimes.
- [ ] Unable to reproduce.

**Regression ?**
- [ ] Yes.
- [ ] No.
- [x] New Feature.
 
**Steps to reproduce ?**

1. Go to web app via a main account (for the test: admin)
2. Go to app files
3. Do a collab sharing on a file/folder to a 2nd user (for the test: John testing)
4. Connect on the shared user on web app
5. Do a collab sharing to a 3rd user on the shared file/folder
6. Check on the 3rd user who shared the file/folder

**Obtained results:** Shows that the file/folder is shared by the 2nd user but we can only see who is the file/folder's owner in the bubble in the file view

Image

# Environment and versions
| 26.0.4.3  |  

Browser used:
- [x] Chrome Windows
- [x] Firefox Windows

nfebe:
## Summary

Show user bubble to indicate owner of file in sidebar overview (this should not change irrespective of share status)... 

Resolves: https://github.com/nextcloud/server/issues/46178

## Screenshots
| Timeline | State |
|----------| -------------| 
| Before|metadata-before|
| Now (owned by you) |metadata-owned-by-you |
| Now (owned by other) |metadata-owned-by-other|

### TODO
- [x] Finish wip
- [x] Screenshots


apps/files/src/views/Sidebar.vue
```diff
@@ -549,6 +574,12 @@ export default {
 		handleWindowResize() {
 			this.hasLowHeight = document.documentElement.clientHeight < 1024
 		},
+		sharedMultipleTimes() {
+			// Mixed share types
+			if (Array.isArray(this.node.attributes?.['share-types']) && this.node.attributes?.['share-types'].length > 1) {
+				return t('files', 'Shared multiple times with different people')
+			}
+		},
```
Fenn-CS, line 582: Probably useful to provide this information somewhere in the side?

- Maybe  a property to check if it has been shared multiple times
- Another property for the actual text.

Fenn-CS:
### Show original file owner in side bar

Now that some logic to show the owner is in place with a simple "draft" to show this information with the userbubble near other folder metada can @nextcloud/designers chime in?


show-admin

nimishavijay:
Nice! Good idea using the user bubble component. I would suggest we include only the username in the user bubble, and to separate the different metadata we could use a middle dot `·`. So it would look like

```
11.2 MB · 3 weeks ago · Owned by (👱🏼 Alex Doe)
```

Also the user bubble component should be center aligned with the rest of the text :)

sorbaugh (approved):

apps/files/src/views/Sidebar.vue
```diff
@@ -17,14 +17,20 @@
...
+			<div class="sidebar__subname">
+				<NcIconSvgWrapper v-if="fileInfo.isFavourited"
+					:path="mdiStar"
+					:name="t('files', 'Favorite')"
+					inline />
+				<span>{{ size }}</span>
+				<span class="sidebar__subname-separator">.</span>
+				<NcDateTime :timestamp="fileInfo.mtime" />
+				<span class="sidebar__subname-separator">.</span>
+				<span>{{ t('files', 'Owned by') }}</span>
```
nickvergessen, line 29: string concatenation is bad for some languages, see https://docs.nextcloud.com/server/latest/developer_manual/basics/front-end/l10n.html#improving-your-translations

apps/files/src/views/Sidebar.vue
```diff
@@ -288,6 +295,25 @@ export default {
...
+		},
+		currentUserIsOwner() {
+			return this.ownerId === this.currentUser.uid
+		},
+		nodeOwnerLabel() {
+			if (this.currentUserIsOwner) {
+				return t('files', 'you')
+			}
+			const ownerDisplayName = this.node?.attributes?.['owner-display-name']
+			return ownerDisplayName
```
nickvergessen, line 309: can you test with external storage and groupfolders? I think you end up with an "Owned by" without a name at the moment?

Fenn-CS: Tested with group folders and it works fine and ti should because the metadata about ownership which is an attribute is "calculated" and sent by the backend irrespective of source or said in another way even in case of external storage, the files and related tables would still hold information about the object but the actual object would come from wherever?

Screenshot from 2024-09-03 11-44-04


Shields

- If the attribute is not available it would show, "Owned by unknown" for now

apps/files/src/views/Sidebar.vue
Fenn-CS: Used to be a feature in previous sharing. cc @jancborchardt

apps/files/src/views/Sidebar.vue
```diff
@@ -17,14 +17,20 @@
...
-			<NcDateTime :timestamp="fileInfo.mtime" />
+			<div class="sidebar__subname">
+				<NcIconSvgWrapper v-if="fileInfo.isFavourited"
+					:path="mdiStar"
+					:name="t('files', 'Favorite')"
+					inline />
+				<span>{{ size }}</span>
+				<span class="sidebar__subname-separator">.</span>
+				<NcDateTime :timestamp="fileInfo.mtime" />
+				<span class="sidebar__subname-separator">.</span>
```
susnux, line 28: Should probably use the same separator as versions

```suggestion
				<span class="sidebar__subname-separator">•</span>
				<NcDateTime :timestamp="fileInfo.mtime" />
				<span class="sidebar__subname-separator">•</span>
```

apps/files/src/views/Sidebar.vue
```diff
@@ -140,12 +145,14 @@ export default {
 	data() {
 		return {
 			// reactive state
+			currentUser: getCurrentUser(),
```
susnux, line 148: Why this change? This is non reactive (the user will not appear or disappear) and should not made reactive (which happens when passing to the data).

And you can access it exactly the same in the computed below (`this.currentUser`)

apps/files/src/views/Sidebar.vue
jancborchardt: @Fenn-CS if the string concat is not proper, let’s instead do it simpler like `Owner: (avatarbubble)`
Also, make sure to use [middle dots](https://en.wikipedia.org/wiki/Interpunct) to separate the size, date and ownership info: `·`, not regular dots. :)

Fenn-CS: ## Screenshots

| Owned by another | Owned by current user |
| --------------------------- | ---------------------------------- |
| Screenshot from 2024-09-17 15-36-27 | Screenshot from 2024-09-17 15-35-26 |

apps/files/src/views/Sidebar.vue
```diff
@@ -460,6 +486,7 @@ export default {
 				this.fileInfo = await FileInfo(this.davPath)
 				// adding this as fallback because other apps expect it
 				this.fileInfo.dir = this.file.split('/').slice(0, -1).join('/')
+				this.node = await fetchNode({ path: (this.fileInfo.path + '/' + this.fileInfo.name).replace('//', '/') })
```
artonge, line 489: Aren't we fetching the file info already somewhere?

apps/files/src/views/Sidebar.vue
```diff
@@ -123,15 +131,14 @@ export default {
 		NcIconSvgWrapper,
 		SidebarTab,
 		SystemTags,
+		NcUserBubble,
 	},
 
 	setup() {
 		const currentUser = getCurrentUser()
-
```
artonge, line 130: Unnecessary change. Same for other whitespaces.

apps/files/src/views/Sidebar.vue
Fenn-CS: I doubt in the same way we need it here, @skjnldsv is there somewhere else I can pick this up?

sorbaugh (approved):

apps/files/src/views/Sidebar.vue
susnux: As the sidebar is part of the files app, you can use the files store to retirieve the current node without needing to re-fetch it again

Fenn-CS: Using the fileStore as at now breaks the sidebar. With undefined errors.... Looks like the sidebar is trying to reference the fileStore before it is ready.

apps/files/src/views/Sidebar.vue
```diff
@@ -104,13 +112,15 @@ import NcActionButton from '@nextcloud/vue/dist/Components/NcActionButton.js'
...
 import NcIconSvgWrapper from '@nextcloud/vue/dist/Components/NcIconSvgWrapper.js'
+import NcUserBubble from '@nextcloud/vue/dist/Components/NcUserBubble.js'
 
 import FileInfo from '../services/FileInfo.js'
 import LegacyView from '../components/LegacyView.vue'
 import SidebarTab from '../components/SidebarTab.vue'
 import SystemTags from '../../../systemtags/src/components/SystemTags.vue'
 import logger from '../logger.ts'
 
+
```
susnux, line 123:
```suggestion
```

susnux: make eslint happy

apps/files/src/views/Sidebar.vue
susnux: Yes I think the sidebar could be loaded before the files app, meaning Pinia is not yet created... A chick-egg problem we could solve but is unrelated here.
Would need to refactor all of the files stores...

susnux (approved):
We need to migrate the Sidebar to use proper Node data instead of legacy stuff, but until then this is probably the best way without patching legacy OC.Files.Client

PR #46693, opened 2024-07-22
feat: add a specialized writeStream implementation for s3 external storage

icewind1991:
This should both reduce overhead and improve detection of failed writes

### DIFF
--- a/apps/files_external/lib/Lib/Storage/AmazonS3.php
+++ b/apps/files_external/lib/Lib/Storage/AmazonS3.php
@@ -8,6 +8,7 @@ namespace OCA\Files_External\Lib\Storage;
 
 use Aws\S3\Exception\S3Exception;
 use Icewind\Streams\CallbackWrapper;
+use Icewind\Streams\CountWrapper;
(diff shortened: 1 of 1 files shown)


apps/files_external/lib/Lib/Storage/AmazonS3.php
```diff
@@ -754,4 +755,23 @@ public function hasUpdated($path, $time) {
 			return true;
 		}
 	}
+
+	public function writeStream(string $path, $stream, ?int $size = null): int {
+		if ($size === null) {
+			$size = 0;
+			$stream = CountWrapper::wrap($stream, function (int $writtenSize) use (&$size) {
+				$size = $writtenSize;
+			});
```
artonge, line 764: Is it a way to get the final written size? This is triggered on `writeObject`, right? Can you add a comment explaining it?

icewind1991: done

artonge (approved):

sorbaugh (approved):

PR #46761, opened 2024-07-25
fix(core): Document CSRF token endpoint

provokateurin:
## Summary

Exposes the endpoint to OpenAPI as it is necessary to receive a token for calling CSRF-protected endpoints.

## Checklist


bigcat88 (approved):

susnux (approved):
```diff
- fix(core): Document CSRF token endpoint 
+ docs(core): Document CSRF token endpoint
```

provokateurin:
In general yes, but in this case I think it was a bug in the code because it should have not been ignored :D

PR #46779, opened 2024-07-26
fix(testing): Make Testing TextProcessing providers unicode safe

marcelklehr:
## Summary
The Fake TextProcessing providers implemented in the testing app would fail when used in the TaskProcessing API because they do not produce valid unicode. This PR fixes that.

## Checklist

### DIFF
--- a/apps/testing/lib/Provider/FakeTextProcessingProvider.php
+++ b/apps/testing/lib/Provider/FakeTextProcessingProvider.php
@@ -18,10 +18,21 @@ class FakeTextProcessingProvider implements IProvider {
 	}
 
 	public function process(string $prompt): string {
-		return strrev($prompt) . ' (done with FakeTextProcessingProvider)';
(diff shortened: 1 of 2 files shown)


apps/testing/lib/Provider/FakeTextProcessingProvider.php
```diff
@@ -18,10 +18,24 @@ public function getName(): string {
...
+	 * Reverse a miltibyte string.
+	 *
+	 * @param string $string The string to be reversed.
+	 * @param string|null $encoding The character encoding. If it is omitted, the internal character encoding value
+	 *     will be used.
+	 * @return string The reversed string
+	 */
+	private function mb_strrev(string $string, string $encoding = null): string
+	{
+		$chars = mb_str_split($string, 1, $encoding ?: mb_internal_encoding());
```
kyteinsky, line 38: $encoding can remain null. It will use the internal encoding.

kyteinsky (approved):

skjnldsv (approved):

PR #46782, opened 2024-07-26
fix(files): Ensure scrolled file list does not leak through file filters

### LINKED ISSUE #46772: File filter has no background when sticky during scroll
- Have a long file list
- Scroll down

As seen on daily:

Screenshot 2024-07-26 at 10 19 32

susnux:
* Resolves: https://github.com/nextcloud/server/issues/46772

## Summary

1. when scrolling the file list should not be visible behind the filters
2. on mobile we need to remove the padding of the filters to ensure they are accessible

[Bildschirmaufnahme_20240726_132643.webm](https://github.com/user-attachments/assets/75a3b9b1-1770-4275-a016-bc1044919b9f)

## Checklist


skjnldsv (approved):

juliushaertl:
Would be nice to address the cut off select input as well ;) 

Also seems to happen when opening it:
Screenshot 2024-07-26 at 14 50 41

susnux:
> Would be nice to address the cut off select input as well ;)

Unrelated to this and seems like a component issue :)

juliushaertl:
I'm not sure where the of users is originating from but 99dcbxiv seems to be what is assumed as a user id on daily for me causing this one broken entry.

juliusknorr (approved):

susnux:
> I'm not sure where the of users is originating

@juliushaertl two sources:
1. `owner` property of the node
2. `sharee` of the node 

I think there might be a problem if 2 is a group(?) I will investigate this.

PR #46803, opened 2024-07-27
chore: bump cypress workers to 8

skjnldsv:
With added cypress tests over the last weeks, this should speed things a bit :)


szaimen (approved):

solracsf:
Any reason why jobs aren't load balanced?
https://github.com/cypress-io/github-action?tab=readme-ov-file#parallel

skjnldsv:
@solracsf because cypress reduced their number of runs per months for open-source plans. We're exceeding them every time. So we don't rely on cypress dashboard anymore, this is too expensive

solracsf (approved):

PR #46840, opened 2024-07-29
fix(files): Don't try to use missing cache entries

skazi0:
## Summary
The `Cache::get()` function can return `ICacheEntry` or `false`. Added check ensures that `false` return value is not used as an array. This silences: `Trying to access array offset on value of type bool` errors.


skazi0:
Fixed in https://github.com/nextcloud/server/pull/50436

PR #46876, opened 2024-07-30
Update line height

marcoambrosini:
Since we have different font sizes, we should make the line height
dependent on the font size and not a fixed value. The recommended
value for accessibility is 1.5em.
https://www.w3.org/WAI/WCAG21/Understanding/text-spacing.html


apps/theming/css/default.css
```diff
@@ -50,7 +50,7 @@
   --font-face: system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
   --default-font-size: 15px;
```
Antreesy, line 51: Maybe add a small font size, as you're touching it?
```suggestion
  --default-font-size: 15px;
  --font-size-small: 13px;
```

marcoambrosini: Was going to do it in a separate pr, but ok

apps/theming/css/default.css
```diff
@@ -50,7 +50,7 @@
...
   /* 1.5 x font-size for accessibility */
-  --default-line-height: 24px;
+  --default-line-height: 1.5em;
```
ShGKme, line 53: Setting `em` as line-height is not correct. This is an absolute value which means "1.5 * this element font-size". That means if you have 1.5em on a 15px parent, then all the children have 1.5 * 15 = 22.5px line height. Not matter what font-size they have, it's exactly 22.5px even for 100px and 10px font-size.

Only using unitless multiplier means "x1.5 of font-size", so children with 100px font size will have 150px line height, and 10px font-size will have 15px line height.

```suggestion
  --default-line-height: 1.5;
```

ShGKme (changes requested):

apps/theming/css/default.css
ShGKme: Agree of creating PR to no mix creating a new variable (feature) and fixing line-height (design change).

apps/theming/css/default.css
ShGKme: Example: https://codepen.io/ShGKme/pen/abgpYKB

ShGKme:
A bit sad part is that 15 doesn't scale well in int numbers... It makes a text element having 22.5 height, which may result in some side effects. (It is different from 22 or 23 px)

Having also grid baseline 16px, shouldn't we make font-size also 16, so it multiplies by 1.5 well, and lies on `4px * n` grid including 1.5 * 16 = 24px?

ShGKme (approved):

Antreesy (approved):

PR #46887, opened 2024-07-30
fix(files_versions): Catch constraint error on version insertion

come-nc:
* Resolves: # 

## Summary

Catch errors when inserting a version with same fileid and timestamp as the previous one, and retry with incremented timestamp.

## Checklist


solracsf (approved):

solracsf:
A similar logic could be applied to Groupfolders? I can observe these Constraints Violations using Groupfolders too.

AndyScherzinger (approved):

artonge:
@come-nc This solution seems to cause a new issue: https://github.com/nextcloud/server/issues/53151
Looking at it, I am also not sure about the rational. If a version is created, then why create an entity with a different timestamp?

PR #46924, opened 2024-07-31
fix(files_sharing): filter out invalid shareWith in sharee API

skjnldsv:
Prevent this:
image

image

### DIFF
--- a/apps/files_sharing/lib/Controller/ShareesAPIController.php
+++ b/apps/files_sharing/lib/Controller/ShareesAPIController.php
@@ -257,7 +257,7 @@ class ShareesAPIController extends OCSController {
 			$sharees = $this->getAllShareesByType($user, $shareType);
 			$shareTypeResults = [];
 			foreach ($sharees as [$sharee, $displayname]) {
-				if (!isset($this->searchResultTypeMap[$shareType])) {
+				if (!isset($this->searchResultTypeMap[$shareType]) || trim($sharee) === '') {
 					continue;
 				}
 


AndyScherzinger (approved):
LGTM 👍

apps/files_sharing/lib/Controller/ShareesAPIController.php
```diff
@@ -257,7 +257,7 @@ private function getAllSharees(string $user, array $shareTypes): ISearchResult {
 			$sharees = $this->getAllShareesByType($user, $shareType);
 			$shareTypeResults = [];
 			foreach ($sharees as [$sharee, $displayname]) {
-				if (!isset($this->searchResultTypeMap[$shareType])) {
+				if (!isset($this->searchResultTypeMap[$shareType]) || empty($sharee)) {
```
come-nc, line 260: Will break on valid `'0'` value I think, as it’s a valid user id but return true on empty check.
```suggestion
				if (!isset($this->searchResultTypeMap[$shareType]) || ($sharee === '')) {
```

come-nc (dismissed):

apps/files_sharing/lib/Controller/ShareesAPIController.php
come-nc: (do we know why we have shares with empty `getSharedWith` return? Is that normal behavior?)

skjnldsv: yes, it can happens for link shares for example :)
Or mail shares without  recipient yet

come-nc (approved):

PR #46925, opened 2024-07-31
Add small font size variable

marcoambrosini:


Antreesy (approved):

Antreesy:
`feat: add small font size variable`

susnux (approved):

apps/theming/css/default.css
```diff
@@ -49,6 +49,7 @@
   --color-border-maxcontrast: #7d7d7d;
   --font-face: system-ui, -apple-system, 'Segoe UI', Roboto, Oxygen-Sans, Cantarell, Ubuntu, 'Helvetica Neue', 'Noto Sans', 'Liberation Sans', Arial, sans-serif, 'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol', 'Noto Color Emoji';
   --default-font-size: 15px;
+  --font-size-small: 13px;
```
susnux, line 52: Maybe use 12px as that is the currently smallest font size we need to use (and fits the grid baseline)

marcoambrosini: It seemed too small compared to the default one and looked odd when both are used in the same component (e.g.) Line and subline. I think we should bring all the 12px instances up to 13 for now.

PR #46930, opened 2024-07-31
fix(core): app menu notification should be in the top right

### LINKED ISSUE #46929: Top Navigation notification bubble should be in the top right, as it is e.g. on iOS, Android...
Notification bubble should be in the top right, as it is e.g. on iOS, Android and for our own notification icon as well. Then there is also no text overlap issue. 

_Originally posted by @jancborchardt in https://github.com/nextcloud/server/pull/46916#pullrequestreview-2210602951_

skjnldsv:
Fix #46929 
| Before | After |
|:---------:|:------:|
|image|image|


core/src/components/AppMenuIcon.vue
```diff
@@ -56,8 +56,9 @@ $unread-indicator-size: 10px;
 	&__unread {
 		color: var(--color-error);
 		position: absolute;
-		inset-block-end: calc($unread-indicator-size / -2.5);
-		inset-inline-end: calc($unread-indicator-size / -2.5);
+		// Align the dot to the top right corner of the icon
+		top: -5px;
+		right: -5px;
```
susnux, line 61: Why go back to top / right? This will need to be adjusted once we go for proper RTL support (only inset-{inline/block} does work for RTL).

susnux (approved):
Ok, but personally I would not go back to physical properties but try to always stick with logical.

core/src/components/AppMenuIcon.vue
skjnldsv: I need to keep that in mind! Old habits and tiredness, sorry 😭
Thanks for pointing it out! 🙏🫶

skjnldsv: Done!

AndyScherzinger (approved):

PR #46931, opened 2024-07-31
fix: fix passing additional db connection parameters in factory

icewind1991:
From what I can tell, the added logic for the primary/replica broke passing additional query parameters as the `$additionalConnectionParams` got merged with the `['primary' => ..., 'replica' => ...]` array, not the actual connection parameters.

### DIFF
--- a/lib/private/DB/ConnectionFactory.php
+++ b/lib/private/DB/ConnectionFactory.php
@@ -54,16 +54,10 @@ class ConnectionFactory {
 		],
 	];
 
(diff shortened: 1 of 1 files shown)


susnux (approved):

szelga:
could this be backported into 29?

juliusknorr (approved):

GJSchaller:
Just updated to 30.0.0, and the same error message is coming up when I try to convert the database to mySQL.

Issue #40073, opened 2023-08-28
[Bug]: 


ETararenko:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

A problem occurred while sending the email. Please revise your settings. (Error: Email could not be sent. Check your mail server log).
![image](https://github.com/nextcloud/server/assets/68062481/069a5d79-6f8a-475f-81c2-24dea15b1a27)
Share notification mail also could not be sent.
When sharing file, the following message in logging:
![image](https://github.com/nextcloud/server/assets/68062481/ac5bc5b4-618b-4fa6-a16d-3f17308d6ef7)
When i send mail from the server using command
sendemail -f mail_from@***.com -t mail_to@***.com -u subject -m "message" -s mail_server.com:25
mail comes

### Steps to reproduce

When configuring the mail server, when you click send mail, an error occurs.

### Expected behavior

Mail must be sent.

### Installation method

Community Manual installation with Archive

### Nextcloud Server version

27

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [X] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "passwordsalt": "***REMOVED SENSITIVE VALUE***",
        "secret": "***REMOVED SENSITIVE VALUE***",
        "trusted_domains": [
            "*****.****"
        ],
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
[...]

Issue #39538, opened 2023-07-22
[Bug]: Cant sending mail plain text to 127.0.0.1 after update to 26.0.4 or cant use STARTTLS with TLS v1. The mail system require min TLS v1.2!


sonic42:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

After update to nexcloud server  from 25.0.0.x to 26.0.0.4, sending mail is impossible.

The mail system require TLS 1.2 or higher.

Postfix log:
```
postfix/smtpd[24363]: connect from localhost.localdomain[127.0.0.1]
postfix/smtpd[24363]: SSL_accept error from localhost.localdomain[127.0.0.1]: -1
postfix/smtpd[24363]: warning: TLS library problem: error:0A000418:SSL routines::tlsv1 alert unknown ca:ssl/record/rec_layer_s3.c:1586:SSL alert number 48
```

Using Plaintext is impossible. Cannot disable STARTTLS in configuration.
 

### Steps to reproduce

1. Try to send a mail to a mail server with minimum TLS 1.2 required. 

### Expected behavior

Please repair plaintext option or enable TLS 1.2 or higher by default for mail sending.
"Plain text" is not sending a STARTTLS command.

### Installation method

None

### Nextcloud Server version

26

### Operating system

Other

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MariaDB

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

Encryption is Disabled

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
'mail_smtpmode' => 'smtp',
  'mail_domain' => '',
  'mail_from_address' => 'nc',
[...]

Issue #39452, opened 2023-07-18
[Bug]: Mail Component ignoring encryption setting (NC 26)


kwisatz:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Despite encryption set to `none` in the administration settings, the symfony mailer component uses STARTTLS to connect to the set smtp server:

```
{
  "reqId": "muy8yiNzXXcL38qveYFo",
  "level": 0,
  "time": "2023-07-18T09:17:06+00:00",
  "remoteAddr": "85.208.7.xxx",
  "user": "user",
  "app": "core",
  "method": "POST",
  "url": "/settings/admin/mailtest",
  "message": "Sending mail to \"Array\n(\n    [admin@user.lu] => user\n)\n\" with subject \"Email setting test\" failed",
  "userAgent": "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/114.0.0.0 Safari/537.36",
  "version": "26.0.3.2",
  "exception": {
    "Exception": "Symfony\\Component\\Mailer\\Exception\\TransportException",
    "Message": "Unable to connect with STARTTLS: stream_socket_enable_crypto(): Peer certificate CN=`mail.domain.tld' did not match expected CN=`10.24.0.67'",
    "Code": 0,
    "Trace": [
      {
        "function": "Symfony\\Component\\Mailer\\Transport\\Smtp\\Stream\\{closure}",
        "class": "Symfony\\Component\\Mailer\\Transport\\Smtp\\Stream\\SocketStream",
        "type": "->",
        "args": [
          "*** sensitive parameters replaced ***"
        ]
      },
      {
        "file": "/var/www/nextcloud/3rdparty/symfony/mailer/Transport/Smtp/Stream/SocketStream.php",
        "line": 174,
[...]

Issue #38957, opened 2023-06-23
[Bug]: Nextcloud 26.0.2 email notification error


cycu85:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

hello

there is a bug in mail notification setting.

in log i have:

```
Exception: Symfony\Component\Mailer\Transport\Smtp\Stream\SocketStream::setTimeout(): Argument #1 ($timeout) must be of type float, string given, called in /var/www/nextcloud/lib/private/Mail/Mailer.php on line 305 in file '/var/www/nextcloud/3rdparty/symfony/mailer/Transport/Smtp/Stream/SocketStream.php' line 37
```

### Steps to reproduce

in config.php i try change 
`"mail_smtptimeout"  => '30',`  to `"mail_smtptimeout"  => 30, `
but no efect.

so i delete `"mail_smtptimeout"  => '30'` from config.php
still nothing.

chnge setting from web admin panel restore  `"mail_smtptimeout"  => '30',` to config.php

only delete `"mail_smtptimeout"  => '30'`, from config.php
and
change  
`/var/www/nextcloud/3rdparty/symfony/mailer/Transport/Smtp/Stream/SocketStream.php  in line 47
`
from 
`return $this->timeout ?? (float) \ini_get('default_socket_timeout');`
to:
`return $this->timeout ?? (float) \ini_get(30);`

help me to workaround problem

### Expected behavior

can you fix it?

### Installation method

None

### Nextcloud Server version

26

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

PostgreSQL

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
[...]

PR #46957, opened 2024-08-01
#41935 smtp: allow self-signed certificates on localhost

### LINKED ISSUE #41935: E-Mail: Automatically set 'allow_self_signed=true' for 'mail_smtphost = localhost'
The newer versions on Nextcloud uses StartTLS per default, resulting in many bug reports when used with self-signed certificates:

  https://github.com/nextcloud/server/issues/37329
  https://github.com/nextcloud/server/issues/37694
  https://github.com/nextcloud/server/issues/38957
  https://github.com/nextcloud/server/issues/39452
  https://github.com/nextcloud/server/issues/39538
  https://github.com/nextcloud/server/issues/40073

There is a workaround to allow self signed certificates by manually editing config.php:

```
  "mail_smtpstreamoptions" => array(
    'ssl' => array(
        'allow_self_signed' => true,
        'verify_peer' => false,
        'verify_peer_name' => false
    )   
  ),  
```

But since many users will stumble across this problem let me suggest a small change: 

If "mail_smtphost" is set to "localhost" or "127.0.0.1" and the above parameters are not explicitly configured, then assume that the user wants to accept unverified self-signed certificates by default - or do not use StartTLS at all for local connections.

Cybso:
* Resolves: 
  * #41935 
  * #37329 
  * #37694 
  * #38957 
  * #39452 
  * #39538 
  * #40073
  * and maybe more

## Summary

A mailserver listening on localhost can never have a valid TLS certificate and it does not require one. This patch enables `allow_self_signed` and disables `verify_peer` and `verify_peer_name` when `mail_smtphost` equals `localhost` or `127.0.0.1`.

Pull request as suggested by [joshtrichards](https://github.com/joshtrichards) in #41935

### DIFF
--- a/lib/private/Mail/Mailer.php
+++ b/lib/private/Mail/Mailer.php
@@ -280,8 +280,9 @@ class Mailer implements IMailer {
 	protected function getSmtpInstance(): EsmtpTransport {
 		// either null or true - if nothing is passed, let the symfony mailer figure out the configuration by itself
 		$mailSmtpsecure = ($this->config->getSystemValue('mail_smtpsecure', null) === 'ssl') ? true : null;
+		$mailSmtphost = $this->config->getSystemValueString('mail_smtphost', '127.0.0.1');
 		$transport = new EsmtpTransport(
-			$this->config->getSystemValueString('mail_smtphost', '127.0.0.1'),
+			$mailSmtphost,
 			$this->config->getSystemValueInt('mail_smtpport', 25),
 			$mailSmtpsecure,
 			null,
@@ -304,6 +305,17 @@ class Mailer implements IMailer {
 
 			$currentStreamingOptions = array_merge_recursive($currentStreamingOptions, $streamingOptions);
 
+			/** @psalm-suppress InternalMethod */
+			$stream->setStreamOptions($currentStreamingOptions);
+		} else if ($mailSmtphost === 'localhost' || $mailSmtphost === '127.0.0.1') {
+			$currentStreamingOptions = $stream->getStreamOptions();
+			$currentStreamingOptions = array_merge_recursive($currentStreamingOptions, array(
+				'ssl' => array(
+					'allow_self_signed' => true,
+					'verify_peer' => false,
+					'verify_peer_name' => false
+				)
+			));
 			/** @psalm-suppress InternalMethod */
 			$stream->setStreamOptions($currentStreamingOptions);
 		}


lib/private/Mail/Mailer.php
```diff
@@ -304,6 +305,17 @@ protected function getSmtpInstance(): EsmtpTransport {
 
 			$currentStreamingOptions = array_merge_recursive($currentStreamingOptions, $streamingOptions);
 
+			/** @psalm-suppress InternalMethod */
+			$stream->setStreamOptions($currentStreamingOptions);
+		} else if ($mailSmtphost === 'localhost' || $mailSmtphost === '127.0.0.1') {
```
szaimen, line 310: In theory we could even add this if no valid host was provided, e.g. by using such a filter: https://github.com/nextcloud/all-in-one/blob/c9b97220d0175914c3ee8c7199949376f7bfe3b0/php/src/Data/ConfigurationManager.php#L274-L295

However this sounds like magic and not sure if we want to actually do this automatically.

WDYT @nextcloud/server-backend @miaulalala

kesselb:
```
1) Test\Mail\MailerTest::testStreamingOptionsWrongType
Failed asserting that 1 matches expected 0.

/home/runner/actions-runner/_work/server/server/tests/lib/Mail/MailerTest.php:291
```

Cybso:
> ```
> 1) Test\Mail\MailerTest::testStreamingOptionsWrongType
> Failed asserting that 1 matches expected 0.
> 
> /home/runner/actions-runner/_work/server/server/tests/lib/Mail/MailerTest.php:291
> ```

I will adapt the test as soon as it is clear whether this pull request has a chance of being accepted or not.

lib/private/Mail/Mailer.php
```diff
@@ -304,6 +305,17 @@ protected function getSmtpInstance(): EsmtpTransport {
...
+			$currentStreamingOptions = $stream->getStreamOptions();
+			$currentStreamingOptions = array_merge_recursive($currentStreamingOptions, array(
+				'ssl' => array(
+					'allow_self_signed' => true,
+					'verify_peer' => false,
+					'verify_peer_name' => false
+				)
+			));
```
miaulalala, line 318: I'm of two minds here - one is that it indeed makes localhosts work without any setup, which, as you stated has confused many users due to how the Symfony Mailer works. OTOH I'm not entirely happy with overwriting config options per default.

I think there's some middle ground here: check if the `ssl` option is set and only overwrite it if a) the host is `localhost || 127.0.0.1` and b) if the option doesn't already exist in the `$currentStreamingOptions`

I also think it doesn't need to be an elseif but can be an if condition on it's own. Something like:

```suggestion
		} 
		$local = ['localhost', '127.0.0.1']; // You could make this a constant
		if (in_array($mailSmtphost, $local) && !isset($currentStreamingOptions['ssl'])) {
			$currentStreamingOptions = $stream->getStreamOptions();
			$currentStreamingOptions = array_merge_recursive($currentStreamingOptions, array(
				'ssl' => array(
					'allow_self_signed' => true,
					'verify_peer' => false,
					'verify_peer_name' => false
				)
			));
```

miaulalala (changes requested):

kesselb:
https://github.com/symfony/symfony/pull/53621 (but it will take a while until we are on mailer 7.1)

szaimen:
Hi, we have a checkbox now to allow insecure certificates: https://github.com/nextcloud/server/pull/57544/changes#diff-d611085ec819ab74b9abec857410a14d726ea1b1c0b84dc0d99bd88be39f54d0R165-R167 so this can be closed imho

image

miaulalala:
Thanks so much for your time and effort @Cybso ! https://github.com/nextcloud/server/pull/57544/ will hopefully solve the issues with self signed certificates (and servers that advertise TLS but don't actually support it).

PR #46985, opened 2024-08-02
feat(sharebymail): improve share email format

### LINKED ISSUE #15680: Add expiration date on sharing notification mail
When the first share notification mail is sent, it would be nice to have the expiration date on it.

skjnldsv:
|Before|After|
|:---:|:---:|
|image|image|

As discussed:
1. Removed `»` and `«`
2. Add expiration
3. Improve layout

Fix https://github.com/nextcloud/server/issues/15680
cc @jancborchardt @nimishavijay @marcoambrosini


szaimen:
I just looked at the after screenshot and I was wondering if technically these elements could be bold? 
image

Also, the button on the bottom should probably have the same border-radius that we are useing in Nextcloud for buttons?

cc @nextcloud/designers

skjnldsv:
> I just looked at the after screenshot and I was wondering if technically these elements could be bold?

- For `Note` and `Expiration`, the place to fix it is [here](https://github.com/nextcloud/server/blob/693a81bfa367089de4b61ee1ea104c85e6bc6d1c/lib/private/Mail/EMailTemplate.php#L437). This will change for all apps using our email templating (like calendar events) 
image
 
For the folder title, that would be nice, but it's not supported for now.
We don't support custom html. We could add some custom support like `**` to convert to bold during the template processing.
But I'd say this is something for 31 or later

skjnldsv:
> Also, the button on the bottom should probably have the same border-radius that we are useing in Nextcloud for buttons?

This is an issue older than internet :see_no_evil: 
Not all email clients support border radius. :exploding_head: 
Maybe things changed since, but that was the reason this is not possible :see_no_evil:

szaimen:
> > Also, the button on the bottom should probably have the same border-radius that we are useing in Nextcloud for buttons?
> 
> This is an issue older than internet 😁 Not all email clients support border radius. 🤯 Maybe things changed since, but that was the reason this is not possible 🙈
> 
> EDIT: nope, still the case ahahah https://www.caniemail.com/features/css-border-radius/

OMG 🙈

skjnldsv:
Addressed the border radius, see first post

nimishavijay (approved):
Super nice! Very cool that the buttons are rounded now :) 

All looks good, only small change I would suggest is to remove the message in gray at the bottom "X shared Y with you. Click the button below to open it." as it seems redundant now :) 

Approving to not block

sorbaugh (approved):

susnux (approved):

PR #46987, opened 2024-08-02
fix(dav): Ensure share properties are also set on public remote endpoint

susnux:
## Summary

Ensure that public and internal endpoint both provide the necessary data, so we can e.g. on both request `nc:note`.
As the share note is not something only for users but also for public shares.

## Checklist

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Node.php
+++ b/apps/dav/lib/Connector/Sabre/Node.php
@@ -15,6 +15,7 @@ use OCA\DAV\Connector\Sabre\Exception\InvalidPath;
 use OCP\Files\DavUtil;
 use OCP\Files\FileInfo;
 use OCP\Files\IRootFolder;
+use OCP\Files\NotFoundException;
 use OCP\Files\StorageNotAvailableException;
 use OCP\Share\Exceptions\ShareNotFound;
 use OCP\Share\IManager;
@@ -298,15 +299,14 @@ abstract class Node implements \Sabre\DAV\INode {
 	 * @return array
 	 */
 	public function getShareAttributes(): array {
-		$attributes = [];
-
 		try {
-			$storage = $this->info->getStorage();
-		} catch (StorageNotAvailableException $e) {
-			$storage = null;
+			$storage = $this->node->getStorage();
+		} catch (NotFoundException $e) {
+			return [];
 		}
 
-		if ($storage && $storage->instanceOfStorage(\OCA\Files_Sharing\SharedStorage::class)) {
+		$attributes = [];
+		if (method_exists($storage, 'getShare')) {
 			/** @var \OCA\Files_Sharing\SharedStorage $storage */
 			$attributes = $storage->getShare()->getAttributes();
 			if ($attributes === null) {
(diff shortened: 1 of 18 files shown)


skjnldsv (approved):
This is on the very edge between a fix and a feat... 😅🙈
Approved nonetheless!

apps/dav/lib/Connector/Sabre/Node.php
```diff
@@ -298,15 +299,14 @@ public function getSharePermissions($user) {
...
-		} catch (StorageNotAvailableException $e) {
-			$storage = null;
+			$storage = $this->node->getStorage();
+		} catch (NotFoundException $e) {
+			return [];
 		}
 
-		if ($storage && $storage->instanceOfStorage(\OCA\Files_Sharing\SharedStorage::class)) {
+		$attributes = [];
+		if (method_exists($storage, 'getShare')) {
```
come-nc, line 309: Why the change from `instanceOfStorage` to `method_exists`?
Seems way more fragile.

come-nc: Hum ok I see that you’ve added the same method to PublicShareWrapper. Would it be possible to add an interface implemented by both classes and use that in instanceOfStorage?

apps/dav/lib/Connector/Sabre/Node.php
```diff
@@ -319,29 +319,25 @@ public function getShareAttributes(): array {
...
 		}
 
-		// Retrieve note from the share object already loaded into
-		// memory, to avoid additional database queries.
-		$storage = $this->getNode()->getStorage();
-		if (!$storage->instanceOfStorage(\OCA\Files_Sharing\SharedStorage::class)) {
+		if (!method_exists($storage, 'getShare')) {
 			return '';
 		}
 		/** @var \OCA\Files_Sharing\SharedStorage $storage */
```
come-nc, line 332: This psalm typing is not true anymore

susnux: See your other comment it helps Psalm to understand there is that method ;)
(dirty maybe need to fix your other comment).

apps/dav/lib/Connector/Sabre/Node.php
susnux: Problem is the `SharedStorage` is only for internal shares so public shares are not an instance of it.

> Would it be possible to add an interface implemented by both classes and use that in instanceOfStorage?

Like having `IShareStorage` with that method and just check for `instanceOfStorage(IShareStorage)` ?

come-nc: Yes.
Otherwise that will bite us later.

apps/files_sharing/lib/SharedStorage.php
```diff
@@ -35,7 +36,7 @@
 /**
  * Convert target path to source path and pass the function call to the correct storage provider
  */
-class SharedStorage extends \OC\Files\Storage\Wrapper\Jail implements ISharedStorage, IDisableEncryptionStorage {
+class SharedStorage extends \OC\Files\Storage\Wrapper\Jail implements LegacyISharedStorage, ISharedStorage, IDisableEncryptionStorage {
```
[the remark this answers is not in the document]
susnux: Deprecated on purpose but we need it here for backwards compatibility, remove with Nextcloud 34.

miaulalala (approved):

PR #46991, opened 2024-08-02
fix(ProvisioningAPI): set typed config values by via API

blizzz:
* Resolves: #45083

## Summary

On Javascript side we have one method at hand to set an app config value: `CP.AppConfig.setValue()`. As value it takes a `string` and sends this over the Provisioning API to the server. 

Meanwhile we have typed configuration entries. All existing entries are default to the `IAppConfig::VALUE_MIXED` type. New entries however are bound to their type. For instance a new boolean value cannot be set via JavaScript anymore.

This fix solves this on API level. All clients may still send strings, and the Provisioning API's Controller tries to cast the provided value into the target type.

I see this sort of critical myself as there is not good validation of the casted value. On the other hand I do not think it can do much harm, only fail in a different way. Still, an alternative would be to have also dedicated setter on the JS side, and endpoints for each acceptable type. Personally I don't have the time for this approach today anymore. I am totally fine to discard this PR in favor of a better replacement :)

## Todo

* [ ] check were backports are necessary

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [X] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #46992, opened 2024-08-02
turn LDAP's treat remnants as disabled feature  config-independent

blizzz:
## Summary

* [x] requires https://github.com/nextcloud/server/pull/46991 

Since former LDAP users cannot be bound to a server configuration anymore (or theoretically belong to more than one) this feature has to be config independent. This PR:

- changes the existing logic from config-bound to app-global
- including web interface changes
- has a migration step to set the new values based on the old ones – one previous activation enables it for the whole LDAP backend

A backport to 28 will need further adjustments as `IAppConfig` was only introduced in 29.

Screenshot_20240802_181949

## Checklist


come-nc (changes requested):

apps/user_ldap/lib/Settings/AppSettings.php
```diff
@@ -0,0 +1,46 @@
+<?php
+/**
```
come-nc, line 5:
```suggestion
<?php

declare(strict_types=1);

/**
```

apps/user_ldap/lib/User_Proxy.php
```diff
@@ -400,15 +402,18 @@ public function createUser($username, $password) {
 	}
 
 	public function isUserEnabled(string $uid, callable $queryDatabaseValue): bool {
-		return $this->handleRequest($uid, 'isUserEnabled', [$uid, $queryDatabaseValue]);
+		if (!$this->treatRemnantsAsDisabled() && $this->deletedUsersIndex->isUserMarked($uid)) {
```
come-nc:
```suggestion
		if ($this->treatRemnantsAsDisabled() && $this->deletedUsersIndex->isUserMarked($uid)) {
```

I think this is reversed?

apps/user_ldap/lib/User_Proxy.php
```diff
@@ -431,4 +436,12 @@ public function getDisabledUserList(?int $limit = null, int $offset = 0, string
 			)
 		);
 	}
+
+	protected function treatRemnantsAsDisabled(): bool {
+		static $setting;
+		if (!$setting) {
+			$setting = $this->appConfig->getValueBool('user_ldap', 'markRemnantsAsDisabled', false);
+		}
+		return $setting;
```
come-nc:
```suggestion
		return $this->appConfig->getValueBool('user_ldap', 'markRemnantsAsDisabled', false);
```

I would tend to avoid this kind of micro-optimisation which makes code complicated and will bite us later.
If you must, then cache in a property, not in a static var.

blizzz: I think I just wanted to save the function call :shrug:

apps/user_ldap/lib/Configuration.php
```diff
@@ -438,6 +443,14 @@ protected function getLcValue(string $varName): string {
 		return mb_strtolower($this->getValue($varName), 'UTF-8');
 	}
 
+	protected function getGlobalAppValueAsBool(string $varName): bool {
+		static $appConfig;
```
come-nc, line 447: Same as the other one, static var is scary. Cache in a property or do not cache. (There is no way to reset the static var ever, while, the property can be reset by trashing the object and building a new one)

come-nc:
@blizzz It would be nice to rebase this on the vue rewrite and remove the static vars. Then we can merge.

PR #47120, opened 2024-08-07
fix(systemtags): Correctly load tagged files in "tags"-files-view

susnux:
* Regression of https://github.com/nextcloud/server/pull/46921

When splitting the path it is `/{ID}` to the splitted array is `['', '{ID}']` -> we need to use the last part otherwise we only get 404.


skjnldsv (approved):

skjnldsv:
We need tests for this :see_no_evil:

AndyScherzinger (approved):

AndyScherzinger:
Merged to have it in Beta5 but created a follow-up at #47127 to have an issue for implementing tests

susnux:
> We need tests for this 🙈

I am already writing some, but this was more urgent :sweat_smile:

PR #47121, opened 2024-08-07
fix(systemtags): Make inline tags list fully accessible

susnux:
## Summary

1. The label was wrong it mentioned the first and last tag, so the information was duplicated
2. It only mentioned the first and last but not all other
3. All other tags were only added as `title` which is not accessible

## Checklist


skjnldsv (approved):

sorbaugh (approved):

susnux:
I am not sure about backporting?

PR #47135, opened 2024-08-08
fix(systemtags): Sub folders should be opened in files

susnux:
## Summary

Currently this is simply broken and here are two ways:
1. Open subfolders in files view
2. Implement logic to save last request

1 is the way this is now implemented, basically copy-paste from the recent view. 2 is much more complicated because if we get `/2/foo/bar` as the path we need to know the source of `foo`, so we would need at least 2 requests or cache the previous directory. We do not do it like this for any view so lets just stick with 1 for now.

## Checklist


apps/systemtags/src/files_actions/openInFilesAction.ts
```diff
@@ -0,0 +1,44 @@
+/**
+ * SPDX-FileCopyrightText: 2023 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { translate as t } from '@nextcloud/l10n'
+import { type Node, FileType, FileAction, DefaultType } from '@nextcloud/files'
+
+export const action = new FileAction({
+	id: 'systemtags:open-in-files',
```
skjnldsv, line 9: This bother me more than it should 😅😅

skjnldsv (approved):

AndyScherzinger (approved):

PR #47145, opened 2024-08-09
feat: Add AES Encryption/Decryption for User Login Credentials

Anyel-ec:


kesselb:
Hi and thanks for your pull request :+1: 

I'm confused by the pull request description. Everything you listed under "Changes" was already there?

core/src/components/login/LoginForm.vue
```diff
@@ -274,15 +228,40 @@ export default {
 		updateUsername() {
 			this.$emit('update:username', this.user)
 		},
-		submit(event) {
+
+		encrypt(data) {
+			// Generar clave y IV
+			const secretKey = 'my-secret-key'; // Clave en hexadecimal
```
susnux, line 234: What security benefit should this bring? The key is known so it is basically just (revertable) hiding the data if you sniff the packages. But meaning it does not add any security?

Issue #32086, opened 2022-04-22
 User Profile should be usable without public visibility


kolewu:
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments. 

**Is your feature request related to a problem? Please describe.**
The new user profile feature is a valuable addition to every cloud, but it's sad that it always is publicly visible. The most private clouds are just that: private. And so no information should be visible to the public.

The current profile is a choice between "make some information visible to the public" or "don't use profiles". But profiles are useful for private use inside the cloud, too.

**Describe the solution you'd like**
The user profile as a whole should at least get an option "only visible to logged in users". Even better, the administrator should be able to configure the visibility for all accounts.

So I request:
- a user option to make the profile invisible to the public
- administrator options
  - make no profile visible to the public and the user could not change this
  - make no profile visible to the public by default and the user can change it afterwords
- for updates make all profiles not visible to the public and the user could change this

**Describe alternatives you've considered**

**Additional context**

eibex:
This would also help to ensure that potential usernames can't be checked by malicious users using cloud.example.com/{username}

szaimen:
cc @jancborchardt @nimishavijay

szaimen:
see https://github.com/nextcloud/server/pull/31623

nimishavijay:
You can control the information that it displayed on your profile using the Profile visibility settings in the bottom of your personal settings page, and these settings allow you to configure if your information is shown to nobody, only logged in users of the instance or to the public. Would this solve the issue?

It could be useful to make the default visibility of all the information on your Profile to just users of the instance @jancborchardt

eibex:
If everything is set not to public it would be better if you tried to load a user profile while not logged in it would redirect you to the login page (as id the username you entered in the URL did not exist)

kolewu:
> You can control the information that it displayed on your profile using the Profile visibility settings in the bottom of your personal settings page, and these settings allow you to configure if your information is shown to nobody, only logged in users of the instance or to the public. Would this solve the issue?

[...]

PR #47170, opened 2024-08-10
Change default profile visibility to logged-in users only

cherti:
* Resolves: # none, I can open one in addition to sending the fix if the process requires that

## Summary

This commit changes the default visibility of all profile properties to logged-in users only, so that users do not accidentially leak part of their information assuming that a privacy-oriented software like nextcloud would not expose their profile information to the outside by default.
This is especially relevant given that the profile visibility settings only show up on the main screen depending on resolution and size, see screenshots below. There is the "show profile visibility", but it is subtle and I consider it a valid assumption that the default settings for users of a privacy- and user-oriented product like nextcloud are unsurprising and safe by default without the need to take a look. If they are not, it is just a question of time until someone has their profiles with settings they don't want or intend.
This PR only changes the default values, no functionality is added or removed. The change also, as far as I can tell, only affects newly created profiles, and public visibility can be restored with a click in the profile visibility settings if desired.

nextcloud1

nextcloud2

## Checklist

- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included -> there seem to be not tests associated with this particular codepath, and no new code has been introduced


susnux:
Not really sure who to ask if we should adjust those fields, as they are not that private fields and in the default use case you probably want them to be public. For example "role" or "organization" you do not use them for personal use but on enterprise and then you probably want them accessible.

So maybe ask @jancborchardt ?

cherti:
Well, in the current configuration they are world-readable. So every Nextcloud as a service for small businesses and organisations, basically everyone who doesn't have an intranet or that Nextcloud on the intranet, has them world-exposed in the current default configuration. If an organization has enough staff to host this on an intranet, they might have enough staff to change the default configurations if they want to, but especially smaller organisations, also voluntary organisations, are currently vulnerable to everyone having to remember changing that visibility to not having it exposed. It just feels like a very non-sane default especially given who significant portions of the userbase of Nextcloud are consists of.

jancborchardt (changes requested):
So we would like to promote people using the profile, possibly having it useful as some sort of federated address book. If the default is changed like this, people will likely never enable it.

@cherti your use-case of course is reasonable, and to achieve that I would rather say there should be an admin setting (or config.php setting). @AndyScherzinger @sorbaugh

SergioArbarviro:
I thank you @cherti for this commit. It goes in the right direction, but in my views not far enough. As I mentioned in #32086 , I believe that the defaults should be:

- logged in users only for: username, profile text, organisation, position and avatar
- "private" (visible only to the user him/herself) for **all** other fields, in particular: e-mail address, location, telephone number, etc that relate to personal and identifiable data as described in the EU General Data Protection Regulation - GDPR. As a citizen of the EU, I expect a privacy-oriented software like NextCloud, developed moreever in the EU, to comply by default with this legislation.

cherti:
>     * "private" (visible only to the user him/herself) for **all** other fields, in particular: e-mail address, location, telephone number, etc

I agree for the e-mail field, for the rest you could just not fill this. However, for e-mail making this private here seems insufficient, because Nextcloud also provides a built-in address book of all users of that Nextcloud. That would have to be changed as well to make this change effective, and I think this is beyond the scope of this PR.

cherti:
> So we would like to promote people using the profile, possibly having it useful as some sort of federated address book. If the default is changed like this, people will likely never enable it.

The current situation is that people will never use the profile if they found out that it unintentionally by them exposed information to the wide internet. I think that breach of trust will have a significantly higher impact on this not becoming a federated address book than people opting to not use this as a federated address book. If there was an admin setting to change these defaults, this would likely be a non-issue (even though I still think it doesn't act upon the "least surprise" principle given that one of Nextcloud's points of pride is privacy), but there is no such thing. There is no way as an admin to influence this setting except with fiddling with the code.
As a matter of fact, the only reason we are using these profiles (and we actually want to) is because in the setup that we have, I **can** monkeypatch the Nextcloud code to remedy this issue. Otherwise we would flatout disable profiles alltogether. So I think you are actually harming adoption of profiles with the current state of these visibilities.

> @cherti your use-case of course is reasonable, and to achieve that I would rather say there should be an admin setting (or config.php setting). @AndyScherzinger @sorbaugh

I am indeed irritated why this was introduced this way without having an admin setting for it. I spend good chunks of time looking for it before realizing that it simply does not exist.

susnux:
> "private" (visible only to the user him/herself) for **all** other fields, in particular: e-mail address, location, telephone number, etc that relate to personal and identifiable data as described in the EU General Data Protection Regulation - GDPR. As a citizen of the EU, I expect a privacy-oriented software like NextCloud, developed moreever in the EU, to comply by default with this legislation.

What reason would one have to fill in this information and then just see it themself?

cherti:
@jancborchardt it says you requested changes, but I'm unable to see which changes are requested. Can you elaborate?

AndyScherzinger:
@cherti I think it is https://github.com/nextcloud/server/pull/47170#pullrequestreview-2260753012 itself where you referred to earlier already via https://github.com/nextcloud/server/pull/47170#issuecomment-2314567139

And I think that could/would be the way to go and fine to get merged then too 👍 from my perspective

cherti:
Ok, but that is something I cannot provide as my familiarity with PHP and the entire environment does not extend to this.

I wholeheartedly agree that Nextcloud should have such a feature and until that gets implemented, I would propose tuning the defaults in a way that it doesn't accidentially leak people's information, and then change those defaults back once we have a setting that allows to set this, as done in this PR.

Accidentially exposing user's information to the outside doesn't seem to provide a way of instilling confidence to (and hence make people) use this feature, quite the opposite. At least it did not with the userbase I am involved in, why I am currently patching the nextcloud code prior to deployment on every update to prevent such leakage to happen in the future.

AndyScherzinger:
Ah, that I wasn't aware of. So I'll discuss the implementation part with @sorbaugh and @jancborchardt then 👍

cherti:
Given that the scope of this PR is fairly minimal, and given that I know very well how busy everyone is, could we merge this as an immediate fix and then once the feature is fully implemented, see if these values need adjustment again? :)

cherti:
@jancborchardt in https://github.com/nextcloud/server/issues/32086#issuecomment-2348640283 :

> I can not make any promises of course. As Nextcloud is an open community, if anyone with more deep PHP knowledge wants to have a go at it, please go ahead.
> But I would stop discussing here since it is just making the thread longer and anyone who possibly would work on this more confused.

But if you can't make any promises, is there any harm in merging this now as the immediate fix and possibly adjusting it when somebody does the full implementation? Given it's an open community, somebody put in the work to prepare this simple fix for easy merging, after all, and it would resolve most of the issues for the time being until the full implementation of https://github.com/nextcloud/server/issues/32086 is done. :)

susnux:
> is there any harm in merging this now as the immediate fix and possibly adjusting it when somebody does the full implementation

It changes the default state for all to a state which we think is not an optional default state. There is also no way for instances that want the default behavior back to change it. So this PR here is quite a huge breaking change for a very limited use case.
So until the feature (restricting the default visibility) is implemented I would rather recommend you to adjust your user provisioning workflow to set the desired visibility for new users, e.g. using the OCS API:
https://docs.nextcloud.com/server/latest/developer_manual/_static/openapi.html#/operations/core-profile_api-set-visibility

cherti:
So, to understand this correctly, you say this necessitates a fix after the fact which needs to be executed after every user registered? Or after the user activated their profile?

SergioArbarviro:
> > "private" (visible only to the user him/herself) for **all** other fields, in particular: e-mail address, location, telephone number, etc that relate to personal and identifiable data as described in the EU General Data Protection Regulation - GDPR. As a citizen of the EU, I expect a privacy-oriented software like NextCloud, developed moreever in the EU, to comply by default with this legislation.
> 
> What reason would one have to fill in this information and then just see it themself?

There is at least one field that the user has no other choice than providing information on, it is the **e-mail address**, which is a personal data enabling identification, hence requiring protection under the EU GDPR. 

NextCloud **must**, at least in the EU, make this information **private by default**, and not disclose it to anyone without the prior, explicit consent of the user - not, in my views, even to the other users of the NextCloud instance. This stands in particular for open communities.

SergioArbarviro:
> > is there any harm in merging this now as the immediate fix and possibly adjusting it when somebody does the full implementation
> 
> It changes the default state for all to a state which we think is not an optional default state. There is also no way for instances that want the default behavior back to change it. So this PR here is quite a huge breaking change **for a very limited use case**.

I disagree that this is a "very limited use case". Privacy by default is the law in the whole European Union, as per the GDPR. I expect the European Union to constitute a very significant share of the installed base of NextCloud. I would not rate it as a "very limited use case", and don't see on the other hand the use case for disclosing one's personal data by default to the whole Internet. Even FaceBook keeps that data for itselft (and, presumably, for the US intelligence services).

cherti:
Are there actually numbers of how many people actively use the federated address book functionality? Because my personal impression is that Nextcloud is primarily used by institutions and people who want to retain their data and not spread it. As @SergioArbarviro says, this is particularly a topic in Europe, and Nextcloud is recommended in these contexts specifically for privacy reasons, so I'm wondering if actual numbers exist on how this feature is actually used?

joshtrichards:
Related:
- #54158
- nextcloud/documentation#14128

Mondane:
Scope 'local' currently includes not logged in users.

> Local
> Only visible to people on this instance and guests

So how would I achieve this with this PR?

susnux:
Guests are not "not logged in" users but users that does have a user account.
There is anonymous / public access and guests which is  https://github.com/nextcloud/guests

PR #47177, opened 2024-08-12
[Task processing] Fix preferred providers

julien-nc:
The first available provider is selected when running a sync task and when populating the available task type values (optional shapes, enum values etc...).

* Fix the sync task background job: Skip providers that are not the preferred one for their task type
* Fix `IManager:getAvailableTaskTypes()`: Use the preferred provider to extend task type values
* Add task processing providers in the testing app

@marcelklehr Should something be fixed for async workers (when exApps are getting a task via the API)?


bigcat88:
Do we have documentation on what "Preferred Provider" is?

What will be its purpose after this PR?

Application `A` and `B` can handle task `O`

Application `B` is a preferred provider.

It was removed or disabled. Shouldn't application `A` start handling tasks `O` after this?

I just thought that "Preferred Provider" is a priority, that you can always choose the provider that has the highest priority and will always be chosen if it is available - but apparently I am confused.

julien-nc:
@bigcat88 `IManager::getPreferredProvider` returns a provider even if the preferred one (from the settings) does not exist anymore.

So IMO it works fine in both things that were fixed:
* bg job: `IManager::getPreferredProvider` will return a provider that exists so the logic still works
* in `IManager::getAvailableTaskTypes`, `IManager::getPreferredProvider` is used, all good

bigcat88 (approved):
thanks for explanations

marcelklehr (approved):

PR #47180, opened 2024-08-12
Apply group limit on remove from group

come-nc:
## Summary

When removing a user from a group, remove shares which would not be creatable anymore according to group limit on sharing.

## Checklist

### DIFF
--- a/cypress/e2e/files_sharing/limit_to_same_group.cy.ts
+++ b/cypress/e2e/files_sharing/limit_to_same_group.cy.ts
@@ -29,11 +29,15 @@ describe('Limit to sharing to people in the same group', () => {
 	let randomFileName1 = ''
 	let randomFileName2 = ''
 	let randomGroupName = ''
+	let randomGroupName2 = ''
+	let randomGroupName3 = ''
 
 	before(() => {
 		randomFileName1 = Math.random().toString(36).replace(/[^a-z]+/g, '').substring(0, 10) + '.txt'
 		randomFileName2 = Math.random().toString(36).replace(/[^a-z]+/g, '').substring(0, 10) + '.txt'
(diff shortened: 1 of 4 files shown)


miaulalala (approved):

yemkareems (approved):

bigcat88 (approved):

jonas2515:
Tbh this seems not very thought through, it makes the behavior of existing shares on settings and group membership changes super inconsistent.

While now existing shares are removed when manipulating the groups of a user, they aren't removed when enabling the "only share between users of group" setting. Neither are they removed when disabling the whole share API, nor when changing the "limit/exclude sharing to/from groups" settings. I assume the behavior for re-shared shares would also have to be considered.

I don't think a simple action like toggling group membership (or toggling a checkbox in settings) should do such destructive things as deleting shares, can we maybe just remove this behavior again?

@come-nc ping just in case you don't receive notifications for merged PRs.

come-nc:
> While now existing shares are removed when manipulating the groups of a user, they aren't removed when enabling the "only share between users of group" setting. Neither are they removed when disabling the whole share API, nor when changing the "limit/exclude sharing to/from groups" settings. I assume the behavior for re-shared shares would also have to be considered.
> 
> I don't think a simple action like toggling group membership (or toggling a checkbox in settings) should do such destructive things as deleting shares, can we maybe just remove this behavior again?

I think it’s for this reason that changing settings is not directly deleting shares. But it would be a nice feature to have a button to apply current settings to existing shares indeed.

But this PR is not when changing settings but when changing groups.
It’s fixing what was reported as a security issue, see https://github.com/nextcloud/security-advisories/security/advisories/GHSA-35gc-jc6x-29cm
When configuring sharing restriction based on groups, it should be expected that removing a user from a group has consequences.

PR #47182, opened 2024-08-12
perf: improve cleanup of tags/comments

icewind1991:
delete entire chunk at once instead of one-by-one

### DIFF
--- a/apps/files/lib/BackgroundJob/DeleteOrphanedItems.php
+++ b/apps/files/lib/BackgroundJob/DeleteOrphanedItems.php
@@ -66,18 +66,15 @@ class DeleteOrphanedItems extends TimedJob {
 
 		$deleteQuery = $this->connection->getQueryBuilder();
 		$deleteQuery->delete($table)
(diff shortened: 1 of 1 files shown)


sorbaugh (approved):

apps/files/lib/BackgroundJob/DeleteOrphanedItems.php
```diff
@@ -66,18 +66,15 @@ protected function cleanUp($table, $idCol, $typeCol) {
...
-				$deletedInLastChunk++;
-				$deletedEntries += $deleteQuery->setParameter('objectid', (int) $row[$idCol])
-					->execute();
-			}
-			$result->closeCursor();
+			$chunk = $query->executeQuery()->fetchAll(\PDO::FETCH_COLUMN);
+			$deletedInLastChunk = count($chunk);
+
+			$deleteQuery->setParameter('objectid', $chunk, IQueryBuilder::PARAM_INT_ARRAY);
+			$deletedEntries += $deleteQuery->executeStatement();
```
nickvergessen, line 77: Should we wrap with a transaction to make it faster?

come-nc (approved):

PR #47185, opened 2024-08-12
fix: Filename validation should only forbid `create` and `update`

### LINKED ISSUE #44963: 🪟📁📄 Ability to enforce windows compatible file names on the server

susnux:
* Resolves: https://github.com/nextcloud/server/issues/44963

## Summary

1. Fix error messages to be consistent and follow the design rules (as they are used by our mobile apps)
2. `isForbidden` must only check full filename - this is used for other purposes (backwards compatibility + fix issues)
3. DAV is directly using the storage and skips their validation, similar other places do, they and the frontend check only the permissions of the node. So we need to adjust the permissions (remove `create` and `update`) when the node is places within an invalid node.
4. The view needs to check also the path for invalid nodes but still needs to allow read-only access on invalid names.

### Screen recording

Setup - folder before enabling "Windows support":
image

#### Without this ("before")

[Bildschirmaufnahme_20240812_182432.webm](https://github.com/user-attachments/assets/3c02e1da-f7d2-4971-96ca-0c029d376b03)

#### With this ("after")

[Bildschirmaufnahme_20240812_182149.webm](https://github.com/user-attachments/assets/df64c38f-b86a-4001-b206-e70b095dfd21)

## Checklist


come-nc:
Hum, so the New button is greyed out because the name of the folder is invalid, but how can the user understand this?
There should be an error message somewhere  explaining this, no?

lib/private/Files/View.php
```diff
@@ -1517,6 +1544,9 @@ public function getDirectoryContent($directory, $mimetype_filter = '', ?\OCP\Fil
 						if ($sharingDisabled) {
 							$rootEntry['permissions'] = $rootEntry['permissions'] & ~\OCP\Constants::PERMISSION_SHARE;
 						}
+						if ($folderIsReadonly) {
+							$rootEntry['permissions'] = $rootEntry['permissions'] & ~(\OCP\Constants::PERMISSION_CREATE | \OCP\Constants::PERMISSION_UPDATE);
+						}
```
come-nc, line 1549: It’s allowed to share a file with an invalid name?

lib/private/Files/View.php
```diff
@@ -1826,28 +1856,44 @@ private function getPartFileInfo(string $path): \OC\Files\FileInfo {
...
+	public function verifyPath($path, $fileName, bool $readonly = false): void {
 		// All of the view's functions disallow '..' in the path so we can short cut if the path is invalid
 		if (!Filesystem::isValidPath($path ?: '/')) {
 			$l = \OCP\Util::getL10N('lib');
 			throw new InvalidPathException($l->t('Path contains invalid segments'));
 		}
 
+		// Handle filename validation for read-only operations, as we still allow to rename invalid files.
+		if ($readonly) {
+			// There is one exception: Forbidden files like `.htaccess` must not be accessible
```
come-nc, line 1871: That brings a subtle difference between `forbidden_filenames` and `forbidden_filename_extensions`/`forbidden_filename_basenames`/`forbidden_filename_characters`, that will need to be documented

susnux:
> Hum, so the New button is greyed out because the name of the folder is invalid, but how can the user understand this?

Its greyed out because the permission does not include "CREATE". Did not change the front end here, could be done in a follow up, but fixing the permissions.

lib/private/Files/View.php
susnux: From the feature requirements yes, everything except writing.
So users should still be able to work with it, but not create new (idea is to not block access to existing).

lib/private/Files/View.php
```diff
@@ -1097,98 +1099,108 @@ public function free_space($path = '/') {
...
-				$this->lockFile($path, ILockingProvider::LOCK_SHARED);
+		// Verify that the operation is allowed
+		try {
+			$parent = dirname($path);
+			$parentName = basename($parent);
+			$this->verifyPath($parent, basename($path), !in_array('write', $hooks));
+			// If this is a create operation also verify the parent folder name
+			if (in_array('create', $hooks) && $parentName !== '') {
+				$this->verifyPath(dirname($parent), $parentName);
 			}
```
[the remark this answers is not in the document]
susnux: Yes the path is verified recursively (line 1880)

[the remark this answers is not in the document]
come-nc: Ok, I was confused by the fact that we call verifyPath manually on the parent here, but it’s because we do the check for non-readonly this second time.

come-nc (approved):
Approved as `after` is better than `before`. But we should make the situation more clear in the UI in my opinion, or we’ll get confused users about why they do not have write access to the folder, and will look into ACLs and storages and file permissions before thinking about a forbidden name on a parent folder.

sorbaugh (approved):

juliushaertl:
Test failures seem related, looks like it widely fails on install

skjnldsv:
I guess we would need the parent folder permissiosn to check for CREATE too ?
Maybe we should adjust the api and do something like this?

image

skjnldsv:
> I guess we would need the parent folder permissiosn to check for CREATE too ? Maybe we should adjust the api and do something like this?
> 
> image

https://github.com/nextcloud-libraries/nextcloud-files/pull/1124

PR #47196, opened 2024-08-13
do not init navigation data too soon

ArtificialOwl:
do not run `init()` during `add()`:

- if an app calls `NavigationManager::add()` during the registration process of `Application.php`, the route generated (using `IUrlGenerator`) for apps using `<navigations>` in `info.xml` could be empty if app is not fully loaded when `init()` is called.
- `getAll()` that returns data filled by `add()` will call `init()` anyway and generate the correct data using the right routes as all apps are loaded at that point.

Because `$this->entries` is only read during `getAll()`, it should not be an issue to not `init()` during `add()`

Fixes https://github.com/nextcloud/fulltextsearch/issues/842

### DIFF
--- a/lib/private/NavigationManager.php
+++ b/lib/private/NavigationManager.php
@@ -70,7 +70,6 @@ class NavigationManager implements INavigationManager {
 			$this->closureEntries[] = $entry;
 			return;
 		}
-		$this->init();
 
 		$id = $entry['id'];
 


susnux:
This does not work and will break some places, because `init` is required to be called before `add`.
`init` will initialize e.g. the default app and customAppOrder.

Instead this can be easily fixed on the app, simply pass a closure to `add`, then `init` is not called.

sorbaugh (approved):

susnux (changes requested):
This does not work and will break some places, because init is required to be called before add.
init will initialize e.g. the default app and customAppOrder.

Instead this can be easily fixed on the app, simply pass a closure to add, then init is not called.

ArtificialOwl:
> This does not work and will break some places, because init is required to be called before add. init will initialize e.g. the default app and customAppOrder.
> 
> Instead this can be easily fixed on the app, simply pass a closure to add, then init is not called.

After a second reading of the code, can confirm that a Closure will avoid some unexpected behavior !

PR #47197, opened 2024-08-13
fix: update border radius values

marcoambrosini:
Update all instances of `--border-radius-pill` and `--border-radius-large` to `--border-radius-element`

`large` is already mapped to `element` so there's no visual change, just cleaner code

### DIFF
--- a/apps/dashboard/src/DashboardApp.vue
+++ b/apps/dashboard/src/DashboardApp.vue
@@ -610,7 +610,7 @@ export default {
 	background-position: 16px center;
 	padding: 12px 16px;
 	padding-left: 36px;
-	border-radius: var(--border-radius-pill);
+	border-radius: var(--border-radius-element);
 	max-width: 200px;
 	opacity: 1;
 	text-align: center;
@@ -660,7 +660,7 @@ export default {
 			width: 140px;
 			background-color: var(--color-background-hover);
 			border: 2px solid var(--color-main-background);
-			border-radius: var(--border-radius-large);
+			border-radius: var(--border-radius-element);
 			text-align: left;
 			overflow: hidden;
 			text-overflow: ellipsis;
--- a/apps/files/src/components/DragAndDropNotice.vue
+++ b/apps/files/src/components/DragAndDropNotice.vue
@@ -230,7 +230,7 @@ export default defineComponent({
 		max-height: 70%;
 		padding: 0 5vw;
 		border: 2px var(--color-border-dark) dashed;
-		border-radius: var(--border-radius-large);
+		border-radius: var(--border-radius-element);
 	}
 }
 
--- a/apps/settings/css/settings.scss
+++ b/apps/settings/css/settings.scss
@@ -137,7 +137,7 @@ select {
 	padding: 14px 20px;
 	background-color: var(--color-primary-element);
 	color: var(--color-primary-element-text);
-	border-radius: var(--border-radius-pill);
+	border-radius: var(--border-radius-element);
 	border: 1px solid var(--color-primary-element);
 	box-shadow: 0 2px 9px var(--color-box-shadow);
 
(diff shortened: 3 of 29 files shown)


apps/dashboard/src/DashboardApp.vue
```diff
@@ -610,7 +610,7 @@ export default {
 	background-position: 16px center;
 	padding: 12px 16px;
 	padding-left: 36px;
-	border-radius: var(--border-radius-pill);
+	border-radius: var(--border-radius-element);
 	max-width: 200px;
 	opacity: 1;
 	text-align: center;
```
susnux, line 616: The whole "edit-panels" rule can be removed, that is not used anymore.

apps/files/src/components/DragAndDropNotice.vue
```diff
@@ -230,7 +230,7 @@ export default defineComponent({
 		max-height: 70%;
 		padding: 0 5vw;
 		border: 2px var(--color-border-dark) dashed;
-		border-radius: var(--border-radius-large);
+		border-radius: var(--border-radius-element);
```
susnux, line 233: Should it not be `container`? It is a container and not an input element.

```suggestion
		border-radius: var(--border-radius-container);
```

apps/settings/css/settings.scss
```diff
@@ -137,7 +137,7 @@ select {
 	padding: 14px 20px;
```
susnux, line 137: Padding needs than to be adjusted, otherwise this looks weird.

```suggestion
	padding: calc(var(--default-grid-baseline) * 2) calc(var(--default-grid-baseline) * 3);
```

apps/settings/src/components/PersonalInfo/ProfileSection/EditProfileAnchorLink.vue
```diff
@@ -58,7 +58,7 @@ a {
 	line-height: 44px;
 	padding: 0 16px;
```
susnux, line 59: Similar here, the padding looks weird with a that small border radius.

```suggestion
	padding: 0 calc(var(--default-grid-baseline) * 2);
```

apps/systemtags/src/css/fileEntryInlineSystemTags.scss
```diff
@@ -15,7 +15,7 @@
 .files-list__system-tag {
 	padding: 5px 10px;
 	border: 1px solid;
```
susnux, line 17: I think this depends if we really want "edgy" chips¹, if yes we should also adjust the border radius of NcSelect options etc.

¹: https://github.com/nextcloud-libraries/nextcloud-vue/pull/5924

before | after
---|---
Screenshot 2024-08-13 at 15-25-53 Dateien - Nextcloud|Screenshot 2024-08-13 at 15-25-31 Dateien - Nextcloud


---

If we want the small radius, than we need to adjust the padding as well:

```suggestion
	padding: var(--default-grid-baseline) calc(2 * var(--default-grid-baseline));
	border: 1px solid;
```

core/css/apps.scss
```diff
@@ -334,7 +334,7 @@ kbd {
 				box-sizing: border-box;
 				white-space: nowrap;
 				text-overflow: ellipsis;
-				border-radius: var(--border-radius-pill);
+				border-radius: var(--border-radius-element);
```
susnux, line 337: Also here the padding should be adjusted to (padding is 2 lines above the comment able area):

```css
padding: 0 calc(2 * var(--default-grid-baseline));
```

core/src/components/UnifiedSearch/SearchResult.vue
```diff
@@ -111,7 +111,7 @@ $margin: 10px;
     ::v-deep a {
             border-radius: 12px;
```
susnux, line 112:
```suggestion
```

marcoambrosini:
closing as outdated

PR #47203, opened 2024-08-13
fix(AppMenu): Prevent menu entries from jumping on hover

### LINKED ISSUE #47148: [Bug]: Content shift when mouse is over top menu elements

### Bug description

https://github.com/user-attachments/assets/919c74ba-96d9-4cf4-87f3-9394967353c6

### Steps to reproduce

See video

### Expected behavior

Content shift should be avoided.

### Installation method

None

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

susnux:
* Resolves https://github.com/nextcloud/server/issues/47148


Pytal:
Tested and the icons don't move anymore but the text gets offset away from the hovered app which looks a bit strange 🤔
image

jancborchardt (commented):
Same as @Pytal noted – we could make the hovered/active app wider (like now) just without shifting the labels left and right if it, if possible? So that they keep centered below their icon.

susnux:
@jancborchardt if we make the hovered wider, without shifting, then everything will move. If you hover the middle entry of 3 app, then the right one will move to the right because the hovered grows.
That why I instead just shrink the text of the neighbors to allow the hovered one to grow without any movement of the app entries.

I could adjust it to only shrink / grow if needed (so the hovered text is ellipsed), but this is then not CSS only anymore and needs some JS.

jancborchardt:
@susnux so by making it wider I also only mean the text label, not the whole entry (so there is no jumping).
Can the neighbor labels just shrink in width instead of moving?

(Ideal would be some overlap or fade into transparency of the neighbor items, but these don't work with non-plain backgrounds.)

susnux:
@jancborchardt maybe like this:
1. Only grow if needed
2. Keep the label centered

The problem with 2 is, that now the label needs to be quite small (we "waste" the space on one side).

[Bildschirmaufnahme_20240814_120108.webm](https://github.com/user-attachments/assets/831068e7-4970-41c2-887f-d738367fbab1)

(commit pushed)

jancborchardt (approved):
@susnux yeah I understand the issue – your last proposal seems like the best compromise for now. :)
(Maybe for the future it would be good to use some JS indeed to polish it more, considering it is the main nav)

@skjnldsv @Pytal what do you think?

skjnldsv (approved):

PR #47259, opened 2024-08-15
feat: add negative compare-and-delete to imemcache

icewind1991:
Delete the key if the value does not equal some value.

Extracted from sharding work

### DIFF
--- a/lib/private/Memcache/CADTrait.php
+++ b/lib/private/Memcache/CADTrait.php
@@ -35,4 +35,21 @@ trait CADTrait {
 			return false;
 		}
 	}
+
+	public function ncad(string $key, mixed $old): bool {
+		//no native cad, emulate with locking
+		if ($this->add($key . '_lock', true)) {
+			$value = $this->get($key);
+			if ($value !== null && $value !== $old) {
+				$this->remove($key);
+				$this->remove($key . '_lock');
+				return true;
+			} else {
+				$this->remove($key . '_lock');
+				return false;
+			}
+		} else {
+			return false;
+		}
+	}
 }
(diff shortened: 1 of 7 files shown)


sorbaugh (approved):

lib/public/IMemcache.php
```diff
@@ -67,10 +67,24 @@ public function cas($key, $old, $new);
...
 	public function cad($key, $old);
+
+	/**
+	 * Negative compare and delete
+	 *
+	 * Delete $key if the stored value is not equal to $old
+	 *
+	 * @param string $key
+	 * @param mixed $old
+	 * @return bool
```
come-nc, line 86: Add a description for the return, what does the bool mean? false is failure or not found?

lib/public/IMemcache.php
```diff
@@ -67,10 +67,24 @@ public function cas($key, $old, $new);
...
+	 * @since 30.0.0
+	 */
+	public function ncad(string $key, $old): bool;
```
come-nc, line 89:
```suggestion
	public function ncad(string $key, mixed $old): bool;
```
In these cases where we really accept anything we should strongtype it as well, `mixed` is available in `PHP>=8.0`

lib/private/Memcache/CADTrait.php
```diff
@@ -35,4 +35,20 @@ public function cad($key, $old) {
 			return false;
 		}
 	}
+
+	public function ncad(string $key, $old): bool {
+		//no native cas, emulate with locking
```
come-nc, line 40: What is cas, is that a typo for cad?

come-nc (approved):

Issue #45440, opened 2024-05-22
Allow to specify the token of a public share / share renaming


artonge:
- could be a security issue. Mitigations:
   - add a warning?
   - optionally enforce password protection?
   - enforce a minimum set of dictionary words? (minimal 2, optionally configurable?)
- makes no sense for bigger server, so only for community edition
- What is the migration plan when going to Enterprise version?

szaimen:
So should this be a separate app so that it can be removed during building the enterprise archive?

jospoortvliet:
@szaimen yes.
Ideally we adapt and include https://apps.nextcloud.com/apps/cfg_share_links or https://apps.nextcloud.com/apps/sharerenamer in the community edition.

miaulalala:
Dictionary attacks would become a concern, yes. Token enumeration is already possible if highly unlikely, although I assume the endpoint is brute force protected, so it's expensive and time consuming.

Could also lead to token leaks as two identical tokens aren't possible, so a (guest) user might try creating shares at random to guess at other tokens. If we add the `userId` to the token, that could work as a preventative measure so each token only needs to be unique to each user space. The token controller will need to first check the `userId` against the logged-in user so somebody can't fake the `userId` in a request and try collision based guessing that way, but then the benefit of having a custom token is minimized.

If we do this, I'd recommend enforcing a password. 

I'd also suggest a minimum length for the token. Not great if the user chooses something really short.

In short, there are definitely some risks to this, and I can't really see the benefits. Maybe a link shortener integration could be a viable alternative?

jospoortvliet:
> Dictionary attacks would become a concern, yes. Token enumeration is already possible if highly unlikely, although I assume the endpoint is brute force protected, so it's expensive and time consuming.
> 
> Could also lead to token leaks as two identical tokens aren't possible, so a (guest) user might try creating shares at random to guess at other tokens. If we add the `userId` to the token, that could work as a preventative measure so each token only needs to be unique to each user space. The token controller will need to first check the `userId` against the logged-in user so somebody can't fake the `userId` in a request and try collision based guessing that way, but then the benefit of having a custom token is minimized.
> 
> If we do this, I'd recommend enforcing a password.
> 
> I'd also suggest a minimum length for the token. Not great if the user chooses something really short.
> 
> In short, there are definitely some risks to this, and I can't really see the benefits. Maybe a link shortener integration could be a viable alternative?

[...]

PR #47265, opened 2024-08-15
feat(share): make sharelink token length configurable

### LINKED ISSUE #419: Make configurable length of shared link token
For example now, I have this: 
`l33t.host/s/CpjUhnJuFEyMQzn`

And we can have this:
`l33t.host/s/AvJHo`

But It's my personal service for maybe 1000 or 10000 (max!) shared files.
And for this aim even 4-5 characters is enough. Why 15?

In Owncloud we don't have theming, now we have, and it's very cool. So why not?

ernolf:
* Resolves: #419 

## Summary
The share-token length is currently fixed at 15 characters as a constant.
This length allows for $52^{15}$ (54,960,434,128,018,667,122,720,768) possible token variations using CHAR_HUMAN_READABLE:
https://github.com/nextcloud/server/blob/dcdb4bbf8ab2d618e789a631a410ad70829766aa/lib/public/Security/ISecureRandom.php#L55
This level of variation can be considered Military Grade High Entropy and is often much more than sufficient for many use cases. Reducing the token length requires direct code modifications.

## Changes
This Pull Request introduces the ability to configure the token length for the Share API.
Administrators can set the token length dynamically through the database in a range from 6 up to 32:

```bash
occ config:app:set --value=8 -- core shareapi_token_length
```
This example sets the token length to 8 characters. 
If no changes are made, the default value of 15 will continue to be used, ensuring backward compatibility.

Even at the minimum length of 6, there are still $52^6$ (19,770,609,664) possible variations.

## Security Consideration
Reducing the token length increases the chance of token collisions, as it drastically reduces the number of possible combinations. It’s important to note that when a variable token length is used, this does not automatically mean, that the security decreases, since the total number of possible variations **increases** significantly. This is because, for shorter tokens, we must sum the possible variations for each length from 6 up to the maximum configured length to get the number of theoretically possible variations.
I have been using a token length of 8 for a long time now, although there are still 53,459,728,531,456 different possible variations. Absolutely sufficient for my use case where music producers share music samples and the shares have a very short ttl..
Shorter token lengths can be very helpful in cases where a lot of work is done with short-lived shares.
The admin is therefore responsible for assessing which security requirements he attaches to the token length. Therefore, this PR makes it possible not only to decrease- but to increase the token length up to a maximum of 32 as well, which corresponds to 8,167,835,760,036,914,488,254,418,108,462,708,901,695,678,621,570,564,096 possible variations, which is absolutely an increase in security.

---
**EDIT:**
Added the values:

~* "max" for the maximum token length (32)~  *)
~* "min" for the minimum token length (6)~  *)
* '0' and any other non-numerical value (like "default") for the default token length (15)

This ensures that the token length is "idiot-proof," defaulting to a safe (the default) value not only when the input is empty but also when any invalid value is provided. This makes it more robust and prevents potential errors.

*) simplified with https://github.com/nextcloud/server/pull/47265/commits/ee24fdd276df5ec44323aec6105577e623348b01

---
**EDIT 2:**

I had mistakenly assumed A-Za-z0-9 instead of CHAR_HUMAN_READABLE

---
## Todo
Consideration must be given to how this should be documented.

### DIFF
--- a/lib/private/Share/Constants.php
+++ b/lib/private/Share/Constants.php
@@ -63,7 +63,10 @@ class Constants {
 
 	public const RESPONSE_FORMAT = 'json'; // default response format for ocs calls
 
-	public const TOKEN_LENGTH = 15; // old (oc7) length is 32, keep token length in db at least that for compatibility
+	public const MIN_TOKEN_LENGTH = 6; // 19,770,609,664 different possible variations
+	public const DEFAULT_TOKEN_LENGTH = 15; // 54,960,434,128,018,667,122,720,768 different possible variations
+	public const MAX_TOKEN_LENGTH = 32; // 8,167,835,760,036,914,488,254,418,108,462,708,901,695,678,621,570,564,096 different possible variations
+	public const TOKEN_LENGTH = self::DEFAULT_TOKEN_LENGTH; // old (oc7) length is 32, keep token length in db at least that for compatibility
 
 	protected static $shareTypeUserAndGroups = -1;
 	protected static $shareTypeGroupUserUnique = 2;
--- a/lib/private/Share/Helper.php
+++ b/lib/private/Share/Helper.php
@@ -126,4 +126,13 @@ class Helper extends \OC\Share\Constants {
 
 		return false;
 	}
+
+	public static function getTokenLength(): int {
+		$config = \OCP\Server::get(\OCP\IAppConfig::class);
+		$tokenLength = $config->getValueInt('core', 'shareapi_token_length', self::DEFAULT_TOKEN_LENGTH);
+		$tokenLength = $tokenLength ?: self::DEFAULT_TOKEN_LENGTH;
+
+		// Token length should be within the defined min and max limits
+		return max(self::MIN_TOKEN_LENGTH, min($tokenLength, self::MAX_TOKEN_LENGTH));
+	}
 }
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -656,13 +656,43 @@ class Manager implements IManager {
 				$this->linkCreateChecks($share);
 				$this->setLinkParent($share);
 
-				// For now ignore a set token.
-				$share->setToken(
-					$this->secureRandom->generate(
-						\OC\Share\Constants::TOKEN_LENGTH,
-						\OCP\Security\ISecureRandom::CHAR_HUMAN_READABLE
-					)
-				);
+				// Initial token length
+				$tokenLength = \OC\Share\Helper::getTokenLength();
+
+				do {
+					$tokenExists = false;
+
+					for ($i = 0; $i <= 2; $i++) {
+						// Generate a new token
+						$token = $this->secureRandom->generate(
+							$tokenLength,
+							\OCP\Security\ISecureRandom::CHAR_HUMAN_READABLE
+						);
+
+						try {
+							// Try to fetch a share with the generated token
+							$this->getShareByToken($token);
+							$tokenExists = true; // Token exists, we need to try again
+						} catch (\OCP\Share\Exceptions\ShareNotFound $e) {
+							// Token is unique, exit the loop
+							$tokenExists = false;
+							break;
+						}
+					}
+
+					// If we've reached the maximum attempts and the token still exists, increase the token length
+					if ($tokenExists) {
+						$tokenLength++;
+
+						// Check if the token length exceeds the maximum allowed length
+						if ($tokenLength > \OC\Share\Constants::MAX_TOKEN_LENGTH) {
+							throw new \Exception('Unable to generate a unique share token. Maximum token length exceeded.');
+						}
+					}
+				} while ($tokenExists);
+
+				// Set the unique token
+				$share->setToken($token);
 
 				// Verify the expiration date
 				$share = $this->validateExpirationDateLink($share);


lib/private/Share/Constants.php
```diff
@@ -63,7 +63,10 @@ class Constants {
 
 	public const RESPONSE_FORMAT = 'json'; // default response format for ocs calls
 
-	public const TOKEN_LENGTH = 15; // old (oc7) length is 32, keep token length in db at least that for compatibility
+	public const MIN_TOKEN_LENGTH = 4; // 14,776,336 different possible variations
```
skjnldsv, line 66: Security wise, 14M isn't much variations I would say :/

ernolf: > [skjnldsv](https://github.com/skjnldsv)
Security wise, 14M isn't much variations I would say :/

Yes. I agree that it is not much, but I would like to point out that it is still much safer than using URL shortening services, as is often done by users, or self-generated guessable links using the apps available for this purpose.

Anyone who uses a token length of 4 must know what they are getting into. That such a token length has nothing to do with security and should only be used for extremely short-lived links with no security-related significance.

I have deliberately left it this low here because it is the minimum length that can really be used, but of course, it can also be limited to 5 (916,132,832 possible variations) or 6 (56,800,235,584 possible variations).

Reference: https://github.com/nextcloud/server/issues/419#issuecomment-2291982852

ernolf

juliusknorr (approved):
Code looks good to me, but would leave deciding that to the security people 👍

lib/private/Share/Constants.php
susnux: It is not 14M variations, it is just 7M.
`CHAR_HUMAN_READABLE` contains 52 characters, with a length of 4 this is only `52 ^ 4 = 7'311'616`.

---

I think there is a high risk of causing collisions esp. by human error (e.g. capitalization) when the token is that short. Especially as this makes it less secure for all users on that instance, as URL guessing is now somewhat easy.

I think the use case / requirement to have memorable URLs is very limited, mostly you copy-paste the URL so the length does not matter.
If you **need** it, lets say because you use the URL in print, then [the custom token feature](https://github.com/nextcloud/server/issues/45440) seems to be better suited.

ernolf: > @susnux 
> It is not 14M variations, it is just 7M.
> `CHAR_HUMAN_READABLE` contains 52 characters, with a length of 4 this is only `52 ^ 4 = 7'311'616`.

A-Za-z0-9 = 64

please read the referenced comment.

ernolf: Excuse me so much, you are right
https://github.com/nextcloud/server/blob/dcdb4bbf8ab2d618e789a631a410ad70829766aa/lib/public/Security/ISecureRandom.php#L55
My fault!

Then I would suggest a minimum token length of 6, correspinding 19,770,609,664

skjnldsv: I'm not too convinced yet. There is already a community app to rename tokens. Maybe that is enough for users?

ernolf: > I'm not too convinced yet. There is already a community app to rename tokens. Maybe that is enough for users?

I use both. Token shortening and configurable share links.

I created a script in response to a request in the comunity forum that can set the constant to a selectable value between 4 and 32, which has been downloaded a lot. So I definitely see a demand for it.

skjnldsv: I don't doubt the demand. But security wise, it's very often users will find loopholes or try to simplify things resulting in lower security (aka password on a post-it). Would offering that option from Nextcloud straight away be a real gift?

skjnldsv:
> ```shell
> occ config:app:set --value=8 -- core shareapi_token_length
> ```

Also, as a side note, we tend to avoid those kind of config setters. They're very hidden and not very discoverable.
Either they go into the config.php or we create a UI for it :)

lib/private/Share/Constants.php
ernolf: > Would offering that option from Nextcloud straight away be a real gift?

I understand your concerns. What are the alternatives? Either dirty hacks like my script are used, or URL shortening services are relied upon, which I would consider much less secure. With `cfg_share_links`, security is even lower or practically non-existent from this perspective.

For this reason, I deliberately avoided considering a variable that could be set in the frontend. If this option is to be documented, the security implications must be clearly highlighted. I believe that Nextcloud admins are generally quite security-conscious, and those who prioritize security will certainly not shorten the token length, and might even increase it!

However, I can also live with a minimum token length of 8, though 15 is really a thorn in my side because it's absolute overkill for so many use cases.

skjnldsv: That's why we do not officially recommend either solutions (until now) :)
Anyway, I'll leave that decision to the security team :+1:

susnux: >  Anyway, I'll leave that decision to the security team 👍

Same here.

But @ernolf  I am interested in the reasons behind this, because I can not see the use case of generally shorter share links. For me:
1. Either you share by link ("share digitally") -> length does not matter as you copy paste / click on it
2. Or you share analog (printed or similar), but then random characters are already bad regardless of the length, so a custom link would be useful ("custom share token" feature)

So yes I would be really interested in the use case :)

(Examples for 2 would be e.g. "Upload your pictures from today" or similar things)

ernolf: > @susnux
I am interested in the reasons behind this, because I can not see the use case of generally shorter share links.

Here a thread (in german) about this:
https://help.nextcloud.com/t/freigabe-tooken-bei-share-api-nutzung-kurzen/162967

.. that was the initial reason why I wrote `nc-sharetoken-length` [(part of ebtb)](https://github.com/ernolf/ebtb)

My use case is in the music industry. Huge amounts of music samples are sent to music producers around the world. The envolved users love to simply click on the [+]  to create a link and would never create a link with cfg_share_link in that ammount of created shares. The link tokens used are then fed into scripts that analyze the log files in order to determine exactly when and from where which link token was used, i.e. called up and if the sample was completely downloaded.

All people involved in the process are unanimous in their opinion that a token length of 15 is much too long. Thus I changed the token length to 8

The `nc-shares` script (ebtb as well) creates lists in which the tokens make up a column. If such a column is too wide, working with it becomes unwieldy.

And yes, all of this is possible with longer link tokens too, but it's also an aesthetic issue. Shorter links simply look better :wink:

I am also in direct competition with Dropbox, which uses a token length of 21 characters. The ability to create shorter tokens has given me a significant 'head start' in this area.

`cfg_share_link` has a minimum of 3

> Or you share analog (printed or similar), but then random characters are already bad regardless of the length

But that contradicts the CHAR_HUMAN_READABLE group, because then one could just as well use A-Za-z0-9 and wouldn't need to shorten the characters used by 10, which would allow a higher bit depth with the same token length.
When printing something, QR codes are increasingly used these days, which are now also provided in the UI.

susnux: > But that contradicts the CHAR_HUMAN_READABLE group

its more like remembering `/sgwhurh` is harder than `/emmas-30th` (to stick with the example of uploading photos).

But thank you for the insights! This helps a lot understanding the demands of Nextcloud users :)

ernolf:
> Either they go into the config.php or we create a UI for it :)

I mainly wanted to create this possibility first. I was aware that there would be many objections. For me personally, it would be sufficient if this was handled as a "hidden feature". If it is generally accepted, I would of course be happy to write the integration in the UI within the scope of my skills.

lib/private/Share/Helper.php
```diff
@@ -126,4 +126,29 @@ public static function isSameUserOnSameServer($user1, $server1, $user2, $server2
...
+		}
+
+		// Token length should be within the defined min and max limits
+		if ($tokenLength < self::MIN_TOKEN_LENGTH) {
+			$tokenLength = self::MIN_TOKEN_LENGTH;
+		} elseif ($tokenLength > self::MAX_TOKEN_LENGTH) {
+			$tokenLength = self::MAX_TOKEN_LENGTH;
+		}
+
+		return $tokenLength;
```
nickvergessen, line 152:
```suggestion
		return max(self::MIN_TOKEN_LENGTH, min($tokenLength, self::MAX_TOKEN_LENGTH));
```

lib/private/Share/Helper.php
```diff
@@ -126,4 +126,29 @@ public static function isSameUserOnSameServer($user1, $server1, $user2, $server2
...
+		// Check for 'min', 'max', or non-numeric input
+		if (strtolower($tokenLengthValue) === 'min') {
+			$tokenLength = self::MIN_TOKEN_LENGTH;
+		} elseif (strtolower($tokenLengthValue) === 'max') {
+			$tokenLength = self::MAX_TOKEN_LENGTH;
+		} elseif (!is_numeric($tokenLengthValue) || (int)$tokenLengthValue === 0) {
+			$tokenLength = self::DEFAULT_TOKEN_LENGTH;
+		} else {
+			$tokenLength = (int)$tokenLengthValue;
+		}
```
nickvergessen, line 143: Simplicity is key. Let's simply only allow integers:
```suggestion
		$config = \OCP\Server::get(\OCP\IAppConfig::class);
		$tokenLength = $config->getValueInt('core', 'shareapi_token_length', self::DEFAULT_TOKEN_LENGTH);
		$tokenLength = $tokenLength ?: self::DEFAULT_TOKEN_LENGTH;
```

lib/private/Share/Helper.php
ernolf: resolved with https://github.com/nextcloud/server/pull/47265/commits/ee24fdd276df5ec44323aec6105577e623348b01

lib/private/Share/Helper.php
ernolf: Resolved with https://github.com/nextcloud/server/pull/47265/commits/ee24fdd276df5ec44323aec6105577e623348b01

nickvergessen (approved):

PR #47277, opened 2024-08-16
[occ:bg-job:worker] Add 'stop_after' option

julien-nc:
If the `stop_after` option is set, the worker will stop itself after the time set with the option value.

As the worker does not update the loaded code (after an app upgrade for example) and the app setting values, we will recommend to regularly restart it. This option makes it possible to stop the worker after some time while it is not running a job.

IMO this should be backported to 30 as the worker is less useful without this option.

### DIFF
--- a/core/Command/Background/JobWorker.php
+++ b/core/Command/Background/JobWorker.php
@@ -52,10 +52,25 @@ class JobWorker extends JobBase {
 				'Interval in seconds in which the worker should repeat already processed jobs (set to 0 for no repeat)',
 				5
 			)
+			->addOption(
+				'stop_after',
+				't',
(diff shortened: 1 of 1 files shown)


marcelklehr (approved):

bigcat88 (approved):
I agree that this should be backported in Nextcloud 30

skjnldsv (approved):

juliushaertl:
Maybe something for later but wouldn't it be possible to detect if any app version changed and only exit in that case? Similar how we build the cache buster for frontend assets based on the server and all app versions?

juliusknorr (approved):

PR #47290, opened 2024-08-16
feat(templates): checkbox field type

elzody:
## Summary
This PR will introduce a checkbox type when filling templates.
Part of [nextcloud/richdocuments#3917](https://github.com/nextcloud/richdocuments/issues/3917)

https://github.com/user-attachments/assets/bdf78248-b917-4d29-a5f8-6e0a8118349a

## Note
Tests are implemented from the richdocuments side: [nextcloud/richdocuments#3919](https://github.com/nextcloud/richdocuments/pull/3919)

## Checklist


jancborchardt (changes requested):
Very nice! :) The switch should be left of the text though as per component, no? https://nextcloud-vue-components.netlify.app/#/Components/NcCheckboxRadioSwitch

image

elzody:
@jancborchardt Indeed, it was a bit foolish of me to change this in retrospect. Will change and re-request review when ready :)

lib/public/Files/Template/Field.php
```diff
@@ -13,36 +13,39 @@
...
 	 * @since 30.0.0
 	 */
-	public function __construct(string $index, string $content, FieldType $type, ?string $alias = null, ?int $id = null, ?string $tag = null) {
-		$this->index = $index;
-		$this->alias = $alias;
-		$this->type = $type;
-		$this->id = $id;
-		$this->tag = $tag;
-		$this->content = $content;
+	public function __construct(
```
juliushaertl, line 25: So unfortunately since we are in API freeze this is nothing we can merge normally. However since the richdocuments app is the only app using it, it would be ok from my perspective.

FYI @AndyScherzinger In case you'd speak out against that.

lib/public/Files/Template/Field.php
```diff
@@ -13,36 +13,39 @@
...
-	private string $content;
-	private FieldType $type;
-	private ?string $alias;
-	private ?int $id;
-	private ?string $tag;
+	public ?string $content = null;
+	public ?string $alias = null;
+	public ?string $tag = null;
+	public ?int $id = null;
+	public ?bool $checked = null;
```
juliushaertl, line 20: Do we rather wan to reuse $value and make it a mixed type? That would make it more flexible if we extend further with a `\DateTime` object or others in the future.

lib/public/Files/Template/Field.php
skjnldsv: > So unfortunately since we are in API freeze this is nothing we can merge normally

Because this is planned to be backported?

AndyScherzinger: Yes, since this is build and used only by richdocuments and used for the templating implementation it ships starting with 30.

lib/public/Files/Template/Field.php
```diff
@@ -1,48 +1,55 @@
...
  */
 
+declare(strict_types=1);
+
 namespace OCP\Files\Template;
 
 /**
  * @since 30.0.0
  */
 class Field implements \JsonSerializable {
```
juliushaertl, line 15: Can be abstract if we now always define a subclass for each type

lib/public/Files/Template/Field.php
```diff
@@ -1,48 +1,55 @@
...
 	 * @since 30.0.0
 	 */
-	public function __construct(string $index, string $content, FieldType $type, ?string $alias = null, ?int $id = null, ?string $tag = null) {
-		$this->index = $index;
-		$this->alias = $alias;
-		$this->type = $type;
-		$this->id = $id;
-		$this->tag = $tag;
-		$this->content = $content;
+	public function setValue(mixed $value): void {
```
juliushaertl, line 32: Can also be abstract

lib/public/Files/Template/Field.php
```diff
@@ -1,48 +1,55 @@
...
+			}
+
+			if ($propertyValue instanceof FieldType) {
+				$propertyValue = $propertyValue->value;
+			}
+
+			array_push($jsonProperties, [$propertyName => $propertyValue]);
+		}
+		
+		return array_merge([], ...$jsonProperties);
```
juliushaertl, line 53: I think with the individual classes we do not need the dynamic property iteration here.

juliusknorr (commented):
Generally fine, just some small comments

jancborchardt (approved):
Looks nice design-wise! :)

kesselb (approved):

apps/files/src/components/TemplateFiller.vue
```diff
@@ -29,11 +26,12 @@
 	</NcModal>
 </template>
 
-<script lang="ts">
+<script>
```
susnux, line 29: why reverting to JS?

apps/files/src/components/TemplateFiller.vue
```diff
@@ -65,10 +64,21 @@ export default defineComponent({
 
 	methods: {
 		t,
-		trackInput([value, index]) {
-			this.localFields[index] = {
-				content: value,
+		trackInput([index, name, value]) {
```
susnux, line 67: I would recommend not to use positional array for this but names objects, makes spotting issues easier :sweat_smile:

apps/files/src/components/TemplateFiller.vue
elzody: Wasn't using TypeScript in this file, but if it makes it more consistent with the rest of the files I can leave it as TS.

apps/files/src/components/TemplateFiller.vue
elzody: Do you mean then emitting something like this instead?

```diff
diff --git a/apps/files/src/components/TemplateFiller/TemplateRichTextField.vue b/apps/files/src/components/TemplateFiller/TemplateRichTextField.vue
index e794abd7e16..a6137a1df36 100644
--- a/apps/files/src/components/TemplateFiller/TemplateRichTextField.vue
+++ b/apps/files/src/components/TemplateFiller/TemplateRichTextField.vue
@@ -15,7 +15,7 @@
                        :label="fieldLabel"
                        :label-outside="true"
                        :placeholder="field.content"
-                       @input="$emit('input', [field.index, 'content', value])" />
+                       @input="input" />
        </div>
 </template>
 
@@ -53,6 +53,16 @@ export default defineComponent({
                        return 'text-field' + this.field.index
                },
        },
+       
+       methods: {
+               input() {
+                       $emit('input', {
+                               index: this.field.index,
+                               property: 'content',
+                               value: this.value,
+                       })
+               }
+       },
 })
 </script>
```

PR #47293, opened 2024-08-16
feat(PsrLoggerAdapter): Allow to use `Psr\Log\LogLevel` for `log` method

susnux:
## Summary

There is the `Psr\Log\LogLevel` class defining log level constants, to be fully compatible we should at least support those logging levels. Moreover this is the last part that was still required from `ILogger` interface, as we did not have alternatives for the log level constants.

## Checklist

### DIFF
--- a/lib/private/Log/PsrLoggerAdapter.php
+++ b/lib/private/Log/PsrLoggerAdapter.php
@@ -14,6 +14,7 @@ use OCP\ILogger;
 use OCP\Log\IDataLogger;
 use Psr\Log\InvalidArgumentException;
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
 use Throwable;
 use function array_key_exists;
 use function array_merge;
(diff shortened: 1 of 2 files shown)


come-nc (commented):
I’m not sure I get the point, those const are strings and not int, so we will still not be able to use them instead of the ILogger ones?

The problem is that there is no standard for int levels.
We should host those const elsewhere or undeprecate them in my opinion. But core and apps need a way to reference log levels by name. 

Should we add a backed enum in OCP for that :nerd_face: ?

susnux:
> I’m not sure I get the point, those const are strings and not int, so we will still not be able to use them instead of the ILogger ones?

Well this adds support so that you can call:

```php
LoggerInterface $logger;

// ...

$logger->log(LogLevel::WARNING, 'this is a warning')
$logger->log(LogLevel::DEBUG, 'this is a debug message')
```

So everywhere¹² you use the `ILogger` levels you can now also use the PSR levels.

---

¹ Well not if directly using the `Log` class but that is private and we should then also remove `ILogger` from public scope.
² And of course `level` set in the context would be a "magic" number, this is a problem if you use the "new log message"-event.

ChristophWurst (approved):
Thanks! This is great when passing our logger instance into a library that speaks to psr/log but obviously has no idea about our constants

come-nc:
There is less usage of those consts than I thought. I was expecting to find a lot of `>=` comparison and there is none :thinking: 
Still, the config entry loglevel will stay an int, right?

lib/private/Log/PsrLoggerAdapter.php
```diff
@@ -213,8 +228,11 @@ public function debug($message, array $context = []): void {
 	 * @throws InvalidArgumentException
 	 */
 	public function log($level, $message, array $context = []): void {
+		if (is_string($level)) {
+			$level = self::logLevelToInt($level);
+		}
```
come-nc, line 233: You should apply the same for `$context['level']` I think to avoid problems if the context contains a string level. `logException` will misbehave if it gets a string level.

come-nc:
```suggestion
		if (is_string($level)) {
			$level = self::logLevelToInt($level);
		}
		if (is_string($context['level'])) {
			$context['level'] = self::logLevelToInt($context['level']);
		}
```

come-nc (approved):

PR #47294, opened 2024-08-16
fix(files): Sort tree views correctly by natural order

Pytal:
## Summary

- Correctly sort nodes by natural order instead of order=0 i.e. creating a new folder always appended it to the bottom of the level instead of correct natural order
- Render nodes by natural order as nodes returned from endpoint have an undefined order

## Requires

- https://github.com/nextcloud-libraries/nextcloud-files/pull/1053

## Checklist


apps/files/src/services/FolderTree.ts
```diff
@@ -35,8 +38,19 @@ export const folderTreeId = 'folders'
 
 export const sourceRoot = `${davRemoteURL}/files/${getCurrentUser()?.uid}`
 
+const collator = Intl.Collator(
```
susnux, line 41: Maybe instead simply use `orderBy` from nextcloud files?

susnux (approved):

artonge (approved):

PR #47323, opened 2024-08-19
feat: Provide RFC 5424 compliant `\OCP\LogLevel` with PSR-3 support

susnux:
## Summary
(This is an alternative to the PSR loglevel PR)

Make our logging system PSR-3 compliant by supporting the `\Psr\Log\LogLevel` and publish our supported the log levels in OCP so we can now possibly remove `\OCP\ILogger`

## Checklist

### DIFF
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -448,9 +448,11 @@ class Log implements ILogger, IDataLogger {
 	 * This is required as we currently only support 5 levels from 8 levels specified in the RFC.
 	 */
 	protected function sanitizeLogLevel(int|LogLevel $level): int {
+		// If the log level is an integer and outside of the internal supported range, try if it is just the value of the `LogLevel` enum
 		if (is_int($level) && ($level < ILogger::DEBUG || $level > ILogger::FATAL)) {
 			$level = LogLevel::tryFrom($level);
 		}
+		// If the level is a case of the `LogLevel` enum convert to the supported internal log level (map 0-7 to 0-4)
 		if ($level instanceof LogLevel) {
 			return match ($level) {
 				LogLevel::DEBUG => ILogger::DEBUG,
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -448,7 +448,7 @@ class Log implements ILogger, IDataLogger {
 	 * This is required as we currently only support 5 levels from 8 levels specified in the RFC.
 	 */
 	protected function sanitizeLogLevel(int|LogLevel $level): int {
-		if ($level < ILogger::DEBUG || $level > ILogger::FATAL) {
+		if (is_int($level) && ($level < ILogger::DEBUG || $level > ILogger::FATAL)) {
 			$level = LogLevel::tryFrom($level);
 		}
 		if ($level instanceof LogLevel) {
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -465,6 +465,7 @@ class Log implements ILogger, IDataLogger {
 				LogLevel::EMERGENCY => ILogger::FATAL,
 			};
 		}
+		// If the log level is not a valid internal log level the `tryFrom` has returned `null` so we fall back to the default log level (which is `warning`).
 		return $level ?? ILogger::WARN;
 	}
 }
--- a/lib/private/Log.php
+++ b/lib/private/Log.php
@@ -20,6 +20,7 @@ use OCP\Log\BeforeMessageLoggedEvent;
 use OCP\Log\IDataLogger;
 use OCP\Log\IFileBased;
 use OCP\Log\IWriter;
+use OCP\LogLevel;
 use OCP\Support\CrashReport\IRegistry;
 use Throwable;
 use function array_merge;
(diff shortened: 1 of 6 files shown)


lib/public/ILogger.php
```diff
@@ -19,27 +19,27 @@
...
 	/**
 	 * @since 14.0.0
-	 * @deprecated 20.0.0
+	 * @deprecated 20.0.0 - use \OCP\LogLevel::INFO
 	 */
 	public const INFO = 1;
 	/**
 	 * @since 14.0.0
-	 * @deprecated 20.0.0
+	 * @deprecated 20.0.0 - use \OCP\LogLevel::WARN
```
come-nc, line 32:
```suggestion
	 * @deprecated 20.0.0 - use \OCP\LogLevel::WARNING
```

lib/private/Log.php
```diff
@@ -441,4 +442,27 @@ protected function getSerializer(): ExceptionSerializer {
 		}
 		return $serializer;
 	}
+
+	/**
+	 * Helper needed to convert IETF / PSR-3 logging levels to internal logging level.
+	 * This is required as we currently only support 5 levels from 8 levels specified in the RFC.
+	 */
+	protected function sanitizeLogLevel(int|LogLevel $level): int {
+		if ($level < ILogger::DEBUG || $level > ILogger::FATAL) {
```
come-nc, line 451: There is an is_int test missing, no?

come-nc:
I would remove all method from ILogger and keep only the constants in there, until apps have the time to switch to the new enum.

But I’m afraid of how this works, as the ERROR in the enum is not backed by the same int as before, and same situation with fatal, transition will be messy, no?

lib/public/LogLevel.php
```diff
@@ -0,0 +1,69 @@
...
+	case CRITICAL = 5;
+	/** Action must be taken immediately */
+	case ALERT = 6;
+	/** System is unusable */
+	case EMERGENCY = 7;
+
+	/**
+	 * Create new `\OCP\LogLevel` from a given PSR-3 LogLevel
+	 * @param string $level the PSR-3 log level to convert
+	 * @throws InvalidArgumentException If the `$level` is not a PSR log level
```
susnux, line 38:
```suggestion
	 * @throws InvalidArgumentException If the `$level` is not a PSR log level
	 * @since 31.0.0
```

lib/public/LogLevel.php
```diff
@@ -0,0 +1,69 @@
...
+			PsrLogLevel::ERROR => self::ERROR,
+			PsrLogLevel::CRITICAL => self::CRITICAL,
+			PsrLogLevel::ALERT => self::ALERT,
+			PsrLogLevel::EMERGENCY => self::EMERGENCY,
+			default => throw new InvalidArgumentException($level . ' is not a valid PSR log level'),
+		};
+	}
+
+	/**
+	 * Convert this log level to a PSR-3 `\Psr\Log\LogLevel`
```
susnux, line 55:
```suggestion
	 * Convert this log level to a PSR-3 `\Psr\Log\LogLevel`
	 * @since 31.0.0
```

lib/private/Log.php
susnux: Yes you are right!

```suggestion
		if (is_int($level) && ($level < ILogger::DEBUG || $level > ILogger::FATAL)) {
```

lib/private/Log.php
```diff
@@ -441,4 +442,27 @@ protected function getSerializer(): ExceptionSerializer {
...
 	}
+
+	/**
+	 * Helper needed to convert IETF / PSR-3 logging levels to internal logging level.
+	 * This is required as we currently only support 5 levels from 8 levels specified in the RFC.
+	 */
+	protected function sanitizeLogLevel(int|LogLevel $level): int {
+		if (is_int($level) && ($level < ILogger::DEBUG || $level > ILogger::FATAL)) {
+			$level = LogLevel::tryFrom($level);
+		}
```
susnux, line 453:
```suggestion
		// If the log level is an integer and outside of the internal supported range, try if it is just the value of the `LogLevel` enum
		if (is_int($level) && ($level < ILogger::DEBUG || $level > ILogger::FATAL)) {
			$level = LogLevel::tryFrom($level);
		}
		// If the level is a case of the `LogLevel` enum convert to the supported internal log level (map 0-7 to 0-4)
```

lib/private/Log.php
```diff
@@ -441,4 +442,29 @@ protected function getSerializer(): ExceptionSerializer {
...
+				LogLevel::INFO => ILogger::INFO,
+				LogLevel::NOTICE => ILogger::INFO,
+				LogLevel::WARNING => ILogger::WARN,
+				LogLevel::ERROR => ILogger::ERROR,
+				LogLevel::ALERT => ILogger::ERROR,
+				LogLevel::CRITICAL => ILogger::ERROR,
+				LogLevel::EMERGENCY => ILogger::FATAL,
+			};
+		}
+		return $level ?? ILogger::WARN;
```
susnux, line 468:
```suggestion
		// If the log level is not a valid internal log level the `tryFrom` has returned `null` so we fall back to the default log level (which is `warning`).
		return $level ?? ILogger::WARN;
```

susnux:
lets keep it simple

PR #47324, opened 2024-08-19
perf(core): Add index for jobs last_checked, reserved_at, time_sensitive

### LINKED ISSUE #46126: [Bug]: Missing database indox on query for time sensitive background jobs

### Bug description

The query 4f0fc707a2e82ec699067a969cd17b05

```sql
SELECT
  *
FROM
  `oc_jobs`
WHERE
  (`reserved_at` <= ?)
  AND (`last_checked` <= ?)
  AND (`time_sensitive` = ?)
ORDER BY
  `last_checked` ASC
LIMIT
  ?
```

does not use a covering index. The index `job_lastcheck_reserved` only covers `last_checked` and `reserved_at`, not `time_sensitive`. The database has to scan a bunch of rows every time cron picks the next job outside the maintenance window:

image

### Steps to reproduce

1. Install Nextcloud
2. Configure a maintenance window

### Expected behavior

An index on `(last_checked, reserved_at, time_sensitive)`

### Installation method

None

### Nextcloud Server version

29

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

cc @nickvergessen

provokateurin:
* Resolves: #46126

## Summary

Replaces the two existing indices with a single shared index saving some time on queries previously using both separate indices.

## Checklist

### DIFF
--- a/core/Migrations/Version31000Date20240819122840.php
+++ b/core/Migrations/Version31000Date20240819122840.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC\Core\Migrations;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version31000Date20240819122840 extends SimpleMigrationStep {
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('jobs');
+
+		# Remove previous indices
+		if ($table->hasIndex('job_lastcheck_reserved')) {
+			$table->dropIndex('job_lastcheck_reserved');
+		}
+		if ($table->hasIndex('jobs_time_sensitive')) {
+			$table->dropIndex('jobs_time_sensitive');
+		}
+
+		# Add updated index
+		if (!$table->hasIndex('job_last_reserved_sensitive')) {
+			$table->addIndex(['last_checked', 'reserved_at', 'time_sensitive'], 'job_last_reserved_sensitive');
+		}
+
+		return $schema;
+	}
+}
--- a/core/Application.php
+++ b/core/Application.php
@@ -191,8 +191,8 @@ class Application extends App {
 
 			$event->addMissingIndex(
 				'jobs',
-				'job_lastcheck_reserved',
-				['last_checked', 'reserved_at']
+				'job_last_reserved_sensitive',
+				['last_checked', 'reserved_at', 'time_sensitive']
 			);
 
 			$event->addMissingIndex(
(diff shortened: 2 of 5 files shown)


core/Migrations/Version31000Date20240819122840.php
```diff
@@ -0,0 +1,45 @@
...
+			$table->dropIndex('job_lastcheck_reserved');
+		}
+		if ($table->hasIndex('jobs_time_sensitive')) {
+			$table->dropIndex('jobs_time_sensitive');
+		}
+
+		# Add updated index
+		if (!$table->hasIndex('job_last_reserved_sensitive')) {
+			$table->addIndex(['last_checked', 'reserved_at', 'time_sensitive'], 'job_last_reserved_sensitive');
+		}
```
nickvergessen, line 41:
```suggestion
```

This should be done only on new installations and otherwise the change in `core/Application.php` takes care?
The best to achieve this is modifying the original migration that added the old indexes (you should also comment them out with a reference that `Version31000Date20240819122840.php` deletes them).

provokateurin: So just keep the old indices?

core/Migrations/Version31000Date20240819122840.php
```diff
@@ -0,0 +1,45 @@
...
+		if ($table->hasIndex('job_lastcheck_reserved')) {
+			$table->dropIndex('job_lastcheck_reserved');
+		}
+		if ($table->hasIndex('jobs_time_sensitive')) {
+			$table->dropIndex('jobs_time_sensitive');
+		}
+
+		# Add updated index
+		if (!$table->hasIndex('job_last_reserved_sensitive')) {
+			$table->addIndex(['last_checked', 'reserved_at', 'time_sensitive'], 'job_last_reserved_sensitive');
```
ChristophWurst, line 40: Adding an index can be slow if there is a lot of data. Would it be possible to use https://docs.nextcloud.com/server/latest/developer_manual/basics/storage/migrations.html#replacing-indices instead?

ChristophWurst (changes requested):
``\OC\Core\Migrations\Version13000Date20170718121200::changeSchema`` still creates the old index. Comment it out and add the new index with a comment as well saying that the indices were changed later on

core/Migrations/Version31000Date20240819122840.php
nickvergessen: yeah, that's the first change she did in core/Application

core/Migrations/Version31000Date20240819122840.php
nickvergessen: No, deleting is fine. Just not creating new ones

core/Migrations/Version31000Date20240819122840.php
provokateurin: Ah I didn't know about the possibility to replace indices.

provokateurin:
So should I replace or only add the index in Application.php?

nickvergessen:
Correct way is:

1. Add `$event->addMissingIndex(` section in `core/Application.php`
2. Add migration that deletes the old index
3. Comment out the code that created the old index with a reference to the migration from 2. (could be in another migration file or in `core/Application.php` (like here)
4. Find the migration file that creates the table (or the last one that creates all necessary columns for your index) and add the index there.

core/Migrations/Version31000Date20240819122840.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace OC\Core\Migrations;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version31000Date20240819122840 extends SimpleMigrationStep {
```
AndyScherzinger, line 17: See https://github.com/nextcloud/server/pull/46974/files#diff-a13c08ddffcec08aa3dfe030b79869cd4b164cf35f9a4fff0e22dbe9848b552eR20 for examples, DB migrations for 30+ should be annotated for the new 30+ feature to work 👍

ChristophWurst:
> Correct way is:
> 
>     1. Add `$event->addMissingIndex(` section in `core/Application.php`
> 
>     2. Add migration that deletes the old index
> 
>     3. Comment out the code that created the old index with a reference to the migration from 2. (could be in another migration file or in `core/Application.php` (like here)
> 
>     4. Find the migration file that creates the table (or the last one that creates all necessary columns for your index) and add the index there.

I'm skeptical about this because it will leave instances in an unindexed state after the upgrade, until the admins apply optional indices.
We have the index replacement APIs for this case. The old index is not great but better than no index. Later on the admin can replace the index for best performance.

So my counter proposal is

1) Adjust the original migration to comment out the old index and add the new
2) Drop the old optional index logic
3) Advertise the new index as replacement for the old, if it exists

PR #47331, opened 2024-08-19
fix: adding blurhash to definitions

skalidindi53:
* Ref https://github.com/nextcloud/spreed/pull/13009

## Summary
- [x] Added blurhash to RichObjectString definitions

## Checklist

### DIFF
--- a/lib/public/RichObjectStrings/Definitions.php
+++ b/lib/public/RichObjectStrings/Definitions.php
@@ -368,6 +368,12 @@ class Definitions {
 					'description' => 'The height in pixels if the file is an image',
 					'example' => '1080',
 				],
+				'blurhash' => [
+					'since' => '30.0.0',
+					'required' => false,
+					'description' => 'The blurhash of the image',
+					'example' => 'LEHV9uae2yk8pyo0adR*.7kCMdnj',
+				],
 			],
 		],
 		'forms-form' => [


provokateurin:
Also the changes in lib/composer have to be reverted

provokateurin:
Ah I see what is going on, the lockfiles in vendor-bin have been changed and that is causing the OpenAPI changes. So please remove all changes except for lib/public.

lib/public/RichObjectStrings/Definitions.php
```diff
@@ -368,6 +368,12 @@ class Definitions {
 					'description' => 'The height in pixels if the file is an image',
 					'example' => '1080',
 				],
+				'blurhash' => [
+					'since' => '29.0.4',
```
provokateurin, line 372: This version is not correct, it should be the first version where https://github.com/nextcloud/spreed/pull/13009 is merged (so probably 30.0.0 because it denotes the server version and not the talk version).

provokateurin (changes requested):

lib/public/RichObjectStrings/Definitions.php
skalidindi53: Thank you so much for the help, but I accidentally set DCO to pass even though it shouldn't and I'm not sure how to revert it

skalidindi53: But I'm not really sure how to fix these last 4 failing checks

nickvergessen (approved):
Reverted all unrelated parts

provokateurin (approved):

PR #47340, opened 2024-08-20
fix(federation): Do not overwrite certificate bundle

juliusknorr:
Signed-off-by: Julius Härtl <jus@bitgrid.net>

* Resolves: # 

## Summary

One might still want to verify self-signed certificates and not disable verification entirely. Before this change the `verify` option was always overwritten with true which means that the system certificate store was used instead of the Nextcloud built in one. We should only pass verify if we actually intent to disable verification, else the http client is injecting the cert bundle properly in https://github.com/nextcloud/server/blob/dae7c159f728a90ffa53247d6e033abdae5d2bd6/lib/private/Http/Client/Client.php#L55

## Checklist

### DIFF
--- a/apps/files_sharing/lib/External/Storage.php
+++ b/apps/files_sharing/lib/External/Storage.php
@@ -259,19 +259,12 @@ class Storage extends DAV implements ISharedStorage, IDisableEncryptionStorage, 
 
 		$client = $this->httpClient->newClient();
 		try {
-			$result = $client->get($url, [
-				'timeout' => 10,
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):

juliushaertl:
Added some more cases that I noticed when trying to debug something with @Antreesy where the request failed silently to also make use of the setting and log errors in apps/files_sharing/lib/External/Storage.php

provokateurin (approved):

PR #47346, opened 2024-08-20
store the mountpoint of storages in the mount options and use it to find the encryption keys

icewind1991:
No matter how much I find storages knowing their own mountpoint wrong from ideological level. In practice this info is needed in various places so might as well create a more universal solution for it.


sorbaugh (approved):

skjnldsv (approved):

danxuliu (approved):
Tested (see https://github.com/nextcloud/server/pull/47044#issuecomment-2296308836) and works :+1:

artonge (commented):
What advantage does it bring compared to asking the mount manager? Beside removing one line of code.

come-nc (approved):

PR #47343, opened 2024-08-20
Use logical positioning instead of physical for bidirectional text support

AndyScherzinger:
* Successor of #42471
* For: #31420 
  * Backend part moved to #47349
---

## Summary
This PR is the first step towards adding bidirectional text support to the Nextcloud ecosystem. It affects the entire project; accordingly, each app should have subsequent PRs.

The PR is 90+ percent complete. I decided to make the PR to
- Get early feedback to ensure it is on the right path.
- Discuss some technical challenges.

I had challenges in building the project. Generating `css` files and the `dist` directory during development is something I am uncomfortable with.

Because of supporting Samsung Internet, I needed to take some hacky approach.

**Note:**
We must update many components in [NextCloud/vue](https://github.com/nextcloud-libraries/nextcloud-vue/). The most important one was the `header-menu` vue component which is [already updated](https://github.com/nextcloud-libraries/nextcloud-vue/pull/4965). Others will come soon. First, I need to ensure this PR is getting prepared for merge.

## Screenshots
| Before | After |
|---|---|
|image|image|
|image|image|

## TODO

- [x] Decide about some classes like `.right` or `.menu-left`.
- [ ] Decide about `core/css/tooltip.scss` and its right|left classes
- [ ] Decide about https://github.com/nextcloud/server/blob/6b71288a30d6acbaaeb2f933a44883f3f51d7260/core/src/views/Profile.vue inline styles (for background-position.
- [ ] Add test to ensure we have proper `dir` attribute on the body. (help needed)

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #47349, opened 2024-08-20
Add bidirectional text support - Backend

nickvergessen:
- Backend part split out from #47343

## Checklist

### DIFF
--- a/build/translation-checker.php
+++ b/build/translation-checker.php
@@ -16,6 +16,35 @@ $untranslatedApps = [
 	'testing',
 ];
 
+// Next line only looks messed up, but it works. Don't touch it!
+$rtlCharacters = [
+	'\x{061C}', // ARABIC LETTER MARK
+	'\x{0623}', // ARABIC LETTER ALEF WITH HAMZA ABOVE
+	'\x{200E}', // LEFT-TO-RIGHT MARK
+	'\x{200F}', // RIGHT-TO-LEFT MARK
+	'\x{202A}', // LEFT-TO-RIGHT EMBEDDING
+	'\x{202B}', // RIGHT-TO-LEFT EMBEDDING
+	'\x{202C}', // POP DIRECTIONAL FORMATTING
+	'\x{202D}', // LEFT-TO-RIGHT OVERRIDE
+	'\x{202E}', // RIGHT-TO-LEFT OVERRIDE
+	'\x{2066}', // LEFT-TO-RIGHT ISOLATE
+	'\x{2067}', // RIGHT-TO-LEFT ISOLATE
+	'\x{2068}', // FIRST STRONG ISOLATE
+	'\x{2069}', // POP DIRECTIONAL ISOLATE
+	'\x{206C}', // INHIBIT ARABIC FORM SHAPING
+	'\x{206D}', // ACTIVATE ARABIC FORM SHAPING
+];
+
+$rtlLanguages = [
+	'ar', // Arabic
+	'fa', // Persian
+	'he', // Hebrew
+	'ps', // Pashto,
+	'ug', // 'Uyghurche / Uyghur
+	'ur_PK', // Urdu
+	'uz', // Uzbek Afghan
+];
+
 $valid = 0;
 $errors = [];
 $apps = new \DirectoryIterator(__DIR__ . '/../apps');
@@ -46,6 +75,12 @@ foreach ($directories as $dir) {
 		}
 
 		$content = file_get_contents($file->getPathname());
+
+		$language = pathinfo($file->getFilename(), PATHINFO_FILENAME);
+		if (!in_array($language, $rtlLanguages, true) && preg_match('/[' . implode('', $rtlCharacters) . ']/u', $content)) {
+			$errors[] = $file->getPathname() . "\n" . '  ' . 'Contains a RTL limited character in the translations.' . "\n";
+		}
+
 		$json = json_decode($content, true);
 
 		$translations = json_encode($json['translations']);
--- a/lib/private/L10N/Factory.php
+++ b/lib/private/L10N/Factory.php
@@ -58,6 +58,12 @@ class Factory implements IFactory {
 		'pt_BR', 'pt_PT', 'da', 'fi_FI', 'nb_NO', 'sv', 'tr', 'zh_CN', 'ko'
 	];
 
+	public const RTL_LANGUAGES = [
+		'ae', 'ar', 'arc', 'arz', 'bcc', 'bqi', 'ckb', 'dv', 'fa', 'glk',
+		'ha', 'he', 'khw', 'ks', 'ku', 'mzn', 'nqo', 'pnb', 'ps', 'sd', 'ug',
+		'ur', 'uzs', 'yi',
+	];
+
 	private ICache $cache;
 
 	public function __construct(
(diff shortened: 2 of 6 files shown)


lib/public/L10N/IFactory.php
```diff
@@ -103,6 +103,14 @@ public function languageExists($app, $lang);
 	 */
 	public function localeExists($locale);
 
+	/**
+	 * Return the current language direction from locale
+	 *
+	 * @param string $locale
+	 * @return string
```
susnux, line 110:
```suggestion
	 * @return 'ltr'|'rtl'
	 * @since 31.0.0
```

lib/private/L10N/Factory.php
```diff
@@ -58,6 +58,12 @@ class Factory implements IFactory {
 		'pt_BR', 'pt_PT', 'da', 'fi_FI', 'nb_NO', 'sv', 'tr', 'zh_CN', 'ko'
 	];
 
+	const RTL_LANGUAGES = [
```
susnux, line 61: Maybe make the visibility explicit?

```suggestion
	public const RTL_LANGUAGES = [
```

icewind1991 (dismissed):

provokateurin (dismissed):

lib/private/L10N/Factory.php
```diff
@@ -58,6 +58,10 @@ class Factory implements IFactory {
 		'pt_BR', 'pt_PT', 'da', 'fi_FI', 'nb_NO', 'sv', 'tr', 'zh_CN', 'ko'
 	];
 
+	public const RTL_LANGUAGES = [
+		'ar', 'dv', 'fa', 'ha', 'he', 'ks', 'ku', 'ps', 'sd', 'ug', 'ur_PK', 'uz', 'yi',
```
susnux, line 62: Not sure why you removed this @nickvergessen ?
https://www.localeplanet.com/icu/ur-IN/index.html
https://www.localeplanet.com/icu/ur/index.html
```suggestion
		'ar', 'dv', 'fa', 'ha', 'he', 'ks', 'ku', 'ps', 'sd', 'ug', 'ur', 'ur_IN', 'ur_PK', 'uz', 'yi',
```

lib/private/L10N/Factory.php
```diff
@@ -59,9 +59,7 @@ class Factory implements IFactory {
 	];
 
 	public const RTL_LANGUAGES = [
-		'ae', 'ar', 'arc', 'arz', 'bcc', 'bqi', 'ckb', 'dv', 'fa', 'glk',
-		'ha', 'he', 'khw', 'ks', 'ku', 'mzn', 'nqo', 'pnb', 'ps', 'sd', 'ug',
-		'ur', 'uzs', 'yi',
+		'ar', 'dv', 'fa', 'ha', 'he', 'ks', 'ku', 'ps', 'sd', 'ug', 'ur_PK', 'uz', 'yi',
```
susnux, line 62:
```suggestion
		'ar', 'ckb', 'dv', 'fa', 'ha', 'he', 'ks', 'ku', 'ps', 'sd', 'ug', 'ur_PK', 'uz', 'yi',
```

Same for ckb: https://www.localeplanet.com/icu/ckb/index.html

lib/private/L10N/Factory.php
```diff
@@ -59,9 +59,7 @@ class Factory implements IFactory {
 	];
 
 	public const RTL_LANGUAGES = [
-		'ae', 'ar', 'arc', 'arz', 'bcc', 'bqi', 'ckb', 'dv', 'fa', 'glk',
-		'ha', 'he', 'khw', 'ks', 'ku', 'mzn', 'nqo', 'pnb', 'ps', 'sd', 'ug',
```
susnux: `mzn` is also RTL: https://www.localeplanet.com/icu/mzn/index.html

skjnldsv (dismissed):

lib/private/L10N/Factory.php
nickvergessen: Because we use 2 character [ISO 639-1](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes) for translations. I removed everything that is not available at all on transifex.

I see now that despite the `RTL_LANGUAGE` name of the const it seems to be taking the orientation based on the locale, and I'm not sure about that either.

nickvergessen: I will try to get input from couple of RTL speakers to see if that is intended and also check what other projects do.
Then we either rename the constant to locale and revert the list, or we recheck my list.

nickvergessen: JS lib seems to do something wrong as well:
https://github.com/nextcloud-libraries/nextcloud-l10n/blob/e0b16b54fc946273e31e99e0b16d5e85b4ff9560/lib/locale.ts#L28

Takes the language (which is always 2 letters or `ab_YZ` and compares it to 3 letter locales

nickvergessen: But let's wait with adjusting until we know

susnux: @nickvergessen the direction (RTL vs LTR) is get by the locale not the language.
So we need to consider all valid locales and we also support 3 digit locales. Our local validation considers all locales from `resources/locales.json` as valid -> all the locales I mentioned are included there.

nickvergessen: > the direction (RTL vs LTR) is get by the locale not the language.

If this is the study fact (and not just describing what is done in this PR here), we shall rename the constant, revert the list and fix the JS libary as it is using the language and not the locale to detect RTL.

susnux: This makes sense (but also its sad we use 639-1 instead of 639-2)

skjnldsv (approved):

lib/private/L10N/Factory.php
susnux: I had a look at our transifex page, there seem to be some `639-2` languages, not that popular but with ~1% and growing translations. Do we just disable merging it into server?

skjnldsv:
Frontend has been merged :)

nickvergessen:
Yeah but this one needs more work and deeper analysis of the situation. I will clarify it during the conf.

Frontend devs that want to continue working on issues can simply hardcode the dir in the template files for the time being

lib/private/L10N/Factory.php
nickvergessen: Current feedback is that both would expect it based on language as that defines the direction of most text.

AndyScherzinger (approved):

nickvergessen:
/backport 242f89afb3312f04bbbf7951d9a71092f15978a1 to stable30

nickvergessen:
/backport 242f89afb3312f04bbbf7951d9a71092f15978a1 to stable29

nickvergessen:
/backport 242f89afb3312f04bbbf7951d9a71092f15978a1 to stable28

build/translation-checker.php
```diff
@@ -16,6 +16,35 @@
...
+];
+
+$rtlLanguages = [
+	'ar', // Arabic
+	'fa', // Persian
+	'he', // Hebrew
+	'ps', // Pashto,
+	'ug', // 'Uyghurche / Uyghur
+	'ur_PK', // Urdu
+	'uz', // Uzbek Afghan
```
susnux, line 45: That is wrong Uzbek is writtin in latin script (LTR) only afghan Uzbek is RTL (`uz_AF`)

AndyScherzinger: PR welcome 😉

AndyScherzinger:
@ahangarha so the changes you made with some work on top by further engineers has been merged, so main branch now ships all the changes so far.

From here on the long list of small fixes will start I think and any PR can just target the master branch. Any discussion can happen on the Talk room as aligned some weeks ago.

Hope you are good and I think we are all happy with the current progress already, so let's take it from here and improve all the bits and pieces on bi-directional text support 🚀

ahangarha:
Great Job 👍🏾

build/translation-checker.php
nickvergessen: https://github.com/nextcloud/server/pull/48419

PR #47351, opened 2024-08-20
fix(build): Allow update-apps.sh script to work with non-master branches

provokateurin:
## Summary

The script only worked for apps which use master as their main branch.
This change also makes the script easier to read and also catch errors properly.

## Checklist

### DIFF
--- a/build/update-apps.sh
+++ b/build/update-apps.sh
@@ -10,9 +10,23 @@
 # It automatically:
 # - goes through all apps which are not shipped via server
 # - shows the app name in bold and uses whitespace for separation
-# - changes to master and pulls quietly
+# - changes to master/main and pulls quietly
 # - shows the 3 most recent commits for context
-# - removes branches merged into master
+# - removes branches merged into master/main
 # - … could even do the build steps if they are consistent for the apps (like `make`)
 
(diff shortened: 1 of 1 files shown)


sorbaugh (approved):

susnux (approved):

skjnldsv (approved):

come-nc:
Fails with sh:
```
$ sh build/update-apps.sh
build/update-apps.sh: 18: set: Illegal option -o pipefail
```
Fails with french locale:
```
$ bash build/update-apps.sh

activity
fatal: empty string is not a valid pathspec. please use . instead if you meant to match all paths
```
(because `git remote show origin` does not contain `HEAD branch` but a french translation)

provokateurin:
For the later we could do a stupid `git checkout master || git checkout main`. If both fail then the whole script will fail as expected.

susnux:
> Ugh I didn't know and think about that. Not sure how to make that work. I tried a few alternatives, but none of them worked for me at all.

I think in that script just `export LANG=C` in the beginning should be enough (and should not conflict as long as you do not source it)

provokateurin:
@come-nc can you try that and if it works make a PR?

provokateurin:
> and should not conflict as long as you do not source it

Instead of using export the env variable could just be applied to the git command that is problematic.

susnux:
> Instead of using export the env variable could just be applied to the git command that is problematic.

Sure this also works :)
I just export to have a consistent scripting env. The env is not leaking into your personal shell session (as long as you do not source it, but why should you? :sweat_smile: )

provokateurin:
Yeah it would be weird to do that, but you mentioned it so we can also just prevent it :D

PR #47363, opened 2024-08-20
fix: use empty template by default when pressing enter

### LINKED ISSUE #47320: new file > choose template: enter key closes dialogue instead of opening new document
**To Reproduce**
1. in files, create new text file
2. dialogue for name: press enter
3. dialogue for template: press enter

**Expected behavior**
new document with pre-selected template is  opened

**Screenshots**
dialogue is closed, no file created or opened

**Server details:**
- Nextcloud version: 30.0.0. RC1
- can't check app version, it's on c.nc.com

elzody:
* Resolves: #47320

## Summary
Previously, when pressing enter without clicking on anything in the template picker, it would simply close and nothing would happen -- no document created, etc. The expected behavior is that pressing enter would just create a document based on the default, empty template. This PR makes it so that the empty template preview is focused when the picker is opened, thus pressing enter when the picker opens will now create a file based on the empty template.

https://github.com/user-attachments/assets/f7add0a0-1a7d-4ab3-8627-9fc59e2a61d1

## Checklist


apps/files/src/views/TemplatePicker.vue
```diff
@@ -179,6 +180,11 @@ export default defineComponent({
 
 			// Else, open the picker
 			this.opened = true
+
+			// Set initial focus to the empty template preview
+			this.$nextTick(() => {
+				this.$refs.emptyTemplatePreview?.$el.querySelector('input')?.focus()
```
susnux, line 186: I think this is quite hacky as you assume there are specific DOM elements within an other component.
Instead it would be cleaner if you expose a method in `TemplatePreview` like `focus()` and just call:

```suggestion
				this.$refs.emptyTemplatePreview?.focus()
```

elzody: Indeed makes sense, thanks for the helpful feedback

apps/files/src/components/TemplatePreview.vue
```diff
@@ -124,6 +124,9 @@ export default {
 		onFailure() {
 			this.failedPreview = true
 		},
+		focus() {
+			this.$el.querySelector('input').focus()
```
susnux, line 128: You can also add a `ref="input"` to the input element in the template and then simply

```suggestion
			this.$refs.input?.focus()
```

susnux (approved):
Thank you :)

juliusknorr (approved):

PR #47371, opened 2024-08-21
fix: Disable auto-zoom on iOS

### LINKED ISSUE #47328: Auto zoom when using Nextcloud on Safari iOS
I used our web interface a couple of times now on mobile safari and if feels quite strange and is partially also annoying that Safari is automatically zooming to input elements and also contenteditables (e.g. in text) when you focus them.

This used to work differently and in my perspective also should for a web app that is not a pure content page.

This was introduced in https://github.com/nextcloud/server/commit/ac736b2f82adfe73176fa78620e617ff17dbc91e as a accessibility fix where I still agree that manually zooming should be possible, however maybe we can get rid of the automatic zoomin.

I found https://stackoverflow.com/questions/2989263/disable-auto-zoom-in-input-text-tag-safari-on-iphone as a possible option.

@nextcloud/designers What do you think about that? Any arguments against changing this?

susnux:
* Resolves https://github.com/nextcloud/server/issues/47328

## Summary

When using iOS and focusing an input element the view should not be zoomed. So if we set a maximum scale iOS will not auto-zoom but still allow users to zoom. But we can not do this by default as this will disable user zoom on Chrome.

## Checklist

### DIFF
--- a/lib/private/TemplateLayout.php
+++ b/lib/private/TemplateLayout.php
@@ -289,7 +289,7 @@ class TemplateLayout extends \OC_Template {
 		}
 
 		$request = \OCP\Server::get(IRequest::class);
-		if ($request->isUserAgent([Request::USER_AGENT_CLIENT_IOS, Request::USER_AGENT_SAFARI])) {
+		if ($request->isUserAgent([Request::USER_AGENT_CLIENT_IOS, Request::USER_AGENT_SAFARI, Request::USER_AGENT_SAFARI_MOBILE])) {
 			// Prevent auto zoom with iOS but still allow user zoom
 			// On chrome (and others) this does not work (will also disable user zoom)
 			$this->assign('viewport_maximum_scale', '1.0');
--- a/lib/private/TemplateLayout.php
+++ b/lib/private/TemplateLayout.php
@@ -8,6 +8,7 @@
 namespace OC;
 
 use bantu\IniGetWrapper\IniGetWrapper;
+use OC\AppFramework\Http\Request;
 use OC\Authentication\Token\IProvider;
 use OC\Files\FilenameValidator;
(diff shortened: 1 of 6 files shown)


skjnldsv (approved):

jancborchardt (approved):

lib/private/TemplateLayout.php
```diff
@@ -286,6 +288,13 @@ public function __construct($renderAs, $appId = '') {
 			}
 		}
 
+		$request = \OCP\Server::get(IRequest::class);
+		if ($request->isUserAgent([Request::USER_AGENT_CLIENT_IOS, Request::USER_AGENT_SAFARI])) {
```
juliushaertl, line 292: Just tested and those two do not match mobile Safari, the following should work adding the following match group `(Mobile\/[0-9.A-Z]+)`, not sure if we want to:

- Include it in USER_AGENT_SAFARI: `Mozilla\/5\.0 \([^)]+\) AppleWebKit\/[0-9.]+ \(KHTML, like Gecko\) Version\/[0-9.]+ (Mobile\/[0-9.A-Z]+)? Safari\/[0-9.A-Z]+`
- Have a separate USER_AGENT_SAFARI_MOBILE: `Mozilla\/5\.0 \([^)]+\) AppleWebKit\/[0-9.]+ \(KHTML, like Gecko\) Version\/[0-9.]+ (Mobile\/[0-9.A-Z]+) Safari\/[0-9.A-Z]+`

susnux: Thanks for testing! I chose the second as we also have the mobile chrome variant

juliusknorr (approved):
Awesome, tested and works now 👍 

Just Ci is not happy yet

PR #47403, opened 2024-08-21
feat(Security): Allow defining a password context for password validation and generation

susnux:
## Summary

This would allow us to implement https://github.com/nextcloud/password_policy/issues/117.
Basically add a context to the events to define for which use case the password needs to be validated / generated.

## Checklist


lib/public/Security/Events/ValidatePasswordPolicyEvent.php
```diff
@@ -9,26 +9,40 @@
...
+ * is invalid, an `\OCP\HintException` will be thrown.
  * @since 18.0.0
  */
 class ValidatePasswordPolicyEvent extends Event {
-	/** @var string */
-	private $password;
 
 	/**
-	 * @since 18.0.0
+	 * @since 18.0.0 - $context parameter since 31.0.0
```
nickvergessen, line 24:
```suggestion
	 * @since 18.0.0
	 * @since 31.0.0 - $context parameter added
```

lib/public/Security/PasswordContext.php
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * @package OCP\Security
+ * @since 31.0.0
+ */
+enum PasswordContext: int {
+	/**
+	 * Password used for an user account
+	 * @since 31.0.0
+	 */
+	case ACCOUNT = 1;
```
nickvergessen, line 22: Could also use strings so that when you see the value anywhere you can understand it without checking code?

nickvergessen (commented):
Looks cool otherwise

lib/public/Security/Events/ValidatePasswordPolicyEvent.php
susnux: I saw this in the other places and wanted to be consistent, but that looks much cleaner :)

lib/public/Security/PasswordContext.php
susnux: sure, I see no drawback doing so

nickvergessen (approved):

provokateurin (approved):

lib/public/Security/PasswordContext.php
```diff
@@ -0,0 +1,29 @@
...
+namespace OCP\Security;
+
+/**
+ * Define the context in which a password is used.
+ * This allows setting a context for password validation and password generation.
+ *
+ * @package OCP\Security
+ * @since 31.0.0
+ */
+enum PasswordContext: string {
```
come-nc, line 17: Out of curiosity, why back it up with string?
A basic enum would be enough for the use-case I think.

susnux: I think it easier to work with especially if you need to serialize. In the first place I used `int` but string was recommended.

We could also make it a plain enum, this would allow to make it a backed enum later, but the other way around would be a breaking change.

PR #47412, opened 2024-08-22
feat(files_sharing): reminder for link shares with expiration date

### LINKED ISSUE #45676: Reminder for link shares with expiration date
### How to use GitHub

Currently when you share a file drop or generally a share link with an expiration date, people might forget to upload something right away. It would be nice to remind them to upload.

Nextcloud should send a reminder via email to the sharee(s) if 1 day before the expiration date the folder is still empty.
We need to check:
- Do you have write permissions
- Is there an expiration date
- Is the folder empty

The "1 day" value is flexible. It might be a bit short notice but it’s good since it only reminds if it is already urgent. But there should be no setting for this anywhere, not for admins nor sharers.

As discussed @AndyScherzinger @artonge @jospoortvliet 

--

Since we don’t save the name of the uploader, currently we can only do this generally. If at some point we do store uploaders, we could refine it to send you a reminder if you specifically did not upload anything yet.

Could also be extended to link shares _without_ expiration date, similar to the nudges in Mail, to follow up after 4 days. But this might be annoying since stuff gets shared all the time.

Luka-sama:
* Resolves: #45676

## Summary

This pull request implements the feature as described in the associated issue. A cron job runs hourly to check if any empty folders have less than a day remaining before their expiration date. The feature supports two types of shares: user shares and email shares.

Additionally, a migration was added to include a 'reminder sent' flag, ensuring that reminders are neither skipped nor sent multiple times when the job runs at irregular intervals, especially when using webcron or ajax instead of cron.

This solution can be easily extended to send reminders for link shares without an expiration date (e.g. after 4 days), if the maintainers choose to implement this feature.

grafik

## Checklist

### DIFF
--- a/apps/files_sharing/appinfo/info.xml
+++ b/apps/files_sharing/appinfo/info.xml
@@ -14,7 +14,7 @@
 Turning the feature off removes shared files and folders on the server for all share recipients, and also on the sync clients and mobile apps. More information is available in the Nextcloud Documentation.
 
 	</description>
-	<version>1.23.0</version>
+	<version>1.23.1</version>
 	<licence>agpl</licence>
 	<author>Michael Gapczynski</author>
 	<author>Bjoern Schiessle</author>
@@ -33,6 +33,7 @@ Turning the feature off removes shared files and folders on the server for all s
 	<background-jobs>
 		<job>OCA\Files_Sharing\DeleteOrphanedSharesJob</job>
 		<job>OCA\Files_Sharing\ExpireSharesJob</job>
+		<job>OCA\Files_Sharing\SharesReminderJob</job>
 		<job>OCA\Files_Sharing\BackgroundJob\FederatedSharesDiscoverJob</job>
 	</background-jobs>
 
--- a/apps/files_sharing/lib/Migration/Version31000Date20240821142813.php
+++ b/apps/files_sharing/lib/Migration/Version31000Date20240821142813.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OCA\Files_Sharing\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version31000Date20240821142813 extends SimpleMigrationStep {
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		$schema = $schemaClosure();
+		$table = $schema->getTable('share');
+		$table->addColumn('reminder_sent', Types::BOOLEAN, [
+			'notnull' => false,
+			'default' => false,
+		]);
+		return $schema;
+	}
+
+}
--- a/apps/files_sharing/lib/SharesReminderJob.php
+++ b/apps/files_sharing/lib/SharesReminderJob.php
@@ -132,6 +132,9 @@ class SharesReminderJob extends TimedJob {
 		$reminderInfo = [];
 		if ($share->getShareType() == IShare::TYPE_USER) {
 			$user = $this->userManager->get($sharedWith);
+			if ($user === null) {
+				return null;
+			}
 			$reminderInfo['email'] = $user->getEMailAddress();
 			$reminderInfo['userLang'] = $this->l10nFactory->getUserLanguage($user);
 			$reminderInfo['folderLink'] = $this->urlGenerator->linkToRouteAbsolute('files.view.index', [
--- a/apps/files_sharing/lib/SharesReminderJob.php
+++ b/apps/files_sharing/lib/SharesReminderJob.php
@@ -32,6 +32,7 @@ use Psr\Log\LoggerInterface;
  */
 class SharesReminderJob extends TimedJob {
 	private const SECONDS_BEFORE_REMINDER = 86400;
+	private const CHUNK_SIZE = 1000;
 
 	public function __construct(
 		ITimeFactory                     $time,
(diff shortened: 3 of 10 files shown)


apps/files_sharing/lib/SharesReminderJob.php
```diff
@@ -0,0 +1,261 @@
...
+						$qb->expr()->eq('share_type', $qb->expr()->literal(IShare::TYPE_EMAIL))
+					),
+					$qb->expr()->eq('item_type', $qb->expr()->literal('folder')),
+					$qb->expr()->gte('expiration', $qb->createNamedParameter($minDate->format('Y-m-d H:i:s'))),
+					$qb->expr()->lt('expiration', $qb->createNamedParameter($maxDate->format('Y-m-d H:i:s'))),
+					$qb->expr()->eq('reminder_sent', $qb->createNamedParameter(
+						false, IQueryBuilder::PARAM_BOOL
+					))
+				)
+			);
```
come-nc, line 92: Should this go into a mapper or manager of some kind? I know in other apps we usually do not hit the DB directly like that, not sure for shares.

apps/files_sharing/appinfo/info.xml
```diff
@@ -14,7 +14,7 @@
 Turning the feature off removes shared files and folders on the server for all share recipients, and also on the sync clients and mobile apps. More information is available in the Nextcloud Documentation.
 
 	</description>
-	<version>1.23.0</version>
+	<version>1.23.1</version>
```
come-nc, line 17: That’s needed for the migration to run I suppose?

apps/files_sharing/lib/SharesReminderJob.php
```diff
@@ -0,0 +1,261 @@
...
+				$foldersByEmail[$mailTo] = [];
+			}
+			$foldersByEmail[$mailTo][] = ['name' => $share->getNode()->getName(), 'link' => $link];
+
+			$share->setReminderSent(true);
+			$qb = $this->db->getQueryBuilder();
+			$qb->update('share')
+				->where($qb->expr()->eq('id', $qb->createNamedParameter($share->getId())))
+				->set('reminder_sent', $qb->createNamedParameter($share->getReminderSent()))
+				->execute();
```
come-nc, line 168:
```suggestion
				->executeStatement();
```
But again, I’d like to see that moved somewhere else.
The updateShare method seems a bit much for such a small flag change though, so I don’t know :thinking:

lib/private/Share20/Share.php
```diff
@@ -72,6 +72,8 @@ class Share implements IShare {
 	private $nodeCacheEntry;
 	/** @var bool */
 	private $hideDownload = false;
+	/** @var bool */
+	private $reminderSent = false;
```
come-nc, line 76:
```suggestion
	private bool $reminderSent = false;
```

lib/private/Share20/Share.php
```diff
@@ -613,4 +615,13 @@ public function setHideDownload(bool $hide): IShare {
...
 	}
+
+	public function setReminderSent(bool $reminderSent): IShare {
+		$this->reminderSent = $reminderSent;
+		return $this;
+	}
+
+	public function getReminderSent(): bool {
+		return $this->reminderSent;
+	}
```
come-nc, line 626: But is the information fetch from DB and put into this object by the manager or the providers?

apps/files_sharing/lib/SharesReminderJob.php
```diff
@@ -0,0 +1,261 @@
...
+	 * Makes the background job do its work
+	 *
+	 * @param array $argument unused argument
+	 * @throws Exception if a database error occurs
+	 */
+	public function run(mixed $argument): void {
+		$shares = $this->getShares();
+		[$foldersByEmail, $langByEmail] = $this->prepareReminders($shares);
+		$this->sendReminders($foldersByEmail, $langByEmail);
+	}
```
come-nc, line 62: You need to put a limit here, and work only on a chunk of possible reminders, like `1000`. Otherwise you can run into memory issues and such.
Also, you are first gathering all the info and then sending all the email, consider using Generators to send shares as you find them.
Also, you need to consider what happens if your job crashes or is stopped mid-process. In the current state it sets the flag reminder_sent on all shares first, which means it won’t get sent for other shares if it crashes on a share before. This is better than re-sending reminders, but it would still be better to flag shares one by one just before sending the email.

skjnldsv:
Great idea @Luka-sama !!
Take your time to address Come's comments :+1:

apps/files_sharing/lib/SharesReminderJob.php
Luka-sama: I could move this to the share manager, but I'm not sure if I should, since the query is very specific to this particular job 🤔

apps/files_sharing/appinfo/info.xml
Luka-sama: Yes, with a new version number the user will be required to do "occ upgrade" (via web interface or console). Also, it seems logical to me to update the version when the new feature comes out.
But of course I could be wrong as I'm new to nextcloud and don't know all the details of how it works here.

lib/private/Share20/Share.php
Luka-sama: Should I maybe update other properties to this format as well (to make it consistent)?

Luka-sama:
Thank you @come-nc and @skjnldsv! The idea comes from the issue, but I found it a very interesting task to implement it here :)
That's a very nice idea with generators! I will fix/address this and other problems in the next days.

apps/files_sharing/lib/SharesReminderJob.php
Luka-sama: I think updateShare works well here, the overhead (if any) is quite insignificant.

lib/private/Share20/Share.php
Luka-sama: > Should I maybe update other properties to this format as well (to make it consistent)?

I decided against it because it breaks a lot of code (e.g. some int properties are sometimes set to null). So I did the strict typing only for my property.

lib/private/Share20/Share.php
Luka-sama: Now it does, thanks for noticing!

apps/files_sharing/lib/SharesReminderJob.php
Luka-sama: Fixed, thank you!

Luka-sama:
So, I've fixed the problems that @come-nc found!
To do this, I moved the condition checking (write permissions and folder emptiness) to the database query - to avoid the situation where the job checks the same 1000 shares every time.

apps/files_sharing/lib/SharesReminderJob.php
```diff
@@ -0,0 +1,213 @@
...
+	/**
+	 * Makes the background job do its work
+	 *
+	 * @param array $argument unused argument
+	 * @throws Exception if a database error occurs
+	 */
+	public function run(mixed $argument): void {
+		foreach ($this->getShares() as $share) {
+			$reminderInfo = $this->prepareReminder($share);
+			$this->sendReminder($reminderInfo);
```
come-nc, line 62:
```suggestion
			$reminderInfo = $this->prepareReminder($share);
			if ($reminderInfo !== null) {
				$this->sendReminder($reminderInfo);
			}
```

apps/sharebymail/lib/ShareByMailProvider.php
```diff
@@ -755,6 +755,7 @@ public function update(IShare $share, ?string $plainTextPassword = null): IShare
 			->set('hide_download', $qb->createNamedParameter((int)$share->getHideDownload(), IQueryBuilder::PARAM_INT))
 			->set('attributes', $qb->createNamedParameter($shareAttributes))
 			->set('mail_send', $qb->createNamedParameter((int)$share->getMailSend(), IQueryBuilder::PARAM_INT))
+			->set('reminder_sent', $qb->createNamedParameter((int)$share->getReminderSent(), IQueryBuilder::PARAM_INT))
```
come-nc, line 758: Why not PARAM_BOOL?

come-nc (commented):
`reminder_sent` should be a bool in DB as well, otherwise good.
Only thing a bit questionable is how would other providers add support for this, but this can be refactored in later.

apps/sharebymail/lib/ShareByMailProvider.php
Luka-sama: I guess I was following the example of other fields that mostly use int for boolean data. However, some fields use bool – and since the tests passed, I've changed to bool as well.

Luka-sama:
Thank you for your code review!
Yes, currently only two types of shares are supported (both in providers and in the job itself). However, support for other types can easily be added if needed.

come-nc (approved):
What I meant was that this could’ve been added in a more extensible way, for instance with an interface for providers which want to implement the feature, and which adds a method to get which email address the reminder should be sent to.
Just thinking out loud, I have no idea if this would’ve been a good idea and all it implies, just a suggestion for future work to think about extensibility by applications.

Approved :+1:

Luka-sama:
You're right, that's definitely a good idea for the future to make the feature more universal this way.

Thanks!

skjnldsv (approved):

skjnldsv:
Ok, let's go! :rocket:

Pytal:
This broke CI on master cc @skjnldsv

apps/files_sharing/lib/Migration/Version31000Date20240821142813.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace OCA\Files_Sharing\Migration;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\DB\Types;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+class Version31000Date20240821142813 extends SimpleMigrationStep {
```
susnux, line 18: As this target 31 it should have a migration annotation, otherwise we have no information of the impact.

Luka-sama: I have added it:
https://github.com/nextcloud/server/commit/d758442
However, my changes are no longer displayed here in this pull request, probably because the branch is already merged into master.

susnux: @Luka-sama yes because it is already merged. But is also not about a doc-comment but about annotation for the kind of the migration (`OCP\Migration\Attributes\MigrationAttribute`).
Nevertheless thank you very much!!

Luka-sama: @susnux Oh, I see! I must have used the older migrations as an example, the attributes seem to be something new (from the 30th version).
Should I add them too? Except that it would probably be an overkill to make a new pull request just for that 😅 (or not?)

PR #47415, opened 2024-08-22
fix: Trusted server icon recovery without addressbook change

pabzm:
## Summary

Always set server status to OK after successful runs

Previously if a server status got set to failure, it stayed that way
until an addressbook-sync found changes. Now the server status is set to
OK after each successful sync check (if that's not the case already),
regardless of addressbook changes.

Additionally a log warning was added in two places where the server status gets set to failure when a job gets removed due to being too old. This would've helped in debugging a problem, so now there's a logging hint for next time.

## Checklist

### DIFF
--- a/apps/federation/lib/BackgroundJob/GetSharedSecret.php
+++ b/apps/federation/lib/BackgroundJob/GetSharedSecret.php
@@ -90,6 +90,7 @@ class GetSharedSecret extends Job {
 		// kill job after 30 days of trying
 		$deadline = $currentTime - $this->maxLifespan;
 		if ($created < $deadline) {
+			$this->logger->warning("The job to get the shared secret job is too old and gets stopped now without retention. Setting server status of '{$target}' to failure.");
 			$this->retainJob = false;
 			$this->trustedServers->setServerStatus($target, TrustedServers::STATUS_FAILURE);
 			return;
--- a/apps/federation/lib/BackgroundJob/RequestSharedSecret.php
+++ b/apps/federation/lib/BackgroundJob/RequestSharedSecret.php
@@ -93,6 +93,7 @@ class RequestSharedSecret extends Job {
 		// kill job after 30 days of trying
 		$deadline = $currentTime - $this->maxLifespan;
 		if ($created < $deadline) {
+			$this->logger->warning("The job to request the shared secret job is too old and gets stopped now without retention. Setting server status of '{$target}' to failure.");
 			$this->retainJob = false;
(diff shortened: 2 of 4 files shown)


pabzm:
This is my first PR in this repo, I'm a little unsure about some details (e.g. those checkboxes). Please let me know if there's something I should change or do.

apps/federation/tests/SyncFederationAddressbooksTest.php
```diff
@@ -90,4 +90,35 @@ public function testException() {
...
+		/** @var DbHandler | MockObject $dbHandler */
+		$dbHandler = $this->getMockBuilder('OCA\Federation\DbHandler')
+			->disableOriginalConstructor()
+			->getMock();
+		$dbHandler->method('getAllServer')
+			->willReturn([
+				[
+					'url' => 'https://cloud.drop.box',
+					'url_hash' => 'sha1',
+					'shared_secret' => 'iloveowncloud',
```
ChristophWurst, line 104: oh no

apps/federation/lib/BackgroundJob/GetSharedSecret.php
```diff
@@ -90,6 +90,7 @@ protected function run($argument) {
 		// kill job after 30 days of trying
 		$deadline = $currentTime - $this->maxLifespan;
 		if ($created < $deadline) {
+			$this->logger->warning("This job is too old and gets stopped now without retention. Setting server status of '{$target}' to failure.");
```
ChristophWurst, line 93: nit: *this job* makes it hard to debug. Log something like *The job to get the shared secret*

apps/federation/lib/BackgroundJob/RequestSharedSecret.php
```diff
@@ -93,6 +93,7 @@ protected function run($argument) {
 		// kill job after 30 days of trying
 		$deadline = $currentTime - $this->maxLifespan;
 		if ($created < $deadline) {
+			$this->logger->warning("This job is too old and gets stopped now without retention. Setting server status of '{$target}' to failure.");
```
ChristophWurst, line 96: Same here

ChristophWurst (approved):

miaulalala (approved):

apps/federation/tests/SyncFederationAddressbooksTest.php
pabzm: Fixed❗

PR #47425, opened 2024-08-22
fix: promote re-shares when deleting the parent share

come-nc:
* Resolves: # 

## Summary

Same as #43025 but without the new command.

* ~Catch share exception in transfer ownership and delete invalid shares~
* Promote reshares into direct shares when a user loses access to a file after an unshare

## Checklist

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -18,6 +18,7 @@ use OCP\Files\Folder;
 use OCP\Files\IRootFolder;
 use OCP\Files\Mount\IMountManager;
 use OCP\Files\Node;
+use OCP\Files\NotFoundException;
 use OCP\HintException;
 use OCP\IConfig;
 use OCP\IDateTimeZone;
@@ -1039,6 +1040,71 @@ class Manager implements IManager {
 		return $deletedShares;
 	}
 
+	public function deleteReshare(IShare $share) {
+		// Skip if node not found
+		try {
+			$node = $share->getNode();
+		} catch (NotFoundException) {
+			return;
+		}
+
+		$userIds = [];
+
+		if  ($share->getShareType() === IShare::TYPE_USER) {
+			$userIds[] = $share->getSharedWith();
+		}
+
+		if ($share->getShareType() === IShare::TYPE_GROUP) {
+			$group = $this->groupManager->get($share->getSharedWith());
+			$users = $group->getUsers();
+
+			foreach ($users as $user) {
+				// Skip if share owner is member of shared group
+				if ($user->getUID() === $share->getShareOwner()) {
+					continue;
+				}
+				$userIds[] = $user->getUID();
+			}
+		}
+
+		$reshareRecords = [];
+		$shareTypes = [
+			IShare::TYPE_GROUP,
+			IShare::TYPE_USER,
+			IShare::TYPE_LINK,
+			IShare::TYPE_REMOTE,
+			IShare::TYPE_EMAIL
+		];
+
+		foreach ($userIds as $userId) {
+			foreach ($shareTypes as $shareType) {
+				$provider = $this->factory->getProviderForType($shareType);
+				$shares = $provider->getSharesBy($userId, $shareType, $node, false, -1, 0);
+				foreach ($shares as $child) {
+					$reshareRecords[] = $child;
+				}
+			}
+
+			if ($share->getNodeType() === 'folder') {
+				$sharesInFolder = $this->getSharesInFolder($userId, $node, true);
+
+				foreach ($sharesInFolder as $shares) {
+					foreach ($shares as $child) {
+						$reshareRecords[] = $child;
+					}
+				}
+			}
+		}
+
+		foreach ($reshareRecords as $child) {
+			try {
+				$this->generalCreateChecks($child);
+			} catch (GenericShareException $e) {
+				$this->deleteShare($child);
+			}
+		}
+	}
+
 	/**
 	 * Delete a share
 	 *
@@ -1061,6 +1127,9 @@ class Manager implements IManager {
 		// Do the actual delete
 		$provider = $this->factory->getProviderForType($share->getShareType());
 		$provider->delete($share);
+
+		// Delete shares that shared by the "share with user/group"
+		$this->deleteReshare($share);
 
 		$this->dispatcher->dispatchTyped(new ShareDeletedEvent($share));
 	}
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -783,6 +783,7 @@ class Manager implements IManager {
 	 * @param IShare $share
 	 * @return IShare The share object
 	 * @throws \InvalidArgumentException
+	 * @throws GenericShareException
 	 */
 	public function updateShare(IShare $share) {
 		$expirationDateUpdated = false;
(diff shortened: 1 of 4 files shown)


lib/private/Share20/Manager.php
```diff
@@ -1017,6 +1018,71 @@ protected function deleteChildren(IShare $share) {
...
+		// Skip if node not found
+		try {
+			$node = $share->getNode();
+		} catch (NotFoundException) {
+			return;
+		}
+
+		$userIds = [];
+
+		if  ($share->getShareType() === IShare::TYPE_USER) {
```
solracsf, line 1031:
```suggestion
		if ($share->getShareType() === IShare::TYPE_USER) {
```

luka-nextcloud (dismissed):

sorbaugh (dismissed):

come-nc:
The failing tests are related.

The test failing is testing this scenario:

> 	Scenario: transferring ownership of folder reshared with another user
> 		Given user "user0" exists
> 		And user "user1" exists
> 		And user "user2" exists
> 		And user "user3" exists
> 		And User "user3" created a folder "/test"
> 		And User "user3" uploads file "data/textfile.txt" to "/test/somefile.txt"
> 		And folder "/test" of user "user3" is shared with user "user0" with permissions 31
> 		And user "user0" accepts last share
> 		And folder "/test" of user "user0" is shared with user "user2" with permissions 31
> 		And user "user2" accepts last share
> 		When transferring ownership from "user0" to "user1"
> 		And the command was successful
> 		And As an "user2"
> 		**Then Downloaded content when downloading file "/test/somefile.txt" with range "bytes=0-6" should be "This is"**
> 		And using old dav path
> 		And as "user0" the folder "/test" exists
> 		And using received transfer folder of "user1" as dav path
> 		And as "user1" the folder "/test" does not exist
> 		And As an "user0"
> 		And Getting info of last share
> 		And the OCS status code should be "100"
> 		And Share fields of last share match with
> 			| uid_owner | user0 |
> 			| uid_file_owner | user3 |
> 			| share_with | user2 |

It fails at the step in bold.
The problem here is that we ask the command to transfer a reshare from user0 to user1 without transferring the incoming share. That would result in a situation where user1 shares the folder without having access to it. The test actually expects it to fail, it checks that user1 does not see the folder.
But with the new code, the share is seen as invalid and is deleted. While the old code would let it in place.

If I reorder the transferring code and use ` --transfer-incoming-shares=1` then I can successfully transfer the share.

The more I think about it the more it feels like it’s not the job of transfer-ownership to delete broken shares.
The fix in the share manager is enough to avoid share breakage on share deletion.
What would remain to fix is to behave better when encountering an already broken share.
From original bug report I see 2 occurences listed:
https://github.com/nextcloud/server/pull/43025#issuecomment-2189118746
https://github.com/nextcloud/server/pull/43025#issuecomment-1914759785

come-nc:
After internal discussion, reshares should **not** be deleted when share is deleted but turned into direct shares instead.
So something more along the line of what the repair share proposition from Luka did.

We should still be careful about what that means for transfer-ownership when incomming shares are not transfered.

lib/private/Share20/Manager.php
```diff
@@ -1010,6 +1012,79 @@ protected function deleteChildren(IShare $share) {
...
+			}
+		}
+
+		foreach ($reshareRecords as $child) {
+			try {
+				$this->generalCreateChecks($child);
+			} catch (GenericShareException $e) {
+				$this->logger->debug('Promote reshare because of exception '.$e->getMessage(), ['exception' => $e, 'fullId' => $child->getFullId()]);
+				try {
+					$child->setSharedBy($share->getSharedBy());
```
icewind1991, line 1079: I'm having problems understanding the flow here, why is the actual update done in the `catch` block?

lib/private/Share20/Manager.php
```diff
@@ -1010,6 +1012,79 @@ protected function deleteChildren(IShare $share) {
 		return $deletedShares;
 	}
 
+	/* Promote reshares into direct shares so that target user keeps access */
```
susnux, line 1015:
```suggestion
	/** Promote re-shares into direct shares so that target user keeps access */
```

susnux:
To make sure I understand this correct:
1. A shares F with B
2. B re-shares  F with C
3. A deletes share of F with B
4. B -> C share of F is converted to a share A -> C of F?

I think there was a discussion recently with @jancborchardt if this is expected behavior.
If I remember correctly if the original sharer now has control over the shares this should be fine, but C should not have access to F after B is removed without that A can control this.

lib/private/Share20/Manager.php
come-nc: Because if the exception does not fire it means the sharer still has access to the file through another way (another group share, share of a parent folder, groupfolder, you name it) and there is no need to update the reshare.

come-nc:
> To make sure I understand this correct:
> 
> 1. A shares F with B
> 2. B re-shares  F with C
> 3. A deletes share of F with B
> 4. B -> C share of F is converted to a share A -> C of F?

Yes that is correct

> I think there was a discussion recently with @jancborchardt if this is expected behavior. If I remember correctly if the original sharer now has control over the shares this should be fine, but C should not have access to F after B is removed without that A can control this.

Before deleting the share to B, A sees shares both to B and C in the UI. There is no visible distinction between the share and the reshare.
After deleting the share to B, the share to C is untouched in the UI, so A knows that he need to delete it if needed. As it’s now a direct share he can also edit the permissions on it.

lib/private/Share20/Manager.php
icewind1991: Ok, I didn't catch that the check was done for the original reshare, not the updated one.

Some comment about this logic would be appreciated

lib/private/Share20/Manager.php
```diff
@@ -1010,6 +1012,81 @@ protected function deleteChildren(IShare $share) {
...
+					foreach ($shares as $child) {
+						$reshareRecords[] = $child;
+					}
+				}
+			}
+		}
+
+		foreach ($reshareRecords as $child) {
+			try {
+				/* Check if the share is still valid (means the resharer still has access to the file through another mean */
```
solracsf, line 1075:
```suggestion
				/* Check if the share is still valid (means the resharer still has access to the file through another mean) */
```

icewind1991 (approved):

come-nc:
Rebased and fixed code style.

yemkareems (approved):

lib/private/Share20/Manager.php
```diff
@@ -1039,6 +1041,81 @@ protected function deleteChildren(IShare $share) {
...
+			foreach ($shareTypes as $shareType) {
+				$provider = $this->factory->getProviderForType($shareType);
+				$shares = $provider->getSharesBy($userId, $shareType, $node, false, -1, 0);
+				foreach ($shares as $child) {
+					$reshareRecords[] = $child;
+				}
+			}
+
+			if ($node instanceof Folder) {
+				$sharesInFolder = $this->getSharesInFolder($userId, $node, false);
```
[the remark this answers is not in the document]
come-nc: Shallow was removed in https://github.com/nextcloud/server/pull/44460

So, indeed, no idea how to get all shares in a folder subtree apart from recursing which would kill performance.
@icewind1991 Is there a trick we can use?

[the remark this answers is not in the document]
icewind1991: List all shares for a user and then filter them by path.

It's far from great but should be better than recursing into folders

icewind1991 (approved):

come-nc:
I’m not sure if cypress failures can be related or not. Should be double check as sharing sidebar is on the failure screenshot.

susnux:
> I’m not sure if cypress failures can be related or not. Should be double check as sharing sidebar is on the failure screenshot.

Never saw that failing but is is 99,99% unrelated because what is tested is purely frontend behavior and unrelated to your changes

luka-nextcloud:
@come-nc I think need to create backport PRs for 28, 29, 30.

Issue #44287, opened 2024-03-18
[Bug]: Missing remote share indicator on receiving shares


blizzz:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

When a user receives and accepts a federated share, the share indicator in the files table is missing – it is an empty icon, due to the fact that the avatar endpoint returned a 404 – there is no support for federated avatars in server so far:

![federates-share-indicator](https://github.com/nextcloud/server/assets/2184312/3d46b96b-6fc1-4c28-a34c-bc74dfd96a4b)

Back at NC 27, the uid of the remote user was shown at least. Not great, but better than nothing.

![federates-share-indicator-27](https://github.com/nextcloud/server/assets/2184312/91bc89a8-6259-47c9-9cc3-2f04181b2022)

### Steps to reproduce

1. As User A on Nextcloud F share a file with User B on Nextcloud L
2. As User B, accept the share and refresh the files page
3. See the incoming file, and check the share indicator

### Expected behavior

- the remote avartar, the remote user name, at least remote user id, or an icon that the files was received from a remote user

### Installation method

None

### Nextcloud Server version

28

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

### Is this bug present after an update or on a fresh install?

None

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [ ] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

skjnldsv:
Duplicate of https://github.com/nextcloud/server/issues/41725

skjnldsv:
I guess we can also use the Guest avatar endpoint to show some avatar too :)

Fenn-CS:
:eyes:  https://github.com/nextcloud/server/pull/47441

PR #47287, opened 2024-08-16
fix: Access node owner by top level `owner` property

susnux:
* Resolves https://github.com/nextcloud/server/issues/47267

## Summary

As suggested by @skjnldsv in the previous PR we should not use `owner-id` directly but just use the top level `owner` attribute (which will be set to owner-id anyways by default).
This fixes the mentioned issue, moreover I added some tests for the inline status action and a regression test for this bug.

(We nevertheless should fix the wrong type of the `owner-id` attribute in `davResultToNode` and cast it to string, but we can do so later in `@nextcloud/files`)

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [ ] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #47441, opened 2024-08-22
[stable28] fix(FileList): Show correct avatar for federated share owner

nfebe:
## Screenshot

Screenshot from 2024-09-18 16-54-57

Resolves : https://github.com/nextcloud/server/issues/44287


apps/files_sharing/src/utils/AccountIcon.ts
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+import { generateUrl } from '@nextcloud/router'
+import GuestSvg from '@mdi/svg/svg/account.svg?raw'
+
+const isDarkMode = window?.matchMedia?.('(prefers-color-scheme: dark)')?.matches === true
+	|| document.querySelector('[data-themes*=dark]') !== null
+
+export const generateAvatarSvg = (userId: string, isGuest = false) => {
+	if (isGuest) {
```
Fenn-CS, line 12: This `isGuest` is always a false negative in the positive case. I guess because https://github.com/nextcloud/server/pull/47287 is absent.

cc: @susnux ??

Pytal:
Federated user avatars should be shown with initials I believe rather than account.svg

apps/files_sharing/src/utils/AccountIcon.ts
susnux: `isGuest` is set by `node.attributes.remote_id` so should be unrelated or am I missing something?

susnux: (I mean within `isExternal`)

But I do not see that this attribute is somewhere set?
Maybe we should check for sharetype TYPE_REMOTE ?

Fenn-CS:
> Federated user avatars should be shown with initials I believe rather than account.svg

Yes that was the goal, but it seems like that would require a much larger refactoring as the FileActions interface turns to only accept svgs? @skjnldsv  is this the case?

susnux:
> Yes that was the goal, but it seems like that would require a much larger refactoring as the FileActions interface turns to only accept svgs?

Sadly our guest avatar is always a PNG, but you should be able to include that PNG in a SVG:
```svg
<svg xmlns="http://www.w3.org/2000/svg"
     xmlns:xlink="http://www.w3.org/1999/xlink">
    <image
        width="64" height="64"
        xlink:href="GUEST_URL"
        />
</svg>
```

(`GUEST_URL` = `http://domain.tld/index.php/avatar/guest/{GUEST_NAME}/64`)

Fenn-CS:
> Sadly our guest avatar is always a PNG, but you should be able to include that PNG in a SVG:

A second look showed me earlier that, the existing code should actually do the wrapping and should work, and the point of failure is that the guest detection does not actually work so it always tries to handle guest avatar generation as internal users.

@susnux Is that part of what https://github.com/nextcloud/server/pull/47287 resolved? Is it portable down to 28?

lib/private/User/LazyUser.php
```diff
@@ -167,4 +172,16 @@ public function getManagerUids(): array {
...
+		try {
+			// If getUser succeeds then user is definitely not federated
+			// This could fail for other reasons (especially a race condition where a user is deleted)
+			// But it's what we have now
+			$this->getUser();
+			return false;
+		} catch (\Exception $e) {
+			return true;
+		}
+	}
```
come-nc, line 186: Not sure I like that but I’m not sure what’s the better solution.
Why are you targetting stable28?

Fenn-CS: I would slowly forward port to master, this bug is currently properly reproducible on 28.

apps/files_sharing/src/actions/sharingStatusAction.ts
```diff
@@ -110,7 +98,8 @@
 
 		const ownerId = node?.attributes?.['owner-id']
 		if (ownerId && (ownerId !== getCurrentUser()?.uid || isExternal(node))) {
-			return generateAvatarSvg(ownerId, isExternal(node))
+			const sanitizeId = (id: string) => id.replace(/[^a-zA-Z0-9._%+-@]+/g, '').replace(/\//g, '')
```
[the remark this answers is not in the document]
skjnldsv: Good suggestion from Sir Copilot @Fenn-CS

lib/private/User/User.php
```diff
@@ -618,4 +618,10 @@ public function triggerChange($feature, $value = null, $oldValue = null) {
 			$this->emitter->emit('\OC\User', 'changeUser', [$this, $feature, $value, $oldValue]);
 		}
 	}
+
+	public function isFederated(): bool {
+		// Federated users are instantiated via LazyUser
+		// DAV also only uses LazyUser
+		return false;
```
skjnldsv, line 625: Still not 100% sure of this. 
@icewind1991 please confirm

icewind1991: `IUser` should only refer to local users. The fact that `$node->getOwner` will return an `IUser` for federated users is a bug and should be fixed.

come-nc (approved):

skjnldsv (approved):

PR #47465, opened 2024-08-24
fix(files): Reset drop notice on firefox

susnux:
## Summary
On Firefox there is an old bug where when you move a dragged file outside the window the `dragleave` event is never emitted. So we just use a timeout to reset the drag over state.

Also a small change: Use the ID of the main container instead of relying on tag name and class. (The ID is guaranteed as other APIs rely on it, while the class is just used internally).

### Screen recording

This is how the bug looks (before this PR):

[Bildschirmaufnahme_20240824_120143.webm](https://github.com/user-attachments/assets/878da007-35c3-454f-8aa6-d42e6af0028e)

## Checklist


susnux:
Firefox bugs that cause this:
1. https://bugzilla.mozilla.org/show_bug.cgi?id=1427644
2. https://bugzilla.mozilla.org/show_bug.cgi?id=656164

And yes those are open since Firefox 4(!)

skjnldsv (approved):

apps/files/src/components/DragAndDropNotice.vue
```diff
@@ -86,18 +88,29 @@ export default defineComponent({
...
+			return debounce(() => {
+				this.dragover = false
+			}, 3000)
+		},
 	},
 
 	mounted() {
 		// Add events on parent to cover both the table and DragAndDrop notice
-		const mainContent = window.document.querySelector('main.app-content') as HTMLElement
+		const mainContent = window.document.getElementById('app-content-vue') as HTMLElement
```
artonge, line 106: To stay explicit about targeting a `main` element, as the var name suggest?

```suggestion
		const mainContent = window.document.getElementById('main#app-content-vue') as HTMLElement
```

apps/files/src/components/DragAndDropNotice.vue
```diff
@@ -112,6 +125,7 @@ export default defineComponent({
 			if (isForeignFile) {
 				// Only handle uploading of outside files (not Nextcloud files)
 				this.dragover = true
+				this.resetDragOver()
```
artonge, line 128: So the drag will time out after 3s of inactivity, and as long as the user moves, the timeout is pushed back, right?

apps/files/src/components/DragAndDropNotice.vue
```diff
@@ -187,15 +203,21 @@ export default defineComponent({
...
+					// Keep params but change fileid
 					params: {
-						view: this.$route.params?.view ?? 'files',
-						fileid: parseInt(lastUpload.response!.headers['oc-fileid']),
+						...this.$route.params,
+						fileid: String(lastUpload.response!.headers['oc-fileid']),
+					},
+					// Query should only contain the dir, but not `openfile`
+					query: {
+						dir: this.$route.query.dir,
```
artonge, line 214: If we only want to remove openfile, then:

```js
delete this.$route.query.openfile
...
query: {
	...this.$route.query
}
```

apps/files/src/components/DragAndDropNotice.vue
```diff
@@ -187,15 +203,21 @@ export default defineComponent({
 			if (lastUpload !== undefined) {
 				logger.debug('Scrolling to last upload in current folder', { lastUpload })
 				this.$router.push({
-					...this.$route,
+					path: this.$route.path,
```
artonge, line 206: We not keep `...this.$route,`?

apps/files/src/components/DragAndDropNotice.vue
susnux: No as stated in the initial post the element tag name is not part of any contracts.
Only the ID is guaranteed to work, the idea behind is fixing this right now because with Vue3 this will break¹.
Side effect: Query by ID is much faster than including the tag name (does not matter that much but the query is factor 2-10 faster).

¹: Because our template looks like this `<main id="content">...</main>` and our Vue app will be mounted to `#content`.
With Vue2 the element would be replaced, so it looks like `<main id="app-content-vue">...</main>`.
But Vue3 will render within and not replace so it looks like `<main id="content"><div id="app-content-vue">...</div></main>`.

apps/files/src/components/DragAndDropNotice.vue
susnux: > So the drag will time out after 3s of inactivity, and as long as the user moves, the timeout is pushed back, right?

Yes, exactly. Or even without move but still on the page. As per documentation:
> The drag event is fired **every few hundred milliseconds** as an element or text selection is being dragged by the user.

apps/files/src/components/DragAndDropNotice.vue
susnux: `vue-router` bug in the Vue2 version, this causes Typescript errors as `this.$route` is from type `Route` while `$router.push` expects `RawLocation`, which both are incompatible because their property `name` has incompatible types.

apps/files/src/components/DragAndDropNotice.vue
susnux: Sure if you think that is cleaner :)
Adjusted the code.

apps/files/src/components/DragAndDropNotice.vue
```diff
@@ -186,16 +203,21 @@ export default defineComponent({
...
 					params: {
-						view: this.$route.params?.view ?? 'files',
-						fileid: parseInt(lastUpload.response!.headers['oc-fileid']),
+						...this.$route.params,
+						fileid: String(lastUpload.response!.headers['oc-fileid']),
 					},
-				})
+				}
+				// Remove open file from query
+				delete location.query?.openfile
```
artonge, line 215: You're not restoring the previous `query` anymore, right? Currently, `location.query?` will always be undefined.

apps/files/src/components/DragAndDropNotice.vue
```diff
@@ -186,16 +203,21 @@ export default defineComponent({
...
-					...this.$route,
+				const location: RawLocation = {
+					path: this.$route.path,
+					// Keep params but change file id
 					params: {
-						view: this.$route.params?.view ?? 'files',
-						fileid: parseInt(lastUpload.response!.headers['oc-fileid']),
+						...this.$route.params,
+						fileid: String(lastUpload.response!.headers['oc-fileid']),
 					},
```
susnux, line 212:
```suggestion
					},
					query: {
						...this.$route.query,
					},
```

apps/files/src/components/DragAndDropNotice.vue
susnux: fixed it! Now this is always true so can remove the optional chaining:

```suggestion
				delete location.query.openfile
```

susnux:
BTW not sure if this has side effects, but to prevent them if they exist I do not like
`delete this.$route.query.XY` but prefer creating a copy first so we do not trigger side effects until the `push`.

artonge (approved):

PR #47467, opened 2024-08-24
fix(files): Adjust margin at the file list bottom

susnux:
## Summary
IIRC this was added when the drop notice was placed on the bottom, but the notice is now always set on the top. So this just causes weird whitespace on the files list.

### Screenshots (scrolled fully down)

before | after
---|---
Screenshot 2024-08-24 at 17-11-47 Files - Nextcloud|Screenshot 2024-08-24 at 17-34-36 Files - Nextcloud
Screenshot 2024-08-24 at 17-12-38 Files - Nextcloud|Screenshot 2024-08-24 at 17-35-02 Files - Nextcloud

## Checklist


skjnldsv:
> IIRC this was added when the drop notice was placed on the bottom, but the notice is now always set on the top. So this just causes weird whitespace on the files list.

This was also more visually appealing.
It gives a feeling of completion and you know you've reached the bottom.
I don't really mind, whatever we go for here, just adding some old input I remember hearing about :)

artonge (approved):

jancborchardt (changes requested):
Yes, as @skjnldsv mentioned, the bottom spacing is intentional, so that:
- It is obvious you reached the end and it’s not loading anymore
- The last items and summary are nicely readable in the center of the screen and not stuck at the bottom

(Some other apps like some text editors do this similarly.)

susnux:
> It is obvious you reached the end and it’s not loading anymore

I thought exactly the opposite, as scrolling into a blank page looks like something needs to load there :sweat_smile: 

But if this was a design decision, then maybe at least adjust the number to not be hard coded pixels but something variable like `25vh`, so it works on all screens?

artonge:
> (Some other apps like some text editors do this similarly.)

But this is not a text editor. Cursed words incoming: "What's the competition doing?". I don't think this is expected from a file list, or any list in general, no?

jancborchardt:
> But if this was a design decision, then maybe at least adjust the number to not be hard coded pixels but something variable like 25vh, so it works on all screens?

Sounds good @susnux! :) 

@artonge it looks better, and some times inbetween the space was also missing and it resulted in it looking off. It’s just weird when the last files and the summary is stuck to the bottom of the screen.

susnux:
I can understand both (personally I also tend to say a list that overflows one page should only be scroll able to the bottom), but also the design decision to have the margin seems reasonable, so I will adjust the PR to be relative to screen size.

(Because whats worse is e.g. VS code where you can scroll a whole page below the last line of code :see_no_evil: )

jancborchardt (approved):
Thank you! :)

susnux:
Done looks like this now:

before | after
---|---
Screen Shot 2024-08-26 at 16 29 46|Screen Shot 2024-08-26 at 16 28 53

PR #47474, opened 2024-08-25
fix(cs-fixer): Correctly ignore files ignored by git

provokateurin:
## Summary

Many files ignored by git were never ignored by cs-fixer due to https://github.com/symfony/symfony/issues/37781.
If it was actually working as intended, then the code for ignoring the additional app directories wouldn't have been necessary.

This implementation is more generic and really ignores the files ignored by git.
Instead of scanning 7075 files (also from additional apps, your count will be different) it only has to scan 4493.
While being faster, this also stops cs-fixer from messing with formatting in additional apps, if they were placed in the `apps/` directory.

I manually added back the ignores for build/stubs and composer since we don't control or care about their content formatting wise.
As for the config dir, we might want to remove the exclusion and have the config.sample.php correctly formatted. Let me know if I should change that.

According to https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/config.rst#configuring-paths the `notPath` calls were also wrong because they are only for files and `exclude` should have been used instead.

## Checklist

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -14,29 +14,23 @@ $config = new Config();
 $config
 	->setParallelConfig(ParallelConfigFactory::detect())
 	->getFinder()
-	->ignoreVCSIgnored(true)
 	->exclude('config')
-	->exclude('data')
-	->notPath('3rdparty')
-	->notPath('build/integration/vendor')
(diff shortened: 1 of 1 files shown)


kesselb (approved):
Nice one

ChristophWurst (approved):

PR #47495, opened 2024-08-26
feat: Delegate is-encrypted handling to the e2ee app

artonge:
The is-encrypted property is signaling the E2EE status, not the SSE one. So it does not make sense to keep it in the server repo.

- E2EE counterpart: https://github.com/nextcloud/end_to_end_encryption/pull/733


come-nc (approved):

come-nc:
Wait, but FileInfo::isEncrypted is used for SSE and not E2EE, no?

artonge:
> Wait, but FileInfo::isEncrypted is used for SSE and not E2EE, no?

Not sure that I get the concern, but in this case, `isEncrypted()` was called inside a `$node instanceof Directory` condition, so this was only targeting E2EE folders and not SSE files.

sorbaugh (approved):

PR #47507, opened 2024-08-27
feat(settings): Add setup check for apcu cache expunge

### LINKED ISSUE #46282: Setup check for apcu cache statistics
### How to use GitHub

We noticed that with a too low [apc.shm_size](https://www.php.net/manual/en/apcu.configuration.php#ini.apcu.shm-size) the local cache could be invalidated very frequently where apcu purges all keys. Frequent invalidation can potentially have an impact on performance so we should add a setup check to give hints to admins for that.

- Warn about frequent invalidation mabye based on uptime and cache full count
- Other measurements depending on hit/miss rate, insert rate or cache growth if measurable could be also taken into account

Expose this to monitoring, but the serverinfo app already does that and likely gets overseen if admins don't dive too deep into metrics.

**Additional context**

Official script for getting insights https://raw.githubusercontent.com/krakjoe/apcu/master/apc.php

juliusknorr:
* Resolves: https://github.com/nextcloud/server/issues/46282

## Summary

This adds a basic setup check to warn admins if their APCu cache is close to be full or has been expunged. Ideally we would be able to calculate something like a rate for how long the cache is kept until cleaned up, but i couldn't find a reasonable metric and we cannot get the startup time of the php service itself.

## Checklist

### DIFF
--- a/apps/settings/composer/composer/autoload_classmap.php
+++ b/apps/settings/composer/composer/autoload_classmap.php
@@ -108,6 +108,7 @@ return array(
     'OCA\\Settings\\SetupChecks\\MysqlUnicodeSupport' => $baseDir . '/../lib/SetupChecks/MysqlUnicodeSupport.php',
     'OCA\\Settings\\SetupChecks\\OcxProviders' => $baseDir . '/../lib/SetupChecks/OcxProviders.php',
     'OCA\\Settings\\SetupChecks\\OverwriteCliUrl' => $baseDir . '/../lib/SetupChecks/OverwriteCliUrl.php',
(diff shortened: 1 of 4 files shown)


skjnldsv (approved):

skjnldsv:
Nice!

nickvergessen (approved):

come-nc (commented):
Does that apply to cli as well? Remember that setupchecks can be run by occ and in this case they use the php cli ini file which is not configured the same as web.

come-nc:
> Right, we cannot really check in cli as there is no shared memory with what ever SAPI is used.

Then the check should check the CLI flag and bail out, take inspiration from other php config related checks.

Issue #47479, opened 2024-08-23
[Bug]: "Could not set global default apps"


fulldecent:
### ⚠️ This issue respects the following points: ⚠️

- [X] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [X] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [X] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [X] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Cannot set an external link as the global default app

### Steps to reproduce

1. Install app: External sites
2. Create an external site (will show up on top bar)
3. Install app: Theming
4. Go to Administration settings > Theming
5. Turn on: Use custom default app
6. X out one of the existing ones (there seems to be a limit of two)
7. Add your own external site that you created above

### Expected behavior

External site is added as the custom default app

---

Actual behavior: error message:

> Could not set global default apps

![Screenshot 2024-08-23 at 13 36 31](https://github.com/user-attachments/assets/23d84a46-7e73-4225-8a6b-37b54a549f86)

### Installation method

Official All-in-One appliance

### Nextcloud Server version

28

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.1

### Web server

Apache (supported)

### Database engine version

MySQL

### Is this bug present after an update or on a fresh install?

Fresh Nextcloud Server install

### Are you using the Nextcloud Server Encryption module?

None

### What user-backends are you using?

- [X] Default user-backend _(database)_
- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

provokateurin:
Unfortunately it is quite a severe bug that needs a bunch of refactoring.
Luckily this issue doesn't have a high impact since most apps register one navigation entry at most and use the same value for `id` and `app`.

fulldecent:
@provokateurin thank you for looking into this. Could you please spell out the details of your research and the parts necessary for such a refactor?

provokateurin:
I'm already working on it! You'll be able to see it in the PR (both the changes and the explanation of the problem).

[...]

PR #47508, opened 2024-08-27
fix(Router): Load attribute routes of all apps when not app is specified

provokateurin:
## Summary

Found while debugging https://github.com/nextcloud/server/issues/47479 (also necessary for the fix to it).

The problem was that when `loadRoutes(null)` was called the attribute routes should have been loaded for all apps since no app was specified. It only loaded the traditional routes.php files and then continued to set `loaded = true`. Any future call would return early, even if an app was specified because everything must be loaded already.
This logic now matches the loading process of routes.php which it should have done from the start.
Unfortunately back then I didn't find this problem and it seems to only trigger in specific places where the app is not known (e.g. when loading the `/index.php` path where the server has to load the routes first and then has to figure out which app should be used).

## Checklist

### DIFF
--- a/lib/private/Route/Router.php
+++ b/lib/private/Route/Router.php
@@ -111,9 +111,14 @@ class Router implements IRouter {
 		if ($this->loaded) {
 			return;
 		}
+		$this->eventLogger->start('route:load:' . $requestedApp, 'Loading Routes for ' . $requestedApp);
 		if (is_null($app)) {
 			$this->loaded = true;
 			$routingFiles = $this->getRoutingFiles();
+
+			foreach (\OC_App::getEnabledApps() as $enabledApp) {
+				$this->loadAttributeRoutes($enabledApp);
+			}
 		} else {
 			if (isset($this->loadedApps[$app])) {
 				return;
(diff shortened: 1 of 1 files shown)


lib/private/Route/Router.php
```diff
@@ -125,23 +129,10 @@ public function loadRoutes($app = null) {
...
-				$this->root->addCollection($collection);
 
-				// Also add the OCS collection
-				$collection = $this->getCollection($requestedApp . '.ocs');
-				$collection->addPrefix('/ocsapp');
-				$this->root->addCollection($collection);
-			}
+			$this->loadAttributeRoutes($app);
 		}
+		$this->eventLogger->start('route:load:' . $requestedApp, 'Loading Routes for ' . $requestedApp);
```
nickvergessen, line 135: this is now not the start anymore?

lib/private/Route/Router.php
```diff
@@ -413,6 +404,27 @@ protected function fixLegacyRootName(string $routeName): string {
 		return $routeName;
 	}
 
+	private function loadAttributeRoutes(string $app): void {
+		if (!in_array($app, \OC_App::getEnabledApps())) {
```
come-nc, line 408:
```suggestion
		if (!$this->appManager->isEnabledForUser($app)) {
```

come-nc (approved):

artonge (approved):

PR #47519, opened 2024-08-27
feat(transfer-ownership): Correctly react to encrypted files

come-nc:
* Resolves: # 

## Summary

1. For E2EE encrypted files, we abort the transfer.
2. For SSE encrypted files, we abort only if not using master key. 
3. Also fixed the check for when the path to a single file is used.

## Checklist

### DIFF
--- a/apps/files/lib/Service/OwnershipTransferService.php
+++ b/apps/files/lib/Service/OwnershipTransferService.php
@@ -245,21 +245,21 @@ class OwnershipTransferService {
 								return false;
 							}
 							if ($fileInfo->isEncrypted()) {
-								/* Encrypted folder means e2ee encrypted */
+								/* Encrypted folder means e2ee encrypted, we cannot transfer it */
 								$encryptedFiles[] = $fileInfo;
 							}
 							return true;
 						}
 						$progress->advance();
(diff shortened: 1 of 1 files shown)


ChristophWurst (approved):
This is not my area of expertise but the changes make sense and look good

apps/files/lib/Service/OwnershipTransferService.php
```diff
@@ -204,41 +204,64 @@ private function walkFiles(View $view, $path, Closure $callBack) {
...
+							}
+							if ($fileInfo->isEncrypted()) {
+								/* Encrypted folder means e2ee encrypted */
+								$encryptedFiles[] = $fileInfo;
+							}
+							return true;
+						}
+						$progress->advance();
+						if ($fileInfo->isEncrypted() && !$masterKeyEnabled) {
+							/* Encrypted file means SSE, we only care if we are using user keys */
```
artonge, line 255: What does "care" mean? Does that better convey the situation?

```suggestion
							/* Encrypted file means SSE, we can only transfert them if master key is enabled. */
```

come-nc: Yeah but no t on transfer

artonge (approved):

PR #47521, opened 2024-08-27
fix(logger): ignore session logging during setup

ArtificialOwl:
This fix a regression from #47105 that lock installation process on 29.0.5 (but no 30 😕 ) when using the web interface

### DIFF
--- a/lib/private/Session/Internal.php
+++ b/lib/private/Session/Internal.php
@@ -29,8 +29,10 @@ class Internal extends Session {
 	 * @param string $name
 	 * @throws \Exception
 	 */
-	public function __construct(string $name,
-		private LoggerInterface $logger) {
+	public function __construct(
+		string $name,
+		private ?LoggerInterface $logger,
+	) {
 		set_error_handler([$this, 'trapError']);
 		$this->invoke('session_name', [$name]);
 		$this->invoke('session_cache_limiter', ['']);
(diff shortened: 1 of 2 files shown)


nickvergessen:
Or we wrap the code with an `isInstalled()` check?

ArtificialOwl:
if can we ignore slow session logging during installation process, I could check this in `base.php` and send a null logger to the Session

nickvergessen (approved):
I think both makes sense.

ArtificialOwl:
@nickvergessen please check this new way of fixing the issue.

lib/private/Session/Internal.php
```diff
@@ -29,8 +29,10 @@ class Internal extends Session {
 	 * @param string $name
 	 * @throws \Exception
 	 */
-	public function __construct(string $name,
-		private LoggerInterface $logger) {
+	public function __construct(
+		string $name,
+		private ?LoggerInterface $logger = null,
```
nickvergessen, line 34:
```suggestion
		private ?LoggerInterface $logger,
```

ArtificialOwl:
Thanks. Fixed, squashed and rebased

artonge (approved):

AndyScherzinger (approved):

joshtrichards:
> (no idea if an urgent new release of 29 is needed)

This breaks web based installations. While `occ maintenance:install` is a workaround, this is a serious bug. I think we need to do a new urgent release.

PR #47522, opened 2024-08-27
[TaskProcessing] Add manager::runTask method

julien-nc:
Add `OCP\TaskProcessing\IManager::runTask` method to run tasks synchronously.
* If the provider is synchronous (Php), the task runs in the same process.
* If not, runTask polls the task that should be picked by an external provider ASAP.

The changeset looks big but it's just moving stuff from the background job to the manager and refactoring the manager a bit to factorize stuff between `scheduleTask` and `runTask`.

Additional change: Fix the condition to reschedule the background job. As discussed with @kyteinsky, knowing that some synchronous providers can handle currently scheduled tasks is not enough. They have to be a preferred provider (selected in the AI admin settings).

This should be backported to stable30.

### DIFF
--- a/lib/private/TaskProcessing/Manager.php
+++ b/lib/private/TaskProcessing/Manager.php
@@ -716,53 +716,67 @@ class Manager implements IManager {
 		if (!$this->canHandleTask($task)) {
 			throw new \OCP\TaskProcessing\Exception\PreConditionNotMetException('No task processing provider is installed that can handle this task type: ' . $task->getTaskTypeId());
 		}
-		$taskTypes = $this->getAvailableTaskTypes();
-		$inputShape = $taskTypes[$task->getTaskTypeId()]['inputShape'];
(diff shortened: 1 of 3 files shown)


bigcat88:
just to clarify, this doesn't affect our APIs in any way, everything remains compatible with what we did - is it just adding a synchronous method?

julien-nc:
@bigcat88 Yes, just adding something.

bigcat88 (approved):

AndyScherzinger (approved):

AndyScherzinger:
@julien-nc is there a way for the old, synchronous API to call this new one? or the implementation behind it? This way the mobile apps won't have to migrate right away (which would else also need to be managed right away now).

julien-nc:
@AndyScherzinger I would have to check with Marcel. IMO it could work, we would only do it for the hardcoded task types in text processing.

PR #47523, opened 2024-08-27
Use NavigationManager instead of AppManager to handle custom apps order

### LINKED ISSUE #47479: [Bug]: "Could not set global default apps"

provokateurin:
* Resolves: https://github.com/nextcloud/server/issues/47479

## Summary

The custom defaults "apps" and order were using the apps and not navigation entries (see the linked issue for more details).
The confusion lead to a few problems that can only be solved by only using the NavigationManager and not the AppManager.
Most of the logic is the same, except for checking against available navigation entries and not apps.

I tried touching as little as possible, all the frontend code is still the same and uses the incorrect "app" over "navigation entry" concept. On the frontend it still makes a bit sense, since the concept is a lot easier to grasp for users.
The config keys on the backend are kept the same to avoid migrations and deprecations.

In theory one could spend a lot more time on trying to fix this, but I already tried that a bit and it is a rabbit hole that is too deep for fixing and backporting.

The last commit is a fix that is useful even without all the other changes. In case an app is disabled while also being used for a custom order, the frontend would display "undefined" entries.

As for backporting, I kept the old broken interface by providing a proxy implementation. I hope this is good enough to be backported.

## Checklist


lib/private/App/AppManager.php
```diff
@@ -82,6 +83,14 @@ public function __construct(
 	) {
 	}
 
+	private function getNavigationManager(): INavigationManager {
+		if ($this->navigationManager !== null) {
+			return $this->navigationManager;
+		}
+		$this->navigationManager = \OCP\Server::get(INavigationManager::class);
```
susnux, line 90: Could save one line ;)

```suggestion
		if ($this->navigationManager === null) {
			$this->navigationManager = \OCP\Server::get(INavigationManager::class);
		}
```

lib/private/URLGenerator.php
```diff
@@ -288,14 +298,14 @@ public function linkToDefaultPageUrl(): string {
...
-		$appId = $this->getAppManager()->getDefaultAppForUser();
-
-		if ($this->config->getSystemValueBool('htaccess.IgnoreFrontController', false)
-			|| getenv('front_controller_active') === 'true') {
-			return $this->getAbsoluteURL('/apps/' . $appId . '/');
+		$entryId = $this->getNavigationManager()->getDefaultEntryIdForUser();
+		$entry = $this->getNavigationManager()->get($entryId);
+		$href = (string)$entry['href'];
+		if ($href === '') {
+			throw new \InvalidArgumentException('Missing href');
```
susnux, line 305:
```suggestion
			throw new \InvalidArgumentException('Default navigation entry is missing href');
```

tests/lib/App/AppManagerTest.php
```diff
@@ -733,159 +733,6 @@ public function testGetAppRestriction() {
 		$this->assertEquals(['foo'], $this->manager->getAppRestriction('test3'));
 	}
 
-	public function provideDefaultApps(): array {
```
susnux, line 736: The function still exists, no?

susnux: It is deprecated right? I saw you added tests for the new functions, so that would be enough for me :)

provokateurin: Yes. The new test is just the same as this removed test but adapted to the changes.

lib/private/App/AppManager.php
```diff
@@ -820,60 +828,42 @@ public function getDefaultEnabledApps(): array {
...
+	 * @inheritdoc
+	 */
 	public function getDefaultApps(): array {
-		return explode(',', $this->config->getSystemValueString('defaultapp', 'dashboard,files'));
+		$ids = $this->getNavigationManager()->getDefaultEntryIds();
+
+		return array_map(function (string $id) {
+			$entry = $this->getNavigationManager()->get($id);
+			return (string)$entry['app'];
+		}, $ids);
```
come-nc, line 849:
```suggestion
		return array_values(array_unique(array_map(function (string $id) {
			$entry = $this->getNavigationManager()->get($id);
			return (string)$entry['app'];
		}, $ids)));
```
Should remove duplicates, no?

lib/private/NavigationManager.php
```diff
@@ -93,7 +101,7 @@ public function add($entry) {
 			}
 
 			// This is the default app that will always be shown first
-			$entry['default'] = ($entry['app'] ?? false) === $this->defaultApp;
+			$entry['default'] = ($entry['app'] ?? false) === $this->defaultEntry;
```
come-nc, line 104: This is comparing app id with entry id, no?

lib/private/NavigationManager.php
```diff
@@ -401,4 +409,80 @@ private function isSubadmin() {
...
+
+	public function get(string $id): array|null {
+		foreach ($this->getAll() as $entry) {
+			if ($entry['id'] === $id) {
+				return $entry;
+			}
+		}
+
+		return null;
+	}
```
come-nc, line 421: Looks innefficient, should we add an index? How often is that called?

lib/private/NavigationManager.php
```diff
@@ -401,4 +409,80 @@ private function isSubadmin() {
...
+		}
+		return array_filter($ids);
+	}
+
+	public function setDefaultEntryIds(array $ids): void {
+		$entryIds = array_map(static fn (array $entry) => $entry['id'], $this->getAll());
+
+		foreach ($ids as $id) {
+			if (!in_array($id, $entryIds, true)) {
+				$this->logger->debug('Can not set unavailable entry as default entry', ['missing_entry' => $id]);
```
come-nc, line 481:
```suggestion
				$this->logger->debug('Cannot set unavailable entry as default entry', ['missing_entry' => $id]);
```

lib/private/TemplateLayout.php
```diff
@@ -99,11 +99,10 @@ public function __construct($renderAs, $appId = '') {
 			$logoUrl = $this->config->getSystemValueString('logo_url', '');
 			$this->assign('logoUrl', $logoUrl);
 
-			// Set default app name
-			$defaultApp = \OC::$server->getAppManager()->getDefaultAppForUser();
-			$defaultAppInfo = \OC::$server->getAppManager()->getAppInfo($defaultApp);
-			$l10n = \OC::$server->get(IFactory::class)->get($defaultApp);
-			$this->assign('defaultAppName', $l10n->t($defaultAppInfo['name']));
+			// Set default entry name
+			$defaultEntryId = \OC::$server->getNavigationManager()->getDefaultEntryIdForUser();
```
come-nc, line 103:
```suggestion
			$defaultEntryId = \OCP\Server::get(INavigationManager::class)->getDefaultEntryIdForUser();
```
Same for the other line.

come-nc (commented):
What happens if you actually set an external site as default entry, you cannot use Nextcloud anymore? (login would redirect to external site, no?)

lib/private/App/AppManager.php
come-nc: Could save all of them:
```php
return $this->navigationManager ??= \OCP\Server::get(INavigationManager::class);
```

But who’s counting?

come-nc:
Is the comment in https://github.com/nextcloud/server/blob/558877c8422c1a7083d083b2fe0be553f3047ba0/config/config.sample.php#L1169-L1178 still true after your PR is merged?

lib/private/NavigationManager.php
provokateurin: Agreed, I already found it weird that the entries are a list and not a hashmap.

provokateurin: Ah it already is, I just didn't check...

provokateurin:
Cypress failure is related for once :sweat_smile:

provokateurin:
Fix will not be backported to any version due to the new interfaces and the close release of 30.

lib/private/URLGenerator.php
```diff
@@ -288,14 +298,17 @@ public function linkToDefaultPageUrl(): string {
...
-		if ($this->config->getSystemValueBool('htaccess.IgnoreFrontController', false)
-			|| getenv('front_controller_active') === 'true') {
-			return $this->getAbsoluteURL('/apps/' . $appId . '/');
+		$entryId = $this->getNavigationManager()->getDefaultEntryIdForUser();
+		$entry = $this->getNavigationManager()->get($entryId);
+		$href = (string)$entry['href'];
+		if ($href === '') {
+			throw new \InvalidArgumentException('Default navigation entry is missing href: ' . $entryId);
+		}
+		if (str_starts_with($href, '/index.php/') && ($this->config->getSystemValueBool('htaccess.IgnoreFrontController', false) || getenv('front_controller_active') === 'true')) {
```
come-nc, line 307: Does that check works on instances with pretty url without the index.php?

provokateurin: The check is exactly for that purpose. It will only change the href if pretty-urls are available. We should create a common function for this check as it is used a few times in multiple places and it's not really self-explanatory.

come-nc (approved):

provokateurin:
Cypress issues are resolved (caused by bugs in the backend). @susnux can you review again?

susnux (approved):

nickvergessen:
This actually broke the guests app, which is going nasty stuff:
https://github.com/nextcloud/guests/blob/master/lib/FilteredNavigationManager.php

PR #47524, opened 2024-08-27
Fix/filename validation 2

susnux:
* Resolves: # 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/Connector/Sabre/FilesPlugin.php
+++ b/apps/dav/lib/Connector/Sabre/FilesPlugin.php
@@ -8,8 +8,11 @@
 namespace OCA\DAV\Connector\Sabre;
 
 use OC\AppFramework\Http\Request;
+use OCA\DAV\Connector\Sabre\Exception\InvalidPath;
 use OCP\Constants;
 use OCP\Files\ForbiddenException;
+use OCP\Files\IFilenameValidator;
+use OCP\Files\InvalidPathException;
 use OCP\Files\StorageNotAvailableException;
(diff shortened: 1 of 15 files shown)


apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -142,33 +140,67 @@ public function initialize(Server $server) {
...
 	 * @param string $source source path
-	 * @param string $destination destination path
-	 * @throws Forbidden
-	 * @throws NotFound
+	 * @param string $target target path
+	 * @throws NotFound If the source does not exist
+	 * @throws InvalidPath If the target is invalid
 	 */
-	public function checkMove($source, $destination) {
+	public function checkCopy($source, $target) {
```
susnux, line 154:
```suggestion
	public function checkCopy($source, $target): void {
```

apps/dav/lib/Connector/Sabre/FilesPlugin.php
```diff
@@ -142,33 +140,67 @@ public function initialize(Server $server) {
...
+	/**
+	 * Plugin that checks if a move can actually be performed.
+	 *
+	 * @param string $source source path
+	 * @param string $target target path
+	 * @throws Forbidden If the source is not deletable
+	 * @throws NotFound If the source does not exist
+	 * @throws InvalidPath If the target name is invalid
+	 */
+	public function checkMove($source, $target) {
```
susnux, line 191:
```suggestion
	public function checkMove(string $source, string $target): void {
```

PR #47535, opened 2024-08-27
Fix/move image to oc namespace

come-nc:
## Summary

Move legacy `OC_Image` into `\OC\Image` and update code style and remove deprecated references in there.
This class is exposed through `OCP\Image` which is used by applications with `new \OCP\Image()` and then `\OCP\IImage` methods can be used on it.

This still leaves a reference to `OC` from `OCP` which we try to avoid.
But this would mean adding an `IImageFactory` which applications needs to DI, not sure this is worth it.

## Checklist


lib/private/Image.php
```diff
@@ -171,7 +166,7 @@ public function show(?string $mimeType = null): bool {
 		if ($mimeType === null) {
 			$mimeType = $this->mimeType();
 		}
-		header('Content-Type: ' . $mimeType);
+		header('Content-Type: ' . ($mimeType ?? ''));
```
provokateurin, line 169: Should it be emitted if null? I'm not sure how clients would handle an empty Content-Type header. Or even fallback to `application/octet-stream` to be correct?

lib/public/Image.php
artonge, line 1: Maybe silly question, why not directly exposing the private `Image.php`?

lib/private/Image.php
artonge: Wouldn't the web server fill that based on the extension?

lib/public/Image.php
come-nc: Because then any change in the class would be a breaking change as it would be part of OCP.
While currently, as long as the class respects OCP\IImage interface it can be modified.
The point of OCP\Image class is only to allow for building an instance. It would be more logical if it contained a public constructor with no parameters though.

lib/private/Image.php
come-nc: It can only be null if image is invalid I think. But yeah you’re right it’s better to not call `header` at all if it’s null.

artonge (approved):

provokateurin (approved):

provokateurin:
This broke Talk https://github.com/nextcloud/spreed/pull/13197 and using the OCP replacement does not work.
I think the problem is the \OCP\Image depends on \OC\Image which does not exist in the dependency and thus all methods are missing.
Also some of the methods are only in \OC\Image and not in \OCP\Image or \OCP\IImage and thus are not visible.

PR #47541, opened 2024-08-27
ci: Reduce required run time

nickvergessen:
- ci: Run things we only run on 1 PHP version on the preinstalled 8.1 which installs ~2m faster
- ci: Don't run things normally and with coverage flag on the same configuration

### DIFF
--- a/.github/workflows/object-storage-azure.yml
+++ b/.github/workflows/object-storage-azure.yml
@@ -45,7 +45,7 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['8.1', '8.2', '8.3']
+        php-versions: ['8.1', '8.2']
         include:
           - php-versions: '8.3'
             coverage: true
--- a/.github/workflows/autocheckers.yml
+++ b/.github/workflows/autocheckers.yml
@@ -45,7 +45,7 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['8.3']
+        php-versions: ['8.1']
 
     name: PHP checkers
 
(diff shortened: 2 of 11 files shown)


susnux (approved):
looks good to me, hope github images move to 24.04 soon :)

provokateurin (approved):

.github/workflows/object-storage-azure.yml
```diff
@@ -45,7 +45,7 @@ jobs:
 
     strategy:
       matrix:
-        php-versions: ['8.1', '8.2', '8.3']
+        php-versions: ['8.1', '8.2']
         include:
           - php-versions: '8.3'
             coverage: true
```
come-nc, line 51: Not sure whether the include will not add back 8.3?
Also, would prefer `8.1,8.3` over `8.1,8.2`.

susnux: This change here will result in:
- PHP 8.1 coverage :x: 
- PHP 8.2 coverage :x: 
- PHP 8.3 coverage :white_check_mark:

solracsf:
Question: why don't use `ubuntu-20.04` image which has all php versions pre-installed?

https://github.com/shivammathur/setup-php?tab=readme-ov-file#github-hosted-runners

.github/workflows/object-storage-azure.yml
nickvergessen: > Not sure whether the include will not add back 8.3?

It will and that's exactly the fix. We ran 8.3 (and in other cases 8.2) twice. Once with the coverage flag and once without. But the only difference is if we generate and upload the report, so we can remove the 8.3 run that is duplicated by the run that also uploads the coverage report

susnux:
> Question: why don't use `ubuntu-20.04` image which has all php versions pre-installed?

Because we use self-hosted runners ;)

st3iny (approved):
:rocket:

PR #47581, opened 2024-08-28
fix: Do not fail to get internal path on NonExistingFile

juliusknorr:
Follow up to:
- https://github.com/nextcloud/server/pull/47252
- https://github.com/nextcloud/server/pull/44871

## Summary

If we cannot get the internal path from the file info (e.g. when creating a new file), but we can obtain it from the parent node through the mountpoint.

Fixes errors log spam when creating a new file and having admin_audit enabled when being called in https://github.com/nextcloud/server/blob/af6de04e9e141466dc229e444ff3f146f4a34765/apps/admin_audit/lib/Actions/Files.php#L166

### DIFF
--- a/lib/private/Files/Node/NonExistingFile.php
+++ b/lib/private/Files/Node/NonExistingFile.php
@@ -38,6 +38,14 @@ class NonExistingFile extends File {
 		}
 	}
 
+	public function getInternalPath() {
+		if ($this->fileInfo) {
+			return parent::getInternalPath();
+		} else {
+			return $this->getParent()->getMountPoint()->getInternalPath($this->getPath());
+		}
+	}
+
 	public function stat() {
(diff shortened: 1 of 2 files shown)


nickvergessen:
Just for the record, if you did this because you see it on our instance, the stable30 backport was merged after the last RC

artonge (approved):

nickvergessen:
NonExistingFolder needs the same fix? and did you also try "read" operation paths from admin_audit

juliushaertl:
Reads are not affected as the file already exists then, but added the same for folder creation on NonExistingFolder

nickvergessen (approved):

juliushaertl:
/backport 61108882b9405bbd3f69213e430201e37e32631b to stable30

PR #47601, opened 2024-08-29
[TaskProcessing] Use taskprocessing in TextProcessing and STT managers

julien-nc:
When
* scheduling a text processing task
* running a text processing task
* scheduling a STT task
* running a STT task

a task processing task actually runs. So these old APIs still benefit from the new task processing providers.

@marcelklehr Like we discussed, it might be good to remove the AI admin settings STT section and adjust the text processing one so it does not show preferences for `FreePromptTaskType`, `HeadlineTaskType`, `SummaryTaskType` and `TopicsTaskType`.

Also, @marcelklehr Should we only run the task processing task in those manager's "run" methods and get rid of the 2nd part which still runs the task like before in case the task processing task has failed?


AndyScherzinger (approved):

AndyScherzinger (approved):

bigcat88 (approved):
looks good

lib/private/SpeechToText/SpeechToTextManager.php
```diff
@@ -112,7 +116,28 @@ public function cancelScheduledFileTranscription(File $file, ?string $userId, st
...
+			$resultTask = $this->taskProcessingManager->runTask($taskProcessingTask);
+			if ($resultTask->getStatus() === Task::STATUS_SUCCESSFUL) {
+				$output = $resultTask->getOutput();
+				if (isset($output['output']) && is_string($output['output'])) {
+					return $output['output'];
+				}
+			}
+		} catch (Throwable $e) {
+			$this->logger->debug('Failed to run a Speech-to-text job from STTManager with TaskProcessing for file ' . $file->getId(), ['exception' => $e]);
+		}
```
marcelklehr, line 139: @julien-nc What do you think about adding a bit more logic here to

1. Avoid calling task processing if audio2text is not available
2. Properly pass on the failure if it is available and fails

instead of a generic fall back

lib/private/TaskProcessing/Manager.php
```diff
@@ -98,8 +96,34 @@ public function __construct(
...
+				$providers[$class] = $this->serverContainer->get($class);
+			} catch (\Throwable $e) {
+				$this->logger->error('Failed to load Text processing provider ' . $class, [
+					'exception' => $e,
+				]);
+			}
+		}
+
+		return $providers;
+	}
```
marcelklehr, line 123: Instead of duplicating the logic, what do you think about having a `getTextProcessingManager()` method that summons the manager to avoid the circular dependency?

PR #47605, opened 2024-08-29
feat: Use the blurhash in Files

artonge:
List view | Grid view
-- | --
Screenshot from 2024-08-29 14-30-30 | Screenshot from 2024-08-29 14-30-26


apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -14,16 +14,19 @@
...
-			alt=""
-			class="files-list__row-icon-preview"
-			:class="{'files-list__row-icon-preview--loaded': backgroundFailed === false}"
-			loading="lazy"
-			:src="previewUrl"
-			@error="onBackgroundError"
-			@load="backgroundFailed = false">
+		<!-- Decorative images, should not be aria documented -->
+		<span v-else-if="previewUrl" class="files-list__row-icon-preview-container">
+			<canvas v-if="hasBlurhash && (backgroundFailed === true || !backgroundLoaded)" ref="canvas" class="files-list__row-icon-blurhash" />
```
susnux, line 19: Canvas have weird semantics for screen readers (or the problem is there is no standard how to work with them).
So I would recommend:
1. Either add `role="img"` and an empty aria label
2. Or probably better: Add `aria-hidden="true"`

artonge: Then please: https://github.com/nextcloud/photos/pull/2628 :D

apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -107,6 +111,7 @@ export default Vue.extend({
 	data() {
 		return {
 			backgroundFailed: undefined as boolean | undefined,
+			backgroundLoaded: false as boolean,
```
susnux, line 114:
```suggestion
			backgroundLoaded: false,
```

Not needed as `boolean` is inferred from `false`

apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -206,24 +211,53 @@ export default Vue.extend({
...
+		drawBlurhash() {
+			const width = this.$refs.canvas.width
+			const height = this.$refs.canvas.height
+
+			const pixels = decode(this.source.attributes['metadata-blurhash'], width, height)
+
+			const ctx = this.$refs.canvas.getContext('2d')
+			const imageData = ctx.createImageData(width, height)
+			imageData.data.set(pixels)
+			ctx.putImageData(imageData, 0, 0)
```
susnux, line 260: There are some type issues, so I would recommend to first cast the canvas ref properly and also check for null on `getContext` (this is a valid return value):

```suggestion
		drawBlurhash() {
			const canvas = this.$refs.canvas as HTMLCanvasElement
			const width = canvas.width
			const height = canvas.height

			const pixels = decode(this.source.attributes['metadata-blurhash'], width, height)

			const ctx = canvas.getContext('2d')
			if (ctx === null) {
				logger.error('Cannot create context for blurhash canvas')
				return
			}
			const imageData = ctx.createImageData(width, height)
			imageData.data.set(pixels)
			ctx.putImageData(imageData, 0, 0)
```

susnux (commented):
Looks super nice! Just a few a11y and TS comments :)

apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -14,16 +14,19 @@
...
-			alt=""
-			class="files-list__row-icon-preview"
-			:class="{'files-list__row-icon-preview--loaded': backgroundFailed === false}"
-			loading="lazy"
-			:src="previewUrl"
-			@error="onBackgroundError"
-			@load="backgroundFailed = false">
+		<!-- Decorative images, should not be aria documented -->
+		<span v-else-if="previewUrl" class="files-list__row-icon-preview-container">
+			<canvas v-if="hasBlurhash && (backgroundFailed === true || !backgroundLoaded)" ref="canvas" class="files-list__row-icon-blurhash" aria-hidden="true"/>
```
susnux, line 19:
```suggestion
			<canvas v-if="hasBlurhash && (backgroundFailed === true || !backgroundLoaded)"
				ref="canvas"
				aria-hidden="true"
				class="files-list__row-icon-blurhash" />
```

Make ESLint happy

apps/files/src/components/FileEntry/FileEntryPreview.vue
```diff
@@ -14,16 +14,19 @@
...
+			<img v-if="backgroundFailed !== true"
```
susnux, line 20: Prevent that both are shown, or am I missing something here?

```suggestion
			<img v-else-if="backgroundFailed !== true"
```

artonge: No, we need both at the beginning, else the preview is never loaded by the ``. The canvas will then be hidden when the preview is loaded.

susnux (approved):

skjnldsv (approved):

PR #47613, opened 2024-08-29
Use app path dir instead of hardcoded path for theme

akhil1508:
## Summary

- https://github.com/nextcloud/server/commit/d0d9b398432d74145c0df686aed434450255b6b9 rolled  back https://github.com/nextcloud/server/commit/5e20c3e7b97598a8e632d9c5ccdef7de7813a4e8
  - This was a fix for https://github.com/nextcloud/server/issues/40897

-  So now the issue closed above is no longer working
- My fix is just to use the generic app path name instead of a hardcoded `/apps`

## Checklist

### DIFF
--- a/lib/private/Template/JSResourceLocator.php
+++ b/lib/private/Template/JSResourceLocator.php
@@ -42,6 +42,7 @@ class JSResourceLocator extends ResourceLocator {
 				// Handle symlinks
 				$appRoot = realpath($appRoot);
 			}
+			$appDirName = basename($appRoot);
 			// Get the app webroot
 			$appWebRoot = dirname($this->appManager->getAppWebPath($app));
 		} catch (AppPathNotFoundException $e) {
@@ -57,7 +58,7 @@ class JSResourceLocator extends ResourceLocator {
 			$found += $this->appendScriptIfExist($this->serverroot, $script);
 			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.$script);
 			$found += $this->appendScriptIfExist($appRoot, $script, $appWebRoot);
-			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.'apps/'.$script);
+			$found += $this->appendScriptIfExist($this->serverroot, $theme_dir.$appDirName.'/'.$script);
 
 			if ($found) {
 				return;


susnux:
I do not see how this is a fix for https://github.com/nextcloud/server/issues/40897

That issue was about apps installed in `apps-extra` or `apps_custom` or any other app directory.
But this PR is about the legacy (deprecated) theme handling, where all overwritten translations are located in `themes/{theme_name}/apps/{appid}/l10n/{lanuage}.js`.

I tested custom app paths and they work, e.g. `calendar` installed in `apps-foo` is loading `apps-foo/calendar/l10n/de.js` correctly.

But custom overwrites in `themes/___/apps-foo/` were never supported.

Or do I miss something here @akhil1508 ?

susnux (changes requested):
I do not see how this is a fix for https://github.com/nextcloud/server/issues/40897

That issue was about apps installed in `apps-extra` or `apps_custom` or any other app directory.
But this PR is about the legacy (deprecated) theme handling, where all overwritten translations are located in `themes/{theme_name}/apps/{appid}/l10n/{language}.js`.

I tested custom app paths and they work, e.g. `calendar` installed in `apps-foo` is loading `apps-foo/calendar/l10n/de.js` correctly.

But custom overwrites in `themes/___/apps-foo/` were never supported.

Or do I miss something here @akhil1508 ?

akhil1508:
> I do not see how this is a fix for https://github.com/nextcloud/server/issues/40897

@susnux https://github.com/nextcloud/server/issues/40897 is an issue for changing translations of custom apps via the theme. So this change allows that by delivering them from theme.

> But custom overwrites in themes/___/apps-foo/ were never supported.

Ah I did not know this, I was not aware that this is an unsupported feature.  So the correct way to do it with the legacy theme is to just add the "custom" apps into the `apps` folder in theme?

> But this PR is about the legacy (deprecated) theme handling, where all overwritten translations are located in themes/{theme_name}/apps/{appid}/l10n/{language}.js.

Could you point me to what the new way of overwriting translations is as this one is deprecated? Would be very helpful to me :)

susnux:
> My current theme structure is like `apps/ other_apps/ ...`, and with this PR's change the translations from `other_apps` are also taken into account. So if I want to avoid having this patch, I just move the folders in `other_apps/` into `apps/` ?

Yes thats the way to go:

- Real app: `apps-extra/my-super-app`
- Custom translation: `themes/my-theme/apps/my-super-app/l10n/...`

While you have multiple app roots, the theming overwrite are always in  `apps` for all app roots :) 

> Could you point me to what the new way of overwriting translations is(or could be) as this one is deprecated? Would be very helpful to me :)

There is none, this is one of the remaining discussions before we can remove legacy (pre NC20) themes.
But it would be super helpful if you can tell us your use case for overwriting translations, is it about "theming" or are there issues with the provided translations? Or even something different?
This could really help us to find a pleasing solution for this topic :smiley:

akhil1508:
@susnux Thanks for the inputs :))

>  But it would be super helpful if you can tell us your use case for overwriting translations, is it about "theming" or are there issues with the provided translations? Or even something different?

 It's mainly to customize the branding, look and feel:

- An example could be the [Memories](https://github.com/pulsejet/memories) app. We call it "Gallery" to be in line with the branding. So we need to overwrite its translations.
- Apart from translations, custom logo, favicons and some other assets to replace the inbuilt ones for a custom look (but now inline svgs are the current way to go so not a direct drop-in replacement for a lot of cases)
- In some very rare cases, script replacements to change functionality, look and feel

susnux:
Thank you very much for your feedback!

> * An example could be the [Memories](https://github.com/pulsejet/memories) app. We call it "Gallery" to be in line with the branding. So we need to overwrite its translations.

This is a pretty valid use case, I think we should add a proper non-legacy way to archive this behavior!

> * Apart from translations: custom logo, favicons and some other assets to replace the inbuilt ones for a custom look (but now inline svgs are the current way to go so not a direct drop-in replacement for a lot of cases)

Well but you can directly set a custom logo and favicon: [Admin theming](https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/theming.html)
But custom app icons probably make also sense, this is something I more often read.
For the other icons, yes we use inline SVGs mostly so no theming is possible here and probably in the future also the left over places will use inline SVGs.

> * In some very rare cases, script replacements to change functionality, look and feel

This will probably not work much longer, there are mostly non vanilla scripts left, as everything is moving to pre-bundled scripts. So I do not think we can keep this behavior. Instead we should consider proper APIs to allow the desired behavior if needed.

susnux:
I will close this PR as this is not a supported use-case please just use `themes/___/apps/...` instead :)

PR #47628, opened 2024-08-30
fix(UnifiedSearch): Restore normal dialog size

nfebe:
| Before | After |
| --------- | ------ |
| Screenshot from 2024-08-30 10-10-14 | Screenshot from 2024-08-30 10-07-34 |


sorbaugh (approved):

marcoambrosini (approved):

core/src/components/UnifiedSearch/UnifiedSearchModal.vue
```diff
@@ -724,7 +725,8 @@ export default defineComponent({
 	&__no-content {
 		display: flex;
 		align-items: center;
-		height: 100%;
+		margin-top: 0.5em;
```
susnux, line 728: What is that margin for? Should be already covered in the height, no?

susnux (approved):
Makes sense

core/src/components/UnifiedSearch/UnifiedSearchModal.vue
Fenn-CS: The borders are overlapping with elements above....

susnux: ok thank you :)

jancborchardt (approved):
Good stuff, thanks @Fenn-CS!

PR #47648, opened 2024-08-30
refactor(style): Remove repeated semicolon

Pytal:
## Summary

* refactor(style): Remove repeated semicolon

## Checklist

### DIFF
--- a/core/css/public.scss
+++ b/core/css/public.scss
@@ -70,7 +70,7 @@
 	/* public footer */
 	footer {
 		position: fixed;
-		bottom: var(--body-container-margin);;
+		bottom: var(--body-container-margin);
 		background-color: var(--color-main-background);
 		border-radius: var(--body-container-radius);
 		box-sizing: border-box;


susnux (changes requested):
The value is valid, `line-height` can just be a numeric factor.
This is essential for calculations of different elements with different font sizes, so this is just the relative line height.

See also https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_30.html#line-height

---

This value `default-line-height` can be used on any element regardless if the element uses `font-size: var(--default-font-size);` or `font-size: var(--font-size-small)`.
Instead if you need the line height of an element you do not need any variable, but simply can use `1lh` this is the CSS unit for one line height of that element.

core/css/public.scss
```diff
@@ -3,11 +3,11 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 #body-public {
-	--footer-height: calc(var(--default-line-height) + 2 * var(--default-grid-baseline));;
+	--footer-height: calc(var(--default-line-height) + 2 * var(--default-grid-baseline));
```
susnux, line 6: So instead you need to multiply the line height with the font size, or much easier:

```suggestion
	--footer-height: calc(1lh + 2 * var(--default-grid-baseline));
```

core/css/public.scss
```diff
@@ -3,11 +3,11 @@
...
 
 	&:has(.footer__legal-links),
 	&:has(.footer__simple-sign-up) {
-		--footer-height: calc(2 * var(--default-line-height) + 3 * var(--default-grid-baseline));;
+		--footer-height: calc(2 * var(--default-line-height) + 3 * var(--default-grid-baseline));
```
susnux, line 10:
```suggestion
		--footer-height: calc(2lh + 3 * var(--default-grid-baseline));
```

susnux (approved):

AndyScherzinger (approved):

skjnldsv (approved):

PR #46784, opened 2024-07-26
fix(styles): Make sure footer with legal links is shown correctly

susnux:
* Resolves: https://github.com/nextcloud/server/issues/36485
* Resolves: https://github.com/nextcloud/server/issues/34399
* Resolves: https://github.com/nextcloud/server/issues/34305

## Summary

Fix public pages footer height. If there are legal links or a signup link or both we need different heights of the footer.
Otherwise the footer content will be clipped and overlay the app.

Adjusted colors in the screen shots to make clear what is which element:

before | after
---|---
Screenshot 2024-07-26 at 15-01-20 Dateien - Nextcloud|Screenshot 2024-07-26 at 15-00-15 Dateien - Nextcloud
Screenshot 2024-07-26 at 15-01-01 Dateien - Nextcloud|Screenshot 2024-07-26 at 14-59-50 Dateien - Nextcloud

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [ ] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #47649, opened 2024-08-30
fix: Fix inaccessible content on public pages due to overlapping footer

### LINKED ISSUE #47426: [Bug]: Long footer may hide content information, e.g. on public upload folders

### Bug description

When opening a public link with a folder upload, the information about the uploaded files might not be visible. Please see the screencast – first window is current Nextcloud master, the other is NC 28.

[Screencast_20240822_155030.webm](https://github.com/user-attachments/assets/4e38772b-62f5-44ce-84bc-f0badb2e19fc)

### Steps to reproduce

0. Set a long slogan in the Theming settings
1. Share an upload-only folder via public link
2. Open that link on mobile, or in a browser size of a mobile browser
3. Upload a file

### Expected behavior

The list of uploaded files is clearly visible, at least it can be scrolled. 

### Installation method

None

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

Pytal:
## Resolves

- Fix #47426

Adds a non-invasive script to set the correct styling whenever the footer overlaps the main content so it is still accessible by scrolling.

Previous non-dynamic footer height was set in https://github.com/nextcloud/server/pull/46784

Before | After
--- | ---
image | image

## Checklist


core/css/public.scss
```diff
@@ -3,19 +3,7 @@
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
 #body-public {
```
susnux, line 5: I would keep the default value as an initial value before the script is initiated (can take some time depending on your internet connection and loading order) and to help understand that this value exists ;)

```suggestion
#body-public {
    // This value is updated by JavaScript to match the footers actual size
    --footer-height: calc(2lh + 2 * var(--default-grid-baseline));
```

apps/files_sharing/src/public-share.ts
susnux, line 1: I think this better belongs into `core` because this affects all templates rendered as `public`.

apps/files_sharing/lib/Listener/BeforePublicShareRenderedListener.php
susnux, line 1: As also written below, I think this is a general problem of the public layout. So we should move that into `core`.
But for the script loading this could simply be done in `core/Listener/BeforeTemplateRenderedListener.php`.
There we already do similar checks, we just need to check for `RENDER_AS_PUBLIC` and in that case add the script.

susnux (commented):
Works nicely, but I would move this into `core` as this is not files sharing related, but also affects other templates rendered as public (same e.g. for forms).

core/css/public.scss
```diff
@@ -51,6 +41,7 @@
 
 	#content {
 		min-height: var(--body-height, calc(100% - var(--footer-height)));
+		padding-block-end: var(--footer-height);
 	}
 
 	#app-content-vue {
```
susnux, line 47: (unrelated:)
This will be funny once we have Vue3 apps, then there is double the padding, as currently Vue2 will replace `#content` with `#app-content-vue`, while Vue3 does not replace the mount point but the inner HTML.

susnux (approved):
looks very clean!

skjnldsv (approved):

PR #47658, opened 2024-08-31
IUserConfig

ArtificialOwl:
- [x] replace `IConfig`'s preferences-related method
- [x] tests


lib/public/UserPreferences/ValueTypeDefinition.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace OCP\UserPreferences;
+
+/**
+ * Listing of value type definition for user preferences
+ *
+ * @see IUserPreferences
+ * @since 31.0.0
+ */
+enum ValueTypeDefinition: string {
```
susnux, line 17: Just curious, what is the use case for the definitions? Meaning in addition to the int backed ValueType enum 🤔

lib/private/UserPreferences.php
artonge, line 1: If possible, please split that file before it gets out of hand.

core/Migrations/Version31000Date20240814184402.php
artonge, line 1: Ideally, split the type and the flags to two distinct columns.

`type` with values `1, 2, 3, 4, ...`. That could also be some flag values, but then drop the `MIXED` type value
`flags` with value `1, 2, 4, 8, ...` or any combination of them.

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1675 @@
...
+			return;
+		}
+
+		if (($lazy ?? true) !== false) { // if lazy is null or true, we debug log
+			$this->logger->debug('The loading of lazy UserPreferences values have been requested', ['exception' => new \RuntimeException('ignorable exception')]);
+		}
+
+		$qb = $this->connection->getQueryBuilder();
+		$qb->from('preferences');
+		$qb->select('userid', 'appid', 'configkey', 'configvalue', 'type');
```
artonge, line 1523: For "extra" perf, don't select the `userid` column and use the `$userId` var.

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1675 @@
...
+		 */
+		if (!$inserted) {
+			$currType = $this->valueTypes[$userId][$app][$key] ?? 0;
+			if ($currType === 0) { // this might happen when switching lazy loading status
+				$this->loadPreferencesAll($userId);
+				$currType = $this->valueTypes[$userId][$app][$key] ?? 0;
+			}
+
+			/**
+			 * This should only happen during the upgrade process from 28 to 29.
```
come-nc, line 1042: A bit late for that.

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1675 @@
...
+			 * We only log a warning and set it to VALUE_MIXED.
+			 */
+			if ($currType === 0) {
+				$this->logger->warning('Value type is set to zero (0) in database. This is fine only during the upgrade process from 28 to 29.', ['app' => $app, 'key' => $key]);
```
come-nc, line 1046:
```suggestion
				$this->logger->warning('Value type is set to zero (0) in database. This is fine only during the upgrade process from 30 to 31.', ['app' => $app, 'key' => $key]);
```

come-nc (commented):
Stopped at line 1050 of UserPreferences, to be continued…

I dislike the sensitive flag thing, it should be a bool just like lazy so that all types var can be typed to the enum.

lib/public/UserPreferences/IUserPreferences.php
artonge, line 1: Maybe let's keep it private for one major version to have the flexibility to change the API down the road?

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1675 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC;
```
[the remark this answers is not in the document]
susnux: Maybe move to `\OC\UserPreferences\Manager` ?

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1777 @@
...
+		if ($valueType !== null) {
+//			$valueFlag = $valueType->value;
+//			if (ValueType::tryFrom($valueFlag) === null) {
+//				throw new InvalidArgumentException('Unknown value type');
+//			}
+		}
+	}
+
+	private function loadPreferencesAll(string $userId): void {
+		$this->loadPreferences($userId, null);
```
susnux, line 1606: I think this is a bit confusing, better be explicit:
```suggestion
		$this->loadPreferences($userId, false);
```

core/Migrations/Version31000Date20240814184402.php
```diff
@@ -0,0 +1,54 @@
...
+#[DropIndex(table: 'preferences', type: IndexType::INDEX, description: 'remove previous app/key index', notes: ['will be re-created to include \'indexed\' field'])]
+#[AddIndex(table: 'preferences', type: IndexType::INDEX, description: 'new index including user+lazy')]
+#[AddIndex(table: 'preferences', type: IndexType::INDEX, description: 'new index including app/key and indexed')]
+class Version31000Date20240814184402 extends SimpleMigrationStep {
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('preferences');
+		$table->addColumn('lazy', Types::SMALLINT, ['notnull' => true, 'default' => 0, 'length' => 1, 'unsigned' => true]);
```
susnux, line 38: I wonder why we are not using the doctrine boolean type? Then we do not need any magic conversation from boolean to 1 or 0 and can let doctrine do the conversation (any always have the proper PHP type).

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1777 @@
...
+				return (int)$value;
+			case ValueType::FLOAT:
+				return (float)$value;
+			case ValueType::BOOL:
+				return in_array(strtolower($value), ['1', 'true', 'yes', 'on']);
+			case ValueType::ARRAY:
+				try {
+					return json_decode($value, true, flags: JSON_THROW_ON_ERROR);
+				} catch (JsonException) {
+					// ignoreable
```
susnux, line 1748: Is this really ignoreable? Does this not mean we have encoded an invalid value meaning we have a bug that we should log to trace it?

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1777 @@
...
+ *
+ * ### What is lazy loading ?
+ * In order to avoid loading useless user preferences into memory for each request,
+ * only non-lazy values are now loaded.
+ *
+ * Once a value that is lazy is requested, all lazy values will be loaded.
+ *
+ * Similarly, some methods from this class are marked with a warning about ignoring
+ * lazy loading. Use them wisely and only on parts of the code that are called
+ * during specific requests or actions to avoid loading the lazy values all the time.
```
susnux, line 40: This section should be on the public interface as you refer to it in the public interface documentation.

susnux (commented):
Just to note this somewhere (feel free to ignore):
For me the name if a bit confusing (I know the DB table is called like that),
as we have `IConfig` and `IAppConfig` I would expect `IUserConfig` as the name.

lib/private/UserPreferences.php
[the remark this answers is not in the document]
ArtificialOwl: Moving everything into `\OC\Config` and `\OCP\Config`

ArtificialOwl:
> Just to note this somewhere (feel free to ignore): For me the name if a bit confusing (I know the DB table is called like that), as we have `IConfig` and `IAppConfig` I would expect `IUserConfig` as the name.

While I think grouping API might happens if we start today to create this API within `\[OC|OCP]\Config`/` but I am still fine with IUserPreferences

lib/private/UserPreferences.php
ArtificialOwl: So, I thought about it .. even started to split into 3: 
- `IUserPreferences` for basic getters/setters on user preferences.
- `IUserPreferencesExtra` that extends`IUserPreferences` and add all direct value manipulation on the cache like `is[Lazy|Sensitive|Indexed]`, `update[Lazy|Sensitive|Indexed]`, ...
- `IAdvancedUserPreferences` for search

but it makes everything bad. I will keep it like this (using a single file) until PHP can define a class from multiple files `:-]`

_Please note that there is a lot of single line method that request 10 lines of PHPDoc ..._

lib/private/UserPreferences.php
```diff
@@ -0,0 +1,1675 @@
...
+		}
+
+		$userIds = [];
+		$result = $qb->executeQuery();
+		$rows = $result->fetchAll();
+		foreach ($rows as $row) {
+			$userIds[] = $row['userid'];
+		}
+
+		return $userIds;
```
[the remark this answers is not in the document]
ArtificialOwl: added `Generator<string>`

lib/private/UserPreferences.php
come-nc: > it makes everything bad

**Reference needed**

artonge (approved):

susnux (commented):
Code is fine, but still want to confirm we want this name.

Reason: We use *app config* and *system config* and this was *config->getUserValue*.
So `UserConfig` might make more sense.

Yet all other places we have words for this refer it as **user settings** (documentation, occ commands etc), so maybe `UserSettings`. The only place we have that call it preferences is the DB table.

I am fine with anything regarding this if its agreed :)

lib/public/IConfig.php
```diff
@@ -245,7 +245,7 @@ public function deleteAppFromAllUsers($appName);
 	 * @param string $appName the app to get the user for
 	 * @param string $key the key to get the user for
 	 * @param string $value the value to get the user for
-	 * @return list<string> of user IDs
+	 * @return array<string> of user IDs
```
come-nc, line 248: Why the change? What are the keys now?

ArtificialOwl: switched back !

susnux (approved):
:partying_face:

kesselb:
Please send a follow-up and increase the version to trigger the migration.

st3iny:
> Please send a follow-up and increase the version to trigger the migration.

An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'type' in 'field list' /home/richard/src/nextcloud/master/lib/private/DB/Exceptions/DbalException.php at 54

lib/private/AllConfig.php
```diff
@@ -224,64 +227,33 @@ public function deleteAppValues($appName) {
 	 * @param string $key the key under which the value is being stored
 	 * @param string|float|int $value the value that you want to store
 	 * @param string $preCondition only update if the config value was previously the value passed as $preCondition
+	 *
 	 * @throws \OCP\PreConditionNotMetException if a precondition is specified and is not met
 	 * @throws \UnexpectedValueException when trying to store an unexpected value
+	 * @deprecated 31.0.0 - use {@see IUserConfig} directly
```
nickvergessen, line 233: https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/api.html#php-unstable-api
> Code from the OCP namespace must never mention anything coming from the NCU namespace. It can not require it as an argument, constant or return something from NCU.

So you can not deprecate this and neither the public interface.

lib/private/AllConfig.php
```diff
@@ -291,31 +263,39 @@ public function setUserValue($userId, $appName, $key, $value, $preCondition = nu
 	 * @param string $appName the appName that we stored the value under
 	 * @param string $key the key under which the value is being stored
 	 * @param mixed $default the default value to be returned if the value isn't set
+	 *
 	 * @return string
+	 * @deprecated 31.0.0 - use {@see IUserConfig} directly
```
nickvergessen, line 268: same for all methods

nickvergessen:
:boom: This PR breaks talks tests.
We set a user config and fetch it in another place of the code, but receive the default value instead:
https://github.com/nextcloud/spreed/actions/runs/11911817197/job/33195693787?pr=13750#step:15:2199

lib/private/Config/UserConfig.php
```diff
@@ -0,0 +1,1806 @@
...
+	 * @see getValueInt()
+	 * @see getValueFloat()
+	 * @see getValueBool()
+	 * @see getValueArray()
+	 */
+	public function getValueMixed(
+		string $userId,
+		string $app,
+		string $key,
+		string $default = '',
```
nickvergessen, line 537: Breaking change, before `mixed` was documented and no type was casted.
At least null is used by apps and was used even in server's own method (setUserValue)

lib/private/Config/UserConfig.php
```diff
@@ -0,0 +1,1806 @@
...
+		?bool $lazy = false,
+	): string {
```
nickvergessen, line 539: When the default value is "widen" again, the return type also needs to be

skjnldsv:
Please also document this

lib/private/AllConfig.php
```diff
@@ -224,64 +227,33 @@ public function deleteAppValues($appName) {
...
-					->andWhere($qb->expr()->eq('configkey', $qb->createNamedParameter($key)));
-				$qb->executeStatement();
-
-				$this->userCache[$userId][$appName][$key] = (string)$value;
-				return;
+		/** @var UserConfig $userPreferences */
+		$userPreferences = \OCP\Server::get(IUserConfig::class);
+		if ($preCondition !== null) {
+			try {
+				if ($userPreferences->getValueMixed($userId, $appName, $key) !== (string)$preCondition) {
```
nickvergessen, line 249: this is a behaviour change, previously the precondition did not fail, when the value is not set in the database.
- Fixed in https://github.com/nextcloud/server/pull/49386

lib/unstable/Config/Exceptions/IncorrectTypeException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace NCU\Config\Exceptions;
```
skjnldsv, line 9: Please create a documentation for this new Namespace and the way it should work (when is it going to become stable, when can devs use it... etc)

come-nc: https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/api.html#php-unstable-api

core/Migrations/Version31000Date20240814184402.php
```diff
@@ -0,0 +1,54 @@
...
+		$table->addColumn('type', Types::SMALLINT, ['notnull' => true, 'default' => 0, 'unsigned' => true]);
+		$table->addColumn('flags', Types::INTEGER, ['notnull' => true, 'default' => 0, 'unsigned' => true]);
+		$table->addColumn('indexed', Types::STRING, ['notnull' => false, 'default' => '', 'length' => 64]);
+
+		// removing this index from Version13000Date20170718121200
+		// $table->addIndex(['appid', 'configkey'], 'preferences_app_key');
+		if ($table->hasIndex('preferences_app_key')) {
+			$table->dropIndex('preferences_app_key');
```
ChristophWurst, line 46: for my dev instance running mariadb this DROP did not happen, so I now have redundant indexes:

image

The migration code seems correct, but something seems to prevent the DROP. @ArtificialOwl could you check the upgrade path?

nickvergessen: See https://github.com/nextcloud/server/issues/49638

The index is dropped... But next repair step run adds it again

PR #47662, opened 2024-08-31
fix(RichObjectStrings/Validator): Validate key value types of rich object parameters

provokateurin:
## Summary

A lot of apps set invalid rich object parameters for notifications. Most common are integer ids were the casting was forgotten.

Technically this is not a breaking change since the apps triggering this behavior are just broken, but we are way to late into the cycle to backport this to 30 and have everyone fix their apps, so I'm not doing that (let alone backport it to older versions).

Not sure if this needs to have some upgrade docs, please let me know.

## Checklist

### DIFF
--- a/apps/files_sharing/lib/Notification/Notifier.php
+++ b/apps/files_sharing/lib/Notification/Notifier.php
@@ -126,9 +126,9 @@ class Notifier implements INotifier {
 				[
 					'node' => [
 						'type' => 'file',
-						'id' => $node->getId(),
+						'id' => (string)$node->getId(),
 						'name' => $node->getName(),
-						'path' => $path,
+						'path' => (string)$path,
 					],
 				]
 			);
--- a/lib/private/Activity/Event.php
+++ b/lib/private/Activity/Event.php
@@ -259,7 +259,7 @@ class Event implements IEvent {
 	}
 
 	/**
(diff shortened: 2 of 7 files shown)


provokateurin:
One non-breaking backportable alternative would be to just cast the keys and values automatically and only log a debug messages. That doesn't force everyone to immediately fix their apps and we could still make it throw an error later when hopefully everyone has seen the debug logs.

lib/private/Notification/Notification.php
```diff
@@ -209,6 +209,14 @@ private function richToParsed(string $message, array $parameters): string {
 					throw new \InvalidArgumentException("Invalid rich object, {$requiredField} field is missing");
 				}
 			}
+			foreach ($parameter as $key => $value) {
```
nickvergessen, line 212: This should be done by the [IValidator](https://github.com/nextcloud/server/blob/master/lib/private/RichObjectStrings/Validator.php) instead and the setRich* methods should call it (so its also run when a parsed message was set by the provider already)

nickvergessen (changes requested):

provokateurin:
Adjusted all the other uses of the validator to have the right types.

come-nc (approved):

nickvergessen (approved):

PR #47670, opened 2024-09-01
fix: Properly calculate intermediate icon for touch icon generation

juliusknorr:
Fix touch icon rendering that is used by iOS devices if you add a webpage to the home screen. While the previous logic used to work it seems these days imagick does not have a resolution set for svg files, so we need to change calculation slightly so that our intermediate image has the right sizing for the final icon.

(The resolution naming seems confusing here, it is rather the density https://www.php.net/manual/en/imagick.setimageresolution.php#96182)

| Before | After |
|---|---|
| image | image |


susnux (approved):
code makes sense

AndyScherzinger (approved):

PR #47675, opened 2024-09-02
chore(db): Make the query more simple to read

solracsf:
## Checklist

### DIFF
--- a/lib/private/DB/Adapter.php
+++ b/lib/private/DB/Adapter.php
@@ -46,11 +46,10 @@ class Adapter {
 	/**
 	 * Create an exclusive read+write lock on a table
 	 *
-	 * @param string $tableName
 	 * @throws Exception
 	 * @since 9.1.0
 	 */
-	public function lockTable($tableName) {
+	public function lockTable(string $tableName): void {
 		$this->conn->beginTransaction();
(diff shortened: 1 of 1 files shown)


lib/private/DB/Adapter.php
```diff
@@ -46,11 +46,10 @@ public function fixupStatement($statement) {
 	/**
 	 * Create an exclusive read+write lock on a table
 	 *
-	 * @param string $tableName
 	 * @throws Exception
 	 * @since 9.1.0
 	 */
-	public function lockTable($tableName) {
+	public function lockTable(string $tableName) {
```
susnux, line 52: When adding types that we can also add the return type

```suggestion
	public function lockTable(string $tableName): void {
```

solracsf: That gave me a psalm error... 😞 See initail commit at https://github.com/nextcloud/server/pull/47675/commits/dfcf3132dda2c74ca96a3972ae8d78b4eba55b88

susnux: Ah yes, because all the implementing classes (AdapterSqlite etc) also would need that typing.

Altahrim (approved):
I would prefer to get rid of `insertIfNotExist` since it's deprecated since Nextcloud 15 :p 
This function is unsafe since column names are not escaped

skjnldsv (approved):

PR #47676, opened 2024-09-02
fix(OCP): Fix Image interface

provokateurin:
* Resolves: https://github.com/nextcloud/server/pull/47535#issuecomment-2324017343

## Summary

Using a private class in the public interface leads to no methods being resolved. Thus it needs to implement \OCP\IImage.
\OCP\IImage is also missing some methods that were relied on in OC_Image by other apps (e.g. Talk https://github.com/nextcloud/spreed/pull/13197) and needed to be added.

## Checklist

### DIFF
--- a/lib/public/Image.php
+++ b/lib/public/Image.php
@@ -14,5 +14,11 @@ namespace OCP;
  * This class provides functions to handle images
  * @since 6.0.0
  */
-class Image extends \OC\Image {
+class Image extends \OC\Image implements \OCP\IImage {
+	/**
+	 * @since 31.0.0
+	 */
+	public function __construct() {
+		parent::__construct();
+	}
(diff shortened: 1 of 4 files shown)


lib/public/Image.php
```diff
@@ -14,5 +14,8 @@
  * This class provides functions to handle images
  * @since 6.0.0
  */
-class Image extends \OC\Image {
+class Image extends \OC\Image implements \OCP\IImage {
+	public function __construct() {
```
come-nc, line 18:
```suggestion
	/**
	 * @since 31.0.0
	 */
	public function __construct() {
```

come-nc (approved):

artonge (approved):

PR #47685, opened 2024-09-02
fix: Move OC_API into \OC\ApiHelper in standard namespace

come-nc:
## Summary

It’s only used by ocs/v1.php

## Checklist


come-nc:
I now see that v1.php only uses this in error handling, and that it’s kind of duplicated code with V1Response/V2Response.
But I’m not sure how hard it would be to switch over to using VXResponse classes instead.

come-nc:
Integration test failure is related:

- [ ] In maintenance mode, even v1 returns 503 and not 200. (Not sure it makes sense but should be fixed for BC)

The taint errors I don’t know what to do with. The code path does not make sense, and in any case I’d expect json_encode output to always be safe to output. It’s only unsafe in a `<script>` tag and it’s not the case here.

lib/private/OCS/ApiHelper.php
```diff
@@ -0,0 +1,80 @@
...
+use OCP\IRequest;
+use OCP\Server;
+
+class ApiHelper {
+	/**
+	 * Respond to a call
+	 * @psalm-taint-escape html
+	 * @param int $httpStatusCode force the HTTP status code, only used for the special case of maintenance mode which return 503 even for v1
+	 */
+	public static function respond(int $statusCode, string $statusMessage, array $headers = [], ?int $httpStatusCode = null): void {
```
artonge, line 26: Maybe more explicit?

```suggestion
	public static function respond(int $statusCode, string $statusMessage, array $headers = [], ?int $overrideStatusCode = null): void {
```

artonge (approved):

nickvergessen (approved):
https://github.com/nextcloud/server/blob/af6de04e9e141466dc229e444ff3f146f4a34765/build/psalm-baseline.xml#L2863-L2870 will be removed automatically?

come-nc:
> https://github.com/nextcloud/server/blob/af6de04e9e141466dc229e444ff3f146f4a34765/build/psalm-baseline.xml#L2863-L2870
> 
> will be removed automatically?

Yes there’s a job for updating the baseline.

PR #3007, opened 2017-01-10
Remove non required db requests

Loki3000:
Remove SQL requests with empty UID for guests.

PR #47686, opened 2024-09-02
fix: Move login via email logic to local backend

### LINKED ISSUE #528: Account does not get blocked despite wrong password when using email address for login
### How to use GitHub

### Steps to reproduce
1. Under Administration/Security/Password Policy: set "Number of login attempts before the user account is blocked" to e.g. 3.
2. Logout and try to login as an exisiting user, using the user name (not the email address) and a wrong password. The user account gets blocked after above set number of tries.
3. After re-enabling the account, try to login using the email address (not the user name) and a wrong password. The user account does not get blocked after above set number of tries

### Expected behaviour
The account should get blocked, no matter if the user name or its email address is used.

### Actual behaviour
The account does not get blocked if the email address is used.

### Server configuration

**Nextcloud version:** 21.0.5

**Updated from an older Nextcloud/ownCloud or fresh install:** updated

### LINKED ISSUE #5221: Login with e-mail address should be moved to the DB user backend
### Steps to reproduce
1. Open Nextcloud web portal with the desired browser.
2. Enter as user login the e-mail address of a LDAP  user and the corresponding password.
3. Login is successfully.

### Expected behaviour
Login should not work, because when I check the user login within the LDAP settings of Nextcloud administration, I get the response that the user was not found. 

### Actual behaviour
When I check the user login within the Nextcloud administration (http://nextcloud-test.test.de/index.php/settings/admin/ldap), I get following response:

"User not found. Please check your login attributes and username. Effective filter (to copy-and-paste for command line validation):
(&(&(objectclass=inetOrgPerson)(mail=*)(!(mail=*@domain1.de))(!(mail=*@domain2.de)))(uid=groupware-test1@test.de))"

The userid of this user is "gtest1".

The following LDAP filter will be used to find the users:
(&(&(objectclass=inetOrgPerson)(mail=*)(!(mail=*@domain1.de))(!(mail=*@domain2.de)))(uid=%uid))

So there is an inconsistence between this check and the login mechanism. 

### Additional information
I already wrote an issue because of a problem with capital letters and Outlook (#5154), therefore I checked this here again: As soon as the e-mail address contains a capital letter, the login also fails. A login with capitals letters within the userid works fine.

### Server configuration

**Operating system**: Ubuntu 16.04.2 LTS

**Web server:** Apache/2.4.18

**Database:**  sqlite3, Version: 3.11.0

**PHP version:** 7.0.15

**Nextcloud version:** 12.0

**Updated from an older Nextcloud/ownCloud or fresh install:** First install was a 12 beta, which was upgraded to the current version last week.

**Where did you install Nextcloud from:**

**Signing status:**
<details>
<summary>Signing status</summary>
No errors have been found.
</details>

**List of activated apps:**
<details>
<summary>App list</summary>
Enabled:
  - activity: 2.5.2
  - bruteforcesettings: 1.0.2
  - calendar: 1.5.3
  - comments: 1.2.0
  - contacts: 1.5.3
  - dav: 1.3.0
  - federatedfilesharing: 1.2.0
  - files: 1.7.2
  - files_pdfviewer: 1.1.1
  - files_sharing: 1.4.0
  - files_texteditor: 2.4.1
  - files_trashbin: 1.2.0
  - files_versions: 1.5.0
  - files_videoplayer: 1.1.0
  - firstrunwizard: 2.1
  - gallery: 17.0.0
  - logreader: 2.0.0
  - lookup_server_connector: 1.0.0
  - nextcloud_announcements: 1.1
  - notifications: 2.0.0
  - oauth2: 1.0.5
  - password_policy: 1.2.2
  - provisioning_api: 1.2.0
  - serverinfo: 1.2.0
  - sharebymail: 1.2.0
  - survey_client: 1.0.0
  - systemtags: 1.2.0
  - theming: 1.3.0
  - twofactor_backupcodes: 1.1.1
  - updatenotification: 1.2.0
  - user_ldap: 1.2.1
  - workflowengine: 1.2.0
Disabled:
  - admin_audit
  - encryption
  - federation
  - files_external
  - spreed
  - user_external
</details>

**Nextcloud configuration:**
<details>
<summary>Nextcloud configuration</summary>
{
    "system": {
        "trusted_domains": [
            "nextcloud-test.test.de"
        ],
        "datadirectory": "\/nextcloud-data",
        "overwrite.cli.url": "http:\/\/nextcloud-test.test.de",
        "dbtype": "sqlite3",
        "version": "12.0.0.29",
        "dbname": "nextcloud",
        "dbhost": "127.0.0.1",
        "dbport": "",
        "dbtableprefix": "oc_",
        "instanceid": "och55ujpywqp",
        "logtimezone": "UTC",
        "installed": true,
        "ldapIgnoreNamingRules": false,
        "ldapProviderFactory": "\\OCA\\User_LDAP\\LDAPProviderFactory",
        "loglevel": 2,
        "maintenance": false,
        "updater.release.channel": "stable",
        "mail_from_address": "no-reply",
        "mail_smtpmode": "sendmail",
        "mail_smtpauthtype": "LOGIN",
        "mail_domain": "test.de",
        "theme": ""
    }
}
</details>

**Are you using external storage, if yes which one:** no

**Are you using encryption:** no

**Are you using an external user-backend, if yes which one:** LDAP

#### LDAP configuration 
<details>
<summary>LDAP config</summary>
+-------------------------------+-----------------------------------------------------------------------------------------------------+
| Configuration                 |                                                                                                     |
+-------------------------------+-----------------------------------------------------------------------------------------------------+
| hasMemberOfFilterSupport      | 0                                                                                                   |
| hasPagedResultSupport         |                                                                                                     |
| homeFolderNamingRule          |                                                                                                     |
| lastJpegPhotoLookup           | 0                                                                                                   |
| ldapAgentName                 | uid=readonly,ou=special-users,dc=test=de                                                       |
| ldapAgentPassword             | ***                                                                                                 |
| ldapAttributesForGroupSearch  |                                                                                                     |
| ldapAttributesForUserSearch   | sn;givenName;uid;mail                                                                               |
| ldapBackupHost                |                                                                                                     |
| ldapBackupPort                |                                                                                                     |
| ldapBase                      | dc=test,dc=de                                                                                     |
| ldapBaseGroups                | ou=groups,dc=test,dc=de                                                                           |
| ldapBaseUsers                 | ou=users,dc=test,dc=de                                                                            |
| ldapCacheTTL                  | 600                                                                                                 |
| ldapConfigurationActive       | 1                                                                                                   |
| ldapDefaultPPolicyDN          |                                                                                                     |
| ldapDynamicGroupMemberURL     |                                                                                                     |
| ldapEmailAttribute            | mail                                                                                                |
| ldapExperiencedAdmin          | 1                                                                                                   |
| ldapExpertUUIDGroupAttr       | cn                                                                                                  |
| ldapExpertUUIDUserAttr        | uid                                                                                                 |
| ldapExpertUsernameAttr        |                                                                                                     |
| ldapGidNumber                 | gidNumber                                                                                           |
| ldapGroupDisplayName          | cn                                                                                                  |
| ldapGroupFilter               | (&(|(objectclass=groupOfUniqueNames))(!(cn=studenten*)))                                            |
| ldapGroupFilterGroups         |                                                                                                     |
| ldapGroupFilterMode           | 0                                                                                                   |
| ldapGroupFilterObjectclass    |                                                                                                     |
| ldapGroupMemberAssocAttr      | uniqueMember                                                                                        |
| ldapHost                      | ldaps://idm.test.de                                                                          |
| ldapIgnoreNamingRules         |                                                                                                     |
| ldapLoginFilter               | (&(&(objectclass=inetOrgPerson)(mail=*)(!(mail=*@domain1.edu))(!(mail=*@domain2.de)))(uid=%uid)) |
| ldapLoginFilterAttributes     |                                                                                                     |
| ldapLoginFilterEmail          | 0                                                                                                   |
| ldapLoginFilterMode           | 0                                                                                                   |
| ldapLoginFilterUsername       | 1                                                                                                   |
| ldapNestedGroups              | 0                                                                                                   |
| ldapOverrideMainServer        |                                                                                                     |
| ldapPagingSize                | 500                                                                                                 |
| ldapPort                      | 636                                                                                                 |
| ldapQuotaAttribute            |                                                                                                     |
| ldapQuotaDefault              |                                                                                                     |
| ldapTLS                       | 0                                                                                                   |
| ldapUserDisplayName           | mail                                                                                                |
| ldapUserDisplayName2          |                                                                                                     |
| ldapUserFilter                | (&(objectclass=inetOrgPerson)(mail=*)(!(mail=*@domain1.edu))(!(mail=*@domain2.de)))              |
| ldapUserFilterGroups          |                                                                                                     |
| ldapUserFilterMode            | 0                                                                                                   |
| ldapUserFilterObjectclass     |                                                                                                     |
| ldapUuidGroupAttribute        | auto                                                                                                |
| ldapUuidUserAttribute         | auto                                                                                                |
| turnOffCertCheck              | 0                                                                                                   |
| turnOnPasswordChange          | 0                                                                                                   |
| useMemberOfToDetectMembership | 1                                                                                                   |
+-------------------------------+-----------------------------------------------------------------------------------------------------+
</details>

### Client configuration
**Browser:** FF, Chrome

**Operating system:** Ubuntu 16

susnux:
* Resolves: https://github.com/nextcloud/server/issues/5221
* Resolves: https://github.com/nextcloud/password_policy/issues/528

## Summary
Backends can decide which names they accept for login, e.g. with user_ldap you can configure arbitrary login fields. This was a hacky approach to allow login via email, so instead this is now only handled by the local user backend.

This also fixes some other related problems:
Other logic relys on `backend::get()` which was not handling email, so e.g. password policy could not block users logged in via email if they use out-dated passwords.
Similar for other integrations, as the user backend was not consistent with what is a login name and what not.

## Checklist


nickvergessen:
This could log out people from their instance unexpectedly when they log in with email instead of user id with LDAP. Similarly it breaks the https://github.com/nextcloud/user_external/ app users that logged in with email.

susnux:
> when they log in with email instead of user id with LDAP.

For LDAP we use the login attribute filter, so this should not be affected see the workaround in the removed login flow file.

> it breaks the https://github.com/nextcloud/user_external/ app users that logged in with email.

This could be true but in that case we should fix that app, no?
Because it would currently only allow email login *after* the first UID login.
Similar as it [was discussed here](https://github.com/nextcloud/server/issues/5221#issuecomment-345661127), though that discussion is quite old ("planned for NC14") I would still agree with the reasoning here.

But maybe we need to pause this for 32 instead?

nickvergessen:
> But maybe we need to pause this for 32 instead?

Sound like a good idea to merge next week after stable31 is branched off, and then leave the user_external app an issue what they need to do

lib/private/User/Database.php
```diff
@@ -65,15 +62,18 @@ public function __construct($eventDispatcher = null, $table = 'users') {
 		$this->cache = new CappedMemoryCache();
 		$this->table = $table;
 		$this->eventDispatcher = $eventDispatcher ?? \OCP\Server::get(IEventDispatcher::class);
+		$this->config = \OCP\Server::get(IConfig::class);
+		$this->dbConnection = null;
 	}
 
 	/**
 	 * FIXME: This function should not be required!
 	 */
```
come-nc, line 71:
```suggestion
```

lib/private/User/Database.php
```diff
@@ -368,46 +369,62 @@ public function checkPassword(string $loginName, string $password) {
...
-		if (!isset($this->cache[$uid])) {
-			//guests $uid could be NULL or ''
-			if ($uid === '') {
-				$this->cache[$uid] = false;
-				return true;
-			}
+		//guests $uid could be NULL or ''
+		if ($loginName === '') {
+			$this->cache[$loginName] = false;
+			return true;
```
come-nc, line 383: according to phpdoc return true here means user was found, but here it was not?

lib/private/User/Database.php
```diff
@@ -368,46 +369,62 @@ public function checkPassword(string $loginName, string $password) {
...
-	 * @param string $uid the username
+	 * @param string $loginName the username or email
 	 * @return boolean true if user was found, false otherwise
 	 */
-	private function loadUser($uid) {
-		$this->fixDI();
+	private function loadUser(string $loginName, bool $tryEmail = true): bool {
+		if (isset($this->cache[$loginName])) {
+			return true;
+		}
```
come-nc, line 378:
```suggestion
		if (isset($this->cache[$loginName])) {
			return $this->cache[$loginName] !== false;
		}
```

I think?

lib/private/User/Database.php
```diff
@@ -368,46 +369,62 @@ public function checkPassword(string $loginName, string $password) {
...
-					'uid' => (string)$row['uid'],
-					'displayname' => (string)$row['displayname'],
-					'password' => (string)$row['password'],
-				];
-			} else {
-				$this->cache[$uid] = false;
-				return false;
+		// Not found by UID so we try also for email, load uid for email.
+		if ($tryEmail) {
+			@[$emailUId] = $this->config->getUsersForUserValue('settings', 'email', mb_strtolower($loginName));
```
come-nc, line 411: no `@` operator please. Especially on a list construction like this it’s highly confusing.

lib/private/User/Database.php
```diff
@@ -368,46 +369,62 @@ public function checkPassword(string $loginName, string $password) {
...
+			// If found, try loading it
+			if ($emailUId !== null && $emailUId !== $loginName) {
+				$result = $this->loadUser($emailUId, false);
+				if ($result) {
+					// Also add cache result for the email
+					$this->cache[$loginName] = $this->cache[$emailUId];
+					// Set a reference to the uid cache entry for also delete email entry on user delete
+					$this->cache[$emailUId]['email'] = $loginName;
+				}
+				return $result;
```
come-nc, line 421: Why is a false result not cached here while it’s cached for the other situations?

lib/private/User/Database.php
susnux: I tried but this is unrelated to this change here.
Also it breaks a lot of tests so lets not do it in the PR. If you like mark this as a thing we can look into on the next hack week :notebook:

come-nc: I was only suggesting to remove the `FIXME` that you fixed here. Are you saying you will revert this part of the PR?

lib/private/User/Database.php
susnux: >  This operator is controversial. It may hide useful errors messages from function calls. A lot of programmers will avoid it even for hight cost. For assignments it is safe though.

Thought would be ok here as it seems to be safe on assignment.
The only reason for it is to silence the PHP notice on potentially null value, but that is safe, as if there is no first key in the array then `emailUId` is simply `null`.

lib/private/User/Database.php
susnux: Yes

lib/private/User/Database.php
susnux: I think this was a hack for guest users but seems not needed anymore.
Ref: https://github.com/nextcloud/server/pull/3007

```suggestion
			return false;
```

lib/private/User/Database.php
susnux: Related question from my side: Should we cache email in general? Because in theory email and uid could conflict.

Also cc @nickvergessen to not introduce potential security issues (yet I do not know how to use that, because uid is tried first so the real user would be returned and then the password (hopefully) does not match. So from my point of view in this case the user with email login will fail to login at all and must use their uid).

lib/private/User/Database.php
```diff
@@ -368,46 +369,62 @@ public function checkPassword(string $loginName, string $password) {
...
```
susnux, line 421:
```suggestion
				} else {
					$this->cache[$loginName] = false
				}
				return $result;
```

susnux, line 421:
```suggestion
				} else {
					$this->cache[$loginName] = false;
				}
				return $result;
```

susnux, line 421:
```suggestion
					$this->cache[$emailUId]['email'] = $loginName;
					return true;
				}
```

lib/private/User/Database.php
susnux: @come-nc put the `return` inside the `if`.
Now if the user is found it is cached and `true` is returned.
If the user is not found than that is cached by the second `loadUser` call (from the try email block). So both cases are cached.

lib/private/User/Database.php
susnux: So the alternative without `@` would be this?

```suggestion
			$emailUId = reset($this->config->getUsersForUserValue('settings', 'email', mb_strtolower($loginName)));
```

and then check for `$emailUId !== false` below

come-nc: Yes, reset sounds good.
Alternative is:
```suggestion
			[$emailUId] = [...$this->config->getUsersForUserValue('settings', 'email', mb_strtolower($loginName)), null];
```

Not pretty.

come-nc: case should be `emailUid`, not `emailUId`. But the name is confusing actually.
Can we simply use `$uid` here?

lib/private/User/Database.php
susnux: No I would leave that in if nothing speaks against it.

Not sure, ideally that function is really not required - meaning we could just DI it in the constructor.
But somehow in installation it seems that a User is created before a db connection could be established...

lib/private/User/Database.php
susnux: > case should be

was thinking like: email **u**ser **i**d

>  Can we simply use $uid here?

Agree that's better!

skjnldsv (approved):
Looks sane (aside from a few static-code-analysis)

susnux:
> aside from a few static-code-analysis

Really not sure why psalm reports that, the type is `string|null` but psalm thinks it is `null`.

lib/private/User/Database.php
```diff
@@ -448,7 +466,7 @@ public function userExists($uid) {
 	 */
 	public function getHome(string $uid) {
 		if ($this->userExists($uid)) {
-			return \OC::$server->getConfig()->getSystemValueString('datadirectory', \OC::$SERVERROOT . '/data') . '/' . $uid;
+			return \OCP\Server::get(IConfig::class)->getSystemValueString('datadirectory', \OC::$SERVERROOT . '/data') . '/' . $uid;
```
come-nc, line 469:
```suggestion
			return $this>config->getSystemValueString('datadirectory', \OC::$SERVERROOT . '/data') . '/' . $uid;
```
You added config to properties

come-nc (approved):

lib/private/User/Database.php
susnux:
```suggestion
			return $this>config->getSystemValueString('datadirectory', \OC::$SERVERROOT . '/data') . '/' . $uid;
```

PR #47709, opened 2024-09-03
fix: only keep major as server version

skjnldsv:
For the releases, will lower the amount of times we need to compile during releases, this is often more cumbersome than necessary


susnux (approved):
Should work, problem here is we do not do semantic versioning here, but currently this should not break anything.

blizzz (approved):

skjnldsv:
> /compile rebase /

For some reason Github is not showing me merged PR...

PR #47711, opened 2024-09-03
fix(OC_Helper): Use correct binary notation for calculating human file sizes

provokateurin:
## Summary

Follows proper IEC 60027-2 notation.

## Checklist


provokateurin:
Ok let's just stick to the wrong but established notation.

susnux:
@provokateurin yes it was decided not to confuse existing users by switch the naming schema: https://github.com/nextcloud/server/pull/40621

In general it is a bit difficult, most users are spoiled by Windows using the wrong notation. But then there is Mac which is even weirder.

PR #47770, opened 2024-09-05
fix(dav): Always respond custom error page on exceptions

artonge:
In order to match what we are doing with regular controlers.

### DIFF
--- a/apps/dav/lib/Files/ErrorPagePlugin.php
+++ b/apps/dav/lib/Files/ErrorPagePlugin.php
@@ -7,17 +7,22 @@
  */
 namespace OCA\DAV\Files;
 
-use OC\AppFramework\Http\Request;
 use OC_Template;
 use OCP\AppFramework\Http\ContentSecurityPolicy;
+use OCP\IConfig;
 use OCP\IRequest;
 use Sabre\DAV\Exception;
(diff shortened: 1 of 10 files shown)


apps/dav/lib/Files/ErrorPagePlugin.php
```diff
@@ -98,4 +94,14 @@ public function sendResponse() {
 		$this->server->sapi->sendResponse($this->server->httpResponse);
 		exit();
 	}
+
+	private function acceptHtml(): bool {
+		foreach (explode(',', $this->request->getHeader('Accept')) as $part) {
+			$subparts = explode(';', $part);
+			if (str_ends_with($subparts[0], '/html')) {
+				return true;
+			}
```
provokateurin, line 103: IMO checking for the full mimetype `text/html` is better as HTML only has this single mimetype that is used, but this should work as well :woman_shrugging:

provokateurin (approved):

apps/dav/lib/Files/ErrorPagePlugin.php
artonge: Let's keep it like that :)

AndyScherzinger (approved):

PR #47782, opened 2024-09-05
fix(l10n): Rephrasing some exceptions to be translated

### LINKED ISSUE #47773: chore(l10n): Bad english needs to be improved
https://github.com/nextcloud/server/blob/332b3efdf0df6bd489d83f58e8ffa677749d0774/lib/private/Share20/Manager.php#L133

Reported at Transifex.

Should "SharedWith" a variable?

@solracsf Please help and improve

solracsf:
* Resolves: https://github.com/nextcloud/server/issues/47773

## Checklist

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -130,34 +130,34 @@ class Manager implements IManager {
 		if ($share->getShareType() === IShare::TYPE_USER) {
 			// We expect a valid user as sharedWith for user shares
 			if (!$this->userManager->userExists($share->getSharedWith())) {
-				throw new \InvalidArgumentException($this->l->t('SharedWith is not a valid user'));
+				throw new \InvalidArgumentException($this->l->t('Share recipient is not a valid user'));
 			}
 		} elseif ($share->getShareType() === IShare::TYPE_GROUP) {
 			// We expect a valid group as sharedWith for group shares
 			if (!$this->groupManager->groupExists($share->getSharedWith())) {
-				throw new \InvalidArgumentException($this->l->t('SharedWith is not a valid group'));
+				throw new \InvalidArgumentException($this->l->t('Share recipient is not a valid group'));
 			}
 		} elseif ($share->getShareType() === IShare::TYPE_LINK) {
 			// No check for TYPE_EMAIL here as we have a recipient for them
 			if ($share->getSharedWith() !== null) {
-				throw new \InvalidArgumentException($this->l->t('SharedWith should be empty'));
+				throw new \InvalidArgumentException($this->l->t('Share recipient should be empty'));
 			}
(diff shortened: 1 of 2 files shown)


lib/private/Share20/Manager.php
```diff
@@ -130,34 +130,34 @@ protected function generalCreateChecks(IShare $share, bool $isUpdate = false) {
 		if ($share->getShareType() === IShare::TYPE_USER) {
 			// We expect a valid user as sharedWith for user shares
 			if (!$this->userManager->userExists($share->getSharedWith())) {
-				throw new \InvalidArgumentException($this->l->t('SharedWith is not a valid user'));
+				throw new \InvalidArgumentException($this->l->t('Share recipient is not a valid user'));
```
susnux, line 133: I still think this should have been never be translated.

susnux (approved):
New messages seem to be fine.

Still I think we should not translate those errors, as this are not user facing exception (no HintException). Meaning those are developers only and translating them makes debugging log files much harder.

solracsf:
In the other side, if the user understands the Exception, he may be able to solve the situation without opening an issue... A coin has 2 sides 🫣

provokateurin (approved):
I agree with @susnux, all non user facing error messages should not be translated. The fix is good though.

provokateurin:
> In the other side, if the user understands the Exception, he may be able to solve the situation without opening an issue... A coin has 2 sides 🫣

They can use a translator for that. The harder part is if we get a translated error message and have to figure out what the original message was. A translator might not give the exact original version and it is much harder to find the source (unless you got a stacktrace).

susnux:
> In the other side, if the user understands the Exception, he may be able to solve the situation without opening an issue... 

But the user will not see this, or if they see it they can not fix it because those errors happen only because of invalid API usage -> broken apps or server -> if the user can fix it they are developers and probably should also understand English in our code base. On the otherhand as @provokateurin mentioned for us this makes reading issues / debugging much harder as we need to guess which error message was meant.

rakekniven (approved):

joshtrichards:
Motivated by translation requirements or not (and whether for users or developers), I quite like the added clarity of these rephrasings. Good stuff!

susnux:
@solracsf you need to adjust the tests, they expect some strings that now have been changed

joshtrichards:
Tests are failing:

```
There were 2 failures:

1) Test\Share20\ManagerTest::testMoveShareUserNotRecipient
Failed asserting that exception message 'Invalid share recipient' contains 'Invalid recipient'.

2) Test\Share20\ManagerTest::testMoveShareGroupNotRecipient
Failed asserting that exception message 'Invalid share recipient' contains 'Invalid recipient'.

```

PR #47790, opened 2024-09-05
[stable29] fix(caldav): Do not load IMipPlugin before user auth and session is c…

SebastianKrupinski:
Back port for #45081

Do not load IMipPlugin before user is authenticated and user session is initialized


apps/dav/tests/unit/CalDAV/Schedule/IMipPluginTest.php
```diff
@@ -107,8 +111,16 @@ protected function setUp(): void {
 		$this->timeFactory->method('getTime')->willReturn(1496912528); // 2017-01-01
 
 		$this->config = $this->createMock(IConfig::class);
+		
+		$this->user = $this->createMock(IUser::class);
+		/*
+		$this->user->method('getUID');
+		$this->user->method('getDisplayName');
+		*/
```
miaulalala, line 119:
```suggestion
```

miaulalala (changes requested):
Please remove the commented out code, otherwise it looks good!

apps/dav/tests/unit/CalDAV/Schedule/IMipPluginTest.php
SebastianKrupinski: Done

miaulalala (approved):

Altahrim (approved):

PR #47792, opened 2024-09-05
chore(files): Use public API where possible

susnux:
## Summary

This is not fixing all issues in the helper, but at least where possible use the public methods.
For the rest I am not sure what the best solution would be, just ignore? Make the data access public (probably not)...

## Checklist

### DIFF
--- a/apps/files/lib/Helper.php
+++ b/apps/files/lib/Helper.php
@@ -119,16 +119,18 @@ class Helper {
 	public static function formatFileInfo(FileInfo $i) {
 		$entry = [];
 
-		$entry['id'] = $i['fileid'];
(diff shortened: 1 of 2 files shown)


provokateurin (approved):

provokateurin:
I think you could cast it to the internal type (for safety) and then access the getters from there? It would at least ensure you never call the getters on some other class.

come-nc (approved):
Yeah maybe wrap in an `if (instanceof)` the arrayaccess part, maybe with an else clause logging an error to be safe.
Or clean it up and add getters for this info (or find a way to access it using public API if it exists)

PR #47801, opened 2024-09-06
fix(TaskProcessing): increase error_message column length

kyteinsky:
## Summary
Increase the length of the error message that can be stored and truncate it before updating the db.
And make the errors more visible since the previous error message was masked by throwing from an exception.

## Checklist

### DIFF
--- a/core/Migrations/Version30000Date20240906095113.php
+++ b/core/Migrations/Version30000Date20240906095113.php
@@ -1,0 +1,40 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC\Core\Migrations;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\Attributes\ModifyColumn;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+#[ModifyColumn(table: 'taskprocessing_tasks', name: 'error_message', description: 'Increase column length to 4000 bytes to support longer error messages')]
+class Version30000Date20240906095113 extends SimpleMigrationStep {
+
+	/**
+	 * @param IOutput $output
+	 * @param Closure(): ISchemaWrapper $schemaClosure
+	 * @param array $options
+	 * @return null|ISchemaWrapper
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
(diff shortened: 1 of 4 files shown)


kyteinsky:
There needs to be a Version31000 migration for NC 31, yes?

marcelklehr:
> There needs to be a Version31000 migration for NC 31, yes?

I don't think so :thinking:

marcelklehr (approved):

marcelklehr:
Ah, this is for stable30?

marcelklehr:
Would perhaps be better to submit this PR against master and then backport?

bigcat88 (approved):

kyteinsky:
> Would perhaps be better to submit this PR against master and then backport?

Yup, now that a separate PR is not required.

core/Migrations/Version30000Date20240906095113.php
```diff
@@ -0,0 +1,40 @@
...
+	 */
+	public function changeSchema(IOutput $output, Closure $schemaClosure, array $options): ?ISchemaWrapper {
+		/** @var ISchemaWrapper $schema */
+		$schema = $schemaClosure();
+
+		$table = $schema->getTable('taskprocessing_tasks');
+		$column = $table->getColumn('error_message');
+
+		if ($column->getLength() < 2048) {
+			$column->setLength(2048);
```
marcelklehr, line 35: Can you make sure this supports utf8mb4? I vaguely remember a limit of 4096 bytes, which is 1000 chars under utf8mb4

kyteinsky:
oops, sorry, changed base branch :see_no_evil:

core/Migrations/Version30000Date20240906095113.php
marcelklehr: The substr is good, I think. I'm more worried about the column configuration being invalid when utf8mb4 is turned on or something

kyteinsky: let's change it then :v: 
better safe than sorry

provokateurin (changes requested):
All the composer changes need to be reverted (except for the autoload_*.php ones)

kyteinsky:
done, thanks Kate!

provokateurin (approved):

core/Migrations/Version30000Date20240906095113.php
```diff
@@ -0,0 +1,40 @@
...
+
+namespace OC\Core\Migrations;
+
+use Closure;
+use OCP\DB\ISchemaWrapper;
+use OCP\Migration\Attributes\ModifyColumn;
+use OCP\Migration\IOutput;
+use OCP\Migration\SimpleMigrationStep;
+
+#[ModifyColumn(table: 'taskprocessing_tasks', name: 'error_message', description: 'Increase column length to 4000 bytes to support longer error messages')]
```
susnux, line 18: :blue_heart:

susnux (approved):

PR #47808, opened 2024-09-06
chore(i18n): Spelling of ID

rakekniven:
Reported at Transifex

### DIFF
--- a/lib/private/Share20/Manager.php
+++ b/lib/private/Share20/Manager.php
@@ -762,7 +762,7 @@ class Manager implements IManager {
 		try {
 			$originalShare = $this->getShareById($share->getFullId());
 		} catch (\UnexpectedValueException $e) {
-			throw new \InvalidArgumentException($this->l->t('Share does not have a full id'));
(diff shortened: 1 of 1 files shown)


joshtrichards (approved):

susnux:
> why adding milestone NC31?

Because every PR against master is milestone 31 or later, of course a backport for stable30 is possible, then that backport PR will target milestone 30.0.x

lib/private/Share20/Manager.php
```diff
@@ -762,7 +762,7 @@ public function updateShare(IShare $share) {
 		try {
 			$originalShare = $this->getShareById($share->getFullId());
 		} catch (\UnexpectedValueException $e) {
-			throw new \InvalidArgumentException($this->l->t('Share does not have a full id'));
+			throw new \InvalidArgumentException($this->l->t('Share does not have a full ID'));
```
susnux, line 765: Yet another place that should never been translated, this will never reach users directly but only logs.

susnux (approved):
makes sense, though should have been never translated

susnux:
@rakekniven was not yet backported to 30, no backport atm

PR #47837, opened 2024-09-09
fix(files): Never return a null ETag in DAV

provokateurin:
## Summary

Introduced by https://github.com/nextcloud/server/pull/47731. Psalm should have already complained before, but never did. There was also no baseline update so idk why we only see it now.

## Checklist

### DIFF
--- a/lib/private/Files/Storage/DAV.php
+++ b/lib/private/Files/Storage/DAV.php
@@ -721,10 +721,9 @@ class DAV extends Common {
 		return $stat ? $stat['permissions'] : 0;
 	}
 
-	/** {@inheritdoc} */
 	public function getETag($path) {
 		$meta = $this->getMetaData($path);
-		return $meta ? $meta['etag'] : null;
+		return $meta ? $meta['etag'] : false;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


nickvergessen (approved):

artonge (approved):

lib/private/Files/Storage/DAV.php
```diff
@@ -724,7 +724,7 @@ public function getPermissions($path) {
 	/** {@inheritdoc} */
 	public function getETag($path) {
 		$meta = $this->getMetaData($path);
-		return $meta ? $meta['etag'] : null;
+		return $meta ? $meta['etag'] : parent::getETag($path);
```
come-nc, line 727: Why `parent::getETag($path)` and not `false`?
What are the side effects of this?
https://github.com/nextcloud/server/pull/47731 only turned the if into a ternary it did not introduce the `null` return.
The OCP Storage interface documents `string|false` as return type.

provokateurin: The common interface it inherits says it can only be string. I didn't look further but you're right it should be false then.

come-nc (approved):

PR #47841, opened 2024-09-09
fix(files_sharing): Open single-file-share by default

susnux:
## Summary

Bring back the behavior that single file shares open the viewer by default.
We can not use a redirect like internally, so we need to adjust the route in the front end by checking if it is a single-file-share and if yes request the files app to open it.

## Checklist


mejo- (approved):
Nice, tested and works well. Thanks for fixing this straight away :heart:

artonge (approved):

PR #47846, opened 2024-09-09
fix: provision api's status codes

hamza221:
## Summary
`addUser`calls `editUser` which creates inconsistencies in status codes 

## TODO
- [x] Get this approved 
- [x] Backport 
- [x] Update documentation https://github.com/nextcloud/documentation/pull/12249

## More info 
# OLD

## CODE:

# New user ( affected by  Edit user exceptions )

101: Invalid password value/argument 

102: User already exists | Sub-admin group does not exist 

103: Cannot create sub-admins for admin group 

104: Group does not exist 

105: Insufficient privileges for group

106: No group specified (required for sub-admins) 

107: Hint exceptions 

108: An email address is required, to send a password link to the user. 

110: Required email address was not provided

111: Could not create non-existing user ID

# Edit user:

101: Invalid argument 

102: Invalid Quota /password / language / locale / first day of the week 

103: No permission to edit a field / (password policy / Setting the password is not supported by the users backend) 

998: user Not found

## Documentation

# New user:

- 100 - successful
- 101 - invalid input data
- 102 - username already exists
- 103 - unknown error occurred whilst adding the user
- 104 - group does not exist
- 105 - insufficient privileges for group
- 106 - no group specified (required for subadmins)
- 107 - all errors that contain a hint - for example “Password is among the 1,000,000 most common ones. Please make it unique.” (this code was added in 12.0.6 & 13.0.1)
- 108 - password and email empty. Must set password or an email
- 109 - invitation email cannot be send

# Edit user:

- 100 - successful
- 101 - user not found
- 102 - invalid input data

# NEW

# New user ( affected by  Edit user exceptions )

101: Invalid password value/argument 

102: User already exists 

103: Cannot create sub-admins for admin group 

104: Group does not exist 

105: Insufficient privileges for group

106: No group specified (required for sub-admins) 

107: Hint exceptions 

108: An email address is required, to send a password link to the user. 

109: Sub-admin group does not exist 

110: Required email address was not provided

111: Could not create non-existing user ID

# Edit user:

101: Invalid argument /Invalid Quota /password / language / locale / first day of the week 

107: password policy (hint exception)

112:  Setting the password is not supported by the users backend

113: editing field not allowed/ field doesn’t exist

## Checklist

### DIFF
--- a/apps/provisioning_api/lib/Controller/UsersController.php
+++ b/apps/provisioning_api/lib/Controller/UsersController.php
@@ -488,7 +488,7 @@ class UsersController extends AUserData {
 				$group = $this->groupManager->get($groupid);
 				// Check if group exists
 				if ($group === null) {
-					throw new OCSException($this->l10n->t('Sub-admin group does not exist'), 102);
+					throw new OCSException($this->l10n->t('Sub-admin group does not exist'), 109);
 				}
 				// Check if trying to make subadmin of admin group
 				if ($group->getGID() === 'admin') {
@@ -508,7 +508,7 @@ class UsersController extends AUserData {
 		}
 		if ($password === '') {
 			if ($email === '') {
-				throw new OCSException($this->l10n->t('To send a password link to the user an email address is required.'), 108);
+				throw new OCSException($this->l10n->t('An email address is required, to send a password link to the user.'), 108);
 			}
 
 			$passwordEvent = new GenerateSecurePasswordEvent();
@@ -1011,7 +1011,7 @@ class UsersController extends AUserData {
 		}
 		// Check if permitted to edit this field
 		if (!in_array($key, $permittedFields)) {
-			throw new OCSException('', 103);
+			throw new OCSException('', 113);
 		}
 		// Process the edit
 		switch ($key) {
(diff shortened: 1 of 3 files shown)


ChristophWurst:
This is to align with https://docs.nextcloud.com/server/latest/admin_manual/configuration_user/instruction_set_for_users.html, right?

hamza221:
> This is to align with https://docs.nextcloud.com/server/latest/admin_manual/configuration_user/instruction_set_for_users.html, right?

The next step would be to update that too, once this gets approved, both code and documentation have inconsistencies

nickvergessen:
These number codes are so ...
I think if we touch/break the existing codes anyway we should migrate to something "readable"

Any specific reason you want to touch them? Or where is this originating from?

nickvergessen (approved):
I checked the Talk Android and iOS mobile clients as those are also using the API. But doesn't seem they are checking the exact error code, but are basically either OK or show whatever the message is.
So seems kind of Yolo in that regard and we should just make sure that code matches docs afterwards

nickvergessen:
But we have integration tests that ensure the numbers stay. So better to keep those and fix the docs?

ChristophWurst (approved):

PR #47860, opened 2024-09-10
fix: Only write once to template instead of create/copy

juliusknorr:
Avoid triggering two file operations (create/copy) and write the template content on creation directly to also trigger a NodeWrittenEvent


mejo- (approved):
Seems sensible.

marcelklehr (approved):
:tada: Thanks!

PR #47889, opened 2024-09-11
fix: `user:settings` command when user is not available

susnux:
## Summary

If `ignore-missing-user` all sub commands work, except listing all settings for a user like `occ user:settings --ignore-missing-user user core`.

## Checklist

### DIFF
--- a/core/Command/User/Setting.php
+++ b/core/Command/User/Setting.php
@@ -219,7 +219,7 @@ class Setting extends Base {
 		}
 	}
 
-	protected function getUserSettings($uid, $app) {
+	protected function getUserSettings(string $uid, string $app): array {
 		$settings = $this->config->getAllUserValues($uid);
 		if ($app !== '') {
 			if (isset($settings[$app])) {
@@ -230,7 +230,10 @@ class Setting extends Base {
 		}
 
 		$user = $this->userManager->get($uid);
-		$settings['settings']['display_name'] = $user->getDisplayName();
+		if ($user !== null) {
+			// Only add the display name if the user exists
+			$settings['settings']['display_name'] = $user->getDisplayName();
+		}
 
 		return $settings;
 	}


skjnldsv (approved):

core/Command/User/Setting.php
```diff
@@ -230,7 +231,10 @@ protected function getUserSettings($uid, $app) {
 		}
 
 		$user = $this->userManager->get($uid);
-		$settings['settings']['display_name'] = $user->getDisplayName();
+		if ($user !== null || $ignoreMissingUser === false) {
+			// Only add the display name if the user exists (or is expected to exist)
+			$settings['settings']['display_name'] = $user?->getDisplayName();
+		}
```
come-nc, line 237: But simply changing for `$user?->` would be enough, no?
All the rest of the PR is unnecessary to my understanding.

susnux: This will wrongly output:
```
- core
  - something: 1234
- settings
  - display_name:
```

while with the current changes it correctly only lists the existing settings:
This will wrongly output:
```
- core
  - something: 1234
```

come-nc: Then just using `if ($user !== null)` only would work? Why show the empty displayname when ignoreMissingUser is false?

susnux:
```suggestion
		if ($user !== null) {
			// Only add the display name if the user exists
			$settings['settings']['display_name'] = $user->getDisplayName();
		}
```

core/Command/User/Setting.php
```diff
@@ -213,13 +213,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
-			$settings = $this->getUserSettings($uid, $app);
+			$ignoreMissingUser = $input->getOption('ignore-missing-user');
+			$settings = $this->getUserSettings($uid, $app, $ignoreMissingUser);
 			$this->writeArrayInOutputFormat($input, $output, $settings);
 			return 0;
 		}
 	}
 
-	protected function getUserSettings($uid, $app) {
+	protected function getUserSettings(string $uid, string $app, bool $ignoreMissingUser = false) {
```
come-nc, line 223:
```suggestion
	protected function getUserSettings(string $uid, string $app): array {
```

The bool is unused now.

core/Command/User/Setting.php
```diff
@@ -213,13 +213,14 @@ protected function execute(InputInterface $input, OutputInterface $output): int
 				return 1;
 			}
 		} else {
-			$settings = $this->getUserSettings($uid, $app);
+			$ignoreMissingUser = $input->getOption('ignore-missing-user');
+			$settings = $this->getUserSettings($uid, $app, $ignoreMissingUser);
```
come-nc, line 217:
```suggestion
			$settings = $this->getUserSettings($uid, $app);
```

come-nc (approved):

PR #47896, opened 2024-09-11
fix: Make user removal more resilient

### LINKED ISSUE #47900: User data not deleted if `User::delete` failed
Current flow is:

1. Emit `BeforeUserDeletedEvent`
2. Delete user from backend
3. Delete core user data
4. Emit `UserDeletedEvent`
5. (listern on that event: Remove user files)

If anywhere between 2 and 5 and exception occurs, e.g. database connection is lost, the state is undefined.
This means the user is already removed from backend, so we have no associated user anymore, but some data might still be available.
It is not possible to recover from this state because:
1. The user is gone (we could check if there are remnants by checking user preferences for userIds not available on any registered backend)
2. Our API (especially the events to let apps delete their user data) require real users not only user ids

While this case is probably logged (unexpected exception), removing all user data is hard for admins as it is unclear which tables and files to remove.

susnux:
* Resolves: https://github.com/nextcloud/server/issues/47900

## Summary

Currently there is a problem if an exception is thrown in `User::delete`, because at that point the user is already removed from the backend, but not all data is deleted.

There is no way to recover from this state, as the user is gone no information is available anymore. This means the data is still available on the server but can not removed by any API anymore.

The solution here is to first set a flag and backup the user home, this can be used to recover failed user deletions in a way the delete can be re-tried.

### Discussion

I am not 100% sure this is the best way, but as out API needs real user instances to delete user data, I think we need to go with the "FailedUsersBackend".
I tested this with injecting various exceptions in random places in the `User::delete` method and then let the background job cleanup the state. In my test cases this worked as expected.

## Checklist


lib/private/User/FailedUsersBackend.php
```diff
@@ -0,0 +1,50 @@
...
+	public function getHome(string $uid): string|false {
+		return $this->config->getUserValue($uid, 'core', 'deleted.backup-home') ?: false;
+	}
+
+	public function getUsers($search = '', $limit = null, $offset = null)
+	{
+		return $this->config->getUsersForUserValue('core', 'deleted', 'true');
+	}
+
+}
```
susnux, line 50:
```suggestion
}

```

lib/composer/composer/autoload_static.php
```diff
@@ -32,6 +36,10 @@ class ComposerStaticInit749170dad3f5e7f9ca158f5a9f04f6a2
         array (
             0 => __DIR__ . '/../../..' . '/lib/public',
         ),
+        'Bamarni\\Composer\\Bin\\' => 
+        array (
+            0 => __DIR__ . '/..' . '/bamarni/composer-bin-plugin/src',
+        ),
```
susnux, line 42:
```suggestion
```

lib/composer/composer/autoload_static.php
```diff
@@ -17,6 +17,10 @@ class ComposerStaticInit749170dad3f5e7f9ca158f5a9f04f6a2
             'OC\\' => 3,
             'OCP\\' => 4,
         ),
+        'B' => 
+        array (
+            'Bamarni\\Composer\\Bin\\' => 21,
+        ),
```
susnux, line 23:
```suggestion
```

lib/composer/composer/autoload_psr4.php
```diff
@@ -9,5 +9,6 @@
     'OC\\Core\\' => array($baseDir . '/core'),
     'OC\\' => array($baseDir . '/lib/private'),
     'OCP\\' => array($baseDir . '/lib/public'),
+    'Bamarni\\Composer\\Bin\\' => array($vendorDir . '/bamarni/composer-bin-plugin/src'),
```
susnux, line 12:
```suggestion
```

lib/composer/composer/installed.php
```diff
@@ -3,21 +3,30 @@
...
         ),
+        'bamarni/composer-bin-plugin' => array(
+            'pretty_version' => '1.8.2',
+            'version' => '1.8.2.0',
+            'reference' => '92fd7b1e6e9cdae19b0d57369d8ad31a37b6a880',
+            'type' => 'composer-plugin',
+            'install_path' => __DIR__ . '/../bamarni/composer-bin-plugin',
+            'aliases' => array(),
+            'dev_requirement' => true,
+        ),
```
susnux, line 30:
```suggestion
```

lib/composer/composer/installed.php
```diff
@@ -3,17 +3,17 @@
...
         'aliases' => array(),
-        'dev' => false,
+        'dev' => true,
     ),
     'versions' => array(
         '__root__' => array(
             'pretty_version' => 'dev-master',
             'version' => 'dev-master',
-            'reference' => 'b99276fdfbac6b1ff243807b8b5b8161f0f67a24',
+            'reference' => '027fc052e78cb8401b3805c26a349c859ba7f275',
```
susnux, line 16:
```suggestion
            'reference' => 'b99276fdfbac6b1ff243807b8b5b8161f0f67a24',
```

lib/composer/composer/installed.php
```diff
@@ -3,11 +3,11 @@
         'name' => '__root__',
         'pretty_version' => 'dev-master',
         'version' => 'dev-master',
-        'reference' => 'b99276fdfbac6b1ff243807b8b5b8161f0f67a24',
+        'reference' => '027fc052e78cb8401b3805c26a349c859ba7f275',
```
susnux, line 6:
```suggestion
        'reference' => 'b99276fdfbac6b1ff243807b8b5b8161f0f67a24',
```

lib/composer/composer/installed.php
```diff
@@ -7,7 +7,7 @@
         'type' => 'library',
         'install_path' => __DIR__ . '/../../../',
         'aliases' => array(),
-        'dev' => false,
+        'dev' => true,
```
susnux, line 10:
```suggestion
        'dev' => false,
```

lib/composer/composer/installed.json
```diff
@@ -1,5 +1,68 @@
...
+                "type": "git",
+                "url": "https://github.com/bamarni/composer-bin-plugin.git",
+                "reference": "92fd7b1e6e9cdae19b0d57369d8ad31a37b6a880"
+            },
+            "dist": {
+                "type": "zip",
+                "url": "https://api.github.com/repos/bamarni/composer-bin-plugin/zipball/92fd7b1e6e9cdae19b0d57369d8ad31a37b6a880",
+                "reference": "92fd7b1e6e9cdae19b0d57369d8ad31a37b6a880",
+                "shasum": ""
+            },
```
susnux, line 17:
```suggestion
    "packages": [],
    "dev": false,
    "dev-package-names": []
```

lib/composer/composer/installed.json
```diff
@@ -2,4 +2,54 @@
...
+            "license": [
+                "MIT"
+            ],
+            "description": "No conflicts for your bin dependencies",
+            "keywords": [
+                "composer",
+                "conflict",
+                "dependency",
+                "executable",
+                "isolation",
```
susnux, line 41:
```suggestion
```

lib/composer/composer/installed.json
```diff
@@ -2,4 +2,17 @@
...
+                "issues": "https://github.com/bamarni/composer-bin-plugin/issues",
+                "source": "https://github.com/bamarni/composer-bin-plugin/tree/1.8.2"
+            },
+            "install-path": "../bamarni/composer-bin-plugin"
+        }
+    ],
+    "dev": true,
+    "dev-package-names": [
+        "bamarni/composer-bin-plugin"
+    ]
```
susnux, line 17:
```suggestion
```

lib/private/User/FailedUsersBackend.php
```diff
@@ -0,0 +1,50 @@
+<?php
+
+/**
+ * SPDX-FileCopyrightText: 2016-2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-FileCopyrightText: 2016 ownCloud, Inc.
+ * SPDX-License-Identifier: AGPL-3.0-only
```
susnux, line 6:
```suggestion
 * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
 * SPDX-License-Identifier: AGPL-3.0-or-later
```

susnux:
@blizzz I think you got a point here, we need to cover the case where `backend->deleteUser` threw an exception.
(if it returned we are always in a defined state)

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
```diff
@@ -0,0 +1,55 @@
...
+use OC\User\FailedUsersBackend;
+use OC\User\Manager;
+use OC\User\User;
+use OCP\AppFramework\Utility\ITimeFactory;
+use OCP\BackgroundJob\TimedJob;
+use OCP\EventDispatcher\IEventDispatcher;
+use OCP\IConfig;
+use Psr\Log\LoggerInterface;
+
+class CleanupDeletedUsers extends TimedJob {
```
blizzz, line 20: imo should only run in maintenance window. And it run far too often. This should be an edge case, not normal one. Once during maintenance window is fine.

lib/private/User/FailedUsersBackend.php
```diff
@@ -0,0 +1,46 @@
...
+use OCP\IConfig;
+use OCP\IUserBackend;
+use OCP\User\Backend\IGetHomeBackend;
+
+/**
+ * This is a "fake" backend for users that were deleted,
+ * but not properly removed from Nextcloud (e.g. an exception occurred).
+ * This backend is only needed because some APIs in user-deleted-events require a "real" user with backend.
+ */
+class FailedUsersBackend extends Backend implements IGetHomeBackend, IUserBackend {
```
blizzz, line 18: i think the naming gives a wrong and/or unclear impression.

```suggestion
class PartiallyDeletedUsersBackend extends Backend implements IGetHomeBackend, IUserBackend {
```

maybe?

blizzz (commented):
Not tested, but looking ok

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
```diff
@@ -0,0 +1,55 @@
...
+		$backend = new FailedUsersBackend($this->config);
+		$users = $backend->getUsers();
+
+		if (empty($users)) {
+			$this->logger->debug('No failed deleted users found.');
+			return;
+		}
+
+		foreach ($users as $userId) {
+			try {
```
susnux, line 41: Maybe check if this user still exists on any backend and if yes, then just fetch that user and re-try delete it.
That would cover the case where the `deleteUser` method of the backed failed due to an exception.

blizzz:
> @blizzz I think you got a point here, we need to cover the case where `backend->deleteUser` threw an exception. (if it returned we are always in a defined state)

that would be a different case, yes.

blizzz:
> > @blizzz I think you got a point here, we need to cover the case where `backend->deleteUser` threw an exception. (if it returned we are always in a defined state)
> 
> that would be a different case, yes.

… but then i am not sure this can be dealt with generically. maybe with a retry, but there might be other things have happened or not happened in the backend.

susnux:
> … but then i am not sure this can be dealt with generically. maybe with a retry, but there might be other things have happened or not happened in the backend.

I just noticed this makes not much sense:
If the user is still in any user backend -> remove the "deleted" flag. As we then still have access to it and can use the normal delete methods.
If anything backend related happened the backend should log that (and the admin can proceed with that information).

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
susnux: After think about it a bit more: If the user exists in any backend -> remove the flag and skip.

susnux: done like that

lib/private/User/FailedUsersBackend.php
susnux: sounds better :sweat_smile:

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
susnux: makes sense, adjusted the code

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
```diff
@@ -0,0 +1,63 @@
...
+			}
+
+			try {
+				$user = new User(
+					$userId,
+					$backend,
+					\OCP\Server::get(IEventDispatcher::class),
+					config: $this->config,
+				);
+				$user->delete();
```
artonge, line 56: Is the delete operation idempotent?

lib/private/User/User.php
```diff
@@ -249,45 +238,75 @@ public function delete() {
...
-					$this->dispatcher->dispatchTyped(new BeforeUserRemovedEvent($group, $this));
-					$group->removeUser($this);
-					$this->dispatcher->dispatchTyped(new UserRemovedEvent($group, $this));
-				}
+
+		// Set delete flag on the user - this is needed to ensure that the user data is removed if there happen any exception in the backend
+		// because we can not restore the user meaning we could not rollback to any stable state otherwise.
+		$this->config->setUserValue($this->uid, 'core', 'deleted', 'true');
+		// We also need to backup the home path as this can not be reconstructed later if the original backend uses custom home paths
+		$this->config->setUserValue($this->uid, 'core', 'deleted.backup-home', $this->getHome());
```
artonge, line 246: Maybe more explicit, I was initially expecting a boolean indicating whether the home has been backed up.

```suggestion
		$this->config->setUserValue($this->uid, 'core', 'deleted.home-path', $this->getHome());
```

lib/private/User/User.php
```diff
@@ -249,45 +238,75 @@ public function delete() {
...
+		try {
+			// We need to create a transaction to make sure we are in a defined state
+			// because if all user values are removed also the flag is gone, but if an exception happens (e.g. database lost connection on the set operation)
+			// exactly here we are in an undefined state as the data is still present but the user does not exist on the system anymore.
+			$database->beginTransaction();
+			// Remove all user settings
+			$this->config->deleteAllUserValues($this->uid);
+			// But again set flag that this user is about to be deleted
+			$this->config->setUserValue($this->uid, 'core', 'deleted', 'true');
+			$this->config->setUserValue($this->uid, 'core', 'deleted.backup-home', $this->getHome());
```
artonge, line 292: Already set above, no?

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
artonge: After reading through it, looks like it is :).

artonge (approved):

lib/private/User/User.php
```diff
@@ -249,45 +238,75 @@ public function delete() {
...
-				}
+
+		// Set delete flag on the user - this is needed to ensure that the user data is removed if there happen any exception in the backend
+		// because we can not restore the user meaning we could not rollback to any stable state otherwise.
+		$this->config->setUserValue($this->uid, 'core', 'deleted', 'true');
+		// We also need to backup the home path as this can not be reconstructed later if the original backend uses custom home paths
+		$this->config->setUserValue($this->uid, 'core', 'deleted.backup-home', $this->getHome());
+
+		// Try to delete the user on the backend
+		$result = $this->backend?->deleteUser($this->uid);
```
come-nc, line 249: In what situation is `$this->backend` null?

lib/private/User/User.php
come-nc: No it’s deleted by `deleteAllUserValues`

lib/private/User/User.php
```diff
@@ -249,45 +238,75 @@ public function delete() {
...
+			$this->config->setUserValue($this->uid, 'core', 'deleted', 'true');
+			$this->config->setUserValue($this->uid, 'core', 'deleted.backup-home', $this->getHome());
+			// Commit the transaction so we are in a defined state: either the preferences are removed or an exception occurred but the delete flag is still present
+			$database->commit();
+		} catch (\Throwable $e) {
+			$database->rollback();
+			throw $e;
+		}
+
+		if ($this->emitter) {
```
come-nc, line 300:
```suggestion
		if ($this->emitter !== null) {
```

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
```diff
@@ -0,0 +1,63 @@
...
+				continue;
+			}
+
+			try {
+				$user = new User(
+					$userId,
+					$backend,
+					\OCP\Server::get(IEventDispatcher::class),
+					config: $this->config,
+				);
```
come-nc, line 55: You need to set an emmiter or the hook will not fire and a lot of stuff is connected to it.

come-nc:
I like the idea, not sure I like the fact that it runs daily.

susnux:
> I like the idea, not sure I like the fact that it runs daily.

It simply skips if there is nothing to do, otherwise it only does things in the maintenance window.

lib/private/User/User.php
susnux: To be honest: No idea. The backend is not required in the constructor and there are some code places where the user is constructed without any backend.

So two ways:
1. Make this null safe as is now
2. Or assert the backend is set and fail if not.

lib/private/User/BackgroundJobs/CleanupDeletedUsers.php
susnux: sad, but true. fixed.

lib/private/User/User.php
come-nc: At the very least log the information that the backend is missing

come-nc: But if backend is null you should return false as user deletion did not happen.
So I would add a check for null backend, if yes log and return false.

come-nc (commented):
I still disagree that this should run daily.
It should be a repair-step. We have ton of other repair steps or occ commands to fix problems, non of them run daily.

lib/private/User/User.php
susnux: This makes sense I implemented as suggestest

susnux:
> I still disagree that this should run daily.
> It should be a repair-step. We have ton of other repair steps or occ commands to fix problems, non of them run daily.

But it is not expensive, yet it ensures that you are compliant with data protection laws as deleted users are really deleted.

come-nc (approved):
Ok, legal requirement is a good reason to run it daily.

provokateurin:
Please include https://github.com/nextcloud/server/pull/48610 in the backports @susnux

PR #47909, opened 2024-09-11
feat: add pagination compatibility to cardav backend

hamza221:
* Part of : https://github.com/nextcloud/contacts/issues/3942

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/CardDAV/AddressBook.php
+++ b/apps/dav/lib/CardDAV/AddressBook.php
@@ -157,7 +157,9 @@ class AddressBook extends \Sabre\CardDAV\AddressBook implements IShareable, IMov
 	}
 
 	public function getChildren() {
-		$objs = $this->carddavBackend->getCards($this->addressBookInfo['id']);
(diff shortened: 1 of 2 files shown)


ChristophWurst:
I'd prefer a cursor based pagination if possible

https://medium.com/@maryam-bit/offset-vs-cursor-based-pagination-choosing-the-best-approach-2e93702a118b

It's better for performance and more robust when the address book changes while the Contacts app is open

apps/dav/lib/CardDAV/AddressBook.php
```diff
@@ -157,7 +157,9 @@ public function getChild($name) {
 	}
 
 	public function getChildren() {
-		$objs = $this->carddavBackend->getCards($this->addressBookInfo['id']);
+		$offset = $_SERVER['HTTP_X_NEXTCLOUD_OFFSET'];
+		$limit = $_SERVER['HTTP_X_NEXTCLOUD_LIMIT'];
```
ChristophWurst, line 161: We should be able to use the standardized limit: https://datatracker.ietf.org/doc/html/rfc6352#section-8.6.1

PR #47914, opened 2024-09-12
fix(LDAP): check index before accessing it

blizzz:
## Summary

Fixes `Undefined array key 1 at /var/www/html/apps/user_ldap/lib/Access.php#320`.

## Checklist

### DIFF
--- a/apps/user_ldap/lib/Access.php
+++ b/apps/user_ldap/lib/Access.php
@@ -305,20 +305,19 @@ class Access extends LDAPUtility {
 	 * @return array If a range was detected with keys 'values', 'attributeName',
 	 *               'attributeFull' and 'rangeHigh', otherwise empty.
 	 */
-	public function extractRangeData($result, $attribute) {
+	public function extractRangeData(array $result, string $attribute): array {
 		$keys = array_keys($result);
 		foreach ($keys as $key) {
 			if ($key !== $attribute && str_starts_with((string)$key, $attribute)) {
 				$queryData = explode(';', (string)$key);
-				if (str_starts_with($queryData[1], 'range=')) {
+				if (isset($queryData[1]) && str_starts_with($queryData[1], 'range=')) {
 					$high = substr($queryData[1], 1 + strpos($queryData[1], '-'));
-					$data = [
+					return [
 						'values' => $result[$key],
 						'attributeName' => $queryData[0],
 						'attributeFull' => $key,
 						'rangeHigh' => $high,
 					];
-					return $data;
 				}
 			}
 		}


apps/user_ldap/lib/Access.php
```diff
@@ -305,20 +305,19 @@ public function extractAttributeValuesFromResult($result, $attribute) {
...
 	 *               'attributeFull' and 'rangeHigh', otherwise empty.
 	 */
-	public function extractRangeData($result, $attribute) {
+	public function extractRangeData(array $result, string $attribute): array {
 		$keys = array_keys($result);
 		foreach ($keys as $key) {
 			if ($key !== $attribute && str_starts_with((string)$key, $attribute)) {
 				$queryData = explode(';', (string)$key);
-				if (str_starts_with($queryData[1], 'range=')) {
+				if ($queryData && isset($queryData[1]) && str_starts_with($queryData[1], 'range=')) {
```
provokateurin, line 313:
```suggestion
				if ($queryData !== null && isset($queryData[1]) && str_starts_with($queryData[1], 'range=')) {
```

provokateurin (approved):

apps/user_ldap/lib/Access.php
provokateurin: Wait did you want to check that it is not empty?
I hate risky comparisons, you never know the intention :see_no_evil:

apps/user_ldap/lib/Access.php
```diff
@@ -305,20 +305,19 @@
...
```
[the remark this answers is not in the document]
provokateurin: @blizzz please take a look at this

apps/user_ldap/lib/Access.php
blizzz: I specifically check for index 1.

hweihwang (approved):

apps/user_ldap/lib/Access.php
provokateurin: So isset should be enough? Not sure what the first condition is supposed to check then. Psalm also complains about it because the condition is redundant.

apps/user_ldap/lib/Access.php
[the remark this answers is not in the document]
blizzz: Theoretically explode may return false, though it won't here. So could safe the first test. But if the result should be empty, there is nothing further to do either.

apps/user_ldap/lib/Access.php
blizzz: Psalm did not complain locally :shrug:

apps/user_ldap/lib/Access.php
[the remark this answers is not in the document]
blizzz: > Theoretically explode may return false, though it won't here. So could safe the first test. But if the result should be empty, there is nothing further to do either.

Oh, the `false` is from before 8.0 :bearded_person:  Then i remove the first check.

apps/user_ldap/lib/Access.php
provokateurin: Yeah locally warnings are not displayed. Only CI shows them (but they can be ignored as psalm will not fail on them).

PR #47920, opened 2024-09-12
feat: Reset route if neither the Viewer of the Sidebar is open

artonge:
When the viewer or the sidebar is opened, we add the fileid to the route.
When both of them are closed, we do not remove the fileid from the route.
This means that, upon reload, the sidebar will be opened even though it was closed previously.

This PR ensure that the fileid is removed from the route when both the Sidebar and the Viewer are closed.

https://github.com/user-attachments/assets/f363228c-d365-49e3-954c-0bbfaae19763


apps/files/src/components/FilesListVirtual.vue
```diff
@@ -251,12 +256,31 @@ export default defineComponent({
 			}
 		},
 
+		handleSideBarCloseEvent() {
+			if (OCA.Viewer.file === '') {
```
susnux, line 260: It feels wrong having access to `OCA.Viewer` here, is is not enough the check if `openFile` is set?
That would make it independent of the Viewer in this place.

Meaning the viewer should reset `openfile` on close.
So on sidebar close there are only two states:
1. `openfile` is set: do not remove the fileid
2. `openfile` is not set: also remove the fileid

artonge: Makes sense as we are also relying on `openfile` for the other way.

susnux (approved):
makes sense (ideally add a cypress tests), only thing I am not sure about if this really should be inside the virtual file list. But code makes sense and seems to work fine.

apps/files/src/components/FilesListVirtual.vue
```diff
@@ -83,6 +83,7 @@ import VirtualList from './VirtualList.vue'
 import logger from '../logger.ts'
 import FilesListTableHeaderActions from './FilesListTableHeaderActions.vue'
 import FileListFilters from './FileListFilters.vue'
+import { subscribe, unsubscribe } from '@nextcloud/event-bus'
```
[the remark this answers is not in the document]
artonge: :people_hugging:

skjnldsv (approved):

skjnldsv (changes requested):
Wait for it

skjnldsv (approved):

skjnldsv:
Backports ?

PR #47927, opened 2024-09-12
Migrate away from OC_App to IAppManager

come-nc:
Follow-up of https://github.com/nextcloud/server/pull/44025

## Summary

Migrate most calls of `OC_App::getAppPath`, `OC_App::cleanAppId`, `OC_App::getAllApps` to calls on the AppManager instance. The later two were added to the OCP interface for the occasion.

## Checklist

### DIFF
--- a/lib/private/legacy/OC_App.php
+++ b/lib/private/legacy/OC_App.php
@@ -45,8 +45,7 @@ class OC_App {
 	 * @psalm-taint-escape html
 	 * @psalm-taint-escape has_quotes
 	 *
-	 * @param string $app AppId that needs to be cleaned
-	 * @return string
+	 * @deprecated 31.0.0 use IAppManager::cleanAppId
 	 */
 	public static function cleanAppId(string $app): string {
 		return str_replace(['<', '>', '"', "'", '\0', '/', '\\', '..'], '', $app);
--- a/lib/private/legacy/OC_App.php
+++ b/lib/private/legacy/OC_App.php
@@ -468,30 +468,10 @@ class OC_App {
 	 * get a list of all apps in the apps folder
 	 *
 	 * @return string[] an array of app names (string IDs)
-	 * @todo: change the name of this method to getInstalledApps, which is more accurate
+	 * @deprecated 31.0.0 Use IAppManager::getAllAppsInAppsFolders instead
 	 */
 	public static function getAllApps(): array {
-		$apps = [];
-
-		foreach (OC::$APPSROOTS as $apps_dir) {
-			if (!is_readable($apps_dir['path'])) {
-				\OCP\Server::get(LoggerInterface::class)->warning('unable to read app folder : ' . $apps_dir['path'], ['app' => 'core']);
-				continue;
-			}
-			$dh = opendir($apps_dir['path']);
-
-			if (is_resource($dh)) {
-				while (($file = readdir($dh)) !== false) {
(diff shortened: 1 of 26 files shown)


lib/public/App/IAppManager.php
```diff
@@ -282,4 +280,24 @@ public function setDefaultApps(array $defaultApps): void;
...
+
+	/**
+	 * Clean the appId from forbidden characters
+	 *
+	 * @psalm-taint-escape file
+	 * @psalm-taint-escape include
+	 * @psalm-taint-escape html
+	 * @psalm-taint-escape has_quotes
+	 *
+	 * @since 30.0.0
```
come-nc, line 292:
```suggestion
	 * @since 31.0.0
```

tests/lib/App/AppManagerTest.php
provokateurin, line 1: Is this semantic change unavoidable?
To me it looks like it could end up affecting some places where only a null check is performed (and not also an empty string check).

provokateurin (approved):
Nice cleanup!

lib/private/legacy/OC_App.php
```diff
@@ -469,30 +468,10 @@ public static function getAlternativeLogIns(): array {
...
-			}
-		}
-
-		$apps = array_unique($apps);
-
-		return $apps;
+		return \OCP\Server::get(IAppManager::class)->getAllAppsInAppsFolders();
 	}
 
 	/**
```
susnux, line 477: Can we also deprecate this in favor of `\OCP\Support\Subscription\IRegistry::delegateGetSupportedApps` ?

tests/lib/App/AppManagerTest.php
come-nc: I only added strong typing to the method matching the previous phpdoc typing.
But the mock was returning null, which never happens in the real methods, and that was crashing as the method is not allowed to return null anymore.

susnux: But it was never documented to return `null` at all

lib/private/legacy/OC_App.php
nickvergessen: I would do the opposite :see_no_evil:

tests/lib/App/AppManagerTest.php
come-nc: > But it was never documented to return `null` at all

Yes, and it never did, expect in the tests because it’s mocked.

lib/private/legacy/OC_App.php
come-nc: Why the opposite?

@susnux yeah we should. But I won’t do it in this PR to keep it mergeable. This kind of cleanup needs to be merged quickly or it conflicts fast.

nickvergessen: > Why the opposite?

From my perspective the subscription methods are "internal" and are used by (other) public components.
All the delegate*() functions are only called from "lib/private/ implementing public interfaces", settings app, support app or update checks.

tests/lib/App/AppManagerTest.php
provokateurin: Ah ok :+1:

lib/private/legacy/OC_App.php
come-nc: Maybe getSupportedApps can be moved to the AppManager then.

In the mean time, would one of you two approve this here PR? :-P

susnux (approved):
:broom:

Issue #39278, opened 2023-07-10
Improving the UI/UX for Theming/Appearance/Accessibility management


joshtrichards:
Also know as: Global Theming, Appearance and accessibility, and User Theming

The Theming versus the User Appearance/Accessibility + User Theming stuff is a bit confusing. I mean in terms of:

- nomenclature usage
- organization and presentation in the user interface (in both Administration and Personal contexts)
- naming in the global config <-- this last one is likely due to legacy and mostly a documentation issue IMO

In the UI we have...

...the **Administration->"Theming"**: (`/settings/admin/theming`) page:

- Name (Organization?)
- Web link
- Slogan
- Color (Background color?)
- Logo
- Background and login image
- Legal notice link
- Privacy notice link
- Header logo
- Favicon
- Toggle: Disable user theming?
  - *"Although you can select and customize your instance, users can change their background and colors. If you want to enforce your customization, you can toggle this on."*

I'd say the Admin page is actually pretty solid, with only minor potential tweaks (more on that in my proposal list).

Then we have also in the UI...

...the **Personal->"Appearance and accessibility"** (`/settings/user/theming`) page:

- While unstated, this is effectively **"Appearance, accessibility, and user theming"** (all combined) 
- Color/Contrast Modes: System default, Light, Dark, High contrast, Dark with high contrast
- Toggle: Dyslexia font
- Toggle: Keyboard shortcuts
- Global theming overrides (aka: "User theming"): Background image and/or background color (if user theming is permitted by the admin)

I have some ideas to make the Personal page clearer (more on that in my proposal list).

Then in the system `config.php` we have:

- `theme` <-- I believe this is legacy (i.e. only applicable if not using the theming app)
- `enforce_theme` <-- Essentially legacy as it's no longer necessary with theming app having its own equivalent parameter

I propose we:

- Strike the word "theme" from the various "Color/Contrast Modes" included under **"Appearance and accessibility" (`/settings/user/theming`)"** page since "theme" is a misnomer there
- Add a heading above the list of "Color/Contrast Modes" that says... "Color/Contrast Modes" since that's what they really are and since they're entirely independent (from the user's perspective) from the actual admin configured theme
- Add the heading "User Theming" just above the "Background" section on the *"Appearance and accessibility" (`/settings/user/theming`)"** page to clarify that's where the user adjustable (if permitted) theme overrides are
[...]

PR #47928, opened 2024-09-12
fix(a11y): Update OpenDyslexic from 0.920 to 2.001

nickvergessen:
- Ref https://github.com/nextcloud/server/issues/39278

## Summary

Normal font | OpenDyslexic v0.920 | OpenDyslexic v2.001
---|---|---
Bildschirmfoto vom 2024-09-12 16-39-55 | Bildschirmfoto vom 2024-09-12 16-39-20 | Bildschirmfoto vom 2024-09-12 16-39-38

## Checklist


apps/theming/lib/Themes/DyslexiaFont.php
```diff
@@ -43,30 +43,22 @@ public function getCSSVariables(): array {
 	}
 
 	public function getCustomCss(): string {
-		$fontPathWoff = $this->urlGenerator->linkTo('theming', 'fonts/OpenDyslexic-Regular.woff');
```
nickvergessen, line 46: otf is supported by all supported browsers (IE was the last one), so we can drop the others

jancborchardt (approved):
Wow, amazing improvement! FYI @xBelladonna if you want to check it out. :)

apps/theming/lib/Themes/DyslexiaFont.php
solracsf: Beside browser support, WOFF is generaly more lightweight too.

image

nickvergessen: But WOFF is not provided by the creator and I don't know what happens with licenses when using a convertor of any kind `¯\_(ツ)_/¯` The files are ~41KB and ~42KB. I really don't care too much about such a little size to risk license issues

nickvergessen: That being said, the old font file was ~250KB and the WOFF was ~120KB there is made a bit more sense

skjnldsv (approved):

skjnldsv:
I think this should be backported

PR #47933, opened 2024-09-13
fix(config): Throw PreconditionException always when it didn't match

nickvergessen:
## Summary

Explanation | Screenshot
---|---
Previously even when the precondition did not match, the call "passed" when the after value was the expected one. This however can lead to race conditions, duplicate code executions and other things. | grafik

## Checklist

### DIFF
--- a/lib/private/AllConfig.php
+++ b/lib/private/AllConfig.php
@@ -242,10 +242,10 @@ class AllConfig implements IConfig {
 		$prevValue = $this->getUserValue($userId, $appName, $key, null);
 
 		if ($prevValue !== null) {
-			if ($prevValue === (string)$value) {
-				return;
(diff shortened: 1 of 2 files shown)


artonge (approved):

come-nc (approved):
Does this make any sense without a transaction or a lock around it anyway? What prevents the DB value to be changed by another process between precondition check and value update?

nickvergessen:
> What prevents the DB value to be changed by another process between precondition check and value update?

There is another check inside that takes care of it (no changed row while the precondition is in the WHERE clause): https://github.com/nextcloud/server/blob/edaa4d1432dcbc705daf4a0f5a5cd0c967b7e52e/lib/private/DB/Connection.php#L620-L622

PR #47947, opened 2024-09-13
feat: Make `CheckServerResponseTrait` public and provide as `OCP\SetupCheck\CheckServerResponseTrait`

susnux:
## Summary

This trait is used by other apps for creating setup checks, so we should provide it instead apps using private API.
Note that I removed the `runHead` as this is just sugar but having it in OCP could make things in the future harder...

## Checklist

### DIFF
--- a/apps/dav/lib/SetupChecks/WebdavEndpoint.php
+++ b/apps/dav/lib/SetupChecks/WebdavEndpoint.php
@@ -9,11 +9,11 @@ declare(strict_types=1);
 
 namespace OCA\DAV\SetupChecks;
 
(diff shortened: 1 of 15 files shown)


come-nc (approved):

susnux:
BTW alternative I considered: Making this an abstract class to make class properties more clear.
But huge con: You can only inherit one, so that would not be extensible with others in the future.

I think a Trait is best fit here.

nickvergessen (approved):
- But please lets merge https://github.com/nextcloud/server/pull/48063 first as it needs a backport to 30

nickvergessen:
> :boom: The autoloaders are not up to date

PR #47959, opened 2024-09-14
chore(files): Provide old node on moved

Pytal:
## Summary

Provides the old node in the event payload instead of just the source

## Requires

- https://github.com/nextcloud-libraries/nextcloud-files/pull/1077

## Checklist


susnux:
Would be a breaking change, no?

Pytal:
> Would be a breaking change, no?

I suppose if we consider events to be versioned then yes, in which case this can just be for >=31

Pytal:
> I'm a bit against this.
> 
> A node isn't supposed to be static. It can change but the object it represents is still the same node. If we start cloning nodes creating versions, we lose the purpose of it.
> 
> Why was this needed in the first place?

In a call you suggested providing the old node instead of just the old source as it is now

We can close this if you want?

PR #47966, opened 2024-09-14
Revert "[stable30] fix(files_external): broken credentials dialog"

skjnldsv:
Merged during 30.0.0 freeze


nickvergessen:
Merged after the final, so should be fine.
And I'm not aware of a rebuild

PR #47984, opened 2024-09-14
build: Add script to fix strict-types

provokateurin:
## Summary

Only a test to see what happens.

Anna: How much do you want to break?
Kate: Yes!

## Checklist

Signed-off-by: provokateurin <kate@provokateurin.de>


nickvergessen:
Had a recent crossing:
https://github.com/nextcloud/files_accesscontrol/issues/591
So at least should not be backported, so we have enough time to notice issues. :see_no_evil: 
But corner-cases are the tripping point with those

provokateurin:
Exactly what we thought as well, so we're doing individual fixes and backporting those and then we can hopefully turn this on later once everything we found is fixed.

provokateurin:
Rector can do this as well

PR #47999, opened 2024-09-15
fix(ci): Use latest buildjet version

SystemKeeper:
## Summary

@susnux @skjnldsv We need to update the buildjet action, 25s now on restore, see https://github.com/nextcloud/server/actions/runs/10869657438/job/30161269276#step:2:12

We are suffering from a bug in node that keeps node running when there are unfullfilled promises, ref
* https://github.com/actions/cache/issues/1442
* https://github.com/actions/cache/blob/704facf57e6136b1bc63b828d79edcd491f0ee84/src/restoreImpl.ts#L102-L109

Seems that version we are using does not include the upstream fix. We can fix it to the latest hash or just use v3, as you like.

## Checklist

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -80,7 +80,7 @@ jobs:
         run: npm run cypress:version
 
       - name: Save context
-        uses: buildjet/cache/save@v3
+        uses: buildjet/cache/save@v4
         with:
           key: cypress-context-${{ github.run_id }}
           path: ./
@@ -103,7 +103,7 @@ jobs:
 
     steps:
       - name: Restore context
-        uses: buildjet/cache/restore@v3
+        uses: buildjet/cache/restore@v4
         with:
           fail-on-cache-miss: true
           key: cypress-context-${{ github.run_id }}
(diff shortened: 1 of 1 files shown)


skjnldsv:
They even have a v4

.github/workflows/cypress.yml
```diff
@@ -103,7 +103,7 @@ jobs:
 
     steps:
       - name: Restore context
-        uses: buildjet/cache/restore@e376f15c6ec6dc595375c78633174c7e5f92dc0e # v3
+        uses: buildjet/cache/restore@v3
```
skjnldsv, line 106:
```suggestion
        uses: buildjet/cache/restore@v4
```

.github/workflows/cypress.yml
```diff
@@ -80,7 +80,7 @@ jobs:
         run: npm run cypress:version
 
       - name: Save context
-        uses: buildjet/cache/save@e376f15c6ec6dc595375c78633174c7e5f92dc0e # v3
+        uses: buildjet/cache/save@v3
```
skjnldsv, line 83:
```suggestion
        uses: buildjet/cache/save@v4
```

skjnldsv (approved):

PR #48003, opened 2024-09-15
fix: Restore missing dot in FilesListVirtual.vue

artonge:
Follow up of https://github.com/nextcloud/server/pull/47920


skjnldsv (approved):

susnux (approved):

skjnldsv:
@susnux we are wondering how this was not caught by eslint.
We tried a bit with @artonge and we're not able to make it detect that error. I think this is quite an issue. Do you have any insight? :)

PR #48008, opened 2024-09-15
fix(Entity): Fix magic setter call for custom strong typed setters

provokateurin:
For reference: https://github.com/nextcloud/server/pull/47984

## Summary

When enforcing strict types in Entity.php the fromRow method can fail if the the type to used to store the data in the db is different from the target php type (e.g. bool in sqlite is integer 0/1).
Usually magic methods are used and those don't have strict types, but if you have a custom setter it will trigger an error because the argument type does not match.

## Checklist

### DIFF
--- a/lib/public/AppFramework/Db/Entity.php
+++ b/lib/public/AppFramework/Db/Entity.php
@@ -52,9 +52,8 @@ abstract class Entity {
 		$instance = new static();
 
 		foreach ($row as $key => $value) {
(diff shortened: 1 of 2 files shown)


provokateurin:
Enforcing strict types triggers some Oracle null related problems, so I'm not going to bother with that right now (also makes backporting easier).

miaulalala (approved):

skjnldsv (approved):

PR #48009, opened 2024-09-15
fix: Remove OCP\Files\Storage interface deprecated since version 9

come-nc:
* Resolves: # 

## Summary

Remove deprecated interface and its last uses (mostly in tests).

## Checklist


skjnldsv (approved):

lib/private/Files/Storage/Common.php
```diff
@@ -310,23 +310,31 @@ protected function getCacheDependencies(): CacheDependencies {
...
 	public function getCache($path = '', $storage = null) {
 		if (!$storage) {
 			$storage = $this;
 		}
+		/** @psalm-suppress NoInterfaceProperties The isset check is safe */
 		if (!isset($storage->cache)) {
 			$storage->cache = new Cache($storage, $this->getCacheDependencies());
 		}
+		/** @psalm-suppress NullableReturnStatement False-positive, as the if above avoids this being null */
+		/** @psalm-suppress NoInterfaceProperties Legacy */
```
icewind1991, line 325: maybe add a check if `$storage` is an instance of `Common` and error otherwise.

(same for `getScanner`/`getPropagator`)

lib/public/Files/Storage/IStorage.php
```diff
@@ -279,6 +279,15 @@ public function hash($type, $path, $raw = false);
...
 	public function free_space($path);
 
+	/**
+	 * search for occurrences of $query in file names
+	 *
+	 * @param string $query
+	 * @return array|bool
+	 * @since 6.0.0
+	 */
+	public function search($query);
```
icewind1991, line 289: Since this was only part of the deprecated interface before, it should be safe to just remove this method now.

lib/public/Files/Storage/IStorage.php
```diff
@@ -401,11 +410,19 @@ public function getAvailability();
 	/**
 	 * @since 9.0.0
 	 * @param bool $isAvailable
+	 * @return void
 	 */
 	public function setAvailability($isAvailable);
 
+	/**
+	 * @since 12.0.0
```
icewind1991, line 418: Should this be `@since 31.0.0` instead?

lib/private/Files/Storage/Common.php
come-nc: Is that safe? $storage could be Common with a wrapper around, no?

lib/public/Files/Storage/IStorage.php
come-nc: It was not clear to me since it comes from the other interface what should be put here.

lib/public/Files/Storage/IStorage.php
come-nc: Do you know whether this is used still?

icewind1991: I don't believe it is. (and if it is it **really** shouldn't be)

come-nc: Removing it gives me this:
```
ERROR: UndefinedInterfaceMethod - lib/private/Files/Storage/Wrapper/Encoding.php:391:26 - Method OC\Files\Storage\Storage::search does not exist (see https://psalm.dev/181)
                return $this->storage->search($query);


ERROR: UndefinedInterfaceMethod - lib/private/Files/Storage/Wrapper/Jail.php:331:38 - Method OC\Files\Storage\Storage::search does not exist (see https://psalm.dev/181)
                return $this->getWrapperStorage()->search($query);


ERROR: UndefinedInterfaceMethod - lib/private/Files/Storage/Wrapper/Wrapper.php:326:38 - Method OC\Files\Storage\Storage::search does not exist (see https://psalm.dev/181)
                return $this->getWrapperStorage()->search($query);
```
Do I just remove it from both the interface and the wrappers?

lib/private/Files/Storage/Common.php
susnux: > Common with a wrapper

Does work if you use `instanceOfStorage` instead of native `instanceof`

lib/public/Files/Storage/IStorage.php
come-nc: > I don't believe it is. (and if it is it **really** shouldn't be)

Awesome, removing it then!

lib/private/Files/Storage/Common.php
come-nc: Well let’s see.

lib/public/Files/Storage/IStorage.php
susnux:
```suggestion
	 * @since 12.0.0
	 * @since 31.0.0 Moved from \OCP\Files\Storage
```

apps/dav/tests/unit/Connector/Sabre/FileTest.php
```diff
@@ -73,10 +73,10 @@ protected function tearDown(): void {
 	}
 
 	/**
-	 * @return MockObject|Storage
+	 * @return MockObject|IStorage
```
susnux, line 76:
```suggestion
	 * @return MockObject&IStorage
```

apps/files_trashbin/lib/Trashbin.php
```diff
@@ -139,7 +140,7 @@
 		}
 	}
 
-	private static function setUpTrash($user) {
+	private static function setUpTrash($user): void {
```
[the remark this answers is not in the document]
susnux:
```suggestion
	private static function setUpTrash(string $user): void {
```

susnux (approved):

JulienFS:
Thank you so much ! It's been super hard to dive into the code because of the amount of legacy stuff. I'm glad this part is gone :partying_face:

PR #48010, opened 2024-09-15
feat(db): switch from settype to casts

miaulalala:
Signed-off-by: Anna Larch <anna@nextcloud.com>

## Summary
casting directly to the proper type is up to 6 times faster than using `settype`

Also added the types for the methods, let's see if things go boom (although they might not without strict types)

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/lib/public/AppFramework/Db/Entity.php
+++ b/lib/public/AppFramework/Db/Entity.php
@@ -13,6 +13,7 @@ use function substr;
 /**
  * @method int getId()
  * @method void setId(int $id)
+ * @psalm-type AllowedTypes = 'json'|'blob'|'datetime'|'string'|'int'|'integer'|'bool'|'boolean'|'float'|'double'|'array'|'object'
  * @since 7.0.0
  * @psalm-consistent-constructor
  */
@@ -23,6 +24,7 @@ abstract class Entity {
 	public $id;
 
 	private array $_updatedFields = [];
+	/** @var array<string, AllowedTypes> */
 	private array $_fieldTypes = ['id' => 'integer'];
 
 	/**
@@ -64,10 +66,10 @@ abstract class Entity {
 
 
 	/**
-	 * @return array with attribute and type
+	 * @return array<string, AllowedTypes> with attribute and type
 	 * @since 7.0.0
 	 */
-	public function getFieldTypes() {
+	public function getFieldTypes(): array {
 		return $this->_fieldTypes;
 	}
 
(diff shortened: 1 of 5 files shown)


lib/public/AppFramework/Db/Entity.php
```diff
@@ -76,50 +78,63 @@ public function getFieldTypes() {
...
-				$type = $this->_fieldTypes[$name];
-				if ($type === 'blob') {
-					// (B)LOB is treated as string when we read from the DB
-					if (is_resource($args[0])) {
-						$args[0] = stream_get_contents($args[0]);
-					}
-					$type = 'string';
+		if (!property_exists($this, $name)) {
+			throw new \BadFunctionCallException($name .
+				' is not a valid attribute');
```
provokateurin, line 96:
```suggestion
			throw new \BadFunctionCallException($name . ' is not a valid attribute');
```

lib/public/AppFramework/Db/Entity.php
```diff
@@ -76,50 +78,63 @@ public function getFieldTypes() {
...
+		}
+
+		if ($args[0] === $this->$name) {
+			return;
+		}
```
provokateurin, line 101: I still don't understand the purpose of this return :thinking:

provokateurin (approved):
Noice!

lib/public/AppFramework/Db/Entity.php
miaulalala: It was already there and I don't wanna touch it :scream: 

(I think it has to do with marking a column as updated or not)

provokateurin: I know ;)

kesselb: Do nothing if the current value is equal to the new one

AndyScherzinger (approved):

lib/public/AppFramework/Db/Entity.php
provokateurin: But why is the variable called `name`? (also notice the dollar sign in the variable name lol)

AndyScherzinger:
@miaulalala do you have a rough number how much faster the overall API call will then be with this improvement?

miaulalala:
I don't know - i can do some comparisons against stable30 and master with blackfire?

AndyScherzinger:
Only if it is not too much work. Was just thinking about it, in case it would be worth mentioning with v31, than marketing would need some numbers / comparison.

provokateurin:
Just claim the `up to 6x` :joy: Technically not wrong is the best way to be right

lib/public/AppFramework/Db/Entity.php
come-nc: The $ sign is not part of the name this is a dynamic property get.
It just checks if current value is already the one we’re setting and bails out. $name is the name of the property.

provokateurin: Ah I see, I didn't know this! Well I did know that this stuff is possible, but the syntax is horrific and I didn't recognize it

kesselb: Is `$this->{$name}` less horrific? ;)

kesselb: Or we move the call to a method with a self explaining name like isPropertyDifferent($name, $args[0])

provokateurin: > Is $this->{$name} less horrific? ;)

At least more recognizable to me.

blizzz:
The added return type hints in lib/public/AppFramework/Db/Entity.php might be breaking, cf. https://github.com/nextcloud/officeonline/pull/611. They are also not documented in https://docs.nextcloud.com/server/latest/developer_manual/app_publishing_maintenance/app_upgrade_guide/upgrade_to_31.html.

PR #48047, opened 2024-09-15
refactor: Replace class strings with ::class constants

ChristophWurst:
* Resolves: # 

## TODO

- [x] ...

## Checklist


miaulalala (approved):

st3iny (approved):

apps/federatedfilesharing/tests/FederatedShareProviderTest.php
nickvergessen, line 1: What happened here?

lib/base.php
```diff
@@ -725,7 +725,7 @@ public static function init(): void {
 		\OCP\Util::connectHook(
 			'\OCA\Files_Sharing\API\Server2Server',
 			'preLoginNameUsedAsUserName',
-			'\OC\User\Database',
+			\OC\User\Database::class,
```
nickvergessen, line 728: This should breaking the events:
https://3v4l.org/ccpQf

If you adjusted the original event to also no longer provide the `\` old listeners will be broken.
Same for the `listen()` emitter things.

Potential solution: strip leading slashes when listeners and emits are down inside Util/Emitter-trait

nickvergessen (changes requested):

ChristophWurst:
Conflicts like crazy -> redo in smaller steps

PR #48090, opened 2024-09-16
fix(settings): disable Discover when appstore is disabled

ShGKme:
## Summary

Even when the App Store is disabled, the default page (also listed in the navigation) is Discovery, which doesn't work. And when there is no internet, it results in error: "Avertissement appstoreFetcher Could not get apps from the appstore".

Solution — if the app store is disabled:
- Remove "Discovery" from the navigation
- Use the old "Your apps" page as the default

P.S. It is still possible to open any category, including supported or non-existed by direct link and see an error. Might make sense to explicitly define the list of available categories and check before route enter/update.

🏚️ Before | 🏡 After
---|---
image | image

## Checklist


apps/settings/src/router/routes.ts
```diff
@@ -2,8 +2,11 @@
  * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
  * SPDX-License-Identifier: AGPL-3.0-or-later
  */
+import { loadState } from '@nextcloud/initial-state'
 import type { RouteConfig } from 'vue-router'
```
susnux, line 6: Types :up: 

```suggestion
import type { RouteConfig } from 'vue-router'
import { loadState } from '@nextcloud/initial-state'
```

ShGKme: ESLint does not complain ._.

skjnldsv: We're the eslint

nickvergessen (approved):

susnux (approved):

PR #48098, opened 2024-09-16
feat: Move to ZipFolderPlugin for downloading multiple-nodes

### LINKED ISSUE #41037: Honor `Accept` header when downloading directories
### How to use GitHub

The file format one gets when downloading a directory seems to depend on the HTTP `User-Agent` and does not honor the `Accept` header.

macos user agents seem to get `tar` files, while everyone else gets `zip` files.

This makes it hard to write instructions to e.g. download files and unpack them via the command line (e.g. curl / wget) that are valid cross platform.

It would be great if Nextcloud would honor the `Accept` header when downloading a directory, so that e.g. it is guaranteed that:

```
$ curl -LO -H 'Accept: application/x-tar' <share-link>
# or
$ curl -LO -H 'Accept: application/zip' <share-link>
```

always produces a tar file and a zip file respectively, independent of platform / user agent.

A possible alternative would be a query parameter for the filetype.

susnux:
Fix https://github.com/nextcloud/server/issues/41037

## Summary

This contains multiple "things":
- :sparkle: Allow `GET` on folders to download them
  - if no accept header for ZIP is sent then (for browser download) allow GET params but if nothing set then fallback to Sabre -> should be non-breaking
- :bug: This now also allows downloading "select-multiple" files form e.g. "recent" or "favorites" view where the files are not really located in the same folder.

## Checklist


apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
```diff
@@ -0,0 +1,133 @@
...
+		} elseif ($node instanceof NcFolder) {
+			$streamer->addEmptyDir($filename);
+			$content = $node->getDirectoryListing();
+			foreach ($content as $subNode) {
+				$this->streamNode($streamer, $subNode, $rootPath);
+			}
+		}
+	}
+
+	public function handleDownload(Request $request, Response $response) {
```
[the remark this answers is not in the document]
susnux: how to type `void|false`?

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
```diff
@@ -0,0 +1,136 @@
...
+		$node = $this->tree->getNodeForPath($request->getPath());
+		if (!($node instanceof ICollection)) {
+			// skip if not a collection
+			return null;
+		}
+
+		if (!($node instanceof \OCA\DAV\Connector\Sabre\Directory)) {
+			// only handle directories
+			return null;
+		}
```
come-nc, line 95:
```suggestion
		if (!($node instanceof \OCA\DAV\Connector\Sabre\Directory)) {
			// only handle directories
			return null;
		}
```
directory is always a collection?

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
```diff
@@ -0,0 +1,136 @@
...
+
+		$query = $request->getQueryParameters();
+		$filesParam = $query['files'] ?? '';
+		$files = $request->getHeaderAsArray('X-NC-Files');
```
come-nc, line 99: What happens if both are filled?

susnux: It will be overwritten with the GET params as they need to take precedence to allow "plain browser GET".

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
```diff
@@ -0,0 +1,127 @@
...
+	 * This initializes the plugin.
+	 *
+	 * This function is called by \Sabre\DAV\Server, after
+	 * addPlugin is called.
+	 *
+	 * This method should set up the required event subscriptions.
+	 */
+	public function initialize(Server $server): void {
+		$this->server = $server;
+		$this->server->on('method:GET', $this->handleDownload(...), 100);
```
skjnldsv, line 53: What is the `...` for? :thinking:

come-nc: It is the same as `[$this, 'handleDownload']` but without typo risk. (Actually I think it’s the same as `Closure::fromCallable([$this, 'handleDownload'])` but you get the idea)

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
```diff
@@ -0,0 +1,127 @@
...
+		} elseif ($node instanceof NcFolder) {
+			$streamer->addEmptyDir($filename);
+			$content = $node->getDirectoryListing();
+			foreach ($content as $subNode) {
+				$this->streamNode($streamer, $subNode, $rootPath);
+			}
+		}
+	}
+
+	public function handleDownload(Request $request, Response $response): null|false {
```
come-nc, line 80: Do you really need to return `null|false`? What’s the meaning of the return type?
To support PHP 8.1 you’ll have to use `bool|null` and only type `false|null` in phpdoc.

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
skjnldsv: Arthur solved the issue, I'm smarter now :sparkles:

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
susnux:
```suggestion
	/**
	 * @return false|null
	 */
	public function handleDownload(Request $request, Response $response): bool|null {
```

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
```diff
@@ -0,0 +1,127 @@
...
+
+		$folder = $node->getNode();
+		$content = empty($files) ? $folder->getDirectoryListing() : [];
+		foreach ($files as $path) {
+			$child = $node->getChild($path);
+			assert($child instanceof Node);
+			$content[] = $child->getNode();
+		}
+
+		$streamer = new Streamer($tarRequest, -1, count($content));
```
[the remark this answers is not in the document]
susnux: It is only used for checking if their is accept tar **OR** zip, if nothing of both then it just falls back to Sabre default behavior.
So in this place we only have either tar === true or/and zip === true. 
Meaning if tar is set we provide a tar ball, otherwise a zip archive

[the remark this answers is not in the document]
skjnldsv: Right, just felt weird to have an unused variable :see_no_evil:

come-nc (approved):

skjnldsv (approved):

apps/dav/lib/Connector/Sabre/ZipFolderPlugin.php
```diff
@@ -0,0 +1,136 @@
...
+		// The preferred way would be headers, but this is not possible for simple browser requests ("links")
+		// so we also need to support GET parameters
+		if ($filesParam !== '') {
+			$files = array_map(fn (string $name) => trim($name), explode(',', $filesParam));
+		}
+
+		$accept = $request->getHeaderAsArray('Accept');
+		$acceptParam = $query['accept'] ?? '';
+		if ($acceptParam !== '') {
+			$accept = array_map(fn (string $name) => strtolower(trim($name)), explode(',', $acceptParam));
```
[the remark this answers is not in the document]
susnux: https://github.com/vimeo/psalm/issues/11075

PR #48129, opened 2024-09-17
chore: add `application/zstd`

### LINKED ISSUE #34892: Support media type "zstd" and show archive icon
Zstandard is a new compression method and many applications use it already.
See https://www.rfc-editor.org/rfc/rfc8878.html

Currently, nextcloud does not display the file extension `zst` as an archive type.

## Example:

Each day I create some zip files and one zst file.
The activity app show it like this

Bildschirmfoto 2022-10-30 um 17 41 16

What do I like to achieve:
* Nextcloud should handle zst with the correct MIME type
* Nextcloud should show archive icon for zst files.

I already tried to modify the mimetype handling on my NC v25.0.0 like described here:
* https://docs.nextcloud.com/server/latest/admin_manual/configuration_mimetypes/index.html
* https://portal.nextcloud.com/article/mimetypes-management-19.html

But no success.

Should I provide an PR of my work even if it is not working?

skjnldsv:
image

Remember to run `occ maintenance:mimetype:update-db` to manually update existing instances
fix https://github.com/nextcloud/server/issues/34892


rakekniven (approved):
I am not sure about all changes here, but the zstd related lines look promising.

skjnldsv:
Yeah, I also sorted the other file, seems people were adding the changes at the end :shrug:

kesselb:
> Yeah, I also sorted the other file, seems people were adding the changes at the end 🤷

Thanks :pray:

kesselb (approved):

juliushaertl:
In case you want to also support files that are already on existing instances, make sure to also add to the repair step: https://github.com/nextcloud/server/blob/2eaeeeeafe5fd9377dbaa079a89b2d56ff0bb245/lib/private/Repair/RepairMimeTypes.php#L38

susnux:
> make sure to also add to the repair step: 

Yes please!

skjnldsv:
Done and sorted it by introduced version too

PR #48131, opened 2024-09-17
fix(caldav): explicitly check from component types

miaulalala:
* Resolves: # 

## Summary

Ignore everything but the specific component types that are defined as valid in the RFC. We should only denormalize data that is actually valid, and skip any invalid data structures.  Otherwise we run into the danger of returning trash data.

## TODO

- [x] ...

## Checklist


apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -2988,20 +2988,22 @@ public function getDenormalizedData(string $calendarData): array {
...
-						$hasDTSTART = true;
-					}
-				}
-				// Track first component type and uid
-				if ($uid === null) {
-					$componentType = $component->name;
-					$uid = (string)$component->UID;
-				}
+			$allowedCoponents = $this->propertyMap;
+			if(!in_array($component->name, ['VEVENT', 'VJOURNAL', 'VTODO'])) {
```
ChristophWurst, line 2992:
```suggestion
			if (!in_array($component->name, ['VEVENT', 'VJOURNAL', 'VTODO'])) {
```

ChristophWurst (approved):

nickvergessen:
Could you make the PR description a bit more verbose?
Any issue that is fixed with this? Or example data set that now behaves different?

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -2995,20 +2995,21 @@ public function getDenormalizedData(string $calendarData): array {
...
-					if ($component->DTSTART) {
-						$hasDTSTART = true;
-					}
-				}
-				// Track first component type and uid
-				if ($uid === null) {
-					$componentType = $component->name;
-					$uid = (string)$component->UID;
-				}
+			if (!in_array($component->name, ['VEVENT', 'VJOURNAL', 'VTODO'])) {
```
SebastianKrupinski, line 2998: FYI, you can use $vObject->getBaseComponent(s) this will only return VEVENT, VJOURNAL, VTODO and omit VTIMEZONE. Then you don't need to check for this.

miaulalala: Can you please take over the PR? :pray: I don't have time to focus on this.

SebastianKrupinski: No problem will do!

SebastianKrupinski (approved):

apps/dav/lib/CalDAV/CalDavBackend.php
```diff
@@ -3396,99 +3394,83 @@ public function restoreChanges(int $calendarId, int $calendarType = self::CALEND
...
-		foreach ($vObject->getComponents() as $component) {
-			if ($component->name !== 'VTIMEZONE') {
-				// Finding all VEVENTs, and track them
-				if ($component->name === 'VEVENT') {
-					$vEvents[] = $component;
-					if ($component->DTSTART) {
-						$hasDTSTART = true;
+		$components = $vObject->getBaseComponents();
+		if (count($components) !== 1) {
+			throw new BadRequest('Invalid calendar object must contain exactly one VJOURNAL, VEVENT, or VTODO component type');
```
tcitworld:
```suggestion
			throw new BadRequest('A valid calendar object must contain at least one VJOURNAL, VEVENT, or VTODO component type');
```

SebastianKrupinski: Done

kesselb (approved):
Some tests would be nice ;)

The exceptions should be documented. 

getDenormalizedData might be good fit to move into a own helper / service.

SebastianKrupinski:
> Some tests would be nice ;)
> 
> The exceptions should be documented.
> 
> getDenormalizedData might be good fit to move into a own helper / service.

I added a few test cases for the other component types

Issue #5080, opened 2017-05-23
Calendar invites: Administrator shall choose if server's or user's email address is used


tuergeist:
## Feature Request

I refer to #2345 and #2778. 

If a Nextcloud instance serves for users from different domains (which is the case at least for me), it's annoying that calendar invites are sent with the server's email address instead of the user's once. For me it would be fine if the server's address is taken if there is no user email configured.

To avoid problems with mailservers that might not accept sending emails for more than one domain, this behavior this could be made optional.
(In my case all domains are hosted and handled by one email server and one nextcloud server - I would not like to set up a nextcloud instance per domain).

(If someone is so nice and point me to the place where I can change it in the code myself, would be fine too)

## Implementation

Delegate sending emails to apps. The main target is the Mail app but others may offer this integration too.

## Work packages

- [x] Add a mail sending provider mechanism
- [x] Implement the mail sending provider for Mail
- [x] Add an admin setting to choose which email system should be used (system or mail providers)
- [x] Adjust CalDAV backend to use the providers and system email as fallback
- [x] Developer documentation for the new API
- [x] Admin documentation for the setup and any configuration options

georgehrke:
@tuergeist This is the place to go:
https://github.com/nextcloud/server/blob/master/apps/dav/lib/CalDAV/Schedule/IMipPlugin.php#L109

Feel free to send a PR that implements this optionally. :)

georgehrke:
(Message provides a `setFrom` method, see https://github.com/nextcloud/server/blob/master/lib/private/Mail/Message.php#L81)

tuergeist:
Hi @georgehrke, IMHO the problem is the sender's address which I cannot change easily in neither code locations you pointed to. [Or I'm just too stupid]

tuergeist:
Aaahh. Got it.

ntnsndr:
I'd add that this is an important issue for us at MayFirst.org, who use the same NextCloud across a great many domains and organizational identities. We really need these invitation emails to be coming from the users who send them. Thanks to all who have the capacity and willingness to work on this.

skjnldsv:
@georgehrke :trumpet: could you update this? :)

georgehrke:
cc @jancborchardt This is what we talked about on Talk.

georgehrke:
to summarise the problem:
[...]

PR #48134, opened 2024-09-17
feat: mail provider settings

SebastianKrupinski:
* Resolves: #https://github.com/nextcloud/server/issues/5080


apps/settings/lib/Listener/MailSettingsEventListener.php
```diff
@@ -0,0 +1,106 @@
...
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\IConfig;
+use OCP\IL10N;
+use OCP\Settings\DeclarativeSettingsTypes;
+use OCP\Settings\Events\DeclarativeSettingsGetValueEvent;
+use OCP\Settings\Events\DeclarativeSettingsRegisterFormEvent;
+use OCP\Settings\Events\DeclarativeSettingsSetValueEvent;
+
+/** @template-implements IEventListener<DeclarativeSettingsRegisterFormEvent> */
```
susnux, line 20:
```suggestion
/** @template-implements IEventListener<DeclarativeSettingsGetValueEvent|DeclarativeSettingsRegisterFormEvent|DeclarativeSettingsSetValueEvent> */
```

apps/settings/lib/Listener/MailSettingsEventListener.php
```diff
@@ -0,0 +1,106 @@
...
+		
+		$event->setValue(
+			match($event->getFieldId()) {
+				'mail_providers_disabled' => $this->config->getSystemValueInt('mail_providers_disabled', 0),
+			}
+		);
+
+	}
+
+	private function handleSetValue(Event $event) {
```
susnux, line 95:
```suggestion
	private function handleSetValue(Event $event): void {
```

apps/settings/lib/Listener/MailSettingsEventListener.php
SebastianKrupinski: Ty

SebastianKrupinski:
@marcoambrosini

apps/settings/lib/Listener/MailSettingsEventListener.php
```diff
@@ -0,0 +1,108 @@
...
+	private function handleRegister(DeclarativeSettingsRegisterFormEvent $event): void {
+
+		$event->registerSchema(Application::APP_ID, [
+			'id' => 'mail-provider-support',
+			'priority' => 11,
+			'section_type' => DeclarativeSettingsTypes::SECTION_TYPE_ADMIN,
+			'section_id' => 'server',
+			'storage_type' => DeclarativeSettingsTypes::STORAGE_TYPE_EXTERNAL,
+			'title' => $this->l->t('System Mails'),
+			'description' => $this->l->t('System e-mails are messages generated automatically by Nextcloud. They are sent for example when a share is created or when inviting attendees to a calendar event..'),
```
kesselb, line 62: > They are sent for example when a share is created

I find it weird to use "share is created" as an example because we don't use the mail provider there.

apps/settings/lib/Listener/MailSettingsEventListener.php
```diff
@@ -0,0 +1,108 @@
...
+			}
+		);
+
+	}
+
+	private function handleSetValue(DeclarativeSettingsSetValueEvent $event): void {
+
+		switch ($event->getFieldId()) {
+			case 'mail_providers_disabled':
+				$this->config->setSystemValue('mail_providers_disabled', $event->getValue());
```
kesselb, line 101: I don't know how you planned the feature with @ChristophWurst, but changing values in config.php via web admin is tricky. 

Examples:

1) config_is_read_only = true
2) multiple config files (the additional configuration files overwrite the values in config.php, setSystemValue modifies only config.php not the additional configuration files)
3) The unfinished and undocumented "overwrite configuration variables via env" feature may overload the variable as well.

That's all unrelated to this pull request, but the incomplete handling in our configuration classes is the reason that we usually don't change config.php via web admin. 

I recommend against adding a ui with system config as backend.

apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
```diff
@@ -265,12 +265,14 @@ public function schedule(Message $iTipMessage) {
 		$mailService = null;
 
 		try {
-			// retrieve user object
-			$user = $this->userSession->getUser();
-			// evaluate if user object exist
-			if ($user !== null) {
-				// retrieve appropriate service with the same address as sender
-				$mailService = $this->mailManager->findServiceByAddress($user->getUID(), $sender);
+			if ($this->config->getSystemValueInt('mail_providers_disabled', 0) === 0) {
```
kesselb, line 268:
```suggestion
			if ($this->config->getSystemValueInt('mail_providers_disabled', 0) === 0) {
```

Must be documented in https://github.com/nextcloud/server/blob/master/config/config.sample.php.

https://docs.nextcloud.com/server/latest/admin_manual/configuration_server/config_sample_php_parameters.html is also generated from the config.sample.php.

kesselb, line 268:
```suggestion
			if ($this->config->getSystemValueBool('mail_providers_disabled', false)) {
```

kesselb, line 268: Easier to use/read: mail_providers_enabled with default true

kesselb (dismissed):
config.sample.php needs an update

apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
SebastianKrupinski: This is what design wanted...

apps/settings/lib/Listener/MailSettingsEventListener.php
SebastianKrupinski: This feature functions identical to the new Windows support feature that was put in last month

apps/settings/lib/Listener/MailSettingsEventListener.php
kesselb: I'm quite sure that design wanted a description of what the feature actually does. I can't imagine that design asked you to mention in the description that this feature toggle affects the behavior of the sharebyemail provider, whereas it doesn't affect it.

apps/settings/lib/Listener/MailSettingsEventListener.php
provokateurin, line 1: Please at least use https://github.com/nextcloud/server/blob/master/apps/testing/lib/Settings/DeclarativeSettingsForm.php if you really want to do this.

apps/settings/lib/Listener/MailSettingsEventListener.php
marcoambrosini: Yep, if we don't use the mail provider there than the example is wrong and we should change it to something that is actually affected by the setting

SebastianKrupinski:
image

Latest version

marcoambrosini (approved):
✅ from design

kesselb:
The Linter is failing.

apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
```diff
@@ -265,12 +265,14 @@ public function schedule(Message $iTipMessage) {
 		$mailService = null;
 
 		try {
-			// retrieve user object
-			$user = $this->userSession->getUser();
-			// evaluate if user object exist
-			if ($user !== null) {
-				// retrieve appropriate service with the same address as sender
-				$mailService = $this->mailManager->findServiceByAddress($user->getUID(), $sender);
+			if ($this->config->getValueInt('core', 'mail_providers_disabled', 0) === 0) {
```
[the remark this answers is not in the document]
kesselb: Yep, therefore I suggested it ;)

apps/settings/lib/Settings/Admin/SystemMail.php
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	public function getSchema(): array {
+		return [
+			'id' => 'mail-provider-support',
+			'priority' => 11,
+			'section_type' => DeclarativeSettingsTypes::SECTION_TYPE_ADMIN,
+			'section_id' => 'server',
+			'storage_type' => DeclarativeSettingsTypes::STORAGE_TYPE_EXTERNAL,
+			'title' => $this->l->t('System Mails'),
```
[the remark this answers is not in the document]
kesselb: I spoke with Marco, and he's okay with adjusting the text to make the feature easier to understand.

apps/settings/lib/Settings/Admin/SystemMail.php
```diff
@@ -0,0 +1,52 @@
...
+			'description' => $this->l->t('System emails are automatically generated by Nextcloud when people perform actions like inviting others to calendar events.'),
```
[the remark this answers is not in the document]
kesselb: I believe it would be really helpful to adjust the description here, making the function easier to understand.

apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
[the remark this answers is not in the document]
kesselb: Just in case you meant "does not" instead of "does" ;)

`DeclarativeSettingsTypes::STORAGE_TYPE_INTERNAL`: Does not support booleans

`DeclarativeSettingsTypes::STORAGE_TYPE_EXTERNAL`: The controller will emit an even, the rest is up to you and therefore we can use booleans here.

apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
SebastianKrupinski: The forms do not work properly with boolean types, to make it less confusing I made it an integer, this way we don't need to cast them. But sure I've made change.

apps/dav/lib/CalDAV/Schedule/IMipPlugin.php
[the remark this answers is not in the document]
SebastianKrupinski: >DeclarativeSettingsTypes::STORAGE_TYPE_INTERNAL: Does not support booleans

>DeclarativeSettingsTypes::STORAGE_TYPE_EXTERNAL: The controller will emit an even, the rest is up to you and therefore we can use booleans here.

Its not the retrieving and saving that does not work properly, the actual forms don't handle boolean properly. So they need to be case to integers either way, which then is confusing. But sure, done.

apps/settings/lib/Listener/SystemMailSettingsListener.php
```diff
@@ -0,0 +1,65 @@
...
+		if ($event instanceof DeclarativeSettingsSetValueEvent) {
+			$this->handleSetValue($event);
+			return;
+		}
+		
+	}
+
+	private function handleGetValue(DeclarativeSettingsGetValueEvent $event): void {
+		
+		$event->setValue(
```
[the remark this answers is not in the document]
SebastianKrupinski: Sure

apps/settings/lib/Listener/SystemMailSettingsListener.php
```diff
@@ -0,0 +1,65 @@
...
+			match($event->getFieldId()) {
+				'mail_providers_disabled' => $this->config->getValueInt('core', 'mail_providers_disabled', 0),
+			}
+		);
+
+	}
+
+	private function handleSetValue(DeclarativeSettingsSetValueEvent $event): void {
+
+		switch ($event->getFieldId()) {
```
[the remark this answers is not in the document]
SebastianKrupinski: sure

apps/settings/lib/Settings/Admin/SystemMail.php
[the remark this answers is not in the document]
keunes: This setting & explanation would only be available/visible from a certain Nextcloud version, right? If so, the 'version 30' could be dropped I think? (Only the Mail version would need to be mentioned, as the other condition is already fulfilled by definition).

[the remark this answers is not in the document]
kesselb: > Mail provider enables sending emails directly through the user’s personal email account. At present, this functionality is limited to calendar invitations. It requires Nextcloud Mail 4.1 and an email account in Nextcloud Mail that matches the user's email address in Nextcloud.

kesselb (approved):

PR #48137, opened 2024-09-17
feat: Add OCP interface to format richtext into string

come-nc:
* Resolves: # 

## Summary

Adds an interface into OCP to format richtext into a string, fixing a code duplication from activity, notifications and setupchecks.
Solves this TODO: https://github.com/nextcloud/server/blob/f8dde2d254ded9760cfa7a24178e1f1ca141ba41/core/Command/SetupChecks.php#L33

## Checklist


provokateurin (approved):
Nice one :+1:

sorbaugh (approved):

susnux (approved):

lib/private/RichObjectStrings/RichTextFormatter.php
```diff
@@ -0,0 +1,36 @@
...
+class RichTextFormatter implements IRichTextFormatter {
+	/**
+	 * @throws \InvalidArgumentException if a parameter has no name or no type
+	 */
+	public function richToParsed(string $message, array $parameters): string {
+		$placeholders = [];
+		$replacements = [];
+		foreach ($parameters as $placeholder => $parameter) {
+			$placeholders[] = '{' . $placeholder . '}';
+			foreach (['name','type'] as $requiredField) {
```
nickvergessen, line 23:
```suggestion
			foreach (['name', 'type'] as $requiredField) {
```

nickvergessen:
Otherwise good and looking forward to it

lib/private/RichObjectStrings/RichTextFormatter.php
susnux: why is this not covered by cs-fixed? :eyes:

nickvergessen: I guess https://github.com/nextcloud/server/blob/master/.php-cs-fixer.dist.php#L23-L35 is broken?

provokateurin: Why would it be broken?

susnux: Because this change is not flagged an cs error

provokateurin: But do we have a rule for this?
The code @nickvergessen mentioned is only there to ignore git-ignored files.

come-nc: This is an array, not a function call so a space is not expected there I think

lib/public/RichObjectStrings/IRichTextFormatter.php
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Parse rich text and format it with the richobjects
+ *
+ * @since 31.0.0
+ */
+interface IRichTextFormatter {
+	/**
+	 * @since 31.0.0
+	 * @param string $message
+	 * @param array<string,array<string,string>> $parameters
```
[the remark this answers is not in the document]
come-nc: I’m always afraid to break tools by doing that because the type and the name of the parameter is space separated to adding space in the type looks wrong.

lib/private/RichObjectStrings/RichTextFormatter.php
provokateurin: But I agree that we should have a rule for it.

PR #48152, opened 2024-09-17
chore(legacy): Introduce public version class and drop version methods from OC_Util

juliusknorr:
- Introduce a ServerVersion class with properly typed methods for getting different version parts and strings
- Remove and migrate any usages of OC_Util to the new API

Next step could be to deprecated related OCP\Util methods.


lib/public/Util.php
```diff
@@ -37,7 +37,7 @@ class Util {
 	 * @since 4.0.0
```
come-nc, line 37: `@deprecated`

lib/public/Util.php
```diff
@@ -68,7 +68,7 @@ public static function setChannel($channel) {
 	 * @since 8.1.0
```
come-nc, line 68: same

come-nc (approved):

lib/private/legacy/OC_Defaults.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
+use OCP\IConfig;
```
nickvergessen, line 3: should be after license header

lib/public/ServerVersion.php
```diff
@@ -0,0 +1,94 @@
+<?php
```
nickvergessen, line 1: declare strict

lib/public/ServerVersion.php
```diff
@@ -0,0 +1,94 @@
...
+ * @since 31.0.0
+ */
+class ServerVersion {
+
+	private array $version;
+	private string $versionString;
+	private string $build;
+	private string $channel;
+
+	public function __construct() {
```
nickvergessen, line 19: Required since

lib/public/ServerVersion.php
```diff
@@ -0,0 +1,94 @@
...
+
+	/**
+	 * @since 31.0.0
+	 */
+	public function getVersionString(): string {
+		return $this->versionString;
+	}
+
+	/**
+	 * @since 31.0.0
```
nickvergessen, line 69:
```suggestion
	 * @psalm-return 'beta'|'stable'|'enterprise'|'git'
	 * @since 31.0.0
```

nickvergessen (dismissed):

susnux (approved):

nickvergessen:
Since this was defacto-standard, please document the new API and mention the removal in the docs

PR #48145, opened 2024-09-17
fix: Drop unnecessary exit

artonge:
This is making the CI stop for no good reason.

PR #48177, opened 2024-09-18
fix: Override start method of \Sabre\DAV\Server to remove exception o…

artonge:
Follow-up of https://github.com/nextcloud/server/pull/48145

### DIFF
--- a/apps/dav/lib/Connector/Sabre/Server.php
+++ b/apps/dav/lib/Connector/Sabre/Server.php
@@ -25,4 +25,27 @@ class Server extends \Sabre\DAV\Server {
 		self::$exposeVersion = false;
 		$this->enablePropfindDepthInfinity = true;
 	}
+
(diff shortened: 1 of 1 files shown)


apps/dav/lib/Connector/Sabre/Server.php
```diff
@@ -25,4 +25,27 @@ public function __construct($treeOrNode = null) {
 		self::$exposeVersion = false;
 		$this->enablePropfindDepthInfinity = true;
 	}
+
+	// Copied from 3rdparty/sabre/dav/lib/DAV/Server.php
+	// Should be them exact same without the exception output.
+	public function start() {
```
come-nc, line 31:
```suggestion
	public function start(): void {
```

icewind1991 (approved):

come-nc (approved):

nickvergessen (approved):

Issue #45630, opened 2022-06-01
Streamline doctrine/pdo/oci


kesselb:
### Describe the task

We discovered some weirdness with our groupware apps on oracle db.  

**Problem:** Booleans are not properly stored as 0/1 but null/false. We need some hacks in our code to make it work in properly (e.g. the explicit check for `true`).

Example SQL
![image](https://user-images.githubusercontent.com/3902676/171370692-b2a0abad-c489-4fde-a5f4-ced0e73adcac.png)

Example Code
![image](https://user-images.githubusercontent.com/3902676/171370789-e6e70f82-8b5a-4541-ad37-bb39db3f5596.png)

**Findings:**

1. Our type definitions (https://github.com/nextcloud/server/blob/010924916c74e62089131ff9f62b8b0e96ff539b/lib/public/DB/QueryBuilder/IQueryBuilder.php#L42-L61) are mapped to pdo types. 

2. But for oci we are not using pdo but the (recommended) native driver: https://github.com/nextcloud/server/blob/010924916c74e62089131ff9f62b8b0e96ff539b/lib/private/DB/ConnectionFactory.php#L61-L66

3. On read/write something to the database via doctrine/dbal we reach getBindingInfo (https://github.com/nextcloud/3rdparty/blob/2ae1a1d6f688ae8394d6559ee673fecbee975db4/doctrine/dbal/src/Connection.php#L1681-L1695). Example for a boolean value: `getBindingInfo($value = true, $type = 5)`. 5 is not a string and not an instance of Type so type is changed to `ParameterType::STRING`. _If we pass `Types::BOOLEAN` (or an instance of Type) doctrine would convert the value to the right value for the given platform (https://github.com/nextcloud/3rdparty/blob/2ae1a1d6f688ae8394d6559ee673fecbee975db4/doctrine/dbal/src/Platforms/AbstractPlatform.php#L2623-L2652)._ 
4.  Finally we end up in `3rdparty/doctrine/dbal/src/Driver/OCI8/Statement.php` (https://github.com/nextcloud/3rdparty/blob/2ae1a1d6f688ae8394d6559ee673fecbee975db4/doctrine/dbal/src/Driver/OCI8/Statement.php#L108-L114) with `$variable = false, $type = 5, $this->convertParameterType($type) = 1)`. I assume that oci driver or oracle db then fallback to some default logic as the provided input is invalid (field is `number(1)`, value is `false` and type `string`).

**Solutions**

[...]

PR #48182, opened 2024-09-18
fix: fix share reminder job for oracle

icewind1991:
This was always broken but nobody noticed because of CI issues

### DIFF
--- a/lib/public/DB/QueryBuilder/IQueryBuilder.php
+++ b/lib/public/DB/QueryBuilder/IQueryBuilder.php
@@ -10,6 +10,7 @@ namespace OCP\DB\QueryBuilder;
 use Doctrine\DBAL\ArrayParameterType;
 use Doctrine\DBAL\Connection;
 use Doctrine\DBAL\ParameterType;
+use Doctrine\DBAL\Types\Types;
 use OCP\DB\Exception;
 use OCP\DB\IResult;
 use OCP\IDBConnection;
@@ -28,7 +29,7 @@ interface IQueryBuilder {
 	/**
 	 * @since 9.0.0
 	 */
-	public const PARAM_BOOL = ParameterType::BOOLEAN;
+	public const PARAM_BOOL = Types::BOOLEAN;
 	/**
 	 * @since 9.0.0
 	 */
--- a/apps/files_sharing/lib/SharesReminderJob.php
+++ b/apps/files_sharing/lib/SharesReminderJob.php
@@ -89,8 +89,8 @@ class SharesReminderJob extends TimedJob {
 						$qb->expr()->eq('s.share_type', $qb->expr()->literal(IShare::TYPE_EMAIL))
 					),
 					$qb->expr()->eq('s.item_type', $qb->expr()->literal('folder')),
-					$qb->expr()->gte('s.expiration', $qb->createNamedParameter($minDate->format('Y-m-d H:i:s'))),
-					$qb->expr()->lt('s.expiration', $qb->createNamedParameter($maxDate->format('Y-m-d H:i:s'))),
+					$qb->expr()->gte('s.expiration', $qb->createNamedParameter($minDate, IQueryBuilder::PARAM_DATE)),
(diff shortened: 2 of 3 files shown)


artonge (approved):

come-nc (approved):

nickvergessen (approved):

lib/public/DB/QueryBuilder/IQueryBuilder.php
```diff
@@ -28,7 +29,7 @@ interface IQueryBuilder {
 	/**
 	 * @since 9.0.0
 	 */
-	public const PARAM_BOOL = ParameterType::BOOLEAN;
+	public const PARAM_BOOL = Types::BOOLEAN;
```
nickvergessen, line 32: This sounds like a serious bug in doctrine which we should report and get addressed upstream?
The binding type should be `ParameterType::`
https://github.com/nextcloud/3rdparty/blob/9ca60e9d2b12f711a756c35d37fd19c178db685b/doctrine/dbal/src/Types/BooleanType.php#L56-L59
Doc block of ParameterType also states that it is for statement parameters:
https://github.com/nextcloud/3rdparty/blob/a9be8d0755e6d786c34e44ef29e88d8f9b90c65b/doctrine/dbal/src/ParameterType.php#L6

The `Types` are the dbal types

nickvergessen:
Backport?

nickvergessen:
- At least the BOOL fix I have to cherry-pick to https://github.com/nextcloud/server/pull/48286

lib/public/DB/QueryBuilder/IQueryBuilder.php
nickvergessen: Let me try if bumping doctrine to 3.9.1 fixes it https://github.com/nextcloud/server/pull/48306

nickvergessen: Okay, still fails with 3.9.1

kesselb: Thanks! 

I've debugged the weirdness with OCI and booleans a while ago: https://github.com/nextcloud/server/issues/45630

The type constants from doctrine/dbal allows us to use doctrine type conversion logic. I was unsure back then about changing it because apps, especially on OCI, may depend on the broken behavior. 

Should we also update the other constants to use the doctrine types?

nickvergessen: > Should we also update the other constants to use the doctrine types?

As said I think the switch that was done here is actually wrong. It uses the DBAL architectural type now and not the PDO parameter type anymore.

kesselb: > As said I think the switch that was done here is actually wrong. It uses the DBAL architectural type now and not the PDO parameter type anymore.

Depends ;)

Using the DBAL type leads to doctrine taking care of the type juggeling for us. That's probably good, but also a risky change. We don't know if the results are still the same.

We are not using the pdo driver for OCI. I think that's the reason for the weird behavior with OCI, especially with booleans. The OCI driver just didn't know what to do with  type "5" (because that's the pdo constant) and fallbacks to string as default. If we use the dbal type doctrine will convert the boolean to 0/1 and set type int.

I've discussed https://github.com/nextcloud/server/issues/45630 with Christoph a while ago, and we didn't know how to move forward :shrug:

AndyScherzinger:
> Backport?

For v30 it sounds like a safe bet. For former versions, if an issue that it should be fixed too

lib/public/DB/QueryBuilder/IQueryBuilder.php
icewind1991: Note that we already use the DBAL types for datetime and json

AndyScherzinger:
@nickvergessen does your backport cover 30?

nickvergessen:
I only backported the database/querybuilder change, not the reminder job stuff

AndyScherzinger:
@icewind1991 does this need backporting (broken on 30 or older?)

icewind1991:
The [share reminder](https://github.com/nextcloud/server/pull/47412) didn't get backported so it's not required to backport this

PR #48183, opened 2024-09-18
feat(deps): Upgrade phpseclib to v3

### LINKED ISSUE #21373: Implement ed25519 certificates support for sftp external storage
I currently manage several nextcloud v19 and v18 distributions on which I have activated the module for storing external folders, and often I find myself managing this module with rsa key authentication, it is possible to also give it authentication with public / private signature certificates also with elliptic curve certificates see keys ed25519? currently, as can also be seen from the attached authentication via sftp, it is possible only by using private or public key rsa non ed25519 or ecdsa

Immagine

### LINKED ISSUE #49229: [Bug]: Uploading files larger than 2GB to SFTP external storage backend fails

- [x] This is a **bug**, not a question or a configuration/webserver/proxy issue.
- [x] This issue is **not** already reported on [Github](https://github.com/nextcloud/server/issues?q=is%3Aopen+is%3Aissue+label%3Abug) OR [Nextcloud Community Forum](https://help.nextcloud.com/) _(I've searched it)_.
- [x] Nextcloud Server **is** up to date. See [Maintenance and Release Schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule) for supported versions.
- [x] I agree to follow Nextcloud's [Code of Conduct](https://nextcloud.com/contribute/code-of-conduct/).

### Bug description

Uploading files larger than 2GB to SFTP external storage backend fails. This is on a docker install using the official Nextcloud image version `29.0.9-apache`.

### Steps to reproduce

1. Connect sftp storage backend space
2. Upload or copy file larger than 2GB to it.
3. It fails.

### Expected behavior

Successful file upload.

### Nextcloud Server version

29

### Operating system

Other

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```json

```

### List of activated Apps

```shell

```

### Nextcloud Signing status

```shell

```

### Nextcloud Logs

```json
The log has these entries:

[PHP] Info: Invalid HMAC at /var/www/html/3rdparty/phpseclib/phpseclib/phpseclib/Net/SSH2.php#3717
	COPY /remote.php/dav/files/redacted/test_file_over2gb.dat

[PHP] Info: Connection closed by server at /var/www/html/3rdparty/phpseclib/phpseclib/phpseclib/Net/SSH2.php#4069
	COPY /remote.php/dav/files/redacted/test_file_over2gb.dat

[no app in context] Warning: Failed to copy stream to storage
	COPY /remote.php/dav/files/redacted/test_file_over2gb.dat
```

### Additional info

This issue is probably related to or the same as https://github.com/nextcloud/server/issues/48231

It happens on a docker container install using the official Nextcloud image with version `29.0.9-apache`

According to Claude, it might be that the packaged phpseclib version is too old. In fact, checking the composer.json in my docker container, it says: `"phpseclib/phpseclib": "^2.0.45"`

Claude says: 
> That explains the issue! You're using phpseclib 2.x which has known limitations with files larger than 2GB due to how it handles HMAC calculations with 32-bit integers.

Is this plausible?

When I connect the same storage via FTPS, uploads above 2GB succeed. 
I am not sure when this bug was introduced since it definitely worked in the past, we uploaded >2GB files to the storage using the SFTP backend before.

susnux:
* For https://github.com/nextcloud/3rdparty/pull/1972

## Summary

Updated to v3 - needed to adjust quiet some places.
AI was just used for https://github.com/nextcloud/server/pull/48183/changes/3983f8a2b80df3f16e549d375d6b49177382833c to help finding a working alternative because the old implementation was using v2 private functions which were not PHP private but just prefixed with `_`.
In v3 they made them really private so we have two options:
- drop our custom streams and use the streams provided by the library
- the solution implemented here using reflection to hack around the private fields...

## Checklist

### DIFF
--- a/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
+++ b/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
@@ -12,7 +12,7 @@ use OCA\Files_External\Lib\StorageConfig;
 use OCP\IConfig;
 use OCP\IL10N;
 use OCP\IUser;
-use phpseclib\Crypt\RSA as RSACrypt;
+use phpseclib3\Crypt\RSA as RSACrypt;
 
 /**
  * RSA public key authentication
@@ -41,15 +41,16 @@ class RSA extends AuthMechanism {
 	 * @return void
 	 */
 	public function manipulateStorageConfig(StorageConfig &$storage, ?IUser $user = null) {
-		$auth = new RSACrypt();
-		$auth->setPassword($this->config->getSystemValue('secret', ''));
-		if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
+		try {
+			$auth = RSACrypt::loadPrivateKey(
+				$storage->getBackendOption('private_key'),
+				$this->config->getSystemValue('secret', '')
+			);
+		} catch (\Throwable) {
 			// Add fallback routine for a time where secret was not enforced to be exists
-			$auth->setPassword('');
-			if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
-				throw new \RuntimeException('unable to load private key');
-			}
+			$auth = RSACrypt::loadPrivateKey($storage->getBackendOption('private_key'));
 		}
+
 		$storage->setBackendOption('public_key_auth', $auth);
 	}
 
@@ -57,17 +58,15 @@ class RSA extends AuthMechanism {
 	 * Generate a keypair
 	 *
 	 * @param int $keyLenth
-	 * @return array ['privatekey' => $privateKey, 'publickey' => $publicKey]
 	 */
-	public function createKey($keyLength) {
+	public function createKey($keyLength): RSACrypt\PrivateKey {
 		$rsa = new RSACrypt();
-		$rsa->setPublicKeyFormat(RSACrypt::PUBLIC_FORMAT_OPENSSH);
-		$rsa->setPassword($this->config->getSystemValue('secret', ''));
-
+		
 		if ($keyLength !== 1024 && $keyLength !== 2048 && $keyLength !== 4096) {
 			$keyLength = 1024;
 		}
 
-		return $rsa->createKey($keyLength);
+		return $rsa->createKey($keyLength)
+			->withPassword($this->config->getSystemValue('secret', ''));
 	}
 }
--- a/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
+++ b/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
@@ -66,7 +66,8 @@ class RSA extends AuthMechanism {
 			$keyLength = 1024;
 		}
 
+		$secret = $this->config->getSystemValue('secret', '');
 		return $rsa->createKey($keyLength)
-			->withPassword($this->config->getSystemValue('secret', ''));
+			->withPassword($secret);
 	}
 }
--- a/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
+++ b/apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
@@ -14,7 +14,7 @@ use OCA\Files_External\Lib\StorageConfig;
 use OCP\IConfig;
 use OCP\IL10N;
 use OCP\IUser;
-use phpseclib\Crypt\RSA as RSACrypt;
+use phpseclib3\Crypt\RSA as RSACrypt;
 
 /**
  * RSA public key authentication
@@ -45,15 +45,16 @@ class RSA extends AuthMechanism {
 	 */
 	#[\Override]
 	public function manipulateStorageConfig(StorageConfig &$storage, ?IUser $user = null) {
-		$auth = new RSACrypt();
-		$auth->setPassword($this->config->getSystemValue('secret', ''));
-		if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
+		try {
+			$auth = RSACrypt::loadPrivateKey(
+				$storage->getBackendOption('private_key'),
+				$this->config->getSystemValue('secret', '')
+			);
+		} catch (\Throwable) {
 			// Add fallback routine for a time where secret was not enforced to be exists
-			$auth->setPassword('');
-			if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
-				throw new \RuntimeException('unable to load private key');
-			}
+			$auth = RSACrypt::loadPrivateKey($storage->getBackendOption('private_key'));
 		}
+
 		$storage->setBackendOption('public_key_auth', $auth);
 	}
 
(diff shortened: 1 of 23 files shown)


susnux:
Problem with this:
`Checker::createSignatureData` does create a different signature... No idea what is happening the parameters are the same.

joshtrichards:
Perhaps `phpseclib2_compat` as an intermediate step for v31+ (or at least v32+)?

https://phpseclib.com/docs/why#phpseclib2_compat

apps/files_external/lib/Lib/Auth/PublicKey/RSAPrivateKey.php
```diff
@@ -39,12 +39,12 @@ public function __construct(
 	 * @return void
 	 */
 	public function manipulateStorageConfig(StorageConfig &$storage, ?IUser $user = null) {
-		$auth = new RSACrypt();
-		$auth->setPassword($this->config->getSystemValue('secret', ''));
-		if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
+		$auth = new RSA\PrivateKey();
+		$auth->withPassword($this->config->getSystemValue('secret', ''));
```
joshtrichards:
```suggestion
		$auth = $auth->withPassword($this->config->getSystemValue('secret', ''));
```
v3 is immutable

apps/files_external/lib/Lib/Auth/PublicKey/RSAPrivateKey.php
```diff
@@ -39,12 +39,12 @@ public function __construct(
...
+		if (!$auth->loadPrivateKey($storage->getBackendOption('private_key'))) {
 			// Add fallback routine for a time where secret was not enforced to be exists
-			$auth->setPassword('');
-			if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
+			$auth->withPassword('');
```
joshtrichards:
```suggestion
			$auth = $auth->withPassword('');
```

apps/files_external/lib/MountConfig.php
joshtrichards, line 1: L220 for v3

	private static function getCipher() {
		$cipher = new AES('cbc');

apps/files_external/lib/Lib/Auth/PublicKey/RSA.php
```diff
@@ -41,33 +41,33 @@ public function __construct(
 	 * @return void
 	 */
 	public function manipulateStorageConfig(StorageConfig &$storage, ?IUser $user = null) {
-		$auth = new RSACrypt();
-		$auth->setPassword($this->config->getSystemValue('secret', ''));
-		if (!$auth->loadKey($storage->getBackendOption('private_key'))) {
+		try {
+			$auth = RSACrypt::loadPrivateKey(
```
terrafrost, line 49: Might be nice to support EC / DSA keys as well. One can do that thusly:

```php
$auth = \phpseclib3\Crypt\PublicKeyLoader::load(
    $storage->getBackendOption('private_key'),
    $this->config->getSystemValue('secret', '');
);
if (!$auth instanceof \phpseclib3\Crypt\Common\PrivateKey) {
    throw new \Exception('A private key needs to be loaded - not a ' . $auth::class . ' key');
}
```
I suppose it might also be worthwhile to realize that if no key can be detected (eg. you try to load a PHP file as a key) then a `\phpseclib3\Exception\NoKeyLoadedException` will be thrown.

That said, if more than just RSA keys are going to be supported then renaming the class would prob be appropriate as well.

susnux: Please open a feature request and if you like also implement that then as a contribution.
We would be happy to review such contribution :)

But for this pull request new features are out-of-scope.

apps/files_external/lib/Lib/Storage/SFTPReflection.php
susnux, line 1: Any options on this solution?
Should we keep the "performance" streams with reflection or drop them completely and just use the library streams?

nickvergessen:
Should document this change, as it breaks at least 2 shipped apps (app_api, nextcloud_announcements) by removing phpseclib 2, and integration_discourse because it ships phpseclib 3 without mozart/namespacing it

susnux:
> as it breaks at least 2 shipped apps

we can add the phpseclib_compat package to add intermediate support for those apps.

nickvergessen:
> we can add the phpseclib_compat package to add intermediate support for those apps.

Documenting it now and early should be good enough

miaulalala (changes requested):
Did some benchmarking, scripts and instructions in the linked gist below. 

Ran `SFTPReadStream`/`SFTPWriteStream` against plain `SFTP::get()`/`put()` to check whether the pipelining they're meant to provide actually pays off.

### Setup: 
- standalone PHP script, real phpseclib3 v3.0.55 via Composer, the three files from this branch dropped in unmodified, against a local `atmoz/sftp` container. Tested both on localhost and with ~30ms latency injected via `tc netem` inside the container, to approximate a real remote host.

### Results (medians over 5 runs):

localhost, no injected latency, 20 MiB:
- write: public `put()` 247ms vs. reflection stream 9,659ms (~39x slower)
- read: public `get()` 238ms vs. reflection stream 218ms (roughly even)

~30ms injected latency, 1 MiB:
- write: public `put()` 844ms vs. reflection stream 1,641ms (~1.9x slower)
- read: public `get()` 883ms vs. reflection stream 1,173ms (~1.3x slower)

The reflection streams were slower than the public API in pretty much every condition I tried, not faster. I checked this isn't a benchmark artifact (MD5 of the round-tripped file matches the input, so no silent retries).

`SFTPWriteStream::stream_flush()` sends one 64 KiB packet and immediately blocks on `read_put_responses(1)` per chunk. Fully synchronous, not pipelined. 

phpseclib3's own public `put()` already queues up to `NET_SFTP_UPLOAD_QUEUE_SIZE` (1024) packets before batch-reading responses, i.e. it already does the windowed upload the custom stream's docblock claims to add, and does it better. 

`SFTPReadStream` does have one-chunk read-ahead, which is why it's closer, but it still trails `get()` once latency is added.

Script + full setup/repro instructions: https://gist.github.com/miaulalala/ef2ed83521e515984dc26ed8bd60a154

AI was used to create the scripts and benchmark data, so please reconfirm with your own local setup. If those numbers are confirmed for someone else too, I'd vote for dropping the custom streams and using the library's public API.

ETA: added some latency runs in the gist, too, if you're interested.

CarlSchwan (approved):
I feel like having up to date security relevant dependency is more important than performance

susnux:
@miaulalala thats the wrong comparison, simple put / get will always be faster.
But this is about streams as we need stream API. So the proper comparison would be custom streams vs phpseclib streams.

We optimize for round trips, so latency needs to be taken into account for the write stream.
With that in mind the benchmark results are different:

| RTT | Size | Op | Custom | phpseclib built-in | Speedup (custom ÷ phpseclib) |
|-------|-------|-------|-------------|--------------------|------------------------------|
| ~0 ms (LAN) | 1 MB | write | 1.84 MB/s | 33.58 MB/s | 0.05× |
| ~0 ms (LAN) | 1 MB | read | 297.88 MB/s | 29.48 MB/s | 10.1× |
| ~0 ms (LAN) | 8 MB | write | 1.94 MB/s | 31.38 MB/s | 0.06× |
| ~0 ms (LAN) | 8 MB | read | 315.47 MB/s | 27.30 MB/s | 11.6× |
| ~0 ms (LAN) | 32 MB | write | 1.90 MB/s | 38.23 MB/s | 0.05× |
| ~0 ms (LAN) | 32 MB | read | 420.94 MB/s | 37.20 MB/s | 11.3× |
| ~20 ms | 1 MB | write | 1.31 MB/s | 0.25 MB/s | 5.2× |
| ~20 ms | 1 MB | read | 7.86 MB/s | 0.25 MB/s | 31.8× |
| ~20 ms | 8 MB | write | 1.68 MB/s | 0.25 MB/s | 6.7× |
| ~20 ms | 8 MB | read | 20.02 MB/s | 0.25 MB/s | 80.2× |
| ~20 ms | 32 MB | write | 1.69 MB/s | 0.25 MB/s | 6.7× |
| ~20 ms | 32 MB | read | 21.35 MB/s | 0.25 MB/s | 84.9× |

Benchmark source: https://gist.github.com/susnux/77ccaaf3b3e7b7779d7f9725ea22eb69

miaulalala (approved):

susnux:
Documentation in https://github.com/nextcloud/documentation/pull/15366

PR #48196, opened 2024-09-18
chore: Add proper deprecation dates where missing

susnux:
## Summary

Looked up deprecation date where missing.

## Checklist

### DIFF
--- a/lib/public/Share_Backend.php
+++ b/lib/public/Share_Backend.php
@@ -30,14 +30,14 @@ interface Share_Backend {
 	 * Get a unique name of the item for the specified user
 	 * @param string $itemSource
 	 * @param string|false $shareWith User the item is being shared with
-	 * @param array|null $exclude List of similar item names already existing as shared items @deprecated 7.0.0
 	 * @return string Target name
 	 *
 	 * This function needs to verify that the user does not already have an item with this name.
 	 * If it does generate a new name e.g. name_#
 	 * @since 5.0.0
+	 * @deprecated 31.0.0
 	 */
-	public function generateTarget($itemSource, $shareWith, $exclude = null);
+	public function generateTarget($itemSource, $shareWith);
 
 	/**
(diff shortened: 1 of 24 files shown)


juliusknorr (approved):

lib/public/Share_Backend.php
```diff
@@ -30,7 +30,7 @@ public function isValidSource($itemSource, $uidOwner);
 	 * Get a unique name of the item for the specified user
 	 * @param string $itemSource
 	 * @param string|false $shareWith User the item is being shared with
-	 * @param array|null $exclude List of similar item names already existing as shared items @deprecated since version OC7
+	 * @param array|null $exclude List of similar item names already existing as shared items @deprecated 7.0.0
```
come-nc, line 33:
```suggestion
	 * @param array|null $exclude List of similar item names already existing as shared items 
	 * @deprecated 7.0.0
```

come-nc (approved):

provokateurin (approved):
Can you also extend the regex to enforce this?
And please also backport what is possible.

lib/public/Share_Backend.php
susnux: I think only the property is deprecated?

provokateurin: That's not really possible. And having two annotations in the same line doesn't work syntax wise

come-nc: It’s never used in any application I have locally, nor in server.

come-nc: ping @icewind1991 @nickvergessen What do you think, was the whole method deprecated in OC7 or only the parameter? What do we do with it?

susnux: Original comment author was @nickvergessen in https://github.com/nextcloud/server/commit/dfde04291eae31765b2cb858e1d1260b60d75329

PR #48206, opened 2024-09-19
feat: make systemtags public visible

### LINKED ISSUE #40364: Show tags in public shares
Since the merge of #37938 (Great job!), another use case rised for me: the display of tags in public shares.

I am managing a list of sheet music (along with some records) for a choir, and make it available for all singers. Each song has its own directory, and now, I'm trying to make this list easier to browse for new people, by adding tags with year in it (as 2021 for a song we sang in 2021…). I find it far better than using sub-directories, either for discoverability and for consistency: since we can re-use an old song, it can be both in 2019 and 2023.

Anyway, for now, tags are not a public data, so I open this ticket to track this feature. If it doesn't meet project goals, it can be closed, but it will at least be there for the record.

### How to use GitHub

**Additional context**

skjnldsv:
fix https://github.com/nextcloud/server/issues/40364

image

Of course if the tag is made invisible in the settings, it will also apply.
I'm also thinking if it would make sense to NOT show restricted tags on public
image

### DIFF
--- a/apps/systemtags/lib/Listeners/BeforeSabrePubliclyLoadedListener.php
+++ b/apps/systemtags/lib/Listeners/BeforeSabrePubliclyLoadedListener.php
@@ -1,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\SystemTags\Listeners;
+
+use OCA\DAV\SystemTag\SystemTagPlugin;
+use OCP\BeforeSabrePubliclyLoadedEvent;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Server;
+
+/**
+ * @template-implements IEventListener<BeforeSabrePubliclyLoadedEvent>
+ */
+class BeforeSabrePubliclyLoadedListener implements IEventListener {
+	public function handle(Event $event): void {
+		if (!$event instanceof BeforeSabrePubliclyLoadedEvent) {
+			return;
+		}
+
+		$server = $event->getServer();
+		if ($server === null) {
+			return;
+		}
+
+		$server->addPlugin(Server::get(SystemTagPlugin::class));
+	}
+}
--- a/apps/systemtags/lib/Listeners/BeforeTemplateRenderedListener.php
+++ b/apps/systemtags/lib/Listeners/BeforeTemplateRenderedListener.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\SystemTags\Listeners;
+
+use OCA\Files_Sharing\Event\BeforeTemplateRenderedEvent;
+use OCA\SystemTags\AppInfo\Application;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Util;
+
+/**
+ * @template-implements IEventListener<BeforeTemplateRenderedEvent>
+ */
+class BeforeTemplateRenderedListener implements IEventListener {
+	public function handle(Event $event): void {
+		if (!$event instanceof BeforeTemplateRenderedEvent) {
+			return;
+		}
+		Util::addInitScript(Application::APP_ID, 'init');
+	}
+}
--- a/apps/systemtags/lib/Listeners/LoadAdditionalScriptsListener.php
+++ b/apps/systemtags/lib/Listeners/LoadAdditionalScriptsListener.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\SystemTags\Listeners;
+
+use OCA\Files\Event\LoadAdditionalScriptsEvent;
+use OCA\SystemTags\AppInfo\Application;
+use OCP\EventDispatcher\Event;
+use OCP\EventDispatcher\IEventListener;
+use OCP\Util;
+
+/**
+ * @template-implements IEventListener<LoadAdditionalScriptsEvent>
+ */
+class LoadAdditionalScriptsListener implements IEventListener {
+	public function handle(Event $event): void {
+		if (!$event instanceof LoadAdditionalScriptsEvent) {
+			return;
+		}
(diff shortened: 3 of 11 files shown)


lib/private/Tags.php
```diff
@@ -27,11 +27,6 @@ class Tags implements ITags {
...
 	private array $tags = [];
 
-	/**
-	 * Are we including tags for shared items?
-	 */
-	private bool $includeShared = false;
-
 	/**
 	 * The current user, plus any owners of the items shared with the current
 	 * user, if $this->includeShared === true.
```
susnux, line 32: Is this then not also always only including the real owner?
But in either way this comment makes no sense anymore with the changes

skjnldsv: yep, I'll remove it, misadded file

apps/files_sharing/lib/DefaultPublicShareTemplateProvider.php
```diff
@@ -102,6 +102,7 @@ public function renderPage(IShare $share, string $token, string $path): Template
 		\OCP\Util::addInitScript(Application::APP_ID, 'init-public');
 		\OCP\Util::addScript('files', 'main');
 		\OCP\Util::addStyle('files', 'merged');
+		\OCP\Util::addInitScript('systemtags', 'init');
```
skjnldsv, line 105: This should be moved to a proper listener

susnux: "before files loaded"?

skjnldsv: Yes that was a note for myself :see_no_evil:

apps/systemtags/lib/Listeners/BeforeTemplateRenderedListener.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @template-implements IEventListener<BeforeTemplateRenderedEvent>
+ */
+class BeforeTemplateRenderedListener implements IEventListener {
+	public function handle(Event $event): void {
+		if (!$event instanceof BeforeTemplateRenderedEvent) {
+			return;
+		}
+		Util::addInitScript(Application::APP_ID, 'init');
```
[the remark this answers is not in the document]
come-nc: case is SystemTags

apps/dav/lib/Connector/Sabre/ServerFactory.php
```diff
@@ -139,6 +123,12 @@ public function createServer(string $baseUri,
 			);
 			$server->addPlugin(new \OCA\DAV\Connector\Sabre\QuotaPlugin($view, true));
 			$server->addPlugin(new \OCA\DAV\Connector\Sabre\ChecksumUpdatePlugin());
+			$server->addPlugin(new SystemTagPlugin(
+				\OCP\Server::get(ISystemTagManager::class),
+				\OCP\Server::get(IGroupManager::class),
+				\OCP\Server::get(IUserSession::class),
+				\OCP\Server::get(ISystemTagObjectMapper::class),
+			));
```
come-nc, line 131: Please do that in the systemtags app so that it’s not registered if the app is disabled and we do not clutter the dav app.
See https://github.com/nextcloud/server/blob/8c60ffa0f21414e6e671c567d664a9b9e5253e26/apps/dav/lib/Connector/Sabre/ServerFactory.php#L183-L192

come-nc (changes requested):

apps/dav/lib/Connector/Sabre/ServerFactory.php
skjnldsv: Fixed! :+1:

come-nc (approved):

apps/dav/lib/Server.php
```diff
@@ -197,7 +197,7 @@ public function __construct(IRequest $request, string $baseUri) {
 		}
 
 		// system tags plugins
-		$this->server->addPlugin(\OC::$server->get(SystemTagPlugin::class));
+		// $this->server->addPlugin(\OCP\Server::get(SystemTagPlugin::class));
```
susnux, line 200: ?

apps/systemtags/lib/Listeners/LoadAdditionalScriptsListener.php
```diff
@@ -0,0 +1,26 @@
...
+
+declare(strict_types=1);
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\SystemTags\Listeners;
+
+use OCA\Files\Event\LoadAdditionalScriptsEvent;
+use OCA\Systemtags\AppInfo\Application;
```
susnux, line 11:
```suggestion
use OCA\SystemTags\AppInfo\Application;
```

lib/public/SystemTag/ISystemTagManager.php
```diff
@@ -112,7 +112,7 @@ public function deleteTags($tagIds);
 	 *
 	 * @since 9.1.0
```
susnux, line 113:
```suggestion
	 * @since 9.1.0
	 * @since 31.0.0 `$user` can be null to check anonymous permissions
```

Also the typing above does not fit anymore, should be:
```
* @param IUser|null $user user to check permission for
```

apps/dav/lib/Server.php
skjnldsv: Added in the `apps/systemtags/lib/Listeners/BeforeSabrePubliclyLoadedListener.php` now :)

skjnldsv:
Done @susnux

nickvergessen:
> I'm also thinking if it would make sense to NOT show restricted tags on public

There are 2 flags: can assign, can see. Restricted is false, true. So would be okay to show from that perspective. But I can see how guests add a new angle to it.
I would still show them and instead add a "Show tags on public pages" appconfig?

skjnldsv:
> "Show tags on public pages" appconfig?

I'd prefer to avoid adding yet another config :sob: 
Since we have public, I think it's straightforward enough to show it in public :shrug:

apps/systemtags/lib/Listeners/BeforeSabrePubliclyLoadedListener.php
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @template-implements IEventListener<BeforeSabrePubliclyLoadedEvent>
+ */
+class BeforeSabrePubliclyLoadedListener implements IEventListener {
+	public function handle(Event $event): void {
+		if (!$event instanceof BeforeSabrePubliclyLoadedEvent) {
+			return;
+		}
+		$event->getServer()->addPlugin(Server::get(SystemTagPlugin::class));
```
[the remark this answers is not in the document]
susnux: Maybe for psalm?

```suggestion
		$server = $event->getServer();
		assert($server !== null);
		$server->addPlugin(Server::get(SystemTagPlugin::class));
```

apps/dav/lib/Server.php
come-nc: What about `SabrePluginAddEvent`?

provokateurin (approved):

PR #48207, opened 2024-09-19
fix: Don't crash if disabled user is missing in the database

artonge:
If the user is missing in the DB, then [`getEMailAddress`](https://github.com/nextcloud/server/blob/eaaad6af8b1db4efc631e15a2762606b89564633/lib/private/User/LazyUser.php#L117-L118) will throw as user will be `null`.


come-nc (approved):

come-nc (changes requested):
Actually, please throw a NoUserException and catch that, using try/catch for a typeerror is a bad idea.

Altahrim (approved):

come-nc (approved):

PR #48210, opened 2024-09-19
ci: Install PHPUnit 10

nickvergessen:
## Checklist


nickvergessen:
Still missing:
## No DB
- [x] OCA\Encryption https://github.com/nextcloud/server/pull/53153
- [x] OCA\FederatedFileSharing https://github.com/nextcloud/server/pull/53179
- [x] OCA\Federation https://github.com/nextcloud/server/pull/53175
- [x] OCA\Files_Sharing

## DBs
- [x] OCA\FederatedFileSharing https://github.com/nextcloud/server/pull/53179
- [x] OCA\Federation https://github.com/nextcloud/server/pull/53175
- [x] OCA\Files https://github.com/nextcloud/server/pull/53179
- [x] OCA\Files_Sharing
- [x] OCA\Files_Trashbin https://github.com/nextcloud/server/pull/53175
- [x] Test\TextProcessing https://github.com/nextcloud/server/pull/53179

susnux (approved):

autotest.sh
```diff
@@ -22,8 +22,8 @@ if [ -z "$PHP_EXE" ]; then
 fi
 PHP=$(which "$PHP_EXE")
 if [ -z "$PHPUNIT_EXE" ]; then
-	if [ -f build/integration/vendor/bin/phpunit ]; then
-		PHPUNIT_EXE="./build/integration/vendor/bin/phpunit"
+	if [ -f lib/composer/bin/phpunit ]; then
+		PHPUNIT_EXE="./lib/composer/bin/phpunit"
```
susnux, line 26: no?

```suggestion
	if [ -f vendor-bin/phpunit/bin/phpunit ]; then
		PHPUNIT_EXE="./vendor-bin/phpunit/bin/phpunit"
```

nickvergessen: this was temporary cheating :upside_down_face:

nickvergessen: Ah no, I had fixed this already. We use `bamarni-bin` with linking binaries to the root composer:
https://github.com/nextcloud/server/blob/d21062c3f07ec892cf9a46cb0a9e4623e9d9668f/composer.json#L78-L79
And that makes them available in `lib/composer/bin`. All `vendor-bin/*/` don't have a `bin` directory.

susnux:
```diff
1) OCA\Files_External\Tests\Service\DBConfigServiceTest::testAddApplicable
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     0 => Array (
-        'type' => 3
-        'value' => 'test'
-        'mount_id' => 2
+        'type' => 1
+        'value' => null
+        'mount_id' => '2'
     )
     1 => [...]
     2 => Array (
-        'type' => 1
-        'value' => null
-        'mount_id' => 2
+        'type' => 3
+        'value' => 'test'
+        'mount_id' => '2'
     )
 )

/home/runner/actions-runner/_work/server/server/apps/files_external/tests/Service/DBConfigServiceTest.php:77
```


I wonder how this ever worked, the code does not sort (the query) so the results could be in random order.
I would suggest to just use `assertEqualsCanonicalizing`

nickvergessen:
Yeah or we sort afterwards, both should work.
But yeah I wonder the same. Feels too random, only thing I can see that the static providers are executed faster and therefore DB writes it in 1 flush to the disk. We had similar problems in Talk integration tests.

nickvergessen:
/backport f230abeef84faf4752c201ec7bf3d202939b8dae to stable31

nickvergessen:
/backport f230abeef84faf4752c201ec7bf3d202939b8dae to stable30

nickvergessen:
> [ci: Skip second S3 test temporarily](https://github.com/nextcloud/server/pull/48210/commits/f6931d79bbd65717be63a112b62c3d78fd8695ca)

- This is temporary and unchanged to before.
- PHPUnit 9 actually only accepted a single file and adding a group and executing the full group, also shows the error on PHPUnit 9. @susnux is proofing that in https://github.com/nextcloud/server/pull/53378 and we will tackle the problem there.

skjnldsv (approved):

PR #48219, opened 2024-09-19
fix(IStorage): Use false instead of bool intersection type to match implementations

provokateurin:
## Summary

This adds strong return types to all implementations to match the interface.
The changes escalated a bit, but everything should be fixed at once.
It also fixes some bugs in poorly implemented storages.

Technically this is a breaking change due to changing the signature, but I only did that for private classes and the fix is completely backwards compatible.
The temporary pain of adding the return types to all implementations is justified by fixing all these type mismatches for once.
This change should also allow us to actually add the strong return types to IStorage since all implementations should also be extending Common which already has them.
Please don't hate me for this change :see_no_evil:

I was not able to test files_external due to the terrible test setup which I can not get working locally, so let's see what CI has to say.

## Checklist

### DIFF
--- a/apps/files_external/lib/Lib/Storage/FTP.php
+++ b/apps/files_external/lib/Lib/Storage/FTP.php
@@ -78,15 +78,15 @@ class FTP extends Common {
 		return $this->connection;
 	}
 
-	public function getId() {
+	public function getId(): string {
 		return 'ftp::' . $this->username . '@' . $this->host . '/' . $this->root;
 	}
 
-	protected function buildPath($path) {
+	protected function buildPath($path): string {
 		return rtrim($this->root . '/' . $path, '/');
 	}
 
-	public static function checkDependencies() {
+	public static function checkDependencies(): array|bool {
 		if (function_exists('ftp_login')) {
 			return true;
 		} else {
@@ -94,7 +94,7 @@ class FTP extends Common {
 		}
 	}
 
-	public function filemtime($path) {
+	public function filemtime($path): int|false {
 		$result = $this->getConnection()->mdtm($this->buildPath($path));
 
 		if ($result === -1) {
@@ -135,7 +135,7 @@ class FTP extends Common {
 		}
 	}
 
-	public function rmdir($path) {
+	public function rmdir($path): bool {
 		if ($this->is_dir($path)) {
 			$result = $this->getConnection()->rmdir($this->buildPath($path));
 			// recursive rmdir support depends on the ftp server
@@ -153,7 +153,6 @@ class FTP extends Common {
 
 	/**
 	 * @param string $path
-	 * @return bool
 	 */
 	private function recursiveRmDir($path): bool {
 		$contents = $this->getDirectoryContent($path);
@@ -170,7 +169,7 @@ class FTP extends Common {
 		return $result;
 	}
 
-	public function test() {
+	public function test(): bool {
 		try {
 			return $this->getConnection()->systype() !== false;
 		} catch (\Exception $e) {
@@ -178,7 +177,7 @@ class FTP extends Common {
 		}
 	}
 
-	public function stat($path) {
+	public function stat($path): array|false {
 		if (!$this->file_exists($path)) {
 			return false;
 		}
@@ -188,14 +187,14 @@ class FTP extends Common {
 		];
 	}
 
-	public function file_exists($path) {
+	public function file_exists($path): bool {
 		if ($path === '' || $path === '.' || $path === '/') {
 			return true;
 		}
 		return $this->filetype($path) !== false;
 	}
 
-	public function unlink($path) {
+	public function unlink($path): bool {
 		switch ($this->filetype($path)) {
 			case 'dir':
 				return $this->rmdir($path);
@@ -211,14 +210,14 @@ class FTP extends Common {
 		return IteratorDirectory::wrap($files);
 	}
 
-	public function mkdir($path) {
+	public function mkdir($path): bool {
 		if ($this->is_dir($path)) {
 			return false;
 		}
 		return $this->getConnection()->mkdir($this->buildPath($path)) !== false;
 	}
 
-	public function is_dir($path) {
+	public function is_dir($path): bool {
 		if ($path === '') {
 			return true;
 		}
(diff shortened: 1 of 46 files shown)


apps/files_external/lib/Lib/Storage/SFTP.php
```diff
@@ -456,17 +400,16 @@ public function stat($path) {
...
+	public function constructUrl($path): string {
 		// Do not pass the password here. We want to use the Net_SFTP object
 		// supplied via stream context or fail. We only supply username and
 		// hostname because this might show up in logs (they are not used).
 		$url = 'sftp://' . urlencode($this->user) . '@' . $this->host . ':' . $this->port . $this->root . $path;
 		return $url;
 	}
 
-	public function file_put_contents($path, $data) {
+	public function file_put_contents($path, $data): float|bool|int {
```
come-nc, line 412:
```suggestion
	public function file_put_contents($path, $data): int|float|false {
```

apps/files_external/lib/Lib/Storage/SMB.php
```diff
@@ -628,7 +609,7 @@ public function getDirectoryContent($directory): \Traversable {
 		}
 	}
 
-	public function filetype($path) {
+	public function filetype($path): bool|string {
```
come-nc, line 612: !

apps/files_external/lib/Lib/Storage/Swift.php
```diff
@@ -92,7 +93,7 @@ private function normalizePath(string $path) {
 	 * @throws \OCP\Files\StorageAuthException
 	 * @throws \OCP\Files\StorageNotAvailableException
 	 */
-	private function fetchObject(string $path) {
+	private function fetchObject(string $path): bool|StorageObject {
```
come-nc, line 96: false

come-nc (commented):
To be continued…

provokateurin:
I hope I fixed all the weird typings, but please take another look.

apps/files_external/lib/Lib/Storage/FTP.php
```diff
@@ -78,23 +78,23 @@ protected function getConnection(): FtpConnection {
...
 		return 'ftp::' . $this->username . '@' . $this->host . '/' . $this->root;
 	}
 
-	protected function buildPath($path) {
+	protected function buildPath($path): string {
 		return rtrim($this->root . '/' . $path, '/');
 	}
 
-	public static function checkDependencies() {
+	public static function checkDependencies(): array|bool {
```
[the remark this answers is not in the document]
provokateurin: I'm not sure either, can just infer it from looking at the code. My guess would be:
`true` -> everything fine
`false` -> something is wrong
`array` -> list of missing dependencies

lib/private/Files/Storage/Common.php
```diff
@@ -526,30 +487,24 @@ protected function getFilenameValidator(): IFilenameValidator {
...
 		$this->mountOptions = $options;
 	}
 
 	/**
 	 * @param string $name
 	 * @param mixed $default
-	 * @return mixed
 	 */
-	public function getMountOption($name, $default = null) {
+	public function getMountOption($name, $default = null): mixed {
```
come-nc, line 498: mixed?

provokateurin: I think this is correct, as it accesses a key of the options and returns the value where it can be anything so mixed is necessary.

apps/files_external/lib/Lib/Storage/FTP.php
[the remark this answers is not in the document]
provokateurin: Not really useful to document at the moment because the implementations are not even based on a shared interface yet.

come-nc (approved):

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -511,7 +451,7 @@ protected function verifyUnencryptedSize(string $path, int $unencryptedSize): in
 	 *
 	 * @return int calculated unencrypted size
```
Altahrim, line 452:
```suggestion
	 * @return float|int calculated unencrypted size
```

Altahrim (approved):

lib/private/Files/Storage/Wrapper/Encryption.php
provokateurin: I also noticed a few other places where this probably affects 32-bit support (due to problems somewhere deeper in the code), @come-nc do you want to look into that?
Either way I would just do that as a follow-up and not in this PR.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -511,7 +451,7 @@ protected function verifyUnencryptedSize(string $path, int $unencryptedSize): in
...
 	 */
-	protected function fixUnencryptedSize(string $path, int $size, int $unencryptedSize): int {
+	protected function fixUnencryptedSize(string $path, int $size, int $unencryptedSize): float|int {
```
come-nc, line 454:
```suggestion
	protected function fixUnencryptedSize(string $path, int $size, int $unencryptedSize): int|float {
```
Please always use the order int|float to help with `grep` searches.

lib/private/Files/Storage/Wrapper/Encryption.php
come-nc: Well here you changed the return type so adapt the phpdoc version or remove it.
For the rest, I do not have time to allocate on 32bit int hunting. Do not hesitate to turn all sizes into `int|float` when you encounter them.
Maybe we’d need a test that fakes a 5GB file in encryption tests to make sure that works on 32bit CI.

nickvergessen:
This breaks all apps that have a storage wrapper.
Despite being in OC, it's de-facto standard and should be kept stable...

At least it should be documented

lib/private/Files/Storage/Common.php
```diff
@@ -823,7 +750,7 @@ public function writeStream(string $path, $stream, ?int $size = null): int {
 		return $count;
 	}
 
-	public function getDirectoryContent($directory): \Traversable {
+	public function getDirectoryContent($directory): \Traversable|false {
```
come-nc, line 753: I missed this change, can you explain why the `|false`?
There is no `return false` in there.

provokateurin: Ok I thought I remembered that one of the implementations returns false so every implementation has to handle it, but I was wrong. The only error handling I could find was done using exceptions, so I will revert this.

PR #48220, opened 2024-09-19
chore: Remove old travis related stuff

susnux:
Not used anymore


provokateurin (approved):
I think in the dav app tests there is still a folder called travis

juliusknorr (approved):

PR #48223, opened 2024-09-19
Add rector config

come-nc:
## Summary

And so it begins…

1. Added rector in a vendor-bin
2. Added a mostly-empty configuration
3. Applied to apps folder

Long term goal is to apply rectors from https://github.com/nextcloud-libraries/rector to core code. We’ll need to be extra careful with folders like lib/public because they use some deprecated stuff on purpose I think.
So not sure if the best way forward is to add more rules or more folders.

## Checklist

### DIFF
--- a/rector.php
+++ b/rector.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+use Rector\Config\RectorConfig;
+
+$config = RectorConfig::configure()
+	->withPaths([
+		__DIR__ . '/apps',
+		// __DIR__ . '/config',
+		// __DIR__ . '/core',
+		// __DIR__ . '/lib',
+		// __DIR__ . '/ocs',
+		// __DIR__ . '/ocs-provider',
+		// __DIR__ . '/tests',
+		// __DIR__ . '/themes',
+	])
+	->withSkip([
+		__DIR__ . '/apps/*/3rdparty/*',
+		__DIR__ . '/apps/*/build/stubs/*',
+		__DIR__ . '/apps/*/composer/*',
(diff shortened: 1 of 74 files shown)


rector.php
```diff
@@ -20,6 +20,12 @@
 		// __DIR__ . '/tests',
 		// __DIR__ . '/themes',
 	])
+	->withSkip([
+		__DIR__ . '/apps/*/3rdparty/*',
+		__DIR__ . '/apps/*/build/stubs/*',
+		__DIR__ . '/apps/*/composer/*',
+		__DIR__ . '/apps/*/config/*',
+	])
```
provokateurin, line 28: Please put this into the first commit

provokateurin (changes requested):
Can you add a script like https://github.com/nextcloud/groupfolders/blob/532f734b9d729a3c29a7371faf0bb85a24e0b311/composer.json#L26? This makes it a lot easier to use rector.

rector.php
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+use Rector\Config\RectorConfig;
+
+return RectorConfig::configure()
+	->withPaths([
+		__DIR__ . '/apps',
```
provokateurin, line 14: Make sure to exclude any git ignored paths like cs-fixer

provokateurin (changes requested):

provokateurin:
Uhm is it really supposed to run on 50323 files? To me it looks like it includes way too much.

come-nc:
> Uhm is it really supposed to run on 50323 files? To me it looks like it includes way too much.

1877 for me

provokateurin:
Maybe because I have my cloned apps in apps/ as well, and then it also scans all the vendors of those and so on.
When ignoring everything that git ignores it should work though.

provokateurin (approved):
Works, you only need to make the CI happy.

nickvergessen (approved):

Issue #39180, opened 2023-07-06
show file ownership of files for expired shares


Reis-A:
Right now, when a share expires the notification email looks like this

"There was some activity on Nextcoud
Today at 2:15am
Share for file .... expired"

It would be useful, if that notification also includes the actual owner of that file, since this is not always clear in a bigger nextcloud with many users...

Issue #20869, opened 2020-05-08
inform user about link expiration and allow restoring a link


kolAflash:
Currently a user may not notice when a link is going to expire.
Either because the user forgot, or because the admin set a global expiration default and the user didn't noticed that at all.
This is especially a problem, if a link has been laboriously spread over multiple channels (newsletters, Twitter, ...).
&nbsp;

So the user should be better informed about the expiration of a link.
At first when the link is being created.
And maybe also later, before the link is expiring (e.g. sending an email one day before expiration).
&nbsp;

Additionally it could be helpfull, if there's a possibility to restore an expired link.
At least for a certain time after it's been expired.

Maybe a link should be restoreable for the same time it was active. So if the link was active for six month, it should stay restoreable for another six month.

dajusc:
Meanwhile in NC21 there is an email notification after a link is expired, which is only of limited help.
As @kolAflash proposed it would be very helpful to have:

1. A notification _before_ the link expires
1. A possibility to restore an expired link

Flo-MF:
There is a OCC command available, to send a notification about expiring shares, which you can run by a cronjob:

occ sharing:expiration-notification

https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/file_sharing_configuration.html

joshtrichards:
Adjusted the title to focus on the "restoring a shared link" part of the enhancement request, since sending a notification before a share expires is already a possibility today.

salmart-dev:
Thank you for reporting this user experience issue.

It seems like there is no possibility to re-create a share once it expires as it gets deleted once the system detects it is expired. As a workaround, it should be possible to edit the expiration date of an existing share before it expires, although one would need to keep track of those important shares to do this, which can be laborious.

We will consider implementing this feature in the future depending on the amount of votes it gets 🔧

dexterdeluxe88:
As a user, I want to "re-enable" the expired link, since I have already shared it. I dont care if it re-enables the link or creates a new share but with the same URL under the hood

PhilLab:
Since Nextcloud 31 (Hub 10) you can achieve this by customizing the share link token ([PR](https://github.com/nextcloud/server/pull/49317)):
1. In the Admin settings -> Sharing , allow custom share tokens
2. In the web UI, create a _new_ share on the same file your previous share pointed to
3. Click on its options, where you can select to customize the sharing link
4. Set the token to the token your expired link had.

[...]

Issue #15819, opened 2019-05-31
Show expiration date on receiving shares


skjnldsv:
On the expiration date, we could show a message stating how many time is left.

MorrisJobke:
For internal shares this should go into the sidebar right? For the public shares it should be at a similar location as the note already is or do you have a better idea?

skjnldsv:
I was indeed thinking about user shares, like having a small calendar/time icon and a popup with the remaining time :)
No idea on the public shares! @nextcloud/designers ?

jancborchardt:
Would agree with @MorrisJobke this should go in the sidebar if anything. And for public shares it should also go into the to-be-added sidebar (for comments etc).

skjnldsv:
Something like that? :)
![Capture d’écran_2019-05-31_13-28-54](https://user-images.githubusercontent.com/14975046/58703167-11ed0080-83a8-11e9-853d-fa7e825ff2ab.png)

jancborchardt:
Ah ok, you mean for the sender?

In that case it should either be a subline (like the "Connect items to a project …" is for "Add to a project") or the icon should overlay the 3-dot menu so when you click, the relevant controls are opened. (Or both.)

skjnldsv:
Nono, the mockup is wrong since it comes from the sender part, you're right.
I mean for the receiving share :)

jancborchardt:
Ok, then how about my suggestion above for the sender part, and this for recipient:
![sharing expiry](https://user-images.githubusercontent.com/925062/58704214-883f3200-83ab-11e9-99d1-2fefd57c2a74.png)

skjnldsv:
Awesome!

tobiasKaminsky:
This should then also be done on clients, or?

skjnldsv:
Ideally yes

skjnldsv:
![Capture d’écran_2019-07-12_12-52-45](https://user-images.githubusercontent.com/14975046/61123166-f6474080-a4a3-11e9-9f71-4741d8aa602f.png)

tobiasKaminsky:
I think it should either be 
"expires in 18 days" as in "Folder xy share expires in 18 days"
or
"will expire in 18 days" as in "Folder xy share will expire in 18 days"

tobiasKaminsky:
BTW: how do you get the 18 days info?

skjnldsv:
[...]

Issue #14538, opened 2019-03-05
extend enforced sharing expiration date


wiswedel:
As a member of an entitled user group, I want to be allowed to extend an enforced share expiration date by another amount of the originally enforced timespan.

Example:
- default enforced expiration date: 7 days
- share a file on 1st March > expires on 8th March
- decide to extend the expiration on 7th March > new expiration date: 14th March

Background: Without an extension possibility, I'd have to create and send out a new link.
Permission to extend should be limited to certain groups.

Related issue: #10776

nickvergessen:
> Permission to extend should be limited to certain groups.

Should be all with share permissions, otherwise you just force people again to do the workaround you described already.

szaimen:
I am going to close this since there doesn't seem to be a lot of interest (no upvotes) and no respond since around 2 years. Please reopen if you still want to implement this.

PR #48225, opened 2024-09-19
feat(files_sharing): allow not deleting expiring shares

### LINKED ISSUE #17996: Keep list of expired public shares
I'm not sure if I've got this 100% right, but it seems like the `XX_share` table in the database is being pruned to remove public links that have 'expired'. I can understand the reasoning for this (reduce the size of the `XX_share` table), however the negative usability impact is as follows:

1. I share a public link with Person A outside my organisation, link expires in 30 days
2. After 40 days, Person A gets back to me saying they've only *just* got round to checking my link, and it's expired. They are presented with a fairly unhelpful message `The document could not be found on the server. Maybe the share was deleted or has expired?` (or see image below):

Screenshot 2019-11-19 at 16 55 45

3. I decide to re-share the folder with Person A, but I can't remember which folder I originally shared. Was it folder `Super Secret No.1` or `Super Secret No.2`?

**Solution**

I think public links should not be removed from `XX_share`, or if pruning is *really* necessary, they should be put in a `XX_share_expired` table. Following this:
a) I can now see expired links in the 'shares' section of my account, and restore links if need be (that would be to alter the expiration date to 'some time in the future'
b) We can change the unhelpful message `The document could not be found on the server. Maybe the share was deleted or has expired?` to something like `Access to the document you are trying to access has expired` - since we know it's not a 'deleted' issue.

Other (potential) benefits:
1. Helps avoid token collisions (Ok, the likelihood of this every happening is tiny):
    a) I share a folder with `Person A` for 30 days. Share token is ABCD
    b) Link expires and token ABCD is removed
    c) I share another folder with `Person B`. Share token is ABCD 
    d) There's a chance that `Person A` might access the folder `Person B` by mistake.
2. If I'm logged in and access the expired URL, Nextcloud can tell me "The file XXXX you previously shared expired on YYYY. Do you want to set a new expiration date? [big button for set new expiration date]"

Server: Nextcloud 16.0.4, running on CentOS/Apache (LAMP)

skjnldsv:
Fix https://github.com/nextcloud/server/issues/17996

## Feature
Shares that expires are just deleted. Which leads to confusion.
- We don't even know if the share existed or not....
- With the new file request, we might want to extend the drop period time...
- We need more transparency

### Solution
- Allow admins to disable share deletion on expiration (opt-in)
  
- Adds a new view to list all expired shares
image

### Status
- [x] Add admin setting
- [x] Add api controller
- [x] Add new view
- [ ] Add integration tests


skjnldsv:
All green and ready :rocket:

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -62,388 +59,47 @@
...
-class ShareAPIController extends OCSController {
+class ShareAPIController extends ShareApiControllerFactory {
 
 	private ?Node $lockedNode = null;
-	private string $currentUser;
 
-	/**
-	 * Share20OCS constructor.
-	 */
 	public function __construct(
```
come-nc, line 66: This constructor is empty, you can remove it

apps/files_sharing/lib/Controller/ExpiredShareAPIController.php
```diff
@@ -0,0 +1,95 @@
...
+		protected IGroupManager $groupManager,
+		protected IRootFolder $rootFolder,
+		protected IAppManager $appManager,
+		protected ContainerInterface $serverContainer,
+		protected UserStatusManager $userStatusManager,
+		protected IPreview $previewManager,
+		protected IDateTimeZone $dateTimeZone,
+		protected IURLGenerator $urlGenerator,
+		protected IL10N $l,
+		protected LoggerInterface $logger,
```
come-nc, line 49:
```suggestion
		ShareManager $shareManager,
		string $userId,
		IUserManager $userManager,
		IGroupManager $groupManager,
		IRootFolder $rootFolder,
		IAppManager $appManager,
		ContainerInterface $serverContainer,
		UserStatusManager $userStatusManager,
		IPreview $previewManager,
		IDateTimeZone $dateTimeZone,
		IURLGenerator $urlGenerator,
		IL10N $l,
		LoggerInterface $logger,
```
You are redeclaring properties from parent class, it’s better to simply let the parent class declare them and simply pass the parameters to the constructor.

apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -8,25 +8,25 @@
...
-	private $db;
-
-	public function __construct(ITimeFactory $time, IManager $shareManager, IDBConnection $db) {
-		$this->shareManager = $shareManager;
-		$this->db = $db;
+	public function __construct(
+		ITimeFactory $time,
+		private IManager $shareManager,
+		private IDBConnection $db,
+		private IAppConfig $config,
```
come-nc, line 27:
```suggestion
		private IAppConfig $appConfig,
```
Consistent naming for those helps reading the code without having to check the constructor for each class

apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -42,13 +42,16 @@ public function __construct(ITimeFactory $time, IManager $shareManager, IDBConne
 	 * @param array $argument unused argument
 	 */
 	public function run($argument) {
-		//Current time
+		if ($this->config->getValueString('core', 'shareapi_delete_on_expire', 'yes') === 'no') {
```
come-nc, line 45: This could be typed as boolean, no? I think @ArtificialOwl has handling for string to boolean conversion especially for this situation to be able to type as boolean stuff which was stored as yes/no before IAppConfig.

lib/private/Share20/Manager.php
```diff
@@ -1143,7 +1143,7 @@ public function getSharesInFolder($userId, Folder $node, $reshares = false, $sha
 	/**
 	 * @inheritdoc
 	 */
-	public function getSharesBy($userId, $shareType, $path = null, $reshares = false, $limit = 50, $offset = 0) {
+	public function getSharesBy($userId, $shareType, $path = null, $reshares = false, $limit = 50, $offset = 0, $expired = false) {
```
come-nc, line 1146: `$expired` is not part of the public interface so you should add an @param to document it in the private class. To make it clear whether the meaning is "give me only expired shares" or "give me shares including expired shares".
As an alternative I’d be also fine to rename it to $allowExpired on all methods as that is self-explanatory to me.

come-nc (changes requested):
Reviewed PHP changes only

apps/files_sharing/appinfo/routes.php
```diff
@@ -139,6 +139,14 @@
...
 			'verb' => 'POST',
 		],
+		/*
+		 * Expired Shares
+		 */
+		[
+			'name' => 'ExpiredShareAPI#index',
+			'url' => '/api/v1/expiredshares',
+			'verb' => 'GET',
+		],
```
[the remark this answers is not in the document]
skjnldsv: Yeah, I thought about this too.
i think it's best to migrate the full app to attributes directly, but mixing the two kind of seemed counter intuitive. Especially when looking for the source of the route declaration :shrug:

apps/files_sharing/lib/Controller/ExpiredShareAPIController.php
```diff
@@ -0,0 +1,95 @@
...
+	public function index(): DataResponse {
+		$groupShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_GROUP, null, -1, 0);
+		$roomShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_ROOM, null, -1, 0);
+		$deckShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_DECK, null, -1, 0);
+		$sciencemeshShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_SCIENCEMESH, null, -1, 0);
+		$linkShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_LINK, null, -1, 0);
+		$userShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_USER, null, -1, 0);
+		$emailsShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_EMAIL, null, -1, 0);
+		$circlesShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_CIRCLE, null, -1, 0);
+		$remoteShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_REMOTE, null, -1, 0);
```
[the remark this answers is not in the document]
skjnldsv: I re-used the behaviour of other ShareAPIControllers. I figured I preferred to stay consistent :)

apps/files_sharing/lib/ExpireSharesJob.php
skjnldsv: It's also to stay consistent with how we store the share settings.
You can see the declaration here https://github.com/nextcloud/server/blob/091fccae79ded370458bb73ace075b7005f8bc8d/apps/settings/lib/Settings/Admin/Sharing.php#L40-L75

apps/files_sharing/lib/Controller/ShareAPIController.php
skjnldsv: No, `IConfig`, `IProviderFactory` and `IMailer` are not in the `ShareApiControllerFactory`

come-nc: Sorry, was confused by the redeclaration of parent properties and thought there was only those.

apps/files_sharing/lib/Controller/DeletedShareAPIController.php
```diff
@@ -15,153 +15,65 @@
...
-
-			try {
-				$result = array_merge($result, $this->getSciencemeshShareHelper()->formatShare($share));
-			} catch (QueryException $e) {
-			}
-		}
-
-		return $result;
+		protected ShareManager $shareManager,
+		protected string $userId,
```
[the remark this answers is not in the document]
provokateurin: Yes, needs to be nullable!

skjnldsv:
Addressed @come-nc

apps/files_sharing/lib/Controller/ShareApiControllerFactory.php
```diff
@@ -0,0 +1,628 @@
...
+use OCP\Share\IShare;
+use OCP\UserStatus\IManager as UserStatusManager;
+use Psr\Container\ContainerExceptionInterface;
+use Psr\Container\ContainerInterface;
+use Psr\Log\LoggerInterface;
+
+/**
+ * @psalm-import-type Files_SharingShare from ResponseDefinitions
+ */
+abstract class ShareApiControllerFactory extends OCSController {
```
come-nc, line 39:
```suggestion
abstract class ShareApiControllerBase extends OCSController {
```
I’m still unhappy with the name, this is not a factory.
See https://en.wikipedia.org/wiki/Factory_method_pattern
A Factory has static methods to create instances. This here is simply a base to be used as parent by other controllers.

apps/files_sharing/lib/Controller/DeletedShareAPIController.php
```diff
@@ -15,154 +15,66 @@
...
+		protected IGroupManager $groupManager,
+		protected IRootFolder $rootFolder,
+		protected IAppManager $appManager,
+		protected ContainerInterface $serverContainer,
+		protected UserStatusManager $userStatusManager,
+		protected IPreview $previewManager,
+		protected IDateTimeZone $dateTimeZone,
+		protected IURLGenerator $urlGenerator,
+		protected IL10N $l,
+		protected LoggerInterface $logger,
```
come-nc, line 53:
```suggestion
		ShareManager $shareManager,
		?string $userId,
		IUserManager $userManager,
		IGroupManager $groupManager,
		IRootFolder $rootFolder,
		IAppManager $appManager,
		ContainerInterface $serverContainer,
		UserStatusManager $userStatusManager,
		IPreview $previewManager,
		IDateTimeZone $dateTimeZone,
		IURLGenerator $urlGenerator,
		IL10N $l,
		LoggerInterface $logger,
```

apps/files_sharing/lib/Controller/ExpiredShareAPIController.php
```diff
@@ -0,0 +1,96 @@
...
+			$appManager,
+			$serverContainer,
+			$userStatusManager,
+			$previewManager,
+			$dateTimeZone,
+			$urlGenerator,
+			$l,
+			$logger,
+		);
+	}
```
come-nc, line 67:
```suggestion
```

This one is really empty or did I miss something?

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -65,388 +61,47 @@
...
+		protected IConfig $config,
+		protected IAppManager $appManager,
+		protected ContainerInterface $serverContainer,
+		protected IUserStatusManager $userStatusManager,
+		protected IPreview $previewManager,
+		protected IDateTimeZone $dateTimeZone,
+		protected LoggerInterface $logger,
+		protected IProviderFactory $factory,
+		protected IMailer $mailer,
+		string $userId,
```
come-nc, line 85:
```suggestion
		IManager $shareManager,
		IGroupManager $groupManager,
		IUserManager $userManager,
		IRootFolder $rootFolder,
		IURLGenerator $urlGenerator,
		IL10N $l,
		protected IConfig $config,
		IAppManager $appManager,
		ContainerInterface $serverContainer,
		IUserStatusManager $userStatusManager,
		IPreview $previewManager,
		IDateTimeZone $dateTimeZone,
		LoggerInterface $logger,
		protected IProviderFactory $factory,
		protected IMailer $mailer,
		?string $userId,
```

Ideally I would also reorder to keep promoted properties together, I would say at the end (so, start with parameters from parent constructor in the same order as parent, add you own after)

apps/files_sharing/lib/Controller/ShareApiControllerFactory.php
```diff
@@ -0,0 +1,628 @@
...
+		protected ContainerInterface $serverContainer,
+		protected UserStatusManager $userStatusManager,
+		protected IPreview $previewManager,
+		protected IDateTimeZone $dateTimeZone,
+		protected IURLGenerator $urlGenerator,
+		protected IL10N $l,
+		protected LoggerInterface $logger,
+	) {
+		parent::__construct(Application::APP_ID, $request);
+		$this->currentUser = $userId;
```
come-nc, line 61:
```suggestion
```

Is there any added value to duplicate $this->userId into $this->currentUser?
If this is to have a clearer name, then remove the protected in front of $userId parameter to not turn it into a property, and keep $this->currentUser.

apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -42,13 +42,16 @@ public function __construct(ITimeFactory $time, IManager $shareManager, IDBConne
 	 * @param array $argument unused argument
 	 */
 	public function run($argument) {
-		//Current time
+		if ($this->appConfig->getValueString('core', 'shareapi_delete_on_expire', 'yes') === 'no') {
```
come-nc, line 45: This is still frightening to me because I’m afraid it means we can never change the type of this key to boolean later, because IAppConfig will not let you read as string a value stored as a boolean I think. To be confirmed by @ArtificialOwl

susnux: Probably we need to migrate all `yes` `no`, `1` `0` to proper booleans... So right after the hooks -> events migration ;)

come-nc: Yes, but you can do that for untyped values, but not for the ones typed as string I think

skjnldsv: But that is unrelated to this pr then 😅

apps/files_sharing/lib/Controller/ShareApiControllerFactory.php
skjnldsv: It's here because it was used by the controllers before this pr. 
I want to keep changes to the existing parts to a minimal.

joshtrichards:
Other issues this may address (did not link since I wasn't certain, but do so if appropriate please :smile:  ): 

- #20869
- #14538

Related too:

- #39180
- #15819

apps/files_sharing/lib/Controller/ShareApiControllerFactory.php
skjnldsv: Thanks, I need to take that PR back :see_no_evil:

apps/files_sharing/lib/Controller/ExpiredShareAPIController.php
```diff
@@ -0,0 +1,96 @@
...
+		IGroupManager $groupManager,
+		IRootFolder $rootFolder,
+		IAppManager $appManager,
+		ContainerInterface $serverContainer,
+		UserStatusManager $userStatusManager,
+		IPreview $previewManager,
+		IDateTimeZone $dateTimeZone,
+		IURLGenerator $urlGenerator,
+		IL10N $l,
+		LoggerInterface $logger,
```
nfebe, line 49: Where are these properties and their access modifiers defined?

apps/files_sharing/lib/Controller/ExpiredShareAPIController.php
```diff
@@ -0,0 +1,235 @@
...
+	public function index(): DataResponse {
+		$groupShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_GROUP, null, -1, 0);
+		$roomShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_ROOM, null, -1, 0);
+		$deckShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_DECK, null, -1, 0);
+		$sciencemeshShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_SCIENCEMESH, null, -1, 0);
+		$linkShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_LINK, null, -1, 0);
+		$userShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_USER, null, -1, 0);
+		$emailsShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_EMAIL, null, -1, 0);
+		$circlesShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_CIRCLE, null, -1, 0);
+		$remoteShares = $this->shareManager->getExpiredShares($this->userId, IShare::TYPE_REMOTE, null, -1, 0);
```
nfebe, line 86: No way to get all? `shareType` -1, could be default parameter in share Manager that returns all?

nfebe, line 86: No way to get all? `shareType` -1, could be default parameter in share Manager that returns all?

apps/files_sharing/lib/ExpireSharesJob.php
```diff
@@ -42,13 +41,16 @@ public function __construct(ITimeFactory $time, IManager $shareManager, IDBConne
 	 * @param array $argument unused argument
 	 */
 	public function run($argument) {
-		//Current time
+		if ($this->config->getValueString('core', 'shareapi_delete_on_expire', 'yes') === 'no') {
```
nfebe: I believe this would be a thing for config.php... We should

- Update `config.sample.php` that is an in code source documentation
- Document this somewhere the administration manual

apps/files_sharing/lib/Controller/DeletedShareAPIController.php
```diff
@@ -15,154 +15,66 @@
...
-	private $rootFolder;
-
-	/** @var IAppManager */
-	private $appManager;
-
-	/** @var IServerContainer */
-	private $serverContainer;
-
-	public function __construct(string $appName,
+class DeletedShareAPIController extends ShareApiControllerFactory {
```
nfebe, line 37: Nitpick : Is this `BaseShareApiController` or Factory? Factories are to help easy object creation, testing, decoupling not exactly DRY-ing?

PR #6315, opened 2017-08-30
Implement metadata for email templates of remaining emails

MorrisJobke:
Followup to #6255

@oparoz FYI

@MariusBluem for documentation later

PR #48226, opened 2024-09-19
fix(caldav): Add default reminder to attendees of scheduling messages

lufer22:
* Resolves: Calendar [#6315](https://github.com/nextcloud/calendar/issues/6315)

## Summary
The attendees of new scheduled events now have the attendee default reminder set.

## TODO

- [ ] ...

## Checklist


apps/dav/lib/CalDAV/Schedule/Plugin.php
```diff
@@ -444,6 +445,159 @@ public function propFindDefaultCalendarUrl(PropFind $propFind, INode $node) {
...
+
+		$home = $this->server->tree->getNodeForPath($homePath);
+		$inbox = $this->server->tree->getNodeForPath($inboxPath);
+
+		$currentObject = null;
+		$objectNode = null;
+		$oldICalendarData = null;
+		$isNewNode = false;
+
+		$userDefaultReminder = $this->config->getUserValue($this->stripOffMailTo($iTipMessage->recipient), 'calendar', 'defaultReminder', 'none');
```
tcitworld, line 540: `$iTipMessage->recipient` should be an email, not an userid, so this won't work unless you're using emails as userids

miaulalala:
Nice one, I will take a look soon!

SebastianKrupinski:
Morning @lufer22

Thanks for you effort on this! The code looks good but I do see a logic issue.

- Firstly since you are completely overriding the parent class code I would keep the parent code in the same function name (scheduleLocalDelivery), instead of a new function (scheduleLocalDeliveryHandler), this is minor and not a big deal.

- The bigger problem I see is that you are searching the users by email address. The problem here is, that email address in NC are NOT unique. So you can have multiple users with the same email address. So the logic would break if two user accounts have the same address, and the code would use the wrong defaults.

I will see if there is another option to this, this weekend when I have more time.

ndo84bw:
Hi @lufer22, hope you're well.

Since this PR has been waiting on @SebastianKrupinski's 2024-10-04 review feedback (the `userManager->getByEmail()` resolution being non-unique) for ~20 months without a follow-up from you, I went ahead and opened a successor PR that builds directly on the structure you laid out: nextcloud/server#61162. Your commits are kept in spirit and you're credited with a `Co-Authored-By` trailer on every commit there.

What the successor PR keeps from your work:

* The general approach of injecting a default VALARM during `scheduleLocalDelivery` for incoming REQUESTs.
* The `secondsToIso8601Duration` helper and the `createAlarm` shape.

What it changes / extends:

* Resolves the recipient via `$aclPlugin->getPrincipalByUri($iTipMessage->recipient)` instead of `userManager->getByEmail()`, which addresses Sebastian's blocking review (email is not unique across users).
* Keeps the original `scheduleLocalDelivery` method name (per Sebastian's other comment) instead of forking it as `scheduleLocalDeliveryHandler` - the new logic sits before the parent call as a `applyRecipientReminderPolicy()` hook, so we don't duplicate Sabre's parent.
* Reads the calendar app's typed defaults (`defaultReminderPartDay` / `defaultReminderFullDay`) with the legacy `defaultReminder` and the app-level `calendar.defaultReminder` as cascading fallbacks, instead of just the single legacy key.
* Adds a per-user toggle `calendar:applyDefaultReminderToInvitations` (defaulting to `yes`) so users who don't want it can opt out and admins who want a different fleet default can set the app value.
* Adds a second, related fix in the same code path: preserving the recipient's manually-added VALARMs across organizer updates (per `RECURRENCE-ID`, master and overrides individually). Without this, every organizer update on a recurring series would discard the recipient's reminders.
* Fixes a pre-existing bug where the strip block only operated on the first VEVENT - recurring REQUESTs with overrides retained organizer VALARMs on the overrides; this PR strips per-component.
* Adds unit-test coverage (was missing in the original PR per Sebastian's checklist).

Not closing this PR - would prefer to leave that decision to the maintainers so your contribution stays discoverable for context. If you ever want to come back to the area, the design rationale lives in the successor PR's description.

Thanks for getting the conversation started here in 2024.

PR #48235, opened 2024-09-20
test: re-add object store primary storage phpunit tests

icewind1991:
This was removed with #41003 but the object store related tests that were migrated there are not equivalent.

- re-add tests
- enable "preserve fileid when moving" test for object store
- rework cross-storage move with object store. *again*
- misc test fixes

### DIFF
--- a/.github/workflows/phpunit-object-store-primary.yml
+++ b/.github/workflows/phpunit-object-store-primary.yml
@@ -1,0 +1,121 @@
+# SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+name: PHPUnit primary object store
+on:
+  pull_request:
+  schedule:
+    - cron: "15 2 * * *"
+
+concurrency:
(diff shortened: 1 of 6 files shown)


provokateurin (approved):

.github/workflows/phpunit-object-store-primary.yml
```diff
@@ -0,0 +1,121 @@
+# SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+# SPDX-License-Identifier: MIT
+name: PHPUnit primary object store
```
susnux, line 3: This is what object-storage-s3.yml should do so I would not add a new one but just enable more tests there then.

susnux: So this is more of an integration test?

But what I meant is why not just add this as a step to the other workflow. Then one would only need to configure all services once and less work if something changes.

susnux (approved):
makes sense  :)

icewind1991:
backport for 28 has been merged into https://github.com/nextcloud/server/pull/48213

PR #48248, opened 2024-09-20
feat: Add check for enterprise

Pytal:
## Summary

Part of https://github.com/nextcloud/server/issues/45440

Add a simple check to check if the current instance is on the enterprise edition

## Checklist

### DIFF
--- a/lib/public/Util.php
+++ b/lib/public/Util.php
@@ -74,6 +74,14 @@ class Util {
 	}
 
 	/**
+	 * @since 31.0.0
+	 */
+	public static function isEnterprise() {
+		$versionString = Server::get(ServerVersion::class)->getVersionString();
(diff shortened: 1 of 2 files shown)


skjnldsv (approved):

lib/public/Util.php
```diff
@@ -73,6 +73,14 @@ public static function getChannel() {
 		return \OCP\Server::get(ServerVersion::class)->getChannel();
 	}
 
+	/**
+	 * @since 31.0.0
+	 */
+	public static function isEnterprise() {
```
juliushaertl, line 79: I'd rather add it to the ServerVersion class directly to not have more static methods in this generic utility class.

juliusknorr (changes requested):

PR #48277, opened 2024-09-23
new occ commands to manage system-tags for files

### LINKED ISSUE #32735: Extend occ (tags) with commands to add and remove tags to/from files
I'm always frustrated when trying to deal with flags from a shell script. It's a major pain to deal with several curl lines when all I need would be an 

`occ tag:file:add file tag[,tag]`

and 

`occ tag:file:remove file tag[,tag]`

It can be done with curl (and yes, I've read [https://help.nextcloud.com/t/is-there-a-way-to-add-tags-per-client-api/68849](https://help.nextcloud.com/t/is-there-a-way-to-add-tags-per-client-api/68849)) but that seems to be a bit over the top just because an administrator tries to clean up after a user.

**Additional context**

schaarsc:
* Resolves: #32735

## Summary
adds the following commands to the files app:

- `files:tag-add <target> <tags> <access>`
- `files:tag-delete <target> <tags> <access>`
- `files:tag-delete-all <target>`

## TODO

- [ ] 

## Checklist


apps/files/lib/Command/TagAdd.php
```diff
@@ -0,0 +1,87 @@
...
+		parent::__construct();
+	}
+
+	protected function configure(): void
+	{
+		$this->setName('files:tag-add')
+			->setDescription('Add a system-tag to a file or folder')
+			->addArgument('target', InputArgument::REQUIRED, "file id or path")
+			->addArgument('tags', InputArgument::REQUIRED, "Name of the tag(s) to add, comma separated")
+			->addArgument('access', InputArgument::REQUIRED, 'access level of the tag (public, restricted or invisible)');
```
come-nc, line 32: It feels weird that access is required for assigning an existing tag

apps/files/lib/Command/TagAdd.php
```diff
@@ -0,0 +1,87 @@
+<?php
+declare(strict_types = 1);
+/**
+ * SPDX-FileCopyrightText: Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+namespace OCA\Files\Command;
```
come-nc, line 7: These commands should be in the systemtags application, not files.

apps/files/lib/Command/TagAdd.php
schaarsc: I think it is required because 
- the same tagName can exist with different access level
- I want to work around the missing exif support (https://github.com/nextcloud/photos/issues/226) by automatically creating the tags if does not exist

apps/files/lib/Command/TagAdd.php
schaarsc: sure, I can move the commands to tag, if you prefer that.
since the objectType is hardcoded to 'files', I'd suggest
- tag:files:add
- tag:files:delete
- tag:files:delete-all

if that is ok

apps/files/lib/Command/TagAdd.php
schaarsc: please replace EXIF with metadata/xmp...

I'm using the exiftool to extract  metadata.  The tool is able to extract the XMP information and TagsList is part of XMP, not EXIF

ExifMetadataProvider does indeed store the EXIF part of the metadata, but this only includes things like GPS, Time, size,...

apps/files/lib/Command/TagAdd.php
joshtrichards: Commenting to generate a notification since @schaarsc may not have seen the upvotes to their query. ;-)

schaarsc: those changes have been implemented and are part of 819385c

apps/systemtags/lib/Command/Files/DeleteAll.php
```diff
@@ -0,0 +1,49 @@
...
+use OCP\SystemTag\ISystemTagObjectMapper;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class DeleteAll extends Command {
+
+	public function __construct(
+		private FileUtils $fileUtils,
```
marcelklehr, line 22: I'm not sure if FileUtils is recommended to be used in apps even if they're in server

marcelklehr (approved):
Looks good

come-nc (approved):

marcelklehr:
@schaarsc Psalm is still failing with an error talking about a wrong argument type

marcelklehr:
How do we merge since CI fails because it doesn't want to run on forks?

come-nc:
> How do we merge since CI fails because it doesn't want to run on forks?

Ask a team lead to force-merge

PR #48295, opened 2024-09-23
Correctly look up filename when logged in but inaccessible

### LINKED ISSUE #2067: Failed to export Submissions: Not allowed to write to file
**Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are affected by the same issue. Please don't comment if you have no relevant information to add!**

**Describe the bug**

A Nextcloud instance that did have a working Forms app now generates an error on form submission.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the form
2. Complete it
3. Click _"Submit"_
5. The use has the error _"There was an error submitting the form"_

**Expected behavior**

A successfully submission. 

**Nextcloud (please complete the following information):**
- Nextcloud-Version: 28.0.4.1
- Forms-Version: 4.2.0

**Desktop (please complete the following information):**
 - OS: Debian
 - Browser Firefox

**Smartphone (please complete the following information):**
 - Browser Firefox/
 - Version 126.0

**Browser log**
```
[ERROR] forms: Error while submitting the form 
Object { app: "forms", level: 2, error: {…} }
NcSettingsSection-NSrCZ23G-6nKKlVz9.mjs:2:337806
    value https://cloud.example.org/apps/forms/js/NcSettingsSection-NSrCZ23G-6nKKlVz9.mjs:2
    value https://cloud.example.org/apps/forms/js/NcSettingsSection-NSrCZ23G-6nKKlVz9.mjs:2
    onConfirmedSubmit https://cloud.example.org/apps/forms/js/Submit-PS6lxYc-.mjs:2
```

**Nextcloud log**

The following is written to the Nextcloud log:

> Failed to export Submissions: Not allowed to write to file

**Server logs**

There is nothing in the Apache or PHP-FPM logs.

**TMPDIR**

The `tempdirectory` is writable.

**Additional context**

The error that is produced appears to be here:

https://github.com/nextcloud/forms/blob/1ba3a1404c3a23eedd8010175b60f10c5a70f688/lib/Service/SubmissionService.php#L226

The change to the server (the forms were working prior to this change), was updating from Debian Bullseye to Debian Bookworm, the instance uses PHP-FPM 8.3.

I have tried to duplicate this on another server with the same configuration but haven't been able to.

toad:
Fixes a forms bug by changing how files_versions gets the path for a Node to tolerate the forms app writing files that are not even visible to the logged-in user.

In FileEventsListener::getPathForNode(), if the user is not logged in, and we can't find the file owner, we guess the file owner from the path. Most likely the username in the path has some access to the file, so we can find the filename. This allows writing files when the user is not logged in, for instance where we are filling in a form from a public link.

However, in the case where there is a logged-in user, and the file is on a group share, both the username and the owner will be the same UID. Which is fine as long as we have access to the file.

But in the case where we are actually writing a spreadsheet update for a form submission, the logged in user may not have access to the file. However, it is still a legitimate write by the forms app, so it is safe to use the owner in the path, just as if we were logged out.

I'm not 100% sure about the security implications here and request review. However it provides a simple workaround for our use case:
- Form with an attached spreadsheet
- User is logged in
- User has access to the form but not the spreadsheet
- The spreadsheet is in a group folder
- The versioning app is enabled

The result is an error every time, see the forms bug https://github.com/nextcloud/forms/issues/2067 (note that the actual error has changed since the initial report!)

The alternative is to fix the underlying problem in the forms app, which would be a much bigger diff, see the above.

However, if I am right about the security model here, this is a safe workaround, and may actually be correct.

* Resolves: [forms bug #2067](https://github.com/nextcloud/forms/issues/2067)

## Summary

Fixes a reproducible form submission error due to a conflict between how forms and versioning see permissions.

At this stage I'm asking for review that this is basically the right solution and there are no security implications. I believe it is safe but request review on that assumption!

An alternative fix in forms would be much more work, see their bug.

## TODO

- [ ] Confirm that the fix is safe before proceeding with the rest

## Checklist

### DIFF
--- a/apps/files_versions/lib/Listener/FileEventsListener.php
+++ b/apps/files_versions/lib/Listener/FileEventsListener.php
@@ -361,9 +361,10 @@ class FileEventsListener implements IEventListener {
 
 		$owner = $node->getOwner()?->getUid();
 
-		// If no owner, extract it from the path.
-		// e.g. /user/files/foobar.txt
-		if (!$owner) {
+		// If no owner, extract it from the path, e.g. /user/files/foobar.txt
+		// Also try this if they are the same, because it might be a group folder that the user does not have access to
+		// E.g. where filling in a form with a spreadsheet attached
+		if (!$owner || $owner == $user) {
 			$parts = explode('/', $node->getPath(), 4);
 			if (count($parts) === 4) {
 				$owner = $parts[1];


come-nc (commented):
This does not look like something we should do, or not this way.

Do you have a stack trace for the permission error in forms app?
Why does the files write pass but not the versions write :thinking: 

The issue is specific to groupfolders where there is no owner so getOwner can return a user with no write permission I think. Maybe that’s the root problem to solve, not sure.

toad:
In the error case I was debugging, it's actually in files_exists. If the user has read-only access to the file then it will work. If the user does not have access to the file at all, it fails. Which is not unreasonable - as they said earlier in the ticket, ideally Forms should use a worker thread with its own user. On the other hand I have to assume that running without files_versions doesn't create a massive security hole?

Anyway, the stack trace is:

{"file":"/var/www/html/lib/private/Files/View.php","line":531,"function":"basicOperation","class":"OC\\Files\\View","type":"->","args":["file_exists",null]},
{"file":"/var/www/html/lib/private/Files/Filesystem.php","line":546,"function":"file_exists","class":"OC\\Files\\View","type":"->","args":[null]},
{"file":"/var/www/html/apps/files_versions/lib/Storage.php","line":191,"function":"file_exists","class":"OC\\Files\\Filesystem","type":"::","args":[null]},
{"file":"/var/www/html/apps/files_versions/lib/Listener/FileEventsListener.php","line":197,"function":"store","class":"OCA\\Files_Versions\\Storage","type":"::","args":[null]},
{"file":"/var/www/html/apps/files_versions/lib/Listener/FileEventsListener.php","line":103,"function":"write_hook","class":"OCA\\Files_Versions\\Listener\\FileEventsListener","type":"->","args":[["OC\\Files\\Node\\File"]]},
{"file":"/var/www/html/lib/private/EventDispatcher/ServiceEventListener.php","line":86,"function":"handle","class":"OCA\\Files_Versions\\Listener\\FileEventsListener","type":"->","args":[["OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent"]]},
{"file":"/var/www/html/3rdparty/symfony/event-dispatcher/EventDispatcher.php","line":230,"function":"__invoke","class":"OC\\EventDispatcher\\ServiceEventListener","type":"->","args":[["OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent"],"OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent",["Symfony\\Component\\EventDispatcher\\EventDispatcher"]]},
{"file":"/var/www/html/3rdparty/symfony/event-dispatcher/EventDispatcher.php","line":59,"function":"callListeners","class":"Symfony\\Component\\EventDispatcher\\EventDispatcher","type":"->","args":[[["Closure"],["Closure"]],"OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent",["OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent"]]},
{"file":"/var/www/html/lib/private/EventDispatcher/EventDispatcher.php","line":86,"function":"dispatch","class":"Symfony\\Component\\EventDispatcher\\EventDispatcher","type":"->","args":[["OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent"],"OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent"]},
{"file":"/var/www/html/lib/private/EventDispatcher/EventDispatcher.php","line":98,"function":"dispatch","class":"OC\\EventDispatcher\\EventDispatcher","type":"->","args":["OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent",["OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent"]]},
{"file":"/var/www/html/lib/private/Files/Node/HookConnector.php","line":93,"function":"dispatchTyped","class":"OC\\EventDispatcher\\EventDispatcher","type":"->","args":[["OCP\\Files\\Events\\Node\\BeforeNodeWrittenEvent"]]},
{"file":"/var/www/html/lib/private/legacy/OC_Hook.php","line":105,"function":"write","class":"OC\\Files\\Node\\HookConnector","type":"->","args":[[true,"/uk tech/test/Yet another test form (responses).xlsx"]]},
{"file":"/var/www/html/lib/private/Files/View.php","line":1282,"function":"emit","class":"OC_Hook","type":"::","args":["OC_Filesystem","write",[true,"/uk tech/test/Yet another test form (responses).xlsx"]]},
{"file":"/var/www/html/lib/private/Files/View.php","line":1154,"function":"runHooks","class":"OC\\Files\\View","type":"->","args":[["update","write"],"/uk tech/test/Yet another test form (responses).xlsx"]},
{"file":"/var/www/html/lib/private/Files/View.php","line":683,"function":"basicOperation","class":"OC\\Files\\View","type":"->","args":["file_put_contents","/matthewtoseland/files/uk tech/test/Yet another test form (responses).xlsx",["update","write"],null]},
{"file":"/var/www/html/lib/private/Files/Node/File.php","line":73,"function":"file_put_contents","class":"OC\\Files\\View","type":"->","args":["/matthewtoseland/files/uk tech/test/Yet another test form (responses).xlsx",null]},
{"file":"/var/www/html/custom_apps/forms/lib/Service/SubmissionService.php","line":224,"function":"putContent","class":"OC\\Files\\Node\\File","type":"->","args":[null]},
{"file":"/var/www/html/custom_apps/forms/lib/Controller/ApiController.php","line":1145,"function":"writeFileToCloud","class":"OCA\\Forms\\Service\\SubmissionService","type":"->","args":[["OCA\\Forms\\Db\\Form",11],"/uk tech/test/Yet another test form (responses).xlsx","xlsx","matthewtoseland"]},
{"file":"/var/www/html/lib/private/AppFramework/Http/Dispatcher.php","line":232,"function":"insertSubmission","class":"OCA\\Forms\\Controller\\ApiController","type":"->","args":[11,[["42"],["44"],["Test test text"]],""]},
{"file":"/var/www/html/lib/private/AppFramework/Http/Dispatcher.php","line":138,"function":"executeController","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->","args":[["OCA\\Forms\\Controller\\ApiController"],"insertSubmission"]},
{"file":"/var/www/html/lib/private/AppFramework/App.php","line":184,"function":"dispatch","class":"OC\\AppFramework\\Http\\Dispatcher","type":"->","args":[["OCA\\Forms\\Controller\\ApiController"],"insertSubmission"]},
{"file":"/var/www/html/lib/private/Route/Router.php","line":331,"function":"main","class":"OC\\AppFramework\\App","type":"::","args":["OCA\\Forms\\Controller\\ApiController","insertSubmission",["OC\\AppFramework\\DependencyInjection\\DIContainer"],["v2.4","ocs.forms.api.insertsubmission"]]},
{"file":"/var/www/html/ocs/v1.php","line":66,"function":"match","class":"OC\\Route\\Router","type":"->","args":["/ocsapp/apps/forms/api/v2.4/submission/insert"]},
{"file":"/var/www/html/ocs/v2.php","line":23,"args":["/var/www/html/ocs/v1.php"],"function":"require_once"}]

,"File":"/var/www/html/lib/private/Files/View.php","Line":1138},"message":"OC\\Files\\View::basicOperation(): Argument #2 ($path) must be of type string, null given, called in /var/www/html/lib/private/Files/View.php on line 531 in file '/var/www/html/lib/private/Files/View.php' line 1138","exception":{},"CustomMessage":"OC\\Files\\View::basicOperation(): Argument #2 ($path) must be of type string, null given, called in /var/www/html/lib/private/Files/View.php on line 531 in file '/var/www/html/lib/private/Files/View.php' line 1138"}}

Chartman123:
@toad We'll investigate further in Forms together with @come-nc if there's some problem in getting the correct path/Node. So perhaps there will be a real fix instead of doing some kind of workaround :)

artonge:
The `getPathForNode` method is only here to bridge between the "modern" event parameters providing `Node` objects, and the legacy `Storage` class which uses paths. So there is indeed no security implication, the goal is only to have a path so that `Storage` can do its thing.

There might be a better way to implement `getPathForNode`, but couldn't find one at the time, and the method ended up in the weird owner hunt that it currently is. So any additional weirdness will not be surprising, and probably shouldn't be rejected :).

toad:
> The `getPathForNode` method is only here to bridge between the "modern" event parameters providing `Node` objects, and the legacy `Storage` class which uses paths. So there is indeed no security implication, the goal is only to have a path so that `Storage` can do its thing.
> 
> There might be a better way to implement `getPathForNode`, but couldn't find one at the time, and the method ended up in the weird owner hunt that it currently is. So any additional weirdness will not be surprising, and probably shouldn't be rejected :).
> 
> EDIT: Also happy to have a better way to get the path. @icewind1991 might know.

So that Storage can do what exactly? As I understand it, permissions checks apply at a higher level. Versions does not replace higher level permissions checks, it just hooks in after they've taken place, creates some metadata for the new version, and presumably prevents the old one being deleted?

So it's safe to deploy this as a temporary fix, even though it's probably not the *right* fix?

sorbaugh:
Hi @toad , from what I read it seems the the preferred approach here is to migrate from hooks to events for this part in forms. We'd rather not merge something temporary as that risks committing less attention to the real source of the issue, and we'd rather work on that. Nevertheless your PR brought our attention to this issue, so thank you very much for that!

toad:
> Hi @toad , from what I read it seems the the preferred approach here is to migrate from hooks to events for this part in forms. We'd rather not merge something temporary as that risks committing less attention to the real source of the issue, and we'd rather work on that. Nevertheless your PR brought our attention to this issue, so thank you very much for that!

Is there a public ticket I can follow to monitor the status of the correct fix? Thanks!

PR #48323, opened 2024-09-24
refactor(updatenotification): Migrate legacy code

susnux:
## Summary

1. Remove hook usage and just provide an initial state
2. Replace jQuery code with modern non-deprecated frontend code

Also fixed the appinfo.xml as the order of entries is fixed in the XSD.

### Screen shots
Bildschirmfoto_20240924_170829

## Checklist


provokateurin (approved):

Pytal (approved):

apps/updatenotification/lib/UpdateChecker.php
```diff
@@ -59,13 +55,17 @@
...
 	/**
-	 * @param array $data
+	 * Provide update information as initial state
 	 */
-	public function populateJavaScriptVariables(array $data) {
-		$data['array']['oc_updateState'] = json_encode([
-			'updateAvailable' => true,
-			'updateVersion' => $this->getUpdateState()['updateVersionString'],
-			'updateLink' => $this->getUpdateState()['updateLink'] ?? '',
+	public function setInitialState() {
```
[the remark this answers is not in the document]
susnux:
```suggestion
	public function setInitialState(): void {
```

apps/updatenotification/src/update-notification-legacy.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
come-nc, line 1: Why is the file still named with -legacy?

susnux: Legacy did not meant the content but the usage: It is only needed if you do not have notifications app enabled.

We could also name it `-fallback`

come-nc: fallback would be clearer for me, but no big deal, was just my curiosity when reading this.

PR #48325, opened 2024-09-24
chore(files_versions): Refactor function for lisibility

solracsf:
## Checklist

### DIFF
--- a/apps/files_versions/lib/Listener/FileEventsListener.php
+++ b/apps/files_versions/lib/Listener/FileEventsListener.php
@@ -64,49 +64,20 @@ class FileEventsListener implements IEventListener {
 	}
 
 	public function handle(Event $event): void {
-		if ($event instanceof NodeCreatedEvent) {
-			$this->created($event->getNode());
-		}
-
-		if ($event instanceof BeforeNodeTouchedEvent) {
-			$this->pre_touch_hook($event->getNode());
-		}
-
-		if ($event instanceof NodeTouchedEvent) {
-			$this->touch_hook($event->getNode());
-		}
-
-		if ($event instanceof BeforeNodeWrittenEvent) {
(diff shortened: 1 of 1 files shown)


kesselb:
I don't consider that an improvement :see_no_evil: 

Fine by me, but please fix the commit message ;)

solracsf:
What's wrong with the commit message? 🤔

kesselb:
The commit type should be "refactoring" instead of "chore" https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type.

lisbility => legibility or readability

provokateurin (commented):
I agree with @kesselb, it's not really better than before.
It also abuses the match functionality, so I don't really like that either.

apps/files_versions/lib/Listener/FileEventsListener.php
```diff
@@ -331,7 +302,7 @@ public function pre_renameOrCopy_hook(Node $source, Node $target): void {
 		$manager = Filesystem::getMountManager();
 		$mount = $manager->find($absOldPath);
 		$internalPath = $mount->getInternalPath($absOldPath);
-		if ($internalPath === '' and $mount instanceof MoveableMount) {
+		if ($internalPath === '' && $mount instanceof MoveableMount) {
```
susnux, line 305: We should have a cs fixer rule for this

provokateurin: Rector does this as well, I've never seen anyone use this syntax in recent years so it is only in legacy code anyway.

come-nc:
I agree with the others, I dislike using match like this. It’s harder to read as the calls are not aligned with each other anymore.
Closing. The and thing is not big enough to justify a PR.

PR #48331, opened 2024-09-24
fix(files): Fix having to resolve conflicts twice when dropping files

Pytal:
## Resolves

When dragging and dropping some files into a folder with identically named files, a conflict resolution dialog will pop up and after submitting your selection subsequent conflict dialogs will be shown for each file individually. We want to proceed with the users initial selection instead of repeating the conflict resolution.

## Checklist


susnux (approved):
makes sense (we should also implement recursive conflict resolution for internal moves similar we do with uploads).

szaimen (approved):
Tested and works. Thanks for fixing this! :)

PR #48332, opened 2024-09-25
refactor(encryption): Migrate from hooks to events

susnux:
## Summary

Migrate code from hooks to typed events, also unify the user events as some had already a `getUid` method -> now all provide this convenient helper.

## Checklist


apps/encryption/lib/AppInfo/Application.php
```diff
@@ -49,46 +56,33 @@
...
-					$server->getUserSession(),
-					$container->query(Util::class),
-					$container->query(Session::class),
-					$container->query(Crypt::class),
-					$container->query(Recovery::class))
-			]);
-
-			$hookManager->fireHooks();
-		} else {
+	public function registerEventListeners(IConfig $config, IEventDispatcher $eventDispatcher) {
```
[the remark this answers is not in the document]
susnux:
```suggestion
	public function registerEventListeners(IConfig $config, IEventDispatcher $eventDispatcher): void {
```

apps/encryption/lib/KeyManager.php
```diff
@@ -300,6 +298,10 @@ public function init($uid, $passPhrase) {
 				$passPhrase = $this->getMasterKeyPassword();
 				$privateKey = $this->getSystemPrivateKey($uid);
 			} else {
+				if ($passPhrase === null) {
+					$this->logger->warning('Master key is disabled but not passphrase provided.');
+					return false;
+				}
```
provokateurin, line 304: Looks like a change not entirely related to the refactor and it should also be backported?

apps/encryption/lib/Listeners/UserEventsListener.php
```diff
@@ -0,0 +1,147 @@
...
+			($event instanceof BeforePasswordUpdatedEvent) =>
+				$this->onBeforePasswordUpdated($event->getUser(), $event->getPassword(), $event->getRecoveryPassword()),
+			($event instanceof PasswordUpdatedEvent) =>
+				$this->onPasswordUpdated($event->getUid(), $event->getPassword(), $event->getRecoveryPassword()),
+			($event instanceof BeforePasswordResetEvent) =>
+				$this->onBeforePasswordReset($event->getUid()),
+			($event instanceof PasswordResetEvent) =>
+				$this->onPasswordReset($event->getUid(), $event->getPassword()),
+
+		};
```
provokateurin, line 65: I would rather avoid abusing match for this and just use a few if statements, but maybe that's just personal taste.

provokateurin (approved):

apps/encryption/lib/Listeners/UserEventsListener.php
susnux: I am fine with both, either this or 8 `if` / `elseif` statements (+ the last `elseif` will trigger psalm as from the typing it can not be another value but I would like to be defensive here and only handle if the event is the right class)

apps/encryption/lib/KeyManager.php
susnux: This was a psalm error that was ignored before as it could be possible that null is passed (e.g. with master key) this was simply ignored before.

apps/encryption/lib/AppInfo/Application.php
```diff
@@ -99,14 +97,14 @@ public function registerEncryptionModule(IManager $encryptionManager) {
...
-					$container->getServer()->get(LoggerInterface::class),
-					$container->getServer()->getL10N($container->getAppName())
+					$container->get(Crypt::class),
+					$container->get(KeyManager::class),
+					$container->get(Util::class),
+					$container->get(Session::class),
+					$container->get(EncryptAll::class),
+					$container->get(DecryptAll::class),
+					$container->get(LoggerInterface::class),
+					$container->get(IL10N::class),
```
come-nc, line 107: Is that equivalent to `$container->getServer()->getL10N($container->getAppName())`?
There is a scoped service for IL10N with the app name?

apps/encryption/lib/Listeners/UserEventsListener.php
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	/**
+	 * Startup encryption backend upon user login
+	 *
+	 * This method should never be called for users using client side encryption
```
come-nc, line 71:
```suggestion
```
This comment is obsolete I think as now both SSE and E2EE can be used together.

apps/encryption/lib/Services/PassphraseService.php
```diff
@@ -0,0 +1,142 @@
...
+use OCP\Encryption\Exceptions\GenericEncryptionException;
+use OCP\IUser;
+use OCP\IUserManager;
+use OCP\IUserSession;
+use Psr\Log\LoggerInterface;
+
+class PassphraseService {
+
+	/** @var array<string, bool> */
+	private static $passwordResetUsers = [];
```
come-nc, line 25:
```suggestion
	private static array $passwordResetUsers = [];
```

core/Events/BeforePasswordResetEvent.php
```diff
@@ -34,6 +34,13 @@ public function getUser(): IUser {
 		return $this->user;
 	}
 
+	/**
+	 * @since 31.0.0
```
come-nc, line 38: Any chance of moving these events to OCP without creating a mess?

lib/private/Log/ExceptionSerializer.php
```diff
@@ -152,14 +153,15 @@ public function __construct(
...
-		UserHooks::class => [
-			'login',
-			'postCreateUser',
-			'postDeleteUser',
-			'prePasswordReset',
-			'postPasswordReset',
-			'preSetPassphrase',
-			'setPassphrase',
+		UserEventsListener::class => [
+			'onUserCreated',
```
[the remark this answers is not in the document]
come-nc: When we ditch PHP 8.1 we can start using https://php.watch/versions/8.2/backtrace-parameter-redaction

come-nc (commented):
Not sure I like the getUid thing, especially since it uses a different casing than the method from IUser.
Otherwise, great work!

apps/encryption/lib/Listeners/UserEventsListener.php
artonge: Hopefully :crossed_fingers:

apps/encryption/lib/Services/PassphraseService.php
```diff
@@ -0,0 +1,142 @@
...
+		// if we are in the process to resetting a user password, we have nothing
+		// to do here
+		if (isset(self::$passwordResetUsers[$userId])) {
+			return true;
+		}
+
+		// Check user exists on backend
+		$user = $this->userManager->get($userId);
+		if ($user === null) {
+			return false;
```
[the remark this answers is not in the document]
susnux: This here is not fatal, so this can just return (but I am fine with void or bool)

susnux:
> Not sure I like the getUid thing, especially since it uses a different casing than the method from IUser.

The IUser one is quite old, older than our naming guidelines that prefer Uid over UID (or Url over URL).
But I am fine with removing it again, I just thought it would make sense to unify it as there are already events with `getUid` (exactly that casing).

core/Events/BeforePasswordResetEvent.php
susnux: IIRC the event name is the class string of the event, meaning we can not move without also emitting the old event until it is deprecated enough.

susnux: So I would say this makes sense, but please in a follow up

apps/encryption/lib/Listeners/UserEventsListener.php
susnux: Removed it and replaced with `if` as we closed the PR that changed other places to `match`

apps/encryption/lib/Listeners/UserEventsListener.php
```diff
@@ -0,0 +1,147 @@
...
+	 */
+	private function onUserLogin(string $userId, ?string $password): void {
+		// ensure filesystem is loaded
+		if (!\OC\Files\Filesystem::$loaded) {
+			\OC_Util::setupFS($userId);
+		}
```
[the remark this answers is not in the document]
artonge: @icewind1991 ?

apps/encryption/lib/Services/PassphraseService.php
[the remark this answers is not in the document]
artonge: No strong opinion.

apps/encryption/lib/Services/PassphraseService.php
```diff
@@ -0,0 +1,142 @@
...
+		} else {
+			// admin changed the password for a different user, create new keys and re-encrypt file keys
+			$recoveryPassword = $recoveryPassword ?? '';
+			$this->initMountPoints($user);
+
+			$recoveryKeyId = $this->keyManager->getRecoveryKeyId();
+			$recoveryKey = $this->keyManager->getSystemPrivateKey($recoveryKeyId);
+			try {
+				$this->crypt->decryptPrivateKey($recoveryKey, $recoveryPassword);
+			} catch (\Throwable) {
```
come-nc, line 94:
```suggestion
			} catch (\Exception) {
```
Avoid catch Throwable except for specific things (like, if you have a plugin system and call code from other apps, sure, catch Throwable, but when you call your own code, catch Exception and fix other errors.)

come-nc (commented):
All good apart from the catch Throwable thing.

apps/encryption/lib/Listeners/UserEventsListener.php
```diff
@@ -0,0 +1,144 @@
...
+
+	/**
+	 * Startup encryption backend upon user login
+	 *
+	 */
+	private function onUserLogin(string $userId, ?string $password): void {
+		// ensure filesystem is loaded
+		if (!\OC\Files\Filesystem::$loaded) {
+			\OC_Util::setupFS($userId);
+		}
```
come-nc, line 74:
```suggestion
		$this->setupManager->setupForUser($user);
```
Discussed this with @icewind1991 , it’s safer to keep the call but please use the modern SetupManager class.
You will need the IUser instance from the event.

susnux: Sadly still no OCP :(

come-nc (approved):

PR #48355, opened 2024-09-25
refactor(rector): Extend scope to all PHP files

provokateurin:
## Summary

Currently fails due to a bug in rector: https://github.com/rectorphp/rector/issues/8841

## Checklist

### DIFF
--- a/build/rector.php
+++ b/build/rector.php
@@ -14,15 +14,19 @@ $nextcloudDir = dirname(__DIR__);
 $config = RectorConfig::configure()
 	->withPaths([
 		$nextcloudDir . '/apps',
-		// $nextcloudDir . '/config',
-		// $nextcloudDir . '/core',
-		// $nextcloudDir . '/lib',
-		// $nextcloudDir . '/ocs',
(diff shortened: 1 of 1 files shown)


build/rector.php
```diff
@@ -14,15 +14,19 @@
...
-		// $nextcloudDir . '/themes',
+		$nextcloudDir . '/build',
+		$nextcloudDir . '/config',
+		$nextcloudDir . '/core',
+		$nextcloudDir . '/lib',
+		$nextcloudDir . '/ocs',
+		$nextcloudDir . '/ocs-provider',
+		$nextcloudDir . '/tests',
+		$nextcloudDir . '/themes',
+		$nextcloudDir . '/*.php',
```
come-nc, line 25: If we do so please add them one by one with the commit applying the result to the folder.
But I would prefer to apply more rules to apps folder instead.
Maybe we should have 2 rector config, one for apps and one for core/lib?
We can have matching rector scripts in composer.

provokateurin: Separate configs sounds good :+1:

PR #48362, opened 2024-09-26
refactor(apps): Use PHP 5.5 features

provokateurin:
## Summary

Setting rector PHP level to 5.3 and 5.4 did not lead to any changes so we can start at 5.5 right away.

## Checklist

### DIFF
--- a/apps/admin_audit/lib/AppInfo/Application.php
+++ b/apps/admin_audit/lib/AppInfo/Application.php
@@ -232,6 +232,6 @@ class Application extends App implements IBootstrap {
 	private function trashbinHooks(IAuditLogger $logger): void {
 		$trashActions = new Trashbin($logger);
 		Util::connectHook('\OCP\Trashbin', 'preDelete', $trashActions, 'delete');
-		Util::connectHook('\OCA\Files_Trashbin\Trashbin', 'post_restore', $trashActions, 'restore');
+		Util::connectHook(\OCA\Files_Trashbin\Trashbin::class, 'post_restore', $trashActions, 'restore');
 	}
 }
--- a/apps/comments/tests/Unit/AppInfo/ApplicationTest.php
+++ b/apps/comments/tests/Unit/AppInfo/ApplicationTest.php
@@ -35,12 +35,12 @@ class ApplicationTest extends TestCase {
 		$c = $app->getContainer();
 
 		$services = [
-			'OCA\Comments\Controller\NotificationsController',
-			'OCA\Comments\Activity\Filter',
-			'OCA\Comments\Activity\Listener',
-			'OCA\Comments\Activity\Provider',
-			'OCA\Comments\Activity\Setting',
-			'OCA\Comments\Notification\Listener',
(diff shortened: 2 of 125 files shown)


apps/admin_audit/lib/AppInfo/Application.php
```diff
@@ -232,6 +232,6 @@ private function versionsHooks(IAuditLogger $logger): void {
 	private function trashbinHooks(IAuditLogger $logger): void {
 		$trashActions = new Trashbin($logger);
 		Util::connectHook('\OCP\Trashbin', 'preDelete', $trashActions, 'delete');
-		Util::connectHook('\OCA\Files_Trashbin\Trashbin', 'post_restore', $trashActions, 'restore');
+		Util::connectHook(\OCA\Files_Trashbin\Trashbin::class, 'post_restore', $trashActions, 'restore');
```
nickvergessen, line 235: This was created by Christoph the other day.
It is a breaking change:
```php
var_dump('\OCA\Files_Trashbin\Trashbin' === \OCA\Files_Trashbin\Trashbin::class);
// bool(false)
```

We can only do this, if we modify the hooker itself and basically trim leading slashes on connect and emit

nickvergessen (changes requested):

apps/admin_audit/lib/AppInfo/Application.php
provokateurin: So the comparison fails, but shouldn't the ::class result in the exact same string content?
Maybe I'm missing something here.

provokateurin:
Another problem as Psalm reported it: Class strings for classes from apps outside of server don't work. Using stubs for that should be fine, right? And there shouldn't be any problem with loading that code even if the class is not available?

apps/admin_audit/lib/AppInfo/Application.php
come-nc: `\OCA\Files_Trashbin\Trashbin::class` is `'OCA\Files_Trashbin\Trashbin'` without the first `\`

come-nc:
> Another problem as Psalm reported it: Class strings for classes from apps outside of server don't work. Using stubs for that should be fine, right? And there shouldn't be any problem with loading that code even if the class is not available?

Yeah psalm is annoying with that. It’s perfectly valid to use `\My\Fully\Qualified\Name::class` even if the class does not exists, it still works.

come-nc:
The only problematic place with starting / should be connectHook, which we should get rid of anyway.
So maybe we first migrate away from hooks in apps before applying this?

apps/admin_audit/lib/AppInfo/Application.php
provokateurin: I see, in the preview IntelliJ displayed it to me with the leading backslash :woman_shrugging:

provokateurin:
>  So maybe we first migrate away from hooks in apps before applying this?

Sounds good :+1:

provokateurin:
Closing for now since there is nothing directly actionable here.

PR #48366, opened 2024-09-26
fix(files_sharing): Parse OCM share permissions from OCM and not OCS prop

provokateurin:
## Summary

Checks for OCM permissions prop, but parses them from the OCS permissions prop.

## Checklist

### DIFF
--- a/apps/files_sharing/lib/External/Storage.php
+++ b/apps/files_sharing/lib/External/Storage.php
@@ -345,6 +345,10 @@ class Storage extends DAV implements ISharedStorage, IDisableEncryptionStorage, 
 
 	public function getPermissions($path): int {
 		$response = $this->propfind($path);
+		if ($response === false) {
+			return 0;
+		}
+
 		$ocsPermissions = $response['{http://open-collaboration-services.org/ns}share-permissions'] ?? null;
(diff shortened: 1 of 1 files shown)


apps/files_sharing/lib/External/Storage.php
```diff
@@ -345,14 +345,17 @@ public function isSharable($path): bool {
 
 	public function getPermissions($path): int {
 		$response = $this->propfind($path);
+		$ocsPermissions = $response['{http://open-collaboration-services.org/ns}share-permissions'];
```
come-nc, line 348:
```suggestion
		$ocsPermissions = $response['{http://open-collaboration-services.org/ns}share-permissions'] ?? null;
```
Is something like that not needed? What’s the type of `$response`?

come-nc (commented):
Ok for the logic, but error handling is missing, $reponse could be false, or could be missing the keys.

apps/files_sharing/lib/External/Storage.php
provokateurin: Right, missing key already results in null but also logs an error message.

come-nc (approved):

kesselb (approved):

PR #48371, opened 2024-09-26
Auto import Nextcloud classes in rector runs

come-nc:
## Summary

When using rector, fully qualified names are often outputted by the rules, like `\OCP\Server` or `\OCP\Http\Client\IClientService`. To avoid that, one can use `withImportNames()` to tell rector to import names with use statements, but then it will import all names.

This is not always desirable, for instance `\LDAP\Connection` in user_ldap would be confusing because there is an `\OCA\User_LDAP\Connection` class as well.

So I added a custom "SkipVoter" which skips auto-import for all classes outside of OC/OCA/OCP.

It may still be a bit confusing for common names like IManager, but I think we can live with that? Or I can add other exceptions if needed.
Note that rector is smart enough to not import two classes with the same name, but it will randomly import one of them and keep the fully qualified name for the other one.

I’m doing that so that when we apply Nextcloud rules set later we do not get fully qualified names everywhere but nice use statements instead.

## Checklist

### DIFF
--- a/apps/dav/appinfo/v1/caldav.php
+++ b/apps/dav/appinfo/v1/caldav.php
@@ -10,6 +10,8 @@ use OC\KnownUser\KnownUserService;
 use OCA\DAV\CalDAV\CalDavBackend;
 use OCA\DAV\CalDAV\CalendarRoot;
 use OCA\DAV\CalDAV\DefaultCalendarValidator;
+use OCA\DAV\CalDAV\Proxy\ProxyMapper;
+use OCA\DAV\CalDAV\Schedule\IMipPlugin;
 use OCA\DAV\CalDAV\Security\RateLimitingPlugin;
 use OCA\DAV\CalDAV\Validation\CalDavValidatePlugin;
 use OCA\DAV\Connector\LegacyDAVACL;
@@ -18,6 +20,9 @@ use OCA\DAV\Connector\Sabre\ExceptionLoggerPlugin;
 use OCA\DAV\Connector\Sabre\MaintenancePlugin;
 use OCA\DAV\Connector\Sabre\Principal;
 use OCP\Accounts\IAccountManager;
+use OCP\EventDispatcher\IEventDispatcher;
+use OCP\IConfig;
+use OCP\Server;
 use Psr\Log\LoggerInterface;
 
 $authBackend = new Auth(
@@ -35,7 +40,7 @@ $principalBackend = new Principal(
 	\OC::$server->getShareManager(),
 	\OC::$server->getUserSession(),
 	\OC::$server->getAppManager(),
-	\OC::$server->query(\OCA\DAV\CalDAV\Proxy\ProxyMapper::class),
+	\OC::$server->query(ProxyMapper::class),
 	\OC::$server->get(KnownUserService::class),
 	\OC::$server->getConfig(),
 	\OC::$server->getL10NFactory(),
@@ -45,8 +50,8 @@ $db = \OC::$server->getDatabaseConnection();
 $userManager = \OC::$server->getUserManager();
 $random = \OC::$server->getSecureRandom();
 $logger = \OC::$server->get(LoggerInterface::class);
-$dispatcher = \OC::$server->get(\OCP\EventDispatcher\IEventDispatcher::class);
-$config = \OC::$server->get(\OCP\IConfig::class);
+$dispatcher = \OC::$server->get(IEventDispatcher::class);
+$config = \OC::$server->get(IConfig::class);
 
 $calDavBackend = new CalDavBackend(
 	$db,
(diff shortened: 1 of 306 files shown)


build/rector.php
```diff
@@ -7,10 +7,32 @@
...
 
+use PhpParser\Node;
+use Rector\CodingStyle\Contract\ClassNameImport\ClassNameImportSkipVoterInterface;
 use Rector\Config\RectorConfig;
+use Rector\StaticTypeMapper\ValueObject\Type\FullyQualifiedObjectType;
+use Rector\ValueObject\Application\File;
 
 $nextcloudDir = dirname(__DIR__);
 
+class NextcloudNamespaceSkipVoter implements ClassNameImportSkipVoterInterface {
```
provokateurin, line 18: This should go into nextcloud/rector then?

provokateurin (commented):
I'm not sure if I like it. In quite a few places it is nice to use the fqn to avoid mistakes.
Then there is also the case when two classes have the same name where I would actually prefer that both use the fqn and not one fqn and one import.

build/rector.php
come-nc: It could, yeah.
The thing is we cannot make it configurable, because it’s added as a service by class name only.
But if it stays general like this we can put in nextcloud/rector, yes. Also we can make it extendable with a protected property to override if needed.

come-nc:
> In quite a few places it is nice to use the fqn to avoid mistakes.

Do you have examples in mind so that we can look if there is a pattern we could apply?

> Then there is also the case when two classes have the same name where I would actually prefer that both use the fqn and not one fqn and one import.

That is quite hard to implement I think. The SkipVoter which avoid use conflict is based on the already imported use list: https://github.com/rectorphp/rector/blob/main/rules/CodingStyle/ClassNameImport/ClassNameImportSkipVoter/UsesClassNameImportSkipVoter.php
Doing the same thing without even adding the first use would require collecting all shortnames from the file before deciding whether to import them or not.

apps/dav/appinfo/v1/caldav.php
```diff
@@ -93,14 +100,14 @@
 
 $server->addPlugin(new \Sabre\DAV\Sync\Plugin());
 $server->addPlugin(new \Sabre\CalDAV\ICSExportPlugin());
-$server->addPlugin(new \OCA\DAV\CalDAV\Schedule\Plugin(\OC::$server->getConfig(), \OC::$server->get(LoggerInterface::class), \OC::$server->get(DefaultCalendarValidator::class)));
+$server->addPlugin(new Plugin(\OC::$server->getConfig(), \OC::$server->get(LoggerInterface::class), \OC::$server->get(DefaultCalendarValidator::class)));
```
provokateurin, line 103: For example here it would be better to use the fqn as `Plugin` is a very generic name.

Same thing goes for the `IManager` classes. For those it would be nice to either use the fqn or at least the part before that so it is `Group\IManager` and so on.

Maybe this is not so bad, but to me it seems harder to read the code and easier to make mistakes when changing imports.

provokateurin:
Could we maybe limit this to a few classes like \OCP\Server and \OCP\Util for now? I don't know how many other cases we have where this happens a lot.

come-nc:
> Could we maybe limit this to a few classes like \OCP\Server and \OCP\Util for now? I don't know how many other cases we have where this happens a lot.

My problem is this rule set: https://github.com/nextcloud-libraries/rector/blob/8061985c44798bac1657ba5bc2d6efa5b10e9089/config/nextcloud-25/nextcloud-25-deprecations.php#L17
Without `withImportNames` it will use fully qualified name for all Server::get() calls and that’s ugly.
I can try again a solution like https://github.com/nextcloud-libraries/rector/pull/11/files to instead make our rules add stuff to the use statements independantly of `withImportNames`. I failed before but I have a better understanding of rector internals now.

come-nc:
So, the approach in https://github.com/nextcloud-libraries/rector/pull/11/files failed.

@provokateurin What do you think about using the strategy in this PR (auto import from OC/OCA/OCP), but with a manually maitained list of exempted class names? (The list would be something like Plugin,Manager,IManager,Exception).
Can I try that and see what the result looks like on our apps?

provokateurin:
Sounds good to me, excluding common duplicate names :+1:

come-nc:
> Sounds good to me, excluding common duplicate names 👍

Can you check at the current diff if the exclusion list and the result makes sense to you?

provokateurin (changes requested):
Looks good.
Maybe `Node`, `File` and `Folder` should be excluded as well since they also exist in `\OCA\DAV\Connector\Sabre` but this mostly affects the dav app so I'd leave that decision to @ChristophWurst and the likes.

provokateurin (approved):
Nice :+1:

ChristophWurst (approved):
Clicked into a bunch of random files and can't spot any issues with the imports

:shipit:

st3iny (approved):

come-nc:
@miaulalala @nickvergessen @tcitworld I fixed psalm baseline, only thing missing now is your reviews as code owners.

PR #48381, opened 2024-09-26
fix(ShareAPI): Send mails for mail shares by default

### LINKED ISSUE #48012: [Bug]: 30.0.0 share by email - no email is sent

### Bug description

Hello Devs.

After updating to 30.0.0 sharing files via email does not work as it is supposed to I guess. There is no mail being sent to the sharee. So the one I try to share the files with is not getting notified. Nothing in the logs.
Email in general works. Invitations for talk go out. 

### Steps to reproduce

1. Share file via Email
2. Save
3. No email to the sharee

### Expected behavior

Sharee gets notified.

### Nextcloud Server version

master

### Operating system

Debian/Ubuntu

### PHP engine version

PHP 8.3

### Web server

Apache (supported)

### Database engine version

MySQL

Upgraded to a MAJOR version (ex. 28 to 29)

Encryption is Disabled

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

```shell
{
    "system": {
        "datadirectory": "***REMOVED SENSITIVE VALUE***",
        "dbtype": "mysql",
        "version": "30.0.0.14",
        "dbname": "***REMOVED SENSITIVE VALUE***",
        "dbhost": "***REMOVED SENSITIVE VALUE***",
        "dbtableprefix": "oc_",
        "dbuser": "***REMOVED SENSITIVE VALUE***",
        "dbpassword": "***REMOVED SENSITIVE VALUE***",
        "installed": true,
        "instanceid": "***REMOVED SENSITIVE VALUE***",
        "maintenance": false,
        "maintenance_window_start": 3,
        "theme": "",
        "trusted_domains": [
            "fluffycloud.de"
        ],
        "mail_smtpmode": "sendmail",
        "mail_from_address": "***REMOVED SENSITIVE VALUE***",
        "mail_domain": "***REMOVED SENSITIVE VALUE***",
        "loglevel": 2,
        "log_rotate_size": 104857600,
        "filelocking.enabled": true,
        "memcache.local": "\\OC\\Memcache\\APCu",
        "memcache.locking": "\\OC\\Memcache\\Redis",
        "redis": {
            "host": "***REMOVED SENSITIVE VALUE***",
            "port": 0
        },
        "htaccess.RewriteBase": "\/",
        "updater.release.channel": "stable",
        "preview_max_filesize_image": -1,
        "overwrite.cli.url": "https:\/\/fluffycloud.de\/",
        "mysql.utf8mb4": true,
        "default_phone_region": "DE",
        "defaultapp": "files",
        "app_install_overwrite": [
            "calendar",
            "groupfolders",
            "mail",
            "spreed",
            "passman",
            "audioplayer",
            "polls",
            "passwords",
            "breezedark"
        ]
    }
}
```

### List of activated Apps

```shell
Enabled:
  - activity: 3.0.0
  - admin_audit: 1.20.0
  - calendar: 5.0.0
  - cloud_federation_api: 1.13.0
  - comments: 1.20.1
  - contacts: 6.1.0
  - contactsinteraction: 1.11.0
  - dashboard: 7.10.0
  - dav: 1.31.1
  - federatedfilesharing: 1.20.0
  - files: 2.2.0
  - files_downloadlimit: 3.0.0
  - files_external: 1.22.0
  - files_pdfviewer: 3.0.0
  - files_reminders: 1.3.0
  - files_sharing: 1.22.0
  - files_trashbin: 1.20.1
  - files_versions: 1.23.0
  - firstrunwizard: 3.0.0
  - logreader: 3.0.0
  - lookup_server_connector: 1.18.0
  - nextcloud_announcements: 2.0.0
  - notes: 4.10.1
  - notifications: 3.0.0
  - oauth2: 1.18.1
  - oidc: 0.9.4
  - password_policy: 2.0.0
  - passwords: 2024.7.20
  - photos: 3.0.2
  - polls: 7.2.3
  - privacy: 2.0.0
  - provisioning_api: 1.20.0
  - recommendations: 3.0.0
  - related_resources: 1.5.0
  - richdocuments: 8.5.0
  - serverinfo: 2.0.0
  - settings: 1.13.0
  - sharebymail: 1.20.0
  - spreed: 20.0.0
  - support: 2.0.0
  - survey_client: 2.0.0
  - systemtags: 1.20.0
  - text: 4.1.0
  - theming: 2.5.0
  - twofactor_backupcodes: 1.19.0
  - updatenotification: 1.20.0
  - user_status: 1.10.0
  - viewer: 3.0.0
  - weather_status: 1.10.0
  - webhook_listeners: 1.1.0-dev
  - workflowengine: 2.12.0
Disabled:
  - bruteforcesettings: 3.0.0 (installed 2.4.0)
  - circles: 30.0.0-dev (installed 27.0.0)
  - encryption: 2.18.0
  - federation: 1.20.0 (installed 1.17.0)
  - files_rightclick: 0.15.1 (installed 1.6.0)
  - groupfolders: 18.0.1 (installed 18.0.1)
  - secrets: 2.0.3 (installed 2.0.3)
  - suspicious_login: 8.0.0
  - twofactor_nextcloud_notification: 4.0.0
  - twofactor_totp: 12.0.0-dev (installed 5.0.0)
  - user_ldap: 1.21.0
  - whiteboard: 1.0.1 (installed 1.0.1)
```

### Nextcloud Signing status

```shell
No errors have been found.
```

### Nextcloud Logs

_No response_

### Additional info

_No response_

nfebe:
## Summary

It looks like, the frontend it needs to provide the `sendMail` param for the backend to decide whether mails would be sent.

Our UI does not have that at the moment so it should default to sending emails always for mail shares.

Not exactly sure how this was handled earlier but this is a good starting point.

Resolves : https://github.com/nextcloud/server/issues/48012

## Checklist

### DIFF
--- a/apps/files_sharing/lib/Controller/ShareAPIController.php
+++ b/apps/files_sharing/lib/Controller/ShareAPIController.php
@@ -659,7 +659,16 @@ class ShareAPIController extends OCSController {
 		$this->checkInheritedAttributes($share);
 
 		// Handle mail send
-		if ($sendMail === 'true' || $sendMail === 'false') {
+		if (is_null($sendMail)) {
+			// Define a default behavior when sendMail is not provided
+			if ($shareType === IShare::TYPE_EMAIL && strlen($shareWith) !== 0) {
+				// For email shares, the default is to send the mail
+				$share->setMailSend(true);
+			} else {
+				// For all other share types, the default is to not send the mail
+				$share->setMailSend(false);
+			}
+		} else {
 			$share->setMailSend($sendMail === 'true');
 		}
 


apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -658,8 +658,24 @@ public function createShare(
...
+			if($shareType === IShare::TYPE_EMAIL) {
+				// For email shares, the default is to send the mail
+				$share->setMailSend(true);
+			} else {
+				// For all other share types, the default is to not send the mail
+				$share->setMailSend(false);
+			}
+		} else {
+			// Convert common "truthy" and "falsy" values to proper booleans
+			$sendMailBool = filter_var($sendMail, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
```
provokateurin, line 672:
```suggestion
			$sendMailBool = $sendMail === 'true';
```
Just check if the value is `true` and disregard everything else as false.

provokateurin (approved):
Fix makes sense, curious how it broke though.

apps/files_sharing/lib/Controller/ShareAPIController.php
come-nc: You can directly do `$share->setMailSend($sendMail === 'true');` in the else block.

come-nc (approved):

provokateurin:
Why rebase it so often?

Fenn-CS:
> Why rebase it so often?

Some tests that seem unrelated (cypress especially are failing consistently)...

Re-basing triggers a rerun, but also add the latest updates on master in case something that broke the tests has been resolved. (Because a bunch of other PRs are failing tests too)...

nfebe (commented):
This breaks "File Requests"

https://github.com/nextcloud/server/blob/363ba6b0f5b99583e789695f19dcd96db00508ef/apps/files_sharing/lib/Controller/ShareAPIController.php#L721-L727


Why?

- File requests are sent with an empty string for `shareWith` which the backend uses to check if a shareWith value is available (poorly so)
- The file requests shareType is 4, which is mail share and I am not sure why that is.

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -718,7 +727,7 @@ public function createShare(
 			}
 
 			// Only share by mail have a recipient
-			if (is_string($shareWith) && $shareType === IShare::TYPE_EMAIL) {
+			if (!empty($shareWith) && !is_null($shareWith) && $shareType === IShare::TYPE_EMAIL) {
```
provokateurin, line 730:
```suggestion
			if ($shareType === IShare::TYPE_EMAIL) {
```
You don't need these checks because the comparison already handles these cases (and psalm doesn't like it either)

provokateurin (approved):

provokateurin (changes requested):
Oops

apps/files_sharing/src/components/NewFileRequestDialog.vue
```diff
@@ -284,7 +284,7 @@ export default defineComponent({
 				const request = await axios.post<OCSResponse>(shareUrl, {
 					// Always create a file request, but without mail share
 					// permissions, only a share link will be created.
-					shareType: sharingConfig.isMailShareAllowed ? ShareType.Email : ShareType.Link,
+					shareType: ShareType.Link,
```
provokateurin, line 287: Not sure about this, @skjnldsv what was the idea behind this?

Fenn-CS:
Add some more context on why the tests are failing and some related stuff with a plausible fix:

Share type setting can be changed (current solution), this seems to prevent the sendMail function for file requests as at now but since the links can be copied and shared, it's way better than breaking email shares on prod.

apps/files_sharing/src/components/NewFileRequestDialog.vue
skjnldsv: I only take showers :shower:

skjnldsv (changes requested):
Blocking, will discuss this in chat for easier comprehension

apps/files_sharing/src/components/NewFileRequestDialog.vue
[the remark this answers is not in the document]
skjnldsv: No, a mail share is a link share.
They both behave the same way. A link share does NOT send emails, if we start doing this, then we would break the default sharing rule that admin allows to send emails or not.

If admin allowed email shares, then a file request allow users to add emails and send the notifications
If admin does NOT, then users can only create a share link file request, and have to manually give the link to the recipients :)

skjnldsv:
> This breaks "File Requests"
> 
> * File requests are sent with an empty string for `shareWith` which the backend uses to check if a shareWith value is available (poorly so)
> * The file requests shareType is 4, which is mail share and I am not sure why that is.

File requests allow to add emails in attributes instead of shareWith. ShareWith does NOT support multiple recipients, and that was the only way to /cleanly/ do it without rewriting the API.
That being said, we allow shareWith to be empty (as for link shares), only if you specify some mails as attributes. This also allow to create a share wthout sending the mails straight away and specify the recipients as a later step.
It's a cleaner UX as you can enter emails OR manually copy the link with token, which we would not have without creating the share.

TLDR:
create share > get token > copy link or edit share to set recipients

skjnldsv:
My fix would be to set the `sendMail` param to true by default, as file requests set it to false anyway :thinking: 
https://github.com/nextcloud/server/blob/e089e4081108bf93991130c99012ee095aa873bc/apps/files_sharing/lib/Controller/ShareAPIController.php#L568

apps/files_sharing/src/components/NewFileRequestDialog.vue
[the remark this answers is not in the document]
Fenn-CS: I don't think this is the right place to check for that. 

A file request is (correct me if I am wrong) : Simply a link that you can send to someone to "requests files"...

The means of sending the link can either be:

- Manually copying the link and sending
- Automatic via mail (after you must have already created the link share)

----

Why should the means for sending the "File request" (link) modify change what a file request is? Does the mail not simply the link ?

My thoughts are, if the admin setting must be enforced, it must be enforced at the step where the user is asked to add emails in file request.

The thing already breaks with emails concerning without even knowing whether the user would end up wanting to send the file request link via email or not.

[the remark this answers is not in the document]
Fenn-CS: > If admin allowed email shares, then a file request allow users to add emails and send the notifications
> If admin does NOT, then users can only create a share link file request, and have to manually give the link to the recipients :)

Also this seems pretty mixed up, why should the setting for "allowed email shares" affect whether a link should be sent to users via mail or not? 

It a convenience thing and even if it was not "File drop" (now "File requests") really had nothing to do with email shares?

In other words if we have something that is actually called "e/mail shares" the admin setting that controls or disables that should not control two things because in my opinion a "File requests" is not a mail share. (Because all mail shares are links shares sent via mail, but link shares are not and should mail shares even if they are sent via mail)

Fenn-CS:
> My fix would be to set the sendMail param to true by default, as file requests set it to false anyway 🤔

and what about other things that use the API without specifying what goes to `sendMail`???

susnux:
> and what about other things that use the API without specifying what goes to sendMail???

Agree, we can not change default behavior as this would be a breaking change.

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -719,7 +728,7 @@ public function createShare(
 			}
 
 			// Only share by mail have a recipient
-			if (is_string($shareWith) && $shareType === IShare::TYPE_EMAIL) {
+			if (!empty($shareWith) && $shareType === IShare::TYPE_EMAIL) {
```
susnux, line 731: hm I thought we do not use `empty` for strings?

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -1219,11 +1228,6 @@ public function updateShare(
 		}
 		$this->checkInheritedAttributes($share);
 
-		// Handle mail send
-		if ($sendMail === 'true' || $sendMail === 'false') {
-			$share->setMailSend($sendMail === 'true');
-		}
```
skjnldsv, line 1225: That should still be needed to send the email after creation

Fenn-CS: True

skjnldsv (approved):

apps/files_sharing/lib/Controller/ShareAPIController.php
```diff
@@ -659,7 +659,16 @@ public function createShare(
 		$this->checkInheritedAttributes($share);
 
 		// Handle mail send
-		if ($sendMail === 'true' || $sendMail === 'false') {
+		if (is_null($sendMail)) {
+			// Define a default behavior when sendMail is not provided
+			if ($shareType === IShare::TYPE_EMAIL && strlen($shareWith) !== 0) {
```
provokateurin, line 664:
```suggestion
			if ($shareType === IShare::TYPE_EMAIL && $shareWith !== null && strlen($shareWith) > 0) {
```

$shareWith can be null as well and this would fail without a proper check.

provokateurin (changes requested):

apps/files_sharing/lib/Controller/ShareAPIController.php
Fenn-CS: Actually no, `strlen(null) === 0`  is true.

Fenn-CS: Hmmm, looks like this is deprecated on `php8.2+` though I tested it and it worked.

provokateurin: Ok fine by me then, although it would be better to explicitly check it for the future.

provokateurin (approved):

PR #48417, opened 2024-09-27
[stable29] Fix file item display for users with numerical user IDs

julien-nc:
This is a partial backport of #47287 . I'm not sure it's enough and not sure about the "sharee" part in the `title` function.
Maybe a full backport would be better than this. Wdyt?

I first saw the issue in 29.0.7 (files displayed as shared by their owner if the user ID is numerical) and checked if it was different in stable29. The issue was still there but not really visible because of the `isExternal` check that was wrong.

To reproduce the issue:
* Checkout stable29
* Create a user (database backend works) with a numerical ID (like `123456789`)
* Browse the Files app
* See JS errors like
  image

Check the fix:
* Checkout this branch
* Compile the server assets
* There should be no more error when browsing the files app with the `123456789` user


susnux (approved):
Looks sane, but did not test.
I agree a full backport would make sense, but the sharee title stuff should not be related to the issue (just another issue^^).

kyteinsky (approved):

PR #48418, opened 2024-09-27
feat: configurable request timeout for carddav sync

pabzm:
Big federated setups may need a longer timeout, which they now can configure.

This is motivated by a debugging session, in which we had to raise the timeout but could do so only via a patch.

The default timeout length doesn't change with this PR.

## Checklist

### DIFF
--- a/apps/dav/lib/CardDAV/SyncService.php
+++ b/apps/dav/lib/CardDAV/SyncService.php
@@ -10,6 +10,7 @@ namespace OCA\DAV\CardDAV;
 
 use OCP\AppFramework\Db\TTransactional;
 use OCP\AppFramework\Http;
+use OCP\Http\Client\IClient;
 use OCP\Http\Client\IClientService;
 use OCP\IConfig;
 use OCP\IDBConnection;
@@ -155,7 +156,7 @@ class SyncService {
 			'auth' => [$userName, $sharedSecret],
 			'body' => $this->buildSyncCollectionRequestBody($syncToken),
 			'headers' => ['Content-Type' => 'application/xml'],
-			'timeout' => $this->config->getSystemValueInt('carddav_sync_request_timeout', 30)
+			'timeout' => $this->config->getSystemValueInt('carddav_sync_request_timeout', IClient::DEFAULT_REQUEST_TIMEOUT)
 		];
 
 		$response = $client->request(
--- a/apps/dav/lib/CardDAV/SyncService.php
+++ b/apps/dav/lib/CardDAV/SyncService.php
@@ -11,6 +11,7 @@ namespace OCA\DAV\CardDAV;
 use OCP\AppFramework\Db\TTransactional;
 use OCP\AppFramework\Http;
 use OCP\Http\Client\IClientService;
+use OCP\IConfig;
 use OCP\IDBConnection;
 use OCP\IUser;
 use OCP\IUserManager;
@@ -33,13 +34,15 @@ class SyncService {
 	private Converter $converter;
 	protected string $certPath;
 	private IClientService $clientService;
+	private IConfig $config;
 
 	public function __construct(CardDavBackend $backend,
 		IUserManager $userManager,
 		IDBConnection $dbConnection,
 		LoggerInterface $logger,
 		Converter $converter,
-		IClientService $clientService) {
+		IClientService $clientService,
+		IConfig $config) {
 		$this->backend = $backend;
 		$this->userManager = $userManager;
 		$this->logger = $logger;
@@ -47,6 +50,7 @@ class SyncService {
 		$this->certPath = '';
 		$this->dbConnection = $dbConnection;
 		$this->clientService = $clientService;
+		$this->config = $config;
 	}
 
 	/**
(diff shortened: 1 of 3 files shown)


config/config.sample.php
```diff
@@ -340,6 +340,11 @@
  */
 'davstorage.request_timeout' => 30,
 
+/**
+ * The timeout in seconds for synchronizing address books, e.g. federated system address books (as run by `occ federation:sync-addressbooks`).
+ */
```
joshtrichards, line 345:
```suggestion
/**
 * The timeout in seconds for synchronizing address books, e.g. federated system address books (as run by `occ federation:sync-addressbooks`).
 * 
 * Defaults to ``30`` seconds
 */
```

apps/dav/lib/CardDAV/SyncService.php
```diff
@@ -121,7 +125,8 @@ protected function requestSyncReport(string $url, string $userName, string $addr
 			'auth' => [$userName, $sharedSecret],
 			'base_uri' => $url,
 			'body' => $this->buildSyncCollectionRequestBody($syncToken),
-			'headers' => ['Content-Type' => 'application/xml']
+			'headers' => ['Content-Type' => 'application/xml'],
+			'timeout' => $this->config->getSystemValueInt('carddav_sync_request_timeout', 30)
```
joshtrichards, line 129: Maybe avoid setting it if not specified, otherwise we end up with another magic number spot here.

tcitworld: You can use the existing default constant `RequestOptions::TIMEOUT`:
https://github.com/nextcloud/server/blob/4f88123d2ba649fb6fe2df3d069a297c101dd019/lib/private/Http/Client/Client.php#L56

tcitworld: The 30 is already used as default value for the client. This could be set as a constant in `OCP\Http\IClient`.
https://github.com/nextcloud/server/blob/4f88123d2ba649fb6fe2df3d069a297c101dd019/lib/private/Http/Client/Client.php#L56

pabzm: > Maybe avoid setting it if not specified

I would find it confusing if there'd be an option in the sample config, whose default value maybe wouldn't apply if it's not set. (Currently the default for this setting is the same as the general HTTP client request timeout, but that might change.) Therefore I'd prefer to actually set the value. It's at least findable this way.

> The 30 is already used as default value for the client. This could be set as a constant in OCP\Http\IClient.

I would on the contrary argue that it makes sense to have a longer default timeout for the federation sync requests (as opposed to general HTTP requests), but didn't dare (yet) to change the factual default in this PR. Therefore I would rather not put it in the `Client`.

tcitworld: Sorry I wasn't clear: I suggested to make the existing 30 in `Client` a constant, so that the default value if `carddav_sync_request_timeout` is not set can be used as `Client::DEFAULT_TIMEOUT` instead of just the `30` magic number.

If you want to increase the fallback for this specific case, then maybe a system config setting isn't needed in the first place, and simply set this value here to something higher?

pabzm: I would like to have this configurable, to help debugging without the need for patches. From my point of view the default could be higher than 30s, but I feared that to be a controversial or even breaking change.

kesselb: > Sorry I wasn't clear: I suggested to make the existing 30 in Client a constant, so that the default value if carddav_sync_request_timeout is not set can be used as Client::DEFAULT_TIMEOUT instead of just the 30 magic number.

Sounds good, mind sending a pull request? ;)

tcitworld: Yup, see https://github.com/nextcloud/server/pull/48639

tcitworld, line 129:
```suggestion
			'timeout' => $this->config->getSystemValueInt('carddav_sync_request_timeout', IClient::DEFAULT_REQUEST_TIMEOUT)
```

(add `use OCP\Http\Client\IClient;` as well)

tcitworld (dismissed):

apps/dav/lib/CardDAV/SyncService.php
kesselb: We were thinking of backporting this. To save time and skip another review round, we can just do the constant migration in a follow-up.

miaulalala (commented):
```ci
1) OCA\DAV\Tests\unit\CardDAV\SyncServiceTest::testEmptySync
ArgumentCountError: Too few arguments to function OCA\DAV\CardDAV\SyncService::__construct(), 6 passed in /home/runner/work/server/server/apps/dav/tests/unit/CardDAV/SyncServiceTest.php on line 62 and exactly 7 expected
```

lmk if you need help with the tests

kesselb (approved):

SebastianKrupinski (approved):

PR #48440, opened 2024-09-28
refactor(theming): Reduce duplicated code in `BackgroundService`

susnux:
## Summary

Some functions need a logged-in user so we can only assert a user is logged in on runtime of those functions, but instead of repeating that check we can just use a helper function.

## Checklist


provokateurin (approved):

come-nc (approved):
I want the same thing for controllers please.

Issue #47682, opened 2024-09-02
Chunked upload performance improvements


provokateurin:
# Motivation

Chunked upload is useful in many cases, but also slower than uploading the entire file directly.
Overall it is slower because multiple network requests have to be made.
The speed of requests varies during their lifetime, as it is slower at the beginning and then flattens out at the maximum over time.
The smaller the requests are and the more we make, the worse performance penalty.
Thus to increase chunked upload speed the size of chunks should be increased.

A single upload using cURL is the upper limit of the possible upload speed for any configuration and upload method.
A chunked upload with the chunk size equal to or greater than the file size represents the upper limit for chunked uploads as it only uploads a single chunk.
While reaching the former would be nice, only the latter is achievable (without general performance improvements in WebDAV regardless of the maximum chunk size) and thus represents the theoretical goal.

# Testing methodology

## Input

```
dd if=/dev/random of=1G.bin bs=1G count=1
```

## Scenarios

All tests are running on a local instance using the PHP standalone web server with 10 workers and no extra apps enabled.
The machine has a Ryzen 5 5800X (8 threads, 16 cores), 48GB RAM and a NVMe M.2 Samsung 980 SSD 1TB.
Hardware should not be a bottleneck on this setup and external networking can not have an effect either.

### 1. cURL single upload

Take the `Real` timing value.

```
time curl -X PUT "http://localhost:8080/remote.php/webdav/1G.bin" -u admin:admin --upload-file 1G.bin
```

Runs:
5.412s
5.223s
5.100s
Average: 5.245s

Note: I once saw an outlier that only took about 4.7s, but this never happened again.

### Chunked upload via browser

Open Firefox Devtools and filter network requests by `dav/uploads`.
Upload `1G.bin` via web interface.
Take `Started` of first (`MKCOL`) and the `Downloaded` of the last (`MOVE`) request and subtract them (See the `Timings` tab of each request).
This includes some constant overhead for the `MKCOL` and `MOVE` requests which is not relevant for comparing chunked upload timing results as they all have the same overhead, but when comparing to the cURL scenario it accurately measures the overall time for the upload process.

According to https://firefox-source-docs.mozilla.org/devtools-user/network_monitor/throttling/index.html "Wi-Fi" throttling means a maximum speed of 15 Mbps.
Sadly this is the "fastest" speed one can select for throttling and there is no way to set a custom speed.
[...]

PR #48461, opened 2024-09-30
perf(files): Adjust default chunk size for chunked upload to 100MiB

provokateurin:
## Summary

Step 1 of https://github.com/nextcloud/server/issues/47682

## Checklist

### DIFF
--- a/apps/files/lib/App.php
+++ b/apps/files/lib/App.php
@@ -44,7 +44,7 @@ class App {
 	public static function extendJsConfig($settings): void {
 		$appConfig = json_decode($settings['array']['oc_appconfig'], true);
 
-		$maxChunkSize = (int)Server::get(IConfig::class)->getAppValue('files', 'max_chunk_size', (string)(10 * 1024 * 1024));
+		$maxChunkSize = (int)Server::get(IConfig::class)->getAppValue('files', 'max_chunk_size', (string)(100 * 1024 * 1024));
 		$appConfig['files'] = [
 			'max_chunk_size' => $maxChunkSize
 		];


szaimen (approved):

joshtrichards:
If we do this at 100 (rather just under it, like 99 I suppose?), it's going to hit installations using the Web client via Cloudflare immediately with breakage. This already happened with the Desktop client and is why we have nextcloud/desktop#4278 as a top-3 most reacted to bug at the moment (with solutions proposed such as nextcloud/desktop#5813 and nextcloud/desktop#4826).

At a minimum, we need to document this as a breaking change IMO because people are going to suddenly have problems with our Web client too and therefore will need to figure out how to change the `max_chunk_size`.

Even if documented, we're going to be spending time fielding false bug reports (but legitimate feedback IMO) + hear about it over and over again on the forum.

I want to increase this value too, but we need to decide whether it's acceptable to deal with the headaches introduced before we have the dynamic chunk size stuff in place. Or choose a choose a pragmatic intermediate path I guess like going with an odd number like 99.

I don't like accommodating for an arbitrary outside vendor, but we can't avoid reality. The two biggest external limits are:

- Cloudflare (must be <100MiB) <-- impossible for user to adjust via CF
- Apache (must be <1GiB) <-- possible to adjust and we document it as of nextcloud/documentation#10989

provokateurin:
@joshtrichards Thanks for the hint, so 99 would actually be the limit we should chose to avoid problems? That would at least make sense if Cloudflare limits to 100 and the little overhead each HTTP request has.

joshtrichards:
I just went back and reviewed things again more closely. 100 may be fine actually. The 99 was in my brain to be conservative. :) I don't see any substantial evidence we had any problems until nextcloud/desktop#3600 was merged and released via 3.4.x (which bumped 100->1000 so I guess all was good right at 100).

provokateurin:
Yeah I also just found https://developers.cloudflare.com/workers/platform/limits/#request-limits and it says that the body size is limited to 100MB and not the overall size. I'm not sure if they really mean MB or MiB, as with the former the actual limit in MiB would be around 95.37MiB.
If we find problems later we can fix them, hopefully that happens before final releases (but there is plenty of time until 31).

joshtrichards (approved):

susnux:
@provokateurin how did this influence performance?
Because this can lead to a bad default UX, 100MB is rather huge. Width average upstream it takes a minute per chunk.
This is a rather long time and prone to failed tries, so e.g. from mobile network if that chunk failed you need to wait another minute for the retry.

If the performance improvement is not significant I would recommend to reduce this back to something like 25MiB or 50MiB and let administrators increase the value if they know their network conditions are good. But for the majority I think we need to keep slow network in mind.

provokateurin:
@susnux see the benchmark results from https://github.com/nextcloud/server/issues/47682

PR #48485, opened 2024-10-01
[taskprocessing] Include error message in tasks returned by the API

julien-nc:
This will allow the frontend to display it in a few places.

### DIFF
--- a/lib/public/TaskProcessing/Task.php
+++ b/lib/public/TaskProcessing/Task.php
@@ -266,6 +266,7 @@ final class Task implements \JsonSerializable {
 			'appId' => $this->getAppId(),
 			'input' => $this->getInput(),
 			'output' => $this->getOutput(),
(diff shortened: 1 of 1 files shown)


lib/public/TaskProcessing/Task.php
```diff
@@ -266,6 +266,7 @@ final public function jsonSerialize(): array {
 			'appId' => $this->getAppId(),
 			'input' => $this->getInput(),
 			'output' => $this->getOutput(),
+			'error_message' => $this->getErrorMessage(),
```
marcelklehr, line 269: Good idea! You'll need to update the psalm annotation of this method as well as the open api response definition of the taskprocessing controller

kyteinsky (approved):

marcelklehr (changes requested):
Could be a security risk

PR #48111, opened 2024-09-16
fix: Move storage constructor to specific interface

come-nc:
* Resolves: #32709

## Summary

That allows Wrappers to use DI and not care about the constructor

## Checklist

- Code is [properly formatted](https://docs.nextcloud.com/server/latest/developer_manual/digging_deeper/continuous_integration.html#linting)
- [Sign-off message](https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md) is added to all commits
- [x] Tests ([unit](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#unit-tests), [integration](https://docs.nextcloud.com/server/latest/developer_manual/app_development/tutorial.html#integration-tests), api and/or acceptance) are included
- [x] Screenshots before/after for front-end changes
- [x] Documentation ([manuals](https://github.com/nextcloud/documentation/) or wiki) has been updated or is not required
- [x] [Backports requested](https://github.com/nextcloud/backportbot/#usage) where applicable (ex: critical bugfixes)

PR #48487, opened 2024-10-01
refactor(Encryption): Migrate to strong types

provokateurin:
## Summary

I noticed that `if (!$this->storage->file_exists($this->internalPath)) {` was called with `null` as `$internalPath` which is not allowed.
Added strong types and fixed the tests which were giving wrong/incomplete inputs.

## Checklist

### DIFF
--- a/lib/private/Files/Storage/Wrapper/Encryption.php
+++ b/lib/private/Files/Storage/Wrapper/Encryption.php
@@ -29,76 +29,32 @@ use Psr\Log\LoggerInterface;
 class Encryption extends Wrapper {
 	use LocalTempFileTrait;
 
-	/** @var string */
-	private $mountPoint;
-
-	/** @var \OC\Encryption\Util */
-	private $util;
-
-	/** @var \OCP\Encryption\IManager */
-	private $encryptionManager;
-
-	private LoggerInterface $logger;
-
-	/** @var string */
-	private $uid;
-
-	/** @var array */
-	protected $unencryptedSize;
-
-	/** @var \OCP\Encryption\IFile */
-	private $fileHelper;
-
-	/** @var IMountPoint */
-	private $mount;
-
-	/** @var IStorage */
-	private $keyStorage;
-
-	/** @var Update */
-	private $update;
-
-	/** @var Manager */
-	private $mountManager;
-
-	/** @var array remember for which path we execute the repair step to avoid recursions */
-	private $fixUnencryptedSizeOf = [];
-
-	/** @var ArrayCache */
-	private $arrayCache;
-
+	private string $mountPoint;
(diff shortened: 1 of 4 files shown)


provokateurin:
Wrapper\Encryption has similar troubles, gonna fix that as well.

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -81,7 +56,7 @@ public function __construct(
 		?Util $util = null,
 		?LoggerInterface $logger = null,
 		?IFile $fileHelper = null,
-		$uid = null,
+		?string $uid = null,
```
artonge, line 59: Good occasion to also migrate to property promotion?

```suggestion
		private ?string $uid = null,
```

artonge (approved):

lib/private/Files/Storage/Wrapper/Encryption.php
provokateurin: Good point, I missed that

lib/private/Files/Storage/Wrapper/Encryption.php
```diff
@@ -29,76 +29,32 @@
...
-		?ArrayCache $arrayCache = null,
+		private ?IManager $encryptionManager = null,
+		private ?Util $util = null,
+		private ?LoggerInterface $logger = null,
+		private ?IFile $fileHelper = null,
+		private ?string $uid = null,
+		private ?IStorage $keyStorage = null,
+		private ?Update $update = null,
+		private ?Manager $mountManager = null,
+		private ?ArrayCache $arrayCache = null,
```
come-nc, line 54: You can remove all those `?` and `= null` now thanks to https://github.com/nextcloud/server/pull/48111, that was the point.

provokateurin: It's not possible here because Wrapper still has this constructor:
https://github.com/nextcloud/server/blob/363ba6b0f5b99583e789695f19dcd96db00508ef/lib/private/Files/Storage/Wrapper/Wrapper.php#L39
So we need to use parameters with default values which in turn means they need to be nullable.

come-nc: What? No, you do not have to have the same constructor as the parent class.

provokateurin: Ah you're right: https://www.php.net/manual/en/language.oop5.basic.php#language.oop.lsp
IntelliJ just flagged it and I didn't think about it again.
IMO it invalidates the Liskov substitution principle, but for constructors it kinda makes sense.

come-nc (approved):

PR #48508, opened 2024-10-01
fix(caldav): add missing handlers

miaulalala:
* Resolves: # 

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/dav/lib/CalDAV/WebcalCaching/Connection.php
+++ b/apps/dav/lib/CalDAV/WebcalCaching/Connection.php
@@ -9,15 +9,11 @@ declare(strict_types=1);
 namespace OCA\DAV\CalDAV\WebcalCaching;
 
 use Exception;
-use GuzzleHttp\HandlerStack;
-use GuzzleHttp\Middleware;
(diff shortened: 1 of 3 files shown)


apps/dav/lib/CalDAV/WebcalCaching/Connection.php
```diff
@@ -31,107 +27,82 @@ public function __construct(
...
 		$subscriptionId = $subscription['id'];
 		$url = $this->cleanURL($subscription['source']);
 		if ($url === null) {
 			return null;
 		}
 
-		try {
-			$params = [
-				'allow_redirects' => [
-					'redirects' => 10
```
tcitworld, line 67: No longer accepts redirects?

kesselb: Redirects should still work as they are enabled by default with a maximum of 5 redirects.

Resolving the final location does not work without a middleware, and therefore we are skipping it for now.

provokateurin (approved):

kesselb (approved):

PR #48522, opened 2024-10-02
fix(QueryBuilder): Account for aliases in output columns

### LINKED ISSUE #48521: QueryBuilder output columns incorrect

### Bug description

Use master

### Steps to reproduce

1. Use master
2. try to update your SSE password
 
```
 "File": "/var/www/html/lib/private/DB/QueryBuilder/QueryBuilder.php",
    "Line": 217,
    "message": "More than 30 characters for an output column name are not allowed on Oracle.",
    "query": "SELECT `c`.*, `r`.`id` as `node_rel_id`, `r`.`node_id`, `r`.`node_type`, `r`.`permissions`, `p`.`id` as `page_id`, `p`.`page_type`, `pc`.`id` as `content_id`, `pc`.`order`, `n`.`display_mode` as `display_mode_default`, `s`.`id` as `share_id`, `s`.`receiver`, `s`.`receiver_type`, `n2`.`display_mode` FROM `*PREFIX*tables_contexts_context` `c` LEFT JOIN `*PREFIX*tables_shares` `s` ON (`c`.`id` = `s`.`node_id`) AND (`s`.`node_type` = :dcValue6) AND ((`s`.`sender` = :dcValue1) OR ((`s`.`receiver_type` = :dcValue2) AND (`s`.`receiver` = :dcValue3)) OR ((`s`.`receiver_type` = :dcValue4) AND (`s`.`receiver` IN (:dcValue5)))) LEFT JOIN `*PREFIX*tables_contexts_page` `p` ON `c`.`id` = `p`.`context_id` LEFT JOIN `*PREFIX*tables_contexts_rel_context_node` `r` ON `c`.`id` = `r`.`context_id` LEFT JOIN `*PREFIX*tables_contexts_navigation` `n2` ON (`s`.`id` = `n2`.`share_id`) AND (`n2`.`user_id` = :dcValue8) LEFT JOIN `*PREFIX*tables_contexts_navigation` `n` ON (`s`.`id` = `n`.`share_id`) AND (`n`.`user_id` = :dcValue9) LEFT JOIN `*PREFIX*tables_contexts_page_content` `pc` ON `p`.`id` = `pc`.`page_id` WHERE ((`owner_id` = :dcValue7) OR (`s`.`receiver` IS NOT NULL)) AND ((`pc`.`node_rel_id` = `r`.`id`) OR (`pc`.`node_rel_id` IS NULL)) ORDER BY `pc`.`order` ASC",
    "columns": [
      "display_mode as display_mode_default"
    ],
```

### Expected behavior

No error.

### Nextcloud Server version

master

### Operating system

None

### PHP engine version

None

### Web server

None

### Database engine version

None

None

None

- [ ] LDAP/ Active Directory
- [ ] SSO - SAML
- [ ] Other

### Configuration report

_No response_

### List of activated Apps

_No response_

### Nextcloud Signing status

_No response_

### Nextcloud Logs

_No response_

### Additional info

_No response_

provokateurin:
* Resolves: https://github.com/nextcloud/server/issues/48521

## Summary

https://github.com/nextcloud/server/pull/48361 added error logs when the output column is too long. https://github.com/nextcloud/server/issues/48521 shows the output columns weren't extracted correctly as it contained `display_mode as display_mode_default` which is actually fine because `display_mode_default` is only 20 chars.

## Checklist

### DIFF
--- a/lib/private/DB/QueryBuilder/QueryBuilder.php
+++ b/lib/private/DB/QueryBuilder/QueryBuilder.php
@@ -582,6 +582,9 @@ class QueryBuilder implements IQueryBuilder {
 			if (is_array($column)) {
 				$this->addOutputColumns($column);
 			} elseif (is_string($column) && !str_contains($column, '*')) {
+				if (str_contains(strtolower($column), ' as ')) {
+					[, $column] = preg_split('/ as /i', $column);
+				}
 				if (str_contains($column, '.')) {
 					[, $column] = explode('.', $column);
 				}
@@ -591,14 +594,7 @@ class QueryBuilder implements IQueryBuilder {
 	}
 
 	public function getOutputColumns(): array {
-		return array_unique(array_map(function (string $column) {
-			if (str_contains($column, '.')) {
-				[, $column] = explode('.', $column);
-				return $column;
-			} else {
-				return $column;
-			}
-		}, $this->selectedColumns));
+		return array_unique($this->selectedColumns);
 	}
 
 	/**


lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -591,14 +594,7 @@ private function addOutputColumns(array $columns): void {
 	}
 
 	public function getOutputColumns(): array {
-		return array_unique(array_map(function (string $column) {
```
nickvergessen, line 594: I fear this preparation is required by the Sharding query builder?

nickvergessen: Maybe add a second getter to be on the save side?

provokateurin: `$selectedColumns` is private so the only way to access them is through this getter.
`getOutputColumns()` is also never called in any sharding logic afaict.

I justed removed this logic because it was duplicate with the logic executed before adding a column and therefore useless.

nickvergessen: > getOutputColumns() is also never called in any sharding logic afaict.

I checked this before I commented of course:
https://github.com/nextcloud/server/blob/62f8b6517f4492b220ebd9df415f2b134735768b/lib/private/DB/QueryBuilder/Partitioned/PartitionQuery.php#L54-L73

provokateurin: Ah ok, I didn't know this counted for sharded as well.

Still I don't think there is a problem with my fix.

nickvergessen: `$nullResult` before was:
```
'column' => null,
```

And now is
```
't1.column' => null,
```

nickvergessen: Nvm I missed you are modifying `addOutputColumns(`

lib/private/DB/QueryBuilder/QueryBuilder.php
```diff
@@ -582,6 +582,9 @@ private function addOutputColumns(array $columns): void {
 			if (is_array($column)) {
 				$this->addOutputColumns($column);
 			} elseif (is_string($column) && !str_contains($column, '*')) {
+				if (str_contains(strtolower($column), ' as ')) {
```
nickvergessen, line 585: But that being said, I guess this breaks aliases in [QueryBuilder/Partitioned/PartitionQuery.php](https://github.com/nextcloud/server/blob/62f8b6517f4492b220ebd9df415f2b134735768b/lib/private/DB/QueryBuilder/Partitioned/PartitionQuery.php#L54-L73)

provokateurin: But then it would already be broken with `selectAlias()` which only adds the alias part to the output columns.

nickvergessen: roger, tests also seem green.
Maybe @icewind1991 can give it a quick local test

come-nc (approved):
Looks sane, did not test.

lib/private/DB/QueryBuilder/QueryBuilder.php
icewind1991: Yes, `getOutputColumns` returning the aliases is the expected behavior.

icewind1991 (approved):
Ideally we should discourage apps from doing aliases like this instead of using `selectAlias`

PR #48532, opened 2024-10-02
fix: do not reduce count for subadmins if they are members of group

Antreesy:
* Follow-up to #48437

## Summary

Counter on sidebar doesn't reflect the actual number of group members, if subadmin is a member too.
For each group, counter is reduced by 1, and for 'All accounts', counter is reduced by amount of groups subadmin is in.

There is a comment:
> we prevent subadmins from looking up themselves so we lower the count of the groups he belongs to

but I don't see a point of it, since we fetch and show yourself in the list anyway, so just removing inconsistency

Before | After
-- | --
image | image
_Subadmin: 'alice', member of 'test group' and 'test group 2'_

## TODO

- [ ] ...

## Checklist

### DIFF
--- a/apps/settings/lib/Controller/UsersController.php
+++ b/apps/settings/lib/Controller/UsersController.php
@@ -154,12 +154,6 @@ class UsersController extends Controller {
 				foreach ($groups as $key => $group) {
 					// $userCount += (int)$group['usercount'];
 					$groupsIds[] = $group['id'];
-					// we prevent subadmins from looking up themselves
-					// so we lower the count of the groups he belongs to
-					if (array_key_exists($group['id'], $userGroups)) {
-						$groups[$key]['usercount']--;
-						$userCount -= 1; // we also lower from one the total count
-					}
 				}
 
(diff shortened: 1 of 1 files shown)


apps/settings/lib/Controller/UsersController.php
```diff
@@ -156,10 +156,10 @@ public function usersList(): TemplateResponse {
...
 					// we prevent subadmins from looking up themselves
 					// so we lower the count of the groups he belongs to
-					if (array_key_exists($group['id'], $userGroups)) {
-						$groups[$key]['usercount']--;
-						$userCount -= 1; // we also lower from one the total count
-					}
+					// if (array_key_exists($group['id'], $userGroups)) {
+					// $groups[$key]['usercount']--;
+					// $userCount -= 1; // we also lower from one the total count
+					// }
```
provokateurin, line 162: Just get rid off the whole thing, I agree it doesn't make much sense to do these subtractions.

provokateurin (changes requested):

provokateurin (approved):

blizzz (approved):

PR #48536, opened 2024-10-02
fix: Vue app names

artonge:
- This improves the debugging experience in the vue dev tool.
  - Actually also fixes the files reminder action
- Also includes two minor fixes to clear our two errors
  - Label for `NcSelect` in `NcSharingInput`
  - Usage of `NcListItem` in `Version`. Will need usage of `actions-aria-label` (https://github.com/nextcloud-libraries/nextcloud-vue/pull/6128/files)


susnux (approved):

apps/files_reminders/src/components/SetCustomReminderModal.vue
```diff
@@ -93,16 +93,16 @@ export default Vue.extend({
...
 
-		fileName(): string {
-			return this.node.basename
+		fileName(): string|undefined {
+			return this.node?.basename
 		},
 
 		name() {
-			return t('files_reminders', 'Set reminder for "{fileName}"', { fileName: this.fileName })
+			return this.node ? t('files_reminders', 'Set reminder for "{fileName}"', { fileName: this.fileName }) : ''
```
Pytal, line 105:
```suggestion
			return this.filename ? t('files_reminders', 'Set reminder for "{fileName}"', { fileName: this.fileName }) : ''
```
For less cognitive overhead ;)

Pytal (approved):

apps/files_reminders/src/components/SetCustomReminderModal.vue
artonge: Yes, makes sense :)

apps/files_versions/src/components/Version.vue
```diff
@@ -133,9 +136,8 @@ import { joinPaths } from '@nextcloud/paths'
 import { loadState } from '@nextcloud/initial-state'
 import { Permission, formatFileSize } from '@nextcloud/files'
 import { translate as t } from '@nextcloud/l10n'
-import { defineComponent } from 'vue'
-
 import axios from '@nextcloud/axios'
+import moment from '@nextcloud/moment'
```
Pytal, line 140: We should move away from moment, no?

artonge: When ?

susnux: Yes for new stuff do not use Moment, but as Pytal mentioned could be done in a follow up

skjnldsv (approved):

apps/files_versions/src/components/Version.vue
artonge: Thanks for the suggestion. Done.

apps/files_versions/src/components/Version.vue
```diff
@@ -225,6 +227,10 @@ export default defineComponent({
 			return label
 		},
 
+		versionHumanExplicitDate(): string {
+			return useFormatDateTime(this.version.mtime, { relativeTime: false, format: { dateStyle: 'long', timeStyle: 'medium' } }).formattedTime
```
Pytal, line 231: I believe composables should only be called in setup() 🙈

susnux: > I believe composables should only be called in setup() 🙈

Yes, this might work but is not guaranteed as some calls might require setup-context.

susnux: Just add:

```js
    setup(props) {
        const mTime = computed(() => props.version.mtime)
        const { formattedTime: versionHumanExplicitDate } = useFormatDateTime(mTime, { relativeTime: false, format: { dateStyle: 'long', timeStyle: 'medium' } })
        return {
            versionHumanExplicitDate,
        }
```

artonge: So not really usable in our case

artonge: Reverted to moment

artonge: Ah, I missed susnux comment. Too late

PR #48537, opened 2024-10-02
feat(preview): add command to cleanup preview folder

tcitworld:
## Summary

When changing preview parameters (size, format, …), one can simply delete the `appdata_xxx/preview` folder and run `files:scan-app-data` in order to retrigger updated preview recreation because files are absent.

However, that's not possible when using ObjectStorage as primary storage, as there's no concept of "preview folder".

This command allows to simply delete and recreate the preview folder properly from Nextcloud, so that the abstraction levels handle all the subfolder & file deleting part.

Possible concerns : deleting all preview files from object storage in a single command might take a while, so even if we don't have a timeout because this is a command, it might be nice to have some kind of a warning before. Possibly a confirmation step.

## Checklist

### DIFF
--- a/core/Command/Preview/Cleanup.php
+++ b/core/Command/Preview/Cleanup.php
@@ -1,0 +1,88 @@
+<?php
+
+declare(strict_types=1);
+
+/**
+ * SPDX-FileCopyrightText: 2024 Nextcloud GmbH and Nextcloud contributors
+ * SPDX-License-Identifier: AGPL-3.0-or-later
+ */
+
+namespace OC\Core\Command\Preview;
+
+use OC\Core\Command\Base;
+use OCP\Files\Folder;
+use OCP\Files\IRootFolder;
+use OCP\Files\NotFoundException;
+use OCP\Files\NotPermittedException;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class Cleanup extends Base {
+
+	public function __construct(
+		private IRootFolder     $rootFolder,
+		private LoggerInterface $logger,
+	) {
+		parent::__construct();
+	}
+
+	protected function configure(): void {
+		$this
+			->setName('preview:cleanup')
+			->setDescription('Removes existing preview files');
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		try {
+			$appDataFolder = $this->rootFolder->get($this->rootFolder->getAppDataDirectoryName());
+
+			if (!$appDataFolder instanceof Folder) {
+				$this->logger->error("Previews can't be removed: appdata is not a folder");
+				$output->writeln("Previews can't be removed: appdata is not a folder");
+				return 1;
+			}
+
+			/** @var Folder $previewFolder */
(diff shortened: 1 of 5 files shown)


core/Command/Preview/Cleanup.php
```diff
@@ -0,0 +1,91 @@
...
+use OCP\Files\IRootFolder;
+use OCP\Files\NotFoundException;
+use OCP\Files\NotPermittedException;
+use OCP\IConfig;
+use Psr\Log\LoggerInterface;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class Cleanup extends Command {
```
come-nc, line 22:
```suggestion
class Cleanup extends Base {
```
It’s in core so it can use Base, no?

core/Command/Preview/Cleanup.php
```diff
@@ -0,0 +1,91 @@
...
+	protected function configure(): void {
+		$this
+			->setName('preview:cleanup')
+			->setDescription('Removes existing preview files');
+	}
+
+	protected function execute(InputInterface $input, OutputInterface $output): int {
+		$instanceId = $this->config->getSystemValueString('instanceid');
+		try {
+			$appDataFolder = $this->rootFolder->get("appdata_$instanceId");
```
come-nc, line 41: You should use `IRootFolder::getAppDataDirectoryName`.
There is even IAppData::getFolder, but you get an `ISimpleFolder`, not sure if that’s enough for your usecase, it might be though? There is a `delete` method.

tests/Core/Command/Preview/CleanupTest.php
```diff
@@ -0,0 +1,172 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Test\TestCase;
+
+class CleanupTest extends TestCase {
+	private IConfig|MockObject $config;
+	private IRootFolder|MockObject $rootFolder;
+	private LoggerInterface|MockObject $logger;
+	private InputInterface|MockObject $input;
+	private OutputInterface|MockObject $output;
```
come-nc, line 25:
```suggestion
	private IConfig&MockObject $config;
	private IRootFolder&MockObject $rootFolder;
	private LoggerInterface&MockObject $logger;
	private InputInterface&MockObject $input;
	private OutputInterface&MockObject $output;
```

core/Command/Preview/Cleanup.php
tcitworld: Missed the first one, thanks.
Noticed the second one, but in any case I still need the IRootFolder to get the `appdata_xxx` folder and recreate the `preview` folder.

come-nc (approved):

core/Command/Preview/Cleanup.php
come-nc: No I think IAppData::getFolder would get you the folder and you can call get('previews') on that and do the same thing, but whatever, your current solution works fine.

tcitworld: I tried that and for some reason I can't DI `IAppData`, only `IAppDataFactory`, so I didn't pursue.

provokateurin (approved):

PR #48538, opened 2024-10-02
fix: Return correct list of managers for a user

### LINKED ISSUE #47642: [Bug]: accounts -> error message as group manager when you are not allowed to fetch the line manager details

### Bug description

[Screencast from 2024-08-30 15-51-30.webm](https://github.com/user-attachments/assets/47d20e04-721a-42d7-b231-357a6dcfb6e8)

### Steps to reproduce

1. Have Admin, Alice (group manager for A-Team), Bob (member of A-Team), Jim (member of A-Team), Tom (**not** a member of A-Team)
2. Admin: Set Tom as line manager for Bob
3. Alice: Login and open the account section
4. See a warning 

### Expected behavior

No warning

Pytal:
## Summary

Fix https://github.com/nextcloud/server/issues/47642

Subadmins may not have access to certain users so we only return the uids for users that are accessible.

## Checklist

### DIFF
--- a/apps/provisioning_api/lib/Controller/AUserData.php
+++ b/apps/provisioning_api/lib/Controller/AUserData.php
@@ -8,7 +8,7 @@ declare(strict_types=1);
  */
 namespace OCA\Provisioning_API\Controller;
 
-use OC\Group\Manager;
+use OC\Group\Manager as GroupManager;
 use OC\User\Backend;
 use OC\User\NoUserException;
 use OC_Helper;
@@ -22,7 +22,6 @@ use OCP\AppFramework\OCSController;
 use OCP\Files\NotFoundException;
 use OCP\Group\ISubAdmin;
 use OCP\IConfig;
-use OCP\IGroupManager;
 use OCP\IRequest;
 use OCP\IUser;
 use OCP\IUserManager;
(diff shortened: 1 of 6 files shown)


lib/private/User/User.php
```diff
@@ -536,13 +556,27 @@ public function setQuota($quota) {
 	}
 
 	public function getManagerUids(): array {
+		$user = $this->userSession->getUser();
```
artonge, line 559: To make it cleared as there is already a lot of reference to user in this file.
```suggestion
		$connectedUser = $this->userSession->getUser();
```

artonge (approved):

come-nc (changes requested):
It feels arbitrary to me to be checking permissions for the logged in user in this method.

What if some code from a public page needs to list the managers from a user?
You should filter this from the calling controller, not in the User class.

apps/provisioning_api/lib/Controller/AUserData.php
```diff
@@ -206,6 +191,37 @@ protected function getUserData(string $userId, bool $includeScopes = false): ?ar
...
 	}
 
+	/**
+	 * @return string[]
+	 */
+	protected function getManagers(IUser $user): array {
+		$currentLoggedInUser = $this->userSession->getUser();
+		$isAdmin = $this->groupManager->isAdmin($currentLoggedInUser->getUID());
+		$isDelegatedAdmin = $this->groupManager->isDelegatedAdmin($currentLoggedInUser->getUID());
+		$isSubAdmin = $this->subAdminManager->isSubAdmin($currentLoggedInUser);
```
come-nc, line 201: I may be wrong but getting these values may be expensive in some cases (like doing an LDAP request to check if the user is an group flagged as admin), so it’s better to only compute them if needed rather than all beforehand. I now I’ve seen the same thing in other places.
Only for performance reasons.

come-nc (approved):

apps/provisioning_api/lib/Controller/AUserData.php
Pytal: Adjusted 🚀

PR #48543, opened 2024-10-02
fix(theming): Fix layout of themes

### LINKED ISSUE #48407: Theme previews / selector looks to crowded
Alignment of screenshots, description and radio buttons looks weird and broken.

The section should not be limited in width and the theme options should take a reasonable width.
For example if there is space then all should look like the "system default" entry.

image

Pytal:
## Summary

Fix https://github.com/nextcloud/server/issues/48407

Before | After
--- | ---
image | image

## Checklist


artonge (approved):

ShGKme:
If I'm not mistaken, @marcoambrosini had some plans here, was discussed when `NcSettingsSection` become force limited

susnux (approved):
Better :)
Yet I think limiting the width is an mistake here, like with the profile settings it should take more horizontal space as 1920px is basically the norm today.

marcoambrosini:
Nice improvement!

marcoambrosini:
> Yet I think limiting the width is an mistake here, like with the profile settings it should take more horizontal space as 1920px is basically the norm today.

Since there can be long paragraphs in the settings, I'd like to limit the width to improve readability

Pytal:
> > Yet I think limiting the width is an mistake here, like with the profile settings it should take more horizontal space as 1920px is basically the norm today.
> 
> Since there can be long paragraphs in the settings, I'd like to limit the width to improve readability

The width is force-limited to 900px already on 30+ but do you think we should limit it more?

marcoambrosini:
I think that's ok @Pytal !

